PR #7, opened 2022-06-05
Use cookiecutter to create new tasks 🍪

NouamaneTazi:


KennethEnevoldsen:
@NouamaneTazi seems like this is outdated so will close it.

PR #18, opened 2022-07-08
Fix task splits

Muennighoff:
If you run multiple tasks it will reuse previous splits which may not exist


NouamaneTazi:
Thanks!

NouamaneTazi (approved):

PR #19, opened 2022-07-08
Update README.md

Muennighoff:
Sometimes there are additional kwargs which would cause the current example to not work


NouamaneTazi:
Good catch! Thanks @Muennighoff

PR #26, opened 2022-07-20
Returning Evaluation results

AmrMKayid:
Return the evaluation results dictionary if `return_resutls` is set to `True`

### DIFF
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -168,7 +168,7 @@ class MTEB:
             logger.info(f"\n# Loading dataset for {task.description['name']}")
             task.load_data()
 
-    def run(self, model, verbosity=1, output_folder="results/result", eval_splits=None, return_resutls=False, **kwargs):
+    def run(self, model, verbosity=1, output_folder="results/result", eval_splits=None, return_results=False, **kwargs):
         """
         Run the evaluation pipeline on the selected tasks.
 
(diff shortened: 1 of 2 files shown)


mteb/evaluation/MTEB.py
```diff
@@ -168,7 +168,7 @@ def load_tasks_data(self):
             logger.info(f"\n# Loading dataset for {task.description['name']}")
             task.load_data()
 
-    def run(self, model, verbosity=1, output_folder="results/result", eval_splits=None, **kwargs):
+    def run(self, model, verbosity=1, output_folder="results/result", eval_splits=None, return_resutls=False, **kwargs):
```
amorisot, line 171:
```suggestion
    def run(self, model, verbosity=1, output_folder="results/result", eval_splits=None, return_results=False, **kwargs):
```

NouamaneTazi:
Apologies for the delay. Thank you for the contribution! I think many users would find this handy, would you like to add a small example of how to use the flag in the README?

Muennighoff:
t doesn't hurt always returning them, so I would remove the kwarg & always return them; What do you think @NouamaneTazi ?

NouamaneTazi:
True, I don't see a problem with that neither. As long as we update the related docs.

Muennighoff:
@AmrMKayid If you could make it the default & update the docstring of `run` that would be amazing!

AmrMKayid:
@NouamaneTazi @Muennighoff Thank you very much for the feedback! I have made it the default and updated the docs :))

NouamaneTazi (approved):

NouamaneTazi:
Amazing! Thanks for the clean PR :)

PR #37, opened 2022-08-04
Fix Mind Reference

Muennighoff:
Two other notes:
- The renaming can create confusion as there exists a test set just that I assume we don't have the labels - I clarified it [here](https://huggingface.co/datasets/mteb/mind_small/discussions/2)
- MIND uses AUC & MRR & NDCG scores, not MAP, see https://msnews.github.io/


NouamaneTazi (approved):

NouamaneTazi:
Which score should be the main one? Can you fix it as well?

Muennighoff:
Well I think we should use the same main score for all tasks in one task category, so sticking with MAP is fine I guess.
Albeit it would be nice to also compute AUC & NDCG scores for MIND. Opened a sep issue for that: https://github.com/embeddings-benchmark/mteb/issues/39

Issue #24, opened 2022-07-10
Negative STS22 scores


Muennighoff:
STS22 scores should not be negative https://competitions.codalab.org/competitions/33835#results

NouamaneTazi:
Can you provide a sample where the score is negative please?

Muennighoff:
```python
!pip install -q mteb
from mteb import MTEB
from sentence_transformers import SentenceTransformer
# Define the sentence-transformers model name
model_name = "average_word_embeddings_komninos"
model = SentenceTransformer(model_name)
evaluation = MTEB(tasks=["STS22"])
evaluation.run(model, output_folder=f"results/{model_name}")
```

PR #41, opened 2022-08-07
Fix/sts22

Muennighoff:
Solves https://github.com/embeddings-benchmark/mteb/issues/24
Goes along with https://huggingface.co/datasets/mteb/sts22-crosslingual-sts/discussions/1

### DIFF
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -18,7 +18,11 @@ class AbsTask(ABC):
         if self.data_loaded:
             return
 
-        self.dataset = datasets.load_dataset(self.description["hf_hub_name"])  # TODO: add split argument
+         # TODO: add split argument
+        self.dataset = datasets.load_dataset(
+            self.description["hf_hub_name"], 
+            revision=self.description["revision"]
+        )
         self.data_loaded = True
 
     @abstractmethod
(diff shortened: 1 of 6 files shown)


scripts/sts22-crosslingual-sts/create_data.py
```diff
@@ -44,6 +48,9 @@
     }
 )
 
+# Invert scores from "from most to least similar" to "from least to most similar"
+df["score"] = df["score"] * -1 + 5
```
NouamaneTazi, line 52: We could use the min and max scores defined in the task description

mteb/abstasks/AbsTask.py
```diff
@@ -18,7 +18,11 @@ def load_data(self, **kwargs):
         if self.data_loaded:
             return
 
-        self.dataset = datasets.load_dataset(self.description["hf_hub_name"])  # TODO: add split argument
+        # TODO: add split argument
+        self.dataset = datasets.load_dataset(
+            self.description["hf_hub_name"], 
+            revision=self.description["revision"]
```
NouamaneTazi, line 24: Should we have a default value for `revision`, or do we make it a required field in all tasks' descriptions? (if the latter, then it's better to add the `revision` field in all the tasks in this same PR)

Muennighoff: Made it optional for now, but I think we should add it for all datasets.

NouamaneTazi (approved):
LGTM

Muennighoff:
We should also merge this PR https://huggingface.co/datasets/mteb/sts22-crosslingual-sts/discussions/1 to make the new dataset with inverted scores the default

Issue #34, opened 2022-08-03
Is SciDocs the same as USEB SciDocs?


Muennighoff:
It's inconvenient to have two datasets with the same case insensitive name: `SciDocs` & `SCIDOCS`. E.g. MacOS is by default case insensitive.

Let's rename SciDocs? I'd propose SciDocsS2S or SciDocsRR instead, but maybe someone has a better idea.

Also do I understand correctly that SciDocs is the same SciDocs as in [useb](https://github.com/UKPLab/useb)? I.e. it includes all the tasks of Cite, Co-Cite, Co-Read, Co-View, see their [paper](https://arxiv.org/pdf/2104.06979.pdf) for details[ (+ recomm).](https://github.com/allenai/scidocs)

cc @loicmagne as I think you added it?

Muennighoff:
Still need to figure out whether it's the same as USEB or not

NouamaneTazi:
Can we replace it with the SciDocs that is in useb, to make sure it's the one used?

Muennighoff:
Based on comparing scores I got from USEB [here](https://arxiv.org/pdf/2202.08904.pdf) & from MTEB/SciDocsRR, they're not the same but in the same ballpark.
I think finding out what the exact differences are isn't worth it, but we should always recommend to run the MTEB versions for all datasets (including STS, BEIR etc).

Closing this one for now, feel free to reopen if you feel otherwise.

PR #43, opened 2022-08-14
Add flag to overwrite results

Muennighoff:
Solves https://github.com/embeddings-benchmark/mteb/issues/31

PR #46, opened 2022-08-26
Fix/scidocs

Muennighoff:
Fixes https://github.com/embeddings-benchmark/mteb/issues/34

Should be merged first: https://github.com/embeddings-benchmark/mteb/pull/43


NouamaneTazi (approved):
LGTM! Could be interesting to clarify if it's the same one from `useb` in a comment or so, as per the question in the original issue.

Muennighoff:
Will try to lo

> LGTM! Could be interesting to clarify if it's the same one from `useb` in a comment or so, as per the question in the original issue.

Will look into this at some point. Leaving the issue open for now, but merging this one already 👍

PR #49, opened 2022-08-26
Fix CQADupstack

Muennighoff:
Fixes https://github.com/embeddings-benchmark/mteb/issues/48

Imo this is the better solution as opposed to automatically running all subsets in case it needs to be split up for compute reasons.
Averaging needs to be done on the user side for now, but we can add a script for that in a follow-up.


NouamaneTazi (approved):

mteb/abstasks/BeIRTask.py
```diff
@@ -31,6 +30,8 @@ def load_data(self, eval_splits=None, **kwargs):
         if eval_splits is None:
             eval_splits = self.description["eval_splits"]
         dataset = self.description["beir_name"]
+        dataset, sub_dataset = dataset.split("/") if "cqadupstack" in dataset else (dataset, None)
```
NouamaneTazi, line 33: we could also remove the hardcoded value in here later on

NouamaneTazi (approved):
LGTM

qmin2:
How about MSMARCO ?
It looks like there are 2 versions, MSMARCO and MSMARCOv2
Should I just average two like CQADupstack to get the score from Table:11 in the paper?

Muennighoff:
> How about MSMARCO ? It looks like there are 2 versions, MSMARCO and MSMARCOv2 Should I just average two like CQADupstack to get the score from Table:11 in the paper?

You should use the **dev** split of MSMARCO only. We did not run MSMARCOv2.
Sorry for not having made this clearer - Maybe there's a good place to put this information 🧐

Issue #40, opened 2022-08-07
Add revision to datasets


Muennighoff:
I'd propose to add the commit hash of the revision to tasks:
```python
from mteb import MTEB
from mteb.abstasks.AbsTaskReranking import AbsTaskReranking
from sentence_transformers import SentenceTransformer


class MindSmallReranking(AbsTaskReranking):
    @property
    def description(self):
        return {
            "name": "MindSmallReranking",
            "hf_hub_name": "mteb/mind_small",
            "description": "Microsoft News Dataset: A Large-Scale English Dataset for News Recommendation Research",
            "reference": "https://www.microsoft.com/en-us/research/uploads/prod/2019/03/nl4se18LinkSO.pdf",
            "type": "Reranking",
            "category": "s2s",
            "eval_splits": ["validation"],
            "eval_langs": ["en"],
            "main_score": "map",
            "revision": "75937953179...",
        }

model = SentenceTransformer("average_word_embeddings_komninos")
evaluation = MTEB(tasks=[MindSmallReranking()])
evaluation.run(model)
```
This is then fed into `load_dataset` via` revision=` & added to the results json file.

This partly addresses https://github.com/embeddings-benchmark/mteb/issues/21

NouamaneTazi:
How would `revision` work with `version`? Wouldn't having both make things more complicated for users to check? (aka asking what version of the library did you use, AND which revision for that task)

Muennighoff:
Yeah there are three versions:
- Revision of the model (Probably entirely on the user side, as we don't implement models here yet)
- Revision of the dataset
- Version of this library (& its dependencies)

If we don't have dataset revisions, the current implementation automatically downloads the latest dataset, so it wouldn't be versioned. As each change to a dataset revision is tracked via a git commit in this repo, using just the version of this library (e.g. a commit string) should be enough. We'll know all dataset revisions at that commit from the code.

NouamaneTazi:
Alright seems good to me. Would be cool to have a `versioning` section in the docs that re-explains this to users.
Thanks 🤗

Muennighoff:
As noted in https://github.com/embeddings-benchmark/mteb/pull/41, this Issue can be closed once a revision field has been added to all MTEB datasets

Muennighoff:
I think we can close, as there's now a revision in all MTEB native datasets. 
For BEIR, I think it's not worth it adding logic to support two revisions / dataset, as it may well change in the future.

PR #72, opened 2022-09-27
Fix/revisions

Muennighoff:
Adds revisions to all datasets except Retrieval fixing https://github.com/embeddings-benchmark/mteb/issues/40

Retrieval is a bit more complicated as we'd need a revision for the corpus & the qrels dataset...


NouamaneTazi (approved):
LGTM :)

PR #79, opened 2022-10-01
Add leaderboard instructions

Muennighoff:


README.md
```diff
@@ -178,6 +178,15 @@ evaluation.run(model)
 
 > **Note:** for multilingual tasks, make sure your class also inherits from the `MultilingualTask` class like in [this](https://github.com/embeddings-benchmark/mteb-draft/blob/main/mteb/tasks/Classification/MTOPIntentClassification.py) example.
 
+## Leaderboard
+
+The MTEB Leaderboard is available [here](https://huggingface.co/spaces/mteb/leaderboard). To submit:
+1. Run your model on MTEB
+2. Format the json files into metadata using the script at `scripts/other/mteb_meta.py`. For example
+`python scripts/other/mteb_meta.py path_to_results_folder`, which will create a `mteb_metadata.md` file.
```
NouamaneTazi, line 186: We could move the file to `scripts/mteb_meta.py`, I think it'll be easier to use

scripts/other/mteb_meta.py
```diff
@@ -0,0 +1,109 @@
...
+                    # All MTEB scores are 0-1, multiply them by 100 for reasons:
+                    # 1) It's easier to visually digest (You need two chars less: "0.1" -> "1")
+                    # 2) Others may multiply them by 100, when building on MTEB making it confusing what the range is
+                    # This happend with Text and Code Embeddings paper (OpenAI) vs original BEIR paper
+                    # 3) It's accepted practice (SuperGLUE, GLUE are 0-100)
+                    sub_score * 100,
+                )
+
+META_STRING += "\n" + MARKER
+with open("./mteb_metadata.md", "w") as f:
```
NouamaneTazi, line 108: This overwrites the previous `mteb_metadata.md`, we would want to warn the user of this, or automatically add a suffix

scripts/other/mteb_meta.py
```diff
@@ -0,0 +1,109 @@
...
+    mteb_desc = (
+        MTEB(tasks=[ds_name.replace("CQADupstackRetrieval", "CQADupstackAndroidRetrieval")])
+        .tasks[0]
+        .description
+    )
+    hf_hub_name = mteb_desc.get("hf_hub_name", mteb_desc.get("beir_name"))
+    if "CQADupstack" in ds_name:
+        hf_hub_name = "BeIR/cqadupstack"
+    mteb_type = mteb_desc.get("type")
+    split = "test"
```
NouamaneTazi, line 69: That means we only report test scores? What if we want to report dev scores as well?

scripts/other/mteb_meta.py
```diff
@@ -0,0 +1,109 @@
...
+TAGS = "tags:"
+MTEB_TAG = "- mteb"
+HEADER = "model-index:"
+MODEL = f"- name: {model_name}"
+RES = "  results:"
+
+META_STRING = "\n".join([MARKER, TAGS, MTEB_TAG, HEADER, MODEL, RES])
+
+
+ONE_TASK = "  - task:\n      type: {}\n    dataset:\n      type: {}\n      name: {}\n      config: {}\n      split: {}\n    metrics:"
```
NouamaneTazi, line 55: `config` should probably be renamed `lang`?

scripts/other/mteb_meta.py
```diff
@@ -0,0 +1,109 @@
...
+                if any([x in sub_metric for x in SKIP_KEYS]):
+                    continue
+                META_STRING += "\n" + ONE_METRIC.format(
+                    f"{metric}_{sub_metric}" if metric != sub_metric else metric,
+                    # All MTEB scores are 0-1, multiply them by 100 for reasons:
+                    # 1) It's easier to visually digest (You need two chars less: "0.1" -> "1")
+                    # 2) Others may multiply them by 100, when building on MTEB making it confusing what the range is
+                    # This happend with Text and Code Embeddings paper (OpenAI) vs original BEIR paper
+                    # 3) It's accepted practice (SuperGLUE, GLUE are 0-100)
+                    sub_score * 100,
```
NouamaneTazi, line 104: Nice

NouamaneTazi (changes requested):
Amazing job! Left some comments :)

scripts/other/mteb_meta.py
Muennighoff: There is no lang - This refers to the dataset config, see https://github.com/huggingface/hub-docs/blame/main/modelcard.md

But you remind me that we should also add the revision kwarg 👍

scripts/other/mteb_meta.py
Muennighoff: I was thinking about this for like half an hour 😂 
These are really important decisions we need to make that are hard to fix later

scripts/other/mteb_meta.py
Muennighoff: Yeah we could also add all other split scores & filter for test scores in the space instead. 
The only downside is splits are not indicated in the sidebar, so it will be confusing if looking at the model card:
Screenshot 2022-10-01 at 22 17 38


Do you think we need other splits for any reason?

NouamaneTazi: Not really, I just thought it shouldnt be that difficult to add, since the library alrdy supports that. But yeah good point about the model card

NouamaneTazi (approved):
LGTM :)

PR #82, opened 2022-10-18
Delete mteb_diagram.png

Muennighoff:
I don't think we need the file here as
- It's alrdy here https://github.com/embeddings-benchmark/mtebscripts/blob/main/plotstables/mteb_diagram.png
- It's only a snapshot in time of MTEB - hopefully it is outdated soon
- Also rmved it form the blog cuz didnt feel like it added a lot of value


NouamaneTazi:
* I agree that we don't need it in two repos
* It gives a good overview of the initial release. Makes it easy for newcomers to understand what we already support
* Same as point 2, but your call ;-)

Issue #83, opened 2022-10-24
No 'validation' split for mteb/sickr-sts: raise KeyError during evaluation


mrzhang11:
Hi,

thank you for providing the benchmark and easy-to-use codebase!

When evaluating the sickr-sts task, there exists a KeyError: 'validation'. The reason is that the 'validation' is included in the "eval_splits" of [SickrSTS task description](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/STS/SickrSTS.py) while [mteb/sts15-sts](https://huggingface.co/datasets/mteb/sickr-sts/tree/main) only provides the test set. Should the 'validation' be removed from the task description?

PR #84, opened 2022-10-24
Remove validation

Muennighoff:
Fixes https://github.com/embeddings-benchmark/mteb/issues/83


NouamaneTazi (approved):
LGTM!
Interesting, I thought I had double checked all `eval_splits` in all tasks.
Do you happen to know why we missed this one @Muennighoff? If so, could there be other issues like this one?

Muennighoff:
> LGTM! Interesting, I thought I had double checked all `eval_splits` in all tasks. Do you happen to know why we missed this one @Muennighoff? If so, could there be other issues like this one?

Not sure, I also thought we had removed all of them 🧐

PR #97, opened 2023-02-21
Fix SemmEval description

ahoho:
Unless I'm missing something, I think the [SemmEval](face.co/datasets/mteb/summeval) description is incorrect---the dataset consists of summaries of news articles, not biomedical abstracts.


Muennighoff:
You're right, thanks for noticing this!

Muennighoff (approved):

Issue #101, opened 2023-04-02
Set `n_inits` explicitly in clustering tasks


stephantul:
When running clustering tasks, I keep seeing this warning:

```
FutureWarning: The default value of `n_init` will change from 3 to 'auto' in 1.4. Set the value of `n_init` explicitly to suppress the warning
```

To keep the behavior or MTEB stable across versions of sklearn you should set `n_init` to 3 explicitly. If someone happens to run this with an sklearn version >= 1.4, they would start getting different results. If you want, I can make a PR.

I'm on scikit-learn 1.2.2

Muennighoff:
Great point! Would be amazing if you could open a PR for that 😇

stephantul:
Will do!

stephantul:
@Muennighoff I'm working on a PR, but I'm a bit confused about isort and black. You seem to use them, but when I run them on master, literally everything gets reformatted. Should I include black and isort in the PR?

Muennighoff:
> @Muennighoff I'm working on a PR, but I'm a bit confused about isort and black. You seem to use them, but when I run them on master, literally everything gets reformatted. Should I include black and isort in the PR?

Yeah I think it's fine to include them!

stephantul:
#104 is up. Let me know what you think 😄

PR #104, opened 2023-04-11
Fix clustering warning

stephantul:
This PR does the following:

- isorts everything
- blackens everything
- adds test for clustering evaluation
- fixes the cluster evaluator that kept throwing a warning

### DIFF
--- a/mteb/evaluation/evaluators/ClusteringEvaluator.py
+++ b/mteb/evaluation/evaluators/ClusteringEvaluator.py
@@ -26,7 +26,7 @@ class ClusteringEvaluator(Evaluator):
 
         logger.info("Fitting Mini-Batch K-Means model...")
         clustering_model = sklearn.cluster.MiniBatchKMeans(
-            n_clusters=len(set(self.labels)), batch_size=self.clustering_batch_size
+            n_clusters=len(set(self.labels)), batch_size=self.clustering_batch_size, n_init="auto"
         )
         clustering_model.fit(corpus_embeddings)
         cluster_assignment = clustering_model.labels_
--- a/mteb/evaluation/evaluators/ClusteringEvaluator.py
+++ b/mteb/evaluation/evaluators/ClusteringEvaluator.py
@@ -4,7 +4,6 @@ import numpy as np
 import sklearn
 import sklearn.cluster
 
-
 logger = logging.getLogger(__name__)
 
 from .Evaluator import Evaluator
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -1,9 +1,10 @@
-from abc import ABC, abstractmethod
 import random
+from abc import ABC, abstractmethod
 
 import datasets
 import numpy as np
 import torch
+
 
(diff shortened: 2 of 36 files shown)


mteb/evaluation/evaluators/ClusteringEvaluator.py
```diff
@@ -26,7 +25,7 @@ def __call__(self, model):
 
         logger.info("Fitting Mini-Batch K-Means model...")
         clustering_model = sklearn.cluster.MiniBatchKMeans(
-            n_clusters=len(set(self.labels)), batch_size=self.clustering_batch_size
+            n_clusters=len(set(self.labels)), batch_size=self.clustering_batch_size, n_init="auto"
```
Muennighoff, line 28: Amazing work! To clarify didn't you want to set the value of `n_init` explicitly to 3 instead of `auto`? (https://github.com/embeddings-benchmark/mteb/issues/101#issue-1650808379)

stephantul: Good point! I based myself on [this snippet from the sklearn codebase](https://github.com/scikit-learn/scikit-learn/blob/9aaed4987/sklearn/cluster/_kmeans.py#L877).

The default initialization function for `MiniBatchKMeans` is `kmeans++`, which is the one you also select. If you select `auto`, then the number of inits will be based on the initialization function, and will be 1 for `kmeans++`, and 3 otherwise, suggesting that 3 is too high anyway. But I can change it to 3 for backwards compatibility of course.

Muennighoff: I see, so changing it to `auto` like in this PR would make new clustering scores different from previous ones, but is likely the better choice as it aligns with sklearn, correct?

Changing it to `3`, meanwhile, would ensure backwards compat, right?

stephantul: Yep, I think that's the choice we'll have to make.

Muennighoff: I see, I think breaking backwards compat is fine! I just ran it on Biorxiv w/ `average_word_embeddings_komninos` & it gets a tiny bit worse, but if it's what's recommended by sklearn let's go with this 👍

From:
```json
{
  "dataset_revision": "258694dd0231531bc1fd9de6ceb52a0853c6d908",
  "mteb_dataset_name": "BiorxivClusteringS2S",
  "mteb_version": "1.0.3.dev0",
  "test": {
    "evaluation_time": 10.24,
    "v_measure": 0.208173761965734,
    "v_measure_std": 0.006864121082528612
  }
}
```

to 

```json
{
  "dataset_revision": "258694dd0231531bc1fd9de6ceb52a0853c6d908",
  "mteb_dataset_name": "BiorxivClusteringS2S",
  "mteb_version": "1.0.3.dev0",
  "test": {
    "evaluation_time": 6.62,
    "v_measure": 0.20810107842666378,
    "v_measure_std": 0.008176457910048483
  }
}
```

Muennighoff (approved):

PR #111, opened 2023-04-25
Add @property annotation to description method of AbsTask

permutohedra:
Add the missing @property annotation to AbsTask.description, which prevents typing errors (when checked with mypy) in subclasses.


Muennighoff:
Nice!

Issue #122, opened 2023-07-21
Adding a new retrieval task


KennethEnevoldsen:
Hi big fan of this package/leaderboard. I wanted to extend it to include Scandinavian languages (let me know if you want a PR or prefer that I keep it as a fork).

I however have a few issues with understanding the "relevant_docs" for the retriever. Currently, I am having issues with this specific dataset:

```python
import datasets
from mteb.abstasks.AbsTaskRetrieval import AbsTaskRetrieval

from sentence_transformers import SentenceTransformer

from mteb import MTEB

class SweFAQRetrieval(AbsTaskRetrieval):
    """ """

    @property
    def description(self):
        return {
            "name": "SweFAQ",
            "hf_hub_name": "AI-Sweden/SuperLim",
            "description": "A Swedish FAQ dataset. Available as a part of Superlim",
            "reference": "https://spraakbanken.gu.se/en/resources/superlim",
            "type": "Retrieval",
            "category": "s2p",
            "eval_splits": ["test"],
            "eval_langs": ["sv"],
            "main_score": "ndcg_at_10",
            "revision": "7ebf0b4caa7b2ae39698a889de782c09e6f5ee56",
            "beir_name": "NA",
        }

    def load_data(self, **kwargs):
        """
        Load dataset from HuggingFace hub and convert to MTEB format
        """
        if self.data_loaded:
            return

        self.dataset = datasets.load_dataset(
            self.description["hf_hub_name"], "swefaq", revision=self.description.get("revision")
        )
        self.data_loaded = True

        self.corpus, self.queries, self.relevant_docs = {}, {}, {}
        for split in self.description["eval_splits"]:
            dataset = self.dataset[split]  # type: ignore
            # answers = dataset["candidate_answer"]
            answers = dataset["correct_answer"]
            self.corpus[split] = {
                str(idx): {"_id": str(idx), "title": "", "text": answer} for idx, answer in enumerate(answers)
            }

            questions = dataset["question"]
            self.queries[split] = {str(idx): question for idx, question in enumerate(questions)}

            # Not sure what to do here:
            # first guess: Relevant documents is a mapping between each query and the desired responses (most relevant documents) for that query
            # result: Performance is 0.0 for all metrics
            self.relevant_docs[split] = {str(idx): {str(idx): 1} for idx in range(len(answers))}

            # second guess: Relevant documents is a mapping between each query and the relevant documents a pre-search would return
[...]

PR #124, opened 2023-07-24
Added support for Scandinavian Languages

### LINKED ISSUE #126: Allow toggle for raising errors
Currently, tasks fail silently and are written to a file. It would make it easier to debug if it was possible to toggle the raising of errors.

I can create the PR with a solution if this is something you want.

KennethEnevoldsen:
## Addition
- Added all relevant tasks for Da, No, and Sv I could find. Norwegian includes Norwegian Bokmål and Norwegian Nynorsk. Danish also includes a bitext between Danish and one of its dialects.
- Added a minor fix to the Bitext task

## Task selection:
- Wherever possible I try to keep as streamlined with ScandEval as possible. I.e. if there are cases two versions of the dataset exist, I use the same one as ScandEval.
- I do not include machine-translated datasets (this excludes the QA datasets in ScandEval)
- Where possible I use the test split. If there are no splits I use the train split.
- I do not include non-public datasets such as DaNewsroom

## Potential problems
- Can't get a non-zero performance on the retrieval task. See  #122 (A solution is simply to remove it for the first PR)
- Currently, there is an overweight of sentiment classification tasks.

## Other
- I have added the transform_dataset method. Might be relevant to make that a general method in the AbsTask. In that way, the loading workflow is independent of the transformation workflow (would allow you to e.g. add splits without having to fix the custom transformations of each of the datasets). Currently, it is not implemented in the main function.

Fixes #126


Muennighoff (commented):
Amazing work!

- I like the `dataset_transform`. I think we can leave it like you did for now & maybe we'll make it a generic function later
- We still need to import all tasks in the `__init__.py` files for each task directory I think
- Why did you choose 16 `samples_per_label` for all Classification tasks? @NouamaneTazi @loicmagne  do you remember how we selected the different `samples_per_label` values for CLF tasks? Was it based on how big the dataset is / the number of labels?
- Can you run a few models on the new tasks and share the result files here? I can add a new [leaderboard tab](https://huggingface.co/spaces/mteb/leaderboard) for some of these languages where we have a few datasets. Maybe three new CLF tabs for Danish, Norwegian, Swedish? Or is it more useful if it's just one new CLF tab for "Scandinavia"?
- Do we have results on SweFAQRetrieval from prior work & are we able to reproduce it?

Muennighoff:
1. - 3. Sounds great!
4. Nice let me know when you have the files! Then I will add them to the LB & we can merge this 👍
5. Maybe let's remove this task for now then like you suggested? We can add it later if we manage to fix the bug 👍

KennethEnevoldsen:
Perfect. I have send the results by mail.

I have also removed the task and fixed #126

Muennighoff (approved):
LGTM; Have added all tasks to the leaderboard: https://huggingface.co/spaces/mteb/leaderboard

Let me know if the leaderboard looks okay to you? & then will merge 👍

KennethEnevoldsen:
The only change I would make is change the name of Bitext from other to Danish (it is Danish + a danish dialect). Otherwise I think it looks good!

Muennighoff:
> The only change I would make is to change the name of Bitext from other to Danish (it is Danish + a Danish dialect). Otherwise, I think it looks good!
> 
> Edit: Actually if you wish I am creating an aggregated site for the Scandinavian subsection [here](https://kennethenevoldsen.github.io/scandinavian-embedding-benchmark/) (still working on it). Feel free to link to it. Plan to also add Finnish, Icelandic and Faroese as well in the future (as well as adding them to MTEB).
> 
> Edit: Oh it seems like ScalaNbClassification is in Swedish instead of Norwegian

Fixed & added the link! If you want to link it in a different way, let me know - You can also edit the `app.py` of the leaderboard directly if you want to.

Also FYI all your scores are in this repository: https://huggingface.co/datasets/mteb/results
I renamed some of them to account for the name change in SweRecClassification.

Merging now 🚀

PR #133, opened 2023-08-08
Replaced prints with logging

### LINKED ISSUE #127: Replace print with logging.info
It seems like there are a few print statements the user is unable to disable (notably in the classification task). Which can be a bit frustrating. I would probably turn these into logging.info or similar. I can create a PR if you wish?

KennethEnevoldsen:
As discussed in #127 with @NouamaneTazi 

fixes #127


NouamaneTazi (approved):
LGTM! Thank you for taking care of this 🙌

PR #140, opened 2023-08-26
Add multi gpu eval to readme

NouamaneTazi:


Muennighoff (approved):
Looks good - Do we also want to support multi-node etc. for other tasks apart from Retrieval?

NouamaneTazi:
> Do we also want to support multi-node etc. for other tasks apart from Retrieval?

Ofc! It's on my to-do list, just need some bandwidth to add it

PR #143, opened 2023-08-29
Default to false

Muennighoff:
Needed so that the script in the readme works

### DIFF
--- a/mteb/abstasks/AbsTaskRetrieval.py
+++ b/mteb/abstasks/AbsTaskRetrieval.py
@@ -33,7 +33,7 @@ class AbsTaskRetrieval(AbsTask):
         super().__init__(**kwargs)
 
     @staticmethod
-    def is_dres_compatible(model, is_parallel=True):
+    def is_dres_compatible(model, is_parallel=False):
         methods = DRPES_METHODS if is_parallel else DRES_METHODS
         for method in methods:
             op = getattr(model, method, None)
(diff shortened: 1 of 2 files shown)


NouamaneTazi (approved):

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
     @staticmethod
-    def is_dres_compatible(model, is_parallel=True):
+    def is_dres_compatible(model, is_parallel=False):
         methods = DRPES_METHODS if is_parallel else DRES_METHODS
```
NouamaneTazi, line 37: Actually we don't need this method anymore. We can remove it at once

NouamaneTazi: You can fix it in this PR if you'd like, or can be done in a latter PR

Muennighoff: Do you mean we no longer need the method `is_dres_compatible`? Why?
I think if people want to use their own model it is still helpful to use `is_dres_compatible ` to check that it has the right methods like e.g. https://github.com/microsoft/unilm/blob/b60c741f746877293bb85eed6806736fc8fa0ffd/e5/mteb_beir_eval.py#L95

NouamaneTazi: The `DRPES_METHODS` need to be updated

Muennighoff: Ah I see you mean we do not need the `is_parallel` kwarg anymore as `DRPES_METHODS` is now the same as `DRES_METHODS ` right?

NouamaneTazi: I think so, you might have to double check

Muennighoff: You're right; for your branch that's all we need!

PR #145, opened 2023-08-31
Rely on standard encode kwargs only

Muennighoff:


mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -93,13 +94,11 @@ def compute_metrics(self, model, corpus_model=None, corpus_embeddings: torch.Ten
...
-            show_progress_bar=self.show_progress_bar,
-            batch_size=self.batch_size,
-            convert_to_tensor=True,
-        )
-
+        # We don't know if encode has the kwargs show_progress_bar
+        kwargs = {
+            "show_progress_bar": self.show_progress_bar
+        } if "show_progress_bar" in inspect.signature(model.encode).parameters else {}
+        query_embeddings = np.asarray(model.encode(self.queries, batch_size=self.batch_size, **kwargs))
```
Muennighoff, line 101: If fine with you, I think we can also drop the `show_progress_bar` bar here entirely & rely on people to put it in their encode function if they want it

PR #146, opened 2023-08-31
Update Custom Model `encode` type signature in README

rsmith49:
Ran into a matrix math error when returning `List[np.array]` from my custom Model. Updating here so that others don't run into the same error and have to debug.

Let me know if this should instead be filed as a bug for `List[np.array]` return types not being supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,8 @@ class MyModel():
             batch_size (`int`): Batch size for the encoding
 
         Returns:
-            `List[np.ndarray]` or `List[tensor]`: List of embeddings for the given sentences
+            `np.array` or `tensor`: An array/tensor of embeddings for the given sentences.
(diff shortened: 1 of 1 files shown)


Muennighoff:
Thanks a lot for fixing. For which task did you run into that error?
I think it may be fixed by this PR: https://github.com/embeddings-benchmark/mteb/pull/145

rsmith49:
Oh nice! I unfortunately don't have my computer with me this week so I can't check exactly, but I believe it was the first InformationRetrieval task

KennethEnevoldsen (approved):

KennethEnevoldsen:
@rsmith49 feel free to submit a bug report on this if needed, but for now, I will close the PR.

PR #178, opened 2023-11-15
feat: add code search net eval

bwanglzu:
script is not identical as original code search net, now i'm using:

function docstring to search function code, also very short function docstring (length < 3) has been filtered.

The result is evaluated based on mean reciprocal rank (mrr) since each docstring has 1 matched code.


Muennighoff:
Nice adding CodeSearchNet would be amazing!

Issue #177, opened 2023-11-14
Parallel retrieval is automatically selected during fabric training


Markus28:
We train an embedding model using pytorch lightning fabric and we evaluate it periodically on MTEB during training. To this end, our model conforms to the default SentenceTransformer interface and exposes an `encode` method that accepts a list of strings and returns the embeddings as a numpy array.

On the TRECCOVID task, we get the following exception:

```
  File "/home/markus/finetuner-large-scale-training/large_scale_training/callbacks/evaluation_callback.py", line 138, in _run_sequential_evaluation
    results = self._mteb.run(
  File "/home/markus/.cache/pypoetry/virtualenvs/large-scale-training-8_3_LJm3-py3.10/lib/python3.10/site-packages/mteb/evaluation/MTEB.py", line 289, in run
    raise e
  File "/home/markus/.cache/pypoetry/virtualenvs/large-scale-training-8_3_LJm3-py3.10/lib/python3.10/site-packages/mteb/evaluation/MTEB.py", line 271, in run
    results = task.evaluate(model, split, **kwargs)
  File "/home/markus/.cache/pypoetry/virtualenvs/large-scale-training-8_3_LJm3-py3.10/lib/python3.10/site-packages/mteb/abstasks/AbsTaskRetrieval.py", line 81, in evaluate
    results = retriever.retrieve(corpus, queries)
  File "/home/markus/.cache/pypoetry/virtualenvs/large-scale-training-8_3_LJm3-py3.10/lib/python3.10/site-packages/beir/retrieval/evaluation.py", line 23, in retrieve
    return self.retriever.search(corpus, queries, self.top_k, self.score_function, **kwargs)
  File "/home/markus/.cache/pypoetry/virtualenvs/large-scale-training-8_3_LJm3-py3.10/lib/python3.10/site-packages/beir/retrieval/search/dense/exact_search_multi_gpu.py", line 120, in search
    query_embeddings = torch.cat(query_embeddings, dim=0)
TypeError: expected Tensor as element 0 in argument 0, but got numpy.ndarray
```
This problem has emerged only recently, likely with commit 0dd82a9819d32f264a24dbc057f753efbf54e9d8. It also occurs during training on a single GPU.

[...]

PR #179, opened 2023-11-17
Fix selection of DRES/DRPES

Markus28:
This is a small fix for #177. Previously, DRPES was selected based on the presence of the `RANK` environment variable. This led to issues for us during training with pytorch lightning. 

In this PR, we allow passing a keyword argument `parallel_retrieval` to `MTEB.run`, which is further processed by `AbsRetrieval.evaluate` and `BeIRTask.load_data` to toggle the selection of DRES/DRPES. As a consequence, we also need to  pass all kwargs of `MTEB.run` to `task.load_data`. I am not sure whether this could possibly lead to problems in other parts of MTEB.

### DIFF
--- a/mteb/abstasks/AbsTaskRetrieval.py
+++ b/mteb/abstasks/AbsTaskRetrieval.py
@@ -48,7 +48,6 @@ class AbsTaskRetrieval(AbsTask):
             raise Exception("Retrieval tasks require beir package. Please install it with `pip install mteb[beir]`")
 
         if not self.data_loaded:
-            print("Bla", parallel_retrieval)
             self.load_data(parallel_retrieval=parallel_retrieval)
 
         corpus, queries, relevant_docs = self.corpus[split], self.queries[split], self.relevant_docs[split]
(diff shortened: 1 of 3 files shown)


Muennighoff:
I think we can even remove DRPES entirely - Do you ever use it? I'm not sure if people are even using it & I think doing parallel retrieval is much easier to do on the model side (i.e. in the users encode function).

Markus28:
Our team has indeed not used DRPES, since we decided to execute different tasks on the individual GPUs. However, if I understand the idea of DRPES correctly, I could imagine that it makes sense in some cases. Embedding on the model side and retrieval on the MTEB side are mostly separate, so it would be my understanding that DRPES  could be used in combination with parallelism on the model side, leading to some additional speedups (?).

jingedawang:
I encountered the same problem. This solution should be good. Why it still not be able to merge?

Muennighoff:
Merged sorry! I was hoping @NouamaneTazi would take a look as he wrote the code it affected;
We will refactor the BEIR integration soon to make it more usable!

PR #202, opened 2024-01-15
Added Norwegian Bokmål-Nynorsk bitext mining task

x-tabdeveloping:
This is mainly important for the [Scandinavian Embedding Benchmark](https://kennethenevoldsen.github.io/scandinavian-embedding-benchmark/), but generally speaking, all multiligual embeddings should aim to encode sentences in Bokmål and Nynorsk relatively close to each other, as they are two written variants of the same language.

### DIFF
--- a/mteb/tasks/BitextMining/NorwegianCourtsBitextMining.py
+++ b/mteb/tasks/BitextMining/NorwegianCourtsBitextMining.py
@@ -1,0 +1,42 @@
+import datasets
+
+from mteb.abstasks import AbsTaskBitextMining
+
+
+class NorwegianCourtsBitextMining(AbsTaskBitextMining):
+    @property
+    def description(self):
+        return {
+            "name": "NorwegianCourtsBitextMining",
+            "hf_hub_name": "kardosdrur/norwegian-courts",
+            "description": "Nynorsk and Bokmål parallel corpus from Norwegian courts. "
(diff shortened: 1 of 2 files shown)


Muennighoff (approved):
LGTM, cool! If it looks good to @KennethEnevoldsen too, fine to merge with me!

KennethEnevoldsen (approved):
> they are two written variants of the same language.

It is actually two distinctly different languages. E.g. Danes can generally understand bokmål (if spoken slowly) but can't understand nynorsk.

x-tabdeveloping:
I feel like calling it two different languages is a bit of a stretch. Spoken Norwegian has different and much fuzzier standards than written Norwegian, and to my knowledge it is debated whether a standard Norwegian spoken language exist.
The variant normally considered to be the standard is [Standard Østnorsk](https://no.wikipedia.org/wiki/Standard_%C3%B8stnorsk).
Nynorsk and Bokmål are the two official written languages, and are taught in school and used in media, but there isn't necessarily a direct connection between which Norwegian you write and how you speak.

KennethEnevoldsen:
I am not an expert, so let us not get too much into it. Subjectively, though, I can't understand Norwegian Nynorsk, but at least the southern dialects of bokmål go just fine. However, reading more up on it, it seems like you are right that they are indeed more on a continuum.

PR #209, opened 2024-01-19
only save top-k

hongjin-su:


mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -86,6 +86,15 @@ def evaluate(
...
                 os.makedirs(output_folder)
+            top_k = kwargs.get('top_k', 10)
+            all_qids = []
+            for k in results:
+                all_qids.append(k)
+            for qid in all_qids:
+                doc_ids = sorted(results[qid].items(), key=lambda x: x[1], reverse=True)
+                remove_doc_ids = [d[0] for d in doc_ids[top_k:]]
+                for a_doc_id in remove_doc_ids:
+                    results[qid].pop(a_doc_id)
```
Muennighoff, line 97: feels like this could be simpler with fewer lines of code 🤔

Muennighoff, line 97:
```suggestion
            top_k = kwargs.get('top_k', None)
            if top_k is not None:
                for qid in list(results.keys()):
                    doc_ids = set(sorted(results[qid], key=lambda x: results[qid][x], reverse=True)[:top_k])
                    results[qid] = {k: v for k, v in results[qid].items() if k in doc_ids}
```

Muennighoff: it can be done in half the lines :)

Issue #215, opened 2024-01-24
Passing language codes to model when encoding


avidale:
I am planning to evaluate a few multilingual encoders with MTEB. 
Some of them ([SONAR](https://github.com/facebookresearch/SONAR), for example) require explicitly specifying the input language when encoding a text. 
For monolingual tasks, this is not a problem: I can specify the language in advance, before running the task. 
However, there are cross-lingual tasks, such as BUCC and Tatoeba, which require processing several languages alternatively, so I cannot specify a single language before running a task. 

So I wonder: is it possible to pass the language code, along with the other inputs, to the `model.encode` method?
This would enable evaluating language-dependent models correctly. 

I probably could do this on my internal clone of MTEB, but I am hesitating about committing my code to the public version, because the `encode` method in `SentenceTransformer` (which is the main type of models evaluated here) does not support extra arguments (https://github.com/UKPLab/sentence-transformers/blob/v2.3-release/sentence_transformers/SentenceTransformer.py#L220). However, for the reproducibility purposes, it would still be nice to have my evaluation working with the public MTEB code.

What would you recommend?

Muennighoff:
I think it would be great to have that possibility in MTEB! I see three options:
1) We could use the Python standard library `inspect` to check the encode signature and if it has a language kwarg provide it
2) We could have an additional kwarg in the .evaluate func sth like `pass_language` or so that if activated will pass that kwarg
3) We could check if the model has a method `encode_with_lang` or sth and if so use that instead of the regular encode function and pass it the lang kwarg. Similar to how `encode_query` is detected for retrieval.

If you don't think any of the above is good, feel free to solve it in a different way! I'd be happy to review & merge a PR for this.

KennethEnevoldsen:
@Muennighoff you might be interested in the approach of SEB here, where we allow the model to do dynamic encoding based on the task object (most notably the metadata) e.g. for instruct-type models or for the cohere embedding model.

Muennighoff:
> @Muennighoff you might be interested in the approach of SEB here, where we allow the model to do dynamic encoding based on the task object (most notably the metadata) e.g. for instruct-type models or for the cohere embedding model.

It's a good point. My main worry is that it comes at the cost of simplicity 🤔

KennethEnevoldsen:
In v2 the current encode interface looks like:

```py
    def encode(
        self,
        inputs: DataLoader[BatchedInput],
        *,
        task_metadata: TaskMetadata,
[...]

Issue #139, opened 2023-08-25
Evaluation of SONAR


gulldan:
Hello hi,

I am quite curious to see the performance of SONAR model in terms of multilingual embeddings on XSTS task. I haven't found anything on it on MTEB. 

https://ai.meta.com/research/publications/sonar-sentence-level-multimodal-and-language-agnostic-representations/
https://github.com/facebookresearch/SONAR

gulldan:
Hi,
i tried but i havent enough RAM memory, test needed more then 64 GB, so i cant perform it.

avidale:
I am curious, which of the tasks needs so much memory? Is it possible to optimize it?

SONAR text encoder itself occupies only about 3GB on disk. With a small batch size, and with truncating long texts to 1024 tokens (it doesn't support more anyway), a GoogleColab-sized machine should be enough for embedding any texts with SONAR.

gulldan:
Clustering, STS tasks on bench. 
Problem not about VRAM, but RAM.

avidale:
I did evaluate SONAR encoders (with a few assumptions about language codes).

However, even after adding the output of `scripts/mteb_meta.py` to [the model's readme](https://huggingface.co/facebook/SONAR), I cannot see SONAR on the most of the leaderbord's tabs after hitting the refresh button (an exception is the Eng-X bitext mining tab, where SONAR is the new SOTA). And I cannot understand why.

My current suspect is that `scripts/mteb_meta.py` misses some of the tasks. For example, I have results for `BornholmBitextMining`, but it is missing in `mteb_metadata.md`.

To check this, I attach all the json results: [sonar_results.zip](https://github.com/embeddings-benchmark/mteb/files/14166961/sonar_results.zip). They correspond to all tasks, except of MSMARCOv2 which takes too long to embed, and the Polish retrieval tasks mentioned at https://github.com/embeddings-benchmark/mteb/issues/219. 

@Muennighoff could you please help me sort this out?

avidale:
Yes, thanks, it works!

KennethEnevoldsen:
Will close this issue as it seems to have become stale, though the [Scandinavian embedding benchmark](https://arxiv.org/abs/2406.02396) does evaluate the SONAR models

Muennighoff:
SONAR is also on the [leaderboard](https://huggingface.co/spaces/mteb/leaderboard) with its scores now I think

KennethEnevoldsen:
[...]

PR #216, opened 2024-01-26
[#215] Optionally specify language when encoding texts 

avidale:
**Motivation**: see #215 

**Implementation**: 
- Add the language argument to each abstract task and each evaluator, and pass to the model, if its encode method has the `language` argument.
- For cross-lingual tasks, split the language code in two.
- Include https://github.com/embeddings-benchmark/mteb/pull/221.
- Fix task name `ScalaNnClassification` (it used to be `ScalaNbClassification`, copy-pasted from the previous task).
- For `ScalaNbClassification` and `ScalaNnClassification`, remove redundant language codes, because they are monolingual: `nn` and `nb` are just special cases of `no`.
- Update the readme with a few nits about non-English task.

**Some issues spotted**
- A few tasks are not marked as multilingual or crosslingual, but do contain multiple languages (BornholmBitextMining, NordicLangClassification, ScalaNbClassification, ScalaNnClassification). The last two seem to be monolingual, but are marked with two codes. BornholmBitextMining contains Danish and its Bornholm dialect; both will be interpreted as Danish by the model. 
For NordicLangClassification; the first language code (da) will be passed to the model for the texts in all languages. Using the same code for all languages makes sense (because we want to classify the languages without leaking the answer), but the choice of the code is arbitrary.
- Polish retrieval tasks still don't work: https://github.com/embeddings-benchmark/mteb/issues/219

**Testing**
Relying on the automatic tests. 
Also, succesfully ran the evaluation with SONAR in https://github.com/embeddings-benchmark/mteb/issues/139.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ You can evaluate only on `test` splits of all tasks by doing the following:
 evaluation.run(model, eval_splits=["test"])
 ```
 
-Note that the public leaderboard uses the test splits for all datasets except MSMARCO, where the "dev" split is used.
+Note that the public leaderboard for English uses the test splits for all datasets except MSMARCO, where the "dev" split is used.
 
 ### Using a custom model
 
@@ -137,6 +137,13 @@ model = MyModel()
 evaluation = MTEB(tasks=["Banking77Classification"])
 evaluation.run(model)
 ```
+Please note that some evaluators pass additional arguments to `encode` (such as `show_progress_bar` or `convert_to_tensor`), 
+expecting the [SentenceTransformer.method](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L220) 
+or a compatible one. A custom `encode` method with `**kwargs` will ignore all such extra arguments.
+
+If your model needs the language tag to embed the texts, you can add the `language` argument to its `encode` method.
+If you are running it on multilingual tasks, please make sure that diverse language codes 
+(such as `es`, `ber`, `zh-TW`, `da-bornholm`) are processed correctly.
 
 If you'd like to use different encoding functions for query and corpus when evaluating on Retrieval or Reranking tasks, you can add separate methods for `encode_queries` and `encode_corpus`. If these methods exist, they will be automatically used for those tasks. You can refer to the `DRESModel` at `mteb/mteb/abstasks/AbsTaskRetrieval.py` for an example of these functions.
 
@@ -199,6 +206,11 @@ evaluation.run(model)
 ```
 
 > **Note:** for multilingual tasks, make sure your class also inherits from the `MultilingualTask` class like in [this](https://github.com/embeddings-benchmark/mteb-draft/blob/main/mteb/tasks/Classification/MTOPIntentClassification.py) example.
+
+### Task-specific notes
+* To run the `DKHateClassification` task (Danish language) you need first to manually ask for access to the gated 
+dataset: https://huggingface.co/datasets/DDSC/dkhate.
+
 
 ## Leaderboard
 
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -56,3 +56,11 @@ class AbsTask(ABC):
         :param split: Which datasplit to be used.
         """
         raise NotImplementedError
+
+    def get_language(self):
+        """ Return the first language of the task.
+        This is not meaningful for multilingual or cross-lingual tasks.
+        Also, a few Nordic tasks although not marked as multilingual, contain multiple languages.
+        For them, the first language will be returned.
+        """
+        return self.description['eval_langs'][0]
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -1,9 +1,13 @@
+import logging
 import random
 from abc import ABC, abstractmethod
 
 import datasets
 import numpy as np
 import torch
+
+
+logger = logging.getLogger(__name__)
 
 
 class AbsTask(ABC):
@@ -63,4 +67,8 @@ class AbsTask(ABC):
         Also, a few Nordic tasks although not marked as multilingual, contain multiple languages.
         For them, the first language will be returned.
         """
-        return self.description['eval_langs'][0]
+        langs = self.description['eval_langs']
+        if len(langs) != 1:
+            name = self.description['name']
+            logger.warning(f"For task {name}, the number of languages is not 1: {langs}. Is it multi- or cross-lingual?")
+        return langs[0]
--- a/mteb/utils.py
+++ b/mteb/utils.py
@@ -1,0 +1,47 @@
+import inspect
+
+
+def get_embed_with_lang_func(model, query_specific=False, corpus_specific=False):
+    """
+    If model.encode supports the `language` argument, return this function.
+    Otherwise, return a wrapper of model.encode with this extra argument.
+    This is needed, because some of the models migh embed in a language-specific way.
+    """
+    func = model.encode
+
+    # for reranking tasks, a model (like E5) may have special methods for encoding queries and passages.
+    if query_specific and hasattr(model, 'encode_queries'):
+        func = model.encode_queries
+    if corpus_specific and hasattr(model, 'encode_corpus'):
+        func = model.encode_corpus
+
+    signature = inspect.signature(func)
+    if "language" in signature.parameters:
+        return func
+
+    # TODO: describe the signature of this function with more details
+    # text, batch_size, maybe progress_bar
+    def func_without_language(sentences, *args, language=None, **kwargs):
+        return func(sentences, *args, **kwargs)
+
+    return func_without_language
+
+
+# language codes that contain a dash and should not be split 
+LANGS_WITH_DASH = {
+    "zh-CN", 
+    "zh-TW",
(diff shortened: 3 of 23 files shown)


Muennighoff:
Interesting, didn't realize it was that complex 😅
For BEIR, we need to remove the integration anyways, so we can also leave it for later (cc @tomaarsen)

avidale:
Actually, it is even more complex :(
Because now I am trying to evaluate SONAR with all the languages, and many language codes are ambiguous: does "Kurdish" refer to Northern, Central or Southern Kurdish? Does "Norvegian" refer Bokmal or Nynorsk? Does "Albanian" refer to Tosk or Gheg dialects?
For a model where the output depends on the exact language code, it is kind of nightmare. 
Maybe, for the tasks added in the future, it would make sense to start indicating more specific language codes (e.g. ISO 639-3 code for the language + ISO 15924  code for the script, like "eng_Latn"), to make the set of languages slightly less ambiguous.

Muennighoff:
> Actually, it is even more complex :( Because now I am trying to evaluate SONAR with all the languages, and many language codes are ambiguous: does "Kurdish" refer to Northern, Central or Southern Kurdish? Does "Norvegian" refer Bokmal or Nynorsk? Does "Albanian" refer to Tosk or Gheg dialects? For a model where the output depends on the exact language code, it is kind of nightmare. Maybe, for the tasks added in the future, it would make sense to start indicating more specific language codes (e.g. ISO 639-3 code for the language + ISO 15924 code for the script, like "eng_Latn"), to make the set of languages slightly less ambiguous.

Yes I agree! Ideally we also revamp it for all existing datasets. Though I guess for some datasets, they may not even tell us what exact dialect it is.

Muennighoff (commented):
I think this is really cool & we can probably merge it --- I'm just afraid if it adds to much complexity to the code for a feature that's very specific to few models? Does someone have thoughts here cc @tomaarsen @NouamaneTazi 

I think this problem can also be solved on the user side without changing MTEB by just iterating through the languages one by one (i.e. recreating `MTEB(...)` but change `task_langs`) 🤔

avidale:
> I think this problem can also be solved on the user side without changing MTEB by just iterating through the languages one by one (i.e. recreating MTEB(...) but change task_langs) 🤔

It would be fine for me to do so, if MTEB was writing the results in files named like `{TaskName}_{Language}.json`. 
Currently, if I recreate MTEB for each language, I have also to care about creating a new output directory for each language, and then merging all the per-language results. To me, it doesn't look clean either.

And anyway, this solution wouldn't work for bilingual tasks, such as bitext mining.

mteb/abstasks/AbsTask.py
```diff
@@ -56,3 +60,15 @@ def evaluate(self, model, split="test"):
         :param split: Which datasplit to be used.
         """
         raise NotImplementedError
+
+    def get_language(self):
+        """ Return the first language of the task.
+        This is not meaningful for multilingual or cross-lingual tasks.
+        Also, a few Nordic tasks although not marked as multilingual, contain multiple languages.
+        For them, the first language will be returned.
```
Muennighoff, line 68: So for those Nordic tasks, it will encode based on the wrong language, correct?

Muennighoff:
> > I think this problem can also be solved on the user side without changing MTEB by just iterating through the languages one by one (i.e. recreating MTEB(...) but change task_langs) 🤔
> 
> It would be fine for me to do so, if MTEB was writing the results in files named like `{TaskName}_{Language}.json`. Currently, if I recreate MTEB for each language, I have also to care about creating a new output directory for each language, and then merging all the per-language results. To me, it doesn't look clean either.
> 
> And anyway, this solution wouldn't work for bilingual tasks, such as bitext mining.

Ah yeah good point re: bilingual, you're right

mteb/abstasks/AbsTask.py
avidale: It depends on the language definition. For example, `ScalaNbClassification` contains labels `["no", "nb"]`, which means "Norvegian" (an inclusive language name) and "Bokmål Norvegian" (a specific standard of Norvegian), respectively. 
If the model makes a distinction between the different Norvegian standards, then the first code, `no`, is ambiguous for it  (and model has to choose the dialect arbitrarily), but the second one is allright. If the model treats Bokmål and Nynorsk as a single language, then the code `no` is appropriate, and the code `nb` is not adding any new information.

Muennighoff: I see but for `NordicLangClassification` it would encode using `da` for texts that are in `sv` etc?

avidale: Yes, it would. But because the goal of the task is to determine the language of the text, we can’t make it much better than choose one language code arbitrarily and pass it with all the inputs.

(and anyway, models like SONAR that are supposed to produce language-agnostic sentence representations aren’t expected to score high in this task, so I think that the choice of this single language code won’t matter much)

KennethEnevoldsen: @avidale. What I would do here instead is pass on the language tag to the model and then let it decide on how to handle it (some models might allow for doing multiple languages at a time).

README.md
```diff
@@ -200,6 +207,11 @@ evaluation.run(model)
 
 > **Note:** for multilingual tasks, make sure your class also inherits from the `MultilingualTask` class like in [this](https://github.com/embeddings-benchmark/mteb-draft/blob/main/mteb/tasks/Classification/MTOPIntentClassification.py) example.
 
+### Task-specific notes
+* To run the `DKHateClassification` task (Danish language) you need first to manually ask for access to the gated 
+dataset: https://huggingface.co/datasets/DDSC/dkhate.
```
KennethEnevoldsen, line 212: Isn't it better to create a custom warning?

mteb/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+    return func_without_language
+
+
+# language codes that contain a dash and should not be split 
+LANGS_WITH_DASH = {
+    "zh-CN", 
+    "zh-TW",
+    "da-bornholm",
+    "en-ext",  # ???
```
KennethEnevoldsen, line 35: You might be interested in the PR #251. For now this PR though I believe that this works

KennethEnevoldsen (changes requested):
See comments above this is mainly caused by a conflicting merge.

Muennighoff:
> I think this is really cool & we can probably merge it --- I'm just afraid if it adds to much complexity to the code for a feature that's very specific to few models?

Do you have thoughts on this @KennethEnevoldsen? This was my main concern 🤔

KennethEnevoldsen:
As I understand it (and feel free to correct me if I am wrong) it only adds the optional language input during the encode? I think it is a reasonable change, though I would simplify it by just passing in either the task metadata (which contains the language as well). This should also allow prompt-based models to create prompts based on the task).

avidale:
@KennethEnevoldsen Out of the box, the task metadata contains information about the set of the languages involved into the current task, but not the language of the specific text which is currently being encoded. 

Anyway, I am willing to modify this PR pass to the model something else alongside with the language tag, if you suggest a convenient way to package it.

KennethEnevoldsen:
@avidale Ahh yeah I see especially for the bitext tasks.

Hmm right so I imagine the ideal encode function would look something like:

```python
def encode(
    sentences: list[str], 
    encode_type: Literal["queries", "corpus"] | None # currently formalized in encode_corpus and encode_queries which I think is fine
    language: list[str] | None, # e.g. required for bitext mining. None can e.g. solve cases where language is unknown (e.g. lang clf)
    task: str # or some kind of metadata object - this allow e.g. embedding differently depending on clustering, retrieval etc.. It also allows for prompt-based encoding.
    )
```

@avidale what are your thoughts here?
@Muennighoff any reason to prefer encode_corpus and encode_queries over passing it as an argument?

Muennighoff:
> @avidale Ahh yeah I see especially for the bitext tasks.
> 
> Hmm right so I imagine the ideal encode function would look something like:
> 
> ```python
> def encode(
>     sentences: list[str], 
>     encode_type: Literal["queries", "corpus"] | None # currently formalized in encode_corpus and encode_queries which I think is fine
>     language: list[str] | None, # e.g. required for bitext mining. None can e.g. solve cases where language is unknown (e.g. lang clf)
>     task: str # or some kind of metadata object - this allow e.g. embedding differently depending on clustering, retrieval etc.. It also allows for prompt-based encoding.
>     )
> ```
> 
> @avidale what are your thoughts here? @Muennighoff any reason to prefer encode_corpus and encode_queries over passing it as an argument?

I don't mind removing `encode_corpus` & `encode_queries` & passing it as an argument though it will break backward compat

KennethEnevoldsen:
@Muennighoff when we switch from MTEB to MMTEB might be the best time to break backward compatibility otherwise I am also fine with leaving it as is.

Muennighoff:
> @Muennighoff when we switch from MTEB to MMTEB might be the best time to break backward compatibility otherwise I am also fine with leaving it as is.

I'm also fine either way ---- I think the switch from MTEB to MMTEB would not break backward compact though no?

KennethEnevoldsen:
>  I think the switch from MTEB to MMTEB would not break backward compat though no?

No you are right. It would just seem like a fitting time to do so if we think it streamlines the encode method for future uses.

KennethEnevoldsen:
Will close this PR as it seems to have gone stale.

PR #246, opened 2024-03-15
docs: Add development installation instructions

KennethEnevoldsen:
Added missing development installation instructions.


README.md
```diff
@@ -213,6 +214,30 @@ Advanced scripts with different models are available in the [mteb/mtebscripts re
...
+## Contributing to MTEB
+We welcome contributions such as new datasets to MTEB! This section describes how to set up the repository for development.
+
+### Development Installation
+If you want to submit a dataset or on other ways contribute to MTEB, you can install the package in development mode:
+
+```bash
+git clone https://github.com/embeddings-benchmark/mteb
+pip install -e .
+pip install -r requirement.dev.txt
```
imenelydiaker, line 228: along with `pip install -re requirements.txt` because it contains the packages needed to run mteb no?

imenelydiaker, line 228:
```suggestion
pip install -r requirements.dev.txt
```

imenelydiaker (approved):

README.md
```diff
@@ -213,6 +214,30 @@ Advanced scripts with different models are available in the [mteb/mtebscripts re
...
+```
+
+### Running Tests
+To run the tests, you can use the following command:
+
+```bash
+pytest 
+# or if you want to run on multiple cores
+pytest -n auto
+```
```
imenelydiaker, line 238: There is a make file here: https://github.com/embeddings-benchmark/mteb/blob/main/Makefile
```suggestion
```bash
make test
```
```

File: README.md (hunk 1 above)
Comment (KennethEnevoldsen) [reply README.md:228]: as least when I tested it it was not required.

File: README.md (hunk 2 above)
Comment (KennethEnevoldsen) [reply README.md:238]: Then we should rewrite make test (it currently takes up all cores which at least for me led to it hanging).

**Comment by KennethEnevoldsen** [comment]
I have now addtionally:

1) updated the CI to use the make commands (this allows the user a 1-1 mapping between ci and make commands)
2) Added linting to docs

File: .github/workflows/python-package.yml [hunk 3]
```diff
@@ -27,15 +27,11 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install flake8 pytest
         python -m pip install .
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        python -m pip install -r requirements.dev.txt
```
imenelydiaker, line 31: I think that you should keep the requirements.txt, else the current package won't find the needed dependencies to run. Why do you think it's not useful ?

KennethEnevoldsen: the requirements are already specified in the setup.py which is installed as a part of `python -m pip install .`

Makefile
```diff
@@ -13,25 +13,38 @@ modified_only_fixup:
...
 # Run tests for the library
-
 test:
-	pytest -n auto --dist=loadfile -s -v ./tests/
\ No newline at end of file
+	pytest
+
+# add parllel test for faster execution (can sometimes cause issues with some tests)
+test-parallel:
+	pytest -n auto --dist=loadfile -s -v
```
Muennighoff, line 50:
```suggestion
# add parallel test for faster execution (can sometimes cause issues with some tests)
test-parallel:
	pytest -n auto --dist=loadfile -s -v

```

README.md
```diff
@@ -213,6 +214,39 @@ Advanced scripts with different models are available in the [mteb/mtebscripts re
 4. Hit the Refresh button at the bottom of the leaderboard and you should see your scores 🥇
 5. To have the scores appear without refreshing, you can open an issue on the [Community Tab of the LB](https://huggingface.co/spaces/mteb/leaderboard/discussions) and someone will restart the space to cache your average scores. The cache is updated anyways ~1x/week.
 
+
+
+## Contributing to MTEB
+We welcome contributions such as new datasets to MTEB! This section describes how to set up the repository for development.
+
+### Development Installation
+If you want to submit a dataset or on other ways contribute to MTEB, you can install the package in development mode:
```
Muennighoff, line 223:
```suggestion
If you want to submit a dataset or contribute in other ways to MTEB, you can install the package in development mode:
```

Muennighoff (approved):
Really nice work; should make contributing much easier!

.github/workflows/python-package.yml
Muennighoff: We could also remove the `requirements.dev.txt` and have them as an extra similar to [transformers](https://github.com/huggingface/transformers/blob/00c1d87a7d5c8dfb4554370983b5a3f7c069edd7/setup.py#L445C20-L445C26) ; So one would install dev with something like `pip install -e ".[dev]"` (https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md); I don't mind either one, the current way may actually use less code & is simpler so I think it's fine

Muennighoff:
Are the failing tests expected?

KennethEnevoldsen:
> Are the failing tests expected?

No. Not sure why it happens. Will try to debug it

.github/workflows/python-package.yml
KennethEnevoldsen: I will change it to the ".[dev]" approach. That was all requirements are specified in one place.

PR #249, opened 2024-03-15
tests: do not run tests on collection

MartinBernstorff:
I noticed test collection was quite slow. Turns out one of the tests was run on import, i.e. as soon as pytest collects the tests.

Instead, I have wrapped it as its own test, and collection is now much faster!


KennethEnevoldsen (approved):
Looks good def. a good fix.

Muennighoff (approved):
Great catch!

PR #253, opened 2024-03-18
dev: add workspace defaults in VSCode

MartinBernstorff:
In VSCode, I use Ruff as the default formatter for most of my Python projects. This means that, if I set no settings, I accidentally format files slightly differently than Black as soon as I edit them.

Settings a workspace-specific default in `.vscode/settings.json` and committing it to the repo ensures that people do not accidentally format with the wrong formatter.

This also enables e.g. test-discovery to be setup automatically.

I know this is somewhat opinionated, but I think it improves the user experience. Would love to discuss, and no pressure 👍


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "python.testing.pytestArgs": [
+        "tests"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true,
+    "editor.defaultFormatter": "ms-python.black-formatter"
+}
```
Muennighoff, line 8: Should we add a newline at the end of this file?

Muennighoff (approved):
Good with me, nice!

.vscode/settings.json
KennethEnevoldsen, line 8:
```suggestion
}

```

KennethEnevoldsen: Added it

PR #254, opened 2024-03-18
Added model results to repo and updated CLI to create consistent folder structure.

KennethEnevoldsen:
This is a suggested change. The goal is to make it easier to add model evaluations along with a dataset (as a kind of test).

This includes a few changes:
- the results folder is now added to contain results using the form `results/{model_name}/{task_results}`
- Changes CLI to follow this format
  - Additionally also adds a model_meta.json with the model_name, time of run and versions during run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ mteb -m sentence-transformers/all-MiniLM-L6-v2 \
     -t Banking77Classification  \
     --verbosity 3
 
-# if nothing is specified it with default to saving the results in the results/{model_name} folder
+# if nothing is specified default to saving the results in the results/{model_name} folder
 
 mteb -m sentence-transformers/all-MiniLM-L6-v2
 ```
--- a/README.md
+++ b/README.md
@@ -57,10 +57,13 @@ results = evaluation.run(model, output_folder=f"results/{model_name}")
 ```bash
 mteb --available_tasks
 
-mteb -m average_word_embeddings_komninos \
+mteb -m sentence-transformers/all-MiniLM-L6-v2 \
     -t Banking77Classification  \
-    --output_folder results/average_word_embeddings_komninos \
     --verbosity 3
+
+# if nothing is specified it with default to saving the results in the results/{model_name} folder
+
+mteb -m sentence-transformers/all-MiniLM-L6-v2
 ```
 
 * Using multiple GPUs in parallel can be done by just having a custom encode function that distributes the inputs to multiple GPUs like e.g. [here](https://github.com/microsoft/unilm/blob/b60c741f746877293bb85eed6806736fc8fa0ffd/e5/mteb_eval.py#L60) or [here](https://github.com/ContextualAI/gritlm/blob/09d8630f0c95ac6a456354bcb6f964d7b9b6a609/gritlm/gritlm.py#L75).
--- a/mteb/cmd.py
+++ b/mteb/cmd.py
@@ -10,7 +10,10 @@ example call:
 
 
 import argparse
+import datetime
+import json
 import logging
+from pathlib import Path
 
 from sentence_transformers import SentenceTransformer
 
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -57,10 +57,13 @@ results = evaluation.run(model, output_folder=f"results/{model_name}")
...
 mteb --available_tasks
 
-mteb -m average_word_embeddings_komninos \
+mteb -m sentence-transformers/all-MiniLM-L6-v2 \
     -t Banking77Classification  \
-    --output_folder results/average_word_embeddings_komninos \
     --verbosity 3
+
+# if nothing is specified it with default to saving the results in the results/{model_name} folder
+
```
Muennighoff, line 65:
```suggestion
# if nothing is specified default to saving the results in the results/{model_name} folder

```

Muennighoff (commented):
Interesting - is the intention to have one result file for every dataset? Could be a good idea, so it's easy to get an idea of what kind of performance to expect

KennethEnevoldsen:
Yep exactly. This also makes it easier for us to review dataset submissions as they then also include at least 1-3 model which have run on the data.

KennethEnevoldsen:
The assumption is to have the results at least for newly submitted datasets.

Muennighoff:
Makes sense, should this be specified somewhere e.g. in how to contribute? Also maybe it makes sense to explicitly tell people always to run model X so it's a bit easier to compare if it's always the same model? If it's just 1 model, it should probably be a multilingual one 🤔

KennethEnevoldsen:
> Also maybe it makes sense to explicitly tell people always to run model X so it's a bit easier to compare if it's always the same model? If it's just 1 model, it should probably be a multilingual one 🤔

Yep I plan to make a PR with how to add datasets that include some standard models (small and multilingual). I was thinking e5-multilingual-small and the sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2.

imenelydiaker:
Great idea ! Both model are multilingual so yes why not. Just keep in mind that for some languages they may not perform very well.. [under-represented ones in datasets]

results/sentence-transformers__all-MiniLM-L6-v2/model.json
```diff
@@ -0,0 +1 @@
+{"model_name": "sentence-transformers/all-MiniLM-L6-v2", "time_of_run": "2024-03-18 11:22:22.739054", "versions": {"sentence_transformers": "2.0.0", "transformers": "4.6.1", "pytorch": "1.8.1"}}
```
imenelydiaker, line 1: Maybe add a revision number to make sure it's the same model version that is used? Wdyt?

mteb/cmd.py
```diff
@@ -20,6 +23,25 @@
 logger = logging.getLogger(__name__)
 
 
+def _name_to_path(name: str) -> str:
+    return name.replace("/", "__").replace(" ", "_")
```
imenelydiaker, line 27: I think you can remove `sentence_transformers` prefix from the model name. Just as in `mteb/results` (we already have the info in the metadata)

MartinBernstorff:
I suppose this closes #254 - would add it to the PR description, but can't edit 👍

KennethEnevoldsen:
@MartinBernstorff this is #254

MartinBernstorff:
Ah yeah, have updated my comment 👍

mteb/cmd.py
imenelydiaker: okay 👍

results/sentence-transformers__all-MiniLM-L6-v2/model.json
imenelydiaker: I was thinking about doing the same as for datasets in mteb (`revision_id`). Just specifying  the commit id from the HF repo that stores the model.

KennethEnevoldsen: Yes I would love to do that. However, I am not sure the commit id is available in the model object (you would have to know it beforehand). I would love to add that (but seems like that is outside the scope of this PR)

@imenelydiaker do anyone from your team have the time to give it a go?

imenelydiaker: I've seen this in the `transformers` doc: https://huggingface.co/docs/transformers/main_classes/model#transformers.PreTrainedModel.from_pretrained.revision
But it's for the `PreTrainedModel.from_pretrained()` function. I can check how it can be applied for `SentenceTransformers` 🙂

PR #255, opened 2024-03-18
docs: add dataset schemas

### LINKED ISSUE #250: docs: document schema for datasets in tasks
When I added #247, it was non-obvious for me which columns a dataset should contain.

I propose documenting this at the `AbsTask` level, e.g. for classification something like:

```python
class AbsTaskClassification(AbsTask):
    """
    Abstract class for kNN classification tasks
    The similarity is computed between pairs and the results are ranked. 
    
    Dataset must be a huggingface dataset split into train/test, and contain the following columns:
        text: str
        label: int
    """

### PR DESCRIPTION
**Pull request by MartinBernstorff**
Unfortunately, the diff contains a lot of formatting changes as well, see #256.

I've gone through it manually, and it should only be docstrings which are changed. I've had to make a few educated guesses, so a thorough look-through is definitely recommended.

@Muennighoff @KennethEnevoldsen 

Closes #250.

### DIFF
--- a/mteb/abstasks/AbsTaskReranking.py
+++ b/mteb/abstasks/AbsTaskReranking.py
@@ -10,7 +10,7 @@ class AbsTaskReranking(AbsTask):
     self.queries = {'dev': Dict[id, str], 'test': Dict[id, str]}
     self.relevant_docs = {'dev': Dict[id, set], 'test': Dict[id, set]}
 
-    Dataset must be a huggingface dataset containing a split matching the task description's "eval_splits" and the following columns:
+    self.load_data() must return a huggingface dataset containing a split matching the task description's "eval_splits" and the following columns:
         query: str
         positive: list[str]
         negative: list[str]
--- a/mteb/abstasks/AbsTaskReranking.py
+++ b/mteb/abstasks/AbsTaskReranking.py
@@ -5,10 +5,6 @@ from .AbsTask import AbsTask
 class AbsTaskReranking(AbsTask):
     """
     Abstract class for re-ranking experiments.
-    Child-classes must implement the following properties:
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: mteb/abstasks/AbsTaskReranking.py [hunk 1]
```diff
@@ -5,10 +5,11 @@
...
     Abstract class for re-ranking experiments.
-    Child-classes must implement the following properties:
-    self.corpus = {'dev': Dict[id, str], 'test': Dict[id, str]}         #id => sentence
-    self.queries = {'dev': Dict[id, str], 'test': Dict[id, str]}
-    self.relevant_docs = {'dev': Dict[id, set], 'test': Dict[id, set]}
+
+    self.load_data() must generate a huggingface dataset with a split matching self.description["eval_splits"], and assign it to self.dataset. It must contain the following columns:
+        query: str
+        positive: list[str]
+        negative: list[str]
```
KennethEnevoldsen, line 12: Not sure about this one either, but if you are then I am fine with it. Looks like the old docstring is def. wrong.

mteb/abstasks/AbsTaskSTS.py
```diff
@@ -5,13 +5,15 @@
...
 
+
 class AbsTaskSTS(AbsTask):
     """
-    Abstract class for re-ranking experiments.
-    Child-classes must implement the following properties:
-    self.corpus = {'dev': Dict[id, str], 'test': Dict[id, str]}         #id => sentence
-    self.queries = {'dev': Dict[id, str], 'test': Dict[id, str]}
-    self.relevant_docs = {'dev': Dict[id, set], 'test': Dict[id, set]}
+    Abstract class for STS experiments.
```
KennethEnevoldsen, line 11: Not sure about this one. I would test it.

KennethEnevoldsen (approved):
Generally looks good!

PR #262, opened 2024-03-19
Restructing the readme

KennethEnevoldsen:
This does PR restructes the readme to allow users to get an easier overview. It does the following: 

1) splits up test using parametrize (this makes then run notably faster as they can be run in parallel)
2) Introduced the docs folder
3) Split larger segments of the readme into the docs folder and introduces a table which links to these 
4) Moved setup.py --> pyproject.toml
5) Removed the double specification of version it is now only specified in one place!


imenelydiaker (approved):
That's amazing! The README is way more accessible this way, nice job! 🚀

PR #265, opened 2024-03-19
replaced linter with ruff

### LINKED ISSUE #263: CI: Linting does not fail if there is changes to be made
In a recent PR I noticed that the linting does not fail the CI. I recommend we fix this along with #256 once the #262 and #260 are merged in.

@MartinBernstorff you might be interested to know this one as well as it related to #256.

@Muennighoff a lot of the formatting (isort, black, flake8) can all be done with ruff. Was hoping to simplify the CI a bit by just using that everywhere?

KennethEnevoldsen:
- replaced black, isort, and flake8 with ruff
- split workflows into tests and lint (lint is disabled as it will fail, but once applied to the repo we can enable it) 
- added release workflow (currently disabled as we need to set up trusted publishing)

@Muennighoff I would like to add @MartinBernstorff as a collaborator as well (then I can e.g. assign him as a reviewer). Just want to check in if that is fine with you?

fixes #263


Makefile
```diff
@@ -1,49 +1,17 @@
...
-# Run tests for the library
 test:
+	@echo "--- 🧪 Running tests ---"
 	pytest
 
-# add parllel test for faster execution (can sometimes cause issues with some tests)
 test-parallel:
+	@echo "--- 🧪 Running tests ---"
+	@echo "Note that parallel tests can sometimes cause issues with some tests."
 	pytest -n auto --dist=loadfile -s -v
```
Muennighoff, line 17:
```suggestion
	pytest -n auto --dist=loadfile -s -v

```

Muennighoff (approved):
Ofc! The contributions by @MartinBernstorff have been amazing, really glad we have him as part of our team!

PR #276, opened 2024-03-21
fix: Added sizes to the metadata

KennethEnevoldsen:
this allow for automatic metadata table generations


mteb/abstasks/TaskMetadata.py
```diff
@@ -144,3 +146,6 @@ class TaskMetadata(BaseModel):
 
     text_creation: TEXT_CREATION_METHOD | None
     bibtex_citation: str | None
+
+    n_samples: dict[str, int]
```
MartinBernstorff, line 150: Maybe we should specify these as dict[SPLIT_TYPE, int]? This can be something as simple as a type-alias, but makes it semantically clearer.

MartinBernstorff: Might also be nice to provide a little utility that gets this info, and document the utility here.

mteb/abstasks/TaskMetadata.py
```diff
@@ -144,3 +146,6 @@ class TaskMetadata(BaseModel):
...
+    avg_character_length: dict[str, float]
```
MartinBernstorff, line 151: We've made the fields nullable for the other ones (instead of an empty container, e.g. `{}`). Perhaps we should do that here as well?

MartinBernstorff (commented):
Excellent work, Kenneth! We end up with a lot of empty fields, but at least we document it clearly, and can handle it if we need the data 👍

mteb/abstasks/TaskMetadata.py
```diff
@@ -116,7 +116,9 @@ class TaskMetadata(BaseModel):
         dialect: The dialect of the data, if applicable. Ideally specified as a BCP-47 language tag.
         text_creation: The method of text creation. Includes "found", "created", "machine-translated", "machine-translated and verified", and
             "machine-translated and localized".
-        bibtex_citation: The BibTeX citation for the dataset.
+        bibtex_citation: The BibTeX citation for the dataset.'
+        n_samples: The number of samples in the dataset.
+        avg_character_length: The average character length of the samples in the dataset.
```
MartinBernstorff, line 121: Clarify that this is for the evaluation splits.

mteb/abstasks/TaskMetadata.py
KennethEnevoldsen: Fixed

mteb/abstasks/TaskMetadata.py
KennethEnevoldsen: it would be possible to check it using the eval_splits attr.

KennethEnevoldsen:
@MartinBernstorff it seems like we generally agree on this PR. I have left a comment unresolved as I am not entirely sure what you mean, but we can add it, in a future PR.

PR #283, opened 2024-03-25
tests: speed up tests

MartinBernstorff:
* Make tests run in parallel by default
* Switch BitextMining test to smaller dataset (Borholm)
* Switch to a faster embedding model


KennethEnevoldsen (approved):
lgtm

PR #299, opened 2024-03-29
Allow extending the load_dataset parameters in custom tasks inheriting AbsTask

gariepyalex:
Currently, it is only possible to pass in a `path` and `revision to `load_dataset`. This is fairly limiting, and forces users to implement their own `AbsTask.load_data` function, which relies on the internals of the library. 

This PR allow to specify any parameter supported by `datasets.load_dataset` in custom tasks. Currently, the metadata is specified as such:

```
    my_task = TaskMetadata(
        name="MyTask",
        hf_hub_name="org/dataset",
        revision="1.0",
        ...
    )
```
This is hard to extend as we would need to add new keys to that Pydantic object for each `load_dataset` key we want to support.

This PR proposes to instead have the following structure:

```
    my_task = TaskMetadata(
        name="MyTask",
        dataset={
            "path": "org/dataset",
            "revision": "123"
        }
        ...
    )
```
This allows users to add any key they want (dataset config name, token, etc.). All key/values are passed in to `load_dataset`. **Note that this is done in a backward compatible manner**. A pydantic validator supports the old parameters and populates the `dataset` dictionary.

### Migration
We've included in the PR a migration of all the built-in tasks to avoid logging a deprecation warning when running them. **This is what causes most of the line changes**.

### Refactoring
While migrating the tasks to the new `metadata_dict` to use `dataset` that a lot of time, `AbsTask.load_data` was overridden for the sole purpose of inserting a `dataset_transform` call. This was such a common pattern that this led to a lot of duplication. We propose here to call `dataset_transform` in `AbsTask`, and this default to a no-op. 

### Bug fixes
A few times, the revision of the dataset was not passed to `load_dataset`, leading to a discrepancy between the revision in the metadata and the actual loaded data. These were fixed and a `BUGFIX` comment was added to help PR review. We will remove them after the initial PR review process.

### Testing
I see that the repo contains test suites for all abstract tasks. Please let me know if any additional testing is required from our end.

cc. @gbmarc1


mteb/tasks/BitextMining/multilingual/DiaBLaBitextMining.py
```diff
@@ -42,10 +44,8 @@ def load_data(self, **kwargs):
 
         self.dataset = {}
         for lang in self.langs:
-            self.dataset[lang] = datasets.load_dataset(
-                self.metadata_dict["hf_hub_name"],
-                revision=self.metadata_dict.get("revision", None),
-            )
+            # TODO it looks like the lang is not used here, should we pass `name=lang`?
+            self.dataset[lang] = datasets.load_dataset(**self.metadata_dict["dataset"])
```
gariepyalex, line 48: It looks like the lang is not used here, should we pass `name=lang`?

mteb/tasks/Retrieval/en/NarrativeQARetrieval.py
```diff
@@ -43,7 +45,9 @@ def load_data(self, **kwargs):
             return
 
         data = datasets.load_dataset(
-            self.metadata_dict["hf_hub_name"], split=self._EVAL_SPLIT
+            split=self._EVAL_SPLIT
+            # BUGFIX: the revision is now used
+            ** self.metadata_dict["dataset"],
```
gariepyalex, line 50: BUGFIX: the revision is now used

mteb/tasks/Retrieval/es/SpanishPassageRetrievalS2P.py
```diff
@@ -37,23 +39,23 @@ def load_data(self, **kwargs):
         if self.data_loaded:
             return
 
+        # BUGFIX: the revision is now used
```
gariepyalex, line 42: BUGFIX: the revision is now used

mteb/tasks/Retrieval/fr/AlloprofRetrieval.py
```diff
@@ -37,11 +39,14 @@ def load_data(self, **kwargs):
         if self.data_loaded:
             return
         # fetch both subsets of the dataset
+        # BUGFIX: the revision is now used
         corpus_raw = datasets.load_dataset(
-            self.metadata_dict["hf_hub_name"], "documents"
+            name="documents",
```
gariepyalex, line 44: BUGFIX: the revision is now used

mteb/tasks/Retrieval/fr/SyntecRetrieval.py
```diff
@@ -39,11 +41,14 @@ def load_data(self, **kwargs):
         if self.data_loaded:
             return
         # fetch both subsets of the dataset
+        # BUGFIX: the revision is now used
         corpus_raw = datasets.load_dataset(
-            self.metadata_dict["hf_hub_name"], "documents"
+            name="documents",
```
gariepyalex, line 46: BUGFIX: the revision is now used

mteb/abstasks/AbsTask.py
```diff
@@ -26,23 +26,27 @@ def __init__(self, seed=42, **kwargs):
...
         """
         if self.data_loaded:
             return
-
-        # TODO: add split argument
-        self.dataset = datasets.load_dataset(
-            self.metadata_dict["hf_hub_name"],
-            revision=self.metadata_dict.get("revision", None),
-        )
+        self.dataset = datasets.load_dataset(**self.metadata_dict["dataset"])
```
KennethEnevoldsen, line 42:
```suggestion
        self.dataset = datasets.load_dataset(
             self.metadata. hf_hub_name,
             revision=self.metadata.revision,
             **self.metadata.dataset_kwargs
         )
```

thanks @gariepyalex this seems like a very solid commit. I would generally love that almost always specify the revision (unless we have good reason not to), I am afraid that the current structure would lead to us not specifying the revision. One option might be to allow a dataset_kwargs, which is not a the revision or HF name.

Let me know what you think about this approach.

gariepyalex: Fair feedback. What I like about the proposed solution is all the arguments passed to `load_dataset` are grouped together. `hf_hub_name` is a name that only exist in MTEB, and it is not immediately obvious that it refers to the path argument of `load_dataset`. 

> I am afraid that the current structure would lead to us not specifying the revision.

An alternative would be to add an additional validator on the `dataset` field of the metadata to ensure that the revision is specified:

```python
@field_validator("dataset")
    def _check_dataset_revision_is_specified(cls, dataset):
        if "revision" not in dataset:
            raise ValueError("You must explicitly specify a revision for the dataset (either a SHA or None).")
        if dataset["revision"] is None:
            logging.warning("It is encourage to specify a dataset revision for reproducability")
        return dataset
```

Let me know what you think! I can also implement the `dataset_kwargs` solution you proposed if you prefer.

KennethEnevoldsen: I can see good reason in that. It would also allow us to change out the dataset source reasonably well in the future. 

I would add a validator for dataset["path"] as well.
I would also rename the field to `load_dataset_kwargs` or `datasets_kwargs` to make it clear what it is being passed to.

mteb/abstasks/TaskMetadata.py
```diff
@@ -91,13 +94,17 @@
...
 class TaskMetadata(BaseModel):
     """
     Metadata for a task.
 
     Args:
-        hf_hub_name: The name of the dataset for the task on the Hugging Face Hub.
-        revision: The revision of the dataset for the task on the Hugging Face Hub.
+        dataset: All arguments to pass to datasets.load_dataset to load the dataset for the task. Refer to https://huggingface.co/docs/datasets/v2.18.0/en/package_reference/loading_methods#datasets.load_dataset
+        hf_hub_name: DEPRECATED. The name of the dataset for the task on the Hugging Face Hub.
+        revision: DEPRECATED. The revision of the dataset for the task on the Hugging Face Hub.
```
KennethEnevoldsen, line 107:
```suggestion
```
I would just delete this. It has been here for less than week.

mteb/abstasks/TaskMetadata.py
```diff
@@ -124,8 +131,9 @@ class TaskMetadata(BaseModel):
         avg_character_length: The average character length of the samples in the dataset. This should only be for the splits evaluated on.
     """
 
-    hf_hub_name: str
-    revision: str
+    dataset: dict | None
+    hf_hub_name: str | None = None  # DEPRECATED, use dataset instead
+    revision: str | None = None  # DEPRECATED, use dataset instead
```
KennethEnevoldsen, line 136:
```suggestion
```

KennethEnevoldsen:
This pr. might also relate to the discussion in #301 about creating two dataset sources mteb org + original dataset.

gariepyalex:
I'll address the comments some time tomorrow!

mteb/abstasks/AbsTask.py
gbmarc1: @KennethEnevoldsen 
I understand your point when saying that renaming the `dataset` field to `load_dataset_kwargs` or `datasets_kwargs`. However, to me, it feels a bit strange to name the field that way. Everything seems to have a declarative format in the TaskMetadata object. The kwargs part feels a bit more functional nomenclature.
That being said, I won't die on a hill for that. Let us know if you still want the change.

mteb/tasks/BitextMining/multilingual/DiaBLaBitextMining.py
KennethEnevoldsen: @imenelydiaker?

mteb/abstasks/AbsTask.py
KennethEnevoldsen: Keep it as `dataset`

mteb/tasks/BitextMining/multilingual/DiaBLaBitextMining.py
imenelydiaker: Actually, there is only one split in [DiaBLa](https://huggingface.co/datasets/rbawden/DiaBLa), so even if you specify the language it won't change anything (it may cause an error). 
In the first implementation the language handled was only `"fr-en"`, and that's what the `data_transform()` function builds (first column with French values and second with English values). 
I see that "en-fr" was added to the metadata, but it is exactly the same dataset, I'm not sure it is relevant.

MartinBernstorff:
This seems like a very pragmatic change, with an extremely clean PR description. Nice job! 

Moving the `load_dataset` logic and calling `dataset_transform` from the AbsTask might make some dataset instances slightly less clear for new contributors. AFAICT it's a deduplication vs. indirection/coupling trade-off, and I think it's very worth it in this instance 👍 

As a side-note: I've been looking into codemodification-programming for cases like this, something like [gritql](https://github.com/getgrit/gritql?tab=readme-ov-file). If you have any experience here, I'd love to hear about it in a discussion.

MartinBernstorff (approved):

mteb/abstasks/AbsTask.py
gariepyalex: All validators were added

KennethEnevoldsen:
Perfect @gariepyalex I have set the tests to run assuming they pass it will be merged it - thanks again for the contribution. If you do want to participate in the [MMTEB](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/readme.md) coming up, feel free to add your names and 2 points to the points sheet (if not feel free to ignore this part).

gariepyalex:
Great! I'll add our names there.

Issue #309, opened 2024-04-03
KeyError: 'test.full'


austinmw:
On version 1.5.1:

```
PairClassification
    - PawsX, s2s, multilingual 1 / 7 langs

[/opt/conda/lib/python3.10/site-packages/datasets/load.py:1461](https://vscode-remote+ssh-002dremote-002bdev.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/datasets/load.py:1461): FutureWarning: The repository for paws-x contains custom code which must be executed to correctly load the dataset. You can inspect the repository content at https://hf.co/datasets/paws-x
You can avoid this message in future by passing the argument `trust_remote_code=True`.
Passing `trust_remote_code=True` will be mandatory to load this dataset from the next major release of `datasets`.
  warnings.warn(
Error while evaluating PawsX: 'test.full'
loaded langs: dict_keys(['en'])
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File <timed exec>:151

File [/opt/conda/lib/python3.10/site-packages/mteb/evaluation/MTEB.py:338](https://vscode-remote+ssh-002dremote-002bdev.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/mteb/evaluation/MTEB.py:338), in MTEB.run(self, model, verbosity, output_folder, eval_splits, overwrite_results, raise_error, **kwargs)
    [334](https://vscode-remote+ssh-002dremote-002bdev.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/mteb/evaluation/MTEB.py:334) logger.error(
    [335](https://vscode-remote+ssh-002dremote-002bdev.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/mteb/evaluation/MTEB.py:335)     f"Error while evaluating {task.metadata_dict['name']}: {e}"
    [336](https://vscode-remote+ssh-002dremote-002bdev.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/mteb/evaluation/MTEB.py:336) )
    [337](https://vscode-remote+ssh-002dremote-002bdev.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/mteb/evaluation/MTEB.py:337) if raise_error:
[...]

PR #316, opened 2024-04-04
Fix PawsX eval splits

imenelydiaker:
Fix for issue #309. Eval splits didn't have the correct name.

### DIFF
--- a/mteb/tasks/PairClassification/multilingual/PawsX.py
+++ b/mteb/tasks/PairClassification/multilingual/PawsX.py
@@ -19,7 +19,7 @@ class PawsX(MultilingualTask, AbsTaskPairClassification):
         reference="https://arxiv.org/abs/1908.11828",
         category="s2s",
         type="PairClassification",
-        eval_splits=["test.full", "validation.full"],
+        eval_splits=["test", "validation"],
         eval_langs=["de", "en", "es", "fr", "ja", "ko", "zh"],
         main_score="ap",
         date=None,


KennethEnevoldsen (approved):
Can you run one of the smaller models on this just to check that it works without issue?

MartinBernstorff (approved):

imenelydiaker:
> Looks good! Agree with Kenneth.
> 
> As a meta-question, maybe it's possible to check if the splits exist in the dataset without downloading it? Might be worth it if we get more errors like this.

Yep was thinking about adding it to the test that check if a dataset + revision id exists

KennethEnevoldsen:
> Yep was thinking about adding it to the test that check if a dataset + revision id exists

@imenelydiaker this test already exists (based on a recent pr)

imenelydiaker:
> > As a meta-question, maybe it's possible to check if the splits exist in the dataset without downloading it? Might be worth it if we get more errors like this.
> 
> As I understand there is no API for this (at least documented), but you could probably do something like:
> 
> ```
> >>> ds = load_dataset("paws-x", "en", revision="8a04d940a42cd40658986fdd8e3da561533a3646", streaming=True)
> >>> ds.keys()
> dict_keys(['train', 'test', 'validation'])
> ```
> 
> It only download metadata I believe:
> 
> ```
> Downloading builder script: 100%|██████████| 6.54k/6.54k [00:00<00:00, 9.68MB/s]
> Downloading metadata: 100%|████████████████| 17.6k/17.6k [00:00<00:00, 9.76MB/s]
> Downloading readme: 100%|██████████████████| 11.8k/11.8k [00:00<00:00, 7.18MB/s]
> ```
> 
> Might be a way to generalize the existing tests.

So you just need to download the dataset in streaming mode to not store it locally and checks the keys of the dataset object?

imenelydiaker:
Should I merge the fix and we'll open another PR for the test?

MartinBernstorff:
Yeah, I'll give it a stab today 👍

PR #319, opened 2024-04-05
fix: Added initial language code suggestion

KennethEnevoldsen:
Added initial suggestion for language codes:

It follows ISO standards as well as the ETF BCP 47 language tag.

This is only an example meant for annotation. 

A solution might be to split the tag up into 2 (eval_langs and add a new metadata for scripts).

so it would be:

```
...
eval_langs: list[LANGUAGE_CODE] | Mapping[str, list[LANGUAGE_CODE]] # would rename to languages
scripts: list[SCRIPT_CODE] | Mapping[str, list[SCRIPT_CODE]]
...
```

The problem with this approach is that I could e.g. have:

```
...
eval_langs = ["dan", "eng", "jpn"]
scripts = ["Latn", "Jpan"] 
...
```
Which seems to indicate that eng-Jpan might be valid. A mapping between eval_langs and script also does not work as a e.g. {"jpn": ["jpn-Jpan"], "jpn": ["jpn-Latn"]}  are both valid.

A third solution might be to completely disregard script codes (simpler but potentially an important variable to test for).

### DIFF
--- a/mteb/abstasks/TaskMetadata.py
+++ b/mteb/abstasks/TaskMetadata.py
@@ -113,9 +113,10 @@ class TaskMetadata(BaseModel):
         category: The category of the task. E.g. includes "s2s", "s2p", "p2p" (s=sentence, p=paragraph).
         reference: A URL to the documentation of the task. E.g. a published paper.
         eval_splits: The splits of the dataset used for evaluation.
-        eval_langs: The languages of the dataset used for evaluation. Specified as a list of ISO 639-3 language codes followed by ISO 15924 script
-            codes (e.g. "eng-Latn"). Can be either a list of languages or a dictionary mapping huggingface subsets to lists of languages (e.g. if a
-            the huggingface dataset contain different languages).
+        eval_langs: The languages of the dataset used for evaluation. Langauges follows a ETF BCP 47 standard consisting of "{language}-{script}"
+            tag (e.g. "eng-Latn"). Where language is specified as a list of ISO 639-3 language codes (e.g. "eng") followed by ISO 15924 script codes
+            (e.g. "Latn"). Can be either a list of languages or a dictionary mapping huggingface subsets to lists of languages (e.g. if a the
+            huggingface dataset contain different languages).
         main_score: The main score used for evaluation.
         date: The date when the data was collected. Specified as a tuple of two dates.
         form: The form of the data. Either "spoken", "written".


mteb/abstasks/TaskMetadata.py
```diff
@@ -109,7 +113,10 @@ class TaskMetadata(BaseModel):
         category: The category of the task. E.g. includes "s2s", "s2p", "p2p" (s=sentence, p=paragraph).
         reference: A URL to the documentation of the task. E.g. a published paper.
         eval_splits: The splits of the dataset used for evaluation.
-        eval_langs: The languages of the dataset used for evaluation.
+        eval_langs: The languages of the dataset used for evaluation. Langauges follows a ETF BCP 47 standard consisting of "{language}-{script}"
```
Muennighoff, line 116:
```suggestion
        eval_langs: The languages of the dataset used for evaluation. Languages follow a ETF BCP 47 standard consisting of "{language}-{script}"
```

mteb/tasks/Classification/multilingual/AmazonCounterfactualClassification.py
```diff
@@ -21,7 +21,12 @@ class AmazonCounterfactualClassification(MultilingualTask, AbsTaskClassification
         category="s2s",
         type="Classification",
         eval_splits=["validation", "test"],
-        eval_langs=_LANGUAGES,
+        eval_langs={
+            "en-ext": ["eng-Latn"],
+            "en": ["eng-Latn"],
+            "de": ["deu-Latn"],
+            "ja": ["jpn-Jpan"],
+        },
```
Muennighoff, line 29: I like the dict idea! Do we still need to modify something such that the dict keys are used to fetch the splits from the hf dataset?

Should the values be used for the name in the json result file or the dict key?

KennethEnevoldsen: Yea there will be the need for a few code changes (this is just to make sure we agree on the format beforehand).

Muennighoff: Thing is if the values are used in the json file that it would be ambiguous if the list has more than 1 item no?

KennethEnevoldsen: Ahh I see what you are getting at. You are referring to {"en-ext": ["eng-Latn"]} vs. {"eng-Latn": "en-ext"}. As a "name" (hf denoted the "en-ext" as "name" when calling the load_dataset).

I was thinking the most intuitive was annotating the "name" with their respective languages (could be multiple), however, we could also do it reverse  {"eng-Latn": "multilingual_subset", "dan-Latn": "multilingual_subset"}. Both approaches are valid.

The first one seems more like annotating the "name".

Muennighoff: I meant `"en-ext": ["eng-Latn", ...]` vs `"en-ext": "eng-Latn"` - if it's the former and there's additional items in the list do we just select the first one for the results file?

Muennighoff: (doesn't matter if we leave the results file keys as is, i.e. `en-ext` in that case; it may make more sense to actlly leave results keys as is, since e.g. for this dataset there'd be a clash as it has both `en-ext` & `en` I think)

KennethEnevoldsen: Ahh so list vs. str. Couldn't we imagine that "en-ext" could be multiple languages? (it would be an odd dataset. It would also make the type match in the typing: `Languages | Mapping[str, Languages]`. If not a list it would be `Languages | Mapping[str, Lang]`, which might be a bit confusing, when working with it.

MartinBernstorff (approved):
We've spoken about this at length and after sleeping on it, and I endorse this design 👍

mteb/tasks/Classification/multilingual/AmazonCounterfactualClassification.py
KennethEnevoldsen: Hmm right good question.

Atm. it is the "key", 


I could imagine it looking somewhat like this:

```
{"name": ...
 "scores across subsets": {"en-ext" ...}
 
 # These could essentially be calculated from the above if the scores are loaded with MTEB:
 "scores across language": {"eng_Latn": ...}
 "scores (aggregated)": ...
 }

MartinBernstorff:
I believe this should be closed since it was superseded, feel free to reopen 👍

Issue #251, opened 2024-03-18
Settle on one language standard


KennethEnevoldsen:
Currently, the repository uses multiple language standards. In #245 it was discussed which standard was the ideal one to use. 

I have multiple options:

1) ISO 639-1 (two-letter codes). Is not comprehensive
2) ISO 639-3 (three letter codes). Attempt to be comprehensive also covering ancient languages
3) BCP 47. A "best current practice" standard which consist of:

- Simple Language Tag: "en" for English, using ISO 639-1 code (or ISO 639-3)
- Language and Region: "en-US" for American English, combining ISO 639-1 and ISO 3166-1 codes.
- Language, Script, and Region: "zh-Hant-TW" for Traditional Chinese as used in Taiwan, combining ISO 639-1, ISO 15924, and ISO 3166-1 codes.

Generally I would recommend 2) (potentially 3) using the iso 639-3 standard). Let me know what you guys think.

Muennighoff:
If we want it to apply to all existing datasets, then I think Flores makes 1) / 2) impossible as it differentiates some languages purely on script basis such as `zho_Hant` & `zho_Hans` (https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/BitextMining/multilingual/FloresBitextMining.py)

KennethEnevoldsen:
Ahh then it might be relevant to use that system. "Language_script", where language is ISO 639-3.

avidale:
> Ahh then it might be relevant to use that system. "Language_script", where language is ISO 639-3. Potentially with the added _region. To allow for specification of dialect.

As far as I understand, for most languages, there are already ISO 639-3 codes for all major dialects (English seems to be one of the rare exceptions), so the extra region addition might not be needed.

imenelydiaker:
I'd go for option 2 (a trigram for languages). But we should keep the script we added for language mappings. What we noticed when creating the French benchmark, was that regardless of MTEB language code standards, some datasets used other codes so we had to map them to MTEB.

KennethEnevoldsen:
It sounds like we are doing, language code (three letters). However, There is language such as da-bornholm, which is a regional dialect of Danish and does not have a three-letter code. Should we just code that as "dan" (the code for danish) and then add the dialect in the description/metadata?

imenelydiaker:
[...]

PR #326, opened 2024-04-08
feat: Added new language code standard

KennethEnevoldsen:
This adds a new language code standard using an iso 639-3 code as well as an iso code for scripts. 

I believe the benchmarks are currently preserved as they should be. I also believe all tasks run as intended (I have tested a few of the relevant candidates).

A few things we would probably need to fix:
- [ ] MTEB(task_langs=[...]), currently uses just the user-specified splits, we should instead transfer it to use the standard language codes.
  - Currently the above filter also only filters within tasks (not monolingual tasks). I believe it should do both.
  - While we are changing it, we could also add a few other filters based on domains etc.
- [ ] The above will cause a need to update the benchmark lists (shouldn't be too big)
- [ ] This change does not lead to changes (at least as I understand it) in the results, but it probably should. We have already discussed standardizing the format of the results.

The suggestions above will be major breaking which changes the way the benchmark interface works as well as how results are stored. 

supersedes: #319
fixes: #251


mteb/abstasks/languages.py
```diff
@@ -0,0 +1,17 @@
...
+
+import json
+from pathlib import Path
+
+path_to_lang_codes = Path(__file__).parent / "iso_639_3_to_language.json"
+path_to_lang_scripts = Path(__file__).parent / "iso_15924_to_script.json"
+
+
+with path_to_lang_codes.open("r") as f:
+    ISOCODE639_3_TO_LANGUAGE = json.load(f)
```
MartinBernstorff, line 14: I'd shorten this to ISO_TO_LANGUAGE.

mteb/abstasks/languages.py
```diff
@@ -0,0 +1,17 @@
...
+
+with path_to_lang_scripts.open("r") as f:
+    ISOCODE15924_TO_SCRIPT = json.load(f)
```
MartinBernstorff, line 17: Same as above.

mteb/abstasks/AbsTask.py
```diff
@@ -58,3 +58,24 @@ def evaluate(self, model, split="test"):
         :param split: Which datasplit to be used.
         """
         raise NotImplementedError
+
+    @property
+    def languages(self) -> set[str]:
```
MartinBernstorff, line 63: My impression is that providing these "shortcuts" tend to be annoying for maintenance, but your mileage may vary.

mteb/abstasks/TaskMetadata.py
```diff
@@ -92,7 +95,8 @@
 ]  # Allows the type to be a string, but ensures that the string is a valid date
 
 SPLIT_NAME = str
-
+LANGUAGE_SCRIPT_CODE = str  # a 3-letter ISO 639-3 language code followed by a 4-letter ISO 15924 script code (e.g. "eng-Latn")
```
MartinBernstorff, line 98: ISO_LANGUAGE_SCRIPT? Follows the naming of the other type aliases.

mteb/abstasks/TaskMetadata.py
```diff
@@ -180,3 +187,58 @@ def _check_dataset_revision_is_specified(cls, dataset):
                 "It is encourage to specify a dataset revision for reproducability"
             )
         return dataset
+
+    @field_validator("eval_langs")
+    def _check_eval_langs(cls, eval_langs):
+        """
+        This method checks that the eval_langs are specified as a list of languages.
+        """
+        if isinstance(eval_langs, list):
```
MartinBernstorff, line 196: If someone specifies a tuple, we are not covering. Perhaps have the first check a dict, the rest check sequences?

mteb/abstasks/TaskMetadata.py
```diff
@@ -180,3 +187,58 @@ def _check_dataset_revision_is_specified(cls, dataset):
...
+    @staticmethod
+    def check_language_code(code):
+        """
+        This method checks that the language code (e.g. "eng-Latn") is valid.
+        """
+        lang, script = code.split("-")
+        if lang not in ISOCODE639_3_TO_LANGUAGE:
+            raise ValueError(f"Invalid language code: {lang}")
+        if script not in ISOCODE15924_TO_SCRIPT:
+            raise ValueError(f"Invalid script code: {script}")
```
MartinBernstorff, line 214: Excellent check! Might be worth it to point the user towards the valid values? E.g. point them to the file containing all valid scripts/langs.

mteb/abstasks/TaskMetadata.py
```diff
@@ -180,3 +187,58 @@ def _check_dataset_revision_is_specified(cls, dataset):
...
+            for code in eval_langs:
+                cls.check_language_code(code)
+        elif isinstance(eval_langs, dict):
+            for langs in eval_langs.values():
+                for code in langs:
+                    cls.check_language_code(code)
+        return eval_langs
+
+    @staticmethod
+    def check_language_code(code):
```
MartinBernstorff, line 206: Private?

mteb/evaluation/MTEB.py
```diff
@@ -41,20 +41,14 @@ def __init__(
...
-        task_langs: list of str / None
-            List of languages to be evaluated. if None, all languages will be evaluated. ["en", "de"] will evaluate on "en", "de", "en-de" and "de-en"
-        tasks: list of str / None
-            List of tasks to be evaluated. If specified, we filter tasks based on `task_langs` only
-        version: int / None
-            Version of the benchmark to use. If None, latest is used
-        err_logs_path: str / None
-            Path to save error logs
+        Args:
+            task_types: List of task types (Clustering, Retrieval..) to be evaluated. If None, all tasks will be evaluated
```
MartinBernstorff, line 45: Since this takes a list of [str], we might want to specify where to find valid inputs.

mteb/evaluation/MTEB.py
```diff
@@ -41,20 +41,14 @@ def __init__(
...
+            task_categories: List of task categories (s2s, p2p..) to be evaluated. If None, all tasks will be evaluated
```
MartinBernstorff, line 46: Same as above.

mteb/evaluation/MTEB.py
MartinBernstorff: Ah, I see you only did formatting changes. Feel free to disregard if this is acceptable.

MartinBernstorff (approved):
Good work @Kenneth! Huge diff though, so might very well have missed something, but the stuff I noticed all seems very minor.

mteb/abstasks/AbsTask.py
KennethEnevoldsen: Annoying for maintenance and nice for use. My guess would be that it removes confusion due to the mapping dicts (leading to less issues). All a guess though. I will need to for:

```
MTEB(task_langs=[...]), currently uses just the user-specified splits, we should instead transfer it to use the standard language codes.
```
So rather than implement it in the MTEB object I believe it is better places here.

KennethEnevoldsen:
I will merge this in now due to the many merge conflicts I keep having to solve. Once it is in we can basically launch MMTEB as the other changes are only nice to have, but not need to have and don't cause issues for people adding datasets

Muennighoff:
Do we have to change the language codes in all existing result files?

KennethEnevoldsen:
@Muennighoff atm. I don't believe it changes the results file at all (still uses the hf_name for the multilingual dataset e.g. "fr-en") and for the monolingual datasets, it doesn't include a lang-id. The same is the case for MTEB(task_langs="fr-en"), which should use the same mappings as before and still match the hf-name.

PR #329, opened 2024-04-08
Improve logging when the revision is None

gariepyalex:
This small PR fixes the logging for the missing dataset revision:

- Use the current logger instead of the root logger
- Adds the path of the dataset to the log. Previously, let's say you have in your code base many datasets with missing revision, it would simply log the message a few times at import time without telling you which datasets are problematic.


KennethEnevoldsen (approved):

KennethEnevoldsen:
Thanks @gariepyalex

PR #330, opened 2024-04-09
Add OpenSubtitles Bitext Mining dataset

loicmagne:
Hey, I created a bitext mining task from the [OpenSubtitles](https://opus.nlpl.eu/OpenSubtitles/corpus/version/OpenSubtitles) dataset, which is derived from translations of movie subtitles. I think it could be a great addition as it contains "spoken" sentences which isn't really covered in the benchmark yet I think, and it contains a lot of under-represented languages

However, the dataset is pretty huge, it contains 1759 language pairs, for each language pairs I sampled 1000 sentences randomly. The resulting dataset isn't so big, ~300MB, but loading 1759 splits sequentially with load_dataset takes some time

Let me know wdyt @KennethEnevoldsen @Muennighoff , I'm aware that there are already concerns with the size of some tasks in the benchmark, should I try to reduce the size ?

Also I used the language codes from the original dataset but I can change them if needed

- [ ] I have added points for my submission to the [POINTS.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) file.


mteb/tasks/BitextMining/multilingual/OpenSubtitlesBitextMining.py
```diff
@@ -0,0 +1,1801 @@
...
+        type="BitextMining",
+        category="s2s",
+        eval_splits=["train"],
+        eval_langs=_LANGUAGES_PAIRS,
+        main_score="f1",
+        date=None,
+        form=["spoken"],
+        domains=["Spoken"],
+        task_subtypes=None,
+        license=None,
```
KennethEnevoldsen, line 1788: Needed

mteb/tasks/BitextMining/multilingual/OpenSubtitlesBitextMining.py
```diff
@@ -0,0 +1,1801 @@
...
+        description="Collection of translated movie subtitles from http://www.opensubtitles.org/",
+        reference="http://www.opensubtitles.org/",
+        type="BitextMining",
+        category="s2s",
+        eval_splits=["train"],
+        eval_langs=_LANGUAGES_PAIRS,
+        main_score="f1",
+        date=None,
+        form=["spoken"],
+        domains=["Spoken"],
```
KennethEnevoldsen, line 1786:
```suggestion
        domains=["Spoken", "Fiction", "Non-fiction"]
```

mteb/tasks/BitextMining/multilingual/OpenSubtitlesBitextMining.py
```diff
@@ -0,0 +1,1801 @@
...
+            "revision": "ba04bfee78ae33527649494c5d53b0e0486eb698",
+        },
+        description="Collection of translated movie subtitles from http://www.opensubtitles.org/",
+        reference="http://www.opensubtitles.org/",
+        type="BitextMining",
+        category="s2s",
+        eval_splits=["train"],
+        eval_langs=_LANGUAGES_PAIRS,
+        main_score="f1",
+        date=None,
```
KennethEnevoldsen, line 1784: Should be added (as a range from earliest to latest)

mteb/tasks/BitextMining/multilingual/OpenSubtitlesBitextMining.py
```diff
@@ -0,0 +1,1801 @@
...
+        form=["spoken"],
+        domains=["Spoken"],
+        task_subtypes=None,
+        license=None,
+        socioeconomic_status=None,
+        annotations_creators="derived",
+        dialect=None,
+        text_creation=None,
```
KennethEnevoldsen, line 1792: "found"

mteb/tasks/BitextMining/multilingual/OpenSubtitlesBitextMining.py
```diff
@@ -0,0 +1,1801 @@
...
+        bibtex_citation="""@article{lison2016opensubtitles2016,
+  title={Opensubtitles2016: Extracting large parallel corpora from movie and tv subtitles},
+  author={Lison, Pierre and Tiedemann, J{\"o}rg},
+  year={2016},
+  publisher={European Language Resources Association}
+}""",
+        n_samples={"train": 1000},
```
KennethEnevoldsen, line 1799: I believe it is 1000xsplits in this case

KennethEnevoldsen (commented):
@loicmagne reducing the size would be great. I would run the two models on the current dataset and then try to see how much you can reduce the sample size without change the performance estimate. I would try to see if 100 samples could do.

You might also want to remove pairs which are too easy (could e.g. imagine "da-no" is quite easy).

loicmagne:
Ok I'll run both models and see how I can reduce the size

loicmagne:
I ran both models on the original dataset, based on the results I kept the 250 language pairs with the lowest scores, excluding some very low score that were due to poor data pairing and parsing in the original subtitles

The dataset now has 250k sentences in total but they are pretty short (28 char average) so I think it's better to keep 1000 sentences per language

perfs

KennethEnevoldsen:
Wonderful @loicmagne, this looks good. Is this plot using all the available samples?

In general, I would be more interested in wider coverage than many examples. So 250 sentences pr. 1000 languages are better than 1000 sentences for 250 languages.

One way to see if this influences performance too much is to do the same plot using 250 and 1000 examples pr. language. We are naturally interested in keeping it as small as possible.

loicmagne:
I see, the reason I tried to reduce the number of language pairs is that there is an overhead on the load_dataset method which makes loading a lot of dataset very slow, just loading the original 1700 language pairs would take 15+ minutes even with 100 sentences per language pairs

But I guess in the end the bottleneck for large models will be running inference on sentences so it makes sense to reduce the number of sentences, I'll check how significant are the performance changes when reducing the amount of samples

KennethEnevoldsen:
Hmm, that seems odd. Is there an issue on datasets for this? The best I can find is https://github.com/huggingface/datasets/issues/5499, which seems to suggest it is 4s from loading a cached dataset (still leave us at a high number but better). How is the dataset stored in the cache? (hopefully as .arrow and not .jsonl)

mteb/tasks/BitextMining/multilingual/OpenSubtitles250BitextMining.py
```diff
@@ -0,0 +1,292 @@
+from __future__ import annotations
+
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+from ....abstasks import AbsTaskBitextMining, CrosslingualTask
+
+_LANGUAGES_PAIRS = [
```
KennethEnevoldsen, line 7: #326 added a consistent language standard for MMTEB (which is wonderful as before it was hard to aggregate the scores meaningfully).

However that mean that this list will have to be a mapping e.g.:

```python
# your subset: [languages in that subset using a standard language code of iso 639-3 and ISO script code]
{"da-en": ["dan-Latn", "eng-Latn"],...}
```
I realize this is a bit frustrating during a PR, so took the liberty to create a language map for you, which should get you over the frustrating part:

```
lang_map = {
    "ka": "kat-Geor",
    "ml": "mal-Mlym",
    "br": "bre-Latn",
    "sr": "srp-Cyrl",
    "bg": "bul-Cyrl",
    "kk": "kaz-Cyrl",
    "lv": "lav-Latn",
    "sk": "slk-Latn",
    "fi": "fin-Latn",
    "zh": "zho-Hans",  # Defaulting to Simplified Chinese; special handling for regions
    "nl": "nld-Latn",
    "vi": "vie-Latn",
    "ja": "jpn-Jpan",
    "sv": "swe-Latn",
    "et": "est-Latn",
    "hu": "hun-Latn",
    "eo": "epo-Latn",
    "tr": "tur-Latn",
    "tl": "tgl-Latn",
    "te": "tel-Telu",
    "hr": "hrv-Latn",
    "si": "sin-Sinh",
    "cs": "ces-Latn",
    "is": "isl-Latn",
    "ro": "ron-Latn",
    "de": "deu-Latn",
    "fr": "fra-Latn",
    "hy": "hye-Armn",
    "no": "nor-Latn",
    "da": "dan-Latn",
    "en": "eng-Latn",
    "lt": "lit-Latn",
    "ur": "urd-Arab",
    "id": "ind-Latn",
    "fa": "fas-Arab",
    "uk": "ukr-Cyrl",
    "el": "ell-Grek",
    "bn": "ben-Beng",
    "hi": "hin-Deva",
    "ar": "ara-Arab",
    "he": "heb-Hebr",
    "gl": "glg-Latn",
    "pt": "por-Latn",
    "es": "spa-Latn",
    "th": "tha-Thai",
    "ko": "kor-Kore",
    "ru": "rus-Cyrl",
    "pl": "pol-Latn",
    "mk": "mkd-Cyrl",
    "ms": "msa-Latn",
    "ta": "tam-Taml",
    "af": "afr-Latn",
    "bg": "bul-Cyrl",
    "ca": "cat-Latn",
    "bs": "bos-Latn",
    "fi": "fin-Latn",
    # Additional script codes for zh variants
    "zh_cn": "zho-Hans", # Simplified Chinese
    "zh_tw": "zho-Hant", # Traditional Chinese
}
```

Cheers

loicmagne:
> Hmm, that seems odd. Is there an issue on datasets for this? The best I can find is [huggingface/datasets#5499](https://github.com/huggingface/datasets/issues/5499), which seems to suggest it is 4s from loading a cached dataset (still leave us at a high number but better). How is the dataset stored in the cache? (hopefully as .arrow and not .jsonl)

I just checked on a different computer for this dataset https://huggingface.co/datasets/loicmagne/open-subtitles-250-bitext-mining (250 pairs of languages, 1000 sentences each), it takes 15min to load all the subsets from the HF hub, and 12min to load it from the cache, so yeah there's a crazy overhead, I guess I'll open an issue on the datasets repo

KennethEnevoldsen:
If you examine the ~/.cache/huggingface folder, can you see how it is stored?

KennethEnevoldsen:
related to #342

loicmagne:
Ok I got the results for 1000 sentences per language pairs vs 256 sentences. The performance are strongly correlated even though the task is a bit easier with 256 sentences since there's less sentences to match with, I think 256 sentences is fine. The dataset doesn't look trivial either for most language pairs

256_1000s

Still, the loading time is currently very high even when reducing the number of language pairs, I opened an issue on the datasets repo https://github.com/huggingface/datasets/issues/6800 , I suggest we wait for responses before going further with this dataset

KennethEnevoldsen:
Thanks for working on this @loicmagne, I think it is a good idea to wait as well as it might influence how we cut det data.

KennethEnevoldsen:
@loicmagne, should we do something about this PR? (either fewer languages or the solution you suggested in the PR)

related to: https://github.com/huggingface/datasets/issues/6800

Issue #334, opened 2024-04-10
French MTEB is out of date


ManuelFay:
Hello,
I tried running the french scripts and lots of errors out of the box. 

- AlloprofRetrieval task has an updated dataset so it is required to modify code to access the dataset columns. Furthermore, the "train" split is the only available one so either the dataset should be updated, or both the eval split and the mteb_metadata.py script should be made so that results reflect that. Currently train results are ignored so not exportable in the leaderboard.

- OpusParcus Pair classification exports results on splits named test.full and validation.full. they also seem to be weird compared to old results in which many models have 100% scores so probably something off there... Similarly since the exports are not named like the rest, the export scripts ignore them.

- The Bsard retrieval task presents very small ndcg@10 scores for all models (best is under 3%). Might be completely correct but seems like either a weird task or a main metric that could be better chosen ?

- some datasets, notably Flores seem to run forever (and are not even reported in the main leaderboard french panel ?)

- Syntec retrieval dataset also seems to have been updated causing errors

- MLsum S2s and P2P tasks (french) require slight updates to select the dataset splits

I might forget a few things, best would be to run a mock model through the run_mteb_french script on the main branch, then run the mteb_metadata script, upload them to the HF spaces leaderboard, and verify everything runs smoothly...

In general, the issue here seems with dataset drift. More than fixing these issues for coherent score reporting, maybe it would be nice to have some e2e testing on the whole test suite, or locked datasets for testing.
 The issues are not long to fix but make reported scores incoherent and somewhat arbitrary...

Some models also might need to be retested given the weird 100% scores on the pair classification tasks ...

Thanks for the great work on the leaderboard,
Best

KennethEnevoldsen:
Thanks for the report @ManuelFay

@Muennighoff can you fix 1?

@imenelydiaker the other questions  seems best passed on to you.

### Re. testing on mteb:

We are currently heavily working on updating mteb, which includes updating the tests suite. It has already seen notable improvements (e.g. checking that dataset + revision is valid). 

Currently, most benchmarks are so large that running a model through the benchmark would take way too long for the tests to be feasible. We are currently examining approaches for reducing the size of the benchmark, though we realize that this would invalidate all previous results, thus we take these changes quite serious.

ManuelFay:
I am not proposing a PR is because a lot of my fixes seem hacky and I am not sure I made the correct choices (typically, every time I force using the train set, I am sure it is better fixed at the dataset level).

[...]

PR #341, opened 2024-04-11
Removing bitext mining tasks from fr evaluation script

imenelydiaker:
Related to this issue #334
 
This PR removes bitext mining tasks from the French evaluation script, since these tasks need two input languages to run (which was making Flores run forever btw). They should be run separately. 
The scores are also not included in the French leaderboard since it concerns 2 languages and not one.


KennethEnevoldsen (approved):
Bitext mining of course evaluates the models ability to pair-target language with others and might therefore still be relevant, though it might be relevant to e.g. include only one bitext task.

Whatever you decide it fine with me (for the scandinavian language we also don't include bitext from danish to other languages)

imenelydiaker:
Merging, the Bitext Mining task has its own tab on the leaderboard, let's keep it this way.

PR #344, opened 2024-04-11
docs: Added points for previous submissions

KennethEnevoldsen:
Added points for previous submissions

For a description of points for each PR see commit messages below.

Tagging the relevant persons here in case I have missed something (in a few cases I am unaware if the tasks were the first of their kind for the given language, and thus should gain a bonus point).

@guenthermi, @Muennighoff, @rasdani, @PhilipMay, @slvnwhrl, @staoxiao, @NouamaneTazi, @rafalposwiata, @violenil, @hanhainebula, @taeminlee.

I have added all the points to the main contributor, but assuming that some of the PRs have multiple contributors we can split it out.


Muennighoff (approved):
LGTM I think this is fine - If it's multiple contributors then they can just let us know and we'll split it up I think!
Also if there's a mistake or something, they can just submit a PR to fix it. Most important is to show that we made an effort & care!

imenelydiaker:
I agree with @Muennighoff, let's merge!

PR #350, opened 2024-04-12
Add neuclir

orionw:

Reason for dataset addition:
NeuCLIR 2022 and 2023 are TREC datasets from the [NeuCLIR](https://neuclir.github.io/) track that can be used for the `Retrieval` abstract task. It has high quality annotations and different corpora and qrels for each of three languages: Russian, Farsi, and Chinese (hence why I made them different datasets per language). 

There are a few other datasets in NeuCLIR (technical retrieval) that I can also add, but wanted to go through the PR process first to make sure I got it right. 

I'm not sure about the points, so I didn't modify that file yet. @KennethEnevoldsen any help? Seems like there is no Farsi or Russian in Retrieval yet so 4 * 2 bonus + 6 datasets * 2 = 20? 

One clarification point: NeuCLIR uses nDCG@20, but MTEB does not support that.  Should I add that MTEB or should we use nDCG@10 which will be incomparable to the published results on the task? @Muennighoff 

- [x] I have added points for my submission to the [POINTS.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) file.


Muennighoff:
> One clarification point: NeuCLIR uses nDCG@20, but MTEB does not support that. Should I add that MTEB or should we use nDCG@10 which will be incomparable to the published results on the task?

I think we should switch to nDCG@20 for this dataset then! The evaluation etc. supports it - I think it is just a matter of passing the k_values through somewhere and adding 20 to them for this dataset. Maybe a good solution would be if the code just checks if the last number in the main score i.e. in `ndcg_at_20` is not in the current k_values and adds it if so?

Lmk if someone disagrees!

KennethEnevoldsen:
> NeuCLIR 2022 and 2023 are TREC datasets from the [NeuCLIR](https://neuclir.github.io/) track that can be used for the Retrieval abstract task. It has high quality annotations and different corpora and qrels for each of three languages: Russian, Farsi, and Chinese (hence why I made them different datasets per language).

I believe this can simply be added as one dataset using the MultilingualTask abstract object. This requires that the dataset is in the format: {language, {split: DATA}}. See e.g. other multilingual retrieval tasks.

> I'm not sure about the points, so I didn't modify that file yet. @KennethEnevoldsen any help? Seems like there is no Farsi or Russian in Retrieval yet so 4 * 2 bonus + 6 datasets * 2 = 20?

Given the language restructering it would make it 2 (for the dataset) and 8 bonus points (russian + farsi) x retrieval. So 10 points.

> I think we should switch to nDCG@20

I completely agree.

mteb/tasks/Retrieval/fas/NeuCLIR2022FasRetrieval.py
```diff
@@ -0,0 +1,34 @@
...
+        description="The task involves identifying and retrieving the documents that are relevant to the queries.",
+        reference="https://neuclir.github.io/",
+        dataset={
+            "path": "mteb/neuclir-2022-fas",
+            "revision": "fa3b9068b9318d13ed5f065d9bbd5c8aff2dcfec",
+        },
+        type="Retrieval",
+        category="s2p",
+        eval_splits=["test"],
+        eval_langs=["fas-Arab"],
```
KennethEnevoldsen, line 20: Assuming you combine these datasets you will have to specify eval_langs as a dict:
```suggestion
        eval_langs={"fas": "fas-Arab", 
                              "rus": ...},
```
Where rus and fas are the language keys used for the dataset. You can also do a list if you use the exact same keys.

mteb/tasks/Retrieval/fas/NeuCLIR2022FasRetrieval.py
```diff
@@ -0,0 +1,34 @@
...
+        domains=None,
+        task_subtypes=None,
+        license=None,
+        socioeconomic_status=None,
+        annotations_creators=None,
+        dialect=None,
+        text_creation=None,
+        bibtex_citation=None,
+        n_samples=None,
+        avg_character_length=None,
```
KennethEnevoldsen, line 33: These will have to be specified as well. Let me know if you have any question related to these.

mteb/tasks/Retrieval/zho/NeuCLIR2023ZhoRetrieval.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import annotations
+
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+from ....abstasks.AbsTaskRetrieval import AbsTaskRetrieval
+
+
+class NeuCLIR2023ZhoRetrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="NeuCLIRZho2023",
```
KennethEnevoldsen, line 10: What is the difference between 2023 and 2022? any reason to not just use 2023?

KennethEnevoldsen (changes requested):

orionw:
Thanks @KennethEnevoldsen!

> I believe this can simply be added as one dataset using the MultilingualTask abstract object. This requires that the dataset is in the format: {language, {split: DATA}}. See e.g. other multilingual retrieval tasks.

You think one dataset instead of two?  There are two different years of TREC evaluation (2022, 2023) with different queries and different published results - I would probably recommend 2 unless I'm missing something.

Any pointers to a huggingface dataset that supports that? I only see JinaAI's which use a rather convoluted format.

mteb/tasks/Retrieval/zho/NeuCLIR2023ZhoRetrieval.py
orionw: Each one is a different evaluation set for that year's shared task (which is what a TREC track is)

orionw: If we think there's too much NeuCLIR we can remove one of the years, but they are distinct

mteb/tasks/Retrieval/fas/NeuCLIR2022FasRetrieval.py
orionw: It seems like most of these aren't filled out in the other datasets so I left them blank.. Many of them I don't have the answer for either. I will add as many as I know!

orionw: Do you have a reference to what some of these are?

`annotations_creators` could be qrels or documents or queries?
Not sure what socioeconomic status indicates?

mteb/tasks/Retrieval/zho/NeuCLIR2023ZhoRetrieval.py
KennethEnevoldsen: Thanks for letting me know. Seem like the organizers -- at least as I understand it in the [docs](https://neuclir.github.io/2023#data) -- combined 2022 into 2023 (at least partly, they used it as the dev set). This seems to suggest to me that the datasets are similar enough to warrant using only the datasets. 

@imenelydiaker what are your thoughts on this?

mteb/tasks/Retrieval/fas/NeuCLIR2022FasRetrieval.py
KennethEnevoldsen: > It seems like most of these aren't filled out in the other datasets so I left them blank.. Many of them I don't have the answer for either. I will add as many as I know!

You are completely correct, older datasets are missing their metadata, but new ones should have it.

> Do you have a reference to what some of these are?

You can find a description of the metadata object [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object). Feel free to let me know if the documentation is missing.

> annotations_creators could be qrels or documents or queries? Not sure what socioeconomic status indicates?

Socioeconomic status for e.g. legal documents is "high" (lawyers are typically quite well of), the same goes for text by similar socio-demographic groups (politicians, doctors etc.). Web is often "mixed". The annotation is there as many NLP resources rarely targets low-income groups.

Annotation creators is who generated the annotations (e.g. expert annotated, found, etc.)

mteb/tasks/Retrieval/zho/NeuCLIR2023ZhoRetrieval.py
orionw: I don't think the docs say that- do you mind posting the quote?  

The queries and qrels are different as I said above; the collection is shared but so are retrieval datasets that use Wikipedia (FEVER and NQ both) or MSMarco (TREC DL 19, TREC DL 20, etc.).  As someone who is advised by the creators of this dataset and who uploaded it, I can promise you the datasets are different :) 

I can see an argument for too much data of one style though! If we want less of it, I can understand that.

Alternatively though MMTEB (and MTEB) already has a lot of Wikipedia-style automatically generated data and could benefit from TREC-style data (which is human judged rather than automatically created and has much deeper relevance judgements).

KennethEnevoldsen: > I don't think the docs say they are the same- do you mind posting the quote? They do say they you can use it as a dev set if you want to get better performance on the 2023 version but they are distinct datasets -- no more than I would say FEVER is a dev set just because training on it makes you do better on a different fact checking dataset.

At least as I understand it; NeuCLIR 2023 includes two dev sets one of which is the 2022 eval set:

> The second development set comprises the NeuCLIR 2022 evaluation set.

But I agree with your point. The two splits you compare are not the same. I am sorry if it came across that way.

> I think the biggest reason to not put them as a single dataset is that there are published papers reporting scores on each of them separately. Here are [papers reporting results](https://scholar.google.com/scholar?cites=10465691959091815374&as_sdt=20000005&sciodt=0,21&hl=en) for 2022 and different ones for [2023](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C21&q=neuclir+track+2023+trec&btnG=).

I am not arguing that we combine it into one dataset. I am wondering if the two datasets test the same underlying phenomena. If that is the case, we might consider whether it is worth adding both of them in or whether one is fine. The MTEB is already quite large and we want to make sure that we make the benchmark unpractical to work with. 

I am atm. leaning toward including both datasets. If the datasets is then later revealed to be highly correlated we might consider not adding both datasets to the benchmark list, but I believe that is better to tested empirically.

KennethEnevoldsen: Let's roll with two dataset 🚀

mteb/tasks/Retrieval/fas/NeuCLIR2022FasRetrieval.py
orionw: Thanks, this was very helpful! 

Quick question on the `avg_character_length` and `n_samples` fields, are they the docs/queries/qrels?

KennethEnevoldsen: Ah yes that is a bit unclear atm.: query-document pairs (updated the taskmetadata docstring to also reflect this: #353)

orionw: Thanks! I see how for different tasks it can be different.  

Just to confirm, for retrieval you mean the qrels (the query-document relevance assessments, which can be > 1 per query)?  Does that include non-relevant judgements or only the relevant judgements?

For avg_character_length, I am unclear on what this means ("query-document" pairs meaning the concatenation of the query with each relevant document)?

KennethEnevoldsen: I have previously done it by combining them into one pile and calculating the mean, but the concatenation of the two might be easier to interpret. We should probably create a method on the AbstaskRetrieval to make it consistent. Something you might want to @orionw?

orionw: What do you mean by "create a method on the AbstaskRetrieval"?  Like a literal code method or just a way to do it?

I would suggest probably making it optionally a tuple, of (length query, length document, # of relevance judgements) as that's usually all the descriptive stats people want.  Or if you want it simpler, perhaps just average query + doc length for relevant docs.

mteb/evaluation/MTEB.py
```diff
@@ -295,6 +295,15 @@ def run(
                 logger.info(f"Loading dataset for {task.metadata_dict['name']}")
                 task.load_data(eval_splits=task_eval_splits, **kwargs)
 
+                # be sure the k_values includes the ones in the task
```
orionw, line 298: This is where I check for other non-default k-values (like 20). It seemed like the code for this had to go in this section so it could interact with the task and the evaluator.  

I could make it specific to the abstract task though, but since we have kwargs I think it won't affect the other classes.  

Seem good or any suggestions @KennethEnevoldsen?

mteb/tasks/Retrieval/fas/NeuCLIR2022FasRetrieval.py
KennethEnevoldsen: > Like a literal code method or just a way to do it?

Yes something like:

```
MyRetrievalTask().get_length()
MyRetrievalTask().get_n_samples()
```

 > Or if you want it simpler, perhaps just average query + doc length for relevant docs.
 
I would go for this. For nuanced metrics can always be calculated from the data (this is mostly to get is an overview of the size of the datasets)

mteb/evaluation/MTEB.py
KennethEnevoldsen: Couldn't we add 20 to the default list of metrics calculated (then it would calculate it for all datasets, but I  don't think it takes too long?) - it would also avoid the added complexity here.

orionw: I'm ambivalent.  My guess is most datasets use an existing number that is a power of 10, so I'll remove this code and we can add it back in if we see other datasets with unusual metrics (nDCG @ 30 or something).

mteb/tasks/Retrieval/fas/NeuCLIR2022FasRetrieval.py
orionw: Okay added them. Thanks!

KennethEnevoldsen:
Looks good @orionw. I moved the metadata calculation to the abstask, added a test for it, and updated the missing metadata. Everything looks good now so will set it to merge.

KennethEnevoldsen (approved):

orionw:
Thanks @KennethEnevoldsen and @Muennighoff for your patience and feedback as I learn the ropes!

KennethEnevoldsen:
No problem @orionw, this PR looks really good and making the last changes before merging it in saves me a bit on future merge conflicts ;)

Issue #354, opened 2024-04-12
Aggregating MMTEB datasets


orionw:
Some MMTEB datasets include only some of their respective many languages.

For example [Multilingual Miracl](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/multilingual/MIRACLRetrieval.py) in MMTEB contains only German and Spanish although it has many more.   Others like [Korean Mr.TyDI](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/kor/KoMrtydi.py) are included even though it is multilingual. Same with [Korean Miracl](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/kor/KoMiracl.py) which is included separately from the multilingual Miracl.

These should be consolidated into multilingual datasets with all their available languages (18 in Miracl and 11 in Mr. TyDI).

I can start working on this next week, unless someone beats me to it.

orionw:
Assuming @KennethEnevoldsen agrees

izhx:
[Nandan Thakur](https://github.com/thakur-nandan) may have started   https://github.com/embeddings-benchmark/mteb/issues/198#issuecomment-2050189570 on Miracl

Merging [Mr.TyDI](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/kor/KoMrtydi.py) seems like a good idea. I think we would prefer not to run these massive retrieval tasks too many times...

KennethEnevoldsen:
> Assuming @KennethEnevoldsen agrees

Completely agree, we should merge the relevant datasets. Feel free to merge Mr.TyDI

KennethEnevoldsen:
> I think we would prefer not to run these massive retrieval tasks too many times...

If the dataset is massive we might consider reducing the size as well.

orionw:
On a closer reading of the MiRACL paper, it seems like they expanded the Mr. TyDi work with more annotations.  Assuming that is true and they have the same query set (@thakur-nandan) I would suggest we remove any Mr. TyDi datasets and only include MiRACL.

thakur-nandan:
Hi @orionw and others, I wasn't aware of this issue thread. including @crystina-z to the thread as well.

[...]

PR #363, opened 2024-04-13
Remove Mr. TyDi Datasets

orionw:
MiRACL datasets expand Mr. TyDi datasets. It seems redundant to have the same queries with fewer relevance annotations on the same corpus.

From #354


mteb/tasks/Retrieval/kor/KoMrtydi.py
KennethEnevoldsen, line 1: We will need to remove it from the Korean benchmark as well. It is not on the leaderboard so it should be fine.

KennethEnevoldsen (changes requested):
Only one change, otherwise looks good

mteb/tasks/Retrieval/kor/KoMrtydi.py
orionw: I think I did that, lmk if `run_korean_mteb` wasn't the file you were referring to.

orionw: Otherwise, I think it is updated and ready to go now.

PR #369, opened 2024-04-14
fix: Add Arabic Restaurant Reviews 

mmhamdy:

Reason for dataset addition:

This PR adds a dataset for [Arabic restaurant reviews](https://huggingface.co/datasets/ar_res_reviews) sentiment classification.

- [ ] I have added points for my submission to the [POINTS.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) file.


mteb/tasks/Classification/ara/RestaurantReviewSentimentClassification.py
```diff
@@ -0,0 +1,49 @@
...
+        name="RestaurantReviewSentimentClassification",
+        dataset={
+            "path": "ar_res_reviews",
+            "revision": "d51bf2435d030e0041344f576c5e8d7154828977",
+        },
+        description="Dataset of 8364 restaurant reviews from qaym.com in Arabic for sentiment analysis",
+        reference="https://link.springer.com/chapter/10.1007/978-3-319-18117-2_2",
+        type="Classification",
+        category="s2s",
+        eval_splits=["test"],
```
imenelydiaker, line 18: I see in the hub that there is only a `train` split for this dataset, loading a `test` split will raise an error imo... How did you manage to get the benchmark run on the `test` split?

results/intfloat__multilingual-e5-small/RestaurantReviewSentimentClassification.json
```diff
@@ -0,0 +1,15 @@
+{
+  "dataset_revision": "d51bf2435d030e0041344f576c5e8d7154828977",
+  "mteb_dataset_name": "RestaurantReviewSentimentClassification",
+  "mteb_version": "1.6.3",
+  "test": {
```
imenelydiaker, line 5: Results seem okay, but I wonder how it worked on the `test` split although only a `train` split is available on HF?

mteb/tasks/Classification/ara/RestaurantReviewSentimentClassification.py
mmhamdy: Hi @imenelydiaker, I did notice that and I used the `train` split at first. But since the data loading function applies the data transform after loading the dataset from the hub, the evaluation uses the `train` split for testing (which is what I've encountered). Here we can specify a `test` split as I've accounted for that and created a train/test split in the dataset transform.

KennethEnevoldsen: This should then be "train" then (to indicate that it is data from the train split on HF that is being used). I believe you use the test train split to downsample the data? 

I would instead shuffle with a seed and select the first 2048 samples.

results/intfloat__multilingual-e5-small/RestaurantReviewSentimentClassification.json
KennethEnevoldsen, line 1: Will have to be rerun on the new "train" split

mmhamdy:
Hi @imenelydiaker @KennethEnevoldsen, I've updated the metadata eval split and reran evaluations. sorry for the misunderstanding.

KennethEnevoldsen:
> Hi @imenelydiaker @KennethEnevoldsen, I've updated the metadata eval split and reran evaluations. sorry for the misunderstanding.

No worries! 

Almost everything is looking good now. If you can address the two issues above I believe we should be there.

imenelydiaker (commented):
I'm not a fan of creating a test split out of a train split that is available on HF, imo it should be done on the source dataset. Else, the whole train split can be used to train models and then when evaluating with mteb on this same dataset we will encourage data leakage because the test set is created from the train one... 
I'd prefer to open a PR on the HF dataset repository and suggest to add a test split. Wdyt @KennethEnevoldsen ?

mmhamdy:
> I'm not a fan of creating a test split out of a train split that is available on HF, imo it should be done on the source dataset. Else, the whole train split can be used to train models and then when evaluating with mteb on this same dataset we will encourage data leakage because the test set is created from the train one.

I agree. But a large number of datasets on the hub have only a train split and most of them have no activity. In my opinion, a possible solution is to host the dataset after creating the splits. This will also ensure stability.

KennethEnevoldsen:
I think using the "train" split is valid (multiple datasets use the train split), the best solution would naturally be to rehost it using only the desired splits (which could be new).

Rehosting with a train/test split which has the downside of seeming like we are testing something other than the original "train" split, which we naturally can't. 

@imenelydiaker I agree with you that relabelled in the dataset_transform as a "test" split is problematic, but as long is it is stated as "train" I believe it is fine.

imenelydiaker:
Okay let's keep the train split only.

KennethEnevoldsen:
Merging this into a branch to add the points, then I will merge it into main

PR #379, opened 2024-04-16
Adding Turkish STSBenchmark

asparius:


KennethEnevoldsen (commented):
Hi, thanks for taking the time to make this contribution. To avoid many machine-translated benchmarks, we specifically disallow them:

> We do not include any machine-translated (without verification) datasets in the benchmark.

source: [adding a dataset](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md)

However, we would gladly accept non-translated datasets e.g. could imagine this repo has a few tasks: https://github.com/alisafaya/mukayese?tab=readme-ov-file

e.g. news-cat for clustering
J

asparius:
Thanks for the response, I am not also big fan of translation datasets however I thought it was a verified translation since it was released via a publication. I will be adding mukayese and other non-translated Turkish datasets soon.

KennethEnevoldsen:
> Thanks for the response, I am not also big fan of translation datasets however I thought it was a verified translation since it was released via a publication. I will be adding mukayese and other non-translated Turkish datasets soon.

If you find some evidence of validation (e.g. quote from article). Then we can move on with the dataset. Otherwise, we can focus on the other datasets.

KennethEnevoldsen:
> decent and possible errors will not cause major errors.

A right, but nothing has been done to remedy the errors?

I think, given that there seems to be good alternative datasets, that I would focus on non-translated datasets.

PR #386, opened 2024-04-16
Add TedTalks dataset

davidstap:

Reason for dataset addition:

- [x] I have added points for my submission to the [POINTS.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) file.


mteb/tasks/BitextMining/multilingual/TedTalksBitextMining.py
```diff
@@ -0,0 +1,178 @@
...
+        form=None,
+        domains=["Spoken"],
+        task_subtypes=None,
+        license=None,
+        socioeconomic_status=None,
+        annotations_creators=None,
+        dialect=None,
+        text_creation=None,
+        n_samples={"train": _N},
+        avg_character_length={},
```
KennethEnevoldsen, line 121: these need to be filled out - documentation can be found in taskmetadata object or in the "adding a dataset"-guide.

KennethEnevoldsen (commented):
Looking good. Just a minor comment. Feel free to tag one of us once you are ready for the full review.

davidstap:
This is ready for review now @KennethEnevoldsen . 

This task has a lot of language coverage: a total of 3080 directions. I subsampled the data to 256, which leads to very similar results as 512. Using more data seems unnecessary to me.

I have added points: 2 for new dataset + 2x3 for new languages in BitextMining task (fas_Arab, kur_Arab, msa_Latn) = 8 points.

Thanks for this community effort, it's nice to play a small role in it :)

mteb/tasks/BitextMining/multilingual/TedTalksBitextMining.py
```diff
@@ -0,0 +1,174 @@
...
+    for x in _ISO6393_to_ISO6391.keys():
+        if "-" not in x:
+            for y in _ISO6393_to_ISO6391.keys():
+                if x != y:
+                    pair = f"{x}-{y}"
+                    hf_lang_subset2isolang[pair] = [
+                        x.replace("_", "-"),
+                        y.replace("_", "-"),
+                    ]
+    return hf_lang_subset2isolang
```
imenelydiaker, line 88: This function is the same as in [Flores](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/BitextMining/multilingual/FloresBitextMining.py) right? Maybe we should define it in AbsTaskBitextMining to allow reusing it eaily and not duplicate the code, wdyt?

KennethEnevoldsen: it seems dependent mostly on the label format which is dataset-specific. I would keep it separate as changing one dataset should not lead to changes in another (this makes them independent).

mteb/tasks/BitextMining/multilingual/TedTalksBitextMining.py
```diff
@@ -0,0 +1,174 @@
...
+        reference="https://huggingface.co/datasets/davidstap/ted_talks",
+        type="BitextMining",
+        category="s2s",
+        eval_splits=_SPLIT,
+        eval_langs=_EVAL_LANGS,
+        main_score="f1",
+        date=("2018-06-01", "2018-06-01"),
+        form=["written"],
+        domains=["Spoken"],
+        task_subtypes=None,
```
KennethEnevoldsen, line 112:
```suggestion
        task_subtypes=[],
```

mteb/tasks/BitextMining/multilingual/TedTalksBitextMining.py
```diff
@@ -0,0 +1,174 @@
...
+        license="CC-BY-NC-ND-4.0",
+        socioeconomic_status="medium",
+        annotations_creators="human-annotated",
+        dialect=None,
```
KennethEnevoldsen, line 116:
```suggestion
        dialect=[],
```
This is just to differentiate between annotated (and there is none:  []) and not annotated (None).

KennethEnevoldsen (commented):
Thanks for this great addition @davidstap and I agree with the size of the dataset as well. 

Please add points for the reviewers as well (me and @imenelydiaker) and address the two minor corrections. Then we can get it merged it 🎉

davidstap:
Hey @imenelydiaker @KennethEnevoldsen thanks for your comments! Everything is clear, and I will make the necessary changes on Monday (currently on a cycling trip without access to computer).

mteb/tasks/BitextMining/multilingual/TedTalksBitextMining.py
imenelydiaker: @davidstap you can resolve this conversation 🙂

davidstap:
I made the necessary changes @imenelydiaker @KennethEnevoldsen 

Note that I also changed my points: new dataset (2 points) + 3 new languages for task (3 * 4 points) = 14 points. (I made a mistake earlier because I thought new language counted for 2 points instead of 4.

imenelydiaker (commented):
Following my comment about the point system in PR #409

davidstap:
I have added the points in new format now.

KennethEnevoldsen:
Set to auto-merge assuming tests pass

davidstap:
There seems to be an error with data loading, I'll try to debug tomorrow. LMK if you think there's an obvious error; I don't see it.

imenelydiaker:
Hello @davidstap, I've had a look at the dataset, it seems to use a loading script to load a tar file. Since you created the dataset, maybe we can update it to make better use of the `datasets` library? Have you tried to use subsets and splits (config names)?
One advantage of using splits, is that your dataset can be used in the leaderboard, default is loading train split from HF, and then you create validation and test splits with the script (as I understood it). If the dataset will be used in a benchmark it won't be used since it only contains a train split according to the metadata.
For the subsets, we can have on subset per language or pair pf language, it can be useful for users if they want to use one pair of language only.
Would you mind if I open a PR on your datasets repository?

davidstap:
Hi @imenelydiaker, thanks for your help!

Indeed, a tar file is loaded. The tar file contains train, validation, test splits (following http://phontron.com/data/ted_talks.tar.gz). The huggingface dataset I created supports splits (`train`, `validation`, `test`) and configurations (all possible language pair combinations, e.g. `ar_az`).

For the MMTEB PR I went for a subsample of the training split, because not all validation/test language combinations have data (some have 0 overlap), and I thought it would be nice to maximise language coverage.

Given this, what do you think is the best way forward?

mteb/tasks/BitextMining/multilingual/TedTalksBitextMining.py
```diff
@@ -0,0 +1,173 @@
...
+        form=["written"],
+        domains=["Spoken"],
+        task_subtypes=[],
+        license="CC-BY-NC-ND-4.0",
+        socioeconomic_status="medium",
+        annotations_creators="human-annotated",
+        dialect=[],
+        text_creation="created",
+        n_samples={"train": _N},
+        avg_character_length={},
```
KennethEnevoldsen, line 119: required

KennethEnevoldsen, line 119: This is required

mteb/tasks/BitextMining/multilingual/TedTalksBitextMining.py
```diff
@@ -0,0 +1,173 @@
...
+        date=("2018-06-01", "2018-06-01"),
+        form=["written"],
+        domains=["Spoken"],
+        task_subtypes=[],
+        license="CC-BY-NC-ND-4.0",
+        socioeconomic_status="medium",
+        annotations_creators="human-annotated",
+        dialect=[],
+        text_creation="created",
+        n_samples={"train": _N},
```
KennethEnevoldsen, line 118: How is this number ensured? seems like it downloads the entire "train" and uses all of it (I might be missing something)

KennethEnevoldsen:
@davidstap what I believe @imenelydiaker is suggesting is that the dataset can be uploaded to hugging face using the default apache arrow format instead of a custom file format loaded using a script.

mteb/tasks/BitextMining/multilingual/TedTalksBitextMining.py
davidstap: See line 157: argument `split=f"train[:{_N}]"` in `load_dataset` ensures this.

imenelydiaker:
> Hi @imenelydiaker, thanks for your help!
> 
> Indeed, a tar file is loaded. The tar file contains train, validation, test splits (following http://phontron.com/data/ted_talks.tar.gz). The huggingface dataset I created supports splits (`train`, `validation`, `test`) and configurations (all possible language pair combinations, e.g. `ar_az`).
> 
> For the MMTEB PR I went for a subsample of the training split, because not all validation/test language combinations have data (some have 0 overlap), and I thought it would be nice to maximise language coverage.
> 
> Given this, what do you think is the best way forward?

Just as @KennethEnevoldsen mentionned we can use parquet to load the dataset, and build subsets for each language and build all splits (train, test and validation). 
For the pairs with no translations, I don't really understand how it is handled, because I see in `extend_lang_pairs()` that all possible language pairs are built from `_ISO6393_to_ISO6391`, which results in more than 3000 language pairs

davidstap:
Based on your comments and suggestions I have done some additional investigations:
* The test error resulted from using `load_dataset` instead of `datasets.load_dataset`, so this is trivial to solve.
* Loading all 3080 language pairs is *extremely slow* due to large overhead. When the data is not cached, it takes ~150 min. When the data is cached, it takes X min (currently running). This is problematic because 1) `make test` is extremely slow as a result and 2) running benchmark on this dataset will be slow. The issue seems similar to #330, and is due to the overhead on `load_dataset` when loading a large number of configurations. I'm not deeply familiar with `datasets`, do you think this issue would be solved when using arrow instead? Converting everything to arrow will take some time, so I'm a bit hesitant to just try it.
* I did not realize that the `train` split would not be used for the benchmark, as pointed out by @imenelydiaker. Unfortunately, there are language combinations that do not have any data in `test` or `validation`. One solution would be to create a new modified dataset, where we put data from `train` in `test` and `validation` to ensure all splits have at least 256 sentences. Of course we then lose the splits from the original dataset for some languages, but the splits are pretty arbitrary anyway.

WDYT @KennethEnevoldsen  @imenelydiaker ?

KennethEnevoldsen:
> do you think this issue would be solved when using arrow instead? Converting everything to arrow will take some time, so I'm a bit hesitant to just try it.

Sadly not there is an issue on this on HF datasets (https://github.com/huggingface/datasets/issues/6800). I would add a comment in there if I were you and potentially try the solution suggested by @loicmagne.

> I did not realize that the train split would not be used for the benchmark, as pointed out by @imenelydiaker. 

You are free to use the train split if you want. As long as it is specified in the eval_splits there shouldn't be an issue. I would however downsample such that there are only 256 for each language (which does not seem to be the case now)

loicmagne:
Hey, just chiming in about the data loading issue

I've had similar problem for the [OpenSubtitles dataset](https://github.com/embeddings-benchmark/mteb/pull/330), which contained ~1600 language pairs. In my case, the [dataset](https://huggingface.co/datasets/loicmagne/open-subtitles-bitext-mining) is split into proper subsets for each languages and splits. Still the loading time is similarly very high, this is caused by an overhead from the datasets lib that performs verifications for each subsets https://github.com/huggingface/datasets/issues/5499

I haven't tried directly converting each files to Arrow format, but when the dataset is cached, it is cached in Arrow format, and even then loading the data takes some time. In my experiment it took 12 minutes to load 250 subsets which were cached as Arrow. So I think converting to Arrow would save a couple minutes but there would still be an overhead from network requests

There is a way to load all the subsets at once instead of inside a for loop https://github.com/huggingface/datasets/issues/6800 , which is much faster. However when doing that, all the subsets are merged together, so you lose the information about which sample belongs to which language, which we need for bitext mining

I don't have a perfect solution yet, but I'm planning to change the bitext mining "standard" data format, as it makes most of these tasks very time-consuming to evaluate currently

KennethEnevoldsen:
@loicmagne that sounds wonderful. If it ends up changing performance scores you might find the #481 PR relevant

PR #394, opened 2024-04-17
style: This should fail linting

KennethEnevoldsen:
Seems like linting check always pass, this is just to check the CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,16 @@ lint:
 	ruff format . 			# running ruff formatting
 	ruff check . --fix  	# running ruff linting
 
+lint:
+	@echo "--- 🧹 Running linters ---"
+	ruff format . 			# running ruff formatting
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,6 +12,16 @@ lint:
...
 
+lint:
+	@echo "--- 🧹 Running linters ---"
+	ruff format . 			# running ruff formatting
+	ruff check . --fix  	# running ruff linting
+
+lint-check:
+	@echo "--- 🧹 Check is project is linted ---"
+	ruff format . --check						    # running ruff formatting
+	ruff check **/*.py 						        # running ruff linting
```
KennethEnevoldsen: @MartinBernstorff, this seems to be the solution, it is simply linting that does not fail as it can just fix the files. I actually had this earlier, but removed it as it was unec.  *facepalm*. I have added a comment for it as well.

Issue #402, opened 2024-04-17
Fix MLSUM S2S and P2P tasks


imenelydiaker:
There seem to be a problem with MLSUM clustering tasks as mentionned in issue #334.

imenelydiaker:
Problem: after mteb refactoring the custom `load_dataset()` dunction was removed, which made the dataset unable to load.

PR #405, opened 2024-04-17
fix: add custom load dataset function for MLSUM tasks

imenelydiaker:
Fix for issue #402 and #334.
The custom load_dataset() function was removed, which caused an error. I'll open another PR to make this dataset multilingual and improve the loading process.

### DIFF
--- a/mteb/tasks/Clustering/fra/MLSUMClusteringP2P.py
+++ b/mteb/tasks/Clustering/fra/MLSUMClusteringP2P.py
@@ -45,10 +45,10 @@ class MLSUMClusteringP2P(AbsTaskClustering):
         if self.data_loaded:
             return
         self.dataset = datasets.load_dataset(
-            self.metadata.dataset['path'],
-            self.metadata.dataset['name'],
+            self.metadata.dataset["path"],
+            self.metadata.dataset["name"],
             split=self.metadata.eval_splits[0],
-            revision=self.metadata.dataset['revision'],
+            revision=self.metadata.dataset["revision"],
         )
         self.dataset_transform()
         self.data_loaded = True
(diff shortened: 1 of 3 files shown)


mteb/tasks/Clustering/fra/MLSUMClusteringP2P.py
```diff
@@ -38,6 +38,21 @@ class MLSUMClusteringP2P(AbsTaskClustering):
...
+    def load_data(self, **kwargs):
+        """
+        Load dataset from HuggingFace hub and convert it to the standard format.
+        """
+        if self.data_loaded:
+            return
+        self.dataset = datasets.load_dataset(
+            self.metadata.dataset["path"],
+            self.metadata.dataset["name"],
+            split=self.metadata.eval_splits[0],
```
MartinBernstorff, line 50: @KennethEnevoldsen In `AbsTask`, we pass `**self.metadata_dict["dataset"]` to the `load_dataset` function. Do we want to load all eval_splits instead? 

(I.e. rename the eval_split key to `split`, or add it when unpacking?)

MartinBernstorff:
Thanks for the fix @imenelydiaker, and sorry about adding an error!

This might be more systematic, I've added a comment for Kenneth.

mteb/tasks/Clustering/fra/MLSUMClusteringP2P.py
KennethEnevoldsen: It could save us quite a bit of downloaded data. So it seems like a great idea (though I would probably keep it as `eval_splits`. This is a separate PR though, which I know will break some datasets (as the dataset_transform assume the whole dataset).

KennethEnevoldsen (approved):
Thanks @imenelydiaker for fixing this! Please add points for the bugfix as well (2 points seem fair?)

KennethEnevoldsen:
Added points and set to automerge

imenelydiaker:
> Added points and set to automerge

Thanks @KennethEnevoldsen sorry for the delay 🙂

KennethEnevoldsen:
No worries, thanks for the work you put in!

PR #424, opened 2024-04-18
Faster Clustering: Doe Downsampling work?

KennethEnevoldsen:
So I did a few experiments with downsampling on some of the larger datasets. We seem to get quite consistent performance, but seems like it is consistently slightly higher scores for smaller samples, though the std seems to remain similar. 

This seems to indicate that we can't just keep the existing scores, but estimate them using lower sample size.


mteb/tasks/BitextMining/multilingual/BUCCBitextMining.py
KennethEnevoldsen, line 1: feel free to ignore this

mteb/abstasks/AbsTaskClustering.py
```diff
@@ -66,4 +66,4 @@ def _evaluate_monolingual(self, model, dataset, split="test", **kwargs):
 
         v_mean = np.mean(v_measures)
         v_std = np.std(v_measures)
-        return {"v_measure": v_mean, "v_measure_std": v_std}
+        return {"v_measure": v_mean, "v_measure_std": v_std, "v_measures": v_measures}
```
KennethEnevoldsen, line 69: to allow user to calculate ci etc.

tmp/slowest_segment.py
KennethEnevoldsen, line 1: feel free to ignore

Muennighoff:
Agree with what has been said - we should make sure we're not losing precision in the task & update all scores if they change👍

KennethEnevoldsen:
superseded by #481

PR #433, opened 2024-04-19
add SlovakSentimentClassification dataset

manandey:
- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] I have considered the size of the dataset and reduced it if it is too big (2048 examples is typically large enough for most tasks)
- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 
- [x] I have added points for my submission to the [POINTS.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) file.

PR #437, opened 2024-04-19
docs: Correct review points

KennethEnevoldsen:
Corrections from PR (@isaac-chung it seems like we both corrected your points). The 4 points from be was a 

#403


isaac-chung:
manandey should have 18 after https://github.com/embeddings-benchmark/mteb/pull/433/files

Issue #434, opened 2024-04-19
Add a multilabel classification task


x-tabdeveloping:
For certain datasets , such as [EURLEX](https://huggingface.co/datasets/coastalcph/multi_eurlex) it would be very useful to have a multilabel classification task in the benchmark.

x-tabdeveloping:
I can shoot a bullet but catch me if I fall @KennethEnevoldsen.

PR #440, opened 2024-04-19
WIP: Multilabel classification

x-tabdeveloping:
Working on #434.
I will still have to add a good test task, if anyone has one don't hesitate to comment.


isaac-chung:
Maybe this? https://huggingface.co/datasets/coastalcph/multi_eurlex

x-tabdeveloping:
I'll look into it, thanks @isaac-chung !!

mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -0,0 +1,146 @@
...
+        self.k = k
+        # Run metadata validation by instantiating addressing the attribute
+        # This is quite hacky. Ideally, this would be done in the constructor of
+        # each concrete task, but then we have to duplicate the __init__ method's
+        # interface.
+        if hasattr(self, "metadata"):
+            self.metadata
+
+    def _add_main_score(self, scores):
+        if self.metadata_dict["main_score"] in scores:
```
KennethEnevoldsen, line 57:
```suggestion
        if self.metadata.main_score in scores:
```
I would use the metadata (no dict) generally. This makes code changes easier in the future.

mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -0,0 +1,146 @@
...
+            scores.append(scores_exp)
+
+        if self.n_experiments == 1:
+            return scores[0]
+        else:
+            avg_scores = {k: np.mean([s[k] for s in scores]) for k in scores[0].keys()}
+            std_errors = {
+                k + "_stderr": np.std([s[k] for s in scores]) for k in scores[0].keys()
+            }
+            return {**avg_scores, **std_errors}
```
KennethEnevoldsen, line 130: Hmm, we have previously discussed creating a standard format for results. This is a reasonable time to discuss it (at least the standard format for parts of it) as the inner results keys have multiple scores.

Let us ignore the format of lang, split nesting (that is for another PR)

Generally, all tasks have a singular main score. However for this case we either have to aggregate them, have multiple main scores or produce multiple results files from one task.

For me it seems the most reasonable to produce a list of results (one for each label), which then in a final table will be displayed as MyNewTask(cat1), MyNewTask(cat2)?

mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -0,0 +1,146 @@
...
+            scores_exp, test_cache = evaluator(model, test_cache=test_cache)
+            scores.append(scores_exp)
+
+        if self.n_experiments == 1:
+            return scores[0]
+        else:
+            avg_scores = {k: np.mean([s[k] for s in scores]) for k in scores[0].keys()}
+            std_errors = {
+                k + "_stderr": np.std([s[k] for s in scores]) for k in scores[0].keys()
+            }
```
KennethEnevoldsen, line 129: I would love to also add the raw scores (e.g. to allow calculating win rates, CI etc.)

mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -0,0 +1,146 @@
...
+            None,
+        )  # we store idxs to make the shuffling reproducible
+        for i in range(self.n_experiments):
+            logger.info(
+                "=" * 10 + f" Experiment {i+1}/{self.n_experiments} " + "=" * 10
+            )
+            # Bootstrap `self.samples_per_label` samples per label for each split
+            X_sampled, y_sampled, idxs = self._undersample_data(
+                train_split["text"], train_split["label"], self.samples_per_label, idxs
+            )
```
KennethEnevoldsen, line 109: You could essentially sample the same and embed it twice which seems slightly inefficient, which is slightly annoying. Another solution would be so sample the ids first, embed them all at once, and then sample the embeddings. This also leads to less task switching.

KennethEnevoldsen (commented):
Looks really good @x-tabdeveloping. A few points of discussion, but testing it on a task seems like the best next step.

mteb/abstasks/AbsTaskMultilabelClassification.py
x-tabdeveloping: What should I do about these?
```python
        self.n_experiments = (
            n_experiments
            if n_experiments is not None
            else self.metadata_dict.get("n_experiments", 10)
        )
        self.samples_per_label = (
            samples_per_label
            if samples_per_label is not None
            else self.metadata_dict.get("samples_per_label", 8)
        )
```
Would it be smart to write that also over to `metadata`? Like this for instance:
```python
if n_experiments is None:
     try:
          self.n_experiments = self.metadata.n_experiments
     except AttributeError:
          self.n_experiments = 10
else:
     self.n_experiments = n_experiments
```
or perhaps a bit more concise but less readable:
```python
self.n_experiments = n_experiments or getattr(self.metadata, "n_experiments", 10)
```

mteb/abstasks/AbsTaskMultilabelClassification.py
x-tabdeveloping: Is this something I need to address in this PR, or should we postpone this for a separate issue of its own? If we decide to discuss it here, perhaps we should bring more people on board with this.

mteb/abstasks/AbsTaskMultilabelClassification.py
x-tabdeveloping: I have removed all references to `metadata_dict` can you check if you like it better now? @KennethEnevoldsen

x-tabdeveloping:
I'm currently in the process of adding EURLEX.

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -14,6 +20,75 @@
...
+                model.encode(self.sentences_test, batch_size=self.batch_size)
+            )
+            test_cache = X_test
+        else:
+            X_test = test_cache
+        for metric in ["cosine", "euclidean", "dot"]:
+            if metric == "dot":
+                metric = dot_distance
+            estimator = KNeighborsClassifier(n_neighbors=self.k, metric=metric)
+            classifier = MultiOutputClassifier(estimator, n_jobs=-1)
```
KennethEnevoldsen, line 73: Might be worth considering whether we might want "interacting" estimators or not. E.g. for related categories such as hate-speech / type of hate-speech classification.

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -14,6 +20,75 @@
...
+
+        self.k = k
+
+    def __call__(self, model, test_cache=None):
+        scores = {}
+        max_accuracy = 0
+        max_f1 = 0
+        max_ap = 0
+        X_train = np.asarray(
+            model.encode(self.sentences_train, batch_size=self.batch_size)
```
KennethEnevoldsen, line 60: I have actually moved this out of the evaluator and into the task in #481. This was mostly because it would allow me to encode all documents at once and just pass in the embeddings (which you can then sample from).

mteb/evaluation/evaluators/ClassificationEvaluator.py
KennethEnevoldsen: Sure thing let us. A simple solution from clf pipeline is to allow for multiple classifiers and just pick the best performing one.

x-tabdeveloping:
Currently this PR assumes that all labels in the classification are independent from each other.
This is due to `MultiOutputClassifier` from sklearn, which trains multiple independent classifiers for each label.

Some options we could consider that would fix this:
1. `ClassifierChain` which would be an optimal choice for hierarchical tasks or where the ordering of the labels is trivial. We have to be cautious to order the labels properly though, which might be a pain in a half to do, and I'm not sure whether this should be the specific tasks' or the `AbsTask`'s responsibility.
2. Using a neural network like `MLPClassifier` with multiple outputs. This would be a good option, because it does not need any ordering and does not assume the independence of labels, but it's waaaaaay slower than just using kNN and we would also lose a great deal of conceptual transparency. 

What do you guys think @KennethEnevoldsen @imenelydiaker @isaac-chung ?

mteb/evaluation/evaluators/ClassificationEvaluator.py
x-tabdeveloping: See my comment below

x-tabdeveloping:
I'm currently in the process of running MultiEURLEX on my machine, this might take a fair bit :D

KennethEnevoldsen:
My immediate assumption is just to go for simplicity and then we can always expand to simpler cases in the future.

mteb/abstasks/AbsTaskMultilabelClassification.py
x-tabdeveloping: Again, is this something we should address in this PR or in another one with multiple tasks at once?

x-tabdeveloping:
Regarding the points: Do we count `MultilabelClassification` as a new task for each language contained in EURLEX or should I only add bonus points for those languages that had no classification task prior to this?

x-tabdeveloping:
I have been running the task basically all day on UCloud on the two models, it takes a ridiculous amount of time.

KennethEnevoldsen:
@x-tabdeveloping let us discuss how to potentially speed it up monday

x-tabdeveloping:
Yea I left it running for days and it didn't complete for a single model, I will attempt to make it faster, come by at the office if you have ideas @KennethEnevoldsen

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -14,6 +19,66 @@
...
+        y_train,
+        embeddings_test,
+        y_test,
+        k=1,
+        batch_size=32,
+        limit=None,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        if limit is not None:
```
isaac-chung, line 39:
```suggestion
        if limit:
```

isaac-chung (commented):
Not sure if it helps with speed much, but noticed that the test/validation sets have 5k samples each. Maybe some downsampling could help.

[update]: When I ran this branch on MultiEurlex, it seems stuck here for over 5min:
```
Task: MultiEURLEXMultilabelClassification, split: validation, language: en. Running...
```

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -14,6 +19,66 @@
...
+def dot_distance(a: np.ndarray, b: np.ndarray) -> float:
+    return -np.dot(a, b)
+
+
+class kNNMultiLabelClassificationEvaluator(Evaluator):
+    def __init__(
+        self,
+        embeddings_train,
+        y_train,
+        embeddings_test,
```
isaac-chung, line 31: Looking at the structure of `kNNClassificationEvaluator`, this new evaluator seems to be very different: encode isn't called here, test cache is omitted. Any particular reason for those?

mteb/evaluation/evaluators/ClassificationEvaluator.py
x-tabdeveloping: I frankly don't think this is a good idea, staying explicit about this is better in my opinion

mteb/evaluation/evaluators/ClassificationEvaluator.py
x-tabdeveloping: Yes consult the discussion on the PR, this is all done for performance reasons

x-tabdeveloping: Btw, @KennethEnevoldsen is already assigned as reviewer and is well-versed in the decisions we made here I'm not sure two reviewers are necessary :)

isaac-chung: I see the comment now, thanks!

x-tabdeveloping:
> Not sure if it helps with speed much, but noticed that the test/validation sets have 5k samples each. Maybe some downsampling could help.
> 
> [update]: When I ran this branch on MultiEurlex, it seems stuck here for over 5min:
> 
> ```
> Task: MultiEURLEXMultilabelClassification, split: validation, language: en. Running...
> ```

We were discussing downsampling the test set, I found other issues, however, that were affecting performance way more negatively.
If the performance is still lacking after the changes, I will consider it

x-tabdeveloping:
Test fails because of request timeout, signals a more widespread issue perhaps?

x-tabdeveloping:
It runs way faster now, I think I will be ready to submit the results in a couple of hours.
I'm trying to write subsampling for the test set in the meantime.

x-tabdeveloping:
@KennethEnevoldsen performance is absolutely crap on the task with both models, especially on accuracy. Where do you think the issue may be?

KennethEnevoldsen:
@x-tabdeveloping my guess is that the KNN (euclidian distance I assume) is a poor fit for models using cosine-sim. A solution is to use a different distance metric for KNN or implement a different classifier. I believe a 2 layer MLP is a good alternative.

x-tabdeveloping:
MLP would also be great for colinearity, though it might be slower and the task already takes quite a while to run
 (also the scikit-learn implementation is not exactly the fastest)

x-tabdeveloping:
I can give it a go.

KennethEnevoldsen:
> MLP would also be great for colinearity, though it might be slower and the task already takes quite a while to run
(also the scikit-learn implementation is not exactly the fastest)

What are we talking? I don't believe classification should be limiting us here, but rather embedding time

x-tabdeveloping:
Running on UCloud again, should be able to submit results within a day.

KennethEnevoldsen (approved):
@x-tabdeveloping feel free to merge it in once it is done running!

x-tabdeveloping:
1. It runs very slow, I couldn't complete the runs, maybe we should subsample and limit to the test set instead of the validation set and the test set.
2. Performance is still crap, I have no idea what to do about that, and am at a bit of a loss as to what is happening.

x-tabdeveloping:
I believe some of the performance could be blamed on how we iteratively subsample the training set.
The way we do it now it could happen that we don't have coverage over all labels.
I will look into this real quick.

isaac-chung:
For what it's worth, maybe it might help to debug to use a small dataset.

KennethEnevoldsen:
Yea using a smaller dataset for test seems like the right approach.

> It runs very slow, I couldn't complete the runs, maybe we should subsample and limit to the test set instead of the validation set and the test set.

Hmm any idea about what part is slow? Is it simply running the trained model on the test set? (in which case reducing the test set might be an option)

> Performance is still crap, I have no idea what to do about that, and am at a bit of a loss as to what is happening.

Doing a baseline using a logistic regression on each label is probably a good idea

x-tabdeveloping:
Something's not right with these scores, I will make a deep dive

x-tabdeveloping:
I think there might be something wrong with EURLEX, the dataset itself. I tried isolating the issue and it seems that no classifier is able to perform better on the dataset than a `DummyClassifier`. I investigated whether this has something to do with how I uploaded the dataset to HuggingFace, but the original has the same issues. I will try to run the Abstask with one of the multi-clustering datasets and see what happens.

x-tabdeveloping:
I ran EURLEX in English with all-MiniLM-L6 with multiple classifiers (`MLPClassifier`, `KNN`, `DummyClassifier`).
It would seem that the task is simply incredibly hard, and that accuracy is not exactly a good metric to reflect performance, maybe we should make `lrap` the main score.
Also note that kNN outperforms MLP by quite a bit, I think this is mainly because of the very small training set and the model is quite parameter-rich.

My suggestion is that we roll back to kNN and make LRAP the main score, what do you think @KennethEnevoldsen ?

```
{
    "en": {
      "dummy": {
        "accuracy": 0.0,
        "f1": 0.0,
        "lrap": 0.17113333333332317
      },
      "knn": {
        "accuracy": 0.0396,
        "f1": 0.29540945816583636,
        "lrap": 0.4267690714285629
      },
      "mlp": {
        "accuracy": 0.0082,
        "f1": 0.08189335124049107,
        "lrap": 0.2942032142856986
      }
    },
    "evaluation_time": 270.71
  }
  ```

x-tabdeveloping:
Also including Dummy classifier scores gives us a relatively good idea of chance level in this multilabel case.

KennethEnevoldsen:
I would not include it in the task, but it might be interesting to just have a "random" model as a baseline.

- A couple of thoughts. It might be worth increasing the training set size for the MLP
  - It might be fine with just KNN, alternatively we can do KNN + MLP and take the best (similar to clf)
- It might be worth getting performance scores for subcategories (though in this case, it is 100+ right?)
- I would also like an experiment using the base e5 just to see that larger models actually perform better

x-tabdeveloping:
E5 definitely performs better on the task than paraphrase-multilingual. I'm not sure about the subcategories, might be a bit too much for some tasks. Though we could include it if need be.
In my experiments kNN uniformly performs better even with larger training set sizes. I suppose if it grows even larger it would surpass kNN, but we're already fighting performance issues with the benchmark, I think the less we have to embed the better.

x-tabdeveloping:
Also specific tasks are free to use whatever they want, like if you see an MLP more fit you can specify it in the task.
What are your thoughts on the PR right now @KennethEnevoldsen ? Should we merge or is there something that still should be addressed

KennethEnevoldsen:
I believe it is fine to merge

PR #453, opened 2024-04-19
Set the min Pydantic version to 2

bryant1410:
I was trying to use this package with Pydantic 1.10, and realized I had an import error because it doesn't have `BeforeValidator`. I saw this was available from Pydantic 2, so I set the minimum to 2.

Note I didn't check further, so chances are that a greater minimum is needed (e.g., 2.1).


KennethEnevoldsen:
Thanks @bryant1410, it is indeed correct that pydantic>2 is required. Assuming all tests pass I have set this to merge.

PR #442, opened 2024-04-19
chore: remove useless sub directory

imenelydiaker:
I may have added mteb/results submodule by mistake in PR #405. This PR is to remove it (I'm not assigning myself points, it's not a bug or anything)

PR #461, opened 2024-04-20
ci: Added doc linting rules to ensure google style docstrings.

KennethEnevoldsen:
We have specified google style docstring in the pyproject.toml, but the rule is not enforced. This PR enforces that.

Many of the changes are just autofixes by ruff.


isaac-chung (approved):
Gotta love ruff 🐶 
We saw the `tmp/results` folder before in https://github.com/embeddings-benchmark/mteb/pull/442. Is this intended?

KennethEnevoldsen:
> We saw the tmp/results folder before in https://github.com/embeddings-benchmark/mteb/pull/442. Is this intended?

No I have removed it. I have also removed it locally so it does not happen again

imenelydiaker (approved):

PR #466, opened 2024-04-20
fix: Add Zulu/Isizulu News Classification

isaac-chung:

Reason for dataset addition: Cover [Zulu / isiZulu](https://en.wikipedia.org/wiki/Zulu_language) in Classification


mteb/tasks/Classification/zul/IsiZuluNewsClassification.py
```diff
@@ -0,0 +1,39 @@
...
+        description="isiZulu News Classification Dataset",
+        reference="https://huggingface.co/datasets/dsfsi/za-isizulu-siswati-news",
+        dataset={
+            "path": "isaacchung/isizulu-news",
+            "revision": "55caf0e52693a1ea63b15a4980a73fc137fb862b",
+        },
+        type="Classification",
+        category="s2s",
+        eval_splits=["train"],
+        eval_langs=["zul-Latn"],
```
KennethEnevoldsen, line 21: What about Siswati? (the code seems to be ssw)

KennethEnevoldsen (approved):
Looks good. Only a few comments

imenelydiaker:
Just for my own culture: is Zulu and Swati the same language?

mteb/tasks/Classification/zul/IsiZuluNewsClassification.py
isaac-chung: I only copied over the Zulu one for now. Will do the Siswati one separately. When loading the original dataset I encountered an error. :/

isaac-chung:
> Just for my own culture: is Zulu and Swati the same language?

It seems like they have very close linguistic affiliation. Most comments say that they are close that it's easy to understand each other, but they are not the same. I was also reading https://en.wikipedia.org/wiki/Swazi_language

mteb/tasks/Classification/zul/IsiZuluNewsClassification.py
KennethEnevoldsen: Ahh right. Should they be in the same dataset?

PR #469, opened 2024-04-20
fix: Added encoder interfaces

KennethEnevoldsen:
Added an outline for interfaces for MTEB.

Hope this can start a discussion on how we want the standard interface to look.


mteb/encoder_interface.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
+
+from typing import Any, Protocol, runtime_checkable
+
+import numpy as np
+import torch
+
+
+@runtime_checkable
+class Encoder(Protocol):
```
KennethEnevoldsen, line 10: An example of a protocol. Only specified the interface we want to use from an encoder model.

tests/test_encoder_interfaces.py
```diff
@@ -0,0 +1,17 @@
...
+    model = SentenceTransformer("average_word_embeddings_komninos")
+    assert isinstance(model, Encoder)
+
+
+def test_wrapped_sentence_is_encoder_with_query_corpus_encode():
+    model = SentenceTransformer("average_word_embeddings_komninos")
+    model = DRESModel(model)
+
+    assert isinstance(model, Encoder)
+    assert isinstance(model, EncoderWithQueryCorpusEncode)
```
KennethEnevoldsen, line 17: Checks that we are compatible!

mteb/encoder_interface.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        ...
+
+    def encode_corpus(
+        self, corpus: list[str], prompt: str, **kwargs: Any
+    ) -> torch.Tensor | np.ndarray:
+        """Encodes the given corpus using the encoder.
+
+        Args:
+            corpus: The corpus to encode.
```
KennethEnevoldsen, line 53: This can also be a dict. Is there a good reason to allow this?

mteb/encoder_interface.py
```diff
@@ -0,0 +1,60 @@
...
+import numpy as np
+import torch
+
+
+@runtime_checkable
+class Encoder(Protocol):
+    """The interface for an encoder in MTEB."""
+
+    def encode(
+        self, sentences: list[str], prompt: str, **kwargs: Any
```
KennethEnevoldsen, line 14: prompt does not really allow the model to generate a custom prompt for their task.

A solution might be to add an optional method `.generate_prompt(task_name: str) -> prompt`?

mteb/encoder_interface.py
```diff
@@ -0,0 +1,60 @@
...
+    ) -> torch.Tensor | np.ndarray:
+        """Encodes the given queries using the encoder.
+
+        Args:
+            queries: The queries to encode.
+            prompt: The prompt to use. Useful for prompt-based models.
+            **kwargs: Additional arguments to pass to the encoder.
+
+        Returns:
+            The encoded queries.
```
KennethEnevoldsen, line 43: I believe this atm. can return `list[np.ndarray] | np.ndarray | list[torch.Tensor] | torch.Tensor`, seems like we should simplify it a bit.

mteb/encoder_interface.py
orionw: Like with #457, it would be great if the prompt/instruction could also be `List[str]` to account for when the data provides it

mteb/encoder_interface.py
orionw: I think BEIR loads it as a dict -- may not be relevant, but that's why I've seen it that way.

tests/test_encoder_interfaces.py
orionw: Looks nice!!

mteb/encoder_interface.py
orionw: I was confused by this when I started using mteb, but I missed any historical reasons for allowing multiple types.  They do mostly behave the same for iterating though

mteb/encoder_interface.py
```diff
@@ -0,0 +1,60 @@
...
+            **kwargs: Additional arguments to pass to the encoder.
+
+        Returns:
+            The encoded sentences.
+        """
+        ...
+
+
+class EncoderWithQueryCorpusEncode(Encoder, Protocol):
+    """The interface for an encoder that supports encoding a queries and a corpus."""
```
Muennighoff, line 30:
```suggestion
    """The interface for an encoder that supports encoding queries and a corpus."""
```

mteb/encoder_interface.py
Muennighoff: Yeah sometimes the corpora contain a title and a text in a dictionary i.e. sth like `{'title': '...', 'text': '...' ... }`

mteb/encoder_interface.py
isaac-chung, line 14: Would prompts be `Optional[str]`, like in SentenceTransformer.encode?
```
def encode(
        self,
        sentences: Union[str, List[str]],
        prompt_name: Optional[str] = None,
        prompt: Optional[str] = None,
        ...
```

Same for `encode_queries` and `encode_corpus`.

mteb/encoder_interface.py
KennethEnevoldsen: I think we should at least pick a format either dict or str and string would be more in line with existing implementations in MTEB, but a dict would be nice to keep it consistent with beir (I think either is reasonable).

mteb/encoder_interface.py
KennethEnevoldsen: Should we create an issue on SentenceTransformers on this?

mteb/encoder_interface.py
KennethEnevoldsen: @Muennighoff seems like you commented on the wrong thread :)

mteb/encoder_interface.py
KennethEnevoldsen: Another option to supply the prompt would also be to do something like:

```
bench = MTEB(..., prompts={"task1": "my prompt for task1", "task2": ...})
```

KennethEnevoldsen: Actually, it seems like SentenceTransformer encode has a "prompt_name" as input (thanks to @isaac-chung for pointing that out), which takes a string and looks up the corresponding prompt in a dictionary. I would just use this instead.

mteb/encoder_interface.py
KennethEnevoldsen: Actually, it seems like SentenceTransformer encode has a "prompt_name" as input (thanks to @isaac-chung for pointing that out). This can e.g. be:

```
my_model = SentenceTransformer("e5_model", prompts={“query”: “query: “, “passage”: “passage: “})

my_model.encode(sentences, prompt_name="query")
```

It seems like this can completely replace the `encode_corpus/encode_query` interface.

mteb/encoder_interface.py
KennethEnevoldsen: Ahh, this is great thanks! (especially prompt_name!)

mteb/encoder_interface.py
KennethEnevoldsen: Actually, this gives us an issue if we e.g. both want to customize the prompt according to "query" as well as "task". A solution might be to simply do:

```
my_model.encode(sentences, prompt_name="task_name,query")
```

mteb/encoder_interface.py
orionw: The example above still is only per-task (dataset) instead of per-instance.  It also looks like it forces you to only submit one `prompt_name` that selects a given string prompt to use for the entire dataset. This doesn't allow each query in one dataset to come with its own different instruction per query.

If I'm misunderstanding and the API allows you to put {query->prompt} dict for each query in the dataset and it loops over each one then great! Otherwise I think it may have to be a PR to SentenceTransformers.

Is this a correct assessment of SentenceTransformers @isaac-chung?

mteb/encoder_interface.py
orionw: Same comment as above, not sure this lets us do a unique prompt *per-query* only *per task*

mteb/encoder_interface.py
orionw: Doesn't making this a `str` not allow us to extend it to provide any custom separator tokens? I could be missing something.  I think it's fine for the default to be just concatenating them, but it should be extendable to allow others to provide custom sep tokens

orionw (approved):

isaac-chung:
> The example above still is only per-task (dataset) instead of per-instance. It also looks like it forces you to only submit one prompt_name that selects a given string prompt to use for the entire dataset. This doesn't allow each query in one dataset to come with its own different instruction per query.

This matches my understanding of the current state of what is accepted as a custom model in MTEB and what is offered from SentenceTransformers. Of course, we can call `encode` once per query if we wish to have "a unique prompt per-query", but I don't think that's the most optimized way. If we wish to pass in `prompt_names` or `prompts` as `List[str]` to achieve a unique prompt per query, changes would be needed here and at SentenceTransformers. Right now the same prompt will be applied to all of the sentences, if passed in ([ref](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L346)):
```python
if prompt is not None:
        sentences = [prompt + sentence for sentence in sentences]
```

@tomaarsen could keep me honest here.

mteb/encoder_interface.py
KennethEnevoldsen: Do people use custom separator tokens for title and text?

mteb/encoder_interface.py
KennethEnevoldsen: Yea this would just be a a singular prompt for the task  -which I believe is valid for all tasks except for the FollowIR?

KennethEnevoldsen:
I will merge this in for now. It can always be updated and does not influence the remainder of the code base.

PR #471, opened 2024-04-20
docs: Add points table 2

isaac-chung:
CI was getting this after #468 
```
fatal: pathspec 'docs/mmteb/points_table.md' did not match any files
```

PR #473, opened 2024-04-21
ci: Only create table in ci when pushing to main

isaac-chung:
Fix for  #471 and https://github.com/embeddings-benchmark/mteb/commit/22ca0e4e941b14d62d543d866fb4daecacdd2805.

## issue
1. For users without MTEB push access, the `check-points` CI step will face this error in the PR:
```
Error: Input required and not supplied: token
```
2. Even when new points files were added, the ci workflow did not register changes at the `git diff` step, so it always exits.

Fix is to 
1. split check points into 2 jobs: a) check points, b) create table
2. only run che when pushing to the main branch.
3. run `git add` after `git diff`

### DIFF
--- a/.github/workflows/mmteb.yml
+++ b/.github/workflows/mmteb.yml
@@ -6,7 +6,6 @@ name: mmteb
 on:
   push:
     branches: [main]
-  pull_request:
 
 jobs:
   check-points:
--- a/.github/workflows/mmteb.yml
+++ b/.github/workflows/mmteb.yml
@@ -6,6 +6,7 @@ name: mmteb
 on:
   push:
     branches: [main]
+  pull_request:
 
 jobs:
   check-points:
@@ -29,10 +30,13 @@ jobs:
       - name: Validate jsonl points files
         run: python docs/mmteb/validate_points.py
 
+  create-table:
+    runs-on: ubuntu-latest
+    needs: check-points
+    steps:
       - name: Create table
         run: python docs/mmteb/create_points_table.py
-
-        
+    
       - name: Push table
         # only run on push to main
         if: github.event_name == 'push' && github.ref == 'refs/heads/main'
--- a/.github/workflows/mmteb.yml
+++ b/.github/workflows/mmteb.yml
@@ -13,6 +13,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
+          cache: "pip"
+
+      - name: Validate jsonl points files
+        run: python docs/mmteb/validate_points.py
+
+  create-table:
+    runs-on: ubuntu-latest
+    needs: check-points
+    steps:
+      - uses: actions/checkout@v3
         with:
           token: ${{ secrets.RELEASE }}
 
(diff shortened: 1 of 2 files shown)


isaac-chung:
Though I'm not sure I've found why the CI run when pushing to main did not update the table. e.g. https://github.com/embeddings-benchmark/mteb/actions/runs/8770780417/job/24067725233
The push table step shows `Everything up-to-date`

KennethEnevoldsen:
Hmm but we want to run the score check on the PR. 

A solution might be to split it in two:

```
# runs all all PRs
  check-points:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-python@v4
        with:
          python-version: "3.8"
          cache: "pip"
      - name: Install dependencies
        run: |
          make install
      - name: Validate jsonl points files
        run: python docs/mmteb/validate_points.py
      - name: Create table
        run: python docs/mmteb/create_points_table.py

# seperate file that only runs on main
  create_table:
    steps:
      - uses: actions/checkout@v3
        with:
          token: ${{ secrets.RELEASE }}
      - uses: actions/setup-python@v4
        with:
          python-version: "3.8"
          cache: "pip"
      - name: Install dependencies
        run: |
          make install
          pip install tabulate # for the table creation
      - name: Push table
        # only run on push to main
        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
         run: |
           ...

**Comment by isaac-chung** [comment]
Oh yes, I got it mixed up. Let's do this:
- validate points in every PR
- only create table for pushing to main

What do you think?

**Comment by KennethEnevoldsen** [comment]
Yep, let so go for that

File: .github/workflows/mmteb.yml [hunk 1]
```diff
@@ -26,21 +46,13 @@ jobs:
...
       - name: Push table
-        # only run on push to main
-        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
         run: |
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
           git config --global user.name "github-actions[bot]"
-          git add docs/mmteb/points_table.md
           # check if the branch is up to date with the remote if not push the changes
-          git diff --exit-code || git commit -m "Update points table" && git push
+          git diff --exit-code || git add docs/mmteb/points_table.md && git commit -m "Update points table" && git push
```
KennethEnevoldsen, line 57: I believe you [need a token](https://stackoverflow.com/questions/57921401/push-to-origin-from-github-action) to be allowed to push to main (I might be wrong)

isaac-chung: Ya good call, we do. The job uses the same token as before (line#37), and it was working for at least commit 1c17394f6147acee175cdbb4dfaeb9a49ac96200.

PRs from this morning didn't trigger a commit to the table page was due to `git diff`, and likely not related to the token.

KennethEnevoldsen: Ahh right that is, because the GH changelog did not show you added it (it just as an edit of the previous). This seems like then!

KennethEnevoldsen (approved):

KennethEnevoldsen:
Hmm seems like it didn't @isaac-chung

isaac-chung:
Investigating. I thought I had a solution with `git diff-index --quiet HEAD` but I wanted to test it out on my fork first.

PR #477, opened 2024-04-21
ci: Use if-else instead for points table creation

isaac-chung:
## Issue
While the fix/feature PR would run smoothly through CI, the github-actions bot would fail on the `create-table` job, specifically on the last step. 
```
nothing to commit, working tree clean
Error: Process completed with exit code 1.
```

## Fix
Use an if statement to explicitly check the exit status of `git diff --quiet` and allow proper branching based on whether changes are detected or not.

Also replaced `make install` with pip installing only the needed packages to speed up CI.


isaac-chung:
Note that by adding the line `echo $?` at the end of the original file, I was also able to get the CI step to pass.

KennethEnevoldsen:
Thanks for taking the time on this @isaac-chung!

PR #480, opened 2024-04-22
ToxicChatClassification task added to multilingual

dokato:

Reason for dataset addition:

### DIFF
--- a/mteb/tasks/Classification/multilingual/ToxicChatClassification.py
+++ b/mteb/tasks/Classification/multilingual/ToxicChatClassification.py
@@ -1,0 +1,90 @@
+from __future__ import annotations
+
+from mteb.abstasks import AbsTaskClassification
+from mteb.abstasks.TaskMetadata import TaskMetadata
+import datasets
+
+_LANGS = [
+    "eng-Latn",
+    "pol-Latn",
+    "ron-Latn",
+    "glg-Latn",
+    "lit-Latn",
+    "bre-Latn",
+    "mlt-Latn",
+    "spa-Latn",
+    "swa-Latn",
+    "slv-Latn",
+    "msa-Latn",
+    "tgl-Latn",
+    "ita-Latn",
+    "est-Latn",
+    "mlg-Latn",
+    "xho-Latn",
+    "que-Latn",
+    "eus-Latn",
+    "nor-Latn",
+    "cym-Latn",
+    "cat-Latn",
+    "fra-Latn",
+    "nno-Latn",
+    "nld-Latn",
+    "ido-Latn",
+    "por-Latn",
+    "fin-Latn",
+    "deu-Latn",
+    "dan-Latn",
+]
+
+
+class ToxicChatClassification(AbsTaskClassification):
+    metadata = TaskMetadata(
+        name="ToxicChatClassification",
+        description="""This dataset contains toxicity annotations on 10K user
+            prompts collected from the Vicuna online demo. We utilize a human-AI
+            collaborative annotation framework to guarantee the quality of annotation
+            while maintaining a feasible annotation workload. The details of data
+            collection, pre-processing, and annotation can be found in our paper.
+            We believe that ToxicChat can be a valuable resource to drive further
+            advancements toward building a safe and healthy environment for user-AI
+            interactions.
+            Only human annotated samples are selected here.""",
+        reference="https://aclanthology.org/2023.findings-emnlp.311/",
+        dataset={
+            "path": "lmsys/toxic-chat",
+            "name": "toxicchat0124",
+            "revision": "3e0319203c7162b9c9f8015b594441f979c199bc",
+        },
+        type="Classification",
+        category="s2s",
+        eval_splits=["test"],
+        eval_langs=_LANGS,
+        main_score="accuracy",
+        date=("2023-10-26", "2024-01-31"),
+        form=["written"],
+        domains=["Constructed"],
+        task_subtypes=["Sentiment/Hate speech"],
+        license="cc-by-4.0",
+        socioeconomic_status="high",
+        annotations_creators="expert-annotated",
+        dialect=None,
+        text_creation=None,
+        bibtex_citation=None,
+        n_samples={"test": 1427},
+        avg_character_length={"test": 189.4},
+    )
+
+    def dataset_transform(self):
+        # only use human-annotated data
+        self.dataset = self.dataset['test'].filter(lambda x: x['human_annotation'])
+        keep_cols = ["user_input", "toxicity"]
+        rename_dict = dict(zip(keep_cols, ["text", "label"]))
+        remove_cols = [
+            col for col in self.dataset.column_names if col not in keep_cols
+        ]
+        self.dataset = self.dataset.rename_columns(rename_dict)
+        self.dataset = self.dataset.remove_columns(remove_cols)
+        self.dataset = self.dataset.class_encode_column("label")
+        self.dataset = self.dataset.train_test_split(
+            test_size=0.5, seed=self.seed, stratify_by_column="label"
+        )  # balanced sampling across types of hate speech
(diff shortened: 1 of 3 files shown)


mteb/tasks/Classification/multilingual/ToxicChatClassification.py
```diff
@@ -0,0 +1,87 @@
...
+        eval_langs=_LANGS,
+        main_score="accuracy",
+        date=("2023-10-26", "2024-01-31"),
+        form=["written"],
+        domains=["Constructed"],
+        task_subtypes=["Sentiment/Hate speech"],
+        license="cc-by-4.0",
+        socioeconomic_status="high",
+        annotations_creators="expert-annotated",
+        dialect=None,
```
isaac-chung, line 69:
```suggestion
        dialect=[],
```

mteb/tasks/Classification/multilingual/ToxicChatClassification.py
```diff
@@ -0,0 +1,87 @@
...
+        text_creation=None,
+        bibtex_citation=None,
```
isaac-chung, line 71: Let's add the citation provided on the dataset page:
```
@misc{lin2023toxicchat,
      title={ToxicChat: Unveiling Hidden Challenges of Toxicity Detection in Real-World User-AI Conversation}, 
      author={Zi Lin and Zihan Wang and Yongqi Tong and Yangkun Wang and Yuxin Guo and Yujia Wang and Jingbo Shang},
      year={2023},
      eprint={2310.17389},
      archivePrefix={arXiv},
      primaryClass={cs.CL}
}

```

mteb/tasks/Classification/multilingual/ToxicChatClassification.py
```diff
@@ -0,0 +1,87 @@
...
+        main_score="accuracy",
+        date=("2023-10-26", "2024-01-31"),
+        form=["written"],
+        domains=["Constructed"],
+        task_subtypes=["Sentiment/Hate speech"],
+        license="cc-by-4.0",
+        socioeconomic_status="high",
+        annotations_creators="expert-annotated",
+        dialect=None,
+        text_creation=None,
```
isaac-chung, line 70:
```suggestion
        text_creation="found",
```

mteb/tasks/Classification/multilingual/ToxicChatClassification.py
```diff
@@ -0,0 +1,87 @@
...
+        # only use human-annotated data
+        self.dataset = self.dataset["test"].filter(lambda x: x["human_annotation"])
+        keep_cols = ["user_input", "toxicity"]
+        rename_dict = dict(zip(keep_cols, ["text", "label"]))
+        remove_cols = [col for col in self.dataset.column_names if col not in keep_cols]
+        self.dataset = self.dataset.rename_columns(rename_dict)
+        self.dataset = self.dataset.remove_columns(remove_cols)
+        self.dataset = self.dataset.class_encode_column("label")
+        self.dataset = self.dataset.train_test_split(
+            test_size=0.5, seed=self.seed, stratify_by_column="label"
```
isaac-chung, line 86: Great that it's stratified! Also, since test_size is 0.5, would `n_samples` be half of ~5k (length of test set) instead of 1427?

isaac-chung (commented):
Thank you for your contribution! Just a few comments below.

Feel free to add points for yourself (6 for each new language in classification - you can do a quick search in the codebase on the lang code, and 2 for each of the rest) and 2 points for me for the review.

mteb/tasks/Classification/multilingual/ToxicChatClassification.py
dokato: Correct, except that I filtered only human annotated rows above.

isaac-chung:
A task is like Classification, Retrieval etc. So this PR is extending language support for an existing task. 

For each of these languages, add 2 + 4 = 6 points as they are new languages in Classification. So that's 9x6=54
```
[
    "glg-Latn",
    "lit-Latn",
    "bre-Latn",
    "mlg-Latn",
    "que-Latn",
    "eus-Latn",
    "nor-Latn",
    "cat-Latn",
    "ido-Latn",
]
```

For each of the rest, add 2 points each. So that's 20x2 = 40.

Which makes the total = 94 pts for you, 2 pts for me. Great work!

mteb/tasks/Classification/multilingual/ToxicChatClassification.py
isaac-chung: Aha good call out. I missed the comment above. Thanks!

isaac-chung (approved):

dokato:
Amazing, thank you @isaac-chung 🎉

isaac-chung:
@imenelydiaker I should have caught this at review time. @dokato before I open a PR to fix points and move this under English, did you mean to add a different dataset that covers add these languages?

dokato:
Sorry, as I'm new to the project I wasn't sure what is the definition of "multilingual" here. Since there were multiple languages in the text, I detected them all and listed as appearing in the set.

> did you mean to add a different dataset that covers add these languages?

Not sure what do you mean by that either? I don't have anything related to toxic chat classification for sure, but I started working on `Davlan/sib200` that clearly has many languages. I'll submit a PR for this soon.

dokato:
BTW happy to help amend it if you instruct me what are best steps?

imenelydiaker:
> Sorry, as I'm new to the project I wasn't sure what is the definition of "multilingual" here. Since there were multiple languages in the text, I detected them all and listed as appearing in the set.
> 

How did you detect them? The problem is that the paper says that 99% of these conversations are in English and the HF dataset card says it's an English dataset also. For multilingual datasets, we require one subset per language to be able to evaluate each language on its own. 
If this dataset contains mixed texts, which means a text with multiple languages in it, we'll take the more important language and the rest of it will be treated as OOV tokens by models I suppose. So in this case, it would be English.

KennethEnevoldsen:
> we require one subset per language to be able to evaluate each language on its own.

We do have datasets where that is not the case (e.g. language classification).

imenelydiaker:
> > we require one subset per language to be able to evaluate each language on its own.
> 
> We do have datasets where that is not the case (e.g. language classification).
> 
I think this [dataset](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Classification/multilingual/IndicLangClassification.py) considers each language as a label, right? 
Here, there's no indicator on the language of the samples, and the paper introducing it doesn't say that there is anything different from English.
The only problem I see when mixing languages, is that we deviate from the initial goal of evaluating a model for a task on a specific language. 
Although, having mixed-languages samples (one sample with multiple languages in the same text) is cool and multilingual models should be able to handle them, but I'm not sure it's right to evaluate a language that is mixed with another one, it becomes familiar and informal languages evaluation, no?

dokato:
> How did you detect them?

I went through all texts and and used `langid.classify` to detect the language code.

dokato:
So step to fix it following @imenelydiaker comments:

1) Move `multilingual/ToxicChatClassification.py` to `eng/ToxicChatClassification.py`
2) Remove _LANG and `eval_langs` becomes `eng-Latn`

Sth else?Should we do sth about other languages or as you said model will treat them as OOS and we can live with that?

KennethEnevoldsen:
@imenelydiaker feel free to create an issue to discuss this further (I believe it is a relevant discussion, but not too related to this)

@dokato these seems like good changes I believe I would change the eval languages to eng-Latn but with a comment in the code that it also contains other languages (potentially in the description as well).

isaac-chung:
Re: Fix, do compare to what's already been included in https://github.com/embeddings-benchmark/mteb/pull/527 by @imenelydiaker .

imenelydiaker:
> So step to fix it following @imenelydiaker comments:
> 
> 1) Move `multilingual/ToxicChatClassification.py` to `eng/ToxicChatClassification.py`
> 2) Remove _LANG and `eval_langs` becomes `eng-Latn`
> 
> Sth else?Should we do sth about other languages or as you said model will treat them as OOS and we can live with that?

I've already addressed this in PR #527

PR #481, opened 2024-04-22
Fast Clustering

KennethEnevoldsen:
This creates an alternative fast clustering for MTEB. It is notably faster. It gains this performance by downsampling and only performing the embed step once. To then regain the robustness of creating clusters multiple times it samples from the embeddings to create clusters.


mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -0,0 +1,189 @@
...
+            cluster_size=self.max_documents_per_cluster,
+            kmean_batch_size=self.k_mean_batch_size,
+            rng_state=random.Random(self.seed),
+        )
+
+        return {"main_score": self.calculate_main_score(scores), "scores": scores}
+
+
+@dataclass
+class MTEBScores:
```
KennethEnevoldsen, line 143: @MartinBernstorff I would love your opinion on this result object. Notably, the format of the scores seems too complex. I would love to discuss it with you.

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -0,0 +1,189 @@
...
+    labels: list[str],
+    n_clusters: int,
+    cluster_size: int,
+    kmean_batch_size: int,
+    rng_state: random.Random = random.Random(),
+) -> ClusteringScores:
+    """Bootstrapped evaluation of clustering performance using V-measure.
+
+    The bootstrapping is done by sampling N samples from the corpus and clustering them. It is done without replacement to get a diverse set of
+    samples.
```
KennethEnevoldsen, line 40: With or without replacement. Don't have any strong intuition about what is the best here.

mteb/tasks/Clustering/eng/ArxivClusteringP2PFast.py
```diff
@@ -0,0 +1,61 @@
...
+        avg_character_length={"test": 1009.98},
+    )
+
+    def dataset_transform(self):
+        sent = self.dataset["test"]["sentences"][0]
+        lab = self.dataset["test"]["labels"][0]
+
+        self.dataset["test"] = datasets.Dataset.from_dict(
+            {"sentences": sent, "labels": lab}
+        )
```
KennethEnevoldsen, line 43: now it is a dataset with two columns sentences and labels list[str], before it was a dataset of two column (list[list[str]]), where the first list is a cluster and the second list is the labels/sentences of that cluster.

mteb/tasks/Clustering/swe/swedn_clustering.py
KennethEnevoldsen, line 1: There was a bug in swedn clustering and it is also too small to really test on so let us just ignore it.

mteb/tasks/Clustering/eng/ArxivClusteringP2PFast.py
```diff
@@ -0,0 +1,61 @@
...
+if __name__ == "__main__":
+    from sentence_transformers import SentenceTransformer
+
+    import mteb
+    from mteb.tasks.Clustering.eng.ArxivClusteringP2P import ArxivClusteringP2P
+
+    old_task = ArxivClusteringP2P()
+    old_task.load_data()
+    # downsample to 1 cluster
+    old_task.dataset["test"] = datasets.Dataset.from_dict(old_task.dataset["test"][:1])
```
KennethEnevoldsen, line 55: instead of running it on 10 clusters I run it on 1. Still 250 000 documents.

tmp/results/ArxivClusteringP2PFast1.json
KennethEnevoldsen, line 1: using slightly different hyperparameters:

```
    max_documents_to_embed = 10000
     max_documents_per_cluster = 1000
     n_clusters = 10
     k_mean_batch_size = 512
```

Generally seems like smaller clusters make the task easier.

mteb/tasks/Clustering/eng/ArxivClusteringP2PFast.py
KennethEnevoldsen: One frustrating part (for backward compatiblity) is that the clusters are not from the same distribution.

E.g. for Swedn we had 2 clusters articles and 2 clusters of summaries (and a bug, but not relevant here). Which naturally isn't the same as sampling from articles and summaries and generating clusters. The same is the case for ArxivClusteringP2P (but I am unsure if this a general case or not - my gut would say that about ~30% has exceptions like these).

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -0,0 +1,189 @@
...
+        scores.append({"v_measure": v_measure})
+
+    return scores
+
+
+class AbsTaskClusteringFast(AbsTask):
+    """Abstract class for Clustering tasks.
+
+    This class embeds the corpus sentences then samples N samples from the corpus and clusters them.
+    he similarity then is calculated using the V-measure metric, which is invariant to the permutation of the labels.
```
isaac-chung, line 71:
```suggestion
    The similarity then is calculated using the V-measure metric, which is invariant to the permutation of the labels.
```

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -0,0 +1,189 @@
...
+    This approach is then repeated K times.
+
+    self.load_data() must generate a huggingface dataset with a split matching self.metadata_dict["eval_splits"], and assign it to self.dataset.
+    It must contain the following columns:
+        sentences: list of str
+        labels: list of str
+    """
```
isaac-chung, line 78:
```suggestion
        sentences: list[str]
        labels: list[str]
    """
```

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -0,0 +1,189 @@
...
+    for _ in range(n_clusters):
+        # sample N samples from the corpus without replacement
+        cluster_indices = rng_state.sample(range(n_embeddings), k=cluster_size)
+
+        _embeddings = embeddings[cluster_indices]
+        _labels = labels_arr[cluster_indices]
+        clustering_model.fit(_embeddings)
+        cluster_assignment = clustering_model.labels_
+        v_measure = v_measure_score(_labels, cluster_assignment)
+        scores.append({"v_measure": v_measure})
```
isaac-chung, line 62: Could we append just `v_measure` instead of the dictionary, like in the normal case? Was there a special purpose to the dict?

mteb/abstasks/AbsTaskClusteringFast.py
isaac-chung: As I understand, bootstrapping would mean sampling with replacement. Since we have downsampled at the encode step, bootstrapping should help "raise" the number of samples same as the original size. I'm not sure if we can achieve that if we sampled without (like in train_test_split). We could use `choices` instead:
```
random.choices(seq, k)
```

tmp/results/SwednClustering.json
```diff
@@ -0,0 +1,17 @@
+{
+  "all": {
+    "evaluation_time": 9.65,
```
isaac-chung, line 3: Looking at the file names, does this mean that normal was 4x faster than the "fast" class? 🤔

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -0,0 +1,189 @@
...
+
+        ds: HFDataset = self.dataset  # type: ignore
+        scores = self._evaluate_monolingual(model, ds, split, **kwargs)
+        return scores
+
+    def _evaluate_monolingual(
+        self, model, dataset: HFDataset, split: str = "test", **kwargs: Any
+    ) -> dict[str, float | ClusteringScores]:
+        sentences = dataset[split]["sentences"][: self.max_documents_to_embed]
+        labels = dataset[split]["labels"][: self.max_documents_to_embed]
```
isaac-chung, line 124: Right now it always takes the first `self.max_documents_to_embed` samples to encode. Should we shuffle those first?

mteb/abstasks/AbsTaskClusteringFast.py
isaac-chung: 1. Seems like `main_score` is a subset of `scores`. Maybe we could drop it?
2. Are we missing `dataset_revision`?

isaac-chung (commented):
Noticed that this PR includes some efforts for a consolidated multilingual MTEBScores class. Could that be a separate PR?

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: Yep, the problem I found was that it inflated the score slightly (due to the same embeddings would naturally end up in the same cluster). I am not sure it is that big of a problem though

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: Yeah, the main score could be dropped. Will add in dataset revision

mteb/tasks/Clustering/eng/ArxivClusteringP2PFast.py
KennethEnevoldsen: > This change is to make it simpler?

Yep as the user-defined clusters (sets of documents) are no longer needed, which allows us to sample clusters with some overlap (allowing for a notably larger sample) 

> In that Swedn case, shouldn't it be two separate datasets then, maybe SwednP2P & SwednS2S Clustering?

Yea that might indeed be a better solution

> I think for most current datasets like arxiv etc all clusters are from the same dist, no?

I think you are right, though arxivClusteringP2P seems to be from multiple dists (as I understand the paper at least)

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: yeah it could just be a list instead

KennethEnevoldsen:
> Noticed that this PR includes some efforts for a consolidated multilingual MTEBScores class. Could that be a separate PR?

Yeah, just had to consider it when designing the task, but I think you are right that it is indeed a separate PR.

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: Good question! 

I see two potential solutions:

1) sample randomly as you suggest
2) take the first n., this will allow for streaming datasets

Either solution if fine with me (It should def. not be the way it is now)

tmp/results/SwednClustering.json
KennethEnevoldsen: Yep in the case of Swedn it is about the same (it is the same amount of documents being embedded), the extra cost comes from the bootstrapping and fitting clusters.

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: I have removed MTEBScore - we can take it in another PR

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: I went with the classic bootstrapping approach you suggested

mteb/tasks/Clustering/eng/ArxivClusteringP2PFast.py
KennethEnevoldsen: I created a new arxiv P2P dataset, let me know if there is issue with it.

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: It is now a list

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: I went with 1)

KennethEnevoldsen:
Thanks for the review @isaac-chung and @Muennighoff.

I have now:
- removed the MTEBScore (for another PR)
- Added the Fast clustering task, but keeping the old implementation 
- Added two new datasets using the old implementation, adding a "superseeded_by" attribute to the old datasets, which will raise a warning.
- I have split swednClustering into P2P and S2S

I believe this is close to ready to merge, let me know what you think

isaac-chung (approved):
Most of my comments were on the MTEBScore class. I did not review the cluster datasets. The rest looks good to me.

KennethEnevoldsen:
I will enable auto-merge on this PR

Muennighoff:
Note that new models are now using this, e.g. in https://huggingface.co/spaces/mteb/leaderboard?task=clustering&language=french the voyage-law-2 model was run with fast clustering while other models were not. Is this okay?

PR #483, opened 2024-04-22
Add Telugu News Classification

SaitejaUtpala:

Reason for dataset addition: **This is first telugu classification in the mteb**

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -216,6 +216,7 @@ The following tables gives you an overview of the tasks in MTEB.
 | [SICKFr](https://huggingface.co/datasets/Lajavaness/SICK-fr) | ['fra'] | STS | s2s |  |  |  |
 | [SNLClustering](https://huggingface.co/datasets/navjordj/SNL_summarization) (Navjord et al., 2023) | ['nob'] | Clustering | p2p | [Encyclopaedic, Non-fiction] | {'test': 2048} | {'test': 1101.3} |
 | [SNLRetrieval](https://huggingface.co/datasets/navjordj/SNL_summarization) (Navjord et al., 2023) | ['nob'] | Retrieval | p2p | [Encyclopaedic, Non-fiction] | {'test': 2048} | {'test': 1101.3} |
+| [SRNCorpusBitextMining](https://arxiv.org/abs/2212.06383) (Zwennicker et al., 2022) | ['nld', 'srn'] | BitextMining | s2s | [Social, Web] | {'test': 256} | {'test': 55.0} |
 | [STS12](https://www.aclweb.org/anthology/S12-1051.pdf) | ['eng'] | STS | s2s |  |  |  |
 | [STS13](https://www.aclweb.org/anthology/S13-1004/) | ['eng'] | STS | s2s |  |  |  |
 | [STS14](https://www.aclweb.org/anthology/S14-1002) | ['eng'] | STS | s2s |  |  |  |


docs/mmteb/points/483.jsonl
```diff
@@ -0,0 +1,2 @@
+{"GitHub": "SaitejaUtpala", "New dataset": 6}
```
isaac-chung, line 1: Telugu has been covered in classification in `MassiveIntentClassification`.
```suggestion
{"GitHub": "SaitejaUtpala", "New dataset": 2}
```

mteb/tasks/Classification/tel/TeluguAndhraJyotiNewsClassification.py
```diff
@@ -0,0 +1,77 @@
...
+        bibtex_citation=None,
+        n_samples={"train": 17312, "test": TEST_SAMPLES},
+        avg_character_length={"train": 1435.53, "test": 1428.28},
+    )
+
+    def dataset_transform(self):
+        seed = 42
+        random.seed(seed)
+        self.dataset = self.dataset.rename_column("body", "text")
+        self.dataset = self.dataset.rename_column("topic", "label")
```
isaac-chung, line 44:
```suggestion
        self.dataset = self.dataset.rename_column("body", "text").rename_column("topic", "label")
```

mteb/tasks/Classification/tel/TeluguAndhraJyotiNewsClassification.py
```diff
@@ -0,0 +1,77 @@
...
+
+        for split in ["test"]:
+            ds = self.dataset[split]
+            # Determine number of classes and samples per class
+            class_count = Counter([sample["label"] for sample in ds])
+            num_classes = len(class_count)
+            total_samples = min(TEST_SAMPLES, len(ds))
+            samples_per_class = total_samples // num_classes
+
+            # Try to maintain class balance
```
isaac-chung, line 54: this is great! I wonder if something like could help:
```
self.dataset = self.dataset["test"].train_test_split(
            test_size=TEST_SAMPLES, seed=42, stratify_by_column="label"
   ) 
```

isaac-chung (changes requested):
Thanks for your contributions! Just a few comments.

mteb/tasks/Classification/tel/TeluguAndhraJyotiNewsClassification.py
SaitejaUtpala: @isaac-chung  I don't think this is stratified sampling !. This is completely different.  label with least examples is maxed out, so that class imbalance is minimized

isaac-chung: I see, thanks!. @imenelydiaker and I are actually going through past datasets and we saw this as well in `VieStudentFeedbackClassification`. I wonder if we'd like to keep our stratified sampling strategy as default for all (to me this is sufficient), or adopt this strategy here.

imenelydiaker: The stratification process is just here because of the subsampling, it allows us to keep the same samples a'd labels proportions as in the original dataset. Note that without the subsampling we would have run the evaluation on the whole dataset even if it's imbalanced.

So I think the question we should be asking is: do we want to have imbalanced classes during the evaluation? Because we're using KNN to evaluate the classification task, after that we compute an accuracy whatever the class balance is.

isaac-chung: Make sense. Having imbalanced classes are realistic and IMO inevitable. The main score for those datasets could be f1 instead of accuracy. Wdyt?

imenelydiaker: Yes I think having f1 as a main score is a good idea! Should we update previous tasks that used accuracy?

isaac-chung: Yes if those datasets are noticeably imbalanced.

Issue #508, opened 2024-04-23
Add helper function to unify how we subsample data


isaac-chung:
Piggy-backing off @imenelydiaker 's [comment](https://github.com/embeddings-benchmark/mteb/pull/424#issuecomment-2063881255):

When the dataset split is beyond 2048, subsampling is often used in this manner without stratification by label:
```python
self.dataset["test"] = (
    self.dataset["test"].shuffle(seed=self.seed).select(range(N_SAMPLES))
)
```
I've primarily seen this in classification datasets, but I imagine this would apply to other tasks as well.

A helper function can be introduced to help ease the contribution efforts when dealing with larger datasets, and to help unify how we subsample data. Example usage:
```python
self.dataset['test'] = stratified_subsampling(self.dataset['test'], N_SAMPLES)
```

imenelydiaker:
Totally agree, we can use `train_test_split` from `datasets` ? Or we can create a the core of the function. There may exist better options here.

I'd add a variable to choose on what to stratify:
```py
self.stratified_subsampling(self.dataset, label, n_samples)
```
Should the method be defined in `AbsTask`?

imenelydiaker:
We should totally point new contributors towards this yes!

Although I think it's posible to replace old subsampling functions, there are not a lot of them for the moment. We also should be able to run evals again on `multilingual-e5-small` and `paraphrase-multilingual-MiniLM-L12-v2`.

If you want we can split the work, you can open a PR with the new subsampling function proposition, and I'll update tasks using the old version and run evaluations again, wdyt?

isaac-chung:
Ya let's do that 💪

KennethEnevoldsen:
Wonderful suggestion @isaac-chung and @imenelydiaker!

PR #513, opened 2024-04-23
Unify subsampling with stratify

isaac-chung:
Addresses https://github.com/embeddings-benchmark/mteb/issues/508

Also added a rerun as a comparison.

### DIFF
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -35,7 +35,7 @@ class AbsTask(ABC):
         """
         pass
 
-    def stratified_subsampling(self, splits: Union[str, list[str]], label: str = "label", n_samples: int = max_n_samples):
+    def stratified_subsampling(self, splits: Union[str, list[str]] = ["test"], label: str = "label", n_samples: int = max_n_samples):
         """Subsamples the dataset with stratification by the supplied label.
         The following kwargs must be provided
         for stratified_subsampling to run:
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -53,7 +53,7 @@ class AbsTask(ABC):
         for split in splits:
             self.dataset[split] = self.dataset[split].train_test_split(
                 test_size=n_samples, seed=self.seed, stratify_by_column=label
-            )
+            )['test'] ## only take the specified test split.
 
     def load_data(self, **kwargs):
         """Load dataset from HuggingFace hub"""
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import random
 from abc import ABC, abstractmethod
-
 from typing import Union
 
 import datasets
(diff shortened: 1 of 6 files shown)


imenelydiaker (approved):
This looks great, thank you !

imenelydiaker (commented):
Just one thing, should we allow subsampling without stratification? For datasets that don't have a label column for example?

isaac-chung:
> Just one thing, should we allow subsampling without stratification? For datasets that don't have a label column for example?

I'd say this base case is covered by
```
self.dataset["test"] = (
    self.dataset["test"].shuffle(seed=self.seed).select(range(TEST_SAMPLES))
)
```

wdyt?

imenelydiaker:
> > Just one thing, should we allow subsampling without stratification? For datasets that don't have a label column for example?
> 
> I'd say this base case is covered by
> 
> ```
> self.dataset["test"] = (
>     self.dataset["test"].shuffle(seed=self.seed).select(range(TEST_SAMPLES))
> )
> ```
> 
> wdyt?

Yep true!

KennethEnevoldsen:
@isaac-chung when I doubt just use fix it bumps the version (which we basically want to do every time we change the codebase)

isaac-chung:
Will do next time. Thanks!

mteb/abstasks/AbsTask.py
```diff
@@ -12,6 +13,7 @@
 
 class AbsTask(ABC):
     metadata: TaskMetadata
+    max_n_samples = 2048
```
KennethEnevoldsen, line 16: @isaac-chung I believe this line is problematic as it can be understood as the max number of samples for the dataset (which is not actually enforced). I would instead move it to `stratified_subsampling` as an argument with a default.

mteb/abstasks/AbsTask.py
```diff
@@ -32,6 +34,31 @@ def dataset_transform(self):
         """
         pass
 
+    def stratified_subsampling(
+        self,
+        splits: Union[str, list[str]] = ["test"],
```
KennethEnevoldsen, line 39: I would just use list[str] (following the mantra: `There should be one-- and preferably only one --obvious way to do it.`)

mteb/abstasks/AbsTask.py
```diff
@@ -32,6 +34,31 @@ def dataset_transform(self):
...
+        label: str = "label",
+        n_samples: int = max_n_samples,
+    ):
+        """Subsamples the dataset with stratification by the supplied label.
+        The following kwargs must be provided
+        for stratified_subsampling to run:
+        - splits: Union[str, list[str]], the splits of the dataset.
+        - label: str, the label with which the stratified sampling is based on.
+        - n_samples: Optional[int], number of samples to subsample. Default is max_n_samples.
+        """
```
KennethEnevoldsen, line 49: reformat to google style docstrings

mteb/abstasks/AbsTask.py
```diff
@@ -32,6 +34,31 @@ def dataset_transform(self):
...
+            splits = [splits]
+
+        ## Can only do this if the label column is of ClassLabel.
+        if not isinstance(self.dataset[splits[0]].features[label], datasets.ClassLabel):
+            self.dataset = self.dataset.class_encode_column(label)
+
+        for split in splits:
+            self.dataset[split] = self.dataset[split].train_test_split(
+                test_size=n_samples, seed=self.seed, stratify_by_column=label
+            )["test"]  ## only take the specified test split.
```
KennethEnevoldsen, line 60: I would instead create the dataset and return it and convert to method to a @staticmethod. This avoids state which can often lead to bugs.

KennethEnevoldsen (commented):
Good thing that you guys merged this in (feel free to do that in the future as well), I have adressed a few issues -  @isaac-chung will you do these in a new PR?

isaac-chung:
PR coming right up.

Issue #520, opened 2024-04-23
Subsampling function removes train split from dataset


imenelydiaker:
Following my second comment in #519, the subsampling function rebuild a new `DatasetDict` that removes the `train` split that is loaded by the `load_dataset` function. This train split is used by the Classification evaluator and thus should not be removed.

isaac-chung:
Would it make more sense to have `stratified_subsampling` handle one split at a time, so that we can have this?
```
self.dataset["test"] = self.stratified_subsampling(self.dataset["test"])
```

That would change the function signature quite a bit:
```
def stratified_subsampling(
       dataset: datasets.Dataset,
        seed: int,
        label: str = "label",
        n_samples: int = 2048,
    ) -> datasets.Dataset:
```

imenelydiaker:
I opened a [PR](https://github.com/embeddings-benchmark/mteb/pull/521) with a proposed change. But I think that what you are proposing is better.

isaac-chung:
I liked yours better actually.

isaac-chung:
Thank you for pushing a fix so quickly!

PR #521, opened 2024-04-23
fix: update current dict and do not remove other splits

imenelydiaker:
Proposed fix for issue #520, the proposition is to update the splits of the current dataset dict instead of creating a new one. This should not update splits that are not mentioned in `splits` argument.


isaac-chung (approved):
👍

PR #522, opened 2024-04-23
Add get_tasks

KennethEnevoldsen:
Added a utility function for fetching tasks (the idea is that in the future it will replace MTEBs own filtering so that all the filtering happens in one space).


isaac-chung (approved):
Nice! Everything is clear.

PR #525, opened 2024-04-23
added create task table

KennethEnevoldsen:
Added CI for creating task table


.github/workflows/docs.yml
```diff
@@ -0,0 +1,42 @@
+# GitHub action for the mmteb contributions
+# this way the CI related to the mmteb contributions will be separated from the linting CI
```
isaac-chung, line 2:
```suggestion
# GitHub action for the task table generation.
```

docs/create_tasks_table.py
```diff
@@ -0,0 +1,86 @@
...
+    """Create (Authors, Year) from bibtex entry (author = {Authors}, year = {Year})"""
+    if bibtex is None:
+        return ""
+    # get authors from bibtex (author = {Authors} or author={Authors})
+    authors = re.search(r"author\s*=\s*{([^}]*)}", bibtex)
+    if authors is None:
+        return ""
+    authors = authors.group(1)
+    authors = [a.split(", ") for a in authors.split(" and ")]
+    # authors = [f"{a[0]} et al." if len(a) > 1 else a[0] for a in authors]
```
isaac-chung, line 19:
```suggestion

```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,42 @@
...
+  create-table:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.RELEASE }}
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
```
isaac-chung, line 20: Other workflow uses 3.10. Should we align at 3.8 or bump others to 3.10 as well?

isaac-chung (commented):
Nice! A few small comments.


Should also merge main as there are some overlaps with https://github.com/embeddings-benchmark/mteb/pull/522.

.github/workflows/docs.yml
KennethEnevoldsen: I don't think it matters. I would just use the most stable version for everything except for tests (3.11 would probably also be fine)

PR #409, opened 2024-04-17
Add NTREX dataset

davidstap:
## Checklist for adding MMTEB dataset

Reason for dataset addition:

- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] I have considered the size of the dataset and reduced it if it is too big (2048 examples is typically large enough for most tasks)
- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 
- [x] I have added points for my submission to the [POINTS.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) file.

PR #543, opened 2024-04-24
Add new languages to NTREX

davidstap:
#409 uses a Huggingface dataset with only some NTREX languages. This PR makes use of a new Huggingface dataset I created, which includes all NTREX languages.

Refer to `NTREXBitextMining.py` lines 12-133 for a list of languages. Currently, I've included languages from #409 and those not yet available for the BitextMining task, leaving others commented out. I'm unsure which languages to include. What are your thoughts, @KennethEnevoldsen? I'll adjust based on your feedback. Possible options:
* Maintain current selection and add new languages to enhance language coverage.
* Include all languages and directions, but this creates 14762 directions and is infeasible (?).
* Include all languages but limit directions, possibly only including those involving English or another few "bridge languages" as either source or target, similar to `TatoebaBitextMining`.


KennethEnevoldsen:
> Include all languages and directions, but this creates 14762 directions and is infeasible (?).

Def. sounds like it. I would probably imagine that some directions are implausible. Danish to Hindi is probably unlikely to occur in practice. Downsampling it in a meaningful way (as you suggest would be nice).

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
```diff
@@ -55,12 +161,12 @@ class NTREXBitextMining(AbsTaskBitextMining, CrosslingualTask):
     metadata = TaskMetadata(
         name="NTREXBitextMining",
         dataset={
-            "path": "xianf/NTREX",
-            "revision": "6047adbf00ceeab9b543bd0bd779b99866deddc0",
+            "path": "davidstap/NTREX",
+            "revision": "fd20d54141b6da952d5c68a2989472892489da0f",
             "trust_remote_code": True,
         },
         description="NTREX is a News Test References for MT Evaluation from English into a total of 128 target languages.",
```
KennethEnevoldsen, line 168: Should be changed in imagine

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
```diff
@@ -75,8 +181,8 @@ class NTREXBitextMining(AbsTaskBitextMining, CrosslingualTask):
         annotations_creators="expert-annotated",
```
KennethEnevoldsen, line 181: Is it really expert annotated?

KennethEnevoldsen (commented):
A few issues, which isn't related to this PR but the dataset

> date=("2022-11-01", "2022-11-01"),

This date seems wrong

KennethEnevoldsen:
hmm currently the large grid would lead to extensive amount of encodes (even though there may be duplicates). It am not sure that is a good way to do it, but that would be a be a very reasonable way to re-use embeddings.

(some sort of embed cache)

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
davidstap: I think so, from the [NTREX paper](https://aclanthology.org/2022.sumeval-1.4/):

> To produce this data set we sent out the original English WMT19 (Barrault et al., 2019) test set (‘newstest2019‘) to professional human translators.

davidstap:
> > Include all languages and directions, but this creates 14762 directions and is infeasible (?).
> 
> Def. sounds like it. I would probably imagine that some directions are implausible. Danish to Hindi is probably unlikely to occur in practice. Downsampling it in a meaningful way (as you suggest would be nice).

Okay I'll create a reasonable downsample at a later stage.

davidstap:
> hmm currently the large grid would lead to extensive amount of encodes (even though there may be duplicates). It am not sure that is a good way to do it, but that would be a be a very reasonable way to re-use embeddings.
> 
> (some sort of embed cache)

Yes there will be a lot of duplicates: the data is fully multi-parallel, so it seems to me that the sentence encodings for some language can be re-used for all combinations with other languages.

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
davidstap: I'm not sure what you mean?

davidstap:
> A few issues, which isn't related to this PR but the dataset
> 
> > date=("2022-11-01", "2022-11-01"),
> 
> This date seems wrong

This is the publication date of the NTREX paper. I thought that was most reasonable given for date, given _The date when the data was collected._ from the docs. Or should it be today?

davidstap:
I changed from draft to ready for review, can you take (another) look @KennethEnevoldsen @imenelydiaker ?

imenelydiaker (commented):
Some scores are surprisingly high, else the implementation looks good to me!

KennethEnevoldsen:
> A few issues, which isn't related to this PR but the dataset
> 
> date=("2022-11-01", "2022-11-01"),
> This date seems wrong

Hi @davidstap sorry for the confusion here. The date is when the text was written (e.g. if found on Wikipedia it could be written between the start of Wikipedia up until the paper release).

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
KennethEnevoldsen: Sorry again for the short comments, seems like I might have been doing a lot of review that day.

The description write from English to 128 languages, which seems to imply only one direction -- English to 128 languages. While I believe this is the description of the NTREX data, the description should be of the given current task (including any modification we might have made).

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
KennethEnevoldsen: Oh wow. 

Then we need to change the text_creation to human translated (and localized if that is the case).

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
```diff
@@ -75,8 +267,8 @@ class NTREXBitextMining(AbsTaskBitextMining, CrosslingualTask):
...
         dialect=[],
         text_creation="created",
-        n_samples={"train": _N},
-        avg_character_length={"train": 120},
+        n_samples={"test": _N},
```
KennethEnevoldsen, line 270: I believe this is _N x number of language pairs? (for the whole dataset)

KennethEnevoldsen (commented):
It seems like a few of the languages pairs obtain close to perfect performance (as @imenelydiaker noted). I am not sure exactly what to do about those, but they leave little space for improvement. We might consider running a "stupid" model e.g. the all lm mini L6 (small english model) to see how that performs and then at least remove all languages which score 99 or above for all three models (we could also do that for the current 2 models)

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
davidstap: I see, will update the description!

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
davidstap: We have the following options:
```
    "found",
    "created",
    "machine-translated",
    "human-translated and localized",
    "machine-translated and verified",
    "machine-translated and localized",

```
I think `human-translated and localized` is the best match, even though I cannot find any information about localization in the NTREX paper.

mteb/tasks/BitextMining/multilingual/NTREXBitextMining.py
davidstap: You're right, will change it to `_N*len(_EVAL_LANGS)` (`EVAL_LANGS` has 1916 language directions)

davidstap:
> It seems like a few of the languages pairs obtain close to perfect performance (as @imenelydiaker noted). I am not sure exactly what to do about those, but they leave little space for improvement. We might consider running a "stupid" model e.g. the all lm mini L6 (small english model) to see how that performs and then at least remove all languages which score 99 or above for all three models (we could also do that for the current 2 models)

I ran the "stupid" model (see `results/sentence-transformers__all-MiniLM-L6-v2/NTREXBitextMining.json`) and for this I did not observe scores that were too high. I think the high scores are due to a lot of named entities and only 256 examples, I'll try a large subset to see what happens to the scores.

Apart from this I think it's ready to merge, I fixed the other suggestions @KennethEnevoldsen

KennethEnevoldsen:
> I think the high scores are due to a lot of named entities and only 256 examples, I'll try a large subset to see what happens to the scores.

@davidstap a lot of NEs? Sounds this be sentences? Might be worth filtering out documents where either pair is less than 4 words)

davidstap:
> > I think the high scores are due to a lot of named entities and only 256 examples, I'll try a large subset to see what happens to the scores.
> 
> @davidstap a lot of NEs? Sounds this be sentences? Might be worth filtering out documents where either pair is less than 4 words)

Sorry I meant named entities in sentences (this is typical for the news domain)

davidstap:
The results for subsampling 256 datapoints vs using the full dataset are summarised in the table below:

|   |  >0.95 | >0.97  | >0.99  |
|---|---|---|---|
|  `256` |297/1917|134/1917|7/1917| 
|  `1997` |170/1917|51/1917|0/1917|

For example, when using the full 1997 dataset, 51/1917 language pairs score higher than 0.97, and 0/1917 score higher than 0.99. As expected, increasing the size of the dataset increases the difficulty. Additionally, the "stupid" model does not produce any scores > 0.99.

@KennethEnevoldsen Let me know what you think, then I'll make your suggested changes and this can be merged.

KennethEnevoldsen:
@imenelydiaker would love a second opinion here as well - what are your thoughts?

imenelydiaker:
> @imenelydiaker would love a second opinion here as well - what are your thoughts?

I just checked some other BitextMining tasks like BUCC, and I've seen prety perfect results too (see attached screenshot). 

This is making me question the whole task, are we evaluating this correctly? Is there still a need for a bitext mining evaluation?
Else, **we can still merge this PR** and open a discussion about BitextMining tasks. Maybe the idea here would be to check the datasets and get only recent/updated ones. Or only keep non-frequent language pairs, since some are so common (e.g. EN<->FR, EN<->DE, etc.) that they gain almost perfect scores.


image

KennethEnevoldsen:
Thanks, @imenelydiaker agree with the points, but it does sound like this is a merge at the moment. Once we have all the tasks merged in and we start running models we can empirically validate if bitext-mining tasks are predictive of performance on other tasks or whether they share a high variance with each other (and thus some of them could be removed). 

I could imagine that for high-resource languages they are less relevant, but for low-resource, this might have notable variance.

@davidstap thanks for the work (dankjewel voor de bijdrage, I seem to recall from my limited duolingo practice ;)). Thanks for taking the time to do the analysis as well. Let us go for 1997 samples pr language and I believe the current set of language pairs is very reasonable, it might be relevant to re-use the paired matrix to downsample some other datasets so might become handy elsewhere as well.

Please do fill out the points. For review please add two points for Imene and I.

davidstap:
Thanks @KennethEnevoldsen @imenelydiaker for your comments. 

I filled out the points: 2 for reviewing for both of you, and 68 for myself (New BitextMining languages (17): `div`, `fas`, `fil`, `fuc`, `hmn`, `lav`, `mey`, `mlg`, `msa`, `nde`, `nep`, `orm`, `pus`, `shi`, `swa`, `tah`, `ven`.)
)

@KennethEnevoldsen happy to help! (Nævn det ikke!) :)

This was my last dataset contribution, I should focus on other stuff for the next few weeks. After that I'm happy to help with evaluating models (I have a lot of available compute) / interpreting BitextMining results / paper writing.

imenelydiaker:
Thank you for this great work @davidstap ! Good luck wih the rest.

@KennethEnevoldsen should we merge this PR?

KennethEnevoldsen:
Thanks David and Imene, I have set this to merge once the tests pass (updated from main branch)

davidstap:
Seems like it doesn’t merge automatically @KennethEnevoldsen

imenelydiaker:
> Seems like it doesn’t merge automatically @KennethEnevoldsen 

I updated the branch, hopefully this time it'll merge.

PR #555, opened 2024-04-24
Fix minor bugs in FollowIR implementation; bump datasets revision id to the fixed version

orionw:
This PR fixes some minor bugs I found when running the full-scale results again:
- The length ablation setup had to be moved in the right place to get the correct keywords and short_instructions
- The counting of the dataset size metadata used all documents, but it was supposed to be relevant documents 
- Updates the dataset revision ID to match the fixed version

It also adds one new piece of functionality:
- Added the ability to download the `previous_results` used for reranking if the path is a link (helpful when establishing a shared baseline)


KennethEnevoldsen (approved):
Thanks @orionw - looks good to me!

PR #560, opened 2024-04-24
feat: Add MLQA dataset for CrossLingual Retrieval

imenelydiaker:
I have made an attempt to add MLQA cross lingual retrieval dataset. I've updated the `AbsTaskRetrieval` to handle cross lingual datasets. I ran the task on one language pair Arabic and German: the corpus is in Arabic and the questions in German, you can see the results in the uploaded json files for `intfloat/multilingual-e5-small`.

I'm still missing this task:

- [x] Better handle the language mapping and loading (I know how to do it, will do tomorrow)

Reason for dataset addition:


mteb/tasks/Retrieval/multilingual/MLQARetrieval.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+
+    def load_data(self, **kwargs):
+        if self.data_loaded:
+            return
+
+        _dataset_raw = {}
+        self.queries, self.corpus, self.relevant_docs = {}, {}, {}
+
+        for lang in _LANGUAGES.keys():
```
imenelydiaker, line 129: This loop loads all language pairs, I'll update it to use `self.langs` and upload only specified language pairs

mteb/tasks/Retrieval/multilingual/MLQARetrieval.py
```diff
@@ -0,0 +1,162 @@
...
+}
+
+
+def extend_lang_pairs() -> dict[str, list[str]]:
+    # add all possible language pairs
+    hf_lang_subset2isolang = {}
+    for x in _ISO6393_to_ISO6391.keys():
+        for y in _ISO6393_to_ISO6391.keys():
+            if x != y:
+                pair = f"{x}_{y}"
```
isaac-chung, line 74: Would each pair be query in xx -> corpus in yy?

imenelydiaker: The dataset has a subset for each language pair xx->yy, I'd prefer not mixing them to avoid errors.

So what I did for now is use each subset: xx->yy on it's own. Since all texts are machine translated. 

You can see the subsets in the _LANGUAGES variable. Although, I'll change the way I'm handling these pairs because this is not very clear to understand.

isaac-chung: Oh yes I'll add to the discussion. I think whichever convention we end up choosing, we just need to write it down somewhere. I tend to forget 🙈 , so it'll be helpful to be able to check from time to time.

imenelydiaker:
@isaac-chung this PR is ready for review. I ran both models on one language pair, but a compelete evaluation is coming (takes some time for 49 pairs).

I didn't add the named tuple as discussed here, as it requires to update all CrossLingual tasks. I prefer adding this in another PR with all necessary refactorings.

For the now, the language is handled this way:
```py
eval_langs = {
"eng-Latn_fra-Latn": ["eng-Latn", "fra-Latn"],
...
}
```

We can call the task  on a language pair (e.g., from English to Spanish) like this:
```bash
mteb -t MLQARetrieval -m intfloat/multilingual-e5-small -l eng-Latn_spa-Latn
```

mteb/tasks/Retrieval/multilingual/MLQARetrieval.py
```diff
@@ -0,0 +1,152 @@
...
+        if self.data_loaded:
+            return
+
+        _dataset_raw = {}
+        self.queries, self.corpus, self.relevant_docs = {}, {}, {}
+
+        for hf_subset, langs in _LANGUAGES.items():
+            lang_pair = (
+                langs[0] + "_" + langs[1]
+            )  # builds a language pair separated by an underscore. e.g., "ara-Arab_eng-Latn"
```
isaac-chung, line 119: What do you think about, instead of `eng-Latn_fra-Latn`, we only use the 3 letter lang code, i.e. `eng_fra` ? As they should be unique, it saves a lot of typing. 

Aside: could also consider making line 118 into its own small function as it's used more than once.

isaac-chung (commented):
I think the rest looks good. Just a few small comments. 

Separating the named tuple into its own PR sounds like a good plan 👍

mteb/tasks/Retrieval/multilingual/MLQARetrieval.py
imenelydiaker: I like this proposition, I'm afraid the format is not adequate to our language standard. I'll give it a try and let you know!

isaac-chung: It should be fine. When passed into `-l`, MTEB calls `_extend_lang_code`, which does a mapping if only a 3 letter code is provided. So both standards should be supported without problems.

Andrian0s (approved):

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -246,7 +246,7 @@ def evaluate(self, model, split="test", **kwargs):
         retriever = RetrievalEvaluator(model, **kwargs)
 
         scores = {}
-        if self.is_multilingual:
+        if self.is_crosslingual or self.is_multilingual:
             for lang in self.langs:
                 logger.info(f"Language: {lang}")
                 corpus, queries, relevant_docs = (
```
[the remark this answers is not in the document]
imenelydiaker: True, should I name it eval_split as in this task: [AbsTaskSTS](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/abstasks/AbsTaskSTS.py) ?

mteb/tasks/Retrieval/multilingual/MLQARetrieval.py
```diff
@@ -0,0 +1,154 @@
...
+        dialect=[],
+        text_creation="found",
+        bibtex_citation="""@article{lewis2019mlqa,
+        title = {MLQA: Evaluating Cross-lingual Extractive Question Answering},
+        author = {Lewis, Patrick and Oguz, Barlas and Rinott, Ruty and Riedel, Sebastian and Schwenk, Holger},
+        journal = {arXiv preprint arXiv:1910.07475},
+        year = 2019,
+        eid = {arXiv: 1910.07475}
+        }""",
+        n_samples={"test": 158083, "validation": 15747},
```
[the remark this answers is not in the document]
imenelydiaker: Yeah for now the number of samples for multilingual and crosslingual tasks is just equal to the sum of samples accross all languages and pairs. We're still thinking about how to implement a faster version of tasks for Retrieval, just as we did in [Clustering,](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/abstasks/AbsTaskClusteringFast.py) do you have any suggestions? We can't simply subsample imo, because it will impact the ndcg score..

imenelydiaker:
@isaac-chung @Andrian0s I fixed the code to build language pair in this format: `lang1-lang2` (see screenshot below), and renamed `_eval_monolingual()` in `AbsTaskRetrieval`.I think it's ready for merging 🙂 
image

mteb/tasks/Retrieval/multilingual/MLQARetrieval.py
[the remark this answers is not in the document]
Andrian0s: I agree, downsampling is quite bad, and the large amount of samples issue itself in this pull request spawns from the amount of language pairs, which is natural and unavaoidable. 

A non trivial solution would be very nice, but I would hope for a proper reasoning defending that choice.

For now, I would keep as is (since the embeddings are currently not calculated multiple times, so all good). I would still suggest to avoid the validation set, test set is enough.

Andrian0s:
> @isaac-chung @Andrian0s I fixed the code to build language pair in this format: `lang1-lang2` (see screenshot below), and renamed `_eval_monolingual()` in `AbsTaskRetrieval`.I think it's ready for merging 🙂
> image

Looks good. If the tests pass, I would also support merging.

imenelydiaker:
Okay I may have broken something, tests are all good locally.. I'm checking

isaac-chung (approved):
Thanks for adding this Imene!

Issue #566, opened 2024-04-25
Tests failing due to not found datasets


isaac-chung:
example test: https://github.com/embeddings-benchmark/mteb/actions/runs/8822732225/job/24235198184?pr=485

```
=========================== short test summary info ============================
FAILED tests/test_all_abstasks.py::test_dataset_availability - AssertionError: Datasets not available on Hugging Face:
  jhu-clsp/core17-instructions - revision e783a88ec6bc4bbdc1ba998edd85edeaf3d820f7
  jhu-clsp/news21-instructions - revision fae860d870141ee0f34fbab67314c65f41921943
  jhu-clsp/robust04-instructions - revision d25e82647b09251f4feaaf5462c9a18445072ba6
assert False
======= 1 failed, 384 passed, [97](https://github.com/embeddings-benchmark/mteb/actions/runs/8822732225/job/24235198184?pr=485#step:5:98) skipped, 2 warnings in 65.58s (0:01:05) =======
make: *** [Makefile:23: test] Error 1
Error: Process completed with exit code 2.
```

isaac-chung:
CC @orionw

isaac-chung:
Cause: Older PRs had not pulled `main`.

KennethEnevoldsen:
@isaac-chung I have set a required for branches to be up to date before merging

orionw:
Ah great, thanks @KennethEnevoldsen and @isaac-chung for the quick debugging!

These dataset’s revision IDs shouldn’t change anymore, but I agree it’s good practice to require them to be up to date before merging, in case someone else makes new tests.

PR #567, opened 2024-04-25
Use old datasets to match revisions

isaac-chung:
Addresses https://github.com/embeddings-benchmark/mteb/issues/566 for now.


isaac-chung:
Hmm seems like there's a blip and we can see those revisions again in the original dataset.

imenelydiaker:
I think the path is the dataset HF name and that's what will be used to find it on the hub

isaac-chung:
Older PRs saw those failures because `main` wasn't pulled. I'll close this PR and the issue as well.

Issue #530, opened 2024-04-23
Slow loading for datasets with a high number of language pairs


loicmagne:
I'm opening this issue to discuss changes to how some datasets are loaded

I wrote more details about this issue here https://github.com/embeddings-benchmark/mteb/pull/386#issuecomment-2072443114. The TLDR is that when you have a dataset with a high number of subsets (in our case this is often language tags), they are currently loaded iteratively one subset at a time, with a `load_dataset` call: https://github.com/embeddings-benchmark/mteb/blob/5370b44435e63a9c00e76db51d01812824b559d0/mteb/abstasks/MultilingualTask.py#L26-L30 https://github.com/embeddings-benchmark/mteb/blob/5370b44435e63a9c00e76db51d01812824b559d0/mteb/abstasks/CrosslingualTask.py#L24-L27
This turns out to be way slower than simply loading the same amount of data in one chunk. The reason is in part because many network request must be done to query each file, but mostly because the datasets lib has a constant overhead when loading datasets where it checks for newer versions of each files. Importantly, even when the dataset is cached, loading it is still slow: in this PR https://github.com/embeddings-benchmark/mteb/pull/330, loading a dataset with 250 subsets would take 15 minutes from the network, and 12 minutes from cache

This issue has been there from the very beginning, but isn't too noticeable when the number of subsets/languages is low, which is the case for most datasets. However for crosslingual tasks like bitext mining where the number of language pairs can be very large, some datasets can take hours to load (https://github.com/embeddings-benchmark/mteb/pull/386, https://github.com/embeddings-benchmark/mteb/pull/330), so it would be relevant to solve this in the context of MMTEB and speeding up things https://github.com/embeddings-benchmark/mteb/issues/381

I ran some experiments with this dataset https://huggingface.co/datasets/loicmagne/open-subtitles-bitext-mining , which contains 1759 language pairs (= subsets), where each subset contains 1000 sentence pairs, which in total is ~300MB of data. Loading it in the iterative way from the network takes around an hour (I don't have the exact number but it's long)

## Solution 1

From https://github.com/huggingface/datasets/issues/6800 there is a way to load multiple subsets all at once with load_datasets:
```python
data_files = "data/*.jsonl"
ds = load_dataset("loicmagne/open-subtitles-250-bitext-mining", data_files=data_files, split="train")
```
or using this loading config:
```python
  - config_name: all
    data_files: data/*.jsonl
```

[...]

PR #572, opened 2024-04-25
Fast loading for cross lingual tasks

loicmagne:
This PR implements the changes discussed here https://github.com/embeddings-benchmark/mteb/issues/530

I only implemented the fast loading for Tatoeba for testing purpose, I'll do another PR to convert others datasets

I created a different dataset on HF for testing purpose (will swap to mteb later), but note that this is fully backward compatible, people with old versions of the MTEB package will be able to load fast datasets as it just add a 'lang' column

I checked that the results between the slow and fast version are the same. Data loading goes from ~10min from network to ~30s on Tatoeba which "only" has ~100 subsets


mteb/tasks/BitextMining/multilingual/TatoebaBitextMining.py
```diff
@@ -147,3 +147,4 @@ class TatoebaBitextMining(AbsTaskBitextMining, CrosslingualTask):
         n_samples={"test": 2000},
         avg_character_length={"test": 39.4},
     )
+    fast_loading = True
```
loicmagne, line 150: I don't know if this is the best place place to toggle it, but I thought that since it's more of an implementation detail it wouldn't be useful to put it as a metadata

mteb/abstasks/CrosslingualTask.py
```diff
@@ -20,6 +20,31 @@ def load_data(self, **kwargs):
...
+            self.fast_load()
+        else:
+            self.slow_load()
+
+    def fast_load(self, **kwargs):
+        """Load all subsets at once, then group by language with Polars"""
+        self.dataset = {}
+        merged_dataset = datasets.load_dataset(
+            **self.metadata_dict["dataset"]
+        )  # load "default" subset
```
loicmagne, line 35: This loads all the subsets and then filter the needed languages. We could load granularly only the required languages as shown here: https://github.com/huggingface/datasets/issues/6800#issuecomment-2057225862

That would require to use a standard data path for each dataset, along with handling train/test/eval splits, I feel like the overhead of downloading more data is ok since datasets shouldn't be huge anyway

loicmagne:
This test is failing because of the mocking I don't exactly understand why

https://github.com/embeddings-benchmark/mteb/blob/28e55225dd2fd4d68f6598fae5f3be43cac7e30b/tests/test_all_abstasks.py#L19

mteb/tasks/BitextMining/multilingual/TatoebaBitextMining.py
KennethEnevoldsen: It is not really metadata. I would move it to:

```
fast_loading = True
metadata = TaskMetadata( ...
)
```

We might not require a toggle if we just convert all the bitext task to the new format.

KennethEnevoldsen:
This looks very good @loicmagne:

A main considerations:
- I think we should just use this as the new default (no reason to support two methods)
  - We can make a script to transfer the datasets and re-upload them
  - There are a few STS tasks that use cross-lingual (we might consider those as well)

mteb/abstasks/CrosslingualTask.py
```diff
@@ -20,6 +20,31 @@ def load_data(self, **kwargs):
...
+        for split in self.metadata.eval_splits:
+            grouped_by_lang = dict(merged_dataset[split].to_polars().group_by("lang"))
+            for lang in self.langs:
+                if lang not in self.dataset:
+                    self.dataset[lang] = {}
+                self.dataset[lang][split] = datasets.Dataset.from_polars(
+                    grouped_by_lang[lang].drop("lang")
```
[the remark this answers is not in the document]
loicmagne: Indeed it's a good idea to document it somewhere

mteb/abstasks/CrosslingualTask.py
loicmagne: The thing is that with load_datasets you can't really select the subsets to load, what you do is to select all the files that you want to load. So if a subset "en-fr" is split into multiple files 0001.jsonl 0002.jsonl etc. we would have to account for that in the loading 

But again maybe we can just make it a constraint to format dataset with one file per split/subset

loicmagne:
I've added documentation for fast loading, and I've made it a mixin instead a specific implementation of CrosslingualTasks. This way we can directly use it with MultilingualTask, as some of these tasks might also benefit it


> This looks very good @loicmagne:
> 
> A main considerations:
> 
>     * I think we should just use this as the new default (no reason to support two methods)
>       
>       * We can make a script to transfer the datasets and re-upload them
>       * There are a few STS tasks that use cross-lingual (we might consider those as well)

Yes I think it would be fair to use this as the default. I'm not really sure how to automate the conversion with a script though, since it also requires changing the config files ? Currently I did it manually depending on the format of each dataset but maybe there's a simpler way

For now I would let "fast loading" as a toggled option until everything is converted

KennethEnevoldsen (approved):
@loicmagne I think this looks very solid. Will you add points (I believe this is worth a solid 10?) and then I will merge it in

Will you also create an issue on converting current datasets to the new format

loicmagne:
I moved the updated Tatoeba dataset to the MTEB org on HF, I think we can merge now @KennethEnevoldsen if everything's ok

I'll do the PR to update other datasets

loicmagne:
@KennethEnevoldsen can we merge this ?

KennethEnevoldsen:
@loicmagne I think we def. can. I will set it to auto-merge

PR #574, opened 2024-04-25
UkrFormalityClassification added

dokato:
Translated English GYAFC data to Ukrainian: https://huggingface.co/datasets/ukr-detect/ukr-formality-dataset-translated-gyafc .

Reason for dataset addition:


mteb/tasks/Classification/ukr/UkrFormalityClassification.py
```diff
@@ -0,0 +1,67 @@
...
+        n_samples={"test": 4853},
+        avg_character_length={"test": 52.39},
+    )
+
+    def dataset_transform(self):
+        self.dataset = self.dataset.rename_column("labels", "label")
+        self.dataset = self.dataset.class_encode_column("label")
+        self.dataset = self.dataset["test"].train_test_split(
+            test_size=0.5, seed=self.seed, stratify_by_column="label"
+        )  # balanced sampling
```
imenelydiaker, line 55: You can use our helper function like this:
```suggestion
        self.dataset = self.stratified_subsampling(self.dataset, seed=self.seef, splits=["test")
```

mteb/tasks/Classification/ukr/UkrFormalityClassification.py
```diff
@@ -0,0 +1,67 @@
...
+if __name__ == "__main__":
+    from sentence_transformers import SentenceTransformer
+
+    from mteb import MTEB
+
+    # intfloat/multilingual-e5-small
+    model_name = "intfloat/multilingual-e5-small"
+    model = SentenceTransformer(model_name)
+    evaluation = MTEB(tasks=[UkrFormalityClassification()])
+    evaluation.run(model)
```
imenelydiaker, line 67: You can also remove this before merging

imenelydiaker (commented):
LGTM, I'll let you address my comments and add points then we'll be good to merge 🙂

mteb/tasks/Classification/ukr/UkrFormalityClassification.py
dokato: I don't think it does the same. I only stratifies "test", leaving large train and validation intact, whereas here, I split the "test" subset into train and test to reduce the overall sample size.

dokato:
Should it be +4 for language in general or lang in task @imenelydiaker ? Also, doe we have some convenient way to list of all lang per task?

imenelydiaker:
> Also, doe we have some convenient way to list of all lang per task?

I'm not sure what you mean here?

mteb/tasks/Classification/ukr/UkrFormalityClassification.py
imenelydiaker: Okay I see, I tend to not add datasets where the train set is not provided, but I see what you mean. Following your comment on PR #575, @isaac-chung so for tasks where the dataset does not contain a train split, we should allow creating it from the test set?

dokato:
> I'm not sure what you mean here?

Like a function that crawls through all tasks and lists unique language for Classification, Retrieval etc. I think as the number of datasets grow, it'd be convenient.

imenelydiaker:
> > I'm not sure what you mean here?
> 
> Like a function that crawls through all tasks and lists unique language for Classification, Retrieval etc. I think as the number of datasets grow, it'd be convenient.

I'm not sure we have this implemented, @KennethEnevoldsen may know if we already have this? Else, feel free to open a PR with the proposed implementation

mteb/tasks/Classification/ukr/UkrFormalityClassification.py
isaac-chung: I think we can allow it, as long as we state that clearly somewhere, like "For datasets without train splits, we create a train split (by doing XYZ)". I'm curious if @KennethEnevoldsen has an opinion on this too.

dokato: Ok, sorry, I overlooked. This actually has training and val split, so I made it with `stratified_subsampling` as requested.

imenelydiaker:
@dokato please update your branch and add points and we'll be good to merge 🙂

PR #580, opened 2024-04-26
Fix instructions in DRESModel

orionw:
I missed one area when committing last time, super sorry about that @KennethEnevoldsen... 

The results are all done though and the leaderboard results PR is in progress


KennethEnevoldsen (approved):
No worries will push this in

PR #545, opened 2024-04-24
SIB200Classification dataset added

dokato:
Dataset from https://huggingface.co/datasets/Davlan/sib200 
Here's the paper: https://arxiv.org/abs/2309.07445 

## Checklist for adding MMTEB dataset

Reason for dataset addition:

- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [x] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 
- [x] I have added points for my submission to the [points folder](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) using the PR number as the filename (e.g. `438.jsonl`).

PR #583, opened 2024-04-26
Add BibleNLP dataset

davidstap:

This dataset has an extremely large coverage of 829 languages, and includes numerous low-resource languages. For this reason I only considered English-centric directions, resulting in a reasonable 1656 pairs. I will add performance results later, but initial testing suggests that while quality is not good for most (low-resource) languages, it is far better than random guesses, so IMO a good addition to the benchmark.


davidstap:
Could you take a look at this PR  @KennethEnevoldsen @imenelydiaker? I'm personally quite excited by the large language coverage.

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
```diff
@@ -0,0 +1,955 @@
...
+    "zpu_Latn",  # Yalálag Zapotec
+    "zpv_Latn",  # Chichicapan Zapotec
+    "zpz_Latn",  # Texmelucan Zapotec
+    "zsr_Latn",  # Southern Rincon Zapotec
+    "ztq_Latn",  # Quioquitani-Quierí Zapotec
+    "zty_Latn",  # Yatee Zapotec
+    "zyp_Latn",  # Zyphe Chin
+]
+
+_SPLIT = ["train"]
```
isaac-chung, line 843: Probably good to note why the train split is chosen here since there are also validation and test splits a well.

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
```diff
@@ -0,0 +1,955 @@
...
+_LANGUAGES_MAPPING = extend_lang_pairs_english_centric()
+
+
+class BibleNLPBitextMining(AbsTaskBitextMining, CrosslingualTask):
+    metadata = TaskMetadata(
+        name="BibleNLPBitextMining",
+        dataset={
+            "path": "davidstap/biblenlp-corpus-mmteb",
+            "revision": "264a18480c529d9e922483839b4b9758e690b762",
+            "split": f"train[:{_N}]",
```
isaac-chung, line 872: Cool - I wasn't familiar with this convention.

isaac-chung (commented):
This is awesome! Just a few comments.

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
davidstap: Makes sense. I went for the train split because the validation/test splits are extremely small in a lot of cases (think 20 sentences). Given this I thought it would be better to subsample the train set.

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
```diff
@@ -0,0 +1,955 @@
...
+            "trust_remote_code": True,
+        },
+        description="Partial Bible translations in 829 languages, aligned by verse.",
+        reference="https://arxiv.org/abs/2304.09919",
+        type="BitextMining",
+        category="s2s",
+        eval_splits=_SPLIT,
+        eval_langs=_LANGUAGES_MAPPING,
+        main_score="f1",
+        date=("2023-04-09", "2023-04-09"),
```
dokato, line 882: Here is an interesting one, cause [for consistency](https://github.com/embeddings-benchmark/mteb/pull/545#discussion_r1577555935) it should be the date of when the source was written 🤔 or maybe in this case a period when the translations where made if such information available? One way or another it requires some adnotation in the comment.

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
```diff
@@ -0,0 +1,955 @@
...
+
+_SPLIT = ["train"]
+
+_N = 256
+
+
+def extend_lang_pairs_english_centric() -> dict[str, list[str]]:
+    # add all language pairs with english as source or target
+    hf_lang_subset2isolang = {}
+    for x in _LANGUAGES:
```
dokato, line 851: if we know that one of the languages is english, we don't need to iterate over twice.
consider sth like:
```py
for lang in _LANGUAGES:
  pair = f"eng_Latn-{lang}"
  hf_lang_subset2isolang[pair] = ['eng_Latn', lang.replace("_", "-")]
  pair = f"{lang}-eng_Latn"
  hf_lang_subset2isolang[pair] = [lang.replace("_", "-"), 'eng_Latn']

```

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
```diff
@@ -0,0 +1,955 @@
...
+        task_subtypes=[],
+        license="CC-BY-SA-4.0",
+        socioeconomic_status="medium",
+        annotations_creators="expert-annotated",
+        dialect=[],
+        text_creation="created",
+        n_samples={"train": _N},
+        avg_character_length={"train": 120},
+        bibtex_citation="""
+@article{akerman2023ebible,
```
dokato, line 894: if you improve indentation here

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
```diff
@@ -0,0 +1,955 @@
...
+                )
+                self.dataset[lang] = datasets.DatasetDict({"train": dataset})
+                seen_pairs.append(hf_lang_name)
+
+        self.dataset_transform()
+        self.data_loaded = True
+
+    def dataset_transform(self) -> None:
+        # Convert to standard format
+        for lang in self.langs:
```
dokato, line 929: if we edit load data already, I don't see why we would need to iterate over `self.langs` twice? maybe we don't need `dataset_transform` at all?

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
```diff
@@ -0,0 +1,955 @@
...
+            l1, l2 = [l.split("_")[0] for l in lang.split("-")]
+
+            self.dataset[lang] = self.dataset[lang].rename_columns(
+                {l1: "sentence1", l2: "sentence2"}
+            )
+
+    @staticmethod
+    def _transform_lang_name_hf(lang):
+        # Transform language name to match huggingface configuration
+        langs = [l.split("_")[0] for l in lang.split("-")]
```
dokato, line 939: if we want to detect english, this looks like a regex job to me

dokato (commented):
Really interesting dataset :) find some of my suggestions below

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
davidstap: You're right, thanks for catching. I lazily copied this over from some other task, will change to your suggestion.

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
davidstap: Sure we can do the transformation in `load_data` to prevent iterating twice, I'll change it.

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
davidstap: I see, interesting indeed. For English, I mostly used the World English Bible (WEB) versions. On Wikipedia I can find: first draft in 1997, completed in 2020, so I will add this information. 

It seems impossible, or at the minimum infeasible, to add information for all other languages.

mteb/tasks/BitextMining/multilingual/BibleNLPBitextMining.py
davidstap: On second thought: This would not be an improvement because it requires different transformations for if else. (The if else is there to prevent double loading of e.g. 'aai_Latn-eng_Latn' and 'eng_Latn-aai_Latn', which is the same huggingface data, and loading data is slow #572).

IMO the current code is a better trade-off between readability and efficiency. The impact of the second loop is negligible; it is a very cheap operation over 829*2 items.

davidstap:
I have addressed your comments @isaac-chung @dokato, thanks for your suggestions!

Results are included now as well. `intfloat/multilingual-e5-small` performs best, but like I pointed out the performance is bad for most low-resource languages. However, almost all directions are way better than random chance (1/256), so this seems like a valuable addition to evaluate (future) multilingual models.

I think this is ready to merge now, do you agree? Let me know, then I'll calculate the points. Thanks!

isaac-chung (approved):
Thanks for iterating! I think it's good enough to merge. In terms of points, the dataset is 2 + 4 * number of new languages added to the BitextMining task. Please check if any of the languages have already been covered in that folder and the multilingual subfolder as well. Then 2 pts each to @dokato and I for the reviews.

davidstap:
Thanks for your quick response! I calculated the number of new languages using [this script](https://gist.github.com/davidstap/787acb00fecb24559439b08cac30c5ea): 756 new languages.

I'll update the points and merge.

davidstap:
Ready to merge now @isaac-chung (I don't have write access)

PR #589, opened 2024-04-27
count_languages func added

dokato:
Following converstaion with @imenelydiaker here: https://github.com/embeddings-benchmark/mteb/pull/574#issuecomment-2078973652 I propose an auxiliary function `count_languages` that summarises count of languages per tasks, domains, categories. I think as the number of tasks grows it might be a convenient way to browse existing sets.

Two doubts here:
- there's inconsistency here in `get_tasks` documentation, where example shows `task_types` accepting only str, but type suggests it should go by list; which one should it be?
- maybe we should change name of this submodule from `get_tasks` to `utils`, `helpers` or something like that as we may add more auxiliary functions for MMTEB analysis later?


mteb/get_tasks.py
```diff
@@ -103,8 +105,8 @@ def get_tasks(
 
     Examples:
         >>> get_tasks(languages=["eng", "deu"], script=["Latn"], domains=["Legal"])
-        >>> get_tasks(languages=["eng"], script=["Latn"], task_type="Classification")
-        >>> get_tasks(languages=["eng"], script=["Latn"], task_type="Clustering", exclude_superseeded_datasets=False)
+        >>> get_tasks(languages=["eng"], script=["Latn"], task_types=["Classification"])
+        >>> get_tasks(languages=["eng"], script=["Latn"], task_types=["Clustering"], exclude_superseeded_datasets=False)
```
imenelydiaker, line 109: I think that making task_types a list will break some scripts that use this function, see: _docs/create_tasks_table.py_.
Have you tried running the tests locally with `make test` or `make pr`?

imenelydiaker (commented):
Thank you for adding this! See below my comment

mteb/get_tasks.py
dokato: Not sure if I follow. `create_tasks_table` doesn't contain any call to `get_tasks` with `task_types` argument. If anything I made this example consistent with typing above. Or in fact I made it work. Previosuly:
```
> get_tasks(languages=["eng"], script=["Latn"], task_types="Clustering", exclude_superseeded=False)
[]
```

now:
```
> get_tasks(languages=["eng"], script=["Latn"], task_types=["Clustering"], exclude_superseeded=False)
[ArxivClusteringP2P(name='ArxivClusteringP2P', languages=['eng']),
....
 MasakhaNEWSClusteringS2S(name='MasakhaNEWSClusteringS2S', languages=['amh', 'eng', 'fra', '...'])]
```
PS I caught another bug here `exclude_superseeded_datasets` -> `exclude_superseeded`

>  Have you tried running the tests locally with make test or make pr?

Yes, but OOC do my local test differ at all form test run by CI? all were green below.

dokato:
Sorry @imenelydiaker @KennethEnevoldsen I’d like to know your stance also on the point 2 above 👆 before I “unwip” it. That is, WDYT of  changing get_tasks.py module name to utils, helpers etc. ?

dokato:
One specific use-case is counting points per task for adding new languages. Second, at some point, we'll be interested in summarizing various categories per language and this makes it explicit.

KennethEnevoldsen:
> One specific use-case is counting points per task for adding new languages. 

This is already covered by get_task


> Second, at some point, we'll be interested in summarizing various categories per language and this makes it explicit.

Seems like we might instead want some kind of kind of overview function for this. I could imagine something like this: 

```
list_of_tasks = get_tasks(...)
type(list_of_tasks)
# MTEBTasks
print(list_of_tasks)
# MTEBTasks(languages: {eng, dan, ...},  tasks=[DKHate, ...], ...)
isisntance(MTEBTasks, list)
# True
```

Where MTEBTasks is an overview object over a lists of tasks, which e.g. allows you to get something like:
```
len(list_of_tasks.languages)
```
or something like:
```
list_of_tasks.to_markdown()
```

to create a markdown report.

mteb/get_tasks.py
imenelydiaker: Okay, I thougth _doc/create_tasks_table.py_ uses filters, but for now the filters are not used so this change won't break anything

dokato:
Thanks @KennethEnevoldsen. I like the idea, although I still think that language summary is a neat feature. See the revised proposal.

KennethEnevoldsen:
Thanks @dokato I think this approach is gonna be much easier to expand in the future - thanks for taking the time to discuss it

KennethEnevoldsen:
Assuming @imenelydiaker agree, feel free to merge

imenelydiaker (approved):

PR #601, opened 2024-04-29
add xsimplusplus in retrieval category

jaygala24:

Reason for dataset addition:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mteb"
-version = "1.7.49"
+version = "1.7.50"
 description = "Massive Text Embedding Benchmark"
 readme = "README.md"
 authors = [


jaygala24:
Adding [xsim++ dataset](https://arxiv.org/abs/2212.05409) spanning over 200 languages in retrieval category task.

mteb/tasks/Retrieval/multilingual/XSimPlusPlusRetrieval.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+class XSimPlusPlusRetrieval(MultilingualTask, AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="XSimPlusPlusRetrieval",
+        dataset={
+            "path": "jaygala24/xsimplusplus",
+            "revision": "07f92f877ea651659f3815884761c49191a1c80c",
+        },
+        description="xsim++ is dataset created to capture more subtle improvements in bitext mining by adding challenging negative examples.",
```
KennethEnevoldsen, line 226: Bitext mining? What is the reason we formulate it as a retrieval task then? Is it because the bitext mining task assumes only positive pairs?

mteb/tasks/Retrieval/multilingual/XSimPlusPlusRetrieval.py
```diff
@@ -0,0 +1,298 @@
...
+        dialect=[],
+        text_creation="machine-translated and verified",
+        bibtex_citation="""@article{chen2023xsim,
+  title     = {xSIM++: An Improved Proxy to Bitext Mining Performance for Low-Resource Languages},
+  author    = {Mingda Chen and Kevin Heffernan and Onur cCelebi and Alexandre Mourachko and Holger Schwenk},
+  journal   = {Annual Meeting of the Association for Computational Linguistics},
+  year      = {2023},
+  doi       = {10.48550/arXiv.2306.12907},
+}""",
+        n_samples={"dev": 997, "devtest": 1012},
```
KennethEnevoldsen, line 249: Is there really only 1012 samples in total?

Any reason to use both the dev and devtest?

KennethEnevoldsen (commented):
A question related to the task formulation

mteb/tasks/Retrieval/multilingual/XSimPlusPlusRetrieval.py
jaygala24: The default formulation for bitext mining expects one-to-one pair mapping between source and target translations where all the examples except the corresponding index act as negative samples. However, in the case of xsim++, there are a variable number of negative samples for each of the English sentences based on the perturbation choices and note that there is no corresponding positive pair (perturbation applied on target language sentences). 

The most appropriate task formulation that we think fits with the xsim++ dataset in MTEB is retrieval where each query (target language sentence) can have a variable number of positive (ground truth English sentence) and negative pairs (perturbed English sentences). This is similar to the [following task](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/multilingual/MIRACLRetrieval.py) in the retrieval category as of now.

mteb/tasks/Retrieval/multilingual/XSimPlusPlusRetrieval.py
jaygala24: The dataset has 1012 samples per language and there are 200 languages so in total it would amount to 200k samples.

The dataset has both dev and devtest splits but we can choose to only include the devtest as the eval split as people usually report the result for the same.

mteb/tasks/Retrieval/multilingual/XSimPlusPlusRetrieval.py
KennethEnevoldsen: Thanks @jaygala24, my suspicion was that it was due to the negative examples as well. Do note that it is possible to introduce new task types to MTEB, however, let us keep it as a retrieval task for now.

mteb/tasks/Retrieval/multilingual/XSimPlusPlusRetrieval.py
KennethEnevoldsen: We should remove the dev set (200k is already on the higher end). Please add the fast loading introduced in #572 to avoid extensive long loading times.

KennethEnevoldsen: Then feel free to add the points as well as updating the scores using the fast implementation (just to show that the implementation work as intended)

mteb/tasks/Retrieval/multilingual/XSimPlusPlusRetrieval.py
Andrian0s: @KennethEnevoldsen 

Ah now I followed through this conversation. Xsim++ was developed to evaluate Bitext mining systems cross-lingually. 

Implementation wise, a Retrieval Evaluation code re-use makes sense. However, the retrieval metrics are not really relevant to the task (there are no evaluations of them within the paper), as they only mention their pairwise models accuracy. Perhaps I would suggest finding a case where this dataset was used beyond pairwise comparison and providing the same metrics.

Conceptually, it is an adversarial evaluation, similar to what Cross Lingual Semantic Discrimination Task (pr coming) will be.  When I am through with that addition, it might make sense to have xsim++ as a tab within that section

Andrian0s: And one more thing, using the retrieval code, you are encoding the "corpus" - which is again similar size text to the original using model.encode_corpus. In models where this matter (e5-multilingual for example), this would give the wrong prompt to the model, altering the results.

Consider adding a binary variable in RetrievalEvaluator within the search function, which will allow for the case that the corpus also needs to be encoded using the encode_queries. Should be an easy fix and will bring difference in results.

PR #604, opened 2024-04-29
get_task now filters languages in multilingual tasks

KennethEnevoldsen:
- get_tasks previously did not filter languages in multilingual datasets, but now it does
- bug fix for bibleNLP 

Added 3 points for get_tasks and 1 point for bugfix


imenelydiaker (approved):
LGTM! Great work 🙂

PR #408, opened 2024-04-17
Add AbsTaskInstructionRetrieval from the FollowIR paper

orionw:
Real PR for FollowIR (draft in #356).

Re-uses as much of the `AbsTaskRetrieval` abstract task as possible, thanks to previous feedback.  Dataset task files are up to date with the newer MTEB metadata.

I left the sample and n_characters metadata empty, I'm updating the dataset slightly (should be done in a few days) and will make a new PR when it's finished to update the dataset revisions and this metadata.

PR #608, opened 2024-05-01
add prompt retrieval

hongjin-su:
Add the prompt retrieval for embedding model evaluation.


Muennighoff:
This is cool! But I think we should integrate it with the existing MTEB interface or does that not make any sense?

cc'ing @dipam7 @KennethEnevoldsen from the issue

Muennighoff:
Here's a great example of a successful integration: https://github.com/embeddings-benchmark/mteb/pull/408
Tagging @orionw here in case he has ideas 😊

KennethEnevoldsen:
Thanks for this PR @hongjin-su, completely agree with @Muennighoff that we would love to integrate this with the existing code-base.

KennethEnevoldsen:
@hongjin-su seems like this PR has gone stale. I will close it for now, but do feel free to re-open it.

PR #611, opened 2024-05-01
fix: Add LegalBench datasets - 4

awinml:
## Checklist for adding MMTEB dataset

Reason for dataset addition: Showcases the model's performance on the [LegalBench](https://hazyresearch.stanford.edu/legalbench/) datasets, enabling evaluation on domain-specific legal texts.

Follow up to #579

## Changes made:

- All the `LegalBenchClassification` tasks have been reorganized for better readability. Add additional citations for a few `ContractNLI` tasks.
- The descriptions for all the tasks have been updated to include more information about the task.

The following 8 datasets have been added in this PR:

1. [CUAD Cap on Liability](https://hazyresearch.stanford.edu/legalbench/tasks/cuad_cap_on_liability.html)
2. [CUAD Change of Control](https://hazyresearch.stanford.edu/legalbench/tasks/cuad_change_of_control.html)
3. [CUAD Competitive Restriction Exception](https://hazyresearch.stanford.edu/legalbench/tasks/cuad_competitive_restriction_exception.html)
4. [CUAD Covenant not to sue](https://hazyresearch.stanford.edu/legalbench/tasks/cuad_covenant_not_to_sue.html)
5. [CUAD Effective Date](https://hazyresearch.stanford.edu/legalbench/tasks/cuad_effective_date.html)
6. [CUAD Exclusivity](https://hazyresearch.stanford.edu/legalbench/tasks/cuad_exclusivity.html)
7. [CUAD Expiration Date](https://hazyresearch.stanford.edu/legalbench/tasks/cuad_expiration_date.html)
8. [CUAD Governing Law](https://hazyresearch.stanford.edu/legalbench/tasks/cuad_governing_law.html)

- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [x] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
[...]

PR #612, opened 2024-05-01
fix: Add points for LegalBench datasets - 4 (#611)

awinml:
## Proposed Changes

Add points for #611


docs/mmteb/points/611.jsonl
```diff
@@ -0,0 +1,2 @@
+{"GitHub": "awinml", "New dataset": 16}
```
KennethEnevoldsen, line 1: Wasn't it just fixes to existing datasets or did I misunderstand?

KennethEnevoldsen: Re-examining the PR it seems like most things look good. The only thing (which you can probably change in this PR), would be that the dates doesn't seem to correspond with when the period text was written. If you could update that I believe everything else is perfect.

KennethEnevoldsen: Actually, to avoid merge conflicts I will let to fix the dates in the new #613.

PR #615, opened 2024-05-02
fix: BSARDRetrieval dataset

wissam-sib:
I did some testing on BSARDRetrieval here: https://colab.research.google.com/drive/1QMoEDoBlsGQlieSStvbSIShvZz44_o5z?usp=sharing
And I was able to produce good results with TF-IDF (recall@100 around 47% comparable to the [paper](https://aclanthology.org/2022.acl-long.468.pdf)). I also tested sentence-camembert-base and the results were way higher than [what we had](https://github.com/embeddings-benchmark/mteb/blob/9b648a0f6419215c31e348c5e840fe6e2bc53ab5/results/dangvantuan/sentence-camembert-base/BSARDRetrieval.json#L37)
It is due to a bug that I am fixing here.

- [x] I have run the '' on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command. 

Additionally to the bug fix:
- [x] Changing score to recall@100


mteb/tasks/Retrieval/fra/BSARDRetrieval.py
```diff
@@ -52,7 +52,7 @@ def load_data(self, **kwargs):
 
         self.queries = {
             self.metadata.eval_splits[0]: {
-                str(q["id"]): " ".join((q["question"] + q["extra_description"]))
+                str(q["id"]): q["question"] + q["extra_description"]
```
imenelydiaker, line 55: What's the difference between these 2 implementations?

wissam-sib: exactly, we were ending up with queries that looked like this: "I  a m a  q u e r y"

imenelydiaker (approved):
LGMT! Thanks for fixing this bug, it wasn't easy to catch, well done! 🚀 
You can add points, 4 points for bug fix seem good to me (wasn't easy to catch), and 2 for review.

imenelydiaker (approved):

PR #624, opened 2024-05-03
Hierarchical clustering

x-tabdeveloping:
Added a naive implementation of hierarchical clustering along with a hierarchical formulation of SNL and results.
If you have any thoughts on how to improve this feel free to chime in.


mteb/abstasks/AbsTaskHierarchicalClustering.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        pred_labels = clustering_model.fit_predict(level_embeddings)
+        # Only evaluate where the level is valid
+        score = v_measure_score(level_labels, pred_labels)
+        v_scores.append(score)
+    return v_scores
+
+
+class AbsTaskHierarchicalClustering(AbsTask):
+    """Abstract class for Hiearchical Clustering tasks."""
```
KennethEnevoldsen, line 56: please make a description of the expected format

results/intfloat__multilingual-e5-small/SNLHierarchicalClustering.json
```diff
@@ -0,0 +1,16 @@
...
+    "evaluation_time": 337.37,
+    "v_measure": 0.6756280728585395,
+    "v_measures": [
+      1.0,
+      0.4103493451179973,
+      0.6244018029078756,
+      0.6676460260153364,
+      0.6757431902514882
+    ]
+  }
```
KennethEnevoldsen, line 15: I would redo the results such that you get multiple scores pr. level (using a boostrap similar to fastclustering) and then make it clear which belongs to which level. So make it 

```
"v_measures": {
       "level 1": [1.0, ...],
       "level 2": [0.4103493451179973, ...], ... 
     }

File: mteb/tasks/HierarchicalClustering/nob/SNLHierarchicalClustering.py [hunk 3]
```diff
@@ -0,0 +1,49 @@
...
+from mteb.abstasks.AbsTaskHierarchicalClustering import AbsTaskHierarchicalClustering
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+
+def split_labels(record: dict) -> dict:
+    record["labels"] = record["labels"].split(",")
+    return record
+
+
+class SNLHierarchicalClustering(AbsTaskHierarchicalClustering):
```
KennethEnevoldsen, line 12: I would add the superseeded_by to the old dataset

mteb/tasks/HierarchicalClustering/nob/SNLHierarchicalClustering.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import annotations
+
+from mteb.abstasks.AbsTaskHierarchicalClustering import AbsTaskHierarchicalClustering
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+
+def split_labels(record: dict) -> dict:
+    record["labels"] = record["labels"].split(",")
```
KennethEnevoldsen, line 8: Seems like the first layer is trivial so would remove that level

results/intfloat__multilingual-e5-small/SNLHierarchicalClustering.json
KennethEnevoldsen: Maybe better to allow for the categories to be named?

mteb/abstasks/AbsTaskHierarchicalClustering.py
```diff
@@ -0,0 +1,117 @@
...
+        for label in labels:
+            if len(label) > i_level:
+                level_labels.append(label[i_level])
+            else:
+                level_labels.append(-1)
+        level_labels = np.array(level_labels)
+        valid_idx = level_labels != -1
+        level_labels = level_labels[valid_idx]
+        level_embeddings = embeddings[valid_idx]
+        clustering_model = MiniBatchKMeans(
```
KennethEnevoldsen, line 42: So we essentially just evaluate it each level? I think this is fine, but it might be possible to integrate this into FastClustering (allow for either 1 or multiple labels). This way we only have one implementation

KennethEnevoldsen (commented):
Looks good. We might consider combining it with fast clustering, but at least we should consider make the level apparent in the results.

mteb/abstasks/AbsTaskHierarchicalClustering.py
x-tabdeveloping: I will look into it!

results/intfloat__multilingual-e5-small/SNLHierarchicalClustering.json
x-tabdeveloping: What do you mean by categories?

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -22,38 +24,51 @@
...
-    for _ in range(n_clusters):
-        # sample N samples from the corpus with replacement
-        cluster_indices = rng_state.choices(range(n_embeddings), k=cluster_size)
+    v_measures = defaultdict(list)
+
+    max_depth = min(max_depth, max(map(len, labels)))
+    # Evaluate on each level til max depth
+    for i_level in range(max_depth):
+        level_labels = []
+        # Assign -1 to gold label if the level is not there
```
KennethEnevoldsen, line 47: How is that handles in the KMeans?

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -133,17 +149,25 @@ def _evaluate_monolingual(
...
+                label = [label]
+            labels.append(label)
 
         v_measures = evaluate_clustering_bootstrapped(
             embeddings,
             downsampled_dataset["labels"],
             n_clusters=self.n_clusters,
             cluster_size=self.max_documents_per_cluster,
             kmean_batch_size=self.k_mean_batch_size,
+            max_depth=self.max_depth,
```
KennethEnevoldsen, line 164: max-depth in the categories list I suppose? I would probably move that to the creation of the list for the specific dataset. Or is there good reason to now have it?

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -22,38 +24,51 @@
...
-    )
-
-    for _ in range(n_clusters):
-        # sample N samples from the corpus with replacement
-        cluster_indices = rng_state.choices(range(n_embeddings), k=cluster_size)
+    v_measures = defaultdict(list)
+
+    max_depth = min(max_depth, max(map(len, labels)))
+    # Evaluate on each level til max depth
+    for i_level in range(max_depth):
```
KennethEnevoldsen, line 45: This seems like it can be performed as a matrix construction from lists

```pyhton
def ragged_list_transpose(input_list: list[list[int]], default: int=-1):
    if not input_list:
        return np.array([])
    
    max_length = max(len(sublist) for sublist in input_list)
    
    arr = np.full((max_length, len(input_list)), default) # -1 array
    
    # Fill the array with actual values from the input list
    for idx, sublist in enumerate(input_list):
        if sublist:  # Check if the sublist is not empty
            arr[:len(sublist), idx] = sublist
    
    return arr.T
```

results/intfloat__multilingual-e5-small/SNLHierarchicalClustering.json
```diff
@@ -0,0 +1,71 @@
+{
+  "dataset_revision": "3d3d27aa7af8941408cefc3991ada5d12a4273d1",
+  "mteb_dataset_name": "SNLHierarchicalClustering",
+  "mteb_version": "1.6.10",
+  "test": {
+    "evaluation_time": 366.1,
+    "v_measure": 0.5648936348151914,
+    "v_measures": {
+      "Level 0": [
```
KennethEnevoldsen, line 9: very sexy, I think this is much better

mteb/abstasks/TaskMetadata.py
```diff
@@ -79,6 +73,7 @@
     "BitextMining",
     "Classification",
     "Clustering",
+    "HierarchicalClustering",
```
KennethEnevoldsen, line 76:
```suggestion
```

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen, line 1: We need to update the docstring to match, but otherwise I believe this looks very solid.

KennethEnevoldsen (commented):
This looks really a few suggestions, but I believe we are getting there.

mteb/abstasks/AbsTaskClusteringFast.py
x-tabdeveloping: Why would this advantageous? I think the implementation I wrote is quite good at communicating intent, and this seems a bit opaque to me. Are there any significant reasons why we should prefer this?

mteb/abstasks/AbsTaskClusteringFast.py
x-tabdeveloping: ? pardon ?

mteb/abstasks/AbsTaskClusteringFast.py
x-tabdeveloping: I mean could have it either way, but this limits it at a reasonable level for all datasets. Also specific tasks wouldn't have to write `dataset_transform()` methods to limit the number of levels, but could simply specify `max_depth`.

```python
class SomeSpecificTaskClustering(AbsTaskClusteringFast):
     max_depth = 3
```

We can also make the default `None` and include all levels, but I still believe it's a helpful attribute to have.

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: after considering it I think you are right. The current approach is more readable, if speed is an issue however this does make it faster, but I believe we care more about readability here.

x-tabdeveloping: Also questionable how much faster it would get.

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: I see that you subset the embeddings. No problems here

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: 0.011833577000004425 vs 0.00961239500000488 so it really isn't much. (using a small sample so it might scale differently)

KennethEnevoldsen (commented):
Everything looks good. 

Will you add points and rerun one of the older Fast Clustering tasks to ensure that the scores doesn't change?

Might be ideal to rename the AbsTaskClusteringFast to AbsTaskClustering and AbsTaskClustering to AbsTaskClusteringSlow (to make the Fast version seem like the default) - this is a seperate PR though

x-tabdeveloping:
Sure thing

KennethEnevoldsen:
I would do: 2 for a new dataset, 2 for new task, 2-4 for fixes (integration of the two tasks).

x-tabdeveloping:
Is this ready for a merge? @KennethEnevoldsen

mteb/tasks/Clustering/nob/SNLHierarchicalClustering.py
```diff
@@ -0,0 +1,52 @@
...
+        dialect=[],
+        task_subtypes=["Thematic clustering"],
+        text_creation="found",
+        bibtex_citation="""@mastersthesis{navjord2023beyond,
+  title={Beyond extractive: advancing abstractive automatic text summarization in Norwegian with transformers},
+  author={Navjord, J{\o}rgen Johnsen and Korsvik, Jon-Mikkel Ryen},
+  year={2023},
+  school={Norwegian University of Life Sciences, {\AA}s}
+}""",
+        n_samples={"test": 2048},
```
KennethEnevoldsen, line 43: is this correct?

KennethEnevoldsen (approved):
Everything looks good here, though I think the sample size might be wrong?

x-tabdeveloping:
Wait a sec, I'm adding a P2P (article) and S2S (ingress) task real quick as with many other clustering datasets.

PR #636, opened 2024-05-05
Add Belebele Retrieval

jupyterjazz:

Belebele is a multiple-choice machine reading comprehension (MRC) dataset spanning 115 distinct languages


mteb/tasks/Retrieval/multilingual/BelebeleRetrieval.py
```diff
@@ -0,0 +1,213 @@
...
+        domains=["Web", "News"],
+        text_creation="created",
+        n_samples={_EVAL_SPLIT: 900},
+        date=("2023-08-31", "2023-08-31"),
+        form=["written"],
+        task_subtypes=["Question answering"],
+        socioeconomic_status="mixed",
+        annotations_creators="expert-annotated",
+        dialect=[],
+        avg_character_length={_EVAL_SPLIT: 329},
```
imenelydiaker, line 157: Can you please sum the number of samples of the test set of each language?

mteb/tasks/Retrieval/multilingual/BelebeleRetrieval.py
```diff
@@ -0,0 +1,213 @@
...
+            belebele_lang = _LANGS[lang][0].replace("-", "_")
+            ds = self.dataset[belebele_lang]
+
+            question_ids = {
+                question: _id for _id, question in enumerate(set(ds["question"]))
+            }
+            context_ids = {
+                passage: _id for _id, passage in enumerate(set(ds["flores_passage"]))
+            }
+            answers = [
```
imenelydiaker, line 186: I'm not sure we need the answers, the corpus containing all the text should be enough for the retrieval task, as the goal is to retrieve the right docs for a question.

imenelydiaker: [GermanQuAD](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/deu/GermanQuADRetrieval.py) and [FQuAD](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/fra/FQuADRetrieval.py) don't use answers, we'll check again these 2 tasks. I think you can remove answers yes.

mteb/tasks/Retrieval/multilingual/BelebeleRetrieval.py
```diff
@@ -0,0 +1,201 @@
...
+        domains=["Web", "News"],
+        text_creation="created",
+        n_samples={_EVAL_SPLIT: 103500},  # number of languages * 900
+        date=("2023-08-31", "2023-08-31"),
+        form=["written"],
+        task_subtypes=["Question answering"],
+        socioeconomic_status="mixed",
+        annotations_creators="expert-annotated",
+        dialect=[],
+        avg_character_length={_EVAL_SPLIT: 329},
```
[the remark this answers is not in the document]
jupyterjazz: I'm sorry I don't fully understand do you want me to add number of samples and avg length for each language (like `avg_character_length={lang1: avg1, lang2: avg2..}` and`n_samples= {lang1: 900, lang2: 900...}`) or apply suggested changes?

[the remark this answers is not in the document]
jupyterjazz: ok, but I think `n_samples` is already like that, right?

I'll recalculate avg_lengths because we removed answers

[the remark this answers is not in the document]
jupyterjazz: `avg_character_length` is 329, `n_samples` I have specified above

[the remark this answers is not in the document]
imenelydiaker: ah okay,  the order of the metadata changed so I didn't see it sorry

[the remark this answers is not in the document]
imenelydiaker: All good then! You can add points and we'll be ready to merge 🙂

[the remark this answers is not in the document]
jupyterjazz: from what I counted, 81 out of the 115 languages are new to the retrieval task, so I'm adding 2 + 81*4 points, please correct me if I'm wrong

docs/mmteb/points/636.jsonl
```diff
@@ -0,0 +1,2 @@
+{"GitHub": "jupyterjazz", "New dataset": 326}
```
imenelydiaker, line 1: It's 2 points for a New dataset even with multiple languages. You may add 4 points per new language if it's not already handled by another task.

jupyterjazz: How many points do you suggest I should add? 81 out of the 115 Belebele languages are not included in other retrieval datasets, that's why I added 2 + 81*4, I think this is what other PRs did as well

Andrian0s:
I would agree we are ready to merge when we fix my minor comments.

docs/mmteb/points/636.jsonl
imenelydiaker: @KennethEnevoldsen we need some help here with the points

jupyterjazz:
@Andrian0s I think your comments were not posted, they're not visible on my side

Andrian0s:
> @Andrian0s I think your comments were not posted, they're not visible on my side

ah, thank you, should be visible now

mteb/tasks/Retrieval/multilingual/BelebeleRetrieval.py
```diff
@@ -0,0 +1,201 @@
...
+        domains=["Web", "News"],
+        text_creation="created",
+        n_samples={_EVAL_SPLIT: 103500},  # number of languages * 900
+        date=("2023-08-31", "2023-08-31"),
+        form=["written"],
+        task_subtypes=["Question answering"],
+        socioeconomic_status="mixed",
+        annotations_creators="expert-annotated",
+        dialect=[],
+        avg_character_length={_EVAL_SPLIT: 568},
```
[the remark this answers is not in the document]
jupyterjazz: It's an average length of query+passage pairs
https://github.com/embeddings-benchmark/mteb/blob/0cf33d73b1f3ff5be1d3689f2aa8abbbe4454c99/mteb/abstasks/TaskMetadata.py#L167

mteb/tasks/Retrieval/multilingual/BelebeleRetrieval.py
```diff
@@ -0,0 +1,201 @@
...
+        },
+        description=(
+            "Belebele is a multiple-choice machine reading comprehension (MRC) dataset spanning 115 distinct languages."
+        ),
+        type="Retrieval",
+        category="s2p",
+        eval_splits=[_EVAL_SPLIT],
+        eval_langs=_LANGS,
+        reference="https://arxiv.org/abs/2308.16884",
+        main_score="ndcg_at_20",
```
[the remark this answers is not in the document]
jupyterjazz: Yes that would make sense if we wanted to get a correct answer across the 4 possible texts given a passage and a question. However, here we treat the problem as a retrieval of passages based on questions. I put 20 initially because I had also included answer retrieval (https://github.com/embeddings-benchmark/mteb/pull/636#discussion_r1590807508), but now that I changed it maybe it makes more sense to have ndcg_at_10, wdyt?

jupyterjazz:
> As far as I know, this dataset spawns from FLORES, which is already covered from datasets likes SIB. Not a retrieval dataset, but still covers the languages

I'm sure the languages are covered if you consider all tasks/datasets on mteb, but from what I understand you get a bonus if the language is new for its task type, in this case for retrieval

mteb/tasks/Retrieval/multilingual/BelebeleRetrieval.py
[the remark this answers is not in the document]
Andrian0s: Why would we treat the task of this dataset differently than it is used and introduced in the paper. 


Unless there is a series of evaluations using this dataset otherwise, you should only evaluate the dataset in the same way as the introducing paper and the evaluation method of it. MCQA would work perfectly fine within the AbsRetrieval task.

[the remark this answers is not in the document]
jupyterjazz: I'm not sure if adapting a dataset to a slightly different task is such a terrible thing 😅

Many reading comprehension datasets are treated that way in the retrieval category:
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/fra/FQuADRetrieval.py
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/deu/GermanQuADRetrieval.py

However, if you think it can be easily implemented within the `AbsRetrieval` class, please suggest how and I'll be glad to do it

docs/mmteb/points/636.jsonl
KennethEnevoldsen: Ahh this actually already happened back in #583 and we have had similar cases before (though never on this scale). We might consider changing the new dataset scores to be a max of 50-100. Both in terms on value added to the benchmark and effort I believe this is reasonable. We will have to update previous scores though.

@jupyterjazz how would you feel about such a change? We know that the points system is imperfect (always will be) and we don't want to discourage contributors who have put many hours into review etc. - I hope that the solution can strike the right balance.

Just checking in with @isaac-chung (reviewer, and large contributor) and @davidstap (previous contributor of large dataset) here as well.

jupyterjazz: Hey @KennethEnevoldsen, thanks for the response. I'm okay with capping the max score of bonuses, I think it's fair. 

Can you please also help us out in this discussion?
https://github.com/embeddings-benchmark/mteb/pull/636#discussion_r1596046376

mteb/tasks/Retrieval/multilingual/BelebeleRetrieval.py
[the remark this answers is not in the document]
imenelydiaker: Yeah we already included *QuAD datasets for retrieval and even summarization ones. Imo it's okay to use it as a retrieval dataset as well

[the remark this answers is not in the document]
Andrian0s: I don't think it's a bad thing to use a dataset for a slightly different task, but why not the original one? PR #645 should be a good starting point.

Query: the passage (a bit unintuitive)
Corpus: the four questions - positive one marked as positive

Then, essentially you want to see if the highest cosine similarity of the answers, and if it's the right one, the model predicted "accurately". 

On MTEB terms, all you need to do is:
metric: recall_at_1 (or precision at 1 but use recall)

The embedding models might perform worse than the LLM's as described in the original paper but I would expect some performance. Please double check this in the results

@imenelydiaker just double checking that it's okay that I request the evaluation to be according to the original of. the paper and not a similar task. My intuition is that if I read the paper and hear it's in the benchmark I want to see the same kind of evaluation to see how the field evolved.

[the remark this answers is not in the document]
imenelydiaker: So if I het it right, the change you're suggesting is to change the main score to recall_at_1 instead of ndcg_at_20? In this case, then okay, all metrics will be computed anyway.

Else, the retrieval evaluator is the same for all tasks so I'm not sure we're changing it for this task only.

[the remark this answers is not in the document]
jupyterjazz: > Query: the passage (a bit unintuitive)
> Corpus: the four questions - positive one marked as positive

I think you meant "the four answers".
Can you clarify how the question itself fits into this? Seems like we only have passages and answers with this setup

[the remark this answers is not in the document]
Andrian0s: > > Query: the passage (a bit unintuitive)
> > Corpus: the four questions - positive one marked as positive
> 
> I think you meant "the four answers". Can you clarify how the question itself fits into this? Seems like we only have passages and answers with this setup

Yes, Four answers. I guess prepending the question to the answers might be sufficient.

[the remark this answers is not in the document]
Andrian0s: Anyway, if within the benchmark there are more QuAD used as a pure retrieval task, feel free to keep as is. I am removing my request to review changes, I think it was worth the consideration anyway! Thank you both

Andrian0s (approved):

mteb/tasks/Retrieval/multilingual/BelebeleRetrieval.py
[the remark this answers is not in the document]
KennethEnevoldsen: I believe atm. even if we were to use recall_at_1 it would not be eq. to the accuracy in the paper. If that is the case I would prefer to use ndcg_at_10 both to avoid unintended comparisons and to keep it in line with MTEB.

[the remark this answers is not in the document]
jupyterjazz: Thanks for the insightful review, @Andrian0s!

I understand your point about aligning our evaluation with the one in the paper. However, I didn't think our setup allowed that, which led me to adapt it. Tbh prepending questions to the answers doesn't seem convincing to me, and also even if we used `recall_at_1`, I think it would apply across the entire corpus, not just the four candidate answers as in the paper, so the comparison wouldn’t hold. MCQA could be a cool new task to implement though

docs/mmteb/points/636.jsonl
isaac-chung: No issues with points cap for a single dataset 👍

imenelydiaker: 50 points seem fair, we'll open an issue to fix points of other PRs. @KennethEnevoldsen @isaac-chung do you agree?

isaac-chung: 👍

davidstap: Saw the tag just now, thanks @KennethEnevoldsen. Not sure how I feel about changing the point system for previous submissions. I get that the change might better reflect effort that went into contributions. Changing the rules now (for future contributions) seems fair, since people can determine whether they participate given these new rules. Changing rules and applying new rules retroactively to previously earned points seems unfair to me.

imenelydiaker: @KennethEnevoldsen @isaac-chung I proposed here a new rule for the pointing system as discussed above: https://github.com/embeddings-benchmark/mteb/pull/666. 

@davidstap Thank you for bringing that up. This change is by no means to diminish any contributions that are already made. We already warned contributors that the system may change in the following line: 
[> **Note**: The purpose of the point system is not to barrier collaboration, but to reward contributions. We might adjust the point requirement lower to accommodate more co-authorship if needed.](https://github.com/embeddings-benchmark/mteb/blob/376568330d79367f1330c3faf7ba3f6533cbd41f/docs/mmteb/readme.md?plain=1#L46)

Contributors will still be recognized as co-authors if they meet the criteria in the points guide.

KennethEnevoldsen: Thanks for the response @davidstap. The specific reason why I would like to reduce previous contributions as one dataset has obtained >3000 points for a single dataset contribution. We should have taken the discussion at the time, but there are many PRs and I understand that it wasn't caught.

Keeping the score as it is suggests to the contributors that the bibleNLP dataset is worth more than ALL the remaining contributions combined. I don't believe that is fair to the other contributors of this project so it is clearly something that we will have to change.

Again this should have been caught in review and I am sad that it wasn't. I hope we can find a solution that fits everyone.

isaac-chung: I'd like to offer my sincere apologies for missing this and not raising this earlier during the original review for the BibleNLP dataset 🙏

davidstap: > @davidstap Thank you for bringing that up. This change is by no means to diminish any contributions that are already made. We already warned contributors that the system may change in the following line:

@imenelydiaker Thanks for the pointer. I read that statement as: we might decide that a lower point _requirement_ (i.e., less than 10 points) merits authorship, not that the point _system_ might change.

davidstap: > Thanks for the response @davidstap. The specific reason why I would like to reduce previous contributions as one dataset has obtained >3000 points for a single dataset contribution. We should have taken the discussion at the time, but there are many PRs and I understand that it wasn't caught.
> 
> Keeping the score as it is suggests to the contributors that the bibleNLP dataset is worth more than ALL the remaining contributions combined. I don't believe that is fair to the other contributors of this project so it is clearly something that we will have to change.
> 
> Again this should have been caught in review and I am sad that it wasn't. I hope we can find a solution that fits everyone.

I get your concern about one dataset contribution racking up a crazy number of points. It seems unfair when that dataset is worth more than all other contributions combined.

OTOH, I think _massively multilingual_ datasets such as BibleNLP, but also FLORES and NTREX, are _extremely_ valuable for a _multilingual_ benchmark. They cover a ton of languages while keeping the domain and other confounders constant, which is crucial when one wants to test how well different languages perform in the same setting. When you're dealing with languages from different datasets, it's hard to compare performance because there are so many other confounding factors at play, like the topic/domain or length of the samples.


That’s why I think it’s much more valuable to have a single dataset with 500 languages, instead of 10 datasets with 50 distinct languages each, even if the set of languages is the same. The new point system (cap at 50/100 points) would not capture this at all. Of course no point system is perfect, but IMO this change would not be an improvement.

imenelydiaker: > OTOH, I think _massively multilingual_ datasets such as BibleNLP, but also FLORES and NTREX, are _extremely_ valuable for a _multilingual_ benchmark. They cover a ton of languages while keeping the domain and other confounders constant, which is crucial when one wants to test how well different languages perform in the same setting. When you're dealing with languages from different datasets, it's hard to compare performance because there are so many other confounding factors at play, like the topic/domain or length of the samples.

BibleNLP, Flores and NTREX can only be used in one task for a multilingual benchmark, although it's be nice to remind that the key idea behind MTEB is the ability to evaluate a model accross different tasks and languages. These datasets are cross-lingual, most of them are parallel texts that were built using translation, so basically they contain the same information across all languages. 

Having distinct datasets with different languages is also valuable since we're not only focused on one style of writing or one source of data. It would clearly be unfair to reward a multilingual dataset more than other contributions like adding a new emmbedding evaluation task. 

This new pointing system will not diminish your contribution, you'll still get 100 points for your two datasets which will get you in the top list of co-authors which is already good. It will just make it fair towards other contributors that have also put efforts into this.

Again sorry for the confusion, we should have fixed this earlier. But we're humans and we make mistakes 🙂

KennethEnevoldsen: > That’s why I think it’s much more valuable to have a single dataset with 500 languages, instead of 10 datasets with 50 distinct languages each, even if the set of languages is the same. The new point system (cap at 50/100 points) would not capture this at all. Of course no point system is perfect, but IMO this change would not be an improvement.

of course, diversity of languages is not the only target of measure. Diversity of tasks is also important. E.g. A model performing well on bible datasets might perform well based on heuristics (e.g. does Jesus appear at the beginning of the sentence), but 50x datasets in 50 different languages cover different domains, tasks and sources and thus I would believe measures a broader spectrum (when we do a factor analysis of task x language scores I believe we will see the same thing as well).

With all of this being said. I would suggest that we set a limit to 50 points for the new datasets but raise the point limit to 100 points for already accepted PRs as well as ongoing PRs. I believe this strikes a fair balance between stating that multilingual datasets are extremely valuable while not overstating their value.

KennethEnevoldsen:
@jupyterjazz, sorry for taking over this PR with a discussion on points. I believe you can merge this as long as it is acceptable to you that we might reduce the point at a later stage (following the discussion above).

docs/mmteb/points/636.jsonl
davidstap: @imenelydiaker @KennethEnevoldsen thanks for your additional comments. It seems I have overstated the importance of extremely multilingual datasets in the context of mmteb. I think this stems from my background in multilingual MT, where these kinds of datasets are very valuable. Thanks again for organizing this community effort ☺️

jupyterjazz:
@KennethEnevoldsen, thanks for the review and the discussion. I changed my score to 100 based on this comment https://github.com/embeddings-benchmark/mteb/pull/636#discussion_r1597467581

If everything looks good, please feel free to merge the PR as I'm not allowed to do it myself. Thanks again!

imenelydiaker:
> @KennethEnevoldsen, thanks for the review and the discussion. I changed my score to 100 based on this comment https://github.com/embeddings-benchmark/mteb/pull/636#discussion_r1597467581
> 
> If everything looks good, please feel free to merge the PR as I'm not allowed to do it myself. Thanks again!

It should be 50 according to this rule: https://github.com/embeddings-benchmark/mteb/pull/666
@KennethEnevoldsen  @isaac-chung

KennethEnevoldsen:
@imenelydiaker let us keep it at 100 for existing contributions (as they started their contributions before the above discussion).

docs/mmteb/points/636.jsonl
KennethEnevoldsen: Thanks for understanding @davidstap. Again we very much appreciate your contributions and this is by no means meant to understate it.

imenelydiaker (approved):
Everything looks good, let's merge! Thanks for your contribution!

Issue #198, opened 2024-01-09
Add MIRACL


Muennighoff:

imenelydiaker:
I may be able do this, is anyone working on it? @Muennighoff @KennethEnevoldsen ?

Should I wait until this PR https://github.com/embeddings-benchmark/mteb/pull/233 is merged ? I think the PR addresses BEIR specifically, so maybe no need to wait until it's merged.

Queries and qrels: https://huggingface.co/datasets/miracl/miracl
Corpus: https://huggingface.co/datasets/miracl/miracl-corpus

Muennighoff:
Parts of MIRACL are already in MTEB I think - Ideally we unify them so that it's just one MIRACL task from which the different datasets can be selected from

imenelydiaker:
> Parts of MIRACL are already in MTEB I think - Ideally we unify them so that it's just one MIRACL task from which the different datasets can be selected from

Yep Jina team added some of it (de and es) here: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/MIRACLRetrieval.py
And the Korean version is here: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Retrieval/KoMiracl.py

The original dataset is here, with all these languages included:  
- Queries and qrels: https://huggingface.co/datasets/miracl/miracl
- Corpus: https://huggingface.co/datasets/miracl/miracl-corpus

Maybe we can just use the original dataset with all provided languages ? 
Do you prefer duplicating the data into a MTEB data repository to make sure it will always be available for the benchmark, or should we just load the dataset from it's original HF repo and transform it on load ?

KennethEnevoldsen:
> Do you prefer duplicating the data into a MTEB data repository to make sure it will always be available for the benchmark, or should we just load the dataset from it's original HF repo and transform it on load ?

Generally, the first option is more robust, but we have multiple datasets atm. which does the second one. The second option is also easier to update (if the dataset is updated). 

For MIRACL though I would just go for option 1 (permissive license) as long there are no planned updates.

thakur-nandan:
Hi @KennethEnevoldsen @imenelydiaker @Muennighoff, 

@crystina-z and I are co-authors of the MIRACL benchmark.

We just saw the announcement on [Twitter/X](https://twitter.com/Muennighoff/status/1778433270843121759) of building out the multilingual MTEB, and I saw this issue is open.

[...]

PR #641, opened 2024-05-06
MIRACL reranking

crystina-z:
Adding MIRACL Reranking as discussed in #198 

Reason for dataset addition:

- [X] I have tested that the dataset runs with the `mteb` package.
- [X] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [X] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [X] `intfloat/multilingual-e5-small`
- [X] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
  (See scores below)
- [X] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
  NOTE: `make test` didn't work for me with error (pytest: error: unrecognized arguments: -n), but `pytest --durations=5` (taking off `-n auto`) passes

----------
Thanks for your patience! We finally have the MIRACL-reranking part ready. Here's the detailed scores "reranking" BM25 top-100:
(BM25, mDPR and Hybrid are the baselines we had before)

  | multilingual-e5-small | paraphrase-multilingual-MiniLM-L12-v2 | BM25 | mDPR | Hybrid
-- | -- | -- | -- | -- | --
ar | 0.715 | 0.413 | 0.481 | 0.499 | 0.525
bn | 0.690 | 0.080 | 0.508 | 0.443 | 0.501
de | 0.456 | 0.361 | 0.226 | 0.490 | 0.408
en | 0.531 | 0.479 | 0.351 | 0.394 | 0.364
es | 0.576 | 0.459 | 0.319 | 0.478 | 0.418
fa | 0.528 | 0.307 | 0.333 | 0.480 | 0.215
fi | 0.740 | 0.541 | 0.551 | 0.472 | 0.602
fr | 0.459 | 0.357 | 0.183 | 0.435 | 0.314
hi | 0.582 | 0.359 | 0.458 | 0.383 | 0.286
id | 0.536 | 0.415 | 0.449 | 0.272 | 0.392
ja | 0.596 | 0.338 | 0.369 | 0.439 | 0.424
ko | 0.766 | 0.427 | 0.419 | 0.419 | 0.483
ru | 0.571 | 0.399 | 0.334 | 0.407 | 0.391
sw | 0.603 | 0.241 | 0.383 | 0.299 | 0.560
te | 0.759 | 0.144 | 0.494 | 0.356 | 0.528
th | 0.705 | 0.445 | 0.484 | 0.358 | 0.517
yo | 0.541 | 0.408 | 0.406 | 0.396 | 0.415
zh | 0.434 | 0.358 | 0.180 | 0.512 | 0.410
avg | 0.599 | 0.363 | 0.385 | 0.418 | 0.431

I noticed that current reranking filters out the queries that have no positive documents in the top-k candidates, which seems to inflate the scores and makes reranking results based on different first-stage systems incomparable. Also the evaluation is based on sklearn, which IIRC yield different ndcg scores from pytrec. Considering these I added another Evaluator just for MIRACL, tho I believe it should be compatible with other Rerank Tasks. Let me know what you think of the new evaluation!


mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
         # ap = np.mean([np.mean(preds[: k + 1]) for k in range(len(preds)) if preds[k]])
         ap = average_precision_score(is_relevant, pred_scores)
         return ap
+
+
+class MIRACLRerankingEvaluator(RerankingEvaluator):
```
KennethEnevoldsen, line 258: @orionw and @Muennighoff if you guys have the time to look at this I would love a second look.

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
...
+            name,
+            similarity_fct,
+            batch_size,
+            use_batched_encoding,
+            limit,
+            **kwargs,
+        )
+        self.k_values = k_values
+
+    def rerank(self, query_emb, docs_emb):
```
KennethEnevoldsen, line 283: Remove the type hint from the docstring to the arguments

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
...
+        samples,
+        mrr_at_k: int = 10,
+        name: str = "",
+        similarity_fct=cos_sim,
+        batch_size: int = 512,
+        use_batched_encoding: bool = True,
+        limit: int = None,
+        k_values: List[int] = [1, 3, 5, 10, 20, 100, 1000],
+        **kwargs,
+    ):
```
KennethEnevoldsen, line 270: A docstring here explaining the difference would be nice for the future

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
...
+class MIRACLRerankingEvaluator(RerankingEvaluator):
+    def __init__(
+        self,
+        samples,
+        mrr_at_k: int = 10,
+        name: str = "",
+        similarity_fct=cos_sim,
+        batch_size: int = 512,
+        use_batched_encoding: bool = True,
+        limit: int = None,
```
KennethEnevoldsen, line 267:
```suggestion
        limit: int | None = None,
```

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
...
+        k_values: List[int] = [1, 3, 5, 10, 20, 100, 1000],
```
KennethEnevoldsen, line 268:
```suggestion
        k_values: list[int] = [1, 3, 5, 10, 20, 100, 1000],
```
as we have: `from __future__ import annotations`

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
         # ap = np.mean([np.mean(preds[: k + 1]) for k in range(len(preds)) if preds[k]])
         ap = average_precision_score(is_relevant, pred_scores)
         return ap
+
+
+class MIRACLRerankingEvaluator(RerankingEvaluator):
+    def __init__(
+        self,
+        samples,
```
KennethEnevoldsen, line 261: A type hint here would be great

mteb/evaluation/evaluators/RerankingEvaluator.py
KennethEnevoldsen: It might be worth considering only keeping one evaluator (while being aware that this change it breaking, the switch from MTEB --> MMTEB might be the right time to do it)

KennethEnevoldsen: (If we choose to replace it I have a few more comments here to make the evaluator easier to maintain in the future)

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
...
+
+        pred_scores = self.similarity_fct(query_emb, docs_emb)
+        if len(pred_scores.shape) > 1:
+            pred_scores = torch.amax(pred_scores, dim=0)
+
+        return {
+            str(i): score.detach().numpy().item() for i, score in enumerate(pred_scores)
+        }
+
+    def compute_metrics_batched(self, model):
```
KennethEnevoldsen, line 308:
```suggestion
    def compute_metrics_batched(self, model: Encoder | EncoderWithQueryCorpusEncode)
```

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
...
+        all documents together
+        """
+        # using encode_queries and encode_corpus functions if they exists,
+        # which can be defined by users to add different instructions for query and passage conveniently
+        encode_queries_func = (
+            model.encode_queries if hasattr(model, "encode_queries") else model.encode
+        )
+        encode_corpus_func = (
+            model.encode_corpus if hasattr(model, "encode_corpus") else model.encode
+        )
```
KennethEnevoldsen, line 319: Can be replaced with `isinstance(model, EncoderWithQueryCorpusEncode)`

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
```diff
@@ -4,35 +4,88 @@
...
+        domains=["Encyclopaedic"],
+        task_subtypes=[],
+        license="CC BY-SA 4.0",
+        socioeconomic_status="low",
+        annotations_creators="expert-annotated",
+        dialect=[],
+        text_creation="created",
+        bibtex_citation=_CITATION,
         n_samples=None,
         avg_character_length=None,
```
KennethEnevoldsen, line 72: Let is add these two as well

docs/mmteb/points/641.jsonl
```diff
@@ -0,0 +1 @@
+{"GitHub": "crystina-z", "New dataset": 2}
```
KennethEnevoldsen, line 1: I would remove the 2 points from adding a dataset and instead add 4 points to (bug)-fixes

Feel free to add a point for the dataset annotation as well

There is also a bonus for tel, ind, saw, yor

so that makes 4 + 1 + 4x4 = 21 points

Please also add the points for the reviewers.

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -251,3 +253,165 @@ def ap_score(is_relevant, pred_scores):
...
+            query = instance["query"]
+            positive = set(instance["positive"])
+            docs = list(instance["candidates"])
+
+            if isinstance(query, str):
+                # .encoding interface requires List[str] as input
+                query = [query]
+            query_emb = np.asarray(
+                encode_queries_func(query, batch_size=self.batch_size)
+            )
```
KennethEnevoldsen, line 405:
```suggestion
            query_emb = np.asarray(
                encode_queries_func([query], batch_size=self.batch_size)
            )
```

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
```diff
@@ -4,35 +4,88 @@
...
-        bibtex_citation=None,
+        eval_splits=[_EVAL_SPLIT],
+        eval_langs=_LANGUAGES,
+        main_score="NDCG@10",
+        date=("2022-06-01", "2023-01-30"),
+        form=["written"],
+        domains=["Encyclopaedic"],
+        task_subtypes=[],
+        license="CC BY-SA 4.0",
+        socioeconomic_status="low",
```
KennethEnevoldsen, line 66:
```suggestion
        socioeconomic_status="mixed",
```
as the documents are from wikipedia

KennethEnevoldsen (commented):
This addition looks really good! I have asked a second pair of eyes on the evaluation, but otherwise I only see a few minor issues.

mteb/evaluation/evaluators/RerankingEvaluator.py
orionw: Looking at the changes, it seems like the reason for a new evaluator class is the `candidates` key instead of `negatives`?  I agree that it would be nice to keep only one `RerankingEvaluator` if it's possible to massage the data into the same format. 

I could be missing something though, @crystina-z is there something I'm missing about the MiRACL data that makes the normal evaluator not work for it? Or would it be possible to rename the keys so that it could fit the same format? 

Thanks for the PR, very excited to see MiRACL added!

crystina-z: heyy! I won't say `candidates` is a replacement for `negatives`, so previously reranking evaluation always includes all the `positives`, whereas it might not be true for reranking: The top-k `candidates` might contain all or partial or even zero positive passages. That's my main motivation to add a separate evaluator to rerank precisely the top-k from the previous stage. But lmk how you think!

Muennighoff: Makes sense, so `candidates` in your evaluator is `positives` + `negatives` in the original evaluator i.e. more or less the same, no? I also think it'd be great to unify the two.

Another small difference I think are the metrics you are calculating (ndcg etc.) but I think it'd be fine to add them to the original evaluator and calculate more metrics there for all reranking evals in the future if those metrics are important to you.

crystina-z: yup `candidates` functions the same with `positives` + `negatives`, tho it doesn't necessarily include all `positives`

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
crystina-z: ah I thought about to ask but forgot. What's the definition for `n_samples` and `avg_character_length` here? Does `n_samples` mean the number of queries or all query-document pairs? and is `avg_character_length` the average char length of all the queries and documents?

mteb/evaluation/evaluators/RerankingEvaluator.py
Muennighoff: Those missing positives can just be filtered out, no? It seems like they wouldn't affect the score, would they? Then maybe your dataet can be turned into the same format as other Reranking tasks and we can use the original evaluator by adding your desired metrics to it.

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
KennethEnevoldsen: You can calculate it using the utility function `task.calculate_length_and_count()` (the formulation isn't perfect but it is consistent)

n_samples is the number of pairs and total length is the sum of length of query and doc in those pairs.

mteb/evaluation/evaluators/RerankingEvaluator.py
crystina-z: oh I believe they would affect the scores? Say among the 100 queries, there are 20 queries with no positive examples returned from the previous stage, then the final reranking score would be divided by all 100 rather than just the 80 queries that have positive documents from the previous stages.

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
crystina-z: sg thanks! I'll add them

mteb/evaluation/evaluators/RerankingEvaluator.py
orionw: I think I may be misunderstanding still - I thought most reranking tasks had some positive (but not all possible ones in all cases) and then the negatives.  The only way to have all the positives present in the Top-K always is if the task was only designed for reranking. 

I think in the MIRACL case it makes sense to include the top 100 (or N) from a standard first-stage model (this seems like BM25, which is great) and then rerank them, even if all positives are not included.  Then the numbers will be realistic as if one had a first stage model and will match other reported numbers on MIRACL reranking. I think this also aligns with other reranking tasks in MTEB, like the SciDocs one, and we can still use the same evaluator.  

Is this the current plan? It might be this is what y'all meant above and I'm just catching up

crystina-z: > Then the numbers will be realistic as if one had a first-stage model and will match other reported numbers on MIRACL reranking.

Yup that's the major concern of us. It'd be nice to have an evaluation aligned with the reranking results outside of mmteb.

Alternatively, do you think we can have multiple evaluator for a task? say an "all-positive-included evaluator" (current RerankingEvaluator) and a "first-stage-only evaluator" (the current MIRACLRerankingEvaluator), and users and specify which evaluators they wanna use. I can update the `mmteb-miracl-reranking` hgf dataset so that it supports both.

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
crystina-z: added!

mteb/evaluation/evaluators/RerankingEvaluator.py
orionw: > say an "all-positive-included evaluator" (current RerankingEvaluator) 

Ah I think I miscommunicated - the current `RerankingEvaluator` does this already. You just have to set the positives to contain only the positives included in the top-K from BM25.  This is what the current [SciDocsReranking](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Reranking/eng/SciDocsReranking.py) is setup to do AFAIK.

The `RerankingEvaluator` is agnostic from what the set of positives are -- so each task can define what they want the set of positives to be. You've set it to be BM25 which I think is a great and thus, there's no need for positives to contain all the positives in the qrels for reranking.

crystina-z: ah right, but the queries without any positives would be skipped during evaluation or that could be bypassed as well?

orionw: Are there queries without positives in the top 100? I'm surprised by that, must be good queries :) 

If we don't include those queries in MTEB I think the eval for MTEB will have a higher score than other rerankers evaluated on MIRACL, so maybe we could keep them just to keep the score comparable? I assume there aren't many queries like this so there won't be too much wasted compute, but could be wrong.

I'm not sure if the current MTEB setup supports positives being an empty list but we can add that to MTEB if we want to keep the eval the same inside MTEB and out

crystina-z: yea supporting positives as an empty list would be nice! I guess we don't even need to really run reranking on them, but just consider those queries when computing the eval metrics (do average on all queries).

I also just quickly summed up the number and percentage of queries that has no positive in top-100 BM25, to get a sense of how much it would impact the scores. It could be non-trivial for some languages (e.g., `zh` has around 30%)

ar: 187 out of 2896 queries (6.46%)
bn: 21 out of 411 queries (5.11%)
en: 82 out of 799 queries (10.26%)
es: 72 out of 648 queries (11.11%)
fa: 117 out of 632 queries (18.51%)
fi: 70 out of 1271 queries (5.51%)
fr: 74 out of 343 queries (21.57%)
hi: 31 out of 350 queries (8.86%)
id: 31 out of 960 queries (3.23%)
ja: 108 out of 860 queries (12.56%)
ko: 27 out of 213 queries (12.68%)
ru: 250 out of 1252 queries (19.97%)
sw: 106 out of 482 queries (21.99%)
te: 138 out of 828 queries (16.67%)
th: 60 out of 733 queries (8.19%)
zh: 121 out of 393 queries (30.79%)
de: 83 out of 305 queries (27.21%)
yo: 30 out of 119 queries (25.21%)

KennethEnevoldsen:
@orionw and @crystina-z I am going over previous PRs and would love to have this merged in. What needs to be done to get it merged in?

crystina-z:
Hi @KennethEnevoldsen! I think the only open conversation above was how to address the queries with no positive in the top-k documents. I feel that might be worth another issue and possibly PR? not too strong tho, I'm also happy to have it done here.

KennethEnevoldsen:
@crystina-z I believe you are right. Ideally, we would like to not have two independent evaluators, but I would rather have this merged.

Can I ask you to do the following minor changes beforehand:

1) Move MIRACLRerankingEvaluator into MIRACLReranking
2) Create an issue on combining the two Evaluators (such that we can address it in the future)
3) Rename the score from NDCG@10 e.g. NDCG@10(MIRACL) such that it is clear that it is calculated differently.
4) Run a model on the task using the CLI so ensure that everything works

KennethEnevoldsen:
I have merged this into a separate branch to address the above concern - will create a new PR with the changes.

PR #645, opened 2024-05-07
Added Crosslingual Semantic Discrimination Task with Four Evaluation Pairs

Andrian0s:
Short Task pitch: We want to investigate the capabilities of multilingual embedding models to perform semantic search across languages. Using an original target sentence (DE), we generate (and massively manually validated) adversarial distractors in the target language. Given the parallel sentence (FR), the models are evaluated in terms of their ability to discriminate the original sentence from the adversarial distractors.

Further plans: Pushed four datasets of this task (WMT19/21 FR->DE x2 and DE->FR x2).  More to be added in the future (under works/verification).

Code Design: 

Original: I maximised reusing AbsTaskRetrieval without extending it. I wrote a separate load_data and shrinked _evaluate_split to minimise extra calculations of evaluations that do not fit within the scope of the task. The rest of the code should work identically to AbsTaskRetrieval. Any new instance of this task simply needs to make a new Metadata object and implement the load_data (since it depends on the language pairs of the metadata).

Updated: Extending AbsTaskRetrieval instead now. I kept the original file structure.

Reason for dataset addition:

Task previously uncovered, different evaluation of Cross-Lingual search of multilingual models.

Results: Performance is similar to the original paper (and therefore a separate implementation) for models except Multilingual-e5 models. Multilingual-e5 models need to take a prefix prompt, which they do not within the TaskRetrieval pipeline. I have confirmed that through manual testing. I will open a separate more detailed issue for someone to pick this up.


mteb/abstasks/AbsTaskCLSD.py
```diff
@@ -0,0 +1,151 @@
...
+from time import time
+
+from ..evaluation.evaluators import RetrievalEvaluator
+from .AbsTask import AbsTask
+
+logger = logging.getLogger(__name__)
+
+
+class AbsTaskCLSD(AbsTask):
+    """Abstract class for Cross Lingual Semantic Discrimination Tasks. Initialised from starting code of AbsTaskRetrieval
```
KennethEnevoldsen, line 15: Please add a line describing the task as well

mteb/abstasks/AbsTaskCLSD.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def load_data(self, **kwargs):
+        """Not implemented, create within the instance. This function must populate:
+        self.corpus: Distractors (Negative) and Parallel (Positive) samples of a search.
+        self.queries: The source text to search with.
+        self.relevant_docs: Denotes which are the relevant documents (The positive).
+        For an example implementation and layout of the huggingface dataset, please look at the CrossLingualSemanticDiscriminationWMT19.py
```
KennethEnevoldsen, line 40: add path

mteb/tasks/CLSD/WMT1921/CrossLingualSemanticDiscriminationWMT19.py
```diff
@@ -0,0 +1,129 @@
...
+        date=("2018-01-01", "2023-12-12"),
+        form=["written"],
+        domains=["News"],
+        task_subtypes=[],
+        license="CC BY-SA 4.0",
+        socioeconomic_status="high",
+        annotations_creators="derived",
+        dialect=[],
+        text_creation="LM-generated and verified",
+        bibtex_citation="to_update",
```
KennethEnevoldsen, line 54: missing

mteb/abstasks/AbsTaskCLSD.py
KennethEnevoldsen: Any reason to not just use a retrieval task?

mteb/abstasks/AbsTaskCLSD.py
Andrian0s: Added

mteb/tasks/CLSD/WMT1921/CrossLingualSemanticDiscriminationWMT19.py
Andrian0s: Will add it once the preprint is available, I added a reminder 👍

KennethEnevoldsen: Let us just keep it as an empty string or now with a comment that a preprint is coming

mteb/abstasks/AbsTaskCLSD.py
KennethEnevoldsen: Thanks for outlining it @Andrian0s, sounds like it is compatible. If that is the case I would much prefer if we keep it the same. This makes maintaining the code notably simpler going forward and since the change is only an added metric to the retrieval task I don't believe it introduced any issues.

CLSD can then become a task subtype.

I will just get a second opinion from @orionw here as well (he did the AbsTaskInstructionRetrieval)

orionw: Thanks for the great work @Andrian0s! 

We extended `AbsTaskInstructionRetrieval` primarily because it evaluates by comparing the difference between two runs as opposed to the standard one run in a retrieval setup. Although there was a different metric, it was more about the fact that the metric needs the model to be run twice over the data and compare rather than the new metric itself -- and no previous abstract task let us run it twice.

You make good points about the pros to separating, but I think the extra time of calculating more metrics is negligible - it should just take at most a few more seconds to calculate the standard metrics beyond accuracy.

--------------

I do think you are right that the [current setup of the dataset](https://huggingface.co/datasets/Andrianos/clsd_wmt19_21/tree/main/wmt19.fr.de) is not in any of the current AbstractTask formats: you have the source, target, and the adversarial sentence as one row in the dataset.

Looking at the task description, I could see this being a retrieval task (a cross-lingual "find the correct sentence among the collection which includes distractors") if you combine all the non-source sentences into one collection. As long as no other sentences are closer to the source than the target and the adversarial examples (which I think is a correct assumption), this should end up the same as your current evaluation setup at recall@1.

On the other hand, your task is also very similar to a classification task, where you want to predict True for the target sentence and False for the adversarial sentences.  This might be truer to your current evaluation if you really only care about accuracy and you're okay with averaging across all instances (whereas retrieval is averaged per source sentence).

However, a third setup that might also work well (although I'm less familiar with it) is that it could probably be formatted as bitext mining. In that case you're looking to mine the correct sentence and not the others - though I think @KennethEnevoldsen is more familiar with this task than me and can weigh in on it.

You are the most familiar with your dataset, but it seems like this task could fit into one of these three topics -- unless I'm misunderstanding something

orionw:
> Multilingual-e5 models need to take a prefix prompt, which they do not within the TaskRetrieval pipeline.

The task prefixes are provided by the model implementation, instead of in the data.  You can test with [this setup for E5](https://github.com/microsoft/unilm/blob/master/e5/mteb_beir_eval.py) as it requires more than what can be provided by the command line arguments.

mteb/tasks/CLSD/WMT1921/CrossLingualSemanticDiscriminationWMT19.py
Andrian0s: Updated

Andrian0s:
> > Multilingual-e5 models need to take a prefix prompt, which they do not within the TaskRetrieval pipeline.
> 
> The task prefixes are provided by the model implementation, instead of in the data. You can test with [this setup for E5](https://github.com/microsoft/unilm/blob/master/e5/mteb_beir_eval.py) as it requires more than what can be provided by the command line arguments.

I was not aware e5 was not intended to work straight out of the box. My apologies for raising this as an issue, still new with this repository.

mteb/abstasks/AbsTaskCLSD.py
Andrian0s: Now removed document

mteb/abstasks/AbsTaskCLSD.py
Andrian0s: Thank you @KennethEnevoldsen  and @orionw for the quick suggestions, promptness really helps.

I appreciate the discussion and I followed Orion's suggested and instead used the AbsRetrieval class. A lot of the evaluations are irrelevant (but not causing an issue), but since we will only use recall_at_1 as the main (which is same as our accuracy, verified), no harm done. 

To be fair, recall_at_3 isn't such a bad insight either. 

I think it's good to go, current implementation is simplistic and sufficient.

Andrian0s:
I assume the comments are resolved, could you give them one last check @KennethEnevoldsen @orionw?

Last thing to discuss is points:

My current proposal:

4 datasets -> 8 points
2 uniquely introduced cross-lingual pairs bonus -> 8 points
Pseudo new task -> 2 points

Total=18

As the second author of the paper is a major contributor for this task I would suggest to give him 10 points to get him on the author list.

The remaining 8 would also push me to 10, which I would be satisfied with. I will probably stick around and help by doing some reviews as I like the initiative, but knowing I am over the 10 line would be comfortable.

Please let me know if this is reasonable.

mteb/abstasks/AbsTaskCLSD.py
KennethEnevoldsen: Thanks for the changes and the considerations here - For now the tab is not required in MTEB, but can be added later

KennethEnevoldsen:
@Andrian0s the points sounds very reasonable. Do also add the points of the reviewer. Otherwise I believe we are ready to merge this PR

KennethEnevoldsen (approved):

orionw (approved):

Andrian0s:
added points, assuming this didn't break the tests, we are ready for merging. Thanks for this!

KennethEnevoldsen:
Updated from main and enabled auto-merge 🥂

PR #623, opened 2024-05-02
XNLI 2.0 subset added 

dokato:
I added extension of XNLI dataset, XNLI 2.0, described in this paper: https://arxiv.org/pdf/2301.06527 for Indian languages.

I collated them into one multilingual set added to: [`mteb/xnli2.0-multi-pair`](https://huggingface.co/datasets/mteb/xnli2.0-multi-pair)

## Checklist for adding MMTEB dataset

Reason for dataset addition:

- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [x] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 
- [ ] I have added points for my submission to the [points folder](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) using the PR number as the filename (e.g. `438.jsonl`).

Issue #419, opened 2024-04-18
Some datasets for languages.


x-tabdeveloping:
I'm gonna practice drums for the rest of the day and probably won't work tomorrow, but for those who are looking to contribute and get some of those juicy points here is some low-hanging fruit in diverse languages:

## Slovak:
- ~~Sentiment: https://huggingface.co/datasets/sepidmnorozy/Slovak_sentiment~~ (as a matter of fact she has loads of Sentiment classification datasets: https://huggingface.co/sepidmnorozy)
- News Summarization: https://huggingface.co/datasets/kiviki/SlovakSum

## Greek:
 - ~~Legal code clustering: https://huggingface.co/datasets/AI-team-UoA/greek_legal_code~~
 - NLI: https://huggingface.co/datasets/Harsit/xnli2.0_greek
 - Medical QA: https://huggingface.co/datasets/ilsp/medical_mcqa_greek

## Maltese:
 - News titles: https://huggingface.co/datasets/MLRS/maltese_news_headlines
 - News categories: https://huggingface.co/datasets/MLRS/maltese_news_categories

dokato:
I'm gonna pick up `kiviki/SlovakSum` if noone is on it yet.

dokato:
On the other hand it seems like the summary task requires:
```
        human_summaries: list[str]
        machine_summaries: list[str]
        relevance: list[float] (the score of the machine generated summaries)
```

and `kiviki/SlovakSum` doesn't have neither machine_summaries nor relevance scores.

x-tabdeveloping:
@dokato Try formulating it as a retrieval task instead :))

wissam-sib:
I can start working on the Maltese dataset if no one is

mariyahendriksen:
I will take care of the sentiment classification dataset for Slovak, if no one claimed it yet

KennethEnevoldsen:
Will close this issue for now - I assume many of these are still relevant to add

@mariyahendriksen do you still want to add the greek medical QA?

PR #649, opened 2024-05-08
Added XNLI V2.0 Greek, Turkish, Russian support

Andrian0s:

Reason for dataset addition:

I picked this up from #419. Everything seems to be working fine. 
To further confirm the quality of the dataset, As a greek native speaker, I read through 20 samples of the dataset to also verify the quality of the underlying dataset. Seems mostly okay. I also outsourced a quick check of Turkish and Russian to a native speaker and the translation quality seems ok and coherent.

Their claim to verifying these datasets is: "We also explored the quality of translation in the XNLI test and dev sets." plus a section where they show qualitative head to head comparisons.

We also already merged in from #623 another subset of this multilingual dataset. I updated the created metadata to match the paper.


Andrian0s:
I might have messed up and merged this in by accident already, trying to figure it out

orionw (approved):
LGTM, thanks for adding!

orionw (changes requested):
Let's merge the XNLIv2 task into the multilingual set, unless that doesn't make sense for a reason I'm missing @Andrian0s

Andrian0s:
> Let's merge the XNLIv2 task into the multilingual set, unless that doesn't make sense for a reason I'm missing @Andrian0s

@orionw 
I looked at the previous PR, seems like a better solution. I will adapt to that. (although that subpart had more of an indian languages theme currently, but why not more anyway)

Meanwhile, how can I upload the new extended set to MTEB huggingface dataset to keep it consistent.

(if you need to provide access, do so to https://huggingface.co/Andrianos)

Else, what are the instructions?

orionw:
Thanks @Andrian0s! You can just upload it any dataset (one you own on your account perhaps?) in the `path` section of the task metadata.  We can migrate it later if needed. 

I think @KennethEnevoldsen has a plan for migrating someday, but for now any dataset is fine.

Andrian0s:
@orionw I updated to follow the new format. However, I was not able to run the results without deploying a temporary hack.

I dived into it and realised it is because XNLI2 is a subclass of XNLI and our automatic check only checks for subclasses of any, AbsTasks, not classes such as XNLI. I am not sure how the previous PR managed to run it, but I don't see how it would be any different when running from command prompt.

I propose: 
I remove the commented hack (see comment in MTEB.py), we merge is as is (as the issue is we have an unreachable XNLI2 which we know works, so no big deal). 

Then, I open an issue with information where the error is, why, and put it up as a good first issue. Then, to verify that their fix works, they can simply try to run XNLI2 from the command prompt without any hack, if it works, their fix is complete.

Tell me what u think about this

Andrian0s:
> @orionw I updated to follow the new format. However, I was not able to run the results without deploying a temporary hack.
> 
> I dived into it and realised it is because XNLI2 is a subclass of XNLI and our automatic check only checks for subclasses of any, AbsTasks, not classes such as XNLI. I am not sure how the previous PR managed to run it, but I don't see how it would be any different when running from command prompt.
> 
> I propose: I remove the commented hack (see comment in MTEB.py), we merge is as is (as the issue is we have an unreachable XNLI2 which we know works, so no big deal).
> 
> Then, I open an issue with information where the error is, why, and put it up as a good first issue. Then, to verify that their fix works, they can simply try to run XNLI2 from the command prompt without any hack, if it works, their fix is complete.
> 
> Tell me what u think about this

Alternatively, on a follow-up thought, if this is not intended new behaviour, I can change XNLI2 to take from the AbsTask and make it work independently, we would basically need to duplicate two functions.

orionw:
Thanks @Andrian0s, it looks a lot better.  

I think I realize the confusion now, sorry I didn't understand - you can't add to the normal XNLI because you don't own that dataset on Huggingface (it's just `xnli` under the default HF account) yet these new additions are still part of XNLI, just machine-translated. Is this a correct understanding?

In this case I see two options: 

1. Make it a new multilingual `XNLITranslated` dataset (or similar name) and move your new languages to that one -- so that we don't have to have the hack but can still include your updates in the multilingual section
2. Clone the original XNLI dataset and combine it with your dataset, then replace the `XNLI` dataset path to be your updated dataset that includes both standard XNLI and the additional translated ones

I think either is fine, both have their pros/cons.  @KennethEnevoldsen do you agree?

Andrian0s:
> Thanks @Andrian0s, it looks a lot better.
> 
> I think I realize the confusion now, sorry I didn't understand - you can't add to the normal XNLI because you don't own that dataset on Huggingface (it's just `xnli` under the default HF account) yet these new additions are still part of XNLI, just machine-translated. Is this a correct understanding?
> 
> In this case I see two options:
> 
> 1. Make it a new multilingual `XNLITranslated` dataset (or similar name) and move your new languages to that one -- so that we don't have to have the hack but can still include your updates in the multilingual section
> 2. Clone the original XNLI dataset and combine it with your dataset, then replace the `XNLI` dataset path to be your updated dataset that includes both standard XNLI and the additional translated ones
> 
> I think either is fine, both have their pros/cons. @KennethEnevoldsen do you agree?

XNLI and XNLI2.0 are separate "translation" instances, MT and verified according to their paper.  They should remain separate results and runs and they should definetely not be mixed.

XNLI2 in the code is a new class that inherits from the XNLI class, pushed in PR #623 and includes a subset of the languages of XNLI2.0

This PR, simply adds three more languages to this subset (which I verified the quality of briefly through native speakers). I did not have access to the hf dataset so I made a copy on my huggingface account.

The thing is, I simply add the new languages, I cannot run XNLI2 class anyway. This is because the XNLI2 class as given in PR #623, initialised it as class XNLI2(XNLI). I do not know how the results in the previous PR were created, I assume through some similar hack.

If I have the correct understanding, we cannot run this through "mteb run -m {model_name} -t XNLI2" because of the ways lines 155-160 on MTEB.py are currently. The reason is, the current code will not consider anything a task unless it directly (no 2 step) extends any of the AbsTasks.  The only way I managed to run it was through the commented hack I showed (which should not be used further).

So we have two options:

1. We keep this as a limitation and I change the class definition and simply duplicate some code to make XNLI2 work.
2. We merge this PR as is and raise this as an issue for someone to solve, with the example of XNLI2 as their indication whether their fix works or not.

I think both options are okay. Option1 allows us a direct solutoin with less chance of breaking things, whilst Option2 allows us greater potential for code reusability and better design in the future (in case we have more tasks that make sense to be defined as  --- class XNLI2(XNLI) -

orionw:
Hmmm, we definitely should be able to run XNLI2 as is -- give me a bit and I will attempt to replicate this

orionw:
You are correct @Andrian0s, I see how MTEB does that check now.  

Can you go with Option 1? Lets just make two classes both inheriting from (MultilingualTask, AbsTaskPairClassification) and keep it that way even though there is some minor deduplication.

Thanks for the investigation! Once you update that I think this should be good to merge.

mteb/tasks/PairClassification/multilingual/XNLI.py
```diff
@@ -103,19 +103,22 @@ def dataset_transform(self):
...
 }
 
 
 class XNLI2(XNLI):
     metadata = TaskMetadata(
         name="XNLI2",
         dataset={
-            "path": "mteb/xnli2.0-multi-pair",
-            "revision": "1a3794e88c74db2a13aebc5e8abc3dc4312a7543",
+            "path": "Andrianos/xnli2.0-multi-pair",
```
imenelydiaker, line 116: Stepping in before merging, don't you think it would be better to update the current mteb dataset with these new languages?

orionw:
@imenelydiaker I think he is adding it to XNLI2, which is what I requested earlier. XNLI2 already exists, he's just adding more languages

imenelydiaker:
> @imenelydiaker I think he is adding it to XNLI2, which is what I requested earlier. XNLI2 already exists, he's just adding more languages

He can update the dataset then by openning a PR and we'll merge it

orionw:
Sure @Andrian0s could make a PR to the dataset from the contents of what he's already done and then he won't need to create a new dataset. Or he could just make a new dataset and we can replace the existing one with it by renaming it, both works 🤷

FWIW, this PR is still needed for MTEB since the languages and size has changed.

imenelydiaker:
> Sure @Andrian0s could make a PR to the dataset from the contents of what he's already done and then he won't need to create a new dataset. Or he could just make a new dataset (since his already combined the two) and we can replace the existing one with it by renaming it, both works 🤷
> 
> FWIW, this PR is still needed for MTEB since the languages and size has changed.

Sorry I deleted my previous comment because I was trying to understand the problem here. Link on how to [make PRs on HF](https://huggingface.co/docs/hub/repositories-pull-requests-discussions).

@orionw I agree with the option you proposed here: 

> Make it a new multilingual XNLITranslated dataset (or similar name) and move your new languages to that one -- so that we don't have to have the hack but can still include your updates in the multilingual section

This would help distinguish the original XNLI and translated ones, I think we'll have multiple translations, so it's okay to open PR on HF and update it.

I also don't mind if XNLI2 is renaled to XNLITranslated and the functions are overwritten again, I find it better to create a task that inherits from `AbsTask` then another task.

Andrian0s:
> > Sure @Andrian0s could make a PR to the dataset from the contents of what he's already done and then he won't need to create a new dataset. Or he could just make a new dataset (since his already combined the two) and we can replace the existing one with it by renaming it, both works 🤷
> > FWIW, this PR is still needed for MTEB since the languages and size has changed.
> 
> Sorry I deleted my previous comment because I was trying to understand the problem here. Link on how to [make PRs on HF](https://huggingface.co/docs/hub/repositories-pull-requests-discussions).
> 
> @orionw I agree with the solution you proposed here:
> 
> 1. Make it a new multilingual XNLITranslated dataset (or similar name) and move your new languages to that one -- so that we don't have to have the hack but can still include your updates in the multilingual section
> 
> This would help distinguish the original XNLI and translated ones, I think we'll have multiple translations, so it's okay to open PRs on HF and update it.
> 
> Here XNLI2 is not supposed to run XNLI languages again, so it makes sense to have it as a separate new task with its own functions. I wouldn't mind if XNLI2 is renamed into `XNLITranslated` and the functions are overwritten again, I find it better to create a task that inherits from `AbsTask` than another existing task.

So basically, the plan that we all agreed on is updated now. I also added points. We are ready to merg.

@orionw @imenelydiaker, would someone approve and we merge?

mteb/tasks/PairClassification/multilingual/XNLI.py
```diff
@@ -103,19 +103,22 @@ def dataset_transform(self):
...
 
-class XNLI2(XNLI):
+class XNLIV2(MultilingualTask, AbsTaskPairClassification):
     metadata = TaskMetadata(
-        name="XNLI2",
+        name="XNLIV2",
         dataset={
-            "path": "mteb/xnli2.0-multi-pair",
-            "revision": "1a3794e88c74db2a13aebc5e8abc3dc4312a7543",
+            "path": "mteb/xnli2.0-multi-pair-13",
```
imenelydiaker, line 116: Sould we remove the old one then https://huggingface.co/datasets/mteb/xnli2.0-multi-pair ? I thought you'd just update `xnli2.0-multi-pair`

imenelydiaker (commented):
LGTM! Thanks for the fixes and the extension!

mteb/tasks/PairClassification/multilingual/XNLI.py
Andrian0s: it didn't allow me, then I made a new one that includes the current languages. I tried to delete the older one but it did not allow me. Could you please try to delete the older one?

imenelydiaker: You couldn't open a PR on the repo? I don't have my laptop with me atm, I can do it tonight

Andrian0s: From my attempts on my own datasets, I haven't found the way to delete through a PR. Probably needs to happen from the interface with a person who has the right access? (I guess I am a member but that might not be enough)

Let me know if u have another idea

imenelydiaker: @Andrian0s I jus opened a PR here: https://huggingface.co/datasets/mteb/xnli2.0-multi-pair/discussions/3
I'll merge it and you'll just need to update the path in this PR to use this repository.

imenelydiaker, line 116:
```suggestion
            "path": "mteb/xnli2.0-multi-pair",
```

mteb/tasks/PairClassification/multilingual/XNLI.py
```diff
@@ -103,19 +103,22 @@ def dataset_transform(self):
...
+            "revision": "c3a8ffbc19ab5d22b0958961490b4a5654b19d89",
```
imenelydiaker, line 117:
```suggestion
            "revision": "5b7d477a8c62cdd18e2fed7e015497c20b4371ad",
```

imenelydiaker (commented):
I added the subsets here: https://huggingface.co/datasets/mteb/xnli2.0-multi-pair.
I'd prefer to have all new subsets of XNLI2.0 added to this repository via PRs, for consistency. I'll delete this one: https://huggingface.co/datasets/mteb/xnli2.0-multi-pair-13 after this PR is merged.

I'll let you update the path and revision_id and we'll merge it. Thanks fo rthis contribution!

Andrian0s:
> I added the subsets here: https://huggingface.co/datasets/mteb/xnli2.0-multi-pair. I'd prefer to have all new subsets of XNLI2.0 added to this repository via PRs, for consistency. I'll delete this one: https://huggingface.co/datasets/mteb/xnli2.0-multi-pair-13 after this PR is merged.
> 
> I'll let you update the path and revision_id and we'll merge it. Thanks fo rthis contribution!

Good thought, should have done that in the first place. Thank you!

Once tests are run again (which they should), I will enable auto-merge

orionw (approved):
LGTM! Thanks @Andrian0s!

PR #658, opened 2024-05-09
WIP: feat: Standardize MTEB results

### LINKED ISSUE #639: Standardizing results format
Sometimes a dataset goes from monolingual to multilingual, e.g. https://github.com/embeddings-benchmark/mteb/pull/556 

This changes the result file structure (adds language dict) & also how the dataset is represented in the leaderboard (with brackets & language code vs without). Now there are some models that ran on `MLSUMClusteringP2P` when it was still monolingual thus their metadata has no `(fr)` e.g. [this one](https://huggingface.co/OrdalieTech/Solon-embeddings-large-0.1/edit/main/README.md), while newer models now have `(fr)` e.g. [this one](https://huggingface.co/datasets/mteb/results/blob/main/results/voyage-law-2/MLSUMClusteringP2P.json). This currently prevents the voyage-law-2 MLSUMClusteringP2P clustering scores from showing up as it is still defined without the `(fr)` [here](https://huggingface.co/spaces/mteb/leaderboard/blob/main/config.yaml#L214). If we change the leaderboard definition to `(fr)` then models with the old one like `Solon-embeddings-large-0.1` will no longer show up.

I'm not sure what is the best solution --- maybe always having a language code even for English datasets & always having the language dict structure? (also mentioned here: https://github.com/embeddings-benchmark/mteb/issues/251#issuecomment-2036569253)

cc @tomaarsen @KennethEnevoldsen & anyone else who may have thoughts 😊

KennethEnevoldsen:
This is the first part of standardizing MTEB results. I believe it is best split up into two PRs as I already had to do quite a few things to get this to work.

To do this it will introduce breaking changes. It implements the following:

- MTEBResults
  - [x] introduces a dataclass for ensuring a standard result format
  - [x] added functionality for summarizing pr. language (optionally we can have it save this)
  - [ ] Added it is the output of MTEB.run
- LangScriptFilter
  - [x] moved existing code to a utility filter for filtering language and scripts
- [x] Changed the abstask.langs to abstasks.hf_subsets to not confuse hf_subsets with iso 639 language codes
- Standardization
  - [x] Standardized the classification task
- [ ] restructuring, I believe the abstasks folder has gotten a bit crowded with non-abstasks stuff (mostly due to me). I would probably move it out of the folder
- [ ] Implement conversion script from old scores

I have generally removed cases where we compute the std or other measure of uncertainty and instead included the entire range (this allows computation of CI, std, win-rate etc.)

Fixes #639


mteb/abstasks/AbsTaskClassification.py
```diff
@@ -144,16 +145,13 @@ def _evaluate_monolingual(
...
-        else:
-            avg_scores = {k: np.mean([s[k] for s in scores]) for k in scores[0].keys()}
-            std_errors = {
-                k + "_stderr": np.std([s[k] for s in scores]) for k in scores[0].keys()
-            }
-            return {**avg_scores, **std_errors}
+        avg_scores: dict[str, Any] = {
+            k: np.mean([s[k] for s in scores]) for k in scores[0].keys()
+        }
+        avg_scores["scores_per_experiment"] = scores
```
KennethEnevoldsen, line 151: I could imagine we would want to standardize this across task types. If so do we want it to be a list of dicts of a dict with lists

KennethEnevoldsen: I am more for the first option (otherwise the we would have e.g. accuracy and accuracies)

mteb/abstasks/languages.py
```diff
@@ -14,3 +17,54 @@
 
 with path_to_lang_scripts.open("r") as f:
     ISO_TO_SCRIPT = json.load(f)
+
+
+@dataclass
+class LangScriptFilter:
```
KennethEnevoldsen, line 23: a utility class for filtering language codes. Previously within abstasks, moved out as I needed it for the results object.

I also believe this implementation is notably more clear.

mteb/abstasks/mteb_result.py
```diff
@@ -0,0 +1,183 @@
...
+
+def eval_langs_as_dict(
+    eval_langs: LANGUAGES,
+) -> dict[HFSubset, list[ISO_LANGUAGE_SCRIPT]]:
+    if isinstance(eval_langs, list):
+        return {"": eval_langs}
+    return eval_langs  # type: ignore
+
+
+class MTEBResults(BaseModel):
```
KennethEnevoldsen, line 34: The main change in this PR. 

Currently, if a dataset has no subset on huggingface a fake subset is added (""). This makes the format standard, but isn't the prettiest I can imagine. An alternative is to use None, which I might prefer, but when serializing it to json None turns to "null". A third option is to convert it to a list of dicts where the dict have 

```
{"train": [{"hf_subset": "en-de" "main_score": 0.5", ...}, {"hf_subset": "en-fr" "main_score": 0.6", ...}, ]
```
This leads to notably less confusion when `"hf_subset": "null"` as the name is specified.

KennethEnevoldsen: Looking this over again I am probably more for the last option.

mteb/abstasks/mteb_result.py
```diff
@@ -0,0 +1,183 @@
...
+            for hf_subset, scores in self.scores[split].items():
+                eval_langs = hf_subset2langs[hf_subset]
+                for lang in eval_langs:
+                    if lang_script_filter.is_valid_language(lang):
+                        values.append(getter(scores))
+                        break
+
+            return aggregation(values)
+
+    def get_main_score(
```
KennethEnevoldsen, line 159: This can extract scores for a specific language, split or script. Useful for going from hf_subset to language codes.

KennethEnevoldsen: See example in docstring.

mteb/abstasks/languages.py
```diff
@@ -14,3 +17,61 @@
...
+
+        filter_scripts = scripts is not None
+        script_codes: set[str] = set(scripts) if filter_scripts else set()
+        return cls(
+            language_scripts=lang_script_codes,
+            scripts=script_codes,
+            languages=normalized_langs,
+        )
+
+    def is_valid_language(self, language: str) -> bool:
```
isaac-chung, line 56: Does this (and `is_valid_script`) function check for containment instead of validity? Would something like `contains_language` be more fitting?

tests/test_langscript_filter.py
```diff
@@ -0,0 +1,15 @@
+
+
+from mteb.abstasks.languages import LangScriptFilter
+
+def test_langscript_filter():
+
+    ls_filter = LangScriptFilter.from_languages_and_scripts(
+        languages=["fra"],
+        scripts=None
```
isaac-chung, line 9: Nice! Would be great to include scripts as well.

mteb/abstasks/mteb_result.py
```diff
@@ -0,0 +1,182 @@
...
+
+            for hf_subset, scores in self.scores[split].items():
+                eval_langs = hf_subset2langs[hf_subset]
+                include_subset = False
+                for lang in eval_langs:
+                    if lang_script_filter.is_valid_language(lang):
+                        include_subset = True
+                        break
+                if include_subset:
+                    values.append(getter(scores))
```
isaac-chung, line 154: Could be simplified?
```suggestion
                    if lang_script_filter.is_valid_language(lang):
                        values.append(getter(scores))
                        break
```

tests/test_mteb_results.py
```diff
@@ -0,0 +1,64 @@
...
+
+    mteb_results = MTEBResults(
+        dataset_revision="1.0",
+        task_name="dummy_task",
+        mteb_version="1.0",
+        scores=scores,
+    )
+
+    assert mteb_results.get_main_score() == 0.55
+    assert mteb_results.get_main_score(languages=["fra"]) == 0.6
```
isaac-chung, line 62: Just to confirm: If languages=`["eng"]`, would the main score be 0.55?

mteb/abstasks/mteb_result.py
```diff
@@ -0,0 +1,182 @@
...
+    def to_disk(self, path: Path) -> None:
+        with path.open("w") as f:
+            f.write(self.model_dump_json())
+
+    @classmethod
+    def from_disk(cls, path: Path) -> MTEBResults:
+        with path.open("r") as f:
+            return cls.model_validate(json.load(f))
+
+    def get_score(
```
isaac-chung, line 111: If this is not intended to be used, could it be merged with `get_main_score`?

mteb/abstasks/mteb_result.py
isaac-chung: I thought the idea was to include the language even when it's monolingual (no subset case) like in https://github.com/embeddings-benchmark/mteb/issues/639#issuecomment-2097788082.

isaac-chung (commented):
Pretty solid! A few comments below. Getting late so will check my work in the AM.

mteb/abstasks/mteb_result.py
KennethEnevoldsen: Yep. The issue I found was that we might loose "performance on split hf_subset X" which might be relevant as it is not recoverable from languages. However with the functionality as it is now we can easily change to use languages instead (which might be a better solution)

mteb/abstasks/languages.py
```diff
@@ -14,3 +17,54 @@
...
+        # normalize to 3 letter language codes
+        normalized_langs = set()
+        filter_lang = languages is not None
+
+        if filter_lang:
+            for lang in languages:
+                lang_script = lang.split("-")
+
+                is_lang_script_code = len(lang_script) == 2
+                if is_lang_script_code:
```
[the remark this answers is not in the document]
KennethEnevoldsen: It is a way to mark the intent of `len(lang_script) == 2` which might otherwise be unclear. In that way, it acts more like a comment (you might be interested in this [video argument](https://www.youtube.com/watch?v=Bf7vDBBOBUA&t=168s))

mteb/abstasks/AbsTask.py
```diff
@@ -126,43 +127,23 @@ def filter_languages(
...
-                    splits_to_keep.append(hf_lang)
-                    continue
-
-                _lang, _script = langscript.split("-")
-                if (filter_lang and _lang in normalized_langs) or not filter_lang:
-                    if script is None or _script in script_codes:
-                        splits_to_keep.append(hf_lang)
+                if lang_script_filter.is_valid_language(langscript):
+                    subsets_to_keep.append(hf_subset)
+                    break
```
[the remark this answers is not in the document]
KennethEnevoldsen: added

mteb/abstasks/languages.py
KennethEnevoldsen: Have renamed langscriptfilter to LanguageScripts and used contains (I think it is better as well)

```
LanguageScripts.contains_language(...) 
```

tests/test_langscript_filter.py
KennethEnevoldsen: I have expanded the test cases to both cover script and other special cases

mteb/abstasks/mteb_result.py
KennethEnevoldsen: Agree. Added this mostly to clarify what the for loops check for.

tests/test_mteb_results.py
KennethEnevoldsen: Yes. Also added it as a test case.

mteb/abstasks/mteb_result.py
KennethEnevoldsen: This is the generalized case of the get_main_score. My guess is that most will use get main score (as it is simplified), but this function is more flexible allowing you to arbitrarily extract from the scores and I could imagine it being used either internally in the package or by users (e.g. to get evaluation times or get uncertainties). get_main_score is just a wrapper.

mteb/abstasks/mteb_result.py
KennethEnevoldsen: I will wait to get some more opinions on this

mteb/abstasks/AbsTaskClassification.py
imenelydiaker: The current format:
```
{
task_name: ...,
eval_time: ..,
lang: { metric: score, ...}
...
}
```
Is okay if we expand it to monolingual tasks also. Unless you have a better suggestion

mteb/abstasks/AbsTask.py
```diff
@@ -126,43 +127,23 @@ def filter_languages(
...
 
-        self.langs = splits_to_keep
+        self.hf_subsets = subsets_to_keep
```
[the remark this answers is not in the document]
x-tabdeveloping: `hf_subsets` is good. Especially since it communicates that we are working around HF Hub's abstractions here, so the name even includes the rationale.

mteb/abstasks/languages.py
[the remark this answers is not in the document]
x-tabdeveloping: I have to agree with @isaac-chung here. Especially with the `filter_lang = languages is not None` line. I had to go back and check what that variable contained and it was a bit hard for me to follow the flow of the code. Naming things in a way that everyone understands is way harder than letting people read the behaviour directly and maybe including a comment where the intent isn't clear.

mteb/abstasks/AbsTaskClassification.py
KennethEnevoldsen: @imenelydiaker this section is more on the format of the scores dict on whether is should be:

```
{"accuracy": 0.5, "accuracies": [0.4, 0.6]}
# or
{"accuracy": 0.5, "scores_per_experiment": [{accuracy": 0.4} , {accuracy": 0.6}]}
``` 
I prefer the second

KennethEnevoldsen: However, I would love your opinion on the result format below

mteb/abstasks/mteb_result.py
KennethEnevoldsen: (@imenelydiaker, @x-tabdeveloping)

mteb/abstasks/AbsTaskClassification.py
imenelydiaker: Agreed the second one with the list is better

mteb/abstasks/AbsTask.py
```diff
@@ -126,43 +127,23 @@ def filter_languages(
...
-                is_lang_script_code = len(lang_script) == 2
-                if is_lang_script_code:
-                    normalized_langs.add(lang_script[0])
-                    lang_script_codes.add(lang)
-                else:
-                    normalized_langs.add(lang)
-
-        filter_scripts = script is not None
-        script_codes: set[str] = set(script) if filter_scripts else set()
+        lang_script_filter = LangScriptFilter.from_languages_and_scripts(
```
x-tabdeveloping: I'm not entirely sure I like this from a locality of behaviour standpoint. Up until now I could just look at the code her and understand what it did, now I have to look in another file in a dataclass for behaviour that directly affects what is happening here.

I might misunderstand the intention here, but I think this method would pretty much do the same thing and is (at least to my eyes) a bit easier to understand.
```python
def filter_languages(
    self, languages: list[str] | None, script: list[str] | None = None
) -> AbsTask:
    splits_to_keep: list[str] = []
    if not isinstance(self.metadata.eval_langs, dict):
        self.langs = self.metadata.eval_langs
        return self
    for hf_subset, langs in self.metadata.eval_langs.items():
        for langscript in langs:
            _lang, _script = langscript.split("-")
            if (languages is None) or (_lang in languages) or (langscript in languages):
                if (script is None) or (_script in script):
                    splits_to_keep.append(hf_subset)
    self.langs = splits_to_keep
    return self
```

mteb/abstasks/mteb_result.py
x-tabdeveloping: Isn't it just called `"default"` if you don't specify any subset when uploading the dataset? I think we could totally use that.

mteb/abstasks/mteb_result.py
```diff
@@ -0,0 +1,183 @@
...
+        0.6
+    """
+
+    dataset_revision: str
+    task_name: str
+    mteb_version: str
+    scores: dict[Split, dict[HFSubset, Scores]]
+
+    @field_validator("scores")
+    def _validate_scores(
```
x-tabdeveloping, line 73: I know it might be a bit late for this, but maybe we should consider pulling in pydantic for all of this. Seems like a lot of validation code to write and maintain

mteb/abstasks/mteb_result.py
x-tabdeveloping: Hmm I'm not sure how I feel about `getter` my feeling is that getter is something you use in programming when you have a private variable whose reference should not be exposed, but its value should be accessible by other objects.
I don't know if I could come up with a better name for this though.

Is it not the main idea though that we should somehow be able to allow users to expose an attribute of a dictionary at arbitrary depths? In that case I would prefer some path to a deeply nested value.
e.g. if you have a Scores dict like this:
```python
{
   "something": {
        "something_else": {"beautiful_score": 0.76}
    }
}
```
Then you could have a path to this score:
```python
get_score(..., score_path=("something", "something_else", "beautiful_score"))
```
Or I've seen some libraries for accessing these with a dot syntax:
```
"something.something_else.beautiful_score"
```

mteb/abstasks/AbsTask.py
KennethEnevoldsen: > I'm not entirely sure I like this from a locality of behaviour standpoint. Up until now I could just look at the code her and understand what it did, now I have to look in another file in a dataclass for behaviour that directly affects what is happening here.

The current format specified `langscripts.contains_language(language) or langscripts.contains_language(script)`. I believe that is more clear than:
`(languages is None) or (_lang in languages) or (langscript in languages)`

As mentioned previously the main idea was also to re-use the functionality. It also allows me to test the functionality directly in a test.

I am also not entirely sure it is equivalent. I would have to test it. E.g. I believe script = ["Cyrl"], language = ["eng-Latn"] with input "eng-Latn" would fail even though it shouldn't.

mteb/abstasks/mteb_result.py
KennethEnevoldsen: Apparently, both works (default is probably better):
Screenshot 2024-05-10 at 12 30 43

@x-tabdeveloping Would you prefer the hf_subsets or just having language codes instead?

tests/test_langscripts.py
x-tabdeveloping, line 1: I fear a number of things with this test:
1. It's a lot of code, comparable to the implementation itself
2. It is very granular, testing internal behaviours of the library. It might just be my views but what actually matters is if the parts of the library that are highly coupled and have to interact work together after changes. What I mean here is that we might miss errors related to language filtering because we test a self-contained dataclass instead of let's say testing, whether after running the whole benchmark with a dummy task we only get the languages in the results that we specified.

tests/test_mteb_results.py
x-tabdeveloping, line 1: This one I really like!! These are the kinds of tests we should have more of I think.

mteb/abstasks/AbsTaskClassification.py
x-tabdeveloping: Agree

mteb/abstasks/mteb_result.py
KennethEnevoldsen: Are we discussing the get_main_score or the getter argument in get_score here?

I don't see a reason why not simply using a function. the path seems to unreasonably limit the user who e.g. wants the mean of recall_at_1 and ndgc_at_10 and it doesn't require custom syntax. 

> The getter function is used to retrieve the variable value

I believe this fits with my intuition as well

mteb/abstasks/mteb_result.py
KennethEnevoldsen: This is pydantic

mteb/abstasks/mteb_result.py
x-tabdeveloping: We should definitely have `hf_subsets`. I really like the way it is handled in multilingual tasks:
```
{"en": ["eng-Latn"], "et-hu": ["est-Latn", "hun-Latn"]}
```
And then we should follow a similar format with the monolingual tasks just with the default split:
```
{"default": ["eng-Latn"]}
```
In theory only internals should care what the splits are named and its not something we should rely on beyond practicalities.

tests/test_langscripts.py
KennethEnevoldsen: The tests suggested by @isaac-chung revealed errors that seem like reason enough to keep it. 

We also test the abstasks filtering through the get_tasks tests as you specify.

It is also hard to debug larger interacting code and being confident that a specific segment works as intended allows me to find the bug easier.

mteb/abstasks/mteb_result.py
x-tabdeveloping: Mmm that's a good example, I accept :D
I'm a bit uncomfortable with the name `getter` but might just be my OOP trauma from earlier years, if others don't take issue with it I'll give it a pass

tests/test_mteb_results.py
KennethEnevoldsen: Thanks, dependency injection is great

mteb/abstasks/AbsTask.py
x-tabdeveloping: Yeah I wasn't sure about what it was doing exactly, like what cases are included and what cases aren't so my code might not do that correctly.
Regarding the point about clarity: I'll try not to be radical, I think the solution you propose is fine, it's just that I don't necessarily like that I can't know what `contains_language` does until I look.

mteb/abstasks/mteb_result.py
x-tabdeveloping: I certainly like the first option with the list. I believe adding the languages is a good idea, as we can't (or shouldn't) rely on the splits being language names.

mteb/abstasks/mteb_result.py
x-tabdeveloping: oh alright, ignore it then xd

tests/test_langscripts.py
x-tabdeveloping: Alright, fair points.

mteb/abstasks/AbsTask.py
KennethEnevoldsen: I think it is a totally fair point - Style is important but hard to get right good with some discussion

mteb/abstasks/AbsTaskClassification.py
isaac-chung: +1 on 2nd

isaac-chung: Is this only done for the main metric? If so, it's more space efficient to use something like
```
{"accuracy": 0.5, "scores_per_experiment": [0.4, 0.6]}
```

tests/test_langscripts.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+test_cases = [
+    LangScriptTestCase(
+        args=dict(languages=["fra"], scripts=None),
+        contains_language=["fra", "fra-Latn"],
+        not_contains_language=["eng"],
+        contains_script=[],
+        not_contains_script=["Latn"],
+    ),
```
isaac-chung, line 25: I didn't find this intuitive - maybe I am not as familiar with how the languages are defined. If `contains_language=["fra", "fra-Latn"],`, why wouldn't `contains_script=["Latn"]` be true?

README.md
```diff
@@ -96,17 +96,18 @@ evaluation = MTEB(task_categories=['S2S']) # Only select sentence2sentence datas
...
+You can also specify which languages to load for multilingual/cross-lingual tasks like below:
 
 ```python
 from mteb.tasks import AmazonReviewsClassification, BUCCBitextMining
 
 evaluation = MTEB(tasks=[
-        AmazonReviewsClassification(langs=["en", "fr"]) # Only load "en" and "fr" subsets of Amazon Reviews
-        BUCCBitextMining(langs=["de-en"]), # Only load "de-en" subset of BUCC
+        AmazonReviewsClassification(hf_subsets=["en", "fr"]) # Only load "en" and "fr" subsets of Amazon Reviews
+        BUCCBitextMining(hf_subsets=["de-en"]), # Only load "de-en" subset of BUCC
```
Muennighoff, line 110: Sorry if I'm missing something obvious, but why is this change needed? 🤔 Can these values ever not be languages?
(changing this breaks existing MTEB scripts)

README.md
```diff
@@ -96,17 +96,18 @@ evaluation = MTEB(task_categories=['S2S']) # Only select sentence2sentence datas
 * by their languages
 
 ```python
+import mteb
 evaluation = MTEB(task_langs=["en", "de"]) # Only select datasets which are "en", "de" or "en-de"
```
Muennighoff, line 100: If you want to add an import here, it should probably be `from mteb import MTEB`, no? But I don't think an import here is needed as it's very clear from the script under `## Usage` I think 🤔

mteb/__init__.py
```diff
@@ -18,4 +18,5 @@
     "MTEB_RETRIEVAL_LAW",
     "MTEB_RETRIEVAL_WITH_INSTRUCTIONS",
     "get_tasks",
+    "TASKS_REGISTRY",
 ]
```
Muennighoff, line 22: Maybe this should be kept alphabetic? (i.e. caps comes before lowercase)

mteb/abstasks/languages.py
```diff
@@ -2,7 +2,10 @@
 Script codes (ISO 15924) obtained from: https://unicode.org/iso15924/iso15924.txt
 """
 
+from __future__ import annotations
```
Muennighoff, line 5: Why is this needed? We don't support Python 2, no?

mteb/abstasks/languages.py
```diff
@@ -2,7 +2,10 @@
...
+
 import json
+from dataclasses import dataclass
 from pathlib import Path
```
Muennighoff, line 9:
```suggestion
from dataclasses import dataclass
import json
from pathlib import Path
```

mteb/abstasks/mteb_result.py
isaac-chung: If the idea is to make it simple for users to specify a metric without custom syntax, could we accept the metric directly instead, and do all the "custom syntax" ourselves? i.e. no getter, no dots. 

Results have this structure. `main_score` is at the same level as other metrics.
```
"aau_Latn-eng_Latn": {
      "accuracy": 0.0703125,
      "f1": 0.05066417191876751,
      "main_score": 0.05066417191876751,
      "precision": 0.044970703125,
      "recall": 0.0703125
    },
```

Option here is for `get_score` to accept a string and defaults to `"main_score"`. That way we don't even need `get_main_score`. Users can directly call `get_score`. Under the hoop (in the method), we handle the logic to extract the specified metric in the structure.
```
metric: str = "main_score"
```

mteb/abstasks/AbsTaskClassification.py
KennethEnevoldsen: It is done for all scores

tests/test_langscripts.py
KennethEnevoldsen: dict(languages=["fra"], scripts=None)
does not contain latn.

It would be impossible to know without having a list of languages and corresponding scripts.

README.md
KennethEnevoldsen: Yea I believe in oscar there is e.g. filtered_dan and unfiltered_dan. They are however often languages, but can be anything.

mteb/abstasks/languages.py
KennethEnevoldsen: it is for the type annotation e.g. list[str] vs List[str]

mteb/abstasks/languages.py
KennethEnevoldsen: I am not sure I get this? Imports are auto-sorted by ruff

tests/test_langscripts.py
isaac-chung: I see, thanks.

mteb/abstasks/languages.py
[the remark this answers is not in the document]
KennethEnevoldsen: fixed

mteb/abstasks/mteb_result.py
KennethEnevoldsen: fixed

README.md
KennethEnevoldsen: fixed

mteb/__init__.py
KennethEnevoldsen: fixed

mteb/abstasks/mteb_result.py
KennethEnevoldsen: I have now refactored out get_main_score such that it is:

```
def get_score(
        ...
        getter: Callable[[Scores], float] = lambda scores: scores["main_score"],
        aggregation: Callable[[list[float]], Any] = np.mean,
    )

...
mteb_results.get_score() # will get main score 
```

mteb/abstasks/mteb_result.py
KennethEnevoldsen: Now uses a list

KennethEnevoldsen:
I believe I have now addressed most concerns, feel free to re-open case threads if I closed them pre-maturely. Otherwise, I feel like this is ready to merge and then I will start the second stage as well.

mteb/abstasks/languages.py
Muennighoff: Oh sorry my bad!

mteb/abstasks/mteb_result.py
isaac-chung: I really like the aggregation option. My only last suggestion is to make the `get_score` signature simpler (sorry if it wasn't clear before):
```diff
def get_score(
        ...
-       getter: Callable[[Scores], float] = lambda scores: scores["main_score"],
+       metric: str = "main_score",
        aggregation: Callable[[list[float]], Any] = np.mean,
    )
```

then in the method, we use `lambda scores: scores[metric]`. That way the user will only need to give us the metric they want, and not a function.

isaac-chung (approved):
In general this looks good to me, thanks for iterating! I only have one final suggestion on `get_scores`. 

Also how would we do points here?

x-tabdeveloping (approved):

Muennighoff:
Do we need to change all the result files in https://huggingface.co/datasets/mteb/results ?

KennethEnevoldsen:
@Muennighoff we can implement a conversion script?

```
MTEBResults.load_legacy_results(path)
```

> Also how would we do points here?

I have added 6 points for the change and 2 points pr. reviewer. Does that seems reasonable?

(will merge this in to start working on the next step, will bring over points from here as well)

x-tabdeveloping:
@KennethEnevoldsen maybe we could add a version to the output format from now on? And then the absence of a version tag would indicate that it's legacy. That way if the protocol changes we can account for it.

PR #671, opened 2024-05-11
fix: ensure that task.languages return a list of languages

KennethEnevoldsen:
And a few minor docstring fixes and type hints.


isaac-chung (approved):
If I were the jury, I'd give you 12 points #EuroVision

PR #678, opened 2024-05-12
 fix: Add LegalBench datasets - 13

awinml:
## Checklist for adding MMTEB dataset

Reason for dataset addition: Showcases the model's performance on the [LegalBench](https://hazyresearch.stanford.edu/legalbench/) datasets, enabling evaluation on domain-specific legal texts.

Follow up to #665 

## Changes made:

The following 10 datasets have been added:

1. [MAUD General Economic and Financial Conditions Subject To Disproportionate Impact Modifier](https://hazyresearch.stanford.edu/legalbench/tasks/maud_general_economic_and_financial_conditions_subject_to_disproportionate_impact_modifier.html)
2. [MAUD Includes Consistent With Past Practice](https://hazyresearch.stanford.edu/legalbench/tasks/maud_includes_consistent_with_past_practice.html)
3. [MAUD Initial Matching Rights Period COR](https://hazyresearch.stanford.edu/legalbench/tasks/maud_initial_matching_rights_period_(cor).html)
4. [MAUD Initial Matching Rights Period FTR](https://hazyresearch.stanford.edu/legalbench/tasks/maud_initial_matching_rights_period_(ftr).html)
5. [MAUDInterveningEventRequiredToOccurAfterSigningAnswer](https://hazyresearch.stanford.edu/legalbench/tasks/maud_intervening_event_-_required_to_occur_after_signing_-_answer.html)
6. [MAUD Knowledge Definition](https://hazyresearch.stanford.edu/legalbench/tasks/maud_knowledge_definition.html)
7. [MAUDLiabilityStandardForNoShopBreachByTargetNonDORepresentatives](https://hazyresearch.stanford.edu/legalbench/tasks/maud_liability_standard_for_no-shop_breach_by_target_non-do_representatives.html)
8. [MAUD Ordinary Course Efforts Standard](https://hazyresearch.stanford.edu/legalbench/tasks/maud_ordinary_course_efforts_standard.html)
9. [MAUD Pandemic Or Other Public Health Event Subject To Disproportionate Impact Modifier](https://hazyresearch.stanford.edu/legalbench/tasks/maud_pandemic_or_other_public_health_event__subject_to_disproportionate_impact_modifier.html)
10. [MAUD Pandemic Or Other Public Health Event Specific Reference To Pandemic Related Governmental Responses Or Measures](https://hazyresearch.stanford.edu/legalbench/tasks/maud_pandemic_or_other_public_health_event_specific_reference_to_pandemic-related_governmental_responses_or_measures.html)

[...]

PR #665, opened 2024-05-11
 fix: Add LegalBench datasets - 12

awinml:
## Checklist for adding MMTEB dataset

Reason for dataset addition: Showcases the model's performance on the [LegalBench](https://hazyresearch.stanford.edu/legalbench/) datasets, enabling evaluation on domain-specific legal texts.

Follow up to #661 

## Changes Made:

The following 10 datasets have been added:

1. [MAUD COR Permitted With Board Fiduciary Determination Only](https://hazyresearch.stanford.edu/legalbench/tasks/maud_cor_permitted_with_board_fiduciary_determination_only.html)
2. [MAUD COR Standard Intervening Event](https://hazyresearch.stanford.edu/legalbench/tasks/maud_cor_standard_(intervening_event).html)
3. [MAUD COR Standard Superior Offer](https://hazyresearch.stanford.edu/legalbench/tasks/maud_cor_standard_(superior_offer).html)
4. [MAUD Definition Contains Knowledge Requirement Answer](https://hazyresearch.stanford.edu/legalbench/tasks/maud_definition_contains_knowledge_requirement_-_answer.html)
5. [MAUD Definition Includes Asset Deals](https://hazyresearch.stanford.edu/legalbench/tasks/maud_definition_includes_asset_deals.html)
6. [MAUD Definition Includes Stock Deals](https://hazyresearch.stanford.edu/legalbench/tasks/maud_definition_includes_stock_deals.html)
7. [MAUD Fiduciary Exception Board Determination Standard](https://hazyresearch.stanford.edu/legalbench/tasks/maud_fiduciary_exception__board_determination_standard.html)
8. [MAUD Fiduciary Exception Board Determination Trigger No Shop](https://hazyresearch.stanford.edu/legalbench/tasks/maud_fiduciary_exception_board_determination_trigger_(no_shop).html)
9. [MAUD Financial Point Of View Is The Sole Consideration](https://hazyresearch.stanford.edu/legalbench/tasks/maud_financial_point_of_view_is_the_sole_consideration.html)
10. [MAUD FLS MAE Standard](https://hazyresearch.stanford.edu/legalbench/tasks/maud_fls_(mae)_standard.html)

The datasets have been included in https://github.com/embeddings-benchmark/mteb/pull/680, only points for them are being added here.

-------------------------------------------------------------------------------------------------------------------------------------------------

- [x] I have tested that the dataset runs with the `mteb` package.
[...]

PR #661, opened 2024-05-10
 fix: Add LegalBench datasets - 11 

awinml:
## Checklist for adding MMTEB dataset

Reason for dataset addition: Showcases the model's performance on the [LegalBench](https://hazyresearch.stanford.edu/legalbench/) datasets, enabling evaluation on domain-specific legal texts.

Follow up to #657 

## Changes Made:

The following 10 datasets have been added:

1. [Ability To Consummate Concept Is Subject To MAE Carveouts](https://hazyresearch.stanford.edu/legalbench/tasks/maud_ability_to_consummate_concept_is_subject_to_mae_carveouts.html)
2. [Accuracy Of Fundamental Target RWS Bringdown Standard](https://hazyresearch.stanford.edu/legalbench/tasks/maud_accuracy_of_fundamental_target_rws_bringdown_standard.html)
3. [Accuracy Of Target Capitalization RW Outstanding Shares Bringdown Standard Answer](https://hazyresearch.stanford.edu/legalbench/tasks/maud_accuracy_of_target_capitalization_rw_(outstanding_shares)_bringdown_standard_answer.html)
4. [Accuracy Of Target General RW Bringdown Timing Answer](https://hazyresearch.stanford.edu/legalbench/tasks/maud_accuracy_of_target_general_rw_bringdown_timing_answer.html)
5. [Additional Matching Rights Period For Modifications Cor](https://hazyresearch.stanford.edu/legalbench/tasks/maud_additional_matching_rights_period_for_modifications_(cor).html)
6. [Application Of Buyer Consent Requirement Negative Interim Covenant](https://hazyresearch.stanford.edu/legalbench/tasks/maud_application_of_buyer_consent_requirement_(negative_interim_covenant).html)
7. [Buyer Consent Requirement Ordinary Course](https://hazyresearch.stanford.edu/legalbench/tasks/maud_buyer_consent_requirement_(ordinary_course).html)
8. [Change In Law Subject To Disproportionate Impact Modifier](https://hazyresearch.stanford.edu/legalbench/tasks/maud_change_in_law__subject_to_disproportionate_impact_modifier.html)
9. [Changes In GAAP Or Other Accounting Principles Subject To Disproportionate Impact Modifier](https://hazyresearch.stanford.edu/legalbench/tasks/maud_changes_in_gaap_or_other_accounting_principles__subject_to_disproportionate_impact_modifier.html)
[...]

PR #680, opened 2024-05-13
 fix: Add LegalBench datasets - 14

awinml:

Reason for dataset addition: Showcases the model's performance on the [LegalBench](https://hazyresearch.stanford.edu/legalbench/) datasets, enabling evaluation on domain-specific legal texts.

Follow up to #678 

## Changes Made:

The following 10 datasets have been added in this PR:

1. [MAUD Relational Language MAE Applies To](https://hazyresearch.stanford.edu/legalbench/tasks/maud_relational_language_(mae)_applies_to.html)
2. [MAUD Specific Performance](https://hazyresearch.stanford.edu/legalbench/tasks/maud_specific_performance.html)
3. [MAUD Tail Period Length](https://hazyresearch.stanford.edu/legalbench/tasks/maud_tail_period_length.html)
4. [MAUD Type Of Consideration](https://hazyresearch.stanford.edu/legalbench/tasks/maud_type_of_consideration.html)
5. [NYS Judicial Ethics](https://hazyresearch.stanford.edu/legalbench/tasks/nys_judicial_ethics.html)
6. [OPP115 Data Retention](https://hazyresearch.stanford.edu/legalbench/tasks/opp115_data_retention.html)
7. [OPP115 Data Security](https://hazyresearch.stanford.edu/legalbench/tasks/opp115_data_security.html)
8. [OPP115 Do Not Track](https://hazyresearch.stanford.edu/legalbench/tasks/opp115_do_not_track.html)
9. [OPP115 First Party Collection Use](https://hazyresearch.stanford.edu/legalbench/tasks/opp115_first_party_collection_use.html)
10. [OPP115 International And Specific Audiences](https://hazyresearch.stanford.edu/legalbench/tasks/opp115_international_and_specific_audiences.html)

------------------------------------------------------------------------------------------------------------------------------------


docs/mmteb/points/680.jsonl
```diff
@@ -0,0 +1 @@
+{"GitHub": "awinml", "New dataset": 20}
```
KennethEnevoldsen, line 1:
```suggestion
{"GitHub": "KennethEnevoldsen", "PR Review": 2}
```

mteb/tasks/Classification/eng/LegalBenchClassification.py
```diff
@@ -4561,3 +4563,560 @@ def dataset_transform(self):
...
+            primaryClass={cs.CL}
+        }
+        @article{wang2023maud,
+            title={MAUD: An Expert-Annotated Legal NLP Dataset for Merger Agreement Understanding},
+            author={Wang, Steven H and Scardigli, Antoine and Tang, Leonard and Chen, Wei and Levkin, Dimitry and Chen, Anya and Ball, Spencer and Woodside, Thomas and Zhang, Oliver and Hendrycks, Dan},
+            journal={arXiv preprint arXiv:2301.00876},
+            year={2023}
+        }
+        """,
+        n_samples={"test": 82},
```
KennethEnevoldsen, line 4608: Unsure if this is large enough to capture a meaningful difference between models. WIll you run it using e5 base to check?

mteb/tasks/Classification/eng/LegalBenchClassification.py
```diff
@@ -4561,3 +4563,560 @@ def dataset_transform(self):
...
+            primaryClass={cs.CL}
+        }
+        @article{wang2023maud,
+            title={MAUD: An Expert-Annotated Legal NLP Dataset for Merger Agreement Understanding},
+            author={Wang, Steven H and Scardigli, Antoine and Tang, Leonard and Chen, Wei and Levkin, Dimitry and Chen, Anya and Ball, Spencer and Woodside, Thomas and Zhang, Oliver and Hendrycks, Dan},
+            journal={arXiv preprint arXiv:2301.00876},
+            year={2023}
+        }
+        """,
+        n_samples={"test": 162},
```
KennethEnevoldsen, line 4664: same here as well - scores seems very similar (potentially hitting a cealing effect in this case)

mteb/tasks/Classification/eng/LegalBenchClassification.py
```diff
@@ -4561,3 +4563,560 @@ def dataset_transform(self):
...
+            primaryClass={cs.CL}
+        }
+        @article{wang2023maud,
+            title={MAUD: An Expert-Annotated Legal NLP Dataset for Merger Agreement Understanding},
+            author={Wang, Steven H and Scardigli, Antoine and Tang, Leonard and Chen, Wei and Levkin, Dimitry and Chen, Anya and Ball, Spencer and Woodside, Thomas and Zhang, Oliver and Hendrycks, Dan},
+            journal={arXiv preprint arXiv:2301.00876},
+            year={2023}
+        }
+        """,
+        n_samples={"test": 159},
```
KennethEnevoldsen, line 4720: Also seems to be the case here. It might be relevant to run a larger model on all these tasks

KennethEnevoldsen (commented):
Looks good. Some concern about size and whether the tasks meaningfully differentiate between models in a few cases. Will you run the e5-base as well?

mteb/tasks/Classification/eng/LegalBenchClassification.py
awinml: I think you are right. The scores are identical with the e5-base model for all the three datasets.

We could combine these datasets, since they are all sampled from the dataset (MAUD). The labels are different, so we would need to remap them to different values.  What do you suggest we do?

KennethEnevoldsen: Hmm unsure if remapping would make a difference here, but it is worth a try if you think it might help.

Otherwise, I would probably either remove them or we might consider including them as is (for completeness), but not include them in an eventual legal benchmark).

@imenelydiaker what is your thought here?

awinml: I have combined all the 34 MAUD datasets into one task. The results are not identical anymore, but not very different either between the models. Have included the results with e5-base as well. 

The MAUD dataset is a tough dataset to perform on, since it originally had over 47,000 labels which were broken down into these 34 subtasks by the LegalBench authors. I think we should include it, as it is a significant part of LegalBench. 

This change includes all the datasets being added in #678, #665 and #661. The points for them have not been added here, we can use the original PR to add the points.

awinml: Unrelated to the above discussion: I would like to highlight that [`test_load_data`](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_all_abstasks.py#L22) fails when `load_data` is modified and the global import of `datasets` is not used. That is, using `datasets.load_dataset()` and `datasets.concatenate_datasets()` instead of just `load_dataset()` and `concatenate_datasets()`.  This is because the whole module has been mocked instead of just `load_dataset()`.

For that reason, I have included both the global and module level imports (please see bb59b5b4323d5d6926a80ff541ed3e923eeefb84) in `LegalBenchClassification.py` . I will fix this and just include the global import for all the datasets in the file, but would prefer to do that in a separate PR so its easier to review.

KennethEnevoldsen: Thanks for the clarification @awinml, yea the mock test does cause some of issues. Feel free to create an issue for it, however it will probably not be our first priority (plenty of new additions atm.)

awinml: Thanks! I will open an issue for the mock test.

What did you think of the combination of MAUD above? This PR should be good to merge otherwise. 
I will proceed to remove the datasets from https://github.com/embeddings-benchmark/mteb/pull/678, https://github.com/embeddings-benchmark/mteb/pull/665 and https://github.com/embeddings-benchmark/mteb/pull/661 and only keep the points there, after your confirmation.

KennethEnevoldsen: I think the combination is a great idea. Solves a lot of the issues I had with the scores while retaining a large part of the dataset.

awinml: That's great, thanks! I'll proceed to clean up the other PRs now.

KennethEnevoldsen:
@awinml is this ready to merge?

awinml:
@KennethEnevoldsen Yes, this is ready to merge. I have already added points for the review.

KennethEnevoldsen:
Was my suspicion as well. Just wanted to check that I didn't miss anything - I have enabled automerge

PR #700, opened 2024-05-14
chore: Fix command in PR template

mrshu:
* The new version of `mteb` doesn't expect the `run` parameter, yielding a command that causes an erorr. This commit fixes that issue.


isaac-chung (approved):
Thanks!

PR #712, opened 2024-05-14
feat: Add carbon emissions estimation

imenelydiaker:
Motivated by this discussion #703, I added this code to compute carbon emissions for each evaluation, the final estimation for a split is added to the results.json file of the task.

I'd appreciate it if someone runs an evaluation and gives me feedback on the logging level, I can switch it to debug if it's too much.

Was thinking about making the carbon tracking optional with an argument to turn on/off by users if they'd like to track it or not. Would love some thoughts on this.


Muennighoff:
Nice! I'd make it optional because I think it slows down evaluation? Maybe you can run a quick speed comparison. 🙌

imenelydiaker:
> Nice! I'd make it optional because I think it slows down evaluation? Maybe you can run a quick speed comparison. 🙌

It doesn't seem like it's slowing the evaluations a lot. See left without CO2, right with CO2:

image

But I agree to make it optional. Thanks!

mteb/evaluation/MTEB.py
```diff
@@ -11,6 +11,7 @@
 from typing import List, Union
 
 import datasets
+from codecarbon import EmissionsTracker
```
Muennighoff, line 14: Should we make the import optional, too?
I.e. moving `from codecarbon import EmissionsTracker` below `if co2_tracker:`?
So people aren't forced to install the package

imenelydiaker: True, updating thanks!

Muennighoff (approved):

mteb/evaluation/MTEB.py
imenelydiaker: I think I should also update the pyptoject.toml, I initially put it in the dependencies so it gets installed with mteb. I don't know where it should go 🤔 

I don't think it should go here since these are dev dependencies: 
```
dev = ["ruff>=0.0.254", "pytest", "pytest-xdist"]
```

Muennighoff: Oh good point; if you want you can add e.g.
`codecarbon = ["codecarbon"]`
so it can be installed via mteb[codecarbon]
or you can also just remove the package requirement and just in the code enclose the `import codecarbon` in a try & except and if it excepts raise an Exception and ask the user to install codecarbon. I like that second option most I think!

imenelydiaker:
I missed the points, I'm adding them

docs/mmteb/points/712.jsonl
```diff
@@ -0,0 +1,2 @@
+{"GitHub": "imenelydiaker", "Bug fixes": 2}
+{"GitHub": "Muennighoff", "Review PR": 2}
```
Muennighoff, line 2: You should get more points than me on this PR 😁

imenelydiaker: Okay, how many do you suggest? 😆

mteb/evaluation/MTEB.py
imenelydiaker: Okay looks great, I'll do that

docs/mmteb/points/712.jsonl
Muennighoff: Not sure but maybe at least double?

imenelydiaker: Okay, I'll add 4, seems fair 🙂

mteb/evaluation/MTEB.py
imenelydiaker: Well, I liked both options, is it okay to keep them both?

Muennighoff: Seems fine! Feel free to merge :)

PR #737, opened 2024-05-15
Added new dataset: XStance pair classification

malteos:
A Multilingual Multi-Target Dataset for Stance Detection in French, German, and Italian (pair classification).

Reason for dataset addition: the first stance detection dataset in French, German, and Italian


mteb/tasks/PairClassification/multilingual/XStance.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    def dataset_transform(self):
+        max_n_samples = 2048
+        _dataset = {}
+        for lang in self.hf_subsets:
+            _dataset[lang] = {}
+            for split in self.metadata.eval_splits:
+                # hf config does not support selection by language, use manual filter instead
+                hf_dataset = self.dataset["de"][split].filter(
```
KranthiGV, line 58: I think this should be `self.dataset[lang][split].....` instead of `self.dataset["de"][split]......`

KranthiGV (approved):
Once the hardcoded "de" bug is fixed, this PR LGTM

mteb/tasks/PairClassification/multilingual/XStance.py
malteos: The language selection does not work with the HF config. That's why I have the Lang hard coded and do the language filtering manually. It's an issue with the HF dataset implementation. See the comment above.

wissam-sib: An alternative, for this to look less strange for an outside reader, would be to overwrite the load data (it's often the case, see OpusparcusPC for example) and there load the dataset and assigning the languages properly:
```
def load_data(self, **kwargs):
    """Load dataset from HuggingFace hub"""
    if self.data_loaded:
        return
    self.dataset = {}
    path = self.metadata_dict["dataset"]["path"]
    revision = self.metadata_dict["dataset"]["revision"]
    raw_dataset = datasets.load_dataset(path, revision=revision)
    for split in self.metadata_dict["eval_splits"]:
        self.dataset[split] = {}
        for lang in self.metadata.eval_langs:
            self.dataset[split][lang] = raw_dataset[split].filter(
                lambda row: row["language"] == lang
            )
    self.dataset_transform(lang) # and in your dataset-transform function you can do whatever ops that are missing
    self.data_loaded = True
```

mteb/tasks/PairClassification/multilingual/XStance.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import annotations
+
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+from ....abstasks import MultilingualTask
+from ....abstasks.AbsTaskPairClassification import AbsTaskPairClassification
+
+
```
Ruqyai, line 8: `N_SAMPLES = 2048` 

I prefer adding here 

then  just pass it to

`     n_samples={"test": N_SAMPLES},
`

` def dataset_transform(self):
        max_n_samples = N_SAMPLES`

KennethEnevoldsen:
@KranthiGV or @malteos does any of you have the time to finish up this PR?

malteos:
I will do it later today! Sorry for the delay.

malteos:
I fixed the hard-coded language in the data loader and added the points.

KennethEnevoldsen:
Fixed tests in #795

Issue #747, opened 2024-05-16
MindSmallReranking dataset errors out due to download/parsing error?


w601sxs:
Tested with the latest mteb on pypi from 4 hours ago as well as installation from latest git main branch:

```
## Evaluating 1 tasks:
───────────────────────────────────────────────── Selected tasks  ─────────────────────────────────────────────────
Reranking
    - MindSmallReranking, s2s

INFO:mteb.evaluation.MTEB:

********************** Evaluating MindSmallReranking **********************
INFO:mteb.evaluation.MTEB:Loading dataset for MindSmallReranking
Repo card metadata block was not found. Setting CardData to empty.
WARNING:huggingface_hub.repocard:Repo card metadata block was not found. Setting CardData to empty.
Failed to read file 'gzip://7a742da40ba0425a72301598ce27d63296c468da48cd98c4ae479b1d88a755a8::/root/.cache/huggingface/datasets/downloads/7a742da40ba0425a72301598ce27d63296c468da48cd98c4ae479b1d88a755a8' with error <class 'pyarrow.lib.ArrowInvalid'>: JSON parse error: Invalid value. in row 0
ERROR:datasets.packaged_modules.json.json:Failed to read file 'gzip://7a742da40ba0425a72301598ce27d63296c468da48cd98c4ae479b1d88a755a8::/root/.cache/huggingface/datasets/downloads/7a742da40ba0425a72301598ce27d63296c468da48cd98c4ae479b1d88a755a8' with error <class 'pyarrow.lib.ArrowInvalid'>: JSON parse error: Invalid value. in row 0
ERROR:mteb.evaluation.MTEB:Error while evaluating MindSmallReranking: An error occurred while generating the dataset
---------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
File /opt/conda/lib/python3.10/site-packages/datasets/packaged_modules/json/json.py:145, in Json._generate_tables(self, files)
    142     with open(
    143         file, encoding=self.config.encoding, errors=self.config.encoding_errors
    144     ) as f:
--> 145         dataset = json.load(f)
    146 except json.JSONDecodeError:

File /opt/conda/lib/python3.10/json/__init__.py:293, in load(fp, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    276 """Deserialize ``fp`` (a ``.read()``-supporting file-like object containing
    277 a JSON document) to a Python object.
    278 
   (...)
[...]

PR #748, opened 2024-05-16
fix: MindSmallReranking not loading

loicmagne:
Following https://github.com/embeddings-benchmark/mteb/issues/747

Converting .gz to .zip fixed the issue


imenelydiaker (approved):
LGTM! Thank you for fixing this

PR #758, opened 2024-05-17
fix: Metadata for Polish STS

mrshu:
* The Polish STS had missing metadata. This commit should fix the situation.

### DIFF
--- a/docs/mmteb/points/758.jsonl
+++ b/docs/mmteb/points/758.jsonl
@@ -1,0 +1,2 @@
+{"GitHub": "isaac-chung", "Review PR": 2}
+{"GitHub": "mrshu", "Bug fixes": 2}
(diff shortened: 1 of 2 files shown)


isaac-chung (approved):
Thanks for adding this! Feel free to add points.

KennethEnevoldsen:
@mrshu will you add the points and make sure that the tests pass as well?

mrshu:
Thanks @KennethEnevoldsen and @isaac-chung -- I believe it should be done.

docs/mmteb/points/758.jsonl
```diff
@@ -0,0 +1,2 @@
+{"GitHub": "isaac-chung", "Review PR": 2}
+{"GitHub": "mrshu", "Bug fixes": 2}
```
isaac-chung, line 2: Filling metadata is closer to dataset annotations:
```
"Dataset annotations": 1, # 1 point for each full dataset annotation
```

PR #760, opened 2024-05-17
Added multilabel stratification to AbsTaskMultilabelClassification

dokato:
This is continuation of the discussion from https://github.com/embeddings-benchmark/mteb/discussions/698. 

cc @x-tabdeveloping

### DIFF
--- a/docs/mmteb/points/760.jsonl
+++ b/docs/mmteb/points/760.jsonl
@@ -1,0 +1,3 @@
+{"GitHub": "dokato", "Bug fixes": 2}
+{"GitHub": "dokato", "New dataset": 6}
+{"GitHub": "x-tabdeveloping", "Review PR": 2}
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -8,10 +8,10 @@ from typing import Any, Sequence
 import datasets
 import numpy as np
 import torch
-from datasets import DatasetDict
-from sklearn.model_selection import train_test_split
+from datasets import Dataset, DatasetDict
 from sklearn.preprocessing import MultiLabelBinarizer
 
+from mteb.abstasks.stratification import _iterative_train_test_split
 from mteb.abstasks.TaskMetadata import TaskMetadata
 from mteb.encoder_interface import Encoder, EncoderWithQueryCorpusEncode
 from mteb.languages import LanguageScripts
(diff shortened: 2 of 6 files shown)


mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -36,6 +38,20 @@ def evaluate_classifier(
...
+    label_set = sorted(label_set)
+
+    multihot_vectors = []
+    for label_list in labels:
+        multihot_vectors.append([1 if x in label_list else 0 for x in label_set])
+
+    return np.array(multihot_vectors)
+
+
 class AbsTaskMultilabelClassification(AbsTask):
```
x-tabdeveloping, line 55: The problem is that this behaviour is not only needed in tasks that always have multiple labels. I implemented this behaviour in AbsTask originally as a backoff for when single label stratification fails, I think we should probably go for that approach.

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "typing_extensions",
     "eval_type_backport",
     "polars>=0.20.22",
+    "skmultilearn",
```
x-tabdeveloping, line 43: Have you had a look at how this iterative subsampling function works in skmultilearn? If we can implement it ourselves, I would probably prefer that over introducing a new dependency. If it would be too complex to implement in MTEB we can of course roll with sklmultilearn.

x-tabdeveloping (commented):
See my comments. Let's continue discussion on this, I think we are going in the right direction but this still needs some thought and work.

mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -36,6 +38,20 @@ def evaluate_classifier(
     return scores
 
 
+def _multihot_encoder(labels):
```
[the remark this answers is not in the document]
dokato: yup, removed

mteb/abstasks/AbsTaskMultilabelClassification.py
dokato: I don't quite get it why we should even try single label stratification on a multilabel problem?

pyproject.toml
dokato: Yea, I had a quick look. It's not terribly complicated but there's a few non-trivial things going on, see:
- https://github.com/scikit-multilearn/scikit-multilearn/blob/master/skmultilearn/model_selection/iterative_stratification.py#L175
- ["A Network Perspective on Stratification of Multi-Label Data", Szymański & Kajdanowicz, 2017](http://proceedings.mlr.press/v74/szyma%C5%84ski17a/szyma%C5%84ski17a.pdf)

KennethEnevoldsen:
hi @dokato and @x-tabdeveloping: I would probably just do the split when creating the dataset on HF. This avoids the extra dependency.

x-tabdeveloping:
Yeah but then we need to reupload a lot of datasets, and I'm wondering if it's too much of a hustle. I am really in favour of not introducing new dependencies if possible.
I'm wondering if it would make sense to just copy the file over from sklmultilearn, simplify it a bit, and call it a day.
It seems pretty self-contained to me at least. What do you think @dokato @KennethEnevoldsen ?

KennethEnevoldsen:
@x-tabdeveloping I believe that is the best solution. We just need to add a reference

dokato:
Thanks for advise @x-tabdeveloping @KennethEnevoldsen. That's what I did, I moved their function (with acknowledgments) and slightly modified so it returns indices `_iterative_train_test_split`.

mteb/abstasks/AbsTaskMultilabelClassification.py
dokato: Cool, thanks for elaborating. I see your points. If useful in other tasks, it should definately be in AbsTask. Though, the place I don't like are those conditions:

```
            try:
                dataset_dict = dataset_dict.class_encode_column(label)
            except ValueError as e:
                if isinstance(dataset_dict[splits[0]][label][0], Sequence):
                    ...
```
IMHO we should just recognise by task type and apply stratified split where it really makes sense, instead of relying on try/except.

x-tabdeveloping: We can't do it, cause ClusteringFast is both hierarchical and flat unfortunately.

x-tabdeveloping:
Sorry for the delay @dokato I'm just mid exam-season, I'll have a look at it right now :D

x-tabdeveloping (commented):
Beautiful! Looks absolutely good to me. Can I ask you to implement it for one or more tasks and see if it works? Thanks for your awesome work and patience

dokato:
Thanks for review @x-tabdeveloping! I rerun Brazilian dataset as all other had small enough "test" splits that didn't require that much of stratification. Hence, I added additionally `KorHateSpeechMLClassification` (https://huggingface.co/datasets/jeanlee/kmhas_korean_hate_speech) also to improve representation of MultiLabel Classification task.

Here's dataset card for `KorHateSpeechMLClassification`.

dokato:
I've changed "Stratified subsampling of test set to 2000 examples." bit to more commonly used in classification 2048 `AbsTaskMultilabelClassification.py`. But now, given that we have proper stratificaiton I wonder if we need that bit at all, or should we just assume that the sampling is done in the `dataset_transform` method? Similarly as is in other tasks. Maybe just a warning if the split if bigger than 2048 would suffice?

dokato:
@x-tabdeveloping struggling to understand what might be going on here. I tried to rebase but it had too many merge conflicts, so I forced push it. But honestly, have no idea why those tests are failing, esp. that they related to some other datasets...

x-tabdeveloping:
Well, since this PR is making changes to quite fundamental things in the library, I would prefer if everything was passing before we went on, and all conflicts were resolved. I think if nothing fixes it probably the best thing to do is just reapply your changes to the current main in a new PR or a new branch, it seems like the most painless option to me. I don't have the slightest cue either what might have gone wrong here unfortunately.

KennethEnevoldsen:
@dokato pulling from main should resolve a lot of the dataset issues.

dokato:
Good shoutout guys, I rebased again. I guess it's ready?

x-tabdeveloping:
I think it looks alright! Thanks for the work and patience @dokato :D

dokato:
No worries, pleasure! hope you're exams went fine ;)

PR #761, opened 2024-05-17
Quick fix CI/CD issue due to dataset redirect

wissam-sib:
You might have noticed that tests have been failing since earlier today. This is due to a dataset path that has been redirected. Changing the path fixes the issue. We may want to have better automated fix in the future.


isaac-chung (approved):
Thanks for fixing this!

wissam-sib:
Adding a screen for reference:
image

PR #759, opened 2024-05-17
Standardized results output

KennethEnevoldsen:
This pull request standardizes the output format of the results in the code. It includes changes to multiple files to ensure consistent formatting and structure of the results.

PR #770, opened 2024-05-20
fix: Added option to load historic result files

KennethEnevoldsen:
- Added test to ensure that a wide variety of old file formats are supported (I tested it using all the files in the results folder, but reduced it to >50 to not have it take ages)
- Added test to ensure that superseeded datasets exists (found one case)
- Added missing imports to a variety of datasets

Extends upon #759

@tomaarsen just making sure that you know that the results format will change (it can all be loaded using `MTEBResults.from_disk(path, load_historic_data=True)`. @Muennighoff is there anyone else who should be made aware of this change?


mteb/abstasks/MTEBResults.py
```diff
@@ -138,9 +142,94 @@ def to_disk(self, path: Path) -> None:
...
+            except Exception as e:
+                raise ValueError(
+                    f"Error loading MTEBResults from disk. You can try to load historic data by setting `load_historic_data=True`. Error: {e}"
+                )
+
+        err: str = ""
+        try:
+            return cls.model_validate(data)
+        except Exception as e:
+            err = f"Error loading MTEBResults from disk: {e}"
```
Muennighoff, line 167: Rather than raising a ValueError, why don't we just output a warning & directly try to load as historic data if it fails?

mteb/abstasks/MTEBResults.py
```diff
@@ -138,9 +142,94 @@ def to_disk(self, path: Path) -> None:
...
+        try:
+            if "mteb_version" in data and Version(data["mteb_version"]) < Version(
+                "1.11.0"
+            ):
+                logger.info(
+                    f"Could not load MTEBResults from disk, got error: {err}. Attempting to load from disk using format from before v1.10.0"
```
Muennighoff, line 173: Shouldn't this be `from before v1.11.0`?

mteb/abstasks/MTEBResults.py
```diff
@@ -138,9 +142,94 @@ def to_disk(self, path: Path) -> None:
...
+
+        dataset_revision = scores.pop("dataset_revision")
+        task_name = scores.pop("mteb_dataset_name")
+        mteb_version = scores.pop("mteb_version")
+
+        # calculate evaluation time across all splits (move to top level)
+        evaluation_time = 0
+        for split, split_score in scores.items():
+            if "evaluation_time" in split_score:
+                evaluation_time += split_score.pop("evaluation_time")
```
Muennighoff, line 194: Why are we doing this? Isn't moving this to the top level a loss of information? 🤔 Maybe having a `total_evaluation_time` at the top level or sth if it is important to have?

tests/historic_results/ArEntail.json
```diff
@@ -0,0 +1,49 @@
+{
```
Muennighoff, line 1: Which model were these run with? Maybe it makes sense to put them in a folder with the model name?

Muennighoff (approved):
Looks great; just minor notes 🙌

mteb/abstasks/MTEBResults.py
KennethEnevoldsen: Yea was considering whether load_historic_data should default be True or False. I can set it to True by default.

mteb/abstasks/MTEBResults.py
KennethEnevoldsen: It is. It is a decision made in the PR #759, I believe it should either be at the top-level or at the hf_subset level (so one level deeper than what it is now), however, that requires a larger refactor. During it this way we can always add it later.

having it somewhere in between creates an inconsistent structure which will require quite a few exceptions and would lead to errors is you have a huggingface subset called "evaluation_time" (which would be odd)

tests/historic_results/ArEntail.json
KennethEnevoldsen: randomly selected from the results folder so multiple (I believe most is e5-mult-small).

mteb/abstasks/MTEBResults.py
Muennighoff: Okay I see; I guess top level only is fine then

KennethEnevoldsen: I would love to add the subset level as well though (totally agree with your point that it is a loss of information)

mteb/abstasks/MTEBResults.py
KennethEnevoldsen: fixed

mteb/abstasks/MTEBResults.py
KennethEnevoldsen: Set default to True

tomaarsen:
Thanks for letting me know. I don't have time to review this in depth, but I'm fine with a format change in theory.

- Tom Aarsen

PR #775, opened 2024-05-21
Remove v_measures from metadata

Muennighoff:
Same as here: https://huggingface.co/datasets/mteb/results/blob/c20da760a46a2ffed061bbb27c49c314d383c45a/results.py#L29

Causes the below otherwise (from https://huggingface.co/raghavlight/SE_v1)

Screenshot 2024-05-20 at 6 48 57 PM


KennethEnevoldsen (approved):

KennethEnevoldsen:
@Muennighoff I suppose this file will have to be updated for the new results format as well?

Muennighoff:
> @Muennighoff I suppose this file will have to be updated for the new results format as well? It also seems to have a lot of manual exceptions (e.g. validation and dev cases). Any reason we wouldn't just add all splits?

Yes good point we need to update it. We can also add all splits, just need to make sure it works fine when pasting it in the metadata README like e.g. [here](https://huggingface.co/GritLM/GritLM-7B/edit/main/README.md)

PR #781, opened 2024-05-22
fix: Added metadata to a bunch of datasets

KennethEnevoldsen:


isaac-chung (approved):
🔥🔥🔥

mteb/tasks/Classification/ara/RestaurantReviewSentimentClassification.py
```diff
@@ -24,7 +24,7 @@ class RestaurantReviewSentimentClassification(AbsTaskClassification):
         form=["written"],
         domains=["Reviews"],
         task_subtypes=["Sentiment/Hate speech"],
-        license=None,
+        license="None specified",
```
isaac-chung, line 27: Nit: We should probably make these consistent - `Not specified` maybe?

PR #789, opened 2024-05-22
fix: select language in multilingual tasks

imenelydiaker:
Fix for bug of selecting a langugae in multilingual tasks.

Joint work with @isaac-chung 🚀


KennethEnevoldsen:
Good catch!

PR #718, opened 2024-05-15
GPT4-o generated queries for 14 languages

rasdani:
## Checklist for adding MMTEB dataset

Reason for dataset addition:
Succinct queries generated by a strong multilingual LLM grounded in Wikipedia articles nicely chunked by Cohere should be a strict improvement over a lot of machine translated versions of SQuAD in different languages.
Wikipedia is probably the highest quality (available) corpus in most languages.
see #378 

- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [ ] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [x] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 
- [ ] I have added points for my submission to the [points folder](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) using the PR number as the filename (e.g. `438.jsonl`).

WIP and I am running query generation over night for the remaining 12 languages on this list:

```
LANG_MAP = {
    "de": "German",
    "bn": "Bengali",
    "it": "Italian",
    "pt": "Portuguese",
    "ru": "Russian",
    "uk": "Ukrainian",
    "nl": "Dutch",
    "cs": "Czech",
    "ro": "Romanian",
    "bg": "Bulgarian",
    "sr": "Serbian",
    "fi": "Finnish",
    "fa": "Persian",
    "hi": "Hindi",
}
```

Draft PR for early feedback. @KennethEnevoldsen @Muennighoff  happy to hear any suggestions :)

PR #796, opened 2024-05-23
Add Norwegian and Swedish to WikipediaRerankingMultilingual and update points

rasdani:
Regarding #718 

I forgot to add Norwegian and Swedish to `WikipediaRerankingMultilingual` as well.
And points need to be updated for [Norwegian, Swedish] x [WikipediaRetrievalMultilingual, WikipediaRerankingMultilingual]

Swedish and Norwegian aren't new in retrieval and we already got 2 points for the `ellamind/wikipedia-2023-11-retrieval-multilingual-queries` dataset in #718, so no new points here. (I assume `-queries`, `-corpus`, `-qrels` don't count as separate datasets.)

In reranking, however, they are new, so 2 * 4 = 8.

I updated the points, such that we should end up next to each other on the paper.
Let me know, if this is correct or did we already hit some limit?


results/intfloat__multilingual-e5-small/WikipediaRerankingMultilingual.json
KennethEnevoldsen, line 1: Seems like there is some merge conflicts here

results/sentence-transformers__paraphrase-multilingual-MiniLM-L12-v2/WikipediaRerankingMultilingual.json
KennethEnevoldsen, line 1: also here

KennethEnevoldsen (approved):

KennethEnevoldsen:
I think this is fine - thanks or adding it in. The test currently fail. An update from main once #803 merges should fix it

PR #806, opened 2024-05-24
replace get_tasks as default filtering.

KennethEnevoldsen:
The intention here is to:

1) move complexity away from the MTEB object
2) ensure that the filters are applied in the same way across the benchmark (currently MTEB filters slightly differently due to not handling the new language codes)
3) deprecate filtering in MTEB going forward (only with a warning atm.)
4) doing it in a two step fashion ensure that users are able to inspect the tasks before they are run (also allow for much more custom filtering on the user end)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ tasks = mteb.get_tasks(categories=["s2s", "p2p"]) # Only select sentence2sentenc
 * by their languages
 
 ```python
-tasks = mteb.get_tasks(languages=["eng", "deu"]) # Only select datasets which contain "eng" or "deu"
+tasks = mteb.get_tasks(languages=["eng", "deu"]) # Only select datasets which contain "eng" or "deu" (iso 639-3 codes)
 ```
 
 You can also specify which languages to load for multilingual/cross-lingual tasks like below:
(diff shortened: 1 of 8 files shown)


isaac-chung:
Could you please get the tests passing first? I have no issues with the intentions described in the PR description.

isaac-chung (approved):
Nice! Looks good.

mteb/evaluation/MTEB.py
```diff
@@ -51,9 +51,12 @@ def __init__(
             tasks: List of tasks to be evaluated. If specified, we filter tasks based on `task_langs` only
             version: Version of the benchmark to use. If None, latest is used
             err_logs_path: Path to save error logs
-            co2_tracker: Whether to enable or disable CO2 emissions tracker
```
imenelydiaker, line 54: Why this was removed? Will the co2_tracker be always enabled?

README.md
```diff
@@ -41,7 +41,7 @@ pip install mteb
 * Using a python script (see [scripts/run_mteb_english.py](https://github.com/embeddings-benchmark/mteb/blob/main/scripts/run_mteb_english.py) and [mteb/mtebscripts](https://github.com/embeddings-benchmark/mtebscripts) for more):
 
 ```python
-from mteb import MTEB
+import mteb
```
imenelydiaker, line 44: The CLI (in `cmd.py`) uses the same process to run tasks, I think it needs to be updated too

mteb/evaluation/MTEB.py
```diff
@@ -76,6 +79,36 @@ def __init__(
...
+    ):
+        if task_types is not None:
+            logger.warning(
+                "The `task_types` argument is deprecated and will be removed in the next release. "
+                + "Please use the `tasks = mteb.get_tasks(... task_types = [...])` to filter tasks instead."
+            )
+        if task_categories is not None:
+            logger.warning(
+                "The `task_categories` argument is deprecated and will be removed in the next release. "
+                + "Please use the `tasks = mteb.get_tasks(... categories = [...])` to filter tasks instead."
```
Muennighoff, line 93:
```suggestion
                + "Please use `tasks = mteb.get_tasks(... categories = [...])` to filter tasks instead."
```

README.md
KennethEnevoldsen: I have created another PR for this as there were a few additional points that we might want to discuss further.

Issue #810, opened 2024-05-24
[BUG] MalteseNewsClassification fails with KeyError


dokato:
Has something changed in MultiLabel classification lately @x-tabdeveloping ? I can't run the Maltese task anymore. Full stacktrace below.

After `dataset_transform` call the data structure looks correct:

```py
DatasetDict({
    train: Dataset({
        features: ['text', 'label'],
        num_rows: 10784
    })
    validation: Dataset({
        features: ['text', 'label'],
        num_rows: 2293
    })
    test: Dataset({
        features: ['text', 'label'],
        num_rows: 2297
    })
})
```

but then in `_evaluate_subset` it only gets:
```
Dataset({
        features: ['text', 'label'],
        num_rows: 10784
    })
```

```
Error while evaluating MalteseNewsClassification: "Column train not in the dataset. Current columns in the dataset: ['text', 'label']"
Traceback (most recent call last):
  File "/Users/user/proj/mteb/mteb/tasks/MultiLabelClassification/mlt/MalteseNewsClassification.py", line 66, in <module>
    evaluation.run(model)
  File "/Users/user/proj/mteb/mteb/evaluation/MTEB.py", line 365, in run
    raise e
  File "/Users/user/proj/mteb/mteb/evaluation/MTEB.py", line 331, in run
    results, tick, tock = self._run_eval(
  File "/Users/user/proj/mteb/mteb/evaluation/MTEB.py", line 227, in _run_eval
    results = task.evaluate(model, split, output_folder=output_folder, **kwargs)
  File "/Users/user/proj/mteb/mteb/abstasks/AbsTask.py", line 115, in evaluate
    scores[hf_subset] = self._evaluate_subset(model, data_split, **kwargs)
  File "/Users/user/proj/mteb/mteb/abstasks/AbsTaskMultilabelClassification.py", line 80, in _evaluate_subset
    train_split = dataset[train_split]
  File "/Users/user/anaconda3/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 2861, in __getitem__
    return self._getitem(key)
  File "/Users/user/anaconda3/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 2845, in _getitem
    pa_subtable = query_table(self._data, key, indices=self._indices)
  File "/Users/user/anaconda3/lib/python3.9/site-packages/datasets/formatting/formatting.py", line 584, in query_table
    _check_valid_column_key(key, table.column_names)
[...]

PR #812, opened 2024-05-24
Fixes MultilabelClassification eval_split

dokato:
This fixes bug reported in https://github.com/embeddings-benchmark/mteb/issues/810 and caused by https://github.com/embeddings-benchmark/mteb/pull/759 .

Not sure if this is the best approach but since old eval_split was working I just copied it from classification abstask.

### DIFF
--- a/docs/mmteb/points/812.jsonl
+++ b/docs/mmteb/points/812.jsonl
@@ -1,0 +1,2 @@
+{"GitHub": "dokato", "Bug fixes": 2}
+{"GitHub": "isaac-chung", "Review PR": 2}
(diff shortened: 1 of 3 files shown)


isaac-chung:
This is ok for the short term. 
To confirm that it actually works, could you please add `MalteseNewsClassification` in `tests/test_mteb.py` as one of the test tasks? (the list starts on line 30)

isaac-chung (commented):
This is ok for the short term. 
To confirm that it actually works, could you please add `MalteseNewsClassification` in `tests/test_mteb.py` as one of the test tasks? (the list starts on line 30)

dokato:
Done ✅

imenelydiaker:
> This is ok for the short term. To confirm that it actually works, could you please add `MalteseNewsClassification` in `tests/test_mteb.py` as one of the test tasks? (the list starts on line 30)


Actually the `evaluate` method was removed by mistake in #759, I think @dokato just put it back so it should be good

isaac-chung (approved):
Thanks!

Issue #823, opened 2024-05-25
Inconsistencies in the naming of some tasks  fast clustering versions


imenelydiaker:
Hello,

I've found the following inconsistencies in the package:

1. Fast version of the clustering sometimes implemented in a new file and sometimes in the same file as the old task. See [ArxivClusteringP2PFast](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Clustering/eng/ArxivClusteringP2PFast.py), [BlurbsClusteringP2PFast](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Clustering/deu/BlurbsClusteringP2PFast.py) and [BigPatentClustering](https://github.com/embeddings-benchmark/mteb/blob/85dbcf84fc58a6e5375d671c36d9a89551574b96/mteb/tasks/Clustering/eng/BigPatentClustering.py#L42).
2. Inconsistencies in naming: in [ArxivClusteringP2PFast](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Clustering/eng/ArxivClusteringP2PFast.py), the task is named `ArxivClusteringP2P.v3`, why? was there a v2 somewhere?
3. Suffixing the name of the task with `.v2` is confusing: MSMARCO has 2 versions, so one of the tasks is named `MSMARCOv2` although it's not a faster version. https://github.com/embeddings-benchmark/mteb/blob/85dbcf84fc58a6e5375d671c36d9a89551574b96/mteb/tasks/Retrieval/eng/MSMARCOv2Retrieval.py#L10

Any opinions? I'd like to update these.
@isaac-chung @KennethEnevoldsen

isaac-chung:
1. My preference is the old and new tasks should be in the same file.
2. Seems like v3 was a typo? there are v2 results committed.
3. This is likely the biggest issue on this naming topic. There was an agreement with v2 at the time [here](https://github.com/embeddings-benchmark/mteb/pull/757#discussion_r1605240988), but given the new evidence I'm leaning towards renaming to `XYZClusteringFast` (instead of `XYZClustering.v2`). This would affect the relevant results and some values within those files.

imenelydiaker:
> 1. My preference is the old and new tasks should be in the same file.
> 2. Seems like v3 was a typo? there are v2 results committed.

Okay I'll update old tasks with these changes.

[...]

PR #740, opened 2024-05-16
fix: Convert StackExchangeClustering to fast

shreeya-dhakal:
Resolve #739

## Checklist for adding MMTEB dataset

Reason for dataset addition:

- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [x] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 
- [ ] I have added points for my submission to the [points folder](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) using the PR number as the filename (e.g. `438.jsonl`).

PR #825, opened 2024-05-27
rename *fast to *v2 for clustering task

KennethEnevoldsen:
Address #823


KennethEnevoldsen:
didn't see #823 before opening this. I will leave it open for now, but we should way for the issue to be resolved.

isaac-chung (commented):
1. Should we update `ArxivClusteringP2PFast` as well?
2. The results file would have a different filename, e.g. `results/intfloat__multilingual-e5-small/BlurbsClusteringS2S.v2.json`
3. More of a note: I realized that we forgot to run the fast version of StackExchange in https://github.com/embeddings-benchmark/mteb/pull/740

KennethEnevoldsen:
seems like we agree on the issue - I have fixed 1) we might run the tasks in another PR

PR #826, opened 2024-05-27
Updated CLI to use new task filter

KennethEnevoldsen:
Updated CLI to use new task filter. Also removed a few flags which seems to not be used.

### DIFF
--- a/mteb/cmd.py
+++ b/mteb/cmd.py
@@ -17,6 +17,7 @@ from pathlib import Path
 
 from sentence_transformers import SentenceTransformer
 
+import mteb
 from mteb import MTEB
 
 logging.basicConfig(level=logging.WARNING)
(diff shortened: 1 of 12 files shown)


mteb/cmd.py
```diff
@@ -84,12 +85,6 @@ def main():
     parser.add_argument(
         "--device", type=int, default=None, help="Device to use for computation"
     )
-    parser.add_argument(
-        "--batch_size", type=int, default=32, help="Batch size for computation"
-    )
```
imenelydiaker, line 89: I think this one is important, it's supposed to be used by the encode function of Evaluators

KennethEnevoldsen: But atm I don't think it is used so I chose to remove it to avoid users thinking that it is used.

Muennighoff (approved):
Amazing!

imenelydiaker (approved):
LGTM, thank you! 🚀

PR #833, opened 2024-05-27
fix: Add MIRACL retrieval

imenelydiaker:
I've added what was missing here: #642.

Remaining: running models `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2` and `intfloat/multilingual-e5-small`.

Edit: `NDCG (no self)@K` scores are not saved with other outputs, updating
Edit 2: Need to fix tests


mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -468,51 +466,74 @@ def evaluate(
...
+            )
+
         for query_id in scores.keys():
             for k in k_values:
                 ndcg[f"NDCG@{k}"] += scores[query_id]["ndcg_cut_" + str(k)]
                 _map[f"MAP@{k}"] += scores[query_id]["map_cut_" + str(k)]
                 recall[f"Recall@{k}"] += scores[query_id]["recall_" + str(k)]
                 precision[f"P@{k}"] += scores[query_id]["P_" + str(k)]
+                if ignore_identical_ids:
+                    ndcg_no_self[f"NDCG (no self)@{k}"] += scores_wo_identical_ids[
```
KennethEnevoldsen, line 520:
```suggestion
                    ndcg_no_self[f"NDCG(no self)@{k}"] += scores_wo_identical_ids[
```
minor preference (either is fine)

imenelydiaker:
> Edit 2: Need to fix tests

Hey, so I'm a bit confused about the retrieval scores. The test is failing because we're not getting the same results when setting `ignore_identical_ids=False` (default option).  I may need some help here 😅 @Muennighoff @KennethEnevoldsen 

> Edit: NDCG (no self)@K scores are not saved with other outputs, updating

The code merges `ndcg(no_self)@k` with `ndcg@k`, so when `ignore_identical_ids` is set to true it replace the `ndcg@k` scores. Is this the expected behavior?

imenelydiaker:
> > Hey, so I'm a bit confused about the retrieval scores. The test is failing because we're not getting the same results when setting ignore_identical_ids=False (default option). I may need some help here 😅 @Muennighoff @KennethEnevoldsen
> 
> It seems problematic - any idea why this is?

The only 2 updates that were made are:
- Removing the if/else condition on `query_id` and `corpus_id` (see code below)
- `ignore_identical_ids` is set to `False` instead of `True` by default (set to `True` for `ArguAna` and `QuoraRetrieval` only, related to the removal of the previous condition)

Thread here: https://github.com/embeddings-benchmark/mteb/pull/642#issuecomment-2098830205
```py
if corpus_id != query_id: # THIS CONDITION WAS REMOVED
    if len(result_heaps[query_id]) < top_k:
        # Push item on the heap
        heapq.heappush(result_heaps[query_id], (score, corpus_id))
    else:
        # If item is larger than the smallest in the heap, push it on the heap then pop the smallest element
        heapq.heappushpop(
            result_heaps[query_id], (score, corpus_id)
        )
```

KennethEnevoldsen:
but wasn't the default before ignore_identical_ids = True?

imenelydiaker:
> but wasn't the default before ignore_identical_ids = True?

Yes but after the discussion https://github.com/embeddings-benchmark/mteb/pull/642#issuecomment-2104947181, we agreed on setting it to False by default and to True for `ArguAna` and `Quora` only (option 2). I'm missing something maybe?

KennethEnevoldsen:
Ahh right, but shouldn't we then we expect that performance scores will change? Which we don't want I assume?

I which case I would do the following:

```
ignore_identical_ids = True # default
ignore_identical_ids = True # for MIRACL
ignore_identical_ids = False # for all current datasets to keep them compatible - alternative keep only do it for datasets in a benchmark
```

imenelydiaker:
After re-examining the comment https://github.com/embeddings-benchmark/mteb/pull/642#issuecomment-2104947181, I think we should keep it to True by default and set it to False for MIRACL (option 1), this would not change current results (see below).

> So, the scores will not be affected as the self-retrieval is double-checked during evaluation as well here with the flag ignore_identical_ids set to True, which is the desirable way to go.

> (1) Keep the code as is with ignore_identical_ids=True but inform users to keep the query_ids and document_ids are distinct from each other, e.g. for MIRACL I pass ignore_identical_ids=False.

KennethEnevoldsen:
@imenelydiaker that is perfectly fine with me

imenelydiaker:
I set back `ignore_identical_ids` default value to `True` but I can't get the tests to pass. 

@KennethEnevoldsen @Muennighoff can someone please have a look at the evaluator?

The retrieval Evaluator test should look something like this (values of NDCG are incorrect, need to be updated): 
```py
  @pytest.mark.parametrize(
      "ignore_identical_ids, expected_metrics",
      [
          (
              False, 
              {
              "ndcg": {'NDCG@1': 1.0, 'NDCG@2': 1.0, 'NDCG@3': 1.0},
              "map": {"MAP@1": 0.25, "MAP@2": 0.25, "MAP@3": 0.25},
              "recall": {"Recall@1": 0.25, "Recall@2": 0.25, "Recall@3": 0.25},
              "precision": {"P@1": 0.5, "P@2": 0.25, "P@3": 0.16667},
              }
          ),
          (
              True, 
              {
              "ndcg": {'NDCG@1': 1.0, 'NDCG@2': 1.0, 'NDCG@3': 1.0, 'NDCG(no self)@1': 0.5, 'NDCG(no self)@2': 0.30657, 'NDCG(no self)@3': 0.30657},
              "map": {"MAP@1": 0.25, "MAP@2": 0.25, "MAP@3": 0.25},
              "recall": {"Recall@1": 0.25, "Recall@2": 0.25, "Recall@3": 0.25},
              "precision": {"P@1": 0.5, "P@2": 0.25, "P@3": 0.16667},
              }
          ),
      ],
  )
  def test_metrics_at_k(self, ignore_identical_ids, expected_metrics):
      # Qid: {Docid: Relevance}
      relevant_docs = {
          "0": {"0": 1, "1": 1},
          "1": {"1": 1},
      }
      results = {
          "0": {"0": 1.0, "1": 0.9, "2": 0.8},
          "1": {"0": 0.0, "1": 1.0, "2": 0.0},
      }

      ndcg, _map, recall, precision = self.evaluator.evaluate(
          relevant_docs,
          results,
          [1, 2, 3],
          ignore_identical_ids=ignore_identical_ids,
      )

      assert ndcg == expected_metrics["ndcg"]
      assert _map == expected_metrics["map"]
      assert recall == expected_metrics["recall"]
      assert precision == expected_metrics["precision"]
```

KennethEnevoldsen:
> @KennethEnevoldsen @Muennighoff can someone please have a look at the evaluator?

@imenelydiaker it seems like the tests don't run - can you give an idea of what is failing then I will have a look at it

KennethEnevoldsen:
@imenelydiaker I am having a look at this as well:

Seems like we expect:
```
 {"NDCG@1": 0.5, "NDCG@2": 0.30657, "NDCG@3": 0.30657}

and get:
{'NDCG@1': 1.0, 'NDCG@2': 1.0, 'NDCG@3': 1.0, 'NDCG(no self)@1': 0.5, 'NDCG(no self)@2': 0.30657, 'NDCG(no self)@3': 0.30657}
```
The no-self (which was the default before as far as I understand) gives the correct result.

As I understand the "no-self"-rule seems to remove cases where the "query id" == "corpus id". This generally seems like a bad idea (queries are generally not in the corpus and might coincidentally have matching IDs if you are not careful during dataset construction). There are naturally a few cases where it matters.

Generally, I would probably remove the `ignore_identical_ids` flag. We are just toggling whether it is computed. It does not determine the main_score on the task. However, datasets that previously used "ndcg_at_10" should now use "ndcg_at_10(no self)" (as far as I understand). Of course, if contributors were careful during dataset construction this should not matter (see the tmp.py file I have pushed)

My conclusion (which might be wrong):
- I believe the scores are correct
- I don't believe the (no self) is a good default, except maybe in a few cases. (I am probably missing something)
- The two methods should be equivalent if the dataset is constructed carefully.
- The current test seems like a bad test since the scores do not match what I would suspect, this is due to the qid and cid matching and then being filtered.


 @Muennighoff would love a second pair of eyes on this
@imenelydiaker let me know if there is something that I might have misunderstood here.

Muennighoff:
Very interesting & detailed investigation!! 🧐 
Maybe we should change the test then to be:

1) Test an example with Identical IDs: If we keep the `ignore_identical_ids` flag, test that it works both with (1.1) `ignore_identical_ids=True` & (1.2) `ignore_identical_ids=False`
2) Test an example without Identical IDs: Same as above test both `ignore_identical_ids=True/False` if we keep it
3) Maybe also adapt `InstructionRetrievalEvaluator` & `test_InstructionRetrievalEvaluator` & add the flag there too? Afaict, there are no identical IDs there for now, but in case a dataset with them gets added in the future? @orionw might know if this makes sense or we don't need to think about it there for now

orionw:
Very interesting issue, good catch everyone! I think this likely doesn't effect InstructionRetrieval right now, but likely will as they have similar dataset setups.  It would be great if this could apply to both!

KennethEnevoldsen:
@imenelydiaker should we finish up this PR soon (seems like it is already starting to see a lot of conflicts)

imenelydiaker:
@KennethEnevoldsen so I'm almost able to recover all results. Turns out all metrics are impacted by the flag `ignore_identical_ids` since the results dict changes when it is set to `True`.

I'm not a fan of how I've done things in this PR. I'd prefer to replace `ndcg` by `ndcg_no_self` when `ignore_identical_ids=True`  like I've done it for other metrics, **as the `ndcg` score with the old dict should not be correct**. We should have only one `ndcg` score and one score for each metric whatever the `ignore_identical_ids` flag value is (it replaces the results dict).

I don't have much experience with retrieval evaluators, can someone have a look at this and help if possible?

mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -518,44 +516,63 @@ def evaluate(
...
 
         map_string = "map_cut." + ",".join([str(k) for k in k_values])
         ndcg_string = "ndcg_cut." + ",".join([str(k) for k in k_values])
+        ndcg_string_no_self = "ndcg_cut." + ",".join([str(k) for k in k_values])
         recall_string = "recall." + ",".join([str(k) for k in k_values])
         precision_string = "P." + ",".join([str(k) for k in k_values])
         evaluator = pytrec_eval.RelevanceEvaluator(
             qrels, {map_string, ndcg_string, recall_string, precision_string}
         )
         scores = evaluator.evaluate(results)
```
imenelydiaker, line 548: Computing scores for a dataset where query ids and document ids are different (new default: `ignore_identical_ids=False`)

imenelydiaker, line 548: Computing scores for a dataset where query ids and document ids are different (new default: `ignore_identical_ids=False`)

imenelydiaker:
@KennethEnevoldsen Here is what I think is the best solution for the evaluator (last commit):
- Set default `ignore_identical_ids` of Retrieval Evaluator to `False`.
- Compute metrics once (with self or without) according to `ignore_identical_ids` flag [this flag replaces the results dict when set to `True`].
- Added tests for abstention and metrics computing (abstention test is failing, can you please have a look at it? [error: `ValueError: not enough values to unpack (expected 5, got 4)`])

KennethEnevoldsen:
Will take a look at this @imenelydiaker either today or tomorrow

KennethEnevoldsen:
So to summarise what I can see has happened so far:

- we now use only NDCG, no "no self" or anything. If you want to ignore identical ids across corpus and query you do it using "ignore_identical_ids". However this is not the default.
- We are able to recover scores on most previous tasks (@imenelydiaker which task weren't you able to recover it on?)
- Added test to ensure that calculations correct
  - currently these fail for abstention

@imenelydiaker can you clarify that this is correct?

I have then fixed the following:
- a minor issue when handling outputs in the test `test_metrics_at_k` (however the code works)
- fixed a minor typo in the nAUC test (this fixed the first case). This case seem to match the case in the previous test which means we reproduce the results. 
- In the second case I can't get the scores to match the expected. I have updated these to:

```
                {
                     "nAUC_NDCG@3_max": 0.9507,
                     "nAUC_NDCG@3_std": 0.9125,
                     "nAUC_NDCG@3_diff1": 0.9507,
                 },
```

This might be a mistake, depending on where the previous scores came from (@imenelydiaker?)

@hgissbkh can I ask you to check out the abstention test as well?

hgissbkh:
> So to summarise what I can see has happened so far:
> 
> * we now use only NDCG, no "no self" or anything. If you want to ignore identical ids across corpus and query you do it using "ignore_identical_ids". However this is not the default.
> * We are able to recover scores on most previous tasks (@imenelydiaker which task weren't you able to recover it on?)
> * Added test to ensure that calculations correct
>   
>   * currently these fail for abstention
> 
> @imenelydiaker can you clarify that this is correct?
> 
> I have then fixed the following:
> 
> * a minor issue when handling outputs in the test `test_metrics_at_k` (however the code works)
> * fixed a minor typo in the nAUC test (this fixed the first case). This case seem to match the case in the previous test which means we reproduce the results.
> * In the second case I can't get the scores to match the expected. I have updated these to:
> 
> ```
> # from
>                 {
>                      "nAUC_NDCG@3_max": 0.9507,
>                      "nAUC_NDCG@3_std": 0.9125,
>                      "nAUC_NDCG@3_diff1": 0.9507,
>                  },
> # to:
>                 {
>                     "nAUC_NDCG@3_max": 0.8368244286523474,
>                     "nAUC_NDCG@3_std": 0.9125701917627439, # seems odd that std is higher than value, when min is 0?
>                     "nAUC_NDCG@3_diff1": 0.950708977119359, # seems odd that diff1 is higher than max?
>                 },
> ```
> 
> This might be a mistake, depending on where the previous scores came from (@imenelydiaker?)
> 
> @hgissbkh can I ask you to check out the abstention test as well?

Hi @KennethEnevoldsen,

First, I wanted to highlight that the observed nAUC_NDCG@3_X values are not inconsistent with each other. Here, X denotes the function we use to assess confidence at the instance level. For instance, nAUC_NDCG@3_max is the normalized nAUC for the NDCG@3 metric when the maximum query-document similarity score is used as a confidence estimator. Therefore, we can have nAUC_NDGC@3_max < nAUC_NDGC@3_diff1 without any problem!
I can check the abstention tests to make sure everything is correct. Just to make sure, you are talking about [tests/test_RetrievalEvaluator.py](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_RetrievalEvaluator.py) and [tests/test_RerankingEvaluator.py](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_RerankingEvaluator.py), right?

Cheers!

KennethEnevoldsen:
> Therefore, we can have nAUC_NDGC@3_max < nAUC_NDGC@3_diff1 without any problem!

Perfect that is very good to hear, then I could imagine the the current implementation is current, but a check would be appreciated.

hgissbkh:
> > Therefore, we can have nAUC_NDGC@3_max < nAUC_NDGC@3_diff1 without any problem!
> 
> Perfect that is very good to hear, then I could imagine the the current implementation is current, but a check would be appreciated.

Hi @KennethEnevoldsen 
Just checked the abstention metrics, everything is fine with the test values
Cheers!

imenelydiaker:
> So to summarise what I can see has happened so far:
> 
> * we now use only NDCG, no "no self" or anything. If you want to ignore identical ids across corpus and query you do it using "ignore_identical_ids". However this is not the default.

Yes the self-retrieval is no default anymore. It can only be triggered by setting `ignore_identical_ids=True` in the task definition.

> * We are able to recover scores on most previous tasks (@imenelydiaker which task weren't you able to recover it on?)

This is being checked rn with @isaac-chung, we can't run all retrieval tasks again (135 tasks), so we are using the returned dataset (corpus and queries) by the `task.load_data()` function for each task to look for matching ids.
Tasks where some `corpus_id == query_id` will have `ignore_identical_ids=True` to match previous MTEB evaluations.
The only thing recovered rn is the test metrics results when `ignore_identical_ids=True`.

> * Added test to ensure that calculations correct

Yes, I've added a test for `ignore_identical_ids=False`, for the metrics and abstention scores. The test parameters can be improved if needed.

>   * currently these fail for abstention

Due to a typo, I think you have already fixed it, thanks !
@hgissbkh just confirmed the values are correct.

KennethEnevoldsen:
Wonderful! Once we have the checks from @isaac-chung I believe we can (finally) merge this one in.

imenelydiaker:
@KennethEnevoldsen I added `ignore_identical_ids=True` where it is required.

@isaac-chung ran [this script](https://colab.research.google.com/drive/1RaiO743LEail4PSybbRBu0RCSb2HUTNS?usp=sharing) that looks into query ids and document ids in each retrieval dataset. If ids are repeated in both, then we get the name of the task and set `ignore_identical_ids=True`. We found this to be more efficient than running all tasks (134) again and comparing scores.


Tasks list:
```
'DanFEVER', 'ArguAna', 'ClimateFEVER', 'FeedbackQARetrieval', 'FEVER', 'FiQA2018', 'MLQuestions', 
'MSMARCO', 'BSARDRetrieval', 'JaGovFaqsRetrieval', 'XPQARetrieval', 'ArguAna-PL', 'FiQA-PL', 
'MSMARCO-PL', 'RiaNewsRetrieval', 'RuBQRetrieval', 'T2Retrieval', 'MMarcoRetrieval', 'EcomRetrieval',
'MedicalRetrieval', 'VideoRetrieval'
```

`XPQA` is being fixed by #918, didn't update the task.

I believe this PR is finally ready! 😄

KennethEnevoldsen:
Will merge this in @imenelydiaker is it worth applying the #918 fix to some of the other datasets (I know it is for DanFEVER)

imenelydiaker:
> Will merge this in @imenelydiaker is it worth applying the #918 fix to some of the other datasets (I know it is for DanFEVER)

Totally worth it if the dataset is not used in any benchmark yet. The idea is to avoid running models again.

KennethEnevoldsen:
Yea - even if it is in a benchmark we have just create a version bump of it

PR #853, opened 2024-05-30
Add BIRCO Doris Mae dataset

bp-high:

Reason for dataset addition:


mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -129,16 +137,37 @@ def compute_metrics_batched(self, model):
...
+                relevance_scores[query_id][doc] = doc_score
+
             all_mrr_scores.append(scores["mrr"])
             all_ap_scores.append(scores["ap"])
 
         mean_ap = np.mean(all_ap_scores)
         mean_mrr = np.mean(all_mrr_scores)
 
-        return {"map": mean_ap, "mrr": mean_mrr}
+        evaluator = pytrec_eval.RelevanceEvaluator(qrels, {"ndcg_cut.10"})
```
orionw, line 161: Can we add this section to inside the `_compute_metrics_instance` function just to keep all the metric calculation in one place? You'll have to pass `qrels` into that function, I think.

At some point we should switch over to just using `pytrec_eval.RelevanceEvaluator` for all of them (like the `RetrievalEvaluator` does) but that doesn't have to be this PR.

Also, like the retrieval evaluator can we do multiple nDCG values (not just 10)?  See [this line](https://github.com/embeddings-benchmark/mteb/blob/50c51646dde7aba5bfe6cf66dc9673a61f87a7e6/mteb/evaluation/evaluators/RetrievalEvaluator.py#L464).

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -224,7 +280,7 @@ def _compute_metrics_instance(self, query_emb, docs_emb, is_relevant):
 
         mrr = self.mrr_at_k_score(is_relevant, pred_scores_argsort, self.mrr_at_k)
         ap = self.ap_score(is_relevant, pred_scores.cpu().tolist())
-        return {"mrr": mrr, "ap": ap}
+        return {"mrr": mrr, "ap": ap}, pred_scores.cpu().tolist()
```
orionw, line 283: We can then not return the scores here, since the calculation will be done in this function.

mteb/tasks/Reranking/eng/BIRCODorisMaeInstructReranking.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+from ....abstasks.AbsTaskReranking import AbsTaskReranking
+
+
+class BIRCODorisMaeInstructReranking(AbsTaskReranking):
```
orionw, line 8: Are there two versions of this dataset?  Why `InstructReranking` and `Reranking`?

orionw (changes requested):
Thanks for adding this! I had a few questions and some minor changes to be made to the evaluation.

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -129,16 +137,37 @@ def compute_metrics_batched(self, model):
             if num_pos == 0 or num_neg == 0:
                 continue
 
+            for doc in instance["positive"]:
+                qrels[query_id][doc] = 1
```
orionw, line 141: It would be best to grab the real qrels here, some of these have different scores (e.g. ClinicalTrials having 1 and 2, which are used for nDCG calculations).

mteb/evaluation/evaluators/RerankingEvaluator.py
bp-high: Yup can do this.

mteb/evaluation/evaluators/RerankingEvaluator.py
bp-high: Okay

mteb/evaluation/evaluators/RerankingEvaluator.py
bp-high: Hmm, Good Point , I think then it would be better to keep the dataset in a retrieval format, rather than changing it to reranker format, this was one thing I was also concerned about.
Screenshot 2024-06-01 at 1 30 35 AM

orionw: Ah I see.  I guess it depends on if @DPWXY used the qrels as is, or if they binarized the qrels to "relevant" and "non-relevant".  

It will be hard to migrate to `retrieval` because there the pool of documents is combined, whereas this is reranking where each pool is separate.

mteb/tasks/Reranking/eng/BIRCODorisMaeInstructReranking.py
orionw: Given that some of these datasets are already in MTEB without the instruction (e.g. Arguana), I think we should only do the `instruction` version to avoid duduplication.

mteb/evaluation/evaluators/RerankingEvaluator.py
DPWXY: We use the real qrels for nCDG calculation. Especially for the Doris-Mae dataset, the qrels are continuous scores rather than binarized "relevant" and "non-relevant". And ClinicalTrials having 0, 1 and 2 relevance scores for nDCG calculations.

orionw: @bp-high why don’t we add qrels as an optional data input to the abstract task? I’m surprised it didn’t have nDCG already, probably because of this issue.

So check for the existence of a dataset split called qrels and if it exists, load it.

KennethEnevoldsen:
@bp-high just a ping on this PR

KennethEnevoldsen:
Will close this PR as it seems to have grown stale. If any of you find the time to finish up the PR feel free to re-open it

PR #858, opened 2024-05-31
Script for converting mteb results to tex

Muennighoff:


scripts/mteb_to_tex.py
```diff
@@ -0,0 +1,225 @@
...
+"""
+import json
+import os
+import sys
+
+from mteb import MTEB
+import numpy as np
+
+### GLOBAL VARIABLES ###
+TASK_LIST_CLASSIFICATION = [
```
KennethEnevoldsen, line 13: Only intended for classical MTEB?

KennethEnevoldsen: probably worth creating a variable for it in the benchmarks.py and just import it

scripts/mteb_to_tex.py
```diff
@@ -0,0 +1,225 @@
...
+MODEL_TO_NAME = {
+    "gritlm-7b": "GritLM-7B",
+}
+
+### LOGIC ###
+results_folder = sys.argv[1].rstrip("/")
+
+all_results = {}
+
+mteb_task_names = [t.metadata.name for t in MTEB().tasks] + ["CQADupstackRetrieval"]
```
KennethEnevoldsen, line 124: why add it here?

scripts/mteb_to_tex.py
```diff
@@ -0,0 +1,225 @@
+"""
```
KennethEnevoldsen, line 1: Seems like it might be worth creating a CLI for some of this, which we can make sure is also tested (such that future edits also make sure to fix it)

Something like:
```
mteb create_table style=tex folder={path_to_folder} benchmark="mteb_en"  
```

KennethEnevoldsen (commented):
Does this work with the new results format?

Muennighoff:
Great points! Yes we need to still
- Make it work with the new results format and use `MTEBResults` instead
- Instead import the global variables

KennethEnevoldsen:
@Muennighoff we might want to introduce a benchmark object which is essentially a list of tasks + some description and potentially a reference.

I am a bit unsure how it should look like but e.g:

```
mteb.benchmarks import MTEB_EN

results = MTEB_EN.load_results_from_folder(path)
```

Issue #834, opened 2024-05-28
Consolidate "convert_to_fast" methods in AbsTaskClusteringFast


isaac-chung:
Currently there are
- `clustering_downsample`: used only by `ArxivClusteringP2P.v2`
- `convert_to_fast`: used only by `TenKGnadClusteringS2S.v2`
- `dataset_transform` from #660 

To make AbsTaskClusteringFast consistent, the proposal is to
1. consolidate into one `convert_to_fast` method that will be called in a dataset's `dataset_transform`. This new method will be refactored to use the logic from the 3rd points above.
2. rerun results for `ArxivClusteringP2P.v2` and `TenKGnadClusteringS2S.v2`

Aside: I've also compared the speed of using dataset filtering (e.g. in `HALClusteringS2S.v2`) vs the current method for filtering out single count labels in clustering datasets (e.g. in `BlurbsClusteringP2P.v2`). The former is 5x slower.
```
HALClusteringS2S.v2: 5.16 s ± 40.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
BlurbsClusteringP2P.v2: 1.03 s ± 39.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

KennethEnevoldsen:
One point here is that `clustering_downsample` it not intended to convert to fast, but rather just downsample "old-format" clustering tasks.

isaac-chung:
Can those be effectively overwritten and replaced by the current version of fast-conversion?

KennethEnevoldsen:
Not in all cases e.g. #828 where it is specified in the comment:

```
# a faster version of BlurbsClusteringP2P, since it does not sample from the same distribution we can't use the AbsTaskClusteringFast, instead we
# we simply downsample each cluster.
```

So clusters can be constructed e.g. using different labels

isaac-chung:
But in #832 we followed the steps in #660 to convert Blurbs to use `AbsTaskClusteringFast`. Should we not have done that?

KennethEnevoldsen:
So one cluster might do "genres" and another do "subgenres", which will lead the new dataset clustering at two different levels at the same time (worse cases include e.g. "newspaper source" and "news/sports/etc"). There might even be duplicates across these categories. I believe the conversion script that I made for one of the datasets specifically checks if the labels are the same across clusters.

isaac-chung:
[...]

PR #862, opened 2024-06-02
fix: add check_label_distribution for ClusteringFast

isaac-chung:
Addresses #834


KennethEnevoldsen:
thanks, @isaac-chung, seems like most (all?) datasets pass the checks. Probably my worry was unwarranted

isaac-chung:
@KennethEnevoldsen no worries, it's good that we checked 👍 

We could merge this just to point back to it if the topic comes up again. What do you think?

isaac-chung:
Added points as well - see if that makes sense?

KennethEnevoldsen:
Looks great! - setting automerge

PR #878, opened 2024-06-04
Added benchmark object

KennethEnevoldsen:
Assuming we want a selection menu for benchmarks in the future leaderboard I have now added a benchmark object. Generally it shouldn't change much otherwise.

The idea is that you would select the benchmark. The description will show + some references. Potentially with a dropdown for the citation. 

- Added benchmark object (should work like a list)
- removed a duplicate task
- Added SEB (we should probably add a few more here as well)


Muennighoff (commented):
Nice, I think we also need to change the README which imports MTEB_MAIN_EN i.e. https://github.com/embeddings-benchmark/mteb?tab=readme-ov-file#dataset-selection

KennethEnevoldsen:
> Nice, I think we also need to change the README which imports MTEB_MAIN_EN i.e. https://github.com/embeddings-benchmark/mteb?tab=readme-ov-file#dataset-selection

Added a test and it does run just fine (since it has an __iter__ method). I also changed the type hint to iterable (weaker than sequence) to reflect that it is the only requirement.

PR #882, opened 2024-06-05
Updated CLI for MTEB

KennethEnevoldsen:
It now includes three main commands one for running, one for getting an overview, and one for creating the metadata for hf.

I also:
- added lower bound on dependencies as it caused a few issues.
- deleted some results with a model revision attached (also created a fix to make sure that doesn't happen as much going forward)
- Added a test for the cli
- Added a quite extensive docstring to the CLI
- made relevant changes to the documentation
- made sure it is backward compatible

## General Checklist


mteb/cli.py
KennethEnevoldsen, line 1: This is the main change

KennethEnevoldsen:
Due to an accidental merge into main from my side followed by a few PR merged I will just merge this in now (we have already discussed it in other issues and also discussed it @x-tabdeveloping). 

Will still implement any changes you might have - just want to make sure that we don't keep breaking changes on the main branch for too long. (we can disallow pushed without a PR to main to avoid this?)

Issue #835, opened 2024-05-28
Paper segment: Comparing scores of Clustering to ClusteringFast


KennethEnevoldsen:
These segments compare the new bootstrapping method (+ multilevel) for clustering methods to the older one. The goal is to establish that it produces similar results (model rank) at a lower cost.

**Details**
1. Using e.g. the English benchmark's clustering tasks, obtain v_measures of 3 models.
2. On the same task's v2 (fast version) and on the same models, obtain v_measures
3. These two methods should then obtain a [spearman ranks correlation](https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient) of ~1. If we just rank them I believe it would be equivalent to just using Spearman ranks correlation on the v_measure
4. Additionally, run a check for duplicate sentences like in #793 (e.g. on #407 )

Multi-level will be covered in #849

isaac-chung:
I'm happy to collaborate and help out here. I'd imagine the evaluation criteria are quite similar to those of the retrieval speedups?

KennethEnevoldsen:
Yep, quite similar feel free to update the first comment. I believe a good case for this is the MTEB(eng) clustering tasks e.g. arxiv clustering (though it might be too large).

Not sure exactly what the most convincing argument is here, but something like:

| model | old rank | new score |
| -- | -- | -- |
| m1 | 1 | 1 (-) |
| m2 | 2 | 3 (downarrow) |
| m3 | 3 | 2 (uparrow) |

might be simple, but hard to scale to many tasks (ideally we want all ranks to stay the same). Significant rank is probably a better metric.

isaac-chung:
> I would probably go with a small, medium, large, and then "close-relative" for the large one (we want to differentiate if we do a spearman's correlation matrix we should have perfect for small, medium, large and close to perfect for large and relatives)

@KennethEnevoldsen I picked e5-small, e5-base, and e5-large. Any preferences on the "close-relative" model?

isaac-chung:
Looks like we mixed up how much to embed and how much to sample for bootstrapping. Added more analyses in the PR above.

KennethEnevoldsen:
A few good pairs for:
- [multilingual-e5-large](https://huggingface.co/intfloat/multilingual-e5-large)

would be:
- [e5-large-v2](https://huggingface.co/intfloat/e5-large-v2)
- [nomic-embed-text-v1](https://huggingface.co/nomic-ai/nomic-embed-text-v1)
[...]

PR #892, opened 2024-06-07
fix: Compare Cluster and ClusterFast scores and speedup

isaac-chung:
## 1) What
Part of #835 .

- Runs english benchmark's clustering (except Arxiv since it's hierarchical) for both Clustering and ClusteringFast on e5-small, e5-base, e5-large, and a few more models.
```
python scripts/run_mteb_english_cluster.py
```

- Calculates the spearman's correlation on the v_measures from both task types
- sum the times to run both benchmarks, divide cluster by clusterfast to obtain speedup multiplier.
```
python scripts/mteb_english_cluster_spearman.py
```

## 2) Hardware
These were run on a single A10.

## 3) Scores on ClusteringFast + Speedup
These are the Spearman's correlation scores and speedups from Clustering to ClusteringFast. The old notes are hidden within each expand.

### 4% n_samples for every task except for RedditClustering and StackExchangeClustering with 32768)
In addition, p-values for each model is reported. 

| Task | Spearman | Significant Spearman | Speedup |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.9505 | 0.9390 | 31.50x |
| BiorxivClusteringS2S        | 0.9890 | 0.9679 | 14.31x |
| MedrxivClusteringP2P        | 0.9615 | 0.8200 | 21.48x |
| MedrxivClusteringS2S        | 0.9560 | 0.9510 | 8.39x |
| RedditClustering            | 0.9670 | 0.9790 | 11.72x |
| RedditClusteringP2P         | 0.9670 | 0.7370 | 22.77x |
| StackExchangeClustering     | 0.9121 | 0.9486 | 9.55x |
| StackExchangeClusteringP2P  | 0.9670 | 0.9497 | 20.20x |
| TwentyNewsgroupsClustering  | 1.0000 | 0.9832 | 5.02x |
| Average                     | 0.9634 | 0.9195 | 16.11x |

<details>
  <summary>Old notes comparing across tasks.</summary>

### a) Current Implementation
|  Model    | Spearman | Speedup |
|-----------|----------|---------|
| BiorxivClusteringP2P        | 0.5 | 34.46x
| BiorxivClusteringS2S        | 1.0 | 20.11x
| MedrxivClusteringP2P        | 0.5 | 17.30x
| MedrxivClusteringS2S        | 1.0 | 10.30x
| RedditClustering            | 1.0 | 23.15x
| RedditClusteringP2P         | 1.0 | 27.32x
| StackExchangeClustering     | 1.0 | 18.71x
| StackExchangeClusteringP2P  | 0.5 | 4.50x
| TwentyNewsgroupsClustering  | 1.0 | 11.44x

### b) New Commit: b8a5be4
|  Model    | Spearman | Speedup |
|-----------|----------|---------|
| BiorxivClusteringP2P      | 0.5 | 3.68x
| BiorxivClusteringS2S      | 1.0 | 3.46x
| MedrxivClusteringP2P      | 1.0 | 1.83x
| MedrxivClusteringS2S      | 1.0 | 1.80x
| RedditClustering          | 1.0 | 17.38x
| RedditClusteringP2P       | 1.0 | 21.55x
| StackExchangeClustering   | 0.5 | 14.43x
| StackExchangeClusteringP2P | 0.5 | 3.68x
| TwentyNewsgroupsClustering | 1.0 | 2.61x

### c) New Commit: 6295dcc (swap max_documents_per_cluster and max_documents_to_embed values)
|  Model    | Spearman | Speedup |
|-----------|----------|---------|
| BiorxivClusteringP2P      | 0.5 | 34.38x
| BiorxivClusteringS2S      | 1.0 | 14.12x
| MedrxivClusteringP2P      | 1.0 | 17.09x
| MedrxivClusteringS2S      | 1.0 | 8.63x
| RedditClustering          | 1.0 | 75.32x
| RedditClusteringP2P       | 1.0 | 156.63x
| StackExchangeClustering   | 1.0 | 46.77x
| StackExchangeClusteringP2P | 1.0 | 29.39x
| TwentyNewsgroupsClustering | 1.0 | 9.68x

### d) New Commit: 3a0d9c4 (c + increasing max_documents_per_cluster to 65536)
|  Model    | Spearman | Speedup |
|-----------|----------|---------|
| BiorxivClusteringP2P      | 0.5 | 31.38x
| BiorxivClusteringS2S      | 1.0 | 10.44x
| MedrxivClusteringP2P      | 0.5 | 15.58x
| MedrxivClusteringS2S      | 1.0 | 5.54x
| RedditClustering          | 1.0 | 50.16x
| RedditClusteringP2P       | 1.0 | 132.17x
| StackExchangeClustering   | 1.0 | 32.46x
| StackExchangeClusteringP2P | 1.0 | 25.43x
| TwentyNewsgroupsClustering | 1.0 | 5.97x

</details>

<details>
  <summary>Old notes comparing across models.</summary>

### a) Current Implementation
| Model    | Spearman | Speedup |
|----------|----------|---------|
| e5-small |  0.7833  | 14.56x  |
| e5-base  |  0.8333  | 15.80x  |
| e5-large |  0.8000  | 16.93x  |

### b) New Commit: b8a5be4
| Model    | Spearman | Speedup |
|----------|----------|---------|
| e5-small |  0.8500  | 5.76x  |
| e5-base  |  0.8833  | 5.97x  |
| e5-large |  0.8000  | 6.17x  |

### c) New Commit: 6295dcc (swap max_documents_per_cluster and max_documents_to_embed values)
| Model    | Spearman | Speedup |
|----------|----------|---------|
| e5-small |  0.8833  | 32.05x  |
| e5-base  |  0.9000  | 40.41x  |
| e5-large |  0.9000  | 52.77x  |

### d) New Commit: 3a0d9c4 (c + increasing max_documents_per_cluster to 65536)
| Model    | Spearman | Speedup |
|----------|----------|---------|
| e5-small |  0.8833  | 22.21x  |
| e5-base  |  0.9000  | 32.11x  |
| e5-large |  0.9500  | 47.11x  |

</details>

## Checklist


scripts/mteb_english_cluster_spearman.py
```diff
@@ -0,0 +1,70 @@
...
+        if version in res.task_name:
+            main_scores_fast.append(main_score)
+            times_fast.append(eval_time)
+        else:
+            main_scores.append(main_score)
+            times.append(eval_time)
+
+    ## Spearman score
+    print(f"{model_name} | {revision}")
+    print(stats.spearmanr(main_scores, main_scores_fast).statistic)
```
KennethEnevoldsen, line 66: So for a model, we get the correlation between scores across tasks (I don't believe this is what we are interested in, but might be missing something?)

I would suspect doing it across models for each task.

So for task A you would get a Spearman correlation (not for the model). For these model this should be a perfect correlation so we might want to add in a few "similar" models in terms of performance.

Another alternative is also just to examine if the two distributions are using a bootrapped significance test:

<details>
<summary>Example</summary>
<br>

```
# determine if two distributions are significantly different

import numpy as np

e5_base_v_measure_twenty_news_v3 = [
    0.3581229581747901,
    0.3714322776700291,
    0.3748205254710915,
    0.42182348635393163,
    0.34802237530625957,
    0.3381987901130682,
    0.4098039181016021,
    0.30702158765390447,
    0.3804081246688891,
    0.3542085586629238,
]

e5_base_v_measure_twenty_news_v1 = [
    0.3408194492148074,
    0.3843101577772905,
    0.2882426609687635,
    0.3192739638121998,
    0.3290549575662491,
    0.3921994383696271,
    0.3456072517347633,
    0.3585725815428564,
    0.35036874936283036,
    0.38337129020566746,
]


def test_significance():
    """Test if two distributions are significantly different using a bootstrapping method"""
    # Set the random seed for reproducibility
    np.random.seed(42)

    # Compute the difference between the two distributions
    diff = np.mean(e5_base_v_measure_twenty_news_v3) - np.mean(
        e5_base_v_measure_twenty_news_v1
    )

    # Concatenate the two distributions
    concat = np.concatenate(
        [e5_base_v_measure_twenty_news_v3, e5_base_v_measure_twenty_news_v1]
    )

    # Number of bootstrap samples
    n_samples = 10000

    # Initialize the bootstrap samples
    bootstrap_diff = np.zeros(n_samples)

    # Generate the bootstrap samples
    for i in range(n_samples):
        # Generate a bootstrap sample
        bootstrap_sample = np.random.choice(concat, size=len(concat), replace=True)

        # Compute the difference between the two bootstrap samples
        bootstrap_diff[i] = np.mean(
            bootstrap_sample[: len(e5_base_v_measure_twenty_news_v3)]
        ) - np.mean(bootstrap_sample[len(e5_base_v_measure_twenty_news_v3) :])

    # Compute the p-value
    # I.e. what is the probability of observing a difference as extreme as the one we observed 
    # given that the null hypothesis is true (i.e. the two distributions are the same)
    p_value = np.mean(bootstrap_diff >= diff)
    print(f"The p-value is: {p_value}")

    return p_value


test_significance() # 0.1254 -> not significant
```
</details>

We might also want to use significant rank (the code above works for that as well) before we do the spearman, but that is probably not the next step.

KennethEnevoldsen (commented):
This looks great @isaac-chung! A few surprising results here. 

I might be afraid that the new method (resampling many documents) produces results closer to the original (which is nice), but what we are really interested in is that they produce the same model ordering and with the same (or better) power to differentiate between similar models.

Though resampling many times does seem to be quite cheap.

scripts/mteb_english_cluster_spearman.py
KennethEnevoldsen: So would probably estimate a table like so:

<!DOCTYPE html>
Task | Spearman | Avg. Speedup
-- | -- | --
Task A | 0.8833 | 32.05x
Task B | 0.9100 | 42.35x
Task C | 0.9000 | 52.77x

(values are random)

mteb/tasks/Clustering/eng/BiorxivClusteringP2P.py
```diff
@@ -10,7 +10,7 @@
 
 class BiorxivClusteringP2PFast(AbsTaskClusteringFast):
     metadata = TaskMetadata(
-        name="BiorxivClusteringP2P.v2",
+        name="BiorxivClusteringP2P.v3-2",
```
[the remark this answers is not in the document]
isaac-chung: Yeah those are meant to be temporary and easy to tell apart since 4 options are presented here. We will keep v2 at the end of the discussion for sure.

scripts/run_mteb_english_cluster.py
```diff
@@ -0,0 +1,48 @@
...
+from mteb import MTEB, get_model
+from mteb.models.e5_models import e5_mult_small, e5_mult_base, e5_mult_large
+
+logging.basicConfig(level=logging.INFO)
+
+logger = logging.getLogger("main")
+
+TASK_LIST_CLUSTERING = [
+    # "ArxivClusteringP2P",  # hierarchical
+    # "ArxivClusteringS2S",  # hierarchical
```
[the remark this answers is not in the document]
isaac-chung: Their fast implementation are covered under the hierarchical side:
```
superseeded_by = "ArXivHierarchicalClusteringP2P"
```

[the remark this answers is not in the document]
KennethEnevoldsen: Ahh yes, but the ranking they produce should still be similar I suppose?

scripts/mteb_english_cluster_spearman.py
isaac-chung: > So would go for a table like so:
> 
> Task	Spearman	Avg. Speedup
> Task A	0.8833	32.05x
> Task B	0.9100	42.35x
> Task C	0.9000	52.77x
> (values are random)

Ah I see now - just pushed a commit to build a table like this, and pasted the results into the PR description. Option c is able to get us closest to perfect all across, and it seems like `BiorxivClusteringP2P` is the same across all options. 

For immediate next steps, I can look into why `BiorxivClusteringP2P` isn't giving perfect scores, and then add runs for a "close-relative" model (one model might be enough? large took quick a long time to run). What do you think?

isaac-chung: For BiorxivClusteringP2P, e5-base ranks 1st in both cases, and when in ClusteringFast, e5-small and e5-large swap ranks (2nd to 3rd and vice versa).

scripts/run_mteb_english_cluster.py
```diff
@@ -0,0 +1,48 @@
...
+for model in MODELS:
+    model_name = model.name
+    revision = model.revision
+
+    model = get_model(model_name=model_name, revision=revision)
+
+    eval_splits = ["test"]
+    evaluation = MTEB(
+        tasks=TASK_LIST, task_langs=["en"]
+    )  # Remove "en" for running all languages
```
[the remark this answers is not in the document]
isaac-chung: Switched.

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -138,9 +138,15 @@ def _evaluate_subset(
...
         )
-        all_v_scores = itertools.chain.from_iterable(v_measures.values())
-        mean_v_measure = np.mean(list(all_v_scores))
-        scores = {"v_measures": v_measures, "v_measure": float(mean_v_measure)}
+        v_measures = list(itertools.chain.from_iterable(all_v_scores.values()))
+
+        mean_v_measure = np.mean(v_measures)
+        v_std = np.std(v_measures)
+        scores = {
+            "v_measures": v_measures,
```
isaac-chung, line 146: Culprit for https://github.com/embeddings-benchmark/mteb/pull/892/files#r1632039634
```suggestion
            "v_measures": all_v_scores,
```

scripts/mteb_english_cluster_spearman.py
KennethEnevoldsen: 1) It might just be that in both cases they are not significantly different from each other (seems like scores are very close) - checking for significant differences might solve the issue here

3) One might be enough. Alternatively we can compare the smaller models:

- [paraphrase-multilingual-MiniLM-L12-v2](https://huggingface.co/sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2) - similar to mult. e5 small
- [all-MiniLM-L12-v2](https://huggingface.co/sentence-transformers/all-MiniLM-L12-v2) similar to mult e5 large

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: Ah right so we just want to make it such that all have the "Level" regardless of whether it is hierarchical or not.

results/intfloat__multilingual-e5-base/d13f1b27baf31030b7fd040960d60d909913633f/ArxivClusteringP2P.json
```diff
@@ -0,0 +1,53 @@
...
+    "test": [
+      {
+        "hf_subset": "default",
+        "languages": [
+          "eng-Latn"
+        ],
+        "main_score": 0.43349302457921374,
+        "v_measure": 0.43349302457921374,
+        "v_measure_std": 0.1364251955618546,
+        "v_measures": [
```
isaac-chung, line 16: Reminder for myself to add Level 0

scripts/mteb_english_cluster_spearman.py
isaac-chung: Added the check for significance as well - I'm not entirely sure how to interpret these results just yet (sans coffee). In the meantime, I could add runs for `paraphrase-multilingual-MiniLM-L12-v2` (good call for a close one for small instead).

KennethEnevoldsen: Ahh right so it seems like I might have gone a bit confused with 1) here. I think the best way is to really to a spearman on the ranks, e.g. for:

```
# three model scores with 10 samples in each
m_1 = np.random.normal(1, 1, 10) # 1 and 2 are similar
m_2 = np.random.normal(1.1, 1, 10) 
m_3 = np.random.normal(5, 1, 10) # 3 is much better

# so we expect rank 3>2>1, but significant rank 3>2=1 (as from only 10 examples we can tell m1 and m2 apart)
```

I actually made the function for it (see below) as we also need it for the task selection.

<details>
<summary>Example</summary>
<be>

```{python}
import numpy as np


def test_significance(dist1, dist2, n_samples=10_000):
    """Test if two distributions are significantly different using a bootstrapping method

    # sanity check:
    # should be significant
    dist1=np.random.normal(1, 1, 10)
    dist2=np.random.normal(4, 1, 10)
    n_samples=10_000
    test_significance(dist1, dist2, n_samples)
    # should not be significant
    test_significance(np.random.normal(1, 1, 10), np.random.normal(1, 1, 10), n_samples=10_000)
    """
    # Compute the difference between the two distributions
    diff = abs(np.mean(dist1) - np.mean(dist2))

    # Concatenate the two distributions
    concat = np.concatenate([dist1, dist2])

    # Initialize the bootstrap samples
    bootstrap_diff = np.zeros(n_samples)

    # Generate the bootstrap samples
    for i in range(n_samples):
        # Generate a bootstrap sample
        bootstrap_sample = np.random.choice(concat, size=len(concat), replace=True)

        # Compute the difference between the two bootstrap samples
        bootstrap_diff[i] = abs(
            np.mean(bootstrap_sample[: len(dist1)])
            - np.mean(bootstrap_sample[len(dist1) :])
        )

    # Compute the p-value
    # I.e. what is the probability of observing a difference as extreme as the one we observed
    # given that the null hypothesis is true (i.e. the two distributions are the same)
    p_value = np.mean(bootstrap_diff >= diff)
    print(f"The p-value is: {p_value}")

    return p_value


def compute_significant_rank(scores: dict, threshold=0.05):
    """Compute significant rank for models.

    Example:
        ```
        scores = {"model1": np.random.normal(1, 1, 10) # 1 and 2 are similar
                "model2": np.random.normal(1.1, 1, 10)
                "model3": np.random.normal(5, 1, 10) # 3 is much better
        ranks = ranks(compute_significant_rank)
        print(ranks)
        # {
        "models": ["model3", "model2", "model1"],
        # "significant rank": [1, 2, 2],
        # "rank": [1, 2, 3],
        # }
        ```
    """
    ranks = {}

    mean_scores = [(m, np.mean(s)) for m, s in scores.items()]
    mean_scores = sorted(mean_scores, key=lambda x: -x[1])  # higher is first

    ranks["models"], _ = zip(*mean_scores)
    ranks["models"] = list(ranks["models"])
    ranks["rank"] = list(range(1, len(ranks["models"]) + 1))

    pairs = [
        (ranks["models"][i], ranks["models"][i + 1])
        for i in range(len(mean_scores) - 1)
    ]

    rank = 1
    ranks["significant rank"] = [rank]  # first model always get rank 1
    for p1, p2 in pairs:
        # test if the two models are significantly different
        p_value = test_significance(scores[p1], scores[p2])

        if p_value < threshold:
            rank += 1

        ranks["significant rank"].append(rank)

    return ranks


scores = {
    "model1": np.random.normal(1, 1, 10),  # 1 and 2 are similar
    "model2": np.random.normal(1.1, 1, 10),
    "model3": np.random.normal(5, 1, 10),  # 3 is much better
}
ranks = compute_significant_rank(scores)
assert ranks == {
    "models": ["model3", "model2", "model1"],
    "rank": [1, 2, 3],
    "significant rank": [1, 2, 2],
}

# we can then do spearman(ranks["significant rank"], ranks_from_another_task["significant rank"])
```

<details>

isaac-chung: Thanks for the details! I felt quite lost before. Let me use this in `scripts/mteb_english_cluster_spearman.py`. I could imagine some operations can be omitted as they can be read from the result object, e.g. average score. Will report back.

KennethEnevoldsen: Hmm wow this turns out to be a much harder problem that I imagined it to be. I redid the code in `compare-scores-clustering-fast-kenneth`. To create the same table but adding a column for significant rank. Also added a column with notes:


| Model | Spearman | Significant Spearman | Speedup | Notes |
| --- | --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.8 | 1.0 | 32.82x | complete agreement |
| BiorxivClusteringS2S        | 1.0 | 0.95 | 14.90x | close to perfect agreement |
| MedrxivClusteringP2P        | 0.8 | nan | 16.43x | no significant difference in classical method. 1, 1, 1, 1 vs 1, 2, 2, 1  |
| MedrxivClusteringS2S        | 1.0 | 0.77 | 7.66x | The new method is better able to differentiate between models [1, 2, 1, 1] vs [1, 4, 3, 2] though agreement on mean rank |
| RedditClustering            | 0.4 | 0.258 | 68.47x | quite different predictions, see scores below for discussion |
| RedditClusteringP2P         | 1.0 | nan | 145.88x | fast approach produces 'rank': [1, 2, 3, 4], 'significant_rank': [1, 2, 3, 4] while classic produces 'rank': [1, 2, 3, 4], 'significant_rank': [1, 1, 1, 1] |
| StackExchangeClustering     | -0.2 | -0.5 | 42.52x | disagreement, again see below | 
| StackExchangeClusteringP2P  | 0.8 | nan | 27.02x | [1, 1, 1, 1] vs [2, 2, 1, 1] (fast) |
| TwentyNewsgroupsClustering  | 1.0 | 0.89 | 8.50x | close to perfect agreement | 

Generally, it seems like the two approaches agree. When only moderately agree it is either due to small variations (e.g. BiorxivClusteringP2P is due to small variation, causing random models to end up on top) or the fast approach is better able to differentiate between the models (e.g. for MedrxivClusteringP2P or StackExchangeClusteringP2P there is little variation, but the fast approach is better to spread it apart. In both cases they generally agree on non-significant rank). 

When there is a high disagreement RedditClustering and StackExchangeClustering it mainly seems to concern the placement of "paraphrase-multilingual-MiniLM-L12-v2", which fast generally ranks higher than the classic implementation (see scores below). This might be worth looking into whether the discrepancy in the new sampling methods might cause this (or if we can toggle something to make them more aligned.

<details>
<summary>Scores from RedditClustering</summary>

{'models': ['multilingual-e5-large', 'multilingual-e5-base', 'paraphrase-multiling...iLM-L12-v2', 'multilingual-e5-small'], 'rank': [1, 2, 3, 4], 'significant_rank': [1, 2, 2, 2]}
{
    "paraphrase-multilingual-MiniLM-L12-v2": [
        0.37277761581154506,
        0.46409492328260454,
        0.3825830217942452,
        0.4188753999297151,
        0.3887993665535172,
        0.3919622314341335,
        0.468578152667782,
        0.40847137153578084,
        0.4175220777386184,
        0.4154785516440778,
        0.391485810594391,
        0.48378790117918613,
        0.46772742914847565,
        0.43365677706385497,
        0.5330771268772484,
        0.39676239704994926,
        0.4132305320441305,
        0.41764820598761504,
        0.4095582509830173,
        0.40708410235702874,
        0.3829832154411731,
        0.39776158641365755,
        0.5720250244093886,
        0.4376112135122882,
        0.386613274337495,
    ],
    "multilingual-e5-small": [
        0.43924415547815493,
        0.4648204019496003,
        0.3634670245059431,
        0.34996271631961057,
        0.3755802403823603,
        0.36625814922396,
        0.39978514515292285,
        0.35586798358803196,
        0.33131383396695796,
        0.37495760001729866,
        0.3779173495615085,
        0.49280279451125475,
        0.39631239594264267,
        0.41286392832069285,
        0.5727645148361832,
        0.34535579003657113,
        0.4344376870664589,
        0.5522002199174861,
        0.3605330364708775,
        0.35072864937515813,
        0.31824360817273484,
        0.3617872826280434,
        0.4803591464281226,
        0.40640128709706164,
        0.34492355007413544,
    ],
    "multilingual-e5-base": [
        0.3899135788336536,
        0.5222611855652423,
        0.4000214183807749,
        0.41669890655780106,
        0.4350556771316963,
        0.43546212061602335,
        0.4704746876298788,
        0.37845532083531574,
        0.3986206963164758,
        0.3968334546709837,
        0.4214596817687342,
        0.5068502845662192,
        0.3973528487984853,
        0.4512344962141797,
        0.5125666672366576,
        0.38427016639848643,
        0.43924342920164716,
        0.5078631117279251,
        0.42861193894654226,
        0.3948749829862104,
        0.3757822633989318,
        0.3952698829310342,
        0.5449360560026447,
        0.41371502319555176,
        0.36971412848087937,
    ],
    "multilingual-e5-large": [
        0.5480269333378893,
        0.4983641774142771,
        0.40428009229498907,
        0.4572569102977324,
        0.5016433521339552,
        0.4200969919045957,
        0.4874222428435611,
        0.424774792761486,
        0.4271797683501383,
        0.40745929515934043,
        0.45283217551666843,
        0.5183701127997492,
        0.4417487747102521,
        0.4659079559947688,
        0.6310998875955774,
        0.4241757302028172,
        0.4936740621416464,
        0.6047135837401434,
        0.456872630260866,
        0.4002594273625891,
        0.37796926993435404,
        0.4154221455363117,
        0.5618674876148675,
        0.47398007910060497,
        0.4313101897959134,
    ],
}

# fast
{'models': ['paraphrase-multiling...iLM-L12-v2', 'multilingual-e5-large', 'multilingual-e5-base', 'multilingual-e5-small'], 'rank': [1, 2, 3, 4], 'significant_rank': [1, 2, 3, 4]}
{
    "paraphrase-multilingual-MiniLM-L12-v2": [
        0.46622833605283226,
        0.4615888627459528,
        0.4669404239007183,
        0.428194799375356,
        0.45183970479940216,
        0.45910320537732935,
        0.4507809974866103,
        0.43758846284607045,
        0.43231853238924606,
        0.42471939281731574,
    ],
    "multilingual-e5-small": [
        0.3635198969630694,
        0.36944623478159166,
        0.3679617095580352,
        0.408354852945145,
        0.3847758827453475,
        0.3886889601017186,
        0.3877530404796728,
        0.36209962807561763,
        0.3672635395156691,
        0.39602123172532905,
    ],
    "multilingual-e5-base": [
        0.4070745141027744,
        0.3961625113995771,
        0.4462316396425147,
        0.4189368671617159,
        0.38190371435856546,
        0.40135636321487783,
        0.39621390462646233,
        0.4269540563592255,
        0.36336432328821466,
        0.43660177382061016,
    ],
    "multilingual-e5-large": [
        0.41197906615410573,
        0.4142187114361218,
        0.4143588005948501,
        0.4502648851328288,
        0.4111933000079036,
        0.42271728211283277,
        0.4442983627188738,
        0.4392277463204552,
        0.44963089957757923,
        0.41537002306688997,
    ],
}

</details>


<details>
<summary>Scores from StackExchangeClustering </summary>


# classic
{
    "models": [
        "multilingual-e5-large",
        "multilingual-e5-base",
        "multilingual-e5-small",
        "paraphrase-multilingual-MiniLM-L12-v2",
    ],
    "rank": [1, 2, 3, 4],
    "significant_rank": [1, 2, 2, 3],
}
{
    "paraphrase-multilingual-MiniLM-L12-v2": [
        0.511467632515496,
        0.48436843024496756,
        0.4460256425335859,
        0.5282684019626085,
        0.5014254912400526,
        0.4503556191599693,
        0.4790446299972,
        0.5690335673943024,
        0.47265317976614685,
        0.47742436086149176,
        0.5980683460660893,
        0.5135362965566143,
        0.5742897900021677,
        0.544466175621605,
        0.44390707965064874,
        0.4820011033553236,
        0.46792590424045855,
        0.4534974347588226,
        0.45444640000971614,
        0.4413160601623461,
        0.47538818140659894,
        0.5340349653884281,
        0.47049165397721143,
        0.4559281762911649,
        0.49324931415716167,
    ],
    "multilingual-e5-small": [
        0.5149556026506005,
        0.5741545702686627,
        0.4469995562276755,
        0.5678006543665266,
        0.4795679317121149,
        0.5022047071386572,
        0.45589623369775,
        0.5537312345148071,
        0.5253922380464688,
        0.5416090263799735,
        0.7025820670938749,
        0.6087271458345636,
        0.6543947895366301,
        0.5845274688607025,
        0.5143120405618706,
        0.5246145874584811,
        0.5048436197861124,
        0.5132818015677498,
        0.4943860431578122,
        0.5039302350689131,
        0.5678584103975669,
        0.4647173198070034,
        0.4738303077323309,
        0.5388959950955831,
        0.5157292215290958,
    ],
    "multilingual-e5-base": [
        0.5506182631851375,
        0.6098472794008257,
        0.44884905130888936,
        0.5516385405128469,
        0.5194284248733142,
        0.4914320610158897,
        0.5029636446257895,
        0.6068214257212708,
        0.5511591838426464,
        0.5625517095297468,
        0.6248970053480281,
        0.6152856080318063,
        0.7018985364271394,
        0.5914797519615147,
        0.5229708722434635,
        0.5475301598702013,
        0.528510437313286,
        0.538358895793387,
        0.5217772773576943,
        0.5409771373530979,
        0.5812137020536748,
        0.48807763749190974,
        0.5173798706624534,
        0.5828855479501713,
        0.527994697972928,
    ],
    "multilingual-e5-large": [
        0.6295997789610954,
        0.6283153032775785,
        0.5093546397212798,
        0.5799467525525586,
        0.5390360305126893,
        0.5349524240915096,
        0.5202410268896647,
        0.6656177534416677,
        0.5951424997183259,
        0.6079780427410808,
        0.6488748469360486,
        0.699905303943518,
        0.680762948936147,
        0.598563961159552,
        0.5598941308286685,
        0.5554098194379443,
        0.5646274661493497,
        0.5565892090916954,
        0.556013393724937,
        0.5663514278951832,
        0.5830354323824856,
        0.5295394730230879,
        0.523939052972823,
        0.600046291358453,
        0.559543313549954,
    ],
}

# fast
{
    "models": [
        "paraphrase-multilingual-MiniLM-L12-v2",
        "multilingual-e5-large",
        "multilingual-e5-base",
        "multilingual-e5-small",
    ],
    "rank": [1, 2, 3, 4],
    "significant_rank": [1, 2, 2, 3],
}
{
    "paraphrase-multilingual-MiniLM-L12-v2": [
        0.5078775025919487,
        0.5328795773287571,
        0.5041532177679843,
        0.5039620209185893,
        0.5172628532940583,
        0.4974148083091552,
        0.5083817280746857,
        0.5167869506766196,
        0.5136630874260778,
        0.5138200271461127,
    ],
    "multilingual-e5-small": [
        0.4689928822090719,
        0.4850198654917769,
        0.4748777356624822,
        0.4835000837721761,
        0.4849813877636553,
        0.482523135937022,
        0.47743696287293425,
        0.47947306477275625,
        0.4790179230592838,
        0.4855516953833751,
    ],
    "multilingual-e5-base": [
        0.5003792121341227,
        0.4803255090234387,
        0.48370283890423094,
        0.479885642927048,
        0.4841346405132491,
        0.5062788279595188,
        0.4920306068212876,
        0.48991023964788155,
        0.4892659983833159,
        0.4822255766130806,
    ],
    "multilingual-e5-large": [
        0.4865439591592704,
        0.4939315831846357,
        0.4916113028118716,
        0.4864586030896486,
        0.5127818968431926,
        0.5115010761143769,
        0.49921994393999636,
        0.4949475949585376,
        0.5026455579663558,
        0.48891647399779137,
    ],
}

</details>


I think the next best steps might be to:

- examine if we can increase max_documents_to_embed (or some other thing) to better align the old and new approaches (especially on the two edge cases)
- currently, the table above does not convey a totally clear image, it might be better plotting it in a meaningful way (no ideas atm)
- It might be worth running more models to see if this pattern is a case of just one oddity rather than a general trend. (we can just use a bunch of smaller models) english e5 small and e5 small v2, all-mini-lm etc.

isaac-chung: Wow thanks. Your update is much clearer on the ranks/sig rank portion. I've incorporated it in the next commit. 

> - examine if we can increase max_documents_to_embed (or some other thing) to better align the old and new approaches (especially on the two edge cases)
> - currently, the table above does not convey a totally clear image, it might be better plotting it in a meaningful way (no ideas atm)
> - It might be worth running more models to see if this pattern is a case of just one oddity rather than a general trend. (we can just use a bunch of smaller models) english e5 small and e5 small v2, all-mini-lm etc.

I can take a look at this in the order 3 -> 1 -> 2 to buy us some time to think about plotting.

KennethEnevoldsen: yea not sure what to do with the plotting (@x-tabdeveloping usually has some good ideas)

isaac-chung: Here's running more models (No. 3) in the last commit [61ff34a](https://github.com/embeddings-benchmark/mteb/pull/892/commits/61ff34aae8727a008333ef48a5e1fca0658a30c1):

| Model | Spearman | Significant Spearman | Speedup |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.7500 | 0.8981 | 31.13x |
| BiorxivClusteringS2S        | 1.0000 | 0.9706 | 12.02x |
| MedrxivClusteringP2P        | 0.7857 | 0.6831 | 15.76x |
| MedrxivClusteringS2S        | 1.0000 | nan | 6.40x |
| RedditClustering            | 0.5714 | 0.6236 | 57.35x |
| RedditClusteringP2P         | 0.9643 | nan | 132.10x |
| StackExchangeClustering     | 0.1071 | -0.1049 | 37.17x |
| StackExchangeClusteringP2P  | 0.9643 | nan | 25.44x |
| TwentyNewsgroupsClustering  | 1.0000 | 0.8141 | 7.07x |

KennethEnevoldsen: I have sent you a few suggestions on plotting as well.

isaac-chung: (No.1) Increasing the max_documents_to_embed by 2x to 4096 (cfb2f95) yields the following. 

| Model | Spearman | Significant Spearman | Speedup |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.7500 | 0.8847 | 17.12x |
| BiorxivClusteringS2S        | 1.0000 | 1.0000 | 9.31x |
| MedrxivClusteringP2P        | 0.8929 | 1.0000 | 8.52x |
| MedrxivClusteringS2S        | 1.0000 | 0.6417 | 4.87x |
| RedditClustering            | 0.5714 | 0.6236 | 45.78x |
| RedditClusteringP2P         | 1.0000 | nan | 81.98x |
| StackExchangeClustering     | 0.2500 | 0.1039 | 31.83x |
| StackExchangeClusteringP2P  | 0.9643 | nan | 15.02x |
| TwentyNewsgroupsClustering  | 1.0000 | 0.7913 | 6.00x |

<br/>----------------

max_documents_to_embed = 4096 and max_documents_per_cluster = 32768.

| Model | Spearman | Significant Spearman | Speedup |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.6786 | 1.0000 | 16.67x |
| BiorxivClusteringS2S        | 1.0000 | 0.9630 | 7.74x |
| MedrxivClusteringP2P        | 1.0000 | 0.6831 | 8.31x |
| MedrxivClusteringS2S        | 1.0000 | 0.6355 | 4.11x |
| RedditClustering            | 0.5714 | 0.6236 | 37.79x |
| RedditClusteringP2P         | 1.0000 | nan | 76.54x |
| StackExchangeClustering     | 0.2500 | -0.1030 | 26.42x |
| StackExchangeClusteringP2P  | 0.9643 | nan | 14.27x |
| TwentyNewsgroupsClustering  | 1.0000 | 0.8767 | 4.91x |

<br/>----------------

max_documents_to_embed = 4096 and max_documents_per_cluster = 65536.

| Model | Spearman | Significant Spearman | Speedup |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.6786 | 1.0000 | 15.76x |
| BiorxivClusteringS2S        | 1.0000 | 0.9630 | 5.97x |
| MedrxivClusteringP2P        | 0.8929 | 0.6831 | 7.87x |
| MedrxivClusteringS2S        | 1.0000 | nan | 3.18x |
| RedditClustering            | 0.6071 | 0.6236 | 28.74x |
| RedditClusteringP2P         | 0.9643 | nan | 69.14x |
| StackExchangeClustering     | 0.1071 | 0.0000 | 20.12x |
| StackExchangeClusteringP2P  | 1.0000 | nan | 13.11x |
| TwentyNewsgroupsClustering  | 0.9643 | 0.9303 | 3.78x |

isaac-chung: This leads me to think that it may not be about an absolute number for a n_samples to embed, and may be more related to the proportion to the original n_samples. We observe that 

1. The smaller sets e.g. BiorxivClusteringP2P reach close to perfect agreement at 2-3% examples embedded. 
2. Big sets like RedditClustering require 8k samples (2%) embedded to reach moderate agreement.
3. StackExchange might need 4% (16k running).

So instead of setting max_documents_to_embed, we could instead set max_percentage_to_embed. This could preserve higher speedups in the other tasks as well.

| Task | Orig Num | Fast Num | Fraction |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        |   75000 |  2048 | 0.027 |
| BiorxivClusteringP2P        |   75000 |  4096 | 0.055 |
| BiorxivClusteringP2P        |   75000 |  8192 | 0.109 |
| BiorxivClusteringP2P        |   75000 | 16384 | 0.218 |
| BiorxivClusteringS2S        |   75000 |  2048 | 0.027 |
| BiorxivClusteringS2S        |   75000 |  4096 | 0.055 |
| BiorxivClusteringS2S        |   75000 |  8192 | 0.109 |
| BiorxivClusteringS2S        |   75000 | 16384 | 0.218 |
| MedrxivClusteringP2P        |  375000 |  2048 | 0.005 |
| MedrxivClusteringP2P        |  375000 |  4096 | 0.011 |
| MedrxivClusteringP2P        |  375000 |  8192 | 0.022 |
| MedrxivClusteringP2P        |  375000 | 16384 | 0.044 |
| MedrxivClusteringS2S        |  375000 |  2048 | 0.005 |
| MedrxivClusteringS2S        |  375000 |  4096 | 0.011 |
| MedrxivClusteringS2S        |  375000 |  8192 | 0.022 |
| MedrxivClusteringS2S        |  375000 | 16384 | 0.044 |
| RedditClustering            |  420464 |  2048 | 0.005 |
| RedditClustering            |  420464 |  4096 | 0.010 |
| RedditClustering            |  420464 |  8192 | 0.019 |
| RedditClustering            |  420464 | 16384 | 0.039 |
| RedditClusteringP2P         |  459399 |  2048 | 0.004 |
| RedditClusteringP2P         |  459399 |  4096 | 0.009 |
| RedditClusteringP2P         |  459399 |  8192 | 0.018 |
| RedditClusteringP2P         |  459399 | 16384 | 0.036 |
| StackExchangeClustering     |  373850 |  2048 | 0.005 |
| StackExchangeClustering     |  373850 |  4096 | 0.011 |
| StackExchangeClustering     |  373850 |  8192 | 0.022 |
| StackExchangeClustering     |  373850 | 16384 | 0.044 |
| StackExchangeClusteringP2P  |   75000 |  2048 | 0.027 |
| StackExchangeClusteringP2P  |   75000 |  4096 | 0.055 |
| StackExchangeClusteringP2P  |   75000 |  8192 | 0.109 |
| StackExchangeClusteringP2P  |   75000 | 16384 | 0.218 |
| TwentyNewsgroupsClustering  |   59545 |  2048 | 0.034 |
| TwentyNewsgroupsClustering  |   59545 |  4096 | 0.069 |
| TwentyNewsgroupsClustering  |   59545 |  8192 | 0.138 |
| TwentyNewsgroupsClustering  |   59545 | 16384 | 0.275 |

KennethEnevoldsen: Yea it would be great if users could use say MTEB(eng, 0.25%) (reduce the size to 25%), however, that raises a problem for retrieval. We could simply supply either. So either a percent or a number. Would probably prefer the default to be set at a number.

It really seems like RedditClustering and StackExchangeClustering are the only two that require a larger sample (probably due to many classes and short documents). What about we just turn them up to let's say 32.768 (just to see if we can hit above the .80 correlation).

Also related to significant rank I realize that the way it does it atm. is slightly problematic.

it compared the significance between:
1,2
2,3 
3,4
etc.

which all might be non-significant (e.g. rank 1, 1, 1, 1...) but the difference between 1, 4 might be so the rank should really be (1, 1, 1, 2). I can make the changes to that. This way the can probably remove the nan's (I also believe that approach is better).

isaac-chung: I think this percentage approach would do well in ClusteringFast. I would defer retrieval-fast to @orionw. I'll run the 4% case to at least at an idea of how it affects the spearman and speedup columns, then incorporate your changes for the significant rank column.

orionw: There are two knobs for retrieval to lower: queries and docs per query.  We can easily subsample queries to X% but for the docs per query we'd need to upload and store hard negatives.  That's not out of the question because we're going to likely be making those anyways.

KennethEnevoldsen: So redid the analysis using the new formulation of significant spearman:

With the latest version of the branch we get:

| Model | Spearman | Significant Spearman | Speedup |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.7500 | 1.0000 | 8.84x |
| BiorxivClusteringS2S        | 1.0000 | 0.9543 | 6.16x |
| MedrxivClusteringP2P        | 0.9286 | 1.0000 | 4.43x |
| MedrxivClusteringS2S        | 1.0000 | nan | 3.27x |
| RedditClustering            | 0.6429 | 0.6417 | 31.63x |
| RedditClusteringP2P         | 1.0000 | nan | 45.95x |
| StackExchangeClustering     | 0.4286 | 0.3208 | 23.76x |
| StackExchangeClusteringP2P  | 0.9643 | nan | 8.13x |
| TwentyNewsgroupsClustering  | 1.0000 | 0.9303 | 4.34x |

And then using the new version significance we get:

| Model | Spearman | Significant Spearman | Speedup |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.7500 | 1.0000 | 8.84x |
| BiorxivClusteringS2S        | 1.0000 | 0.9820 | 6.16x |
| MedrxivClusteringP2P        | 0.9286 | 0.6295 | 4.43x |
| MedrxivClusteringS2S        | 1.0000 | 0.9533 | 3.27x |
| RedditClustering            | 0.6429 | 0.8141 | 31.63x |
| RedditClusteringP2P         | 1.0000 | 0.6236 | 45.95x |
| StackExchangeClustering     | 0.4286 | 0.7276 | 23.76x |
| StackExchangeClusteringP2P  | 0.9643 | 0.8819 | 8.13x |
| TwentyNewsgroupsClustering  | 1.0000 | 0.9630 | 4.34x |

generally, I would say that it is notably better (glad to see), with the exception of MedrxivClusteringP2P.

Where it goes from [1, 2, 2, 2, 2, 2, 2], [1, 2, 2, 2, 2, 2, 2]  to [1, 2, 2, 2, 2, 2, 2], [1, 2, 2, 3, 3, 4, 4] (second is fast). Which is what we should suspect, but it is probably due to the classical method not being able to differentiate the two. Adding more model should show that there is in fact a decent agreement (which the actual rank also shows)

In fact I am considering if [1, 2, 2, 2, 2, 2, 2], [1, 2, 2, 3, 3, 4, 4] is really a disagreement in rank - both approaches agree on the model rank when ignoring for minor variation (that is the significance part), but on is better at segregrating model scores (higher measurement quality).

I think we are getting to a point where we need some figures though to support our arguments.

isaac-chung: This is great! I'm running RedditClustering, StackExchangeClustering, and MedrxivClusteringP2P with 32768 samples. I noticed that the newest verion of MedrxivClusteringP2P only had 17K samples, whereas the prev version (classic) had 375k. I might rerun tasks like BiorxivClustering later as well just to confirm.

Do you have the updated version of significance in your branch? I'd love to give that a spin. Would help with figures as well.

KennethEnevoldsen: @isaac-chung I have added the plots + script to the branch (they can be prettier, but give an overview). 

From them, it seems like we can get the right table by setting a percentage (e.g., 4% seems reasonable) and running a few more models (probably on the higher end of the English benchmark to get a wider spread).

isaac-chung: Forgot to refresh the page, I see them now, thanks again. Let me take a look at the plots.

Here is the table from the latest commit: ( 4% for every task except for RedditClustering, StackExchangeClustering, and MedrxivClusteringP2P with 32768)

| Model | Spearman | Significant Spearman | Speedup |
| --- | --- | --- | --- |
| BiorxivClusteringP2P        | 0.8214 | 0.8981 | 30.56x |
| BiorxivClusteringS2S        | 1.0000 | 0.9633 | 12.13x |
| MedrxivClusteringP2P        | 0.9286 | 0.6417 | 1.15x |
| MedrxivClusteringS2S        | 1.0000 | 0.9136 | 8.44x |
| RedditClustering            | 0.8929 | 0.9129 | 11.37x |
| RedditClusteringP2P         | 1.0000 | 0.6179 | 22.46x |
| StackExchangeClustering     | 0.8571 | 0.9120 | 9.52x |
| StackExchangeClusteringP2P  | 1.0000 | 0.8819 | 19.38x |
| TwentyNewsgroupsClustering  | 1.0000 | 0.9425 | 7.05x |
| Average                     | 0.9444 | 0.8538 | 13.56x |

isaac-chung: > running a few more models (probably on the higher end of the English benchmark to get a wider spread).

@KennethEnevoldsen how about
- [e5-large-v2](https://huggingface.co/intfloat/e5-large-v2) for large
- [paraphrase-multilingual-mpnet-base-v2](https://huggingface.co/sentence-transformers/paraphrase-multilingual-mpnet-base-v2) for medium?

isaac-chung:
Looking ahead to what closes this PR, the following should be completed first:

- [ ] Move plotting scripts and plots to MMTEB scripts
- [ ] Run a few more models (https://github.com/embeddings-benchmark/mteb/pull/892#discussion_r1639429887)

Anything else I'm missing?

scripts/mteb_english_cluster_spearman.py
KennethEnevoldsen: Sounds fine to me. 

One point is that we might simply compare speedup using the current models and then we can re-use the results om MTEB from previous runs (that way we avoid spending a lot of time rerunning and we already have a decent estimate on the speedup)

KennethEnevoldsen: (would probably prefer a few more actually, if we are not running on MTEB this might be fine?)

isaac-chung: The fast version will need to be run regardless as the 4% rule is only introduced in this PR. However the biggest time saving will come from the classic results. 

Which models would you like to be run, in addition to what's been proposed here?

KennethEnevoldsen: I would select e5-base-v2, the mixedbread model, nomic-embed, labSE.

isaac-chung: Added most of these. It went OOM while running nomic_embed_text_v1 on a few datasets. I'll switch to a bigger machine to run the classic+v2 for those.

isaac-chung: Maybe https://huggingface.co/BAAI/bge-base-en-v1.5 instead as they have similar sizes

KennethEnevoldsen:
I think that it good

isaac-chung:
Added plotting scripts and plots [here](https://github.com/embeddings-benchmark/mmtebscripts/pull/1). Will remove them from this PR, then we should be good to merge.

isaac-chung:
@KennethEnevoldsen  just wanted a sanity ✅ before we merge, thanks!

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -91,11 +91,12 @@ class AbsTaskClusteringFast(AbsTask):
         labels: list[str] | list[list[str]]
     """
 
-    max_documents_to_embed = 16_384
-    max_documents_per_cluster = 2048
+    max_fraction_of_documents_to_embed = 0.04
```
KennethEnevoldsen, line 94: I would do:

```suggestion
    max_fraction_of_documents_to_embed = 0.04
    max_document_to_embed = None
```

we also need to document these in the abstask

KennethEnevoldsen: if both is None don't downsample.

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -91,11 +91,12 @@ class AbsTaskClusteringFast(AbsTask):
...
+    max_documents_per_cluster = 16_384
     n_clusters = 10
     k_mean_batch_size = 512
     max_depth = None
+    use_dataset_as_is = False
```
KennethEnevoldsen, line 99:
```suggestion
```

KennethEnevoldsen (approved):
A few minor comments otherwise looks good

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: Oh a note here is also that this will influence existing clustering tasks (some of which are downsampled before they are pushed to the hub). For these just using the current number is probably the best solution.

mteb/abstasks/AbsTaskClusteringFast.py
isaac-chung: This param is used in `RedditFastClusteringS2S` and `StackExchangeClusteringFast`, where the data is subsampled during `dataset_transform` to 32768 n_samples.

KennethEnevoldsen: I think if both of the above is None, that would be the same as having this true?

PR #893, opened 2024-06-07
Fix abstention metric for small datasets

hgissbkh:
Hi MTEB team!
I recently added abstention/calibration metrics to the Retrieval and Reranking tasks and just noticed that there was a slight issue when evaluating on very small datasets (less than 10 samples).
I've made the necessary changes to make it work in this edge case.
Cheers!


KennethEnevoldsen (approved):
Looks great! Thanks for the fix

PR #926, opened 2024-06-14
LLM2Vec models

vaibhavad:
## Checklist

### Adding a model checklist


mteb/models/llm2vec_models.py
```diff
@@ -0,0 +1,141 @@
...
+logger = logging.getLogger(__name__)
+
+
+def llm2vec_loader(**kwargs):
+    try:
+        from llm2vec import LLM2Vec
+    except ImportError:
+        raise ImportError(
+            "LLM2Vec is not installed. Please install it with `pip install llm2vec`."
+        )
```
KennethEnevoldsen, line 18:
```suggestion
        raise ImportError(
            "To use the LLM2Vec models `llm2vec` is required. Please install it with `pip install llm2vec`."
        )
```

KennethEnevoldsen (approved):
I assume you are testing the models atm. otherwise, everything else looks great.

Does it work for instruction retrieval tasks?

vaibhavad:
@KennethEnevoldsen - Yes, I was waiting for E5-instruct to finish so that I can re-use the same workflow for instruction following tasks.

The scores currently seem 2-3 points off our reported scores. I am currently inspecting that, but otherwise everything else is implemented.

KennethEnevoldsen:
Hmm odd, that is probably an error somewhere. Is that a general thing or is it for specific tasks?

KennethEnevoldsen:
> @KennethEnevoldsen - It was an error for retrieval tasks, it is fixed now. For exact replication of the numbers reported in the paper, the instruction needs to have semicolon at the end, e.g. Retrieve semantically similar text:. However, the performance numbers are very close with default implementation that does not have semicolon. I added the provision for the user to provide their own set of instructions. I can use that on the LLM2Vec repo to document how to replicate scores using mteb package.

You are more than welcome to implement the ":" in the LLM encode function if you wish

vaibhavad:
I think the existing list works fine. LLM2Vec used the same instructions as e5-mistral-instruct, so there is anyways a large overlap with the instructions provides. I added the modification for ending the instruction with a semicolon though.

Issue #582, opened 2024-04-26
pair classification inconsistencies


SaitejaUtpala:
```
class AbsTaskPairClassification(AbsTask):
    """Abstract class for PairClassificationTasks
    The similarity is computed between pairs and the results are ranked. Average precision
    is computed to measure how well the methods can be used for pairwise pair classification.

    self.load_data() must generate a huggingface dataset with a split matching self.metadata_dict["eval_splits"], and assign it to self.dataset. It must contain the following columns:
        sent1: list[str]
        sent2: list[str]
        labels: list[int]
    """

    def __init__(self, **kwargs):
        super().__init__(**kwargs)

    def _evaluate_monolingual(self, model, dataset, split="test", **kwargs):
        data_split = dataset[split][0]  # This causes error because it just gets first row of the split
        logging.getLogger(
            "sentence_transformers.evaluation.PairClassificationEvaluator"
        ).setLevel(logging.WARN)
        evaluator = PairClassificationEvaluator(
            data_split["sent1"], data_split["sent2"], data_split["labels"], **kwargs
        )
```

I am working #581 dataset, I found couple of potential issues in 'AbsTaskPairClassification' with 

-  data_split = dataset[split][0]  in _evaluate_monolingual, This causes error because it just gets first row of the split and doesn't whole dataset
-  PairClassificationEvaluator(
            data_split["sent1"], data_split["sent2"], data_split["labels"], **kwargs
        )
also it expects 'sent1', 'sent2', 'labels' instead of 'sentence1', 'sentence2' and 'label' (standard followed in STS and BiText Mining task)

loicmagne:
It's not an error but this is a legacy of how the initial pair classification datasets were formatted, for example [TwitterSemEval2015](https://huggingface.co/datasets/mteb/twittersemeval2015-pairclassification/tree/main):

```python
>>> d = load_dataset('mteb/twittersemeval2015-pairclassification')
Downloading data: 100%|██████████████████████████████████████████████████████████████████████████| 313k/313k [00:00<00:00, 1.35MB/s]
Generating test split: 100%|███████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  7.39 examples/s]
>>> d
DatasetDict({
    test: Dataset({
        features: ['sent1', 'sent2', 'labels'],
        num_rows: 1
    })
})
```

There is a single row, where each row contains a list of sentences. I agree this isn't a very good format and the naming is inconsistent with other tasks so it might make sense to change it

PR #581, opened 2024-04-26
Add Indic xnli pair classification

SaitejaUtpala:
## Checklist for adding MMTEB dataset

Reason for dataset addition:

- [ ] I have tested that the dataset runs with the `mteb` package.
- [ ] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [ ] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [ ] `intfloat/multilingual-e5-small`
- [ ] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [ ] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [ ] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [ ] Run tests locally to make sure nothing is broken using `make test`. 
- [ ] Run the formatter to format the code using `make lint`. 
- [ ] I have added points for my submission to the [points folder](https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md) using the PR number as the filename (e.g. `438.jsonl`).

PR #945, opened 2024-06-17
Fix pair classification inconsistency

dokato:
Addresses issue reported here #582. Brings consistency with standard followed in STS and BiText Mining tasks:  'sentence1', 'sentence2' and 'label'.

### DIFF
--- a/mteb/tasks/PairClassification/ara/ArEntail.py
+++ b/mteb/tasks/PairClassification/ara/ArEntail.py
@@ -47,8 +47,8 @@ class ArEntail(AbsTaskPairClassification):
         for split in self.metadata.eval_splits:
             _dataset[split] = [
                 {
-                    "sent1": self.dataset[split]["premise"],
-                    "sent2": self.dataset[split]["hypothesis"],
+                    "sentence1": self.dataset[split]["premise"],
+                    "sentence2": self.dataset[split]["hypothesis"],
                     "labels": self.dataset[split]["label"],
                 }
             ]
--- a/mteb/tasks/PairClassification/zho/CMTEBPairClassification.py
+++ b/mteb/tasks/PairClassification/zho/CMTEBPairClassification.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from mteb.abstasks.AbsTaskPairClassification import AbsTaskPairClassification
 from mteb.abstasks.TaskMetadata import TaskMetadata
 
-
+# TODO sentence1
 class Ocnli(AbsTaskPairClassification):
     metadata = TaskMetadata(
         name="Ocnli",
--- a/mteb/tasks/PairClassification/zho/CMTEBPairClassification.py
+++ b/mteb/tasks/PairClassification/zho/CMTEBPairClassification.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from mteb.abstasks.AbsTaskPairClassification import AbsTaskPairClassification
 from mteb.abstasks.TaskMetadata import TaskMetadata
 
-# TODO sentence1
+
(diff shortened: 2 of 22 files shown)


mteb/tasks/PairClassification/zho/CMTEBPairClassification.py
```diff
@@ -64,48 +68,52 @@ class Cmnli(AbsTaskPairClassification):
...
-    pages = "4762--4772",
-}""",
+            title = "{CLUE}: A {C}hinese Language Understanding Evaluation Benchmark",
+            author = "Xu, Liang  and
+            Hu, Hai  and
+            Zhang, Xuanwei  and
+            Li, Lu  and
+            Cao, Chenjie  and
+            Li, Yudong  and
+            Xu, Yechen  and
```
KennethEnevoldsen, line 78: I don't think we should change the bibtex in any of these cases.

mteb/tasks/PairClassification/ara/ArEntail.py
KennethEnevoldsen, line 1: does this fix all pair datasets? (I assume it does) Can I ask you to run ~3 of them just to make sure that things work.

KennethEnevoldsen (approved):
Thanks for this @dokato! In the future will you leave the start of the checklist there as well (if you have't run tests/lint just leave it unchecked).

From my reading of the issue, the issue pertains to the structure of the tasks as a "one-line" dataset, which contains all the samples, not the naming of the column.

So, while it is a good fix (for other reasons), it doesn't address the stated issue.

mteb/tasks/PairClassification/zho/CMTEBPairClassification.py
dokato: Well, `ruff` doens't care, and we're not consisnent with it over the datasets. So I admit, it is quite opinionated, but imho lack of identation looks weird and confusing. But if you feel stronlgy about it I can bring it back to what it was...

dokato:
Thanks @KennethEnevoldsen for taking a look. 

> From my reading of the issue, the issue pertains to the structure of the tasks as a "one-line" dataset, which contains all the samples, not the naming of the column.

I think the remark about "one-liner" was only for one specific dataset; others were formatted accordingly , e.g. by creating such lists `_dataset[lang]["test"] = [...]` and evaluation function was adapted. We could change it and modify PairClassificationEvaluator but as long as we remain consistent I don't think it's worth it. I focus here on the second thing mentioned there which is inconsisteny with STS and BiText Mining. And then we can close it!

>  In the future will you leave the start of the checklist there as well (if you have't run tests/lint just leave it unchecked).

Sorry for that... I forgot to WIP it, as I considered `IndicXnliPairClassification` from https://github.com/embeddings-benchmark/mteb/pull/581 as it is a good dataset, adding of which might have been blocked by it. I added it to this PR to make it less trivial.

mteb/tasks/PairClassification/ara/ArEntail.py
dokato: Sure thing, tested:
- PolishPC
- FalseFriendsGermanEnglish
- ArEntail
- and newly added IndicXnliPairClassification

all works fine.

mteb/tasks/PairClassification/zho/CMTEBPairClassification.py
KennethEnevoldsen: Right - @Sakshamrzt, just checking if this affects you?

If it does not I am perfectly fine keeping it as you suggest

KennethEnevoldsen: Let us assume it is fine for now (I think it is)

PR #954, opened 2024-06-18
Backward compatibility fixes for clustering

KennethEnevoldsen:
- [x] ensure that we use the same number of samples for clustering as before

## Checklist


isaac-chung (approved):
Thanks! I had forgotten to add this 😅

Issue #957, opened 2024-06-19
multilingual e5 small CLI failure


isaac-chung:
Running this (added as test case in the branch `e5-model-test`:
```
mteb run -m intfloat/multilingual-e5-small -t BornholmBitextMining --verbosity 3 --output_folder tests/results/test_model --model_revision e4ce9877abf3edfe10b0d82785e83bdcb973e22e
```

gives
```
...
  File "/home/ubuntu/isaac/work/mteb/mteb/evaluation/evaluators/model_encode.py", line 26, in model_encode
    embeddings = model.encode(sentences, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/isaac/work/mteb/mteb/models/e5_models.py", line 134, in encode
    return self.encode_queries(sentences, batch_size=batch_size, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/isaac/work/mteb/mteb/models/e5_models.py", line 138, in encode_queries
    emb = self.mdl.encode(sentences, batch_size=batch_size, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/miniconda3/lib/python3.11/site-packages/sentence_transformers/SentenceTransformer.py", line 451, in encode
    raise ValueError(
ValueError: Prompt name 'BornholmBitextMining' not found in the configured prompts dictionary with keys [].
```

Seems to be related to the changes introduced in https://github.com/embeddings-benchmark/mteb/pull/888

PR #958, opened 2024-06-19
fix: Add E5 model test case in test CLI

isaac-chung:
addresses #957 

## Checklist


tests/test_mteb.py
```diff
@@ -60,15 +60,15 @@ def test_mteb_task(task: Union[str, AbsTask], model_name: str):
...
-        "SummEval",
+        # "TwentyNewsgroupsClustering",
+        # "TwentyNewsgroupsClustering.v2",
+        # "Banking77Classification",
+        # "SciDocsRR",
+        # "SprintDuplicateQuestions",
+        # "NFCorpus",
+        # "MalteseNewsClassification",
+        # "STS12",
+        # "SummEval",
```
KennethEnevoldsen, line 71: Oh, I would keep these

tests/test_mteb.py
```diff
@@ -77,6 +77,9 @@ def test_mteb_with_instructions(task_name: str):
     """
 
     class EncoderWithInstructions(Encoder):
+        def __init__(self, task_name: str):
+            self.prompts = {task_name: "Dummy prompt"}
```
KennethEnevoldsen, line 81: shouldn't be needed

mteb/evaluation/evaluators/model_encode.py
```diff
@@ -14,12 +14,11 @@
...
 ) -> np.ndarray:
-    kwargs["prompt_name"] = task_name
-    if hasattr(model, "prompts") and task_name not in model.prompts:  # type: ignore
+    if hasattr(model, "prompts") and task_name in model.prompts:  # type: ignore
         logger.info(
-            f"Prompt {task_name} not found in model prompts. Removing prompt_name argument."
+            f"Prompt {task_name} found in model prompts. Adding prompt_name argument."
         )
-        kwargs.pop("prompt_name")
+        kwargs["prompt_name"] = task_name
```
KennethEnevoldsen, line 21: It should actually pop it.

KennethEnevoldsen: This handles sentenceTransformers - otherwise we expect that other implementation are able to handle ignoring the prompt_name args.

KennethEnevoldsen (commented):
Thanks for this @isaac-chung I believe the fix is in #953 so I don't believe this is needed. However, it might be worth adding a test for e5.

isaac-chung:
> I believe the fix is in https://github.com/embeddings-benchmark/mteb/pull/953 so I don't believe this is needed.

Ooooooh okay! Should we wait till that PR gets merged then? Or should I add to your PR?

This still fails at the moment. Would love to have a fix either way.
```
mteb run -m intfloat/multilingual-e5-small -t BornholmBitextMining --verbosity 3 --output_folder tests/results/test_model --model_revision e4ce9877abf3edfe10b0d82785e83bdcb973e22e
```

KennethEnevoldsen:
Does it fail on the new PR?

isaac-chung:
It passes here, but fails on master.

KennethEnevoldsen:
But we haven't tested it on #953

isaac-chung:
I can close this PR and add the extra test case in #953 instead, as it's collecting quite a lot of changes 👍

KennethEnevoldsen (approved):

KennethEnevoldsen:
This looks fine to add as it is

PR #970, opened 2024-06-20
Fix GritLM kwarg

Muennighoff:
Not sure this is the cleanest solution but GritLM init does not take arbitrary kwargs and fails if we pass device

### DIFF
--- a/mteb/models/gritlm.py
+++ b/mteb/models/gritlm.py
@@ -14,7 +14,7 @@ def gritlm_loader(**kwargs):
         raise ImportError(
             "GritLM is not installed. Please install it with `pip install gritlm`."
         )
-    kwargs.pop("device") # GritLM does automatic device placement
+    kwargs.pop("device", None) # GritLM does automatic device placement
     return GritLM(**kwargs)
 
 
--- a/mteb/models/gritlm.py
+++ b/mteb/models/gritlm.py
@@ -14,6 +14,7 @@ def gritlm_loader(**kwargs):
         raise ImportError(
             "GritLM is not installed. Please install it with `pip install gritlm`."
         )
+    kwargs.pop("device") # GritLM does automatic device placement
     return GritLM(**kwargs)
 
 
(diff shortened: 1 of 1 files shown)


mteb/models/gritlm.py
```diff
@@ -14,6 +14,7 @@ def gritlm_loader(**kwargs):
         raise ImportError(
             "GritLM is not installed. Please install it with `pip install gritlm`."
         )
+    kwargs.pop("device") # GritLM does automatic device placement
```
isaac-chung, line 17: Could we add a check?
```suggestion
    if "device" in kwargs:
      kwargs.pop("device") # GritLM does automatic device placement
```

isaac-chung (approved):

mteb/models/gritlm.py
Muennighoff: I think kwargs.pop does not fail if the text is not in kwargs but just return None in that case, so fine without checking, no?

isaac-chung: Oh right! Maybe like this?
```suggestion
    kwargs.pop("device", None) # GritLM does automatic device placement
```

I was reading [here](https://docs.python.org/3/library/stdtypes.html#dict.pop) that 
> If default is not given and key is not in the dictionary, a KeyError is raised.

Muennighoff: Oh you're right; let's go with that then!

mteb/models/gritlm.py
```diff
@@ -14,6 +14,7 @@ def gritlm_loader(**kwargs):
         raise ImportError(
             "GritLM is not installed. Please install it with `pip install gritlm`."
         )
+    kwargs.pop("device", None)  # GritLM does automatic device placement
```
KennethEnevoldsen, line 17: What you could do is just remove any kwarg which isn't allowed.

Muennighoff: That would require a list of kwargs that are allowed right? It's a bit complex cuz the kwargs are passed to HF AutoModel which allows lots of kwargs and may even depend on the model that is loaded I think

KennethEnevoldsen: Ahh right then this fix is fine

PR #972, opened 2024-06-21
[WIP] RAR-b second PR & task instruction for IR tasks in general

gowitheflow-1998:
Adding the ability to evaluate RAR-b tasks with task-specific instructions.

The way I am doing it now is by only modifying RAR-b's task-specific classes to take in an extra `include_task_instruction` arg. I have only modified ARCChallenge for now, and it will be good to get some opinions!

I have further included the flexibility to change `instruction` and the format to concatenate instruction by passing it to `format_func` (e.g., putting instructions before or after the queries, concatenated by "\n" or " " can even make some difference to the performance), which has some default format. I think these are very important to accomodate the optimal instruction & instruction format including complex ones that LLM-based embedding models need (e.g., E5Mistral, GritLM).

When we need to customize:

```python
from mteb.tasks.Retrieval import ARCChallenge
import mteb

model = SentenceTransformer("sentence-transformers/all-MiniLM-L6-v2")

def custom_format(instruction, query):
    return f"{instruction}\n{query}" #f"{instruction} {query}" by default

# Initialize the task with custom instruction and format function
arc_challenge_task = ARCChallenge(
    include_task_instruction=True,
    instruction = "Answer the following question carefully", #"Retrieve the answer to this question." by default
    format_func=custom_format
)
evaluation = mteb.MTEB(tasks=[arc_challenge_task])
```

Otherwise, it is still:
```python
tasks = mteb.get_tasks(tasks=["ARCChallenge"])
```

Looking forward to get some opinions before I make changes to the remaining 16 RAR-b tasks! Also, it might be relevant to other IR tasks as well?

## Checklist


KennethEnevoldsen:
@orionw this seems to fall within intruction retrieval. Will you have the time to look at this?

orionw:
Hi @gowitheflow-1998! Thanks for adding RAR-B!

Typically we let the models themselves implement the instruction (see [e5’s codebase](https://github.com/microsoft/unilm/blob/master/e5/utils.py) as an example, but it’s similar in others) because each model will do things differently and `mteb` stores only the datasets and not the models. Otherwise we’d have to store a large dict of instruction mappings per model and then update it for every task, which is unwieldy. 

If I recall the paper correctly, all models were given the same instruction? I suppose then the question of whether to add it to RAR-B datasets depends on whether you want every model to use the same instruction or whether you’re okay with people defining their own. If the latter, I’d leave the instructions out and people will use yours as a default and/or change them when they create their own large mapping of instructions.  If you want it to be these specific instructions, I’d append/prepend them to the queries, perhaps with a separator so that people can break them apart if they want to use them in a specific manner.

@KennethEnevoldsen did recently add the task name to the parameters models receive, so they’ll be able to lookup an instruction based on your dataset name, if you prefer to let models choose their own instructions.

KennethEnevoldsen:
re. task_name, we have a list of default instructions in within mteb/models (which I have considered moving to the task object instead).

orionw:
Perfect @KennethEnevoldsen! If @gowitheflow-1998 thinks it’s a good fit, I would propose we move these instructions there! That way it will establish the default but allow others to change it (assuming that’s the desired functionality, correct me if that’s wrong). 

Re: moving it to the task object, perhaps that’s another PR we can make an issue for.  I can see that being a good fit also. 

However I think the formatting instructions (e.g new lines and other things like “passage:”) should be model defined although I like the proposed method in this PR of being able to override a default functionality.  It might be a good fit in a general model function rather than task specific, for that piece.

gowitheflow-1998:
@orionw @KennethEnevoldsen thanks for the inputs Orion and Kenneth! 

Indeed for each task, I used the same actual `instruction` for all models, but strictly followed the `instruction format` that is optimal to each model to wrap the actual instruction, such as `'Instruct: {}\nQuery: '.format(instruction)` for E5Mistral. For black-box ones like OpenAI's, I simply prepended the instructions like non-LLM models.

Therefore in my own benchmark implementation, I customized the evaluation pipeline for each model class. For examples, for old Sentence Transformers models, we would have to prepend the instruction to the queries before passing them into the model, but for models like GritLM, they can take in the instructions and prepend them on the fly, so we will only need to pass in the instructions along the queries as seperate arguments.

I guess the main question here is that, with the current codebase, is there a way that we can use the same workflow/pipeline with MTEB, for models that do and do not take in instructions on the fly?

And yes although it'll be great to pre-define a task-specific instruction and model-specific format, I think it'll be useful to allow people to customize their own instruction or override the defined instruction format, because for models that aren't tuned with instructions, people might find a different optimal format for tasks!

orionw:
> I guess the main question here is that, with the current codebase, is there a way that we can use the same workflow/pipeline with MTEB, for models that do and do not take in instructions on the fly?

You can add the instructions [here](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/instructions.py).

For the features for each model, the person evaluating would have to define that individually. Alternatively if there are models you want to add to the [models/](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/) folder I think that could work also. It looks like this covers most of what you were saying.

gowitheflow-1998:
thanks @orionw. I have added default instructions to the default instruction lookup dict!

I was trying to confirm that I can reproduce the exact same results with my own implementation for both SentenceTransformer models and user defined models, and found that there're generally two problems at the moment and my proposals accordingly:

1. if using `SentenceTransformer` (which is not `is_dres_compatible` that thus will first be wrapped in DRES and thus using `model_encode`), with the current codebase, the default result will be without instructions, as our defined task prompt isn't by default in `model.prompts` and thus would be removed when calling `model_encode`, so we will have to do something like: 
```python
model.prompts["WinoGrande"] = "Given the following sentence, retrieve an appropriate answer to fill in the missing underscored part. "
tasks = mteb.get_tasks(tasks=["WinoGrande"])
```
With this though, the prompt will be prepended to both `queries` and `documents`, which is not optimal as we both know. A workaround without changing the codebase is that I keep my original implementation last week which pre-processes the queries in the datasets, but turn `include_task_instruction` to `False` by default, so that when people want to evaluate with instructions for this kind of models, they can just:

```python
arc_challenge_task = ARCChallenge(include_task_instruction=True)
evaluation = mteb.MTEB(tasks=[arc_challenge_task])
```

2. For models that people need to define individually such as E5MistralWrapper in the codebase (which are typically made `is_dres_compatible` with `encode_queries` and `encode_corpus` defined already), the current problem is the default setting will be with-instruction, because `RetrievalEvaluator` does:
```python
return self.retriever.search(
    corpus,
    queries,
    self.top_k,
    self.score_function,
    prompt_name=self.task_name,
)
```
which takes in the task_name as prompt_name automatically.

A workaround without adjusting the codebase is when defining models, we can take in an `evaluate_with_instruction` in their init and `encode_queries` differently based on their own init, so whether or not using instructions can be decided on the model loading end, instead of the `RetrievalEvaluator` end. Such as:
```python
class E5Mistral:
    def __init__(self, model_name: str, revision:str, evaluate_with_instruction=False, **kwargs):
        self.evaluate_with_instruction = False
```
and
```python
def encode_queries(self, queries, prompt_name, batch_size:int=32, **kwargs):
    if self.evaluate_with_instruction==True:
        instruction = task_to_instruction(prompt_name)
    else:        
        instruction = ""
    queries = [
        f"Instruct: {instruction}\nQuery: {sentence}"
        for sentence in queries
      ]
```

A short summary (sorry for the long comments!), for SentenceTransformer models and defined models, we might need to customize the pipeline differently to reproduce the exact same results. I can probably open a PR to adjust the codebase later to make it more unified, but if it makes sense at the moment, I'll keep the instructions in the default dict and also finish my original implementation last week for the remaining datasets to accommodate both use cases? @orionw 

cc @KennethEnevoldsen @Muennighoff

orionw:
I think this approach makes a lot of sense, thanks for the detailed comments @gowitheflow-1998!

One clarifying question: is the proposal to define instructions in the task (for `sentencetransformer` models) and in the model classes (for E5-Mistral type models)? If this is the case, I think it may be better to pick only one of the two options to be consistent.  My vote would be for the instructions to be defined at a model level (one option is what you have above in code by manually setting `prompts` or by doing something like the below)

```
class InstructionSentenceTransformer(SentenceTransformer):

    def __init__(self, is_tart, **kwargs):
        super().__init__(**kwargs)
        # load instructions here
        
   def encode_queries(self, queries, batch_size: int, **kwargs):
        queries = [instruction + query for query in queries]
        return self.model.encode(queries, batch_size=batch_size, **kwargs)
```

> I can probably open a PR to adjust the codebase later to make it more unified, but if it makes sense at the moment

Do you mean to extend the `DRES` class at some point to allow for these instructions prefixes? That seems like a good plan for a future PR as then all the models will have the `prompts` object and be on the same page. 


> For models that people need to define individually such as E5MistralWrapper in the codebase (which are typically made is_dres_compatible with encode_queries and encode_corpus defined already), the current problem is the default setting will be with-instruction

I think it's fine for the default to be with instructions, if you're okay with it. It seems easier to set the default instructions and let users opt out rather than asking them to copy all the instructions in. If we want it to be non-instruction the user can define a different set of `prompts` with the values as empty strings.

KennethEnevoldsen:
So my 2 cents here:

We generally want to give the model all the information it may potentially need:


This we want the signature for encode_corpus to also take in prompt_name, however, it might be a good default to have the sentence transformers ignore this for the corpus. A solution is to define a wrapper (essentially `DRESModel`)

It might be ideal to actually have an update to SentenceTransformers on this to allow for encode_queries and encode_corpus commands.

gowitheflow-1998:
thanks @orionw@KennethEnevoldsen! sorry for the late reply.

Indeed I think having a consistent pipeline will be better, and defining at the model level seems like a good solution for all RAR-b use cases now! I will also add some demos from my own repo to show the way to define such models to reproduce RAR-b results with MTEB as well!

and yes an update to both DRES and SentenceTransformer will be good as future PRs at some point I agree, as taking in and customizing instructions become more and more common.

I guess it should be fine to merge my changes to `instruction.py` in this case then? Let me know if you guys think it'll be okay if I modify the remaining task-specific classes as well? This can be an alternative way - people can also pre-process the queries first, and also customize at the model level to not do anything about the instructions on-the-fly.

orionw:
> I guess it should be fine to merge my changes to instructions.py in this case then?

Definitely lets commit the instructions!

> Let me know if you guys think it'll be okay if I modify the remaining task-specific classes as well? This can be an alternative way - people can also pre-process the queries first

I may be confused, why would they need to preprocess the queries?  Is it for the instructions?  If so I think that's fine for now in the model class -- unless you want to change the `DRES` and `SentenceTransformer` in this PR, which would also be great!

gowitheflow-1998:
@orionw yes the plan was people can either add instructions to the queries before encoding (what I defined in `ARCChallengeRetrieval.py`) or do them on-the-fly with custom models with `instructions.py` (what we've been talking about).

In this case I'll remove the dataset class changes and only keep new instructions in `instructions.py` then!

gowitheflow-1998:
@orionw have removed dataset class-based one and updated the instruction lookup to the latest format & added points for previous and this PRs. Not much to add on my end. Feel free to merge it if okay!

orionw (approved):

orionw:
LGTM, thanks @gowitheflow-1998!!

PR #981, opened 2024-06-24
Fix GritLM Retrieval instructions

Muennighoff:
Issue mentioned in https://github.com/embeddings-benchmark/mteb/pull/976

### DIFF
--- a/mteb/cli.py
+++ b/mteb/cli.py
@@ -114,8 +114,6 @@ def run(args: argparse.Namespace) -> None:
         tasks=args.tasks,
     )
 
-    print("Tasks:", tasks)
-
     eval = mteb.MTEB(tasks=tasks)
 
     eval.run(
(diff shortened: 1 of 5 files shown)


KennethEnevoldsen:
Looks good @Muennighoff - I have added a fix for the e5 models as well

Muennighoff:
I think for e5 it was already fine cuz it didn't use the instruction anyways, but also fine to change it like you did 👍

```
        if encode_type == "query":
            sentences = [
                f"Instruction: {instruction}\nQuery: {sentence}"
                for sentence in sentences
            ]
```

KennethEnevoldsen:
sorry I was unclear - changed it for the e5 instruct models only

Muennighoff:
Oh sorry I was unclear - i also meant that for e5 instruct I think it was already fine cuz of the code above, but anyways it's cleaner to explicitly not have an instruction like you changed so all good as is I think! 😁

KennethEnevoldsen:
Oh double misunderstanding there. Well no worries clarity is probably a good thing in this case as you say

PR #985, opened 2024-06-25
Update Retrieval Statistics Checker

orionw:
Changes the way retrieval statistics are calculated to be more helpful for MMTEB planning (calculating qrels, docs, etc.). It also returns a value, so it can be used in other scripts. 

I used this to create [this spreadsheet](https://docs.google.com/spreadsheets/d/12Yldu9whVysP0jhnw1S5lmpTi2hOolU2KO5mWgSYk6I/edit?usp=sharing) for planning.

Example output now:
```
{"test": {"average_document_length": 19706.823653325308, "average_query_length": 1031.0680889324833, "num_documents": 9969, "num_queries": 12234, "average_relevant_docs_per_query": 1.1705084191597188}, "task_name": "GerDaLIRSmall"}
```

## Checklist


orionw:
Also @KennethEnevoldsen @isaac-chung I was planning to downsample all datasets with > 100k docs, which is roughly 40 tasks. See the [spreadsheet](https://docs.google.com/spreadsheets/d/12Yldu9whVysP0jhnw1S5lmpTi2hOolU2KO5mWgSYk6I/edit?usp=sharing) for the specific tasks.

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -376,44 +376,78 @@ def _add_main_score(self, scores: ScoresDict) -> None:
     def calculate_metadata_metrics(self) -> None:
         self.load_data()
 
+        all_details = {}
         for split in self.metadata_dict["eval_splits"]:
+            all_details[split] = {}
             if self.is_multilingual:
                 for lang in self.relevant_docs.keys():
-                    process_language(
+                    print(split, lang)
```
isaac-chung, line 384: Was this intentional or for debugging purposes?

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -376,44 +376,78 @@ def _add_main_score(self, scores: ScoresDict) -> None:
...
+                    split_details = process_language(
                         self.relevant_docs[lang][split],
                         self.queries[lang][split],
                         self.corpus[lang][split],
                         lang,
                     )
+                    all_details[split][lang] = split_details
             else:
-                process_language(
+                print(split)
```
isaac-chung, line 393: Was this intentional or for debugging purposes?

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -376,44 +376,78 @@ def _add_main_score(self, scores: ScoresDict) -> None:
...
+    doc_lens = []
+    for query in queries.values():
+        queries_lens.append(len(query))
+
+    for doc in corpus.values():
+        if isinstance(doc, dict):
+            if "title" in doc:
+                doc_lens.append(len(doc["title"]) + len(doc["text"]))
+            else:
+                doc_lens.append(len(doc["text"]))
```
isaac-chung, line 447: Could be simplified:
```suggestion
            doc_lens.append(len(doc.get("title", "")) + len(doc["text"]))
```

isaac-chung (approved):
Looking great! A few small comments. We could also consider a small test for these parts of the class but not a blocker for merging.

mteb/abstasks/AbsTaskRetrieval.py
orionw: There were the other print statements, so I wasn't sure -- should I keep them in (and make them nicer) or take them out?

mteb/abstasks/AbsTaskRetrieval.py
orionw: Same as above

isaac-chung: For this, as the number of splits are usually 1 or 2, perhaps this can be removed.

mteb/abstasks/AbsTaskRetrieval.py
isaac-chung: Here we often see multilingual datasets with a huge number of languages. Showing process would be nice to have. Actually instead of print, maybe a progress bar could be helpful?
```
for lang in tqdm(self.relevant_docs.keys()):
    ...
```

orionw: I like the tqdm bars, but realized when I switched it to that locally that the results print statements are then unattached to what split they came from.  So I will leave the print statements and add the TQDM bars

orionw:
> We could also consider a small test for these parts of the class but not a blocker for merging.

We do have [one test](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_retrieval_abstask.py#L9), although it just checks that it runs.  I think that's probably fine for now since I don't think anything depends on the exact keys in the dictionaries, just that it runs.

isaac-chung:
Makes sense, thanks! Merging now

KennethEnevoldsen:
Looks much better thanks for this PR. Spreadsheets look more than fine, with the exception of MIRACLRetrieval which is 0?

We should probably also update the scores in the datasets since we have just calculated them (this will also make future PRs use the same scores).

Besides that I think it is good to merge.

orionw:
> exception of MIRACLRetrieval which is 0?

I think this was pre-merge, so it was only partially implemented. I still need to re-run the ones you two fixed, thanks for that!

> We should probably also update the scores in the datasets since we have just calculated them (this will also make future PRs use the same scores).

Agree -- I can go through and make a PR for that also, updating all the scores.

KennethEnevoldsen:
Perfect - then I think this is perfectly legit as it stands.

PR #986, opened 2024-06-25
Add BGE large en v1.5

isaac-chung:
Addresses part of https://github.com/embeddings-benchmark/arena/issues/6

## Checklist

### Adding a model checklist

### DIFF
--- a/mteb/models/bge_models.py
+++ b/mteb/models/bge_models.py
@@ -56,3 +56,12 @@ bge_base_en_v1_5 = ModelMeta(
     revision="a5beb1e3e68b9ab74eb54cfd186867f64f240e1a",
     release_date="2023-09-11",  # initial commit of hf model.
 )
+
+bge_large_en_v1_5 = ModelMeta(
+    loader=partial(BGEWrapper, model_name="BAAI/bge-large-en-v1.5"),  # type: ignore
+    name="BAAI/bge-large-en-v1.5",
+    languages=["eng_Latn"],
+    open_source=True,
+    revision="d4aa6901d3a41ba39fb536a557fa166f842b0e09",
+    release_date="2023-09-12",  # initial commit of hf model.
+)
--- a/mteb/models/bge_models.py
+++ b/mteb/models/bge_models.py
@@ -36,7 +36,9 @@ class BGEWrapper:
             "Represent this sentence for searching relevant passages: " + sentence
             for sentence in queries
         ]
-        emb = self.mdl.encode(sentences, batch_size=batch_size, normalize_embeddings=True, **kwargs)
+        emb = self.mdl.encode(
+            sentences, batch_size=batch_size, normalize_embeddings=True, **kwargs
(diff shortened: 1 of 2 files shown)


mteb/models/jina_models.py
```diff
@@ -0,0 +1,9 @@
+from mteb.model_meta import ModelMeta
+
+paraphrase_multilingual_MiniLM_L12_v2 = ModelMeta(
+    name="jinaai/jina-embeddings-v2-base-en",
+    languages=["eng_Latn"],
+    open_source=True,
+    revision="f84569a8c478c7bb2ea3bc187034859870940deb",
+    release_date="2023-10-30",  # release date of paper
+)
```
Muennighoff, line 9: Oh since Jina works out of the box with ST do we even need to add it? I guess it can already be loaded?

Muennighoff (commented):
Amazing speed 🚀

mteb/models/jina_models.py
isaac-chung: Good call! Model loading worked without being added to `mteb/models/__init__.py`. Will remove

mteb/models/bge_models.py
```diff
@@ -56,3 +56,12 @@ def encode_corpus(
...
 )
+
+bge_large_en_v1_5 = ModelMeta(
+    loader=partial(BGEWrapper, model_name="BAAI/bge-large-en-v1.5"),  # type: ignore
+    name="BAAI/bge-large-en-v1.5",
+    languages=["eng_Latn"],
+    open_source=True,
+    revision="d4aa6901d3a41ba39fb536a557fa166f842b0e09",
+    release_date="2023-09-12",  # initial commit of hf model.
+)
```
Muennighoff, line 67: Don't we need `normalize_embeddings=True` when calling `self.mdl.encode`? 🤔
See
(1) https://huggingface.co/BAAI/bge-large-en-v1.5#using-sentence-transformers
(2) https://github.com/UKPLab/sentence-transformers/blob/2dee8c22f618df3f34f55d406bf97aa88383728a/sentence_transformers/SentenceTransformer.py#L430

Maybe worth checking on a quick retrieval dataset like SciFact if we get similar scores to what the model gets on the mteb leaderboard

KennethEnevoldsen (approved):

mteb/models/bge_models.py
isaac-chung: Good catch! I've run `mteb run -t SciFact -m BAAI/bge-large-en-v1.5 --model_revision d4aa6901d3a41ba39fb536a557fa166f842b0e09`. 

`bge-large-en-v1.5` gives 74.63 from mteb (this PR) and 74.61 from the leaderboard.

isaac-chung: Rerunning with `normalize_embeddings=True` yields 74.63 as well.

Muennighoff (approved):

PR #988, opened 2024-06-25
Update numbers in retrieval statistics

orionw:
Update the retrieval statistics to match the new metadata processing.

## Checklist


orionw:
@KennethEnevoldsen @isaac-chung I replaced one of the keys with the full information. 

How do we want to break this up to fit in the existing schema?

KennethEnevoldsen (approved):
I think this looks very good. We can def. keep this as is.

mteb/tasks/Retrieval/code/CodeSearchNetRetrieval.py
```diff
@@ -38,7 +38,53 @@ class CodeSearchNetRetrieval(MultilingualTask, AbsTaskRetrieval):
         n_samples={
             _EVAL_SPLIT: 1000,
         },
-        avg_character_length={"test": 1196.4609},
+        avg_character_length={
```
isaac-chung, line 41: Just need to fix the pydantic validation errors.

mteb/tasks/Retrieval/code/CodeSearchNetRetrieval.py
```diff
@@ -38,7 +38,53 @@ class CodeSearchNetRetrieval(MultilingualTask, AbsTaskRetrieval):
...
+                },
+                "php": {
+                    "average_document_length": 712.129,
+                    "average_query_length": 162.119,
+                    "num_documents": 1000,
+                    "num_queries": 1000,
+                    "average_relevant_docs_per_query": 1.0,
+                },
+            },
+            "task_name": "CodeSearchNetRetrieval",
```
isaac-chung, line 86: Could remove `task_name` as it's duplicated in the metadata.

mteb/tasks/Retrieval/code/CodeSearchNetRetrieval.py
orionw: Looks like you fixed the pydantic, thank you! I'll update the docstring now

mteb/tasks/Retrieval/code/CodeSearchNetRetrieval.py
orionw: Agree, no need to have that also

orionw:
I think I got every retrieval dataset, except for MSMarcoV2.  My machine unfortunately died every time I tried to calculate it, apparently it requires a lot of RAM.

isaac-chung:
@orionw the pydantic error seems to be related to the presence of `"task_name"` in some dicts, e.g. `mteb/tasks/Retrieval/dan/TV2Nordretrieval.py`. The validation checks for a dict of {str:float} (for other tasks) or {str:{str: dict}} (for this change). The extra key will make it upset.

orionw:
Ah thank you! I tried to delete all of them but clearly missed some. I'll update it

orionw:
Thanks @isaac-chung!! Good luck, it takes quite a while! 

TBH if neither of us can load it, perhaps we need a more efficient dataset loader or simply to remove MSMarcoV2 from our list.  My machine had quite a lot of RAM so it's pretty inaccessible -- looking at the [specs on IR datasets](https://ir-datasets.com/msmarco-passage-v2.html) it has 138 million passages!!

isaac-chung:
Wow!
It seems to be running still. I'll give it till the morning (GMT+3) and report back.

isaac-chung:
It went OOM killed already. Let's merge this and consider our options with MSMARCOv2.

PR #994, opened 2024-06-27
Update DanFever to use avoid using overwrite=True and ignore identical ids

KennethEnevoldsen:
This pull request updates the DanFever module to not ignore identical ids. It also adds a utility helper  when specifying the wrong task (suggesting the closest alternative) and adds the `--overwrite` flag to the command-line interface. Additionally I changed some of the logging to more correctly reflect the severity.

## Checklist


mteb/tasks/Retrieval/dan/DanFeverRetrieval.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def load_data(self, **kwargs):
+        """Load dataset from HuggingFace hub"""
+        if self.data_loaded:
+            return
+        self.dataset = datasets.load_dataset(**self.metadata.dataset)  # type: ignore
+        self.dataset_transform()
+        self.data_loaded = True
+
+    def dataset_transform(self) -> None:
```
imenelydiaker, line 177: This function does the same preprocessing fo tboth tasks right? If it's the case, can we factor it into one function outside of the classes and call it in each `data_transform` ?

KennethEnevoldsen: Oh no that is a mistake 2 sec will fix that

mteb/tasks/Retrieval/dan/DanFeverRetrieval.py
```diff
@@ -0,0 +1,221 @@
...
+                    1 if claim_is_supported else 0
+                )  # negative for refutes claims - is that what we want?
+
+                if claim not in text2id:
+                    text2id[claim] = str(len(text2id))
+                if evidence not in text2id:
+                    text2id[evidence] = len(text2id)
+
+                claim_id = "Q" + str(text2id[claim])
+                evidence_id = "C" + str(text2id[evidence])
```
imenelydiaker: In the old version there was no Q and C I think an that's why we put `ignore_identical_ids=True`
```suggestion
                claim_id = str(text2id[claim])
                evidence_id = str(text2id[evidence])
```

mteb/overview.py
```diff
@@ -226,4 +227,15 @@ def get_task(
...
+        close_matches = difflib.get_close_matches(task_name, TASKS_REGISTRY.keys())
+        if close_matches:
+            suggestion = (
+                f"KeyError: '{task_name}' not found. Did you mean: {close_matches[0]}?"
+            )
+        else:
+            suggestion = (
+                f"KeyError: '{task_name}' not found and no similar keys were found."
+            )
+        raise KeyError(suggestion)
```
imenelydiaker, line 240: If I get this right, older results will be overwritten by the new task? Is there any reason for keeping the old task?

mteb/tasks/Retrieval/dan/DanFeverRetrieval.py
```diff
@@ -0,0 +1,221 @@
...
+                self.queries[split][claim_id] = claim
+                self.corpus[split][evidence_id] = {"title": "", "text": evidence}
+
+                if claim_id not in self.relevant_docs[split]:
+                    self.relevant_docs[split][claim_id] = {}
+
+                self.relevant_docs[split][claim_id][evidence_id] = sim
+
+
+class DanFever(AbsTaskRetrieval):
```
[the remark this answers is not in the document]
imenelydiaker: Okay, looks good :)

KennethEnevoldsen:
Perfect - I also just found out that superseeded_by is a wrong spelling so fixed that as well.

PR #995, opened 2024-06-27
Added utility function for downloading and loading models results

KennethEnevoldsen:
Model results can now be loaded using 

`mteb.load_results()`

This will download the results repo (on github) -- or update it if needed -- and then load then in as a dict (see docstring).

The hope is that this is the start to developing better result fetching + result downloading from MTEB. I could also imagine we could add the option to load specific model result from HF hub, push results to models card (creating a PR) etc.

## Checklist


orionw (approved):
I like this, thanks @KennethEnevoldsen. Then we can test that new changes can still load and also run some Github Actions using it.

PR #996, opened 2024-06-27
docs: Added result tables

KennethEnevoldsen:
Added results tables for MMTEB. Some reuslts are still missing.


KennethEnevoldsen:
Will merge this in as I need the result table construction for another thing (it is only scripts so no notable changes to the package)

vaibhavad (approved):

PR #1001, opened 2024-06-28
install only ruff for lint workflow

Samoed:
Install only ruff for lint workflow instead of all main dependencies.


KennethEnevoldsen:
This will speed up linting slightly. However, it does lead to multiple places of specifying installation. make install simplifies the installation process - as it is set up now any changes to linting commands happen in the make file. E.g. we could replace ruff in make lint-check without having to worry about changing the CI at all.

I am not sure this change is worth it.

Issue #837, opened 2024-05-28
Paper segment: Task selection


KennethEnevoldsen:
The goal of this segment is to create meaningful benchmark subsets with a minimal set of tasks.

I believe the steps are as follows:

1) construct an experimental subset. If people agree I can construct one from the Scandinavian Embedding Benchmark. I imagine that this is both small enough but also we assume there is cross-lingual transfer between languages (so some expected redundancy). 
2) experiment with methods for subsampling (please provide suggestions):
  - [Predicting Performance for Natural Language Processing Tasks](https://arxiv.org/abs/2005.00870)
  - [big-bench approach](https://arxiv.org/abs/2206.04615)
3) Deciding on the best approach

If people agree with 1) we can start running the models on `tasks = mteb.get_tasks(langauges = ["dan", "nno", "nob", "swe"])`. Alternatively, we can also do it on MTEB (english) where we also expect some redundancy (e.g. between arxiv P2P and S2S). The cross-lingual transfer might be more relevant given the multilingual focus.

KennethEnevoldsen:
So @vaibhavad had a look at the first paper and did a bit of a write-up (mix of their and my own ideas). Let me know what you think. 

> Following a similar approach as [xia et al. (2020)](https://arxiv.org/abs/2005.00870), we seek to estimate the performance of an unseen task $U_t_i$ performance on another set of tasks $S_t_j, t_j \in T \ {t_i}$. We will, in our case, leave the target task arbitrary, as it could be any downstream task of the model. Note that keeping the task arbitrary can be equated to estimating a generalization factor of the model (Chollet, ~2020). 
> ...

> Under this formulation, we then consider task selection as a feature reduction/redundancy problem, a well-formulated problem with machine learning (cite missing), where we seek to remove features that either share the same information or aren’t predictive of the target variable. We treat the performance of existing tasks as features. Note the approach used by Xia et al. (2020) follows a ‘forward feature selection’ approach.

This approach leaves a whole slew of known approaches e.g. feature specific approaches e.g. filtering based on Mutual information, feature correlation, or variance within a feature. We don't even need the model in this case for anything other than the argument.

However, we might want to make the model anyway as it might help us in the aggregation step -- predictive performance on task U_i is really what we are interested in when evaluating -- here I don't believe that following Xia et al. is very promising. Their approach seems to avoid the notion that if task t1 is predictive of the model performance on task t2, then I would also imagine it more likely to be predictive of t3.

[...]

Issue #368, opened 2024-04-14
Add a Benchmark for African languages


KennethEnevoldsen:
#### Afro-Asiatic Languages
- [x] Arabic - ara (widely spoken in North Africa)
- [x] Amharic - amh (Ethiopia)
- [x] Somali - som (Somalia)
- [x] Tigrinya - tir (Eritrea and Ethiopia)
- [x] Hausa - hau (Niger-Congo but extensively uses Afro-Asiatic vocabulary, Nigeria and surrounding countries)

#### Niger-Congo Languages
- [x] Swahili - swa (East Africa)
- [x] Yoruba - yor (Nigeria and neighboring countries)
- [x] Igbo - ibo (Nigeria)
- [ ] Akan - aka (Ghana)
- [ ] Zulu - zul (South Africa)
- [x] Xhosa - xho (South Africa)
- [x] Lingala - lin (Congo)

#### Nilo-Saharan Languages
- [ ] Nuer - nus (South Sudan)
- [ ] Dinka - din (South Sudan)
- [ ] Kanuri - kau (Nigeria, Niger)

#### Khoisan Languages (Known for their click consonants)
- [ ] Khoekhoe - naq (Namibia, South Africa)

#### Creole Languages
- [x] Pidgin (Nigerian Pidgin) - pcm (Widely used in informal contexts across Nigeria)

KennethEnevoldsen:
I will close this as not planned but contributors are more than welcome to step in

Issue #367, opened 2024-04-14
Add a Benchmark for Asian Languages


KennethEnevoldsen:
#### East Asian Languages
- [x] Chinese (Mandarin) - cmn
- [ ] Cantonese - yue
- [x] Japanese - jpn
- [x] Korean - kor
- [ ] Mongolian - mon

#### South Asian Languages
**Indic Languages:**

- [x] Hindi - hin
- [ ] Bengali - ben
- [ ] Punjabi - pan
- [ ] Marathi - mar
- [ ] Gujarati - guj
- [ ] Urdu - urd
- [ ] Nepali - nep
- [ ] Sinhala - sin
- [x] Tamil - tam
- [ ] Telugu - tel
- [ ] Kannada - kan
- [ ] Malayalam - mal

- **Dravidian Languages:**
  - Included above (Tamil, Telugu, Kannada, Malayalam)

#### Southeast Asian Languages
- **Austronesian Languages:**
  - [ ] Indonesian - ind
  - [ ] Filipino - fil
  - [ ] Malay - msa
  - [ ] Javanese - jav
- **Tai-Kadai Languages:**
  - [x] Thai - tha
  - [ ] Lao - lao
- **Austroasiatic Languages:**
  - [ ] Vietnamese - vie (see #364) 
  - [ ] Khmer - khm
- [ ] Burmese - mya

#### Central Asian Languages
- **Turkic Languages:**
  - [ ] Kazakh - kaz
  - [ ] Uzbek - uzb
  - [ ] Turkmen - tkm
  - [ ] Kyrgyz - kir
  - [ ] Uighur - uig

#### West Asian (Middle Eastern) Languages
- **Semitic Languages:**
  - [x] Arabic - ara
  - [ ] Hebrew - heb
- **Iranian Languages:**
  - [ ] Persian - fas
  - [ ] Kurdish - kur
  - [ ] Pashto - pus
  - [ ] Dari - prs

rasdani:
I will take a stab at a Bengali benchmark together with a colleague of mine 👍

KennethEnevoldsen:
Wonderful @rasdani feel free to create an issue on this as well so that other can see that you are working on it.

gentaiscool:
I created PRs for Indonesian languages (at least 10+ additions from 2 corpus) and African language. Once, they are approved, I can add the language to the list.

KennethEnevoldsen:
I will close this as not planned, but contributors are more than welcome step in

Issue #365, opened 2024-04-14
Add Lingua Franca Benchmark


KennethEnevoldsen:
This issue is a port of a series of issues to get an overview of the missing languages in the benchmark
This issue proposes the addition of benchmarks for lingua francas—languages widely used beyond the native speakers for communication across diverse linguistic communities. The goal is to evaluate and improve the performance of our tools and services in these pivotal languages, enhancing accessibility and functionality for a broader user base.

- **Scope:** Define the range of languages to be included and the criteria for their selection as lingua francas.

### Criteria for Selection:
- **Geographical Spread:** Languages that are used by large populations across multiple countries.
- **Economic Importance:** Languages predominant in global business and international trade.
- **Cultural and Historical Influence:** Languages that have served as mediums for cultural exchange and historical documentation.

### Proposed List of Lingua Francas:

- [x] 1. **English**
- [x] 2. **French**
- [x] 3. **Spanish**
- [x] 4. **Arabic**
- [x] 5. **Mandarin Chinese**
- [x] 6. **Russian**
- [x] 7. **Portuguese**
- [x] 8. **Swahili** (important in Eastern Africa)
- [x] 9. **Hindi/Urdu** (widely understood in South Asia) 
- [x] 10. **Urdu**

(Urdu is not covered but I am unsure if that is needed, if you know about this feel free to add it)

PR #1003, opened 2024-06-28
Added benchmark construction

KennethEnevoldsen:
This PR adds an outline for benchmark construction.

This is mostly to get an idea for task selection / benchmark construction. (see more at #837) and whether people agree with this approach.

Naturally, once a workflow is set up for creating a benchmark it is quite easy to add new ones so it might also be argued that we should just add a benchmark for all languages we believe is well enough covered. Let me know what you guys think.


isaac-chung:
I think this addresses #368 , #367 , #365 , and #361 as well. Can mention those in the description if that's agreed.

KennethEnevoldsen:
> I think this addresses https://github.com/embeddings-benchmark/mteb/issues/368 , https://github.com/embeddings-benchmark/mteb/issues/367 , https://github.com/embeddings-benchmark/mteb/issues/365 , and https://github.com/embeddings-benchmark/mteb/issues/361 as well. Can mention those in the description if that's agreed.


That is a great idea. We haven't added the lingua franca benchmark but should be easy enough.

mteb/tasks/Classification/zho/CMTEBClassification.py
```diff
@@ -220,21 +220,7 @@ class JDReview(AbsTaskClassification):
...
-}
-""",
+        bibtex_citation="",  # missing
```
Muennighoff, line 223: Maybe it should cite the paper that created CMTEB (https://arxiv.org/abs/2309.07597) 🤔

Muennighoff (commented):
Nice should this file eventually be merged into https://github.com/embeddings-benchmark/mteb/blob/main/mteb/benchmarks.py ?

KennethEnevoldsen:
> Nice should this file eventually be merged into https://github.com/embeddings-benchmark/mteb/blob/main/mteb/benchmarks.py ?

Once we have downsampled the list of tasks - then yes

scripts/mmteb/task_selection/benchmark_construction.py
```diff
@@ -0,0 +1,256 @@
...
+task_lists = {
+    "MTEB(eng, lite)": {
+        "tasks_to_search": MTEB_MAIN_EN,  # tasks to select from
+        "minimally_include": [],  # tasks that must be included in the final benchmark (considered high-quality, representative etc.)
+        "task_types_to_include": [  # task types that should be represented in the final benchmark
+            "Classification",
+            "Clustering",
+            "PairClassification",
+            "Reranking",
+            "Retrieval",
```
isaac-chung, line 60: I might have missed the discussion on this. Was the exclusion of STS and other tasks intentional?

KennethEnevoldsen: No def. Atm they are only not required to be in (however that is also a mistake, the goal is to keep one of each task type).

KennethEnevoldsen:
Will merge this in the for the fixes to datasets - and then create an updated PR for the constructed benchmarks.

KennethEnevoldsen:
Will merge this in for now it only makes minor changes to the repo - future PRs will added more on this

PR #1014, opened 2024-07-01
Added Missing Models

KennethEnevoldsen:
This adds a script for checking for missing models, a script for running them and removed older scripts for running models.

It also add a list of models which currently aren't run.


scripts/mmteb/running_model/check_results.py
```diff
@@ -0,0 +1,220 @@
+"""This script checks that all result have been runs and writes a file: missing_tasks.txt with the missing tasks for each model"""
```
Muennighoff, line 1:
```suggestion
"""This script checks that all result have been run and writes a file: missing_tasks.txt with the missing tasks for each model"""
```

Muennighoff (approved):
Really nice! I will look into those missing_tasks 👍

Muennighoff:
Can you update the `scripts/mmteb/running_model/missing_tasks.txt` here based on the latest pushed results?

KennethEnevoldsen:
Updated! Already notably better - running check_results.py should update it as well.

Muennighoff:
Should be mostly there now? Sorry can you run it once more & update the file? 😅

KennethEnevoldsen:
Yep. Will just merge this in an create a new PR

KennethEnevoldsen:
See #1038

PR #1016, opened 2024-07-01
docs: Add Mariya's points

isaac-chung:
Combining all into one PR, correct the file names, and list contributions. Corrected the jsonl key too (-> `Paper writing`).
- 1011: intro
- 1012: Appendix B

CC @mariyahendriksen for future reference.


isaac-chung:
Wanted to check if the points look good? @KennethEnevoldsen

KennethEnevoldsen:
I think the points are reasonable

PR #1017, opened 2024-07-01
Simplify

Muennighoff:


mteb/abstasks/MultiSubsetLoader.py
```diff
@@ -9,8 +9,7 @@ def load_data(self, **kwargs):
         if self.data_loaded:
             return
 
-        fast_loading = self.fast_loading if hasattr(self, "fast_loading") else False
-        if fast_loading:
+        if hasattr(self, "fast_loading"):
```
KennethEnevoldsen, line 12: if could be false as well I assume?

KennethEnevoldsen (changes requested):
looks good a pull fr

mteb/abstasks/MultiSubsetLoader.py
Muennighoff, line 12:
```suggestion
        if hasattr(self, "fast_loading") and self.fast_loading:
```

Muennighoff: Sorry fixed it!

PR #1026, opened 2024-07-02
fix: Fix broken clustering tasks

KennethEnevoldsen:
Fix clustering tasks using Imenes fix

## Checklist


imenelydiaker (approved):
Look good!

Is it a good Idea to make the subsampling function more robust by choosing the number of samples to min(size of dataset, N_SAMPLES) ?

It can be done later, this fix is already great! Thanks Kenneth!

KennethEnevoldsen:
I will add that in as well!

imenelydiaker:
> I will add that in as well! - I suppose we don't really need to subsample at all if that is the case

Yeah true, it's just that the subsampling function can be applied to all available subsets and splits. Some times for a split you can have multiple subsets which some of do not need subsampling (that's what happenned with MLSUM). It's either adding a condition on the length of the dataset before calling the subsampling function or making subsampling_fn handle it by default. But no rush on this, I'm not sure we have the same problem elsewhere.

imenelydiaker (approved):
LGTM! Thanks 🚀

KennethEnevoldsen:
@imenelydiaker added the downsampling fix. I also found another error due to python 3.8 and a an odd numpy data type.

imenelydiaker:
Perfect, you can merge this PR 😊

PR #1027, opened 2024-07-02
Update figure

x-tabdeveloping:
Updated main figure in the paper with better colors, more up-to-date numbers and reasoning as retrieval.


KennethEnevoldsen (approved):
!

PR #1030, opened 2024-07-02
fix: Added encode_kwargs as the input for encode arguments and added batch_size to CLI

### LINKED ISSUE #1025: Allow passing batch_size via CLI
I think it'd make sense to allow passing the `batch_size`kwarg when using the cli argument `mteb run ...`? Some missing results are due to OOMs (https://github.com/embeddings-benchmark/mteb/pull/1014)

KennethEnevoldsen:
This PR adds encode_kwargs as an argument to MTEB.run(..., encode_kwargs) which allows passing in arguments to model.encode calls. This has the benefit of removing/distinguishing between multiple model arguments present in the code.

Where default were already in place I have replaced kept them, future PRs might standardize or remove these.

I also add a test to ensure that the encode_kwargs in all cases are passed to the model.encode.

I additionally also add the batch_size argument to the CLI.

Fixes #1025 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ evaluation = mteb.MTEB(tasks=MTEB_MAIN_EN, task_langs=["en"])
 
 ### Passing in `encode` arguments
 
-To pass in arguments to the model's `encode` function you can use the encode keyword arguments (`encode_kwargs`):
+To pass in arguments to the model's `encode` function, you can use the encode keyword arguments (`encode_kwargs`):
 
 ```python
 evaluation.run(model, encode_kwargs={"batch_size": 32}


mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -27,63 +26,6 @@ def dot_distance(a: np.ndarray, b: np.ndarray) -> float:
     return -np.dot(a, b)
 
 
-class kNNMultiLabelClassificationEvaluator(Evaluator):
```
Muennighoff, line 30: Why do we not need this one anymore?

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -312,11 +254,16 @@ def __init__(
...
+        encode_kwargs: dict[str, Any] = {},
         limit: int | None = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
+        self.encode_kwargs = encode_kwargs
+
+        if "batch_size" not in self.encode_kwargs:
+            self.encode_kwargs["batch_size"] = 32
+
```
Muennighoff, line 266: Why not default it to  `encode_kwargs: dict[str, Any] = {"batch_size": 32},` like in the other classes above?

Muennighoff (commented):
I'm not sure this is worth it given it breaks backward compat but if you all think it is, then fine with me tho should probably be explained in the docs or sth 🤔 cc @orionw in case of thoughts

orionw:
I may be missing something, how does this break backwards compat? 

I like the idea overall though, it's often confusing to me which one to pass the args too.

Muennighoff:
> I may be missing something, how does this break backwards compat?

Oh maybe I am wrong but I think this code would now fail: https://github.com/ContextualAI/gritlm/blob/fdeb0f78dbb11268233d9777f169df790de3d2b7/evaluation/eval_mteb.py#L1206 (i.e. passing batch_size via regular kwargs)

isaac-chung:
> though, it's often confusing to me which one to pass the args too. Being able to pass all the args to run makes sense.

I feel that way as well. And as this breaks backwards compatibility, I'd suggest keeping the ability to pass `batch_size` into kwargs as is. 

Re-reading https://github.com/embeddings-benchmark/mteb/issues/1025 again, maybe we can move forward with the following:
- having that (pass in `batch_size` via kwargs) in the docs / one of the README examples so people know
- consistently implement `batch_size` for all tasks types

What do you think?

mteb/evaluation/evaluators/ClassificationEvaluator.py
KennethEnevoldsen: The evalautor was never used for multilabel CLF (@x-tabdeveloping). I think this was just an accidental leftover from the PR.

mteb/evaluation/evaluators/ClassificationEvaluator.py
KennethEnevoldsen: Since most places use a default of {} it would just be overwritten. This way we ensure the same behaviour. (I will fix the changed where the encode_kwargs has a default other than {}.

KennethEnevoldsen:
We should def. do something about backward compat. 

I have now implemented 

```
eval.run(batch_size=N)
```

Which does the following:

```
        if "batch_size" in kwargs:
            logger.warning(
                "The `batch_size` argument is deprecated and will be removed in the next release. "
                + "Please use `encode_kwargs = {'batch_size': ...}` to set the batch size instead."
            )
            encode_kwargs["batch_size"] = kwargs["batch_size"]
```

This is not a perfect backward compat. but at least we maintain backward compat on the interface layers. WDYT?

KennethEnevoldsen:
> having that (pass in batch_size via kwargs) in the docs / one of the README examples so people know
consistently implement batch_size for all tasks types
What do you think?

I have added an example to the readme using encode_kwargs. (note that batch_size still works)

isaac-chung:
The current way of handling `batch_size in kwargs looks good to me. To test that, I've modified the test case to the following:
```diff
def test_encode_kwargs_passed_to_all_encodes(task_name: str | AbsTask):
    """Test that all tasks correctly pass down the encode_kwargs to the encoder."""
+   my_encode_kwargs = {"batch_size": 999}
-   my_encode_kwargs = {"no_one_uses_this_args": "but_its_here"}
    class TestEncoder(Encoder):
        def encode(self, sentences, prompt_name: str | None = None, **kwargs):
            assert kwargs == my_encode_kwargs
            return np.zeros((len(sentences), 10))

    if isinstance(task_name, AbsTask):
        tasks = [task_name]
    else:
        tasks = mteb.get_tasks(tasks=[task_name])

    eval = mteb.MTEB(tasks=tasks)

    # Test that the task_name is passed down to the encoder
    model = TestEncoder()
    eval.run(
        model,
        output_folder="tests/results",
        overwrite_results=True,
+       **my_encode_kwargs,
-       encode_kwargs=my_encode_kwargs,
    )
```

And yielded 4 failures + 7 passed when running `pytest tests/test_mteb.py::test_encode_kwargs_passed_to_all_encodes`:
```
================================================================================== short test summary info ==================================================================================
FAILED tests/test_mteb.py::test_encode_kwargs_passed_to_all_encodes[TwitterHjerneRetrieval] - AssertionError: assert {'batch_size'...ss_bar': True} == {'batch_size': 999}
FAILED tests/test_mteb.py::test_encode_kwargs_passed_to_all_encodes[BrazilianToxicTweetsClassification] - AssertionError: assert {'batch_size'...ss_bar': True} == {'batch_size': 999}
FAILED tests/test_mteb.py::test_encode_kwargs_passed_to_all_encodes[FaroeseSTS] - AssertionError: assert {'batch_size'...ss_bar': True} == {'batch_size': 999}
FAILED tests/test_mteb.py::test_encode_kwargs_passed_to_all_encodes[SummEval] - AssertionError: assert {'batch_size'...ss_bar': True} == {'batch_size': 999}
=============================================================================== 4 failed, 7 passed in 38.66s ===================================================
```

mteb/abstasks/AbsTask.py
```diff
@@ -92,6 +94,7 @@ def evaluate(
             model: Sentence embedding method. Implements a encode(sentences) method, that encodes sentences and returns a numpy matrix with the
                 sentence embeddings
             split: Which datasplit to be used.
+            encode_kwargs: Additional keyword arguments that are passed to the models encode method.
```
isaac-chung, line 97: To match README
```suggestion
            encode_kwargs: Additional keyword arguments that are passed to the model's `encode` method.
```

README.md
```diff
@@ -127,6 +127,13 @@ from mteb import MTEB_MAIN_EN
 evaluation = mteb.MTEB(tasks=MTEB_MAIN_EN, task_langs=["en"])
 ```
 
+### Passing in `encode` arguments
+
+To pass in arguments to the model's `encode` function you can use the encode keyword arguments (`encode_kwargs`):
```
isaac-chung, line 132:
```suggestion
To pass in arguments to the model's `encode` function, you can use the encode keyword arguments (`encode_kwargs`):
```

KennethEnevoldsen:
Ahh that sounds great. I am not sure if you are suggesting we change the test or that it was just to make sure that everything works?

mteb/evaluation/evaluators/SummarizationEvaluator.py
```diff
@@ -42,10 +42,18 @@ def __init__(
...
-    def __call__(self, model: Encoder | EncoderWithSimilarity):
+    def __call__(
+        self,
+        model: Encoder | EncoderWithSimilarity,
+        *,
+        encode_kwargs: dict[str, Any] = {},
+    ):
+        # set default for encode_kwargs
+        if "batch_size" not in encode_kwargs:
+            encode_kwargs["batch_size"] = 32
```
isaac-chung, line 55: I noticed that sometimes this if statement is in `__init__()` for some evaluators, and sometimes in `__call__()`. Since the tests are passing, we know the arg will be passed into `encode`, so we're good there. Just a note for us that this might be something we'd want to revisit later to standardize.

isaac-chung (approved):
Looks good. Added a few small comments as well.

isaac-chung:
> Ahh that sounds great. I am not sure if you are suggesting we change the test or that it was just to make sure that everything works?

it was just to make sure that everything works - sanity check for me 🤠

mteb/evaluation/evaluators/SummarizationEvaluator.py
KennethEnevoldsen: Yes you are completely correct. Already partly standardized it, but already made quite large changes and wanted to keep it as simple as possible (especially for retrieval and reranking the classes have also grown excessively large, which makes it hard to debug)

isaac-chung: Makes sense!

KennethEnevoldsen: Perfect will that I will add it in then.

Muennighoff:
I think this PR added a bug: As device now defaults to `None` in the CLI arguments if no device is passed via the CLI & e5 instruct is used, then in its init it will be provided with device=None which will fail as it tries to place the model onto this device of None.

KennethEnevoldsen:
Added fix in #1048

PR #1031, opened 2024-07-02
ci: fix docs ci again

KennethEnevoldsen:
@isaac-chung just so you see that it is fixed.


isaac-chung:
Thanks for taking this!

PR #1034, opened 2024-07-03
add test for results

Samoed:
## Checklist

I have added tests to ensure that the main metric is included in the results of each task. This adjustment was necessary because `create_meta` requires `main_score` to be present in the results. All tasks pass tests except `AbsTaskPairClassification` and its subtasks. These changes are reflected in `test_mteb_load_results.py`.

Additionally, I have incorporated tests to verify the actual results stored in the `/results/` folder. These tests are designed to identify any discrepancies that may arise when datasets are relocated. Once the necessary corrections are made post-relocation, these tests can be safely removed.

I've also marked failing tests with xfail to maintain CI stability.


tests/test_load_results/test_mteb_load_results.py
```diff
@@ -23,3 +46,230 @@ def test_mteb_load_results():
...
+            {
+                "text": ["text"],
+                "human_summaries": [["text"]],
+                "machine_summaries": [["text"]],
+                "relevance": [[0.1]],
+            },
+        ),
+    ],
+)
+def test_load_results_scores(task, dataset):
```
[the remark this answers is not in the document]
KennethEnevoldsen: (general comment)

tests/test_load_results/test_mteb_load_results.py
```diff
@@ -23,3 +46,230 @@ def test_mteb_load_results():
...
+                    subset_keys = list(subset_score.keys())
+                    subset_keys.remove("hf_subset")
+                    subset_keys.remove("languages")
+                    subset_keys = sorted(subset_keys)
+                    assert (
+                        res_keys == subset_keys
+                    ), f"{result_file} for task {task_name} keys not matching expected {res_keys}, actual {subset_keys}"
+
+
+@pytest.mark.skip
```
KennethEnevoldsen, line 187: why skip?

tests/test_load_results/test_mteb_load_results.py
```diff
@@ -23,3 +46,230 @@ def test_mteb_load_results():
...
+
+    encoder = MockEncoder()
+    res = example_task.evaluate(encoder, split="test")["default"]
+    for task_class in all_subclasses_classes:
+        assert (
+            task_class.metadata.main_score in res
+        ), f"{task_class.metadata.name} have main_score {task_class.metadata.main_score} that not in {res.keys()}"
+
+
+@pytest.mark.xfail
```
KennethEnevoldsen, line 245: If we add a fail we should have a docstring which explains why it fails

tests/test_load_results/test_mteb_results.py
KennethEnevoldsen, line 1: will you add a docstring to both of these I am not quite sure how they differ

KennethEnevoldsen (commented):
I am not entirely sure what the intention is of all of these tests. Adding some docstring will make the review easier (and with fewer errors).

However, it seems like we test if the "main_score" is present and that is actually not required (not all older formats have it). However, loading the results using MTEBResults should fix many of these issues. So what we should really test for is that the main_score can be loaded using MTEBResults. We might, however, want to add to MTEBResults a test that checks if the current "main_score" (e.g. 0.66) and the set main score (e.g. "accuracy": 0.66) match. That way if we change the main score of the task we should get an error when loading historic results.

It also seems to test the `results` folder. Which as it stands is not used in any leaderboard. We are currently in the process of designing a good process for the results in the future (@Muennighoff), however, currently existing benchmarks results are to be located at the huggingface repository and MMTEB results and the github [repo](https://github.com/embeddings-benchmark/results/)

Let me know what you think

tests/test_load_results/test_mteb_load_results.py
```diff
@@ -23,3 +46,230 @@ def test_mteb_load_results():
     known_revision = "6d9c09a789ad5dd126b476323fccfeeafcd90509"
     assert known_model in results
     assert known_revision in results[known_model]
+
+
+@pytest.mark.xfail
+@pytest.mark.parametrize("task", MTEB().tasks_cls)
+def test_load_results_main_score_in_real_results(task):
```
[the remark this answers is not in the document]
Samoed: Yes, but when I tried to `create_meta` for models that run before some changes in `STS` create meta just failed, because it needed to have task main score in all scores. This was the problem that I tried to fix with this PR. https://github.com/embeddings-benchmark/mteb/blob/main/mteb/cli.py#L272

tests/test_load_results/test_mteb_load_results.py
Samoed: In this implementation of test, I have problems with creating evaluation `DenseRetrievalExactSearch.__init__() got multiple values for argument 'model'`. And I left it like that to show my pr at start

Samoed:
In `task_main_results`, I've added tests to ensure that the `main_score` of each task is included in the result metrics. This is crucial because currently, not all tasks pass this test, and it's uncertain what their main metric should be.

The tests from `test_mteb_load_results` could eventually be relocated to the repository with the results. This would help ensure that datasets are in the expected format.

tests/test_load_results/test_mteb_load_results.py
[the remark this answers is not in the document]
KennethEnevoldsen: Ahh right. The new fix to STS should probably fix that. This seems like a test of create_meta though rather than of the results format

tests/test_load_results/test_mteb_load_results.py
```diff
@@ -114,9 +106,10 @@ def test_load_results_main_score_in_real_results(task):
     ],
 )
 def test_load_results_scores(task, dataset):
+    """Test that all keys from actual task results presented in real task result"""
```
KennethEnevoldsen, line 109: Hmm instead I don't think this is something we actually want to test for. It prevents us from adding new metrics. I would delete this test. The only thing we really want to test for is the main metric e.g. "accuracy" is there.

tests/test_load_results/test_mteb_results.py
```diff
@@ -160,42 +149,41 @@ def test_revision_layer(json_path, results_path):
         ], f"Unexpected path structure for version {json_version}: '{relative_path}'"
 
 
-@pytest.mark.xfail
+@pytest.mark.xfail(reason="PairClassification and Reranking fail that test")
```
KennethEnevoldsen, line 152: Since tasks type can have different main metrics I would instead check that the main metric of the task is present in its results.

```
results = mteb.load_results()
# results is a dict["model_name", dict["revision": MTEBResults(...)]
# this function loads the results from the results repo

for model, revscores in results.items():
  for rev, mtebscore in revscores.items():
    task = mteb.get_task(mtebscore.task_name)
    score = mtebscore.get_score(getter = lambda scores: scores[task.metadata.main_score])
    assert score == mtebscore.get_score(getter = lambda scores: scores["main_score"])
```

KennethEnevoldsen: This way as long as the main score is present we shouldn't have any issues.

Samoed: Your code tests the existing datasets, but what I've done is check whether the `main_metric` from each task is included in the task evaluation results in the current implementation. This test focuses on a slightly different aspect. I plan to add this code to a separate file, as it serves a distinct purpose from the existing tests.

Now PairClassifier will output `['similarity', 'cosine', 'manhattan', 'euclidean', 'dot', 'max', 'main_score']`, but multilingual PairClassification tasks currently have 'ap' as main metric. This test is focusing to test behavior like that

KennethEnevoldsen: I am not sure I get it. I don't see a case where this fails and I am sure that I want it to fail (and which wouldn't be caught by the above approach).

KennethEnevoldsen: When I run a task i get (ap included):

```
{
  "dataset_revision": "7335288588f14e5a687d97fc979194c2abe6f4e7",
  "evaluation_time": 5.820026159286499,
  "kg_co2_emissions": null,
  "mteb_version": "1.12.59",
  "scores": {
    "test": [
      {
        "cosine": {
          "accuracy": 0.6239067055393586,
          "accuracy_threshold": 0.774509072303772,
          "ap": 0.6484016684922852,
          "f1": 0.6789940828402367,
          "f1_threshold": 0.6167908310890198,
          "precision": 0.5510204081632653,
          "recall": 0.884393063583815
        }, 
        ...
        "hf_subset": "default",
        "languages": [
          "fas-Arab"
        ],
        "main_score": 0.6484016684922852,
        ...
        }
      }
    ]
  },
  "task_name": "FarsTail"
}
```
do you have cases where you don't get ap?

KennethEnevoldsen: but I def. see that it is not in the outermost layers (similar to how it used to be with STS). This seems like format to standardize.

Samoed: Yes it is appearing in "sublevel" of metrics, but I thought it should be in "top level" like in STS. Because previously there was 
```
cosine_pearson

Comment (KennethEnevoldsen) [reply tests/test_load_results/test_mteb_results.py:152]: Completely agree, then we will get "cosine_ap", "manhattan_ap", "max_ap" etc., and convert the main metric to "max_ap". Will you create an issue on this? We will need to discuss the issue problem with a few of the others as well to make sure that we don't break compatibility with the existing leaderboard.

File: tests/test_load_results/test_mteb_load_results.py [hunk 7]
```diff
@@ -23,3 +32,45 @@ def test_mteb_load_results():
...
+                assert (
+                    task.metadata.main_score in subset_score
+                ), f"{result_file} not have {task.metadata.main_score} for task {task_name}"
+                assert (
+                    subset_score[task.metadata.main_score] == subset_score["main_score"]
+                ), result_file
+
+
+@pytest.mark.xfail(reason="Some models results have wrong main score")
+def test_load_results_scores():
```
KennethEnevoldsen, line 62: I believe this is the test of interest. I would remove the one above

tests/test_load_results/test_mteb_results.py
```diff
@@ -145,5 +149,148 @@ def test_revision_layer(json_path, results_path):
...
+            {
+                "text": ["text"],
+                "human_summaries": [["text"]],
+                "machine_summaries": [["text"]],
+                "relevance": [[0.1]],
+            },
+        ),
+    ],
+)
+def test_main_score_in_task_result(task, dataset):
```
KennethEnevoldsen, line 196: I would delete these test and instead move the logic to:

```
tests/test_mteb::test_mteb_task
```

```
def test_mteb_task(task: Union[str, AbsTask], model_name: str):
    """Test that a task can be fetched and run"""
    eval = MTEB(tasks=[task])
    results = eval.run(model, output_folder="tests/results", overwrite_results=True)
    # add test of results here
```

Samoed: I wrote it this way to ensure the test runs quickly without requiring data downloads or complex models. However, I'm open to making the change to align with the suggested solution

KennethEnevoldsen: The goal here is to make tests that are easy to maintain going forward. Would def. be more than happy to create a list of mock datasets for all the dataset and create two versions of the test test_mteb_task one for integration and one using only the mock datasets.

Muennighoff:
Yes, let's remove the results folder from this repository now as it seems to be confusing contributors/users? Should I add all results from Hugging Face to https://github.com/embeddings-benchmark/results? cc @orionw

orionw:
That sounds good to me.  I hope to look at the auto-calculation / Github action stuff this weekend.  If we put all the results in one place it will make it easy to work with (especially if it's a Github instead of a HF dataset).

Samoed:
I updated tests. Now in test running `average_word_embeddings_komninos` on real datasets. It took nearly 9m on my machine. Now `test_mteb_task` failing in two cases. `MIRACLReranking have NDCG@10(MIRACL)` that not in results, and I can't run InsturctionRetrival now. I will create separete PR to fix InsturctionRetrival

KennethEnevoldsen:
@Samoed based on your PRs and a few other elements I have actually restructured the tests in #1070. This also notably speeds up the benchmark (slowest is ~20s on my laptop)

KennethEnevoldsen:
Will close this for now as results have been moved to https://github.com/embeddings-benchmark/results

feel free to add create a PR with the results there. Additionally the test would also be nice to get added, but it would be nice t have https://github.com/embeddings-benchmark/mteb/pull/1072 in first.

This PR have been very helpful though in both highlighting problems and helping us update the tests suite

PR #1038, opened 2024-07-04
Added models x tasks which haven't yet been run

KennethEnevoldsen:


Muennighoff:
Some results that should already have been there for models but are somehow not recognized by the script:
- PAWSX (Maybe due to confusion with PawsX?)
- SemRel24STS

All others should have either been added now or I opened an issue for them (the dialogue ones) 👍 Lmk if there's more I should run?

KennethEnevoldsen:
Will take a look at this

KennethEnevoldsen:
Ahh so it turns out that the macOS file system is case-incensitive (by default) so the PAWSX and PawsX get merged automatically... A bit frustrating, we should probably rename one of them (I would suggest changing PawsX to PAWSXPairClassification). Sadly both of them are used in a benchmark (is it possible to fix this on the french bechmark?)

Muennighoff:
Agree that we should fix this! Also ran into that problem a few times

mteb/tasks/PairClassification/multilingual/PawsXPairClassification.py
```diff
@@ -8,13 +8,13 @@
...
         dataset={
             "path": "google-research-datasets/paws-x",
             "revision": "8a04d940a42cd40658986fdd8e3da561533a3646",
             "trust_remote_code": True,
         },
-        description="",
+        description="{PAWS-X: A Cross-lingual Adversarial Dataset for Paraphrase Identification",
         reference="https://arxiv.org/abs/1908.11828",
         category="s2s",
         type="PairClassification",
```
Muennighoff, line 20: Do you not want to also rename the class? 
Also if we rename this, we should probably rename all corresponding result files. Unfortunately we cannot rename the models meta of other people so we'll still need to allow for that in the leaderboard code.

KennethEnevoldsen:
@Muennighoff figured out that this happens due to some models being run on the "dev" split of the SemRel24STS. This makes it seems like other models don't have it (but it is not specified as required in the task). I have updated the list now to fix these cases. 

Note that this required a lot of fixes to the results repo (due to adding the older files)

Muennighoff:
Amazing! I put the new results here: https://github.com/embeddings-benchmark/results/pull/6 except for the failures here: https://github.com/embeddings-benchmark/mteb/issues/1064

KennethEnevoldsen:
Merging this PR as well @Muennighoff the FP16 case should have been fixes if you want to rerun that one

PR #980, opened 2024-06-23
Voyage bclavie/mmarco-japanese-hard-negatives

jordiclive:
dataset mentioned in #920, difficult Japanese version of MMarcoReranking.

## Checklist

- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 

### Adding datasets checklist

**Reason for dataset addition**: ... 

- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [x] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 

### Adding a model checklist

 - [ ] I have filled out the ModelMeta object to the extent possible
 - [ ] I have ensured that my model can be loaded using
   - [ ] `mteb.get_model(model_name, revision_id)` and
   - [ ] `mteb.get_model_meta(model_name, revision_id)`
 - [ ] I have tested the implementation works on a representative set of tasks.

PR #1057, opened 2024-07-07
points

jordiclive:
points for [dataset](https://github.com/embeddings-benchmark/mteb/pull/980)

### DIFF
--- a/docs/mmteb/points/980.jsonl
+++ b/docs/mmteb/points/980.jsonl
@@ -1,0 +1,2 @@
+{"GitHub": "jordiclive", "New dataset": 3}
+{"GitHub": "dokato", "Review PR": 2}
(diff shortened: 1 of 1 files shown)


docs/mmteb/points/980.jsonl
```diff
@@ -0,0 +1,2 @@
+{"GitHub": "jordiclive", "New dataset": 3}
```
KennethEnevoldsen, line 1: Unsure how you got 3 points here?

KennethEnevoldsen (approved):
Only a minor thing

docs/mmteb/points/980.jsonl
jordiclive: should be 2, right?

PR #1059, opened 2024-07-07
Multi-gpu/simpler models; Add gte

Muennighoff:
Made more models implemented in MTEB Multi-GPU to be faster for encoding. Sentence Transformer doesn't support this nicely afaict so switched to GritLM - verified it leads to the same embeddings for SFR. Since installing gritlm is very [lightweight](https://github.com/ContextualAI/gritlm/blob/fdeb0f78dbb11268233d9777f169df790de3d2b7/setup.py#L61) i think it is fine to also require installing it for gte/sfr but we can also cp over the single file that it is to avoid the installation: https://github.com/ContextualAI/gritlm/blob/main/gritlm/gritlm.py

Also rmv'd the results folder as we're moving it to https://github.com/embeddings-benchmark/results IIUC? Should we move the remaining results in this repo over or just delete them? cc @orionw


mteb/models/e5_instruct.py
```diff
@@ -236,6 +244,7 @@ def loader_inner(**kwargs: Any) -> Encoder:
         "intfloat/multilingual-e5-large-instruct",
         "baa7be480a7de1539afce709c8f13f833a510e0a",
         max_length=512,
+        #max_batch_size=2,
```
KennethEnevoldsen, line 247:
```suggestion
```

mteb/models/e5_instruct.py
```diff
@@ -48,10 +48,17 @@ def __init__(
...
+        device = torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
         self.model = AutoModel.from_pretrained(model_name, **kwargs).to(device)
+        self.model.eval()
+        self.device = device
         self.max_length = max_length
         self.max_batch_size = max_batch_size
+        self.gpu_count = torch.cuda.device_count()
+
+        if self.gpu_count > 1:
+            print(f"----------Using {self.gpu_count} data-parallel GPUs----------")
```
KennethEnevoldsen, line 60: I would use logging instead

mteb/models/salesforce_models.py
```diff
@@ -1,74 +1,58 @@
...
+                kwargs["instruction"] = instruction
+                return super().encode(*args, **kwargs)
+
+            def encode_corpus(self, *args, **kwargs):
+                kwargs["is_query"] = False
+                return super().encode_corpus(*args, **kwargs)
+    except ImportError:
+        raise ImportError(
+            "Please install `pip install gritlm` to use SFR_Embedding_2_R."
+        )
```
KennethEnevoldsen, line 33: Just wrap import statement in try except

KennethEnevoldsen (approved):
Looks good on my end. 

re. removed results folder I would probably add it should we add it to gitignore as well?

re. results I believe it is only really the Russian models/tasks where the results are relevant the rest are from test runs to show that everything works

orionw:
> Should we move the remaining results in this repo over or just delete them? cc @orionw

Yes, we should be good to delete them and move to the other repo!

KennethEnevoldsen:
This PR seems done. I have just updated from main and enable automerge

PR #1079, opened 2024-07-12
fix: ensure that CQADupstackRetrieval is included in results if possible

KennethEnevoldsen:
Example from running mteb create_meta with new code:

```
[...]
  - dataset:
      config: default
      name: MTEB CQADupstackRetrieval (default)
      revision: CQADupstackRetrieval is a combined dataset
      split: test
      type: CQADupstackRetrieval is a combined dataset
    metrics:
    - type: main_score
      value: 46.40058333333334
    - type: ndcg_at_10
      value: 46.40058333333334
    task:
      type: Retrieval
[...]
```

## Checklist


mteb/cli.py
```diff
@@ -251,6 +253,56 @@ def add_run_parser(subparsers) -> None:
     parser.set_defaults(func=run)
 
 
+def potentially_add_cqadupstack_to_results(results: list[mteb.MTEBResults]) -> None:
+    """If all CQADupstack tasks are present in the results, combine them into a single CQADupstackRetrieval task and add it to the results."""
+    TASK_LIST_CQA = set(
```
Muennighoff, line 258: Instead of set([a, b, c]) why not {a, b, c}?

mteb/cli.py
```diff
@@ -251,6 +253,56 @@ def add_run_parser(subparsers) -> None:
...
+            "CQADupstackTexRetrieval",
+            "CQADupstackUnixRetrieval",
+            "CQADupstackWebmastersRetrieval",
+            "CQADupstackWordpressRetrieval",
+        ]
+    )
+
+    task_names = {result.task_name for result in results}
+
+    if not TASK_LIST_CQA.issubset(task_names):
```
Muennighoff, line 277: Hm what if people run tasks separately e.g.  via an array slurm job? Maybe it makes more sense to check if all results are in the result folder 🤔

mteb/cli.py
```diff
@@ -251,6 +253,56 @@ def add_run_parser(subparsers) -> None:
...
+                "hf_subset": "default",
+                "languages": ["eng_Latn"],
+            }
+        ]
+    }
+
+    result = mteb.MTEBResults(
+        task_name="CQADupstackRetrieval",
+        dataset_revision="CQADupstackRetrieval is a combined dataset",
+        mteb_version="NA",
```
Muennighoff, line 298: Why `NA` here? Should we not put a version?

mteb/cli.py
```diff
@@ -251,6 +253,56 @@ def add_run_parser(subparsers) -> None:
...
+        scores=scores,
+        evaluation_time=evaluation_time,
+        kg_co2_emissions=None,
```
Muennighoff, line 301: If we sum up the eval times, shouldn't we sum up the kg_co2 emissions too?

mteb/load_results/mteb_results.py
```diff
@@ -253,6 +256,9 @@ def from_disk(cls, path: Path, load_historic_data: bool = True) -> MTEBResults:
             path: The path to the file to load.
             load_historic_data: Whether to attempt to load historic data from before v1.11.0.
         """
+        import mteb
+
+        mteb.__version__
```
Muennighoff, line 261: Why is this code needed? 🤔

Muennighoff (commented):
Nice! 🙌 Maybe it would be good to also automatically merge it when evaluating not via the cli but importing mteb in Python? So instead of specific to the cli, maybe it should be 

Also maybe we should leave in the merging script at least for some time in case someone wants to merge older results or there are scenarios where the automatic merging is not triggered?

Muennighoff:
Oh and one more question, did we decide to keep the `(default)` in the metadata? Do we still need to fix this: https://huggingface.co/spaces/mteb/leaderboard/discussions/132#668efce065639f1fd9cc3fee ?

mteb/cli.py
KennethEnevoldsen: It is not added to the markdown so I just avoided adding it.

mteb/load_results/mteb_results.py
KennethEnevoldsen: Should be removed

KennethEnevoldsen:
> Oh and one more question, did we decide to keep the (default) in the metadata? Do we still need to fix this: https://huggingface.co/spaces/mteb/leaderboard/discussions/132#668efce065639f1fd9cc3fee ?

This does not fix this case no. I think we should change the leaderboard to allow for default (however that might be harder than simply changing it here? WDYT?

mteb/cli.py
KennethEnevoldsen: it is not in the metadata and never saved so just avoided dealing with potentially multiple versions

mteb/cli.py
KennethEnevoldsen: results is all results loaded from the folder and if they contain the CQA set we add the new metric.

mteb/cli.py
KennethEnevoldsen: yea probably better

KennethEnevoldsen:
> Also maybe we should leave in the merging script at least for some time in case someone wants to merge older results or there are scenarios where the automatic merging is not triggered?

I believe the new should work for older cases as well (due to the backward compatible loading). but I am fine with leaving it in as well

KennethEnevoldsen:
> Nice! 🙌 Maybe it would be good to also automatically merge it when evaluating not via the cli but importing mteb in Python? So instead of specific to the cli, maybe it should be

Seems like this was never finished, not quite sure what you mean?

Muennighoff:
> I believe the new should work for older cases as well (due to the backward compatible loading). but I am fine with leaving it in as well

Nice if it works for older cases to (without having to necessarily run an evaluation but just merge as is) then that's  fine I think! Should it be mentioned somewhere here: https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_model.md#adding-a-model-to-the-mteb-leaderboard ?

> > Nice! 🙌 Maybe it would be good to also automatically merge it when evaluating not via the cli but importing mteb in Python? So instead of specific to the cli, maybe it should be
> 
> Seems like this was never finished, not quite sure what you mean?

Sorry 😅 I meant if people are evaluating via MTEB in Python then it wouldn't be triggered right? So maybe it should check everytime a CQADupstack task is run or something rather than everytime the CLI is used?

Muennighoff:
> This does not fix this case no. I think we should change the leaderboard to allow for default (however that might be harder than simply changing it here? WDYT?

Yeah it seems like changing it here is simpler I think

KennethEnevoldsen:
> Sorry 😅 I meant if people are evaluating via MTEB in Python then it wouldn't be triggered right? So maybe it should check everytime a CQADupstack task is run or something rather than everytime the CLI is used?

Aah right, yes you can't run "CQADupstackRetrieval" within using mteb (cli or python). But if you run all its constituents `mteb create_meta` will create the aggregate score. 

I have considered implementing it as a "benchmark" and then you get a score pr. benchmark similar to a task. Then we can also have that a 'benchmark' as a task within another benchmark, but that requires a lot more utility handling for benchmarks than we currently have (though in the future I would love to add it, e.g. allow benchmarks to implement their own aggregations scores)

Muennighoff:
> > Sorry 😅 I meant if people are evaluating via MTEB in Python then it wouldn't be triggered right? So maybe it should check everytime a CQADupstack task is run or something rather than everytime the CLI is used?
> 
> Aah right, yes you can't run "CQADupstackRetrieval" within using mteb (cli or python). But if you run all its constituents `mteb create_meta` will create the aggregate score.
> 
> I have considered implementing it as a "benchmark" and then you get a score pr. benchmark similar to a task. Then we can also have that a 'benchmark' as a task within another benchmark, but that requires a lot more utility handling for benchmarks than we currently have (though in the future I would love to add it, e.g. allow benchmarks to implement their own aggregations scores)

I guess if `create_meta` does the aggregation it is fine for now 👍 We'll see if people run into issues!

PR #1080, opened 2024-07-12
Remove secret token from documentation workflow

Samoed:
It seems that the workflow breaks when external contributors open PRs because they don't have access to the secret token. As far as I understand, this token is not required for this workflow.
Example https://github.com/embeddings-benchmark/mteb/actions/runs/9913313640/job/27390053173


KennethEnevoldsen:
@Samoed this will fail during the push segment (skipped for PRs). I have create another fix #1081

PR #1096, opened 2024-07-18
ci: Splitting doc CI in two to ensure that it runs on PRs AND on main

KennethEnevoldsen:
I am having a few issues with this CI. I believe this should work, but I am not sure exactly why the previous CI doesn't work (it did for a short while). It might be due to the token not being passed correctly. 

This configuration essentially just takes the previous CI that did work on main and adds a completely separate CI for PRs. 

To avoid having to change stuff in the future it convert the build command into `make build-docs`.

@isaac-chung would love a second pair of eyes on this if you have the time

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -8,7 +8,8 @@ on:
   pull_request:
 
 jobs:
-  create-table:
+  create-table-on-pr:
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
Well I want it to run during a PR as well. Did try to combine it into one (the one we currently have), but couldn't get it to work. I have tried to make it a bit more clear (and also added the conditional to ensure that the push workflow isn't run)

isaac-chung (approved):
This is more clear, thanks!

KennethEnevoldsen:
And it seems to work on main as well - wonders!

PR #1103, opened 2024-07-20
[MIEB] Add more image classification and zero shot classification datasets

isaac-chung:
Second batch of image classification datasets, which covers
- MNIST
  - [x] Linear probe 
  - [x] zero shot
- FER2013
  - [x] Linear probe 
  - [x] zero shot
- Stanford Cars
  - [x] Linear probe 
  - [x] zero shot
- BirdSnap
  - [x] Linear probe 
  - [x] zero shot
- EuroSAT
  - [x] Linear probe 
  - [x] zero shot

## Checklist


gowitheflow-1998 (approved):
Thanks for adding these and uploading Birdsnap! Zero-shot results look pretty good. Only EuroSAT is much lower than the CLIP paper. Any idea? We can put a note in the sheet for the ones that have much lower zero-shot scores and look into them after all classification ones are added. Also for birdsnap the test set seems fewer than the original dataset and CLIP, is it because of fewer available online now? Can look at these after though!

isaac-chung:
For EuroSAT, it might be due to the class label names not having a space in between words (`AnnualCrop` vs `annual crop land` in the paper). It's hard to tell beyond the 5 examples in the paper though (i.e. what should something like `Industrial` be formatted to).

```
In [1]: from datasets import load_dataset
In [2]: ds = load_dataset("timm/eurosat-rgb")
In [3]: ds["test"].features
Out[3]: 
{'image': Image(mode=None, decode=True, id=None),
 'label': ClassLabel(names=['AnnualCrop', 'Forest', 'HerbaceousVegetation', 'Highway', 'Industrial', 'Pasture', 'PermanentCrop', 'Residential', 'River', 'SeaLake'], id=None),
 'image_id': Value(dtype='string', id=None)}
```

isaac-chung:
[This commit](6b49181e) fixes the issue and matches the paper's performance with B/32 @ 49.7.

PR #1105, opened 2024-07-21
[MIEB] Add more image classification and zero shot datasets

isaac-chung:
Third batch of image classification datasets, which covers
- STL-10
  - [x] Linear probe 
  - [x] zero shot
- RESISC45
  - [x] Linear probe 
  - [x] zero shot
- SUN397
  - [x] Linear probe 
  - [x] zero shot
- DTD
  - [x] Linear probe 
  - [x] zero shot

## Checklist


gowitheflow-1998:
looks great! Might worth looking at SUN397 and RESISC45's zero shot results in the next commit. For SUN 397, it might be the current implementation of simply joining all hierarchical categories by `name = " ".join(name.split("/")[1:]).replace("_", " ")`. The single character is joined as well? e.g., `"/s/subway_station/platform"` -> `'s subway station platform'`. Not sure keeping label of what level will align with the original setting (perhaps "subway station"?). For RESISC45 no ideas yet

gowitheflow-1998 (approved):

isaac-chung:
For SUN397, the paper uses `a photo of a podium indoor`. That does not include the single character (not joined as well, as we only take [1:] from the splitted list). So what we have here matches the CLIP paper.

For RESISC45, the paper uses `satellite imagery of roundabout`, which is used here as well.

The actual cause in difference in performance is the dataset size (which also means the actual data used). See table below on the difference between train set and test set sizes: (from CLIP paper table 9, page 39)

| Dataset  | ours, train | ours, test | paper, train | paper, test |
|----------|-------------|------------|--------------|-------------|
| Sun397   | 76127       | 21750      | 19850        | 19850       |
| RESISC45 | 18900       | 6300       | 3150         | 25200       |

This should be fine for the benchmark as we'll run all models on the same version of the dataset.

gowitheflow-1998:
[1:] will start with the single character as the 0 is "" when split by / no?

gowitheflow-1998:
the dataset size for RESISC45 makes sense! I can see a few available ones on HF all use similar splits. In general as long as we align with common splits people use it's good.

isaac-chung:
> [1:] will start with the single character as the 0 is "" when split by / no?

Oh yes you're right. I should have had coffee first next time. Pushed an update to `mieb` in [this commit](77b0e35d).

PR #1106, opened 2024-07-22
fix: fix bug-causing spelling error in function name of e5-mistral-instruct

anpalmak2003:
An error was found in the function's name, which caused a bug, and not all tasks were completed, due to an error in the model's embeddings. There were sometimes zeros and infinity values.

### DIFF
--- a/mteb/models/e5_instruct.py
+++ b/mteb/models/e5_instruct.py
@@ -193,7 +193,7 @@ class E5MistralWrapper(E5InstructWrapper):
             sequence_lengths,
         ]
 
-    def get_embbeding_from_output(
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen (approved):
Thanks for the fix

anpalmak2003:
> Thanks for the fix - this looks fine. I have enabled automerge so it should merge assuming all tests pass.

can I get a point for fixing a bug? https://github.com/embeddings-benchmark/mteb/blob/main/docs/mmteb/points.md

KennethEnevoldsen:
Sure! Feel free to create a PR adding 2 points.

PR #1109, opened 2024-07-24
fix MultilingualSentimentClassification

Samoed:
Now if run MultilingualSentimentClassification with any language as subset it will raise error

Example
```
tasks = mteb.get_tasks(tasks=["MultilingualSentimentClassification"], languages=["eng"])
evaluation = mteb.MTEB(tasks)
evaluation.run(model, output_folder="results", verbosity=2, overwrite_results=True)
```

### DIFF
--- a/mteb/tasks/Classification/multilingual/MultilingualSentimentClassification.py
+++ b/mteb/tasks/Classification/multilingual/MultilingualSentimentClassification.py
@@ -96,8 +96,8 @@ class MultilingualSentimentClassification(AbsTaskClassification, MultilingualTas
     def dataset_transform(self):
         # create a train set from the test set for Welsh language (cym)
         lang = "cym"
-        _dataset = self.dataset[lang]
         if lang in self.dataset.keys():
+            _dataset = self.dataset[lang]
             _dataset = _dataset.class_encode_column("label")
             _dataset = _dataset["test"].train_test_split(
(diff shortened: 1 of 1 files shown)


mteb/tasks/Classification/multilingual/MultilingualSentimentClassification.py
```diff
@@ -96,13 +96,13 @@ class MultilingualSentimentClassification(AbsTaskClassification, MultilingualTas
     def dataset_transform(self):
         # create a train set from the test set for Welsh language (cym)
         lang = "cym"
```
KennethEnevoldsen, line 98: Seem like this line might have been here for testing and needs to be removed as well

KennethEnevoldsen (approved):
Only a minor edit - thanks for the fix.

mteb/tasks/Classification/multilingual/MultilingualSentimentClassification.py
Samoed: This is not for testing, but `cym` have only train split in it. I can move definition of `lang` inside if to make it more clear

KennethEnevoldsen: Ahh I see. Just ran a test on this an it seems to work as expected - thanks for the fix

PR #1114, opened 2024-07-25
fix export type

### LINKED ISSUE #1111: Creating model meta for HF model
I tied to export results to model_meta, but got several errors when pushing to huggingface. 
1. `Error: "model-index[0].results[23].dataset.type" with value "CQADupstackRetrieval is a combined dataset" fails to match the required pattern: /^(?:[\w-]+\/)?[\w-.]+$/` this caused by #1079 maybe this can be renamed to `CQADupstackRetrieval_is_a_combined_dataset`
2. `Error: "model-index[0].results[102].dataset.type" with value " " fails to match the required pattern: /^(?:[\w-]+\/)?[\w-.]+$/` When I tried to export `GPUSpeedTask` and `CPUSpeedTask` this caused by `GPUSpeedTask` and `CPUSpeedTask` doesn't have path in dataset, so task if empty. Maybe this can be ignored, but I'm just warning about this problem

Samoed:
Closes #1111
I've run `mteb create_meta` and got `type: CQADupstackRetrieval_is_a_combined_dataset` in type. I think that other datasets have same problems

### DIFF
--- a/mteb/cli.py
+++ b/mteb/cli.py
@@ -292,7 +292,7 @@ def potentially_add_cqadupstack_to_results(results: list[mteb.MTEBResults]) -> N
 
     result = mteb.MTEBResults(
         task_name="CQADupstackRetrieval",
-        dataset_revision="CQADupstackRetrieval_is_a_combined dataset",
+        dataset_revision="CQADupstackRetrieval_is_a_combined_dataset",
         mteb_version="NA",
         scores=scores,
         evaluation_time=evaluation_time,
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
> I think that other datasets have the same problems

Which ones are you thinking of? CQADupstackRetrieval is the only combined dataset.

Samoed:
Other dummy datasets. I don't know where they are using it, but exporting with this revision can lead to the same problems.

KennethEnevoldsen:
> Other dummy datasets. I don't know where they are using it, but exporting with this revision can lead to the same problems.

Ahh right indeed thanks for the clarification.

The dummy datasets exist only to support backward compatibility. In this regard, CQADupstackRetrieval is the only combined dataset of existing datasets (thus we need to actively handle it), however remaining dummy datasets have been outdated (removed, superseded, or renamed within MTEB) so they are only there to support loading of historic result files.

PR #1116, opened 2024-07-28
Fix test wrappers and export all tasks results in tests

Samoed:
## Checklist

Added mocks for:
- E5Wrapper
- BGEWrapper
- MxbaiWrapper
- GritLMWrapper (it is also used in some other places, but it can't be inherited, so only one test)


tests/test_benchmark/test_benchmark.py
```diff
@@ -11,24 +12,56 @@
...
+    # make sure that all tasks results can be exported
+    args = Namespace(
+        results_folder=output_folder + "/no_model_name_available/no_revision_available",
+        output_path=output_folder + "/model_card.md",
+        overwrite=True,
+    )
+    monkeypatch.setattr(
+        mteb, "get_task", lambda task_name, **kwargs: tasks_dict[task_name]
+    )
+    create_meta(args)
```
Samoed, line 51: I'm not sure that this is a good place for test exporting results, but I don't know where to put it.

KennethEnevoldsen: It seems like you have replaced an existing test, and it now tests something else. I would instead create a new test. I also don't believe that there is a need to monkey patch -- generally, I find that it hints that the code should be rewritten.

What I would do is create a function like:

```
create_meta_cli(args):
  # handle args, get tasks etc. 
  yml_dict = create_meta(task_results) 
  # save to disk
```

We can then test `create_meta` independently without monkeypatching.

tests/test_benchmark/test_gritlm_models.py
```diff
@@ -0,0 +1,63 @@
...
+
+import mteb
+from mteb.models.instructions import task_to_instruction
+from tests.test_benchmark.task_grid import MOCK_TASK_TEST_GRID
+
+pytest.importorskip("gritlm")
+from gritlm import GritLM  # noqa
+
+
+class MockGritLMWrapper(GritLM):
```
KennethEnevoldsen, line 14: Not entirely sure what this tests in the original implementation? Isn't it simply reimplementing the functionality?

tests/test_benchmark/test_benchmark.py
KennethEnevoldsen: This is not breaking as the create_meta cli will remain the same, while the create_meta becomes a testable function that the user can also call independently.

tests/test_benchmark/mock_models.py
```diff
@@ -30,3 +38,41 @@ def __init__(self):
 
     def encode(self, sentences, prompt_name: str | None = None, **kwargs):
         return torch.randn(len(sentences), 10, dtype=torch.bfloat16)
+
+
+class MockSentenceTransformer(SentenceTransformer):
```
KennethEnevoldsen, line 43:
```suggestion
class MockSentenceTransformer(SentenceTransformer):
   """
   A mock implementation of the SentenceTransformer intended to implement just the encode, method using the same arguments.
   """
```

tests/test_benchmark/mock_models.py
KennethEnevoldsen, line 1: Looks great

KennethEnevoldsen (commented):
Thanks for the addition. It is very convenient to have an active test of the model implementations. 

It seems like the test for create_meta needs a bit more work. I am not sure what it is intended to test which isn't already tested in its associated test. Let me know if I am missing something.

tests/test_benchmark/test_gritlm_models.py
Samoed: Yes, a bit. All GritLM wrappers covered inside functions, so I can't just inherit them, but I think test that GritLM models can run tests could be helpful

tests/test_benchmark/test_benchmark.py
Samoed: I'll move it to different PR

Samoed: I needed to monkeypatch it, because create_meta checks if task is existing mteb task, but mock tasks not in main part of library so it gives errors.

tests/test_benchmark/test_gritlm_models.py
KennethEnevoldsen: Hmm, that might be better done by creating an extremely small grim model (>100m parameters) and just testing that instead. Feel free to do that; otherwise, I would remove it for now and leave it for another PR.

tests/test_benchmark/test_benchmark.py
KennethEnevoldsen: Yeah, I can definitely see why you added it, but I think the rewrite is the better option. I can also see why you would do the monkey patch for the PR as well (it is generally less obtrusive than changing code).

tests/test_benchmark/test_gritlm_models.py
Samoed: Tried to run gritlm with `sentence-transformers/all-MiniLM-L6-v2`, but got error, that passed unexpected parameters by gritlm library. Removed this test

KennethEnevoldsen:
Looks good - thanks for the PR!

PR #1127, opened 2024-07-31
Add CLEVR and SciMMIR Image-Text Understanding tasks

imenelydiaker:


gowitheflow-1998:
nice thanks! For CLEVR, can you add the [count task](https://huggingface.co/datasets/clip-benchmark/wds_vtab-clevr_count_all) as well? Also both count and distance are worth adding linear probing.

Didn't realize SciMMIR can be a zero-shot classification as well (it is a retrieval task with image-caption) but I think is quite cool to add the zero-shot classification as you do here! The results indicate this is still a task that's not too easy so worth keeping. I can take care of the retrieval version later unless you have the bandwidth? I have just fixed image memory issues for all retrieval so it should be runnable with it as well.

gowitheflow-1998:
> I'm not aware of how much work it requires to add it as a retieval task. Would it require building a corpus and queries? I'll have a look at it and keep you posted, we'll add it in another PR anyway.

yeah we'll need to build corpus, queries and qrels like other retrieval tasks. No worries. lmk!

imenelydiaker:
> > I'm not aware of how much work it requires to add it as a retieval task. Would it require building a corpus and queries? I'll have a look at it and keep you posted, we'll add it in another PR anyway.
> 
> yeah we'll need to build corpus, queries and qrels like other retrieval tasks. No worries. lmk!

Added CLEVR count task. I think we'd do SciMMIR retrieval tasks in another PR (Text2Image and Image2Text). This one is ready 🙂

gowitheflow-1998 (approved):

gowitheflow-1998:
yeah looks nice! CLEVR's linear probing can be in another pr as well.

PR #1131, opened 2024-08-02
fix: Allow more linient TaskMetadata

### LINKED ISSUE #1121: Task Metadata Pydantic rules are too strict
Hi, wonderful package.

I am trying to run experimentations on my laptop on custom datasets but I end up spending a lot of time configuring each and every parameters of [TaskMetadata](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/abstasks/TaskMetadata.py#L138). It feels too constrained and I end up forced to add meaningless values that bite my later.

For example, I'd like to fetch multiple datasets but I HAVE TO put a revision string for each which makes the ordeal very cumbersome.

I am not using classic datasets but custom ones so being able to just put `None` for most of these parameters would be really helpful.

Any plans adding a `bypass_verification` for Pydantic ??

KennethEnevoldsen:
This fix allows for more lenient TaskMetadata when using the package, but maintain the validation for the package within the tests.

This is intended to ensure flexibility and ease of use, while maintaining high-quality doucementation of tasks

## Checklist

Fixes #1121

### DIFF
--- a/mteb/abstasks/TaskMetadata.py
+++ b/mteb/abstasks/TaskMetadata.py
@@ -174,40 +174,60 @@ class TaskMetadata(BaseModel):
 
     name: str
     description: str
-    type: TASK_TYPE
-    modalities: list[Literal["text"]]
-    category: TASK_CATEGORY
-    reference: STR_URL | None  # URL to documentation, e.g. published paper
+    type: TASK_TYPE | None = None
+    modalities: list[Literal["text"]] = ["text"]
+    category: TASK_CATEGORY | None = None
+    reference: STR_URL | None = None
 
-    eval_splits: list[str]
+    eval_splits: list[str] = ["test"]
     eval_langs: LANGUAGES
-    main_score: str  # Might want a literal here
+    main_score: str
 
-    date: tuple[STR_DATE, STR_DATE] | None  # When the data was collected
-    domains: list[TASK_DOMAIN] | None
-    task_subtypes: list[TASK_SUBTYPE] | None
-    license: str | None
+    date: tuple[STR_DATE, STR_DATE] | None = None
+    domains: list[TASK_DOMAIN] | None = None
+    task_subtypes: list[TASK_SUBTYPE] | None = None
+    license: str | None = None
 
-    annotations_creators: ANNOTATOR_TYPE | None
-    dialect: list[str] | None
+    annotations_creators: ANNOTATOR_TYPE | None = None
+    dialect: list[str] | None = None
 
-    sample_creation: SAMPLE_CREATION_METHOD | None
-    bibtex_citation: str | None
+    sample_creation: SAMPLE_CREATION_METHOD | None = None
+    bibtex_citation: str | None = None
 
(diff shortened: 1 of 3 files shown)


mteb/abstasks/TaskMetadata.py
```diff
@@ -174,40 +174,60 @@ class TaskMetadata(BaseModel):
...
     @field_validator("dataset")
-    def _check_dataset_path_is_specified(cls, dataset):
+    def _check_dataset_revision_is_specified(
+        cls, dataset: dict[str, Any]
+    ) -> dict[str, Any]:
+        cls.dataset_revision_is_specified(dataset)
+        return dataset
+
+    @staticmethod
+    def dataset_path_is_specified(dataset: dict[str, Any]) -> bool:
```
isaac-chung, line 220: Here it seems to either raise an error or return a value that will not be used later on. For static methods like this, maybe we don't need a return value. Would suggest removing the return line as well.
```suggestion
    def dataset_path_is_specified(dataset: dict[str, Any]) -> None:
```

mteb/abstasks/TaskMetadata.py
```diff
@@ -217,19 +237,18 @@ def _check_dataset_revision_is_specified(cls, dataset):
                 "Revision missing for the dataset %s. It is encourage to specify a dataset revision for reproducability.",
                 dataset["path"],
             )
-        return dataset
+        return True
 
-    @field_validator("eval_langs")
-    def _check_eval_langs(cls, eval_langs):
+    def eval_langs_are_valid(self, eval_langs: LANGUAGES) -> bool:
```
isaac-chung, line 242: Same here - the returned value is not used anywhere.
```suggestion
    def eval_langs_are_valid(self, eval_langs: LANGUAGES) -> None:
```

tests/test_TaskMetadata.py
```diff
@@ -383,13 +383,13 @@ def test_filled_metadata_is_filled():
...
 
-def test_all_metadata_is_filled():
+def test_all_metadata_is_filled_and_valid():
     all_tasks = get_tasks()
 
     unfilled_metadata = []
     for task in all_tasks:
         if task.metadata.name not in _HISTORIC_DATASETS:
-            if not task.metadata.is_filled():
+            if not task.metadata.is_filled() and task.metadata.is_validated():
```
isaac-chung, line 392: I wasn't able to find the `is_validated` method. Did you mean `validate_metadata`? Also, might need to wrap the and part in `()` before putting `not` in front 🤔

isaac-chung (commented):
Looks good so far! Added a few suggestions.

tests/test_TaskMetadata.py
KennethEnevoldsen: Good catch, apparently there is a pydantic function called the same.

mteb/abstasks/TaskMetadata.py
KennethEnevoldsen: Agree - I how now reverted it to None

KennethEnevoldsen:
Completely agree with the suggested. I have added them in

PR #1134, opened 2024-08-06
Update the syntax for Python 3.8

bryant1410:
Update the syntax to that of Python 3.8 (the minimum Python version supported by this library). I did this with [pyupgrade](https://github.com/asottile/pyupgrade):

```bash
pyupgrade $(find . -name "*.py" -type f) --py38-plus
```

It includes fixes such as removing unnecessary list creations and updating the typing.


bryant1410:
I reviewed all the changes and they make sense to me.

I had to manually intervene for LaTeX code blocks inside Python string literals. There were unintended escapes (e.g., `"\{"`) that were detected by pyupgrade and converted into escaping the escape (e.g., `"\\{"`). But I preferred to change them to use raw strings (e.g., `r"\{"`), especially given that there were some unintended wrong escapes going on (e.g., `"\'"` will be converted when printing to `"'"`) and it's a nicer solution.

bryant1410:
I also ran:

```bash
dos2unix $(find . -name "*.py" -type f)
```

KennethEnevoldsen:
Thanks @bryant1410! Very much appreciate this PR. Since we already used ruff and would like to integrate it in the CI I have made a PR #1136 and have change the target of this PR to that one.

KennethEnevoldsen (commented):
It seems like the main change now is this r" string. I know that some code depends on the bibtex, which isn't actively included in the test. So I would probably prefer not to include this change.

Thus I will probably close this PR, but I have added you as a co-author to the other PR. 

Again thanks for the suggestion!

bryant1410:
Great!

> It seems like the main change now is this r" string. I know that some code depends on the bibtex, which isn't actively included in the test. So I would probably prefer not to include this change.

Consider that I found some bugs in the BibTeX entries. I feel like the best fix is to actually use raw string literals for these entries.

PR #1137, opened 2024-08-07
Added pyupgrade (using ruff) to lint CI

KennethEnevoldsen:
Make sure that we use the latest available syntax when possible. Also added a ruleset for simplifying comprehensions.

Based on #1134, but using ruff

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,7 +87,7 @@ exclude = ["tests", "results"]
 target-version = "py38"
 
 [tool.ruff.lint]
-select = ["F", "I", "E", "D",  "UP"]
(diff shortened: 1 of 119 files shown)


pyproject.toml
Samoed, line 1: Maybe add to make shure that everywhere annotations
```toml
[tool.ruff.lint.isort]
known-first-party = ["mteb"]
required-imports = ["from __future__ import annotations"]
```

KennethEnevoldsen: Completely agree - have added it!

isaac-chung (approved):
Looks good!

PR #1144, opened 2024-08-11
feat: Cleaned up scripts folder 

KennethEnevoldsen:
- A few of the benchmarks weren't implemented in the benchmark list. The best choice seem to move them benchmarks and remove the script (there were already a few issues). To ensure that they are maintained with the rest. Almost all of these had a task that had been renamed. The new approach will raise an error if this is the case.
- Removed trivial (e.g., running one task) or outdated examples. Removing them ensure that we have less to maintain.
- moved the mteb-specific (paper) script into a scirpts/mteb folder.

**Note:** This caused a change in the order of imports, which turns out that (due to our shady naming practices) causes:

```
from mteb.abstasks import TaskMetadata
```

to import the Module `TaskMetadata` instead of the object. I have instead specified all places to avoid future issues:

```
from mteb.abstasks.TaskMetadata import TaskMetadata
```

I have moved this fix to the associated PR: #1145  to clarify it.

## Checklist


Muennighoff (approved):
Nice! Rather than breaking it up into mteb & mmteb, I would probably just break it up into `scripts/data`, `scripts/running` and maybe `scripts/analysis` and put single files directly into `scripts` rather than having folders with just one file 🤔

But I think the current way is also fine if you prefer!

isaac-chung (approved):
I agree with Niklas' suggestion as well, but the current way is fine too.

KennethEnevoldsen:
That is perfectly fine with me as well. I have changed it to match

PR #1148, opened 2024-08-11
fix: Remove unused tests

KennethEnevoldsen:
These tests are no longer used since results moved to the results repo.

## Checklist


isaac-chung (approved):
LGTM. Is it worth moving some of these to the results repo as well?

KennethEnevoldsen:
I have added a test recently that checks that the results repo can be loaded using `mteb.load_results()`. It is a decent test, but we could probably add these over there as well.

PR #1153, opened 2024-08-12
fix: removed test set for AFQMC with no gold labels

KennethEnevoldsen:
The test set of AFQMC is hidden so has removed it from the default suite. This influences the C-MTEB. Who is the best person to contact for this? (@Muennighoff)

## Checklist

### DIFF
--- a/mteb/tasks/STS/zho/CMTEBSTS.py
+++ b/mteb/tasks/STS/zho/CMTEBSTS.py
@@ -238,7 +238,7 @@ class AFQMC(AbsTaskSTS):
         type="STS",
         category="s2s",
         modalities=["text"],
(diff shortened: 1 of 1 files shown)


Muennighoff:
Yeah it's supposed to only use the validation split: https://github.com/embeddings-benchmark/results/blob/4a19a3001ad42c62c777087c33091a0b555addc6/results.py#L34

It seems good to remove the test split. We should also in the mteb meta make sure that the validation split is taken I think? I'm afraid some prior submissions may have accidentally used the test split in their metadata :/

KennethEnevoldsen:
So it looks like some might just have included both. It sounds like we can just merge this in and be good to go?

PR #1157, opened 2024-08-13
fix: Ensure that summarization only calculates the metric across the statistic.

### LINKED ISSUE #1156: SummarizationEvaluator  mean score issue
Code: https://github.com/embeddings-benchmark/mteb/blob/3a3f9cf9688a6dff860788dc5fa4bf9942b8b512/mteb/evaluation/evaluators/SummarizationEvaluator.py#L160

I have noticed that the final score of SummaryEval is calculated from `np.mean` of `spearman_scores`.

However, it seems that the items in `spearman_scores` are SignificanceResult objects, which contain both **statistic** and **pvalue**.
An example of `spearman_scores` is as follows:
```
[SignificanceResult(statistic=0.36804249483125123, pvalue=0.16074763326319716), SignificanceResult(statistic=0.26212077736283135, pvalue=0.3267287471324084), ...]
```

When applying  `np.mean`, it calculate the mean score of all the **statistic** and **pvalue**， instead of mean of**statistic** .

I don't know if there is some misunderstanding of the code or the evaluation method. Please confirm and check this issue.

KennethEnevoldsen:
Fixes #1156

## Checklist

### DIFF
--- a/mteb/abstasks/AbsTaskSummarization.py
+++ b/mteb/abstasks/AbsTaskSummarization.py
@@ -24,6 +24,8 @@ class AbsTaskSummarization(AbsTask):
         relevance: list[float] (the score of the machine generated summaries)
     """
 
+    evalutor = SummarizationEvaluator
+
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
@@ -42,7 +44,7 @@ class AbsTaskSummarization(AbsTask):
             (np.array(x) - self.min_score) / (self.max_score - self.min_score)
             for x in data_split["relevance"]
         ]
-        evaluator = SummarizationEvaluator(
+        evaluator = self.evalutor(
             machine_summaries=data_split["machine_summaries"],
             human_summaries=data_split["human_summaries"],
             texts=data_split["text"],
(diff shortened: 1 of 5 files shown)


Muennighoff (approved):
Wow great bug fix.

Muennighoff:
Can we run it for 1-2 models to see how the score changes?

KennethEnevoldsen:
```

import mteb

tasks = mteb.get_tasks(tasks=["SummEval"])
tasks[0].load_data()

model = mteb.get_model("sentence-transformers/all-MiniLM-L6-v2")
eval = mteb.MTEB(tasks=tasks)
res = eval.run(model, overwrite_results=True)
# expected score: 30.81
res[0].get_score() # 0.259711030822813
```

and for:
-  `intfloat/multilingual-e5-small` we get: 0.306
- `intfloat/multilingual-e5-base` we get: 0.271 
- `intfloat/multilingual-e5-large` we get: 0.31
- `intfloat/e5-small-v2` we get: 0.326
- `intfloat/e5-large-v2` 0.3453

This seems much better at discriminating between models.

I think, however, we should wrap this in a summEval v2 to avoid people confusing scores within model cards with the newer scores. This will mean that we deliberately keep code with a bug, but we can use the "superseeded_by" to refer people to the newest version of the dataset.

Muennighoff:
> I think, however, we should wrap this in a summEval v2 to avoid people confusing scores within model cards with the newer scores. This will mean that we deliberately keep code with a bug, but we can use the "superseeded_by" to refer people to the newest version of the dataset.

Sounds good!

KennethEnevoldsen:
I have added the changed versioning here (@imenelydiaker letting you know about this as well).

PR #1166, opened 2024-08-16
fix: Added task selection for the EU benchmark

KennethEnevoldsen:
Added task for the EU benchmark for MMTEB. 

- The approach is outlines in the notebook, but mainly follows the approach:
  - select EU tasks
  - remove irrelevant tasks (no license, low quality, code, african etc.)
  - Apply iterative task selection removing the most predictable task (under some conditions, see notebook)
  - Based on the selected task refine the filtering until satisfied

This still needs some review, but generally I think it is close to being finalized. Let me know what you think, I would especially like your opinion on the datasets for languages that you speak.

Note: many of the changes are updates to metadata. This is simply to determine the quality of some of the datasets.


isaac-chung (commented):
Looks good! A small suggestion. I think we're good on the question I had separately.

Minor: I wasn't able to add a comment in the notebook. Typo: `Shengen` -> `Schengen`

mteb/tasks/STS/por/SickBrSTS.py
```diff
@@ -25,7 +25,7 @@ class SickBrSTS(AbsTaskSTS):
         date=("2018-01-01", "2018-09-01"),  # rough estimate
         domains=["Web", "Written"],
         task_subtypes=["Textual Entailment"],
-        license="unknown",
+        license="Unknown",
```
x-tabdeveloping, line 28: What is the difference between "Unknown" and " Not Specified"? Is there a meaningful distinction here?

x-tabdeveloping (approved):
Looks quite reasonable overall.
Since @KennethEnevoldsen left comments on how good the tasks are for the languages he speaks, I though I'd also chime in an say, that it's a bit worrying that no Hungarian tasks are left from an actual authentic Hungarian corpus, even though I'd say that it's very different from most other European languages, especially when it comes to fiction.

x-tabdeveloping:
Also, I don't know if this was a conscious decision but we're missing Romani in the language list. I would personally like to see it, since it's spoken by a large enough minority in Europe.

mteb/tasks/STS/por/SickBrSTS.py
KennethEnevoldsen: No we should standardize them to "Not specified"

KennethEnevoldsen:
> Also, I don't know if this was a conscious decision but we're missing Romani in the language list. I would personally like to see it, since it's spoken by a large enough minority in Europe.

Def. an oversight. Will add it tomorrow

Muennighoff (approved):
Looks great! if you want me to run any of the datasets under `# was added after models were run`, lmk 👍

For the French tasks it'd probably be best to get input from @imenelydiaker if needed

KennethEnevoldsen:
> if you want me to run any of the datasets under # was added after models were run, lmk

let us wait until we have the downsampled retrieval datasets as well from @vaibhavad and @orionw, then we can run the all in one go.

KennethEnevoldsen:
I will set this PR to merge. I will message @imenelydiaker if she can take a look and if any additional concerns are raised then we will naturally redo it it again

PR #1120, opened 2024-07-30
[MIEB] Add jina clip

isaac-chung:
Add Jina AI CLIP v1 model

## Checklist

- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 

### Adding a model checklist

 - [x] I have filled out the ModelMeta object to the extent possible
 - [x] I have ensured that my model can be loaded using
   - [x] `mteb.get_model(model_name, revision_id)` and
   - [x] `mteb.get_model_meta(model_name, revision_id)`
 - [x] I have tested the implementation works on a representative set of tasks.

PR #1051, opened 2024-07-05
Review: Outline of MIEB

KennethEnevoldsen:
This is just a review PR. To let me comment on the changes in MIEB

PR #1167, opened 2024-08-16
Add function for metadata create

Samoed:
## Checklist

I've tested the metadata for each task in both normal and multilingual versions (where available), except for multilingual reranking. `WikipediaRerankingMultilingual` is a bit odd because it contains a list of strings instead of a single string for each row (so it probably won't run at all). Additionally, I couldn't download `MIRACLReranking` due to an error with downloading their dataset.


mteb/abstasks/AbsTaskBitextMining.py
```diff
@@ -100,3 +104,28 @@ def _evaluate_subset(
...
+            sentence2 = self.dataset[split][sent_2]
+        total_s1_len = sum([len(s1) for s1 in sentence1])
+        total_s2_len = sum([len(s2) for s2 in sentence2])
+
+        return {
+            "average_sentence1_length": total_s1_len / len(sentence1),
+            "average_sentence2_length": total_s2_len / len(sentence2),
+            "num_sentence1": len(sentence1),
+            "num_sentence2": len(sentence2),
+        }
```
KennethEnevoldsen, line 131: I would love some kind of documentation on the values in the metadata):

A solution might be to create an object:

```
class BitextDescriptiveStatistics(DescriptiveStatistics):
  """
  add docs here
  """
  ...
  
# we can keep it as a dict using:
class DescriptiveStatistics(TypedDict):
   ...
```

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -190,6 +190,30 @@ def _evaluate_subset(
...
+                total_labels.extend(label)
+            else:
+                total_labels.append(label)
+
+        return {
+            "num_texts": len(sentences),
+            "num_labels": len(labels),
+            "average_text_length": total_text_len / len(sentences),
+            "average_label_count": len(total_labels) / len(labels),
+            "unique_labels": len(set(total_labels)),
```
KennethEnevoldsen, line 214: I would actually really like the labels in this dict as well, but that of course mean that we can introduce quite a long list (and labels are often just 0/1) so for now we should probably just ignore it.

mteb/abstasks/MultilingualTask.py
```diff
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+from abc import ABC
+
 from .AbsTask import AbsTask
 from .MultiSubsetLoader import MultiSubsetLoader
 
 
-class MultilingualTask(MultiSubsetLoader, AbsTask):
+class MultilingualTask(MultiSubsetLoader, AbsTask, ABC):
```
KennethEnevoldsen, line 9: Not sure why this is here?

mteb/tasks/Classification/multilingual/LanguageClassification.py
```diff
@@ -66,7 +66,58 @@ class LanguageClassification(AbsTaskClassification):
...
         descriptive_stats={
             "n_samples": {"test": 2048},
-            "avg_character_length": {"test": 107.8},
+            "avg_character_length": {
+                "test": {
+                    "num_texts": 2048,
+                    "num_labels": 2048,
+                    "average_text_length": 109.546875,
+                    "num_label_17": 102,
+                    "num_label_0": 102,
```
KennethEnevoldsen, line 75: Not entirely sure what is going on here? A documentation on the field might help (see above)

mteb/tasks/Clustering/eng/ArxivClusteringS2S.py
```diff
@@ -38,6 +38,14 @@ class ArxivClusteringS2S(AbsTaskClustering):
 }""",
         descriptive_stats={
             "n_samples": {"test": 732723},
-            "avg_character_length": {"test": 74},
+            "avg_character_length": {
+                "test": {
+                    "average_label_count": 23636.225806451614,
+                    "average_text_length": 23636.225806451614,
```
KennethEnevoldsen, line 44: This seems like a bug?

KennethEnevoldsen:
Thanks for the PR @Samoed! It is looking very promising

mteb/abstasks/MultilingualTask.py
Samoed: `MultilingualTask` is only used as a parent class for multilingual classes, so I thought it could be an ABC class to avoid issues with abstract methods. However, if it doesn't fit, I can remove it.

KennethEnevoldsen: Ahh right, so no functional change? In that case I would probably remove it.

mteb/tasks/Classification/multilingual/LanguageClassification.py
Samoed: num_label_n is count how many times label occurred

KennethEnevoldsen:
Edit: We should probably add a test that ensure that future Abstasks have the correct methods.

mteb/tasks/Classification/multilingual/LanguageClassification.py
KennethEnevoldsen: hmm but that will always be the same as num_texts, we should instead use "number of samples". 

I also think it might be worth writing some of these out (no reason to write it like variable names in string).

mteb/abstasks/AbsTaskBitextMining.py
Samoed: Very good suggestion, but what should be done with "generated" fields, like label count, etc.?

mteb/tasks/Clustering/eng/ArXivHierarchicalClustering.py
```diff
@@ -40,7 +40,15 @@ class ArXivHierarchicalClusteringP2P(AbsTaskClusteringFast):
         bibtex_citation="@misc{arXiv.org e-Print archive, url={https://arxiv.org/} }",
         descriptive_stats={
             "n_samples": {"test": N_SAMPLES},
-            "avg_character_length": {"test": 1009.98},
+            "avg_character_length": {
+                "test": {
+                    "num_texts": 2048,
+                    "num_labels": 2048,
```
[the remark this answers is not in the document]
Samoed: Yeah, num_labels is a bit meaningless since it just counts the total number of label rows. I also don't think the number of levels is useful because not all clustering tasks are hierarchical.

mteb/tasks/Classification/multilingual/LanguageClassification.py
Samoed: No, this refers to how often each label appears.

KennethEnevoldsen: A so num_label_n is how many times label N occurred. It might be better as a nested dict. "label frequency" {"1": 102, ...}

Samoed:
Maybe we should rename somehow `avg_character_length`  in `descriptive_stats` or add new field for dataset data description?

mteb/tasks/Clustering/eng/ArXivHierarchicalClustering.py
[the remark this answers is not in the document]
Samoed: But [mteb/arxiv-clustering-p2p](https://huggingface.co/datasets/mteb/arxiv-clustering-p2p) only has a list of labels, so I'm not sure how to determine if the dataset is hierarchical or not

KennethEnevoldsen:
> Maybe we should rename somehow avg_character_length in descriptive_stats or add new field for dataset data description?

I am not sure what you are getting at here?

mteb/abstasks/AbsTaskBitextMining.py
KennethEnevoldsen: I am sure I see the problem. Can you clarify?

mteb/tasks/Clustering/eng/ArXivHierarchicalClustering.py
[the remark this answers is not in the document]
KennethEnevoldsen: I believe the labels are a list:

{"text": "sample text", "label": [1, 2, 5]} # e.g. news, sports, football
{"text": "sample text", "label": [1, 2, 3]} # news sports basket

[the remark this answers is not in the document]
Samoed: Yes, but number of levels should be for all datasets?

Samoed:
Now all "dataset metadata" stored like
```python
        descriptive_stats={
            "n_samples": {"test": 1000},
            "avg_character_length": {
                "test": {
                    ...
                }
            },
```
but name `avg_character_length` not represents what it contains. So maybe it should be renamed or added new field for TaskMetadata?

KennethEnevoldsen:
> but name avg_character_length not represents what it contains. So maybe it should be renamed or added new field for TaskMetadata?

A right, I believe descriptive_stats should contain all computationally derived metrics. Of course the name of the keys should match the content thus I would rename "avg_character_length" if that is not what it contains (what does it then contain?)

mteb/tasks/Clustering/eng/ArXivHierarchicalClustering.py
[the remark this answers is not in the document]
KennethEnevoldsen: No just for clustering and potentially also multilevelclassification

Samoed:
Maybe rename to dataset_statistics or data_info (data_stat)

KennethEnevoldsen:
> Maybe rename to dataset_statistics or data_info (data_stat)

What are you suggesting to rename?

`descriptive_stats`: Let us keep it as is for now (to avoid too many merge conflict with the ongoing MIEB (e.g. #1091 , #1051 , #1120 etc.)
`avg_character_length`: This should not be renamed to `dataset_statistics` as the value is the average character length of the text (for each split in the dict).

Samoed:
Ok, I won't change the name

mteb/abstasks/AbsTask.py
```diff
@@ -51,6 +52,15 @@ def _multilabel_subsampling(
     return dataset_dict
 
 
+class AbsDescriptiveStatistics(TypedDict):
+    """Abstract class for descriptive statistics.
+
+    num_samples: number of samples in the dataset.
+    """
+
+    num_samples: int
```
KennethEnevoldsen, line 61: Would just do:

```suggestion
class DescriptiveStatistics(TypedDict):
    """Class for descriptive statistics.

    num_samples: number of samples in the dataset.
    """

    num_samples: int
```

Unsure if we even need num_samples specified here, but it all cases use it then it is fine.

mteb/abstasks/AbsTaskBitextMining.py
```diff
@@ -100,3 +115,29 @@ def _evaluate_subset(
...
+            sentence1 = self.dataset[split][sent_1]
+            sentence2 = self.dataset[split][sent_2]
+        total_s1_len = sum([len(s1) for s1 in sentence1])
+        total_s2_len = sum([len(s2) for s2 in sentence2])
+
+        return {
+            "average_sentence1_length": total_s1_len / len(sentence1),
+            "average_sentence2_length": total_s2_len / len(sentence2),
+            "num_samples": len(sentence1),
+        }
```
KennethEnevoldsen, line 143: Should return BitextDescriptiveStatistics(...)

mteb/abstasks/AbsTaskClassification.py
```diff
@@ -14,11 +15,24 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+class ClassificationDescriptiveStatistics(AbsDescriptiveStatistics):
+    """Descriptive statistics for Classification
+
+    average_text_length: Average length of text
+    unique_labels: Number of unique labels
+    labels: dict of label frequencies
```
KennethEnevoldsen, line 28:
```suggestion
    Attributes:
      average_text_length: Average length of text
      unique_labels: Number of unique labels
      labels: dict of label frequencies
```

General correction

mteb/abstasks/AbsTaskClassification.py
```diff
@@ -188,3 +195,51 @@ def _undersample_data(self, X, y, samples_per_label: int, idxs=None):
                 y_sampled.append(y[i])
                 label_counter[y[i]] += 1
         return X_sampled, y_sampled, idxs
+
+    def calculate_metadata_metrics(
+        self,
+    ) -> dict[str, AbsDescriptiveStatistics | dict[str, AbsDescriptiveStatistics]]:
```
KennethEnevoldsen, line 201:
```suggestion
    ) -> dict[str, ClassificationDescriptiveStatistics | dict[str, ClassificationDescriptiveStatistics]]:
```

mteb/abstasks/AbsTaskBitextMining.py
KennethEnevoldsen: seems to be a general correction

mteb/abstasks/AbsTaskInstructionRetrieval.py
```diff
@@ -580,56 +602,59 @@ def create_qrel_diff(self, og_qrels, changed_qrels):
...
+        )
+        qrels_per_doc = num_qrels_non_zero / len(relevant_docs) if len(queries) else 0
+        top_ranked_per_query = (
+            sum(len(docs) for docs in top_ranked.values()) / len(queries)
+            if len(queries)
+            else 0
+        )
+        return {
+            "num_docs": len(corpus),
+            "num_samples": len(queries),
```
KennethEnevoldsen, line 645: Would probably rather have num_queries (which means I would move num_samples out of general dict)

mteb/tasks/BitextMining/dan/BornholmskBitextMining.py
```diff
@@ -43,7 +43,13 @@ class BornholmBitextMining(AbsTaskBitextMining):
...
         descriptive_stats={
             "n_samples": {"test": 500},
-            "avg_character_length": {"test": 89.7},
+            "avg_character_length": {
+                "test": {
+                    "average_sentence1_length": 49.834,
+                    "average_sentence2_length": 38.888,
+                    "num_samples": 500,
+                }
+            },
```
KennethEnevoldsen, line 52: I see what you meant with `avg_character_length`. Sorry about the confusion:

Would probably have it either as Dict[Key, Dict[Split, Value]]:
```
            "n_samples": {"test": 500},
             "average_sentence1_length": {"test": 49.834}...
```

Or Dict[Split, Dict[Key, Value]]:

```
                "test": {
                     "average_sentence1_length": 49.834,
                     "average_sentence2_length": 38.888,
                     "num_samples": 500,
                 }
```
Either is fine by me. The first is closer to the current implementation.

mteb/tasks/STS/multilingual/STS17CrosslingualSTS.py
```diff
@@ -67,7 +67,76 @@ class STS17Crosslingual(AbsTaskSTS, MultilingualTask):
 }""",
         descriptive_stats={
             "n_samples": {"test": 500},
-            "avg_character_length": {"test": 43.3},
+            "avg_character_length": {
+                "test": {
+                    "ko-ko": {
```
KennethEnevoldsen, line 72: This does become quite nested. Might be nice to have the overall metrics as well:

{"test": {"n_samples": ... "average_sentence1_length": .., "hf_subset_descriptive_stats": {"ar-ar": {...}}}

Just to keep the easy overview.

KennethEnevoldsen: This would also make multilingual datasets have the same structure as non-multilingual datasets (which we def. want)

KennethEnevoldsen (commented):
Already looking a lot better. I have added a few comments to make sure we get it right.

mteb/abstasks/AbsTaskBitextMining.py
Samoed: TypedDict does not support init like this. I can make like
```
TypedDict('BitextDescriptiveStatistics', ...)
```
or
```
result: BitextDescriptiveStatistics = ...
```
Maybe we should use dataclass instead of TypedDict?

mteb/tasks/Clustering/eng/ArxivClusteringS2S.py
Samoed: It's not working correctly, because it contains a list of lists for sentences and labels. I will remove metadata update

mteb/tasks/BitextMining/dan/BornholmskBitextMining.py
Samoed: I think the second implementation is better because it allows all metrics for a split to be easily viewed, and the code is clearer

mteb/abstasks/AbsTaskBitextMining.py
KennethEnevoldsen: A data class works just fine as well. 

Though TypedDict does support init:

```
from typing import TypedDict

class Movie(TypedDict):
    name: str

t = Movie(name='The Matrix') # runs without issue
```

Samoed: Hm. Maybe I missed that when reading documentation

mteb/abstasks/AbsTask.py
```diff
@@ -216,7 +216,9 @@ def calculate_metadata_metrics(
                     pbar_lang.set_postfix_str(f"Language: {lang}")
                     print(f"Processing metadata for language {lang}")
                     split_details = self._calculate_metrics_from_split(split, lang)
-                    all_details[split][lang] = split_details
+                    all_details[split]["hf_subset_descriptive_stats"][lang] = (
```
KennethEnevoldsen, line 219:
```suggestion
                    all_details[split]["hf_subset_descriptive_stats"][hf_subset] = (
```
As it is not nec. language

mteb/tasks/BitextMining/multilingual/IN22ConvBitextMining.py
```diff
@@ -101,8 +101,11 @@ class IN22ConvBitextMining(AbsTaskBitextMining, MultilingualTask):
 note={}
 }""",
         descriptive_stats={
-            "avg_character_length": {
-                "test": {
+            "test": {
+                "average_sentence1_length": 54.32948595562498,
+                "average_sentence2_length": 54.32948595562498,
```
KennethEnevoldsen, line 106: This is exactly the same?

mteb/tasks/BitextMining/multilingual/IN22ConvBitextMining.py
```diff
@@ -101,8 +101,11 @@ class IN22ConvBitextMining(AbsTaskBitextMining, MultilingualTask):
 note={}
 }""",
         descriptive_stats={
-            "avg_character_length": {
-                "test": {
+            "test": {
```
KennethEnevoldsen, line 104: These looks very good!!

KennethEnevoldsen (commented):
These changes looks really good! Thanks for taking the time. We need to update: 

https://github.com/embeddings-benchmark/mteb/blob/main/docs/create_tasks_table.py

and:
https://github.com/embeddings-benchmark/mteb/blob/main/docs/tasks.md

As well. I think the changes are fairly minor though (it seems like we only need to change the header of the table) 

Once those are fixed I think we are good to go

mteb/tasks/BitextMining/multilingual/IN22ConvBitextMining.py
Samoed: This is a parallel dataset, so first one language counts as sententece1 and other time as sentence2. At total there all texts in both "sentences".

Samoed:
I've updated the task tables

KennethEnevoldsen:
Wonderful will merge this in then - Thanks for the addition!

Issue #361, opened 2024-04-13
Add a European Benchmark


KennethEnevoldsen:
Add a benchmark for the European languages. This issue gives an overview of the status.

The EU has 24 official languages:

- [ ] Bulgarian - bul
- [ ] Croatian - hrv
- [ ] Czech - ces
- [x] Danish - dan
- [ ] Dutch - nld
- [x] English - eng
- [ ] Estonian - est
- [x] Finnish - fin (covered with only 1 task outside of bitext and translations)
- [x] French - fra
- [x] German - deu
- [ ] Greek - ell
- [ ] Hungarian - hun
- [ ] Irish - gle
- [x] Italian - ita
- [ ] Latvian - lav
- [ ] Lithuanian - lit
- [ ] Maltese - mlt
- [x] Polish - pol
- [x] Portuguese - por
- [ ] Romanian - ron
- [ ] Slovak - slk
- [ ] Slovenian - slv
- [x] Spanish - spa
- [x] Swedish - swe

Additionally to this list we might add (feel free to add languages that I might have missed):

- [ ] Basque (although it is not an official EU language, it is officially recognized within Spain)
- [x] Norwegian Nynorsk (official language of Norway, which is in the Schengen Area but not in the EU)
- [x] Norwegian Bokmål (official language of Norway, which is in the Schengen Area but not in the EU)
- [ ] Icelandic (official language of Iceland, also in the Schengen Area but not in the EU)
- [ ] Albanian (official language of Albania, a candidate for EU membership and part of the Schengen visa regime)
- [ ] Serbian (official language of Serbia, another candidate for EU membership and part of Schengen visa policies)
- [ ] Macedonian (official language of North Macedonia, an EU candidate country and part of the Schengen visa system)

_Note I haven't checked off languages only covered in bitext tasks or in translated tasks_

x-tabdeveloping:
We might get reasonable coverage of Macedonian by including Bulgarian, they are as close as Bokmål and Danish pretty much. Same thing goes for Serbian and Croatian (maybe also Slovenian) up until to nineties Serbo-Croatian was considered done language. Perhaps the reason to include both would be because Serbian is written with cyrillic script (but then we can make sure the model understands that by having Bulgarian, Russian or Ukrainian)

x-tabdeveloping:
Also might consider some minority languages and regional dialects for fairness's and ethics's sake, a handful of examples I can think of:
[...]

PR #1168, opened 2024-08-19
Added contruction of MTEB(eng, v2) and its lite version

KennethEnevoldsen:
Added task for the EU benchmark for MMTEB. 

- The approach is outlines in the notebook, but mainly follows the approach:
  - select EU tasks
  - remove irrelevant tasks (no license, low quality, code, african etc.)
  - Apply iterative task selection removing the most predictable task (under some conditions, see notebook)
  - Based on the selected task refine the filtering until satisfied

This still needs some review, but generally I think it is close to being finalized. Let me know what you think, I would especially like your opinion on the datasets for languages that you speak.

Note: many of the changes are updates to metadata. This is simply to determine the quality of some of the datasets.

(see also #361 )


Muennighoff (commented):
Nice though I think maybe for doing a new core english benchmark  i.e. mtebv2, we should rather start out with all english benchmarks not the original 56? E.g. something like

1. Look at all English tasks in the codebase
2. Filter for some measure of quality i.e. we only want the best
3. Filter for some measure of diversity (similar to what you did here I think)
4. Filter for some measure of redundancy (similar to what you did here I think)
5. Make them not too expensive to run / have a lite version

Muennighoff:
Curious also what @orionw thinks!

orionw:
I like the approach in the notebook, looks good overall.  I also like that it has the human influence and isn't entirely model driven, I think we have some extra knowledge about quality and such that is hard to fit into the model.

Agree with both of you that for a v2 version we should probably select and then prune back via some of these methods (which is somewhat similar to what we're doing now).

KennethEnevoldsen:
I have now reformulated it to be MTEB lite only - I will set this to merge (to avoid conflict), but I knowing that we still have tasks to add.

PR #1173, opened 2024-08-21
add multilingual mocks and test descriptive_stats

Samoed:
## Checklist

### DIFF
--- a/tests/test_benchmark/task_grid.py
+++ b/tests/test_benchmark/task_grid.py
@@ -80,7 +80,7 @@ MOCK_TASK_TEST_GRID = [
     MockMultilingualClusteringFastTask(),
     MockPairClassificationTask(),
     MockMultilingualPairClassificationTask(),
-    MockRerankingTask(),  # 11
+    MockRerankingTask(),
     MockMultilingualRerankingTask(),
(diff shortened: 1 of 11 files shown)


tests/test_benchmark/task_grid.py
```diff
@@ -55,19 +67,31 @@
...
     MockRetrievalTask(),
+    MockMultilingualRetrievalTask(),
     MockSTSTask(),
+    MockMultilingualSTSTask(),
     MockMultilabelClassification(),
+    MockMultilingualMultilabelClassification(),
     MockSummarizationTask(),
+    MockMultilingualSummarizationTask(),
     MockInstructionRetrival(),
+    MockMultilingualInstructionRetrival(),
```
KennethEnevoldsen, line 94: How does this influence runtime of the tests?

KennethEnevoldsen (commented):
slightly concerned that adding the new cases increases runtime of existing tests too much. Otherwise this looks reasonable.

tests/test_benchmark/task_grid.py
Samoed: I looked through CI runs and time nearly same. Locally all tasks run ~20 sec

PR #1174, opened 2024-08-21
fix: Added multilingual and indic benchmark

KennethEnevoldsen:
This adds outlines for the multilingual and indic benchmarks as well as a few adjustments to existing benchmarks due to a few discovered bugs in the previous implementation.

It also add utilites for easily creating a table from a list of tasks:

```
tasks = mteb.get_tasks(tasks=[...])
tasks.to_latex()
tasks.to_dataframe()
```

## Checklist


KennethEnevoldsen:
Merging this in for now -  we will do a final review on the benchmark at the end

docs/mmteb/create_points_table.py
```diff
@@ -6,7 +6,7 @@
 import pandas as pd
 
 
-def load_data(file_path: Path) -> pd.DataFrame:
```
isaac-chung, line 9: This causes the test to fail: https://github.com/embeddings-benchmark/mteb/actions/runs/10652807489/job/29527177315

PR #1179, opened 2024-08-25
refactor: create_meta

Samoed:
## Checklist


KennethEnevoldsen (approved):
so you have moved it to a seperate file and made minor changes to the test. I think this looks very reasonable. It might be worth creating a cli folder and move it all there and split it each into their own files. However, this is still a good so will just merge it in

PR #1181, opened 2024-08-25
add citation rumteb

Samoed:
## Checklist

### DIFF
--- a/mteb/benchmarks.py
+++ b/mteb/benchmarks.py
@@ -156,7 +156,16 @@ MTEB_MAIN_RU = Benchmark(
     ),
     description="Main Russian benchmarks from MTEB",
     reference="https://aclanthology.org/2023.eacl-main.148/",
-    citation=None,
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen (approved):
Ahh saw the publication as well, congatulations! Will you also add the reference within MMTEB as well (down in the appendix)

Also glad to see that your results are in line with the ones on the Scandinavian embedding benchmark as well as the ones within MMTEB, though with the mistral-based e5 performing better in your case. 

Also I see that you incorrectly highlighted BGE-M3 as the best retrieval model in table 5 (while is the XLM-r based e5 instruct)

Samoed:
I've added a citation to the appendix. I'm not the author (and didn't participate in these publications), but I'll contact the author because they have different results for `bge-m3` than what's shown on the leaderboard.

Samoed:
They included new results only in their paper and haven't published them anywhere else. If they don't update it, I'll run them later

Issue #942, opened 2024-06-17
Avoid using global seeds


KennethEnevoldsen:
Currently, we use a global seed (in fact, multiple, which might invalidate each other). This is problematic as computation happening outside the software might also set a global seed, which will then invalidate benchmark results). A solution instead is to define a random_state or pass around an RNG object.

see e.g. https://docs.astral.sh/ruff/rules/numpy-legacy-random/

This is not a pressing issue; the current performance metric aligns with the third decimal (as far as I am aware and can test)

KennethEnevoldsen:
A documented case where this causes problems: https://github.com/embeddings-benchmark/results/pull/19#discussion_r1741573217

ayush1298:
@KennethEnevoldsen I see that https://github.com/embeddings-benchmark/mteb/pull/1193 is already merged in v2 branch. So, can we closed this issue?

KennethEnevoldsen:
Just added NPY to v2 so will close this

PR #1193, opened 2024-09-03
fix: Ensure seed is based on RNG State

KennethEnevoldsen:
The global seed (see also #942) cause variance in performance estimation depending on the order in which tasks are run, we see an example here:

https://github.com/embeddings-benchmark/results/pull/19#discussion_r1741573217

In #942 I propose a more general fix to this issue, but this PR as well as using a consistent seed will naturally causes slightly different scores as RNG states changes. 

We might want to avoid merging this change until we change the major version (and then do #942 fully).

## Checklist

### DIFF
--- a/mteb/abstasks/AbsTaskClassification.py
+++ b/mteb/abstasks/AbsTaskClassification.py
@@ -190,7 +190,8 @@ class AbsTaskClassification(AbsTask):
         y_sampled = []
         if idxs is None:
             idxs = np.arange(len(y))
-        np.random.shuffle(idxs)
+        rng_state = np.random.default_rng(self.seed)
+        rng_state.shuffle(idxs)
         label_counter = defaultdict(int)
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen:
@Samoed and @artemsnegirev would love your opinion on this as well

Samoed:
Maybe it's better to use [fix_seed from transformers](https://huggingface.co/docs/transformers/internal/trainer_utils#transformers.set_seed), because it will fix seed for torch with different devices?

KennethEnevoldsen:
> Maybe it's better to use [fix_seed from transformers](https://huggingface.co/docs/transformers/internal/trainer_utils#transformers.set_seed), because it will fix seed for torch with different devices?

But this will still be influenced by the order of operations? (we essentially do this already within the abstasks). It will also be influenced by code-changes.

Issue #1200, opened 2024-09-05
BadRequestError with OpenAI models for some datasets


HSILA:
Cloning the repo, installing it with pip:
```
git clone https://github.com/embeddings-benchmark/mteb.git
cd mteb
python -m venv venv
source ./venv/bin/activate
pip install -e .
```
And running the following script:
```python
import mteb

tasks = mteb.get_tasks(tasks=["SprintDuplicateQuestions"])
model = mteb.get_model("text-embedding-3-small")
evaluation = mteb.MTEB(tasks=tasks)
evaluation.run(model, output_folder="results")
```
Will throw the following error:
```
  File "/home/user/mteb/venv/lib/python3.11/site-packages/openai/_base_client.py", line 1041, in _request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "'$.input' is invalid. Please check the API reference: https://platform.openai.com/docs/api-reference.", 'type': 'invalid_request_error', 'param': None, 'code': None}}
```
The same error is observed with all of the OpenAI embedding models.

In the [OpenAI's Embedding API Reference](https://platform.openai.com/docs/api-reference/embeddings/create), it is mentioned that:

> The input must not exceed the max input tokens for the model (8192 tokens for text-embedding-ada-002), cannot be an empty string, and any array must be 2048 dimensions or less.

I believe the issue arises from the dimensions of the `sentences` list [here](https://github.com/embeddings-benchmark/mteb/blob/16b222019c595aea7cdbbb04b506736851c6316b/mteb/models/openai_models.py#L34C1-L41C10), when it exceeds 2048, an error is thrown.  For `SprintDuplicateQuestions`, the dimension of the `sentences` list is 8932 which will cause the error. But when you replace `SprintDuplicateQuestions` with `LegalBenchPC` in the above code, you no longer see the error, since the `sentences` in this case has only 1770 elements.

KennethEnevoldsen:
The fix is now merged in - will close this PR

PR #1201, opened 2024-09-05
fix OpenAI BadRequestError by limiting input dimensions to 2048 elements

HSILA:
- Ensure the `sentences` list passed to OpenAI API does not exceed 2048 elements, a fix to issue #1200 
- Reference: [OpenAI's Embedding API documentation](https://platform.openai.com/docs/api-reference/embeddings/create) on input limits

## Checklist

`test_mulitple_mteb_tasks` fails, but the same test fails with the current MTEB repository.

### DIFF
--- a/mteb/models/openai_models.py
+++ b/mteb/models/openai_models.py
@@ -31,14 +31,22 @@ class OpenAIWrapper:
                 "Reducing embedding size available only for text-embedding-3-* models"
             )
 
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen:
@HSILA I will take a look at the test and then return back to this PR - thanks for the quick fix

PR #1210, opened 2024-09-09
fix: Normalize licenses including casing, uses of "-" etc.

### LINKED ISSUE #1165: Standardize license in task metadata
Atm. we have multiple license which are equivalent but specified differently "mit" vs "MIT", are practically unknown ("Other (Commercial)"), custom licenses (typically specified as a link or some text), not specified by the dataset ("unknown", "Not specified", "Not Specified",... ) and not specified by us (None).

We can standardize the following:

- Use a predefined list of licenses to standardize spelling variations
- Use links for custom licenses
- Use "Not specified" for cases where the license is unknown (but we have checked)
- Use None in cases where we haven't checked the license

KennethEnevoldsen:
fixes #1165
 

## Checklist


imenelydiaker (approved):
LGTM! 🙂

isaac-chung (approved):
Nice!

PR #1212, opened 2024-09-10
docs: Fix broken links in docs

Pringled:
Added fixes for a couple of broken links in [adding_a_dataset](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md) and [adding_a_model](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_model.md). 

Note: the script run_mteb_chinese.py seems to be gone entirely.

### DIFF
--- a/docs/adding_a_model.md
+++ b/docs/adding_a_model.md
@@ -29,7 +29,7 @@ mteb run -m {model_name} -t {task_names}
 
 These will save the results in a folder called `results/{model_name}/{model_revision}`.
 
-For reference you can also look at [scripts/run_mteb_english.py](https://github.com/embeddings-benchmark/mteb/blob/main/scripts/data/run_mteb_english.py) for all MTEB English datasets used in the main ranking.
+For reference you can also look at [scripts/data/run_mteb_english.py](https://github.com/embeddings-benchmark/mteb/blob/main/scripts/data/run_mteb_english.py) for all MTEB English datasets used in the main ranking.
 Advanced scripts with different models are available in the [mteb/mtebscripts repo](https://github.com/embeddings-benchmark/mtebscripts).
 
 2. **Format the results using the CLI:**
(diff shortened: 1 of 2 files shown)


docs/adding_a_model.md
```diff
@@ -29,7 +29,7 @@ mteb run -m {model_name} -t {task_names}
 
 These will save the results in a folder called `results/{model_name}/{model_revision}`.
 
-For reference you can also look at [scripts/run_mteb_english.py](https://github.com/embeddings-benchmark/mteb/blob/main/scripts/run_mteb_english.py) for all MTEB English datasets used in the main ranking, or [scripts/run_mteb_chinese.py](https://github.com/embeddings-benchmark/mteb/blob/main/scripts/run_mteb_chinese.py) for the Chinese ones. 
+For reference you can also look at [scripts/data/run_mteb_english.py](https://github.com/embeddings-benchmark/mteb/blob/main/scripts/data/run_mteb_english.py) for all MTEB English datasets used in the main ranking.
```
isaac-chung, line 32: We might want to remove this now that we're away from scripts. wdyt @KennethEnevoldsen ?

KennethEnevoldsen: Agree will try to get the ability to get eval_splits added to benchmarks so that everything is specified "test" and then I will remove them

Pringled: Should I remove this section in this PR already? Also, I noticed that some of the tests are failing, but I think it does not have anything to do with this PR.

isaac-chung: > Should I remove this section in this PR already? Also, I noticed that some of the tests are failing, but I think it does not have anything to do with this PR.

@Pringled if you merge the latest main, the failures should go away. It's related to a missing dataset which has been fixed.

KennethEnevoldsen: > Should I remove this section in this PR already?

I would rather wait until we have the final fixes in.

PR #1215, opened 2024-09-11
fix: Ensure that results are returned even when hitting cache

### LINKED ISSUE #1122: Allow Evaluation should return results from cached runs instead of skipping them
The evaluation [do not run tasks twice](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/evaluation/MTEB.py#L334), that is smart as it saves ressources but why is the task deleted ??

I am using the python SDK and spent several minutes wondering why my code didn't work anymore (do not return any result) and that was the culprit.

I'd thought that the package should use already cached results instead of just not returning anything.

Any design justification ? Is it a bug ?

KennethEnevoldsen:
Fixes #1122

## Checklist

### DIFF
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -278,7 +278,7 @@ class MTEB:
         co2_tracker: bool = False,
         encode_kwargs: dict[str, Any] = {},
         **kwargs,
-    ):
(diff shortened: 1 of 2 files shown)


isaac-chung (approved):
Love this change. LGTM.

PR #1217, opened 2024-09-12
fix: Allow benchmark to specify eval_splits

KennethEnevoldsen:
This PR allow for benchmarks to specify eval. splits. This allow us to fully specify a benchmark within the benchmark object.

To do this it add the following:
- added eval_splits to the Abstask object, which default to metadata.eval_splits
- use the task.eval_splits unless overwritten in mteb.MTEB.run
- added eval_splits arg to mteb.get_tasks, which filter the tasks based on splits
- updated documentation
  - renamed the "Advanced Usage" to "Usage Documentation" to make it more accessible
- added tests where relevant

## Checklist


mteb/overview.py
```diff
@@ -288,6 +297,7 @@ def get_task(
         languages: A list of languages either specified as 3 letter languages codes (ISO 639-3, e.g. "eng") or as script languages codes e.g.
             "eng-Latn". For multilingual tasks this will also remove languages that are not in the specified list.
         script: A list of script codes (ISO 15924 codes). If None, all scripts are included. For multilingual tasks this will also remove scripts
+        eval_splits: A list of evaluation splits to include. If None, all splits are included.
```
isaac-chung, line 300: Could we also add a test case for the 2nd sentence, where we assert the number of splits > 1? (If none...)

README.md
```diff
@@ -126,13 +126,28 @@ evaluation = mteb.MTEB(tasks=[
...
+```
+
+The benchmark specified not only a list of tasks, but also what splits and language to run on. To get an overview of all available benhcmarks simply run:
+
+```python
+import mteb
+benchmarks = mteb.get_benchmarks()
+```
+
+generally we use the naming scheme for benchmarks `MTEB(*)`, where the "*" denotes the target of the benchmark. In case of a language we use the three letter language code. For large groups of language we use the group notation, e.g. `MTEB(Scandinavian)` for Scandinavian languages. External benchmarks implemented in MTEB like `CoIR` use their original name. When using a benchmark from MTEB please cite `mteb` along with the citations of the benchmark which you can access using:
```
isaac-chung, line 147:
```suggestion
Generally we use the naming scheme for benchmarks `MTEB(*)`, where the "*" denotes the target of the benchmark. In case of a language, we use the three letter language code. For large groups of language we use the group notation, e.g. `MTEB(Scandinavian)` for Scandinavian languages. External benchmarks implemented in MTEB like `CoIR` use their original name. When using a benchmark from MTEB please cite `mteb` along with the citations of the benchmark which you can access using:
```

isaac-chung (commented):
Nice! Added some suggestions.

mteb/abstasks/AbsTask.py
```diff
@@ -67,6 +67,7 @@ def __init__(self, seed: int = 42, **kwargs: Any):
         self.data_loaded = False
         self.is_multilingual = False
         self.save_suffix = kwargs.get("save_suffix", "")
+        self._eval_splits = None
```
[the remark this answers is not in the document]
KennethEnevoldsen: yep, also moved all of the out of the init function (they can just be specified outside)

mteb/overview.py
KennethEnevoldsen: In that case, it should be: 
- task.eval_splits == task.metadata.eval_splits # can be 1
I have added a test for that.

isaac-chung:
It seems like there's a relevant test failure

PR #1228, opened 2024-09-20
Update points

Muennighoff:
Small fixes! Seems like there is both kranthigv & KranthiGV in the point list.


KennethEnevoldsen (approved):
Lovely - thanks!

PR #1233, opened 2024-09-23
Image text pair cls

gowitheflow-1998:

1. fix ImageTextPairClassification Evaluator in handling large datasets.
2. add the rest of Compositionality Evaluation tasks (AROCocoOrder & AROFlickrOrder).
3. fix meta data to align with `main` for datasets that haven't been fixed.


mteb/tasks/Image/ImageTextPairClassification/AROCocoOrder.py
```diff
@@ -0,0 +1,55 @@
...
+        bibtex_citation="""@article{hsieh2024sugarcrepe,
+  title={Sugarcrepe: Fixing hackable benchmarks for vision-language compositionality},
+  author={Hsieh, Cheng-Yu and Zhang, Jieyu and Ma, Zixian and Kembhavi, Aniruddha and Krishna, Ranjay},
+  journal={Advances in neural information processing systems},
+  volume={36},
+  year={2024}
+}""",
+        descriptive_stats={
+            "n_samples": {"test": 25010},
+            "avg_character_length": {"test": 1},
```
KennethEnevoldsen, line 53: Just a label?

mteb/evaluation/evaluators/Image/ImageTextPairClassificationEvaluator.py
```diff
@@ -54,47 +95,74 @@ def __call__(
...
-
-        image_embeddings = F.normalize(
-            model.get_image_embeddings(images, batch_size=encode_kwargs["batch_size"]),
-            dim=-1,
-        ).view(num_samples, num_images_per_sample, -1)
-        text_embeddings = F.normalize(
-            model.get_text_embeddings(texts, batch_size=encode_kwargs["batch_size"]),
-            dim=-1,
-        ).view(num_samples, num_texts_per_sample, -1)
+        data_loader = DataLoader(
```
KennethEnevoldsen, line 98: I feel like the batching should be a model concern not a concern of the task.

We do this in main using `encode_kwargs`.

KennethEnevoldsen: I would love to hear what others think though

mteb/tasks/Image/ImageTextPairClassification/AROCocoOrder.py
gowitheflow-1998: currently yes, we haven't thought much about what will be useful meta data for image datasets so it is always 1 at the moment for most datasets. This should be on the to do list

mteb/evaluation/evaluators/Image/ImageTextPairClassificationEvaluator.py
gowitheflow-1998: I agree

gowitheflow-1998: The current pipeline is quite aligned with main I think? If we do `results = evaluation.run(model, encode_kwargs={"batch_size": 1024}, output_folder="results"` for this task, it will be able to adjust as well (the batch size has been used in the construction of `ImageTextDataset` here). 

What is confusing might be that for other mieb tasks, we can do something like `model.get_image_embeddings(images, batch_size=encode_kwargs["batch_size"])` but for this task there is a loop over dataloader to get the images and texts out seperately first (image and text are currently both in the same Dataset Object) so it might look like the batch size is not used in the encoding process. I will aim to make it look better in the next commit.

PR #1246, opened 2024-09-27
fix: @mrshu's name in `points.md`

mrshu:
* Use the diacritic character to be inline with Slovak spelling.


mrshu:
I am very sorry for bothering with this @KennethEnevoldsen but I did notice this small discrepancy in #1241 and would appreciate if this small change could be applied. In TeX it should be as easy as replacing `S` with `\v{S}`.

Thanks a bunch!

isaac-chung (approved):
Thanks. I'll make the edits in the file.

isaac-chung:
Whoops. Should have been `docs` instead of `fix`.

Issue #1250, opened 2024-09-28
Add benchmark CLI options


isaac-chung:
As a follow up to #1209, add CLI options for
- [x] listing all available benchmarks
- [x] running benchmarks by name

PR #1261, opened 2024-09-29
fix: Select benchmarks CLI option

isaac-chung:
Address #1250 point 2.
- Added test cases for a list of Benchmark objects passed into the task parameter
- Added command line option to select Benchmarks

The following can now run. 1) script
```python
import mteb

model_name = "average_word_embeddings_komninos"
benchmark = mteb.get_benchmarks(names=["MTEB(eng)"])
evaluation = mteb.MTEB(tasks=benchmark)
results = evaluation.run(model)
```
and command line:
```
mteb run -m average_word_embeddings_komninos -b 'MTEB(eng)'
```

## Checklist

### DIFF
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -6,6 +6,7 @@ import os
 import traceback
 from copy import copy
 from datetime import datetime
+from itertools import chain
 from pathlib import Path
 from time import time
 from typing import Any, Iterable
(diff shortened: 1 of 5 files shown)


mteb/evaluation/MTEB.py
```diff
@@ -58,6 +59,11 @@ def __init__(
 
         if tasks is not None:
             self._tasks = tasks
+            if (
+                not isinstance(tasks[0], AbsTask) and not isinstance(tasks[0], str)
+            ):  # So that passing in a list of Benchmarks works too. Trying to import Benchmark causes circular import.
```
KennethEnevoldsen, line 64: Hmm, I think we should use Benchmark here, why do we see the circular import? (something we should fix?)

Sidenote: https://github.com/embeddings-benchmark/mteb/issues/1272

KennethEnevoldsen (approved):
only a few minor things.

mteb/evaluation/MTEB.py
isaac-chung: Done. And also commented on the tagged issue.

isaac-chung:
Thanks! 🚀 Enabling auto-merge.

Issue #1211, opened 2024-09-09
Figure out an approach for adding experiments to the leaderboard


KennethEnevoldsen:
We currently do not have a good approach to adding experiment runs to the MTEB leaderboard e.g. experiment with the influence of hyperparameters. Such as using no-prompt for prompt-based models or changing the embedding size.

A solution to implement is to run it with a unique model_name, but there is currently no documentation on how one would need to do that.

---

edit: a potentially better solution is to add experiments. Where we add a layer to the visualization:

```
model_name
| - revision1
|    | - task1.json
     | - ...
     | - experiments
     |    | - exp_name1
     |    |    | -  task1.json
     |    |    | - ...
     |    | - exp_name1
     |    |    | -  task1.json
     |    |    | - ...
```

or potentially a slightly more consistent structure:

```
model_name
| - revision1
     | - default
     |    | -  task1.json
     |    | - ...
     | - exp_name1
     |    | -  task1.json
     |    | - ...
```

Experiment names could be e.g. no-instruct/instruct (whether you use instructions) or emb_size=256

To run a model as an experiment we could do:

```
model = MyModel(**custom_kwargs)
results = mteb.MTEB(model, experiment_name = "no-instruct")
```

KennethEnevoldsen:
This way also allows us to filter experiments on the leaderboard.

orionw:
I like this, gives more flexibility and cleans up the main folder.

The only potentially hard part is what do we do say for GritLM but for each type of prompt it runs?  The user would have to identify some title for the prompt to put down (e.g. E5 prompt or Instructor prompt or "prompt-I-just-made-up").  These will then be non-standard since every user will name it differently (or perhaps run different prompts).

Samoed:
I think there could be a file for each experiment that includes the parameters used for the experiment, similar to `model_meta` or additional kwargs for model can be logged

KennethEnevoldsen:
> The only potentially hard part is what do we do say for GritLM but for each type of prompt it runs? The user would have to identify some title for the prompt to put down (e.g. E5 prompt or Instructor prompt or "prompt-I-just-made-up"). These will then be non-standard since every user will name it differently (or perhaps run different prompts).

Def. agree with this. I would imagine that some standards form, but there will be some custom stuff to allow for flexibility.

> I think there could be a file for each experiment that includes the parameters used for the experiment, similar to model_meta. This can be done by logging additional kwargs for model

I think we can just log this in the model_meta?

[...]

PR #1270, opened 2024-10-02
`e5-mistral` evaluation via vLLM

robertgshaw2-neuralmagic:
- Launch server:

```bash
vllm serve intfloat/e5-mistral-7b-instruct
```

- Run eval script:

```bash
python3 run-e5.py
```

### DIFF
--- a/run-e5.py
+++ b/run-e5.py
@@ -1,0 +1,90 @@
+import asyncio
+
+import numpy as np
+from typing import Any
+from openai import AsyncOpenAI
+
+import mteb
+from mteb.model_meta import ModelMeta
+from mteb.models.text_formatting_utils import corpus_to_texts
+from mteb.models.instructions import task_to_instruction
+
+MODEL = "intfloat/e5-mistral-7b-instruct"
+TASKS = ["STS12"]
+
+class AsyncOpenAIWrapper:
+    def __init__(self, 
+                 model_name: str, 
+                 base_url: str = "http://localhost:8000/v1",
+                 num_concurrent: int = 100) -> None:        
+
+        self._client = AsyncOpenAI(base_url=base_url)
+        self._model_name = model_name
+        self._num_concurrent = num_concurrent
+
+    async def async_encode(self, sentences: list[str]) -> np.ndarray:
+        sublists = [
+            sentences[i : i + self._num_concurrent]
+            for i in range(0, len(sentences), self._num_concurrent)
+        ]
+
+        all_embeddings = []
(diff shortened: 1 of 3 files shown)


run-e5.py
```diff
@@ -0,0 +1,90 @@
+import asyncio
+
+import numpy as np
+from typing import Any
+from openai import AsyncOpenAI
```
Samoed, line 5: You should move imports of external library inside the class to make it optional

run-e5.py
```diff
@@ -0,0 +1,90 @@
...
+
+model = AsyncOpenAIWrapper(model_name=MODEL)
+model.mteb_model_meta = ModelMeta(
+    name=MODEL,
+    revision=None,
+    release_date=None,
+    languages=None,
+    similarity_fn_name=None,)
+
+tasks = mteb.get_tasks(tasks=TASKS)
```
Samoed, line 88: When this file will be imported it will try to evaluate model. I think this should be removed

KennethEnevoldsen (commented):
Very happy to see this PR! We already have an implementation of e5 mistral. However, I assume this is notably faster, but it would be nice we an actual comparison of both scores and speed.

We also implement models in `mteb/models`. I believe we could do this here as well. 

I assume it produces different scores (given vLLM) so it might be worth changing the name (related to #1211), but we can take that after a test.

robertgshaw2-neuralmagic:
> Very happy to see this PR! We already have an implementation of e5 mistral. However, I assume this is notably faster, but it would be nice we an actual comparison of both scores and speed.
> 
> We also implement models in `mteb/models`. I believe we could do this here as well.
> 
> I assume it produces different scores (given vLLM) so it might be worth changing the name (related to #1211), but we can take that after a test.

Hey @KennethEnevoldsen - apologies, I did not meant to submit this PR onto upstream it was just a simple POC for myself to eval the embedding models running in vLLM. I managed to match the scores running through GritLM on a few STS tasks. I made this to automate correctness testing in vLLM

If this is something you would like to see in MTEB, I will clean it up

KennethEnevoldsen:
> If this is something you would like to see in MTEB, I will clean it up

If you can get it to match the performance at a faster speed I would love to have it added (cc @Muennighoff; it might be worth integrating this in GritLM)

KennethEnevoldsen:
Will close this issue as it seems to have gotten stale, though I believe this is a great PR so do feel free to re-open it or add a PR to GritLM.

PR #1275, opened 2024-10-03
fix: derive `results_directory` path from `results_repo` name

dbuades:
The `load_results` function includes a `results_repo` parameter, allowing the use of an alternative Git repository for storing the benchmarking results. This parameter is then passed to the `download_of_results` function. However, the current implementation requires the alternative repository to be named _results_, as this name is hardcoded.

This PR updates the code to derive the results_directory path from the repository name, removing the hardcoded requirement and giving more flexibility in naming the alternative repository.

## Checklist


isaac-chung (approved):
Nice catch, thank you!

PR #1225, opened 2024-09-18
Models refactoring, add stella, BGE, Linq

AlexeyVatolin:
I created the `SentenceTransformerWrapper` class to remove code duplication and make it easier to add new models. I also added the task_type parameter to the model, as it is needed for some instructions.
Added models Stella 400M, 1.5B, Linq-Embed-Mistral and bge-multilingual-gemma2.

## Checklist

- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 

### Adding datasets checklist

**Reason for dataset addition**: ... 

- [ ] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb -m {model_name} -t {task_name}` command.
  - [ ] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [ ] `intfloat/multilingual-e5-small`
- [ ] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [ ] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [ ] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [ ] Run tests locally to make sure nothing is broken using `make test`. 
- [ ] Run the formatter to format the code using `make lint`. 

### Adding a model checklist

 - [x] I have filled out the ModelMeta object to the extent possible
 - [x] I have ensured that my model can be loaded using
   - [x] `mteb.get_model(model_name, revision_id)` and
   - [x] `mteb.get_model_meta(model_name, revision_id)`
 - [x] I have tested the implementation works on a representative set of tasks.

PR #1221, opened 2024-09-15
feat: leverage SentenceTransformers' query/passage specific prompts

dbuades:
Some embedding models require different prompts for encoding queries versus encoding passages. SentenceTransformers provides this functionality through two features:

1. A model may be initialized with a prompt dictionary via the [prompts](https://sbert.net/docs/package_reference/sentence_transformer/SentenceTransformer.html#sentence_transformers.SentenceTransformer) argument.
2. When calling `model.encode()`, we can pass a [prompt_name](https://sbert.net/docs/package_reference/sentence_transformer/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode) argument. If the argument matches a key in the prompt dictionary, that prompt will be used automatically.

The main change in this PR is the modification of the logic in `model_encode.py` to leverage this functionality. To maintain compatibility with instruct models that currently use the task name as `prompt_name`, I added a new `prompt_type` argument and renamed `prompt_name` to `task_name`. The other change is that `encode_queries` and `encode_corpus` now specify `query` and `passage` as their respective `prompt_type` argument, which matches SentenceTransformers' default. As a side benefit, this allows instruct models to also have specific prompts for queries and passages. 

Previously, the way to achieve this was by creating wrappers for these models: `arctic_models.py`, `bge_models.py`, `mxbai_models.py`, and `e5_models.py`. While this worked correctly, it added complexity to maintain.

With the changes in this PR, those wrappers are no longer needed, as we can now specify the `prompts` argument directly in the loader. In the case of arctic_models, since they define the prompts in their [config_sentence_transformers.json](https://huggingface.co/Snowflake/snowflake-arctic-embed-m-v1.5/blob/main/config_sentence_transformers.json), we don't even need to modify the loader.

@KennethEnevoldsen, please let me know your thoughts. I'm happy to discuss any concerns you may have!

## Checklist

- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`.

PR #1278, opened 2024-10-05
feat: Use prompt by task_type

Samoed:
Previously, prompts were selected for each task by `task_name`, but now they can be selected by both `task_name` and `task_type`. I tested `ai-forever/ru-en-RoSBERTa`, and the results matched the leaderboard. I think that `model_encode` can now take prompts and pass them to the model.

## Checklist


mteb/evaluation/evaluators/model_encode.py
```diff
@@ -25,13 +26,16 @@ def model_encode(
...
         prompt_type: The prompt type (e.g. "query" | "passage") to use for building the encoding prompt
         **kwargs: Additional arguments to pass to the model.encode method
     """
     ## The order of priorities are:
     # 1. Composed prompt of task name + prompt type
     # 2. Specific task prompt
-    # 3. Specific prompt type
+    # 3. Composed prompt of task type + prompt type
+    # 4. Specific task type prompt
+    # 5. Specific prompt type
```
KennethEnevoldsen, line 38: Should be moved to docstring

KennethEnevoldsen:
Any reason to not extract it from the task_name?

```
task = mteb.get_task(task_name)
task_type = task.metadata.type
```

I also think the current six-step flow is a bit complicated, might be worth documenting it in the docstring not only in model_encode function but also in the interface

tests/test_benchmark/test_benchmark.py
```diff
@@ -99,8 +109,10 @@ def encode(self, sentences, **kwargs):
 
     if isinstance(task_name, mteb.AbsTask):
         tasks = [task_name]
+        _task_type = task_name.metadata.type
     else:
         tasks = mteb.get_tasks(tasks=[task_name])
+        _task_type = tasks[0].metadata.type
```
KennethEnevoldsen, line 115: Not sure what is happening here?

KennethEnevoldsen (approved):

Samoed:
That's great idea

tests/test_benchmark/test_benchmark.py
Samoed: I was tasting prompts there in the beginning. Will remove it

tests/test_benchmark/conftest.py
```diff
@@ -0,0 +1,12 @@
...
+import pytest
+
+import mteb
+
+from .task_grid import MOCK_TASK_REGISTRY
+
+
+@pytest.fixture
+def mock_mteb_get_task(monkeypatch):
+    monkeypatch.setattr(mteb.overview, "TASKS_REGISTRY", MOCK_TASK_REGISTRY)
```
KennethEnevoldsen, line 12: I don't think we actually want this - we should be able to test mteb.get_tasks() works as intended with the real tasks.

KennethEnevoldsen: But I see that you must have added this to get mteb.get_task() to work within the model_encode - in which case I am probably more a fan of the previous approach (sorry about that), since this change is likely to cause tests which behaves unexpectedly

Samoed: I understand your concerns, but I believe this approach is better because it simplifies the code significantly and allows us to gather more information about tasks. In the future, metadata (e.g., instructions) could be used more extensively, and using `mteb.get_task` inside `model_encode` would be a much simpler solution than passing parameters throughout the code.

KennethEnevoldsen: Hmm I see the point. I also really like the code simplification. 

I think a solution is to change the interface such that it only takes in promptType and task_name (i.e. remove task_type). Then we can introduce a SentenceTransformerWrapper (similar to what is done in #1225), which can then do all the heavy lifting. For the test, we then simply create a wrapper that uses the MOCK_REGISTRY instead of the mteb.get_tasks().

This keeps the simplicity, moved sentence transformer specific transformations to a separate wrapper and removes the need for monkeypatching in tests

let me know what you think about this approach

Samoed: You suggest moving `model_encode` to separate wrapper?

KennethEnevoldsen: yes, right now it is really intended for SentenceTransformers (cc @isaac-chung just to get a second opinion)

isaac-chung: Hmm, I might be missing something but I'm not following how this significantly simplifies the code?

Samoed: Using `mteb.get_task` in `model_encode` or wrappers is simpler than passing arguments from tasks or evaluators.

KennethEnevoldsen: Not sure what part you mean, but using `mteb.get_task` vs passing task_type down through the call stack does notably reduce the information handling (and avoids accidental defaults overwriting e.g. task_type)

isaac-chung: Ah I see, thanks!

Samoed:
I created a separate class for wrapping models and removed the `encode_corpus` and `encode_queries` methods #1284

mteb/models/promptriever_models.py
```diff
@@ -11,12 +11,10 @@
...
 from .repllama_models import RepLLaMAWrapper
 
-logging.basicConfig(level=logging.WARNING)
 logger = logging.getLogger(__name__)
 
-EncodeTypes = Literal["query", "passage"]
-
 
+# todo PromptrieverWrapper(RepLLaMAWrapper
 class PromptrieverWrapper(RepLLaMAWrapper):
```
Samoed, line 18: @orionw Models in this file should use `RepLLaMAWrapper` or `PromptrieverWrapper`?

orionw: The Promptriever models can re-use a lot of code from RepLlama, so they inherit from them. But they are a different model.  

Does this answer the question? Sorry if I misunderstood

Samoed: In this file, you created the `PromptrieverWrapper` class, but the models are using the `RepLLaMAWrapper` class

orionw: Ah nice catch, thank you!  It looks like the differences are minimal (just making sure the query ends with a punctuation) so we’re probably fine but I can also submit a separate PR to make sure nothing breaks (since I apparently haven’t run it).

Thanks again for pointing it out!

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -0,0 +1,189 @@
...
+        if hasattr(self.model, "prompts") and task_to_prompt_name is not None:
+            if model_prompts is None:
+                model_prompts = {v: v + " " for v in task_to_prompt_name.values()}
+            if len(self.model.prompts) == 0:
+                self.model.prompts = model_prompts
+                logger.info(f"Model prompts will be overrided with {model_prompts}")
+            elif override_prompts:
+                # override existing model prompts
+                self.model.prompts = model_prompts
+                logger.info(f"Model prompts will be overrided with {model_prompts}")
```
KennethEnevoldsen, line 61:
```suggestion
                logger.info(f"Model prompts will be overrided with {model_prompts}")

```suggestion
                logger.info(f"Model prompts will be overwritten with {model_prompts}")

```

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Args:
+            model: The SentenceTransformer model to use. Can be a string (model name), a SentenceTransformer model, or a CrossEncoder model.
+            revision: The revision of the model to use.
+            task_to_prompt_name: A dictionary mapping task names to prompt names.
+                First priority is given to the composed prompt of task name + prompt type (query or passage), then to the specific task prompt,
+                then to the composed prompt of task type + prompt type, then to the specific task type prompt,
+                and finally to the specific prompt type.
+            model_prompts: A dictionary mapping prompt names with prompts. If not provided, the model's prompts will be used.
+            override_prompts: Whether to override the model's existing prompts with the provided model_prompts.
```
KennethEnevoldsen, line 38: I would just always do this - if you specify model_prompts then I would expect that the new ones are used.

mteb/evaluation/MTEB.py
```diff
@@ -322,6 +325,8 @@ def run(
 
         Args:
             model: Model to be used for evaluation
+            task_to_prompt_name: Dictionary mapping task names to prompt names. If None, the default prompt names are used.
+            model_prompts: Dictionary mapping task types to prompt names. If None, the default prompt names are used.
```
KennethEnevoldsen, line 329: Shouldn't this be specified in the model? I don't think we should specify it here.

Similar with `task_to_prompt_name`.

Samoed: There's an issue where, if users pass a `SentenceTransformer`, there won't be a mapping from tasks to prompts. I wanted to allow specifying this, but I can remove that option if needed.

KennethEnevoldsen: Can you give an example of this?

Samoed: In the README example:

```python
import mteb

model = SentenceTransformer("sentence-transformers/all-MiniLM-L6-v2")
```
It's not possible to add prompts to the model in this way. However, this may be acceptable since users should create custom models instead.

KennethEnevoldsen: They can specify prompts like this I believe:

```
model = SentenceTransformer("sentence-transformers/all-MiniLM-L6-v2", prompts ={"task_name": "myprompt"}
```

Samoed: Yes, this approach could remove `model_prompts`, but how would the mapping between tasks and `prompt_name` be specified? Should I just keep `task_to_prompt_name`? I don't want to do that because some models in SentenceTransformers already have prompts and I wanted to reuse them

KennethEnevoldsen: I would image that the user would specify the prompt for a given task like:

```
model = SentenceTransformer("sentence-transformers/all-MiniLM-L6-v2", 
    prompts={
      # task specific:
      "BornholmBitextMining": "Retrieve parallel sentences in Danish and Bornholmsk",
      # retrieval with prompt-type:
      "MIRACLReranking-query": "Query: Given a question, retrieve Wikipedia passages that answer the question",
      "MIRACLReranking-corpus": "", # or just leave out
    }
)
```

Let me know if there is something I am missing?

Samoed: I can merge task_to_prompt_name and model prompts and check only model prompts to make this work like this

KennethEnevoldsen:
@Samoed is this ready for review?

Samoed:
Yes, PR ready for review, but I runned `bge-ase-en-v1.5`, `multilingual-e5-large`, `mxbai-embed-large-v1`, but their results, don't match with leaderboard. I'll rerun them with mteb from main to check them.

Samoed:
I have evaluated models and their results are the same with main branch, except for AskUbuntuQuestions. 
[mteb_res.zip](https://github.com/user-attachments/files/17432965/mteb_res.zip)
I tried to reevalute locally AskUbuntuQuestions on multilingual-e5-small model and if I **remove** prompts I get same results for this task as in main, but if I add prompts I get a bit lower results.

Mteb from main branch:
```
INFO:mteb.evaluation.evaluators.model_encode:No combination of task name and prompt type was found in model prompts. Removing prompt_name argument.
INFO:mteb.evaluation.evaluators.model_encode:Encoding 6743 sentences.
INFO:mteb.evaluation.evaluators.RerankingEvaluator:Evaluating...
INFO:mteb.evaluation.MTEB:Evaluation for AskUbuntuDupQuestions on test took 4.56 seconds
INFO:mteb.evaluation.MTEB:Scores: {'default': {'map': 0.5680031628804745, 'mrr': 0.7094974277799763, 'nAUC_map_max': 0.2685051042082004, 'nAUC_map_std': 0.2013157249456535, 'nAUC_map_diff1': 0.04353361902146947, 'nAUC_mrr_max': 0.38623389097835026, 'nAUC_mrr_std': 0.23623505561915978, 'nAUC_mrr_diff1': 0.1194833073518608, 'main_score': 0.5680031628804745}}
```
This branch with prompts:
```
INFO:mteb.models.sentence_transformer_wrapper:Using prompt name=`passage`  for task=AskUbuntuDupQuestions prompt_type=passage
INFO:mteb.models.sentence_transformer_wrapper:Encoding 6743 sentences.
INFO:mteb.evaluation.evaluators.RerankingEvaluator:Evaluating...
INFO:mteb.evaluation.MTEB:Evaluation for AskUbuntuDupQuestions on test took 4.85 seconds
INFO:mteb.evaluation.MTEB:Scores: {'default': {'map': np.float64(0.5642404534571142), 'mrr': np.float64(0.6937211449676823), 'nAUC_map_max': np.float64(0.27381102594997414), 'nAUC_map_std': np.float64(0.1693578941478997), 'nAUC_map_diff1': np.float64(0.08137714630662059), 'nAUC_mrr_max': np.float64(0.38416684118951216), 'nAUC_mrr_std': np.float64(0.22024098705009196), 'nAUC_mrr_diff1': np.float64(0.21548592467525338), 'main_score': np.float64(0.5642404534571142)}}
```
This branch without prompts:
```
INFO:mteb.evaluation.MTEB:Scores: {'default': {'map': np.float64(0.5680031628804745), 'mrr': np.float64(0.7094974277799763), 'nAUC_map_max': np.float64(0.2685051042082004), 'nAUC_map_std': np.float64(0.2013157249456535), 'nAUC_map_diff1': np.float64(0.04353361902146947), 'nAUC_mrr_max': np.float64(0.38623389097835026), 'nAUC_mrr_std': np.float64(0.23623505561915978), 'nAUC_mrr_diff1': np.float64(0.1194833073518608), 'main_score': np.float64(0.5680031628804745)}}
```

mteb/evaluation/evaluators/RerankingEvaluator.py
```diff
@@ -147,22 +136,17 @@ def compute_metrics_individual(self, model):
...
-        encode_corpus_func = (
-            model.encode_corpus if hasattr(model, "encode_corpus") else model.encode
-        )
+
         if self.evaluator_type == "standard":
             results = self._encode_candidates(
-                encode_queries_func=encode_queries_func,
-                encode_corpus_func=encode_corpus_func,
+                encode_queries_func=model.encode,
+                encode_corpus_func=model.encode,
```
KennethEnevoldsen, line 143: This seems like a mistake? Either is should just be 1 encode or we should make it two different ones

KennethEnevoldsen (approved):

KennethEnevoldsen:
>I tested AskUbuntuQuestions locally using the multilingual-e5-small model. When removing prompts, I obtained the same results as the main branch. However, when I added prompts, the results were slightly lower.

>So, should I remove the query and passage prompts for Reranking?

I have highlighted a section where the might be a bug - that might resolve the discrepancy

mteb/evaluation/evaluators/RerankingEvaluator.py
Samoed: I kept the current setup to avoid a refactor, but will do this now. The main issue is that this version uses prompts, whereas the version on the main branch does not. Should I remove the prompts to align the results with the main branch?

Samoed:
@KennethEnevoldsen I tested the commit just before merging #1221 (commit 11518edd03b916a63e84505592f0d1a32a058d49) and obtained the following results:
```bash
mteb run -m intfloat/multilingual-e5-small -t SciDocsRR --lang eng --overwrite
```
```
INFO:mteb.evaluation.MTEB:Scores: {'default': {'map': 0.7825588176371663, 'mrr': 0.9386869847654161, 'nAUC_map_max': 0.5018386842952304, 'nAUC_map_std': 0.6401228473171651, 'nAUC_map_diff1': 0.08046436666347959, 'nAUC_mrr_max': 0.7836770455942673, 'nAUC_mrr_std': 0.7179299106806627, 'nAUC_mrr_diff1': 0.5521448612994969, 'main_score': 0.7825588176371663}}
```
These results match my implementation, suggesting that #1221 did not cover everything, and some parts might have been missed.

mteb/evaluation/evaluators/RerankingEvaluator.py
KennethEnevoldsen: Let us remove the prompt to avoid a major bump but create an issue on it and mark it v2.0.0

KennethEnevoldsen: Would be great with to invalidate "results on [these tasks] before version X.X.X, should be invalid" going forward.

For now, let us include the prompts (that is intended behavior, after all).

Samoed: > Would be great with to invalidate "results on [these tasks] between version X.X.X and X.X.X, should be invalid" going forward.

You mean I should add warning when loading results in splits for reranking tasks?

> For now, let us include the prompts (that is intended behavior, after all).

You mean #1300? Because current implementation passing instructions to wrappers correctly, like previously

KennethEnevoldsen: > You mean I should add warning when loading results in splits for reranking tasks?

More of a general concern. Sometimes, we find a mistake for specific versions of MTEB and we would like to not include those in the benchmark.

KennethEnevoldsen:
I believe this is all ready to merge! Just checked it to see if it introduced any breaking changes, and the only one is the prompt, which I believe is fine. I have changed it to a minor bump though

PR #1292, opened 2024-10-16
fix QBQTC in C-MTEB

ZiyiXia:
When reproducing C-MTEB evaluation, found that the QBQTC class (in STS/zho) should add a metadata_dict() function.


isaac-chung (approved):
Thanks! Nice catch.

PR #1293, opened 2024-10-16
[mieb] Adding MOCOv3 models

isaac-chung:
## Checklist

### Adding a model checklist


isaac-chung:
@gowitheflow-1998 a note that on the same HF collection, there are only the base and large models, but not the small model. In the example the authors shared in [this discussion](https://huggingface.co/nyu-visionx/moco-v3-vit-b/discussions/1), there's an implementation for small as well, but that involves a ton more code. Would base and large be sufficient for our needs?

gowitheflow-1998:
> @gowitheflow-1998 a note that on the same HF collection, there are only the base and large models, but not the small model. In the example the authors shared in [this discussion](https://huggingface.co/nyu-visionx/moco-v3-vit-b/discussions/1), there's an implementation for small as well, but that involves a ton more code. Would base and large be sufficient for our needs?

Thanks, look great! yeah base and large are sufficient this PR. Can add small with a seperate PR when we need it.

gowitheflow-1998 (approved):

PR #1294, opened 2024-10-17
fix: Remove non-existent eval split of CMNLI

ZiyiXia:
Fix the KeyError of [CMNLI dataset](https://huggingface.co/datasets/C-MTEB/CMNLI) in C-MTEB:  'test' does not exist.


isaac-chung (approved):
Thanks @ZiyiXia ! I have also updated the description to specifically spell out what is "wrong" with the key. This helps maintainers understand what problem the PR is solving easier.

PR #1300, opened 2024-10-18
fix: Integrate prompts to task metadata

Samoed:
## Checklist

I have added prompts to all tasks except for reranking because I'm not sure whether the prompts should be `query`/`passage` or just a general prompt. Also, this problem presented in #1278


mteb/abstasks/TaskMetadata.py
```diff
@@ -227,6 +245,10 @@ class TaskMetadata(BaseModel):
     sample_creation: SAMPLE_CREATION_METHOD | None = None
     bibtex_citation: str | None = None
 
+    prompt: str | None = None
```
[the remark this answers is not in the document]
KennethEnevoldsen: I would move it up next to the description in the order of kwargs

mteb/abstasks/TaskMetadata.py
```diff
@@ -165,6 +165,20 @@
...
+    """A dictionary containing the dataset path and revision.
+
+    Args:
+        path: The path to the dataset.
+        revision: The revision of the dataset.
+        trust_remote_code: Whether to trust the remote code.
+    """
+    path: str
+    revision: str
+    trust_remote_code: bool
```
KennethEnevoldsen, line 179: does this then require a trust_remote_code?

KennethEnevoldsen (approved):
Loving the idea! Added a few comments to simplify. 

Can't seem to find where the new prompts are being used. Seems like they are not passed to the model?

mteb/abstasks/TaskMetadata.py
Samoed: No, `total=False` means that not all parameters can be passed to the dict

mteb/abstasks/TaskMetadata.py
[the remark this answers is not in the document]
Samoed: I considered that, and I'm a bit concerned about potential misspellings and other issues. However, while I was writing this, I realized that I could implement validators for that field. I'll change as you suggest

Samoed:
> Seems like they are not passed to the model
Currently, yes, because I'm waiting for https://github.com/embeddings-benchmark/mteb/pull/1278 to merge and pass prompts to wrapper here, but I'll make this PR compatible with the current main if you want.

Samoed:
I will run some instruct models on weekends

Samoed:
I've tested [multilingual-e5-large-instruct](https://github.com/user-attachments/files/17531128/intfloat__multilingual-e5-large-instruct.zip) and got similar results

KennethEnevoldsen:
wonderful - let us get the tests to pass and a review from @Muennighoff or @isaac-chung once that is done, I believe we can merge it in.

isaac-chung (commented):
Thank you so much for working on this! A few suggestions below - nothing major.

Samoed:
I've tested e5-mistral and got very different results. I'll look what's different

Samoed:
I forgot to apply instruct tempalte to `e5-mistral` now results matches. Also fixes #1346

Samoed:
Seems like `BrazilianToxicTweetsClassification` completly broken in tests

KennethEnevoldsen (approved):
Hmm seem like tests fail due to the Brazilian test case... 

Otherwise I believe this change looks great! It will cause some conflict with MIEB (@isaac-chung and @gowitheflow-1998), but I think it is generally a good change for MIEB as well.

mteb/model_meta.py
```diff
@@ -70,11 +70,12 @@ class ModelMeta(BaseModel):
         reference: A URL to the model's page on huggingface or another source.
         languages: The languages the model is intended for specified as a 3 letter language code followed by a script code e.g. "eng-Latn" for English
             in the Latin script.
-        use_instuctions: Whether the model uses instructions E.g. for prompt-based models. This also include models that require a specific format for
+        use_instructions: Whether the model uses instructions E.g. for prompt-based models. This also include models that require a specific format for
             input such as "query: {document}" or "passage: {document}".
         zero_shot_benchmarks: A list of benchmarks on which the model has been evaluated in a zero-shot setting. By default we assume that all models
             are evaluated non-zero-shot unless specified otherwise.
     """
+    model_config = ConfigDict(extra="forbid")
```
Samoed, line 78: For now, I added `extra="forbid"` to make sure that all parameters of model metadata are correct, but I can remove it

Samoed:
@KennethEnevoldsen Can we merge?

Issue #946, opened 2024-06-17
Be able to cache embeddings and load them


orionw:
For most users, being able to cache their embedded docs and/or provide a cached embedding file is probably overkill.

However, there are many situations where it would be helpful to have an option to cache them.  For example, experiments where you alter the query/document set to for speedups (as I'm doing now) or if you're testing the effect of different prefixes/instructions over the same dataset.

I typically use `pyserini` for caching the index so that we can quickly search over it later, but that doesn't integrate nicely with `mteb`.  I think it would fairly straightforward to implement this: (1) we need to take in a flag of whether to cache the embeddings, cache them to a file that corresponds to the dataset and model name and (2) provide an option to read in a cached embedding file. 

I don't have bandwidth for this right now, but if anyone does it would be an excellent addition.

tenzu15:
Hey @orionw ,

I would like to try this if possible!

orionw:
Awesome @tenzu15! It would be great to be able to pass the two flags in the `mteb.run` command.  Something like `cache_embeddings: bool = True` and `cached_embedding_file: str`

This would need to be changed in the [RetrievalEvaluator](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/evaluation/evaluators/RetrievalEvaluator.py) class for now.  If it's useful for other tasks, we can implement it there also. Also cc'ing @KennethEnevoldsen who may have opinions on where this should be added/what the names should be.

But feel free to start @tenzu15. If you have any questions feel free to make a draft PR and cc me!

KennethEnevoldsen:
@orionw wouldn't it be better to implement a more general model wrapping for this such that it works for all tasks?

```
class ModelWrap():
  def __init__(self, model):
     self.model = model
  
  def encode(...):
     embeddings = self.model.encode(...)
     self.store_embeddings(sentences, embeddings)   
     return embeddings
```

isaac-chung:
There's some background discussion related to the topic from https://github.com/embeddings-benchmark/mteb/issues/354#issuecomment-2055435397 as well.

orionw:
+1 @KennethEnevoldsen, I think a wrapper is a great idea and even simpler to implement.

KennethEnevoldsen:
It sounds like we settled on a wrapper here. In which case I don't think it is something that should be within `mteb`. Let me know if you disagree, then I will re-open the issue

orionw:
[...]

PR #1307, opened 2024-10-21
Cache the embeddings when requested

orionw:
This PR would solve #946 and allow the user to cache embeddings with:
```
# define your task and model above as normal
...
# wrap the model with the cache wrapper
from mteb.models.cache_wrapper import CachedEmbeddingWrapper
model_with_cached_emb = CachedEmbeddingWrapper(model, cache_path='path_to_cache_dir')
# run as normal
evaluation.run(model, ...) 
```

I have it right now set to cache the embeddings per task, but this could also be a general cache which could be used across tasks which share documents (like Wikipedia, etc.). I can modify it to have a general cache, but there are perks to the task-specific cache also (like being able to delete just one portion). 

I tried to make it scalable, although I haven't stress tested large datasets yet. 

## Checklist


Samoed:
I created a wrapper for the `SentenceTransformer` model in #1278. I believe it might be possible to combine them in some way

orionw:
Ah shoot, sorry I missed that @Samoed! Where exactly is it?

Samoed:
https://github.com/embeddings-benchmark/mteb/blob/c2cf927461bc53395c1eab83acb4f11472934633/mteb/models/sentence_transformer_wrapper.py

orionw:
Gotcha, thank you @Samoed! I think, and please correct me if I'm wrong, that your wrapper handles the prompting.  I think my wrapper is only for caching and only happens on demand (shouldn't be a default).  I think they can co-exist, but let me know if I am misunderstanding!

orionw:
> Also, I've updated the API of the Encoder interface, and it no longer includes encode_queries and encode_corpus.

Ah, good to know! What has replaced it? Happy to modify this to wrap the function that needs to be wrapped.  The core of this is just wrapping the main encode function so that we can create a map of text->embedding.

Samoed:
In my PR, the wrapper has:
```python
def encode(
        self,
        sentences: Sequence[str],
        *,
        task_name: str,
        **kwargs: Any,
    ) -> np.ndarray:
    ....
    return self.model.encode(
            sentences,
            prompt_name=prompt_name,
            **kwargs,
        )
```
`task_name` is used to get the prompt relevant to the task, and `prompt_type` specifies the type of prompt (e.g., query or passage). However, it's currently difficult to determine which wrapper will be applied first and which parameters should be supported.

README.md
```diff
@@ -369,6 +369,21 @@ df = results_to_dataframe(results)
...
+  <summary>  Caching Embeddings To Re-Use Them </summary>
+
+
+### Caching Embeddings To Re-Use Them
+
+There are times you may want to cache the embeddings so you can re-use them. This may be true if you have multiple query sets for the same corpus (e.g. Wikipedia) or are doing some optimization over the queries (e.g. prompting, other experiments). You can setup a cache by using a simple wrapper, which will save the cache per task in the `cache_embeddings/{task_name}` folder:
+
+```python
+... # define your task and model above as normal
+evaluation.run(model, encode_kwargs={'cache_embeddings': 'path_to_cache_dir', "vector_dim": 512}) # setting this flag will create/use the cache
```
Samoed, line 382: Maybe make this 
```python
evaluation.run(model, cache_embeddings='path_to_cache_dir', vector_dim= 512)
```
Also, is it possible to make it without specifying vector_dim?

orionw: +1, you can not set it it and it will default to 768.  But to create the mmap matrix it needs some dimension.

I could have it wait until it encodes something and then get the dimension/create the matrix? And/or store a dimensionality number in the cache so it could be loaded without it needing to be passed. Does that seem preferable?

mteb/evaluation/CacheWrapper.py
```diff
@@ -0,0 +1,261 @@
...
+class CachedEmbeddingWrapper:
+    def __init__(self, model: Any, cache_path: Union[str, Path], vector_dim: int = 768):
+        self._model = model
+        self.cache_path = Path(cache_path)
+        self.cache_path.mkdir(parents=True, exist_ok=True)
+        self.vector_dim = vector_dim
+
+        if hasattr(model, "encode_queries") and hasattr(model, "encode_corpus"):
+            self.encode_method = "split"
+            self.query_cache = TextVectorMap(
```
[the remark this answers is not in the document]
orionw: Agree! I do that here: https://github.com/embeddings-benchmark/mteb/blob/89db68e943713414e2eca1680085572a8c0790e8/mteb/evaluation/MTEB.py#L378 which is then passed to this class.

README.md
Samoed: I see that you passed default, but what will be if model has 1024 vector size? Will it be loaded correctly? I think that load from cache will be more preferable

mteb/evaluation/CacheWrapper.py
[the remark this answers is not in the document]
Samoed: Ah, I see. But since you're recreating the model, this would result in a wrapper inside another wrapper, leading to multiple caches of the same data.

README.md
orionw: Sounds good to me, added the inferring details in so it doesn't have to be passed.

mteb/evaluation/CacheWrapper.py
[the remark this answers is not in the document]
Samoed: In main branch `task_name` and `prompt_type` currently passing to model_encode and from there to `model.encode` in kwargs in `prompt_name`

Samoed:
Yes, I've made changes so that parameters are passed to all wrappers

KennethEnevoldsen (commented):
Looking great @orionw! Made a few suggestions. Wrapping the SentenceTransformerWrapper should not be a problem at all.

I see that you use a pkl as a wrapper, I was thinking it might be better to use a data format that is [safe](https://www.benfrederickson.com/dont-pickle-your-data/). 
Mostly if people intended to share this data.

mteb/evaluation/CacheWrapper.py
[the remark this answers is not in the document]
orionw: Thanks @Samoed for pointing out this issue.  I've moved it out to avoid this issue!

README.md
```diff
@@ -369,6 +369,21 @@ df = results_to_dataframe(results)
...
+
+
+### Caching Embeddings To Re-Use Them
+
+There are times you may want to cache the embeddings so you can re-use them. This may be true if you have multiple query sets for the same corpus (e.g. Wikipedia) or are doing some optimization over the queries (e.g. prompting, other experiments). You can setup a cache by using a simple wrapper, which will save the cache per task in the `cache_embeddings/{task_name}` folder:
+
+```python
+... # define your task and model above as normal
+evaluation.run(model, encode_kwargs={'cache_embeddings': 'path_to_cache_dir'}) # setting this flag will create/use the cache
+```
```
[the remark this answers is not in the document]
orionw: +1 great feedback, thanks @KennethEnevoldsen! I suppose if it scales too big, we can always switch to ujson or similar. 

Update it to work like this, agree it's much easier to maintain like that.

orionw:
> I see that you use a pkl as a wrapper, I was thinking it might be better to use a data format that is [safe](https://www.benfrederickson.com/dont-pickle-your-data/).

I went with a Pickle for the compression and speed, but agree JSON is better overall. I switched back to it and moved it to the models section of the code.

KennethEnevoldsen (approved):
I believe things are looking ready to merge.

mteb/models/cache_wrapper.py
```diff
@@ -0,0 +1,330 @@
...
+    def close(self):
+        if hasattr(self, "vectors") and self.vectors is not None:
+            self.vectors.flush()
+            del self.vectors
+            self.vectors = None
+        logger.info(f"Closed TextVectorMap in directory: {self.directory}")
+
+
+class CachedEmbeddingWrapper:
+    def __init__(self, model: Any, cache_path: Union[str, Path]):
```
KennethEnevoldsen, line 203:
```suggestion
    def __init__(self, model: Encoder, cache_path: Union[str, Path]):
```

mteb/models/cache_wrapper.py
```diff
@@ -0,0 +1,330 @@
...
+
+        logger.info(
+            f"Initialized CachedEmbeddingWrapper with {self.encode_method} encoding method"
+        )
+
+    def _wrap_split_encode_methods(self):
+        original_encode_queries = self._model.encode_queries
+        original_encode_corpus = self._model.encode_corpus
+
+        def wrapped_encode_queries(
```
KennethEnevoldsen, line 234: I believe we should remove these

mteb/models/cache_wrapper.py
```diff
@@ -0,0 +1,330 @@
...
+        def wrapped_encode(
+            texts: List[str], batch_size: int = 32, **kwargs
+        ) -> np.ndarray:
+            return self._cached_encode(
+                texts, self.cache, original_encode, batch_size, **kwargs
+            )
+
+        self._model.encode = wrapped_encode
+
+    def _cached_encode(
```
KennethEnevoldsen, line 263: I would just call this `encode`

mteb/models/cache_wrapper.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def close(self):
+        if hasattr(self, "vectors") and self.vectors is not None:
+            self.vectors.flush()
+            del self.vectors
+            self.vectors = None
+        logger.info(f"Closed TextVectorMap in directory: {self.directory}")
+
+
+class CachedEmbeddingWrapper:
```
KennethEnevoldsen, line 202:
```suggestion
class CachedEmbeddingWrapper(Encoder):
```

Samoed: I've also created a dummy class for wrappers to indicate that the model is already wrapped. I think any new wrapper classes should inherit from this as well.

KennethEnevoldsen: Will you link to it

Samoed: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/wrapper.py

mteb/models/cache_wrapper.py
orionw: Sure - but does this not cause issues with custom models?

mteb/models/cache_wrapper.py
orionw: Gotcha, so inherit from Wrapper, Encoder then!

isaac-chung:
Sorry I was late to the party. I wish I was able to share some feedback sooner, esp. how this would interact with the user. I found @orionw 's original proposal to use this to be very convenient for the user. This could even be toggled by a flag with a default directory location, and the `path_to_cache_dir` param could be a secondary way to specific where if the user desires to.
```
evaluation.run(model, encode_kwargs={'cache_embeddings': 'path_to_cache_dir'})
```

The reason for me leaning towards the above option is that it does not require importing anything new from the existing imports. The only addition would be a flag (in CLI or the run method). 

That's said, I'm also just happy that we would now have this capability. If we go with this PR as is, the PR description can be updated to include the following:
```python
# define your task and model above as normal
...
# wrap the model with the cache wrapper
from mteb.models.cache_wrapper import CachedEmbeddingWrapper
model_with_cached_emb = CachedEmbeddingWrapper(model, cache_path='path_to_cache_dir')
# run as normal
evaluation.run(model, ...) 
```

mteb/models/cache_wrapper.py
KennethEnevoldsen: I believe most custom models (if not all) have been refactored.

Samoed: If you about models inside models dir, I changed them all

orionw:
Thanks @isaac-chung! I am conflicted - I do think it's easier for the user (especially for CLI) as you mentioned and as I originally had, but I also agree with @KennethEnevoldsen that it's easier to maintain as a separate thing as we don't have to fit it into the MTEB class. 

I'm not sure how used this feature will be, but if it's a niche thing then it's fine to have these additional steps. Perhaps we leave it as is and can refactor into in the main code if we find it has a high usage/demand?

Thanks for the updated description for the PR, updated!

mteb/models/cache_wrapper.py
orionw: I think that could get confusing with the models `encode` and this encode which is just a wrapper. Is this style or was there a particular other reason? If just style, I think I'd prefer to leave it.

mteb/models/overview.py
```diff
@@ -21,6 +21,8 @@
     mxbai_models,
     nomic_models,
     openai_models,
+    promptriever_models,
+    repllama_models,
```
orionw, line 25: I was using these as a test and noticed they weren't here. Adding them but lmk if they were supposed to be elsewhere.

KennethEnevoldsen: This is perfect

KennethEnevoldsen:
I believe this can be merged in once the conflict is resolved

Issue #1297, opened 2024-10-18
Got unsupported ScalarType BFloat16


xujunrt:
In the RerankingEvaluator.py, np.asarray does not support BFloat16. A forced conversion should be added to avoid similar errors. For example, you can first convert the Tensor to float16 and then call numpy_asarray.

isaac-chung:
CC-ing @orionw and @KennethEnevoldsen on this. 

@xujunrt it would be great if you could also share an example of how you're using the Reranking evaluator. Thanks!

KennethEnevoldsen:
This sounds like it is related to #941 

Agree with @xujunrt. This sounds like a problem we could have in more places as well.

KennethEnevoldsen:
We have added mock encoders for a variety of format including BF16 (`MockSentenceTransformersbf16Encoder`) which we test in `test_encoder_dtype_on_task`. I believe this solves this issue, but if it still is there feel free to re-open

PR #1309, opened 2024-10-21
Add implementations of common reranker models

orionw:
Adds many types of rerankers, mainly MonoT5-style with a prediction token. Also includes some custom rerankers such as Jina's and BGE's. 

## Checklist

### Adding a model checklist


mteb/models/rerankers_custom.py
```diff
@@ -0,0 +1,251 @@
...
+            self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        model_args = {}
+        if self.fp_options:
+            model_args["torch_dtype"] = self.fp_options
+
+        try:
+            from FlagEmbedding import FlagReranker
+        except ImportError:
+            raise ImportError(
+                "FlagEmbedding is not installed. Please install it via `pip install FlagEmbedding`"
```
KennethEnevoldsen, line 66:
```suggestion
                "FlagEmbedding is not installed. Please install it via `pip install "mteb[FlagEmbedding]"`"
```

To allows us to lock the dependencies to ensure that it works in the future. Naturally also add to the dependencies.

mteb/models/rerankers_custom.py
```diff
@@ -0,0 +1,251 @@
...
+    release_date="2020-05-28",
+)
+
+# languages unclear: https://huggingface.co/jinaai/jina-reranker-v2-base-multilingual/discussions/28
+jina_reranker_multilingual = ModelMeta(
+    loader=partial(
+        _loader,
+        wrapper=JinaReranker,
+        model_name_or_path="jinaai/jina-reranker-v2-base-multilingual",
+        fp_options="bfloat16",
```
KennethEnevoldsen, line 197: Isn't that an odd default?

KennethEnevoldsen (approved):
A few minor things - but otherwise, great (did you see the issue of #1297?

mteb/models/rerankers_custom.py
orionw: I realize now this may not be common, but my understanding was that bfloat16 was basically identical in scores to fp32 and doesn't have the reliability issues of fp16.  That may be incorrect though. Would you suggest a different fp option?

I was not aware of #1297, but I think that wouldn't apply for rerankers? Could be wrong

orionw: LMK if I should change it @KennethEnevoldsen!

Samoed: I think this should be changed because T4 (in kaggle, colab) doesn't support bfloat16

orionw: Ah that's helpful.  So fp32 then? Or fp16?

orionw: I switched to FP16.  If we want FP32, lmk otherwise I think we're good

KennethEnevoldsen: FP16 is good

PR #1312, opened 2024-10-23
WIP: Leaderboard UI improvements

x-tabdeveloping:
I'm opening this PR so that you can track my progress in ironing out some of the UI problems in the Leaderboard.

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -65,25 +65,26 @@ task_select = gr.Dropdown(
     info="Select specific tasks to include",
 )
 
-css = """
-.scrollable {
-    overflow-y: scroll;
-    max-height: 400px
-}
+head = """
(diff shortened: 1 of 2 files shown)


x-tabdeveloping:
I'm cookin' this is how it lookin'
image

x-tabdeveloping:
I could finally add a scrollbar. It is completely beyond me which Gradio elements are okay with scrolling and why, but `gr.Row()` works for some reason. (`CheckBox`, `Group` and `Column` don't)

KennethEnevoldsen:
A couple of comments for readability:

- remove org from model name (and add reference link #1316)
- multiply scores by 100 and keep one decimal, e.g. 78.1 (@orionw not sure if this also works for followIR?). I believe anything more precise does not translate to a meaningful performance difference
- It might be ideal to move the "performance pr. task" to a separate table in a fold-down menu.

Let me know if you want this to be an issue

x-tabdeveloping:
@KennethEnevoldsen can we make it an issue?
Thanks for the comments btw

KennethEnevoldsen:
done!

PR #1316, opened 2024-10-24
feat: Update metadata for all models

KennethEnevoldsen:
Update metadata for all models. Specifying most (but not all) hyperparameters. This is required for the leaderboard metadata. Which we plan to move here (and only here)

@x-tabdeveloping there seems to be some cases that fails in linting due to the variable being unknown (I imagine it is a bug - will you take a look at it)

@mueninghoff and @orion I wanted to add memory usage as well, but could not find where is saved on the leaderboard. Do any of you guys know?

I am considering splitting open source into:
- open weights
- open (training) code
- open data

We also discussed adding if a model is zeros-shot on specific benchmarks I was thinking of adding this as:

- zeroshot_benchmarks = ["mteb(eng)", "mteb(multilingual)",...]. With a default of [].

Is there any other metadata that we would like to have but do not?

## Checklist


x-tabdeveloping:
@KennethEnevoldsen I know about the issue and fixed it here #1312 . I don't even understand how it got merged into main.

x-tabdeveloping:
Looks pretty good! I was wondering if we wanted to add training datasets to model meta. I think some HuggingFace repos have good info on this in the model card

x-tabdeveloping:
Another thing: I think we could just add the license for each model and then infer whether the model is openly accessible from the license in a property.

mteb/models/voyage_models.py
```diff
@@ -156,6 +156,13 @@ def _batched_encode(
     max_tokens=16000,
     embed_dim=1024,
     open_source=False,
+    n_parameters=None,
+    memory_usage=None,
+    license=None,
+    reference="https://blog.voyageai.com/2024/05/05/voyage-large-2-instruct-instruction-tuned-and-rank-1-on-mteb/",
+    similarity_fn_name="cosine",
+    framework=["Sentence Transformers", "PyTorch"],
```
Samoed, line 164: Voyage models are API only. Maybe change framework to API?

mteb/models/llm2vec_models.py
```diff
@@ -199,6 +262,15 @@ def loader_inner(**kwargs: Any) -> Encoder:
     name="McGill-NLP/LLM2Vec-Sheared-LLaMA-mntp-unsup-simcse",
     languages=["eng_Latn"],
     open_source=True,
-    revision=None,
+    revision="a5943d406c6b016fef3f07906aac183cf1a0b47d",
     release_date="2024-04-09",
+    n_parameters=7_111_000_000,
```
Samoed, line 267: Maybe write in millions parameters?

Samoed (approved):

mteb/models/llm2vec_models.py
x-tabdeveloping: I think this is better, honestly. I would rather have larger numbers than introduce confusion

x-tabdeveloping:
One can get dataset information about models on the hub like this:

```python
from huggingface_hub import ModelCard

card = ModelCard.load("sentence-transformers/all-MiniLM-L6-v2")
print(card.data)
```

```{
    "base_model": null,
    "datasets": [
        "s2orc",
        "flax-sentence-embeddings/stackexchange_xml",
        "ms_marco",
        "gooaq",
        "yahoo_answers_topics",
        "code_search_net",
        "search_qa",
        "eli5",
        "snli",
        "multi_nli",
        "wikihow",
        "natural_questions",
        "trivia_qa",
        "embedding-data/sentence-compression",
        "embedding-data/flickr30k-captions",
        "embedding-data/altlex",
        "embedding-data/simple-wiki",
        "embedding-data/QQP",
        "embedding-data/SPECTER",
        "embedding-data/PAQ_pairs",
        "embedding-data/WikiAnswers"
    ],
    "eval_results": null,
    "language": "en",
    "library_name": "sentence-transformers",
    "license": "apache-2.0",
    "license_name": null,
    "license_link": null,
    "metrics": null,
    "model_name": null,
    "pipeline_tag": "sentence-similarity",
    "tags": [
        "sentence-transformers",
        "feature-extraction",
        "sentence-similarity",
        "transformers"
    ]
}
```

mteb/models/voyage_models.py
KennethEnevoldsen: fixed!

mteb/models/llm2vec_models.py
KennethEnevoldsen: yea did consider it as well. I agree with @x-tabdeveloping, but it is a minor preference.

Samoed:
@KennethEnevoldsen Model memory usage calculated in leaderboard based on model size https://github.com/embeddings-benchmark/leaderboard/blob/5ad8ba29f2cae995a9c760739c656db845cbeb04/refresh.py#L400

KennethEnevoldsen:
> Looks pretty good! I was wondering if we wanted to add training datasets to model meta. I think some HuggingFace repos have good info on this in the model card. This could also help us with determining which benchmarks/tasks are zero-shot for given models.

This only works when the datasets is publicly available (which for the majority of the cases it is not. This also means that you can be zero-shot by accident/default and I want it to be a specific thing that developers consider.


> One can get dataset information about models on the hub like this:

This is nice to know. From this, we could extract:

- license (open source)
- framework (but only partly)
- languages (but we would have to add a mapping - however, there is no 1-1 mapping between the two codes. E.g. Chinese have 3 difference 3-letter codes.

from this, it seems like we could add an "autofill" option.

> @KennethEnevoldsen Model memory usage calculated in leaderboard based on model size https://github.com/embeddings-benchmark/leaderboard/blob/5ad8ba29f2cae995a9c760739c656db845cbeb04/refresh.py#L400

Alright, we could remove memory usage then and potentially add it as a property.

x-tabdeveloping:
But how do we do the zero-shot thing then? Do we just take people's words for it? Like how do we get an idea about what OpenAI does for instance? Or do we just label everything non-zero-shot by default, unless we know?

orionw:
> But how do we do the zero-shot thing then? Do we just take people's words for it? Like how do we get an idea about what OpenAI does for instance? Or do we just label everything non-zero-shot by default, unless we know?

Yeah I would say non-zero-shot by default and people have to opt in. I'm fine with taking people's word for it (e.g. if OpenAI said it, we could probably trust them), although we could an enforce an open-data clause.

mteb/model_meta.py
```diff
@@ -1,24 +1,26 @@
...
-Frameworks = Literal["Sentence Transformers", "PyTorch"]
+logger = logging.getLogger(__name__)
 
-pastdate_adapter = TypeAdapter(date)
-STR_DATE = Annotated[
-    str, BeforeValidator(lambda value: str(pastdate_adapter.validate_python(value)))
-]  # Allows the type to be a string, but ensures that the string is a valid date
+
+FRAMEWORKS = Literal[
+    "Sentence Transformers", "PyTorch", "GritLM", "LLM2Vec", "TensorFlow", "API"
```
orionw, line 21: I think [`Tevatron`](https://github.com/texttron/tevatron/tree/main) is another one that many people have used

mteb/models/e5_instruct.py
```diff
@@ -66,6 +66,15 @@ def encode(
...
     release_date=E5_PAPER_RELEASE_DATE,
+    framework=["GritLM", "PyTorch"],
+    similarity_fn_name="cosine",
+    use_instuctions=True,
+    reference="https://huggingface.co/intfloat/multilingual-e5-large-instruct",
+    n_parameters=560_000_000,
+    memory_usage=None,
+    embed_dim=1024,
+    license="mit",
+    max_tokens=514,
```
orionw, line 77: Is this true? 514 seems like an odd number.

mteb/models/promptriever_models.py
```diff
@@ -55,6 +55,15 @@ def loader_inner(**kwargs: Any) -> Encoder:
...
     revision="01c7f73d771dfac7d292323805ebc428287df4f9-30b14e3813c0fa45facfd01a594580c3fe5ecf23",  # base-peft revision
     release_date="2024-09-15",
+    n_parameters=7_000_000,
+    memory_usage=None,
+    max_tokens=None,
+    embed_dim=None,
+    license="apache-2.0",
+    reference="https://huggingface.co/samaya-ai/promptriever-llama2-7b-v1",
+    similarity_fn_name="cosine",
+    framework=["PyTorch"],
```
orionw, line 65: Is Tevatron (though also PyTorch).  

Max tokens is 4096 and embed dim 4096. 

This is true for all Promptriever models (except the llama3 one which is 8k).

mteb/models/repllama_models.py
```diff
@@ -144,6 +144,15 @@ def loader_inner(**kwargs: Any) -> Encoder:
...
     release_date="2023-10-11",
+    n_parameters=7_000_000,
+    memory_usage=None,
+    max_tokens=None,
+    embed_dim=None,
+    license="apache-2.0",
+    reference="https://huggingface.co/samaya-ai/castorini/repllama-v1-7b-lora-passage",
+    similarity_fn_name="cosine",
+    framework=["PyTorch"],
+    use_instuctions=True,
```
orionw, line 155: No instruction for this one, although it has the query prefix and passage prefix (do those count?)

And 4096 embedding and max tokens. Also tevatron training code with Pytorch.  Same for both of the RepLlama models.

mteb/models/e5_instruct.py
Samoed: Yes, this is true. I think that this is 512 + 2 (query and passage)

mteb/models/cohere_models.py
```diff
@@ -93,6 +93,7 @@ def encode(
     license=None,
     similarity_fn_name="cosine",
     framework=[],
```
Samoed, line 95:
```suggestion
    framework=["API"],
```

mteb/models/cohere_models.py
```diff
@@ -113,4 +114,5 @@ def encode(
...
```
Samoed, line 116:
```suggestion
    framework=["API"],
```

x-tabdeveloping (approved):
I think we should merge this soon, so I can build on it and then increase incremental changes as we go along. This is a huge help with the leaderboard already.

mteb/model_meta.py
x-tabdeveloping: Maybe also AnglE?

KennethEnevoldsen:
> Yeah I would say non-zero-shot by default and people have to opt in. I'm fine with taking people's word for it (e.g. if OpenAI said it, we could probably trust them), although we could an enforce an open-data clause.

We could add an asterix. For cases where training data is unknown. Something like:

text-embed-large*

*Training data is unknown. 

This would encourage open-data.

> I think we should merge this soon, so I can build on it and then increase incremental changes as we go along. This is a huge help with the leaderboard already.

Agree will have this merged on Monday at the latest.

mteb/models/promptriever_models.py
KennethEnevoldsen: added

mteb/models/repllama_models.py
KennethEnevoldsen: added

KennethEnevoldsen: > No instruction for this one, although it has the query prefix and passage prefix (do those count?)

I have counted it as an instruction yes

KennethEnevoldsen:
Added the fixes:

changed open source to open weights.
Added public_training_data, public_training_code, and zero_shot_benchmarks (though I did not fill it out).

PR #1319, opened 2024-10-24
Add jina, uae, stella models

Samoed:
## Checklist

### Adding a model checklist


mteb/models/jina_models.py
```diff
@@ -0,0 +1,127 @@
...
+        if CURRENT_SENTENCE_TRANSFORMERS_VERSION < MIN_SENTENCE_TRANSFORMERS_VERSION:
+            raise RuntimeError(
+                f"sentence_transformers version {st_version} is lower than the required version 3.1.0"
+            )
+        try:
+            import einops  # noqa: F401
+        except ImportError:
+            raise ImportError(
+                "To use the jina-embeddings-v3 models `einops` is required. Please install it with `pip install einops`."
+            )
```
bwanglzu, line 47: just a small feedback, install flash attention will make evaluation much faster, maybe it's worth to give a logging message

mteb/models/jina_models.py
```diff
@@ -0,0 +1,127 @@
...
+            "STS": "text-matching",
+            "PairClassification": "text-matching",
+            "BitextMining": "text-matching",
+            "MultilabelClassification": "classification",
+            "Reranking": "separation",
+            "Summarization": "text-matching",
+        },
+    ),
+    name="jinaai/jina-embeddings-v3",
+    languages=["eng_Latn"],
```
bwanglzu, line 123: you can adopt the XLM-R language code in e5instruct, we continue training the original XLM-R with slight different languages, but it's a subset of XLM-R

mteb/models/jina_models.py
```diff
@@ -0,0 +1,127 @@
...
+    open_source=True,
+    revision="215a6e121fa0183376388ac6b1ae230326bfeaed",
+    release_date="2024-09-05",  # initial commit of hf model.
```
bwanglzu, line 126:
```suggestion
    release_date="2024-09-18",  # official release date
```

mteb/models/jina_models.py
```diff
@@ -0,0 +1,127 @@
...
+        JinaWrapper,
+        model="jinaai/jina-embeddings-v3",
+        revision="215a6e121fa0183376388ac6b1ae230326bfeaed",
+        model_prompts={
+            "Retrieval-query": "retrieval.query",
+            "Retrieval-passage": "retrieval.passage",
+            "Clustering": "separation",
+            "Classification": "classification",
+            "STS": "text-matching",
+            "PairClassification": "text-matching",
```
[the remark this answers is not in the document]
Samoed: My results on PairClassification didn't match. Thank you! I'll reeval my results

mteb/models/jina_models.py
```diff
@@ -0,0 +1,127 @@
...
+            "PairClassification": "classification",
+            "BitextMining": "text-matching",
+            "MultilabelClassification": "classification",
+            "Reranking": "separation",
+            "Summarization": "text-matching",
+        },
+    ),
+    name="jinaai/jina-embeddings-v3",
+    languages=["eng_Latn"],
+    open_source=True,
```
bwanglzu, line 124: i think we need an extra `trust_remote_code=True` here in the loader

bwanglzu, line 124: maybe it's nice to add a small comment, making sure ppl aware that it's an open-weights model, for example:
```suggestion
    open_source=True, # cc-by-nc-4.0
```

Samoed: There is addional field in modelmetdata. I will update metadata a bit later

bwanglzu:
the rest looks good to me, need to run some check to make sure different task adapters (especially retrieval), `task` and `prompt_name` is correctly passed and can reproduce our reported results, i'm running some small testing

Samoed:
I have results. I will paste them soon (currently creating table to compare easily). @bwanglzu Thank you very much!

mteb/models/jina_models.py
Samoed: It will be added in `__init__` of `SentenceTransformerWrapper`

bwanglzu:
perfect thanks @Samoed ! seems our reported on Emotion is lower than what we actually have (lol). 

do you mind to share me your script so that i can run a few more experiments?

Samoed:
[Here is my code](https://www.kaggle.com/samoed/fork-mteb-run)

Samoed:
> BTW some of our reported score might be comes from a smaller context length such as 512

What do you mean? I think that mteb using SentenceTransformer uses all context length

bwanglzu:
> What do you mean? I think that mteb using SentenceTransformer uses all context length

i mean when we submit scores, their might be a small chance it is being submitted by different ppl in the team which utilise slightly different max sequence length, sometimes to speed up evaluation we use 512, sometimes we use full context length which is 8192.

mteb/models/jina_models.py
KennethEnevoldsen: I would set it as a requirement.

I would also specify:

```pip install mteb[jina]```
for installing dependencies to ensure that the required dependencies are specified.

File: mteb/models/e5_instruct.py [hunk 6]
```diff
@@ -46,7 +46,11 @@ def encode(
                 )
             if instruction:
                 kwargs["instruction"] = e5_instruction(instruction)
-            return super().encode(sentences, *args, **kwargs)
+            embeddings = super().encode(sentences, *args, **kwargs)
+            if isinstance(embeddings, torch.Tensor):
+                # sometimes in kwargs can be return_tensors=True
+                embeddings = embeddings.cpu().detach().float().numpy()
```
KennethEnevoldsen, line 52: Should we move this normalization somewhere else? (I believe it used to be in the model_encode function, but it might be nice to have a normalization function.

mteb/models/jina_models.py
```diff
@@ -0,0 +1,235 @@
...
+
+        Args:
+            sentences: The sentences to encode.
+            task_name: The name of the task. Sentence-transformers uses this to
+                determine which prompt to use from a specified dictionary.
+            prompt_type: The name type of prompt. (query or passage)
+            **kwargs: Additional arguments to pass to the encoder.
+
+            The order of priorities for prompt selection are:
+                1. Composed prompt of task name + prompt type (query or passage)
```
KennethEnevoldsen, line 170: is this relevant to have here?

mteb/models/jina_models.py
KennethEnevoldsen: Is this default behavior? (it probably shouldn't be)

mteb/models/jina_models.py
```diff
@@ -0,0 +1,235 @@
...
+    ),
+    name="jinaai/jina-embeddings-v3",
+    languages=XLMR_LANGUAGES,
+    open_source=True,
+    revision="215a6e121fa0183376388ac6b1ae230326bfeaed",
+    release_date="2024-09-18",  # official release date
+    n_parameters=572_000,
+    max_tokens=8194,
+    embed_dim=4096,
+    license="mit",
```
KennethEnevoldsen, line 232:
```suggestion
    license="cc-by-nc-4.0",
```

KennethEnevoldsen (approved):

mteb/models/jina_models.py
Samoed: Yes, this is default. I will change that

mteb/models/jina_models.py
Samoed: Mostly not, I copied this from wrapper

mteb/models/e5_instruct.py
Samoed: I don't know where to put it. Most of the tasks return numpy embeddings and this can be removed, but in STS somehow passed `return_tensors=True`. I tried to debug it, but couldn't find from where it passes

KennethEnevoldsen: hmm will try do a debug on this tomorrow or monday

mteb/models/jina_models.py
KennethEnevoldsen: let us remove it then (also from the other encode)

mteb/models/jina_models.py
Samoed: Do you mean to include both `flash attention` and `einops`? I have some concerns about `flash attention` because it needs to be installed with `pip install flash-attn --no-build-isolation`, and it also requires CUDA 12.1

KennethEnevoldsen: Ahh, yeah, right. Then I would probably throw an info message for it and separate it out to its own optional dependency.

mteb/models/e5_instruct.py
Samoed: I found it #1325!

pyproject.toml
```diff
@@ -59,6 +59,8 @@ speedtask = ["GPUtil>=1.4.0", "psutil>=5.9.8"]
 peft = ["peft>=0.11.0"]
 leaderboard = ["gradio>=4.44.0", "gradio_rangeslider>=0.0.6"]
 flagembedding = ["FlagEmbedding"]
+jina = ["einops"]
+flash_attention = ["flash-attn"]
```
KennethEnevoldsen, line 63: Let us add lower bounds on versions.

KennethEnevoldsen (approved):
Only a minor thing otherwise all good

pyproject.toml
Samoed: @KennethEnevoldsen Done!

Samoed:
@KennethEnevoldsen Is this PR ready for merge?

KennethEnevoldsen (approved):
This looks good! Very happy to have it merged in

Issue #1311, opened 2024-10-23
FSTimeoutError for MLSUMClusteringP2P & MLSUMClusteringS2S


bourdoiscatie:
Hi! 

I've just trained an embedding model in French and would like to test it on the MTEB_FR.
I used the following code:
```
import mteb

benchmark = mteb.get_benchmark("MTEB(fra)")
evaluation = mteb.MTEB(tasks=benchmark)
evaluation.run(my_model, eval_splits=["test"], output_folder=f"results")
```

and everything ran fine until `MLSUMClusteringP2P`, where I got the following error:
```
---------------------------------------------------------------------------
TimeoutError                              Traceback (most recent call last)
File ~/.local/lib/python3.12/site-packages/fsspec/asyn.py:56, in _runner(event, coro, result, timeout)
     55 try:
---> 56     result[0] = await coro
     57 except Exception as ex:

File ~/.local/lib/python3.12/site-packages/fsspec/implementations/http.py:254, in HTTPFileSystem._get_file(self, rpath, lpath, chunk_size, callback, **kwargs)
    253 while chunk:
--> 254     chunk = await r.content.read(chunk_size)
    255     outfile.write(chunk)

File /usr/lib/python3.12/site-packages/aiohttp/streams.py:393, in StreamReader.read(self, n)
    392 while not self._buffer and not self._eof:
--> 393     await self._wait("read")
    395 return self._read_nowait(n)

File /usr/lib/python3.12/site-packages/aiohttp/streams.py:311, in StreamReader._wait(self, func_name)
    310 try:
--> 311     with self._timer:
    312         await waiter

File /usr/lib/python3.12/site-packages/aiohttp/helpers.py:713, in TimerContext.__exit__(self, exc_type, exc_val, exc_tb)
    712 if exc_type is asyncio.CancelledError and self._cancelled:
--> 713     raise asyncio.TimeoutError from None
    714 return None

TimeoutError: 

The above exception was the direct cause of the following exception:

FSTimeoutError                            Traceback (most recent call last)
Cell In[9], line 5
      3 benchmark = mteb.get_benchmark("MTEB(fra)")
      4 evaluation = mteb.MTEB(tasks=benchmark)
----> 5 evaluation.run(model, eval_splits=["test"], output_folder=f"results")

[...]

PR #1322, opened 2024-10-24
Re-upload dataset to hub to avoid using script upload

imenelydiaker:
This should solve issue #1311.
I re-uploaded the dataset to parquet format here: https://huggingface.co/datasets/mteb/mlsum as suggested by @lhoestq, to avoid using the script generation for loading.

### DIFF
--- a/mteb/benchmarks/benchmarks.py
+++ b/mteb/benchmarks/benchmarks.py
@@ -63,6 +63,7 @@ class Benchmark:
             base_results = load_results()
         return base_results.select_tasks(self.tasks)
 
+
 MTEB_MAIN_EN = Benchmark(
     name="MTEB(eng)",
     tasks=get_tasks(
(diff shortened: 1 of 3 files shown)


mteb/benchmarks/benchmarks.py
```diff
@@ -63,6 +63,7 @@ def load_results(
             base_results = load_results()
         return base_results.select_tasks(self.tasks)
 
+
```
isaac-chung, line 66: nit: no-op, maybe can remove?

isaac-chung (approved):
Looks good, as long as you've tried out at least one model on the new dataset.

mteb/benchmarks/benchmarks.py
imenelydiaker: It's the linter 😅

imenelydiaker:
> Looks good, as long as you've tried out at least one model on the new dataset.

Yes I tested it locally with average_kominos model. Let's merge?

KennethEnevoldsen:
Great fix, @imenelydiaker!

Issue #941, opened 2024-06-17
Standardising matrix arrays (adopt Array API spec)


KennethEnevoldsen:
Currently MTEB does not consistently use a ground truth for vector manipulation. A solution would be to adopt the [Array API spec](https://data-apis.org/array-api/latest/index.html). A potentially simpler solution is to use either torch or numpy.

- The benefit of using numpy is that it is very stable and rarely leads to OOM and related GPU issues.
- The benefit of using PyTorch is that it potentially allows evaluation using GPUs (currently, sklearn does not support this, but might in the future, when they adopt they API spec)

partly related to the discussion in: https://github.com/embeddings-benchmark/mteb/issues/883

This is not a pressing concern atm.

Muennighoff:
Good point; I think using numpy would also allow us to make torch an optional dependency which could be nice as I assume some models do not require it so people have to install it just for mteb

KennethEnevoldsen:
#2234 and #1791 both improve tensor compatibility allowing models to output torch tensors and where possible also used operations that work for both numpy and torch arrays.

This is a list of some of the current outstanding issues:

1) Unsqueeze not in array API spec, we should use expand_dims, but torch does not implement this (https://github.com/pytorch/pytorch/issues/56774). We can use reshape instead.
2) What to do with torch.functional? We would have to reimplent these to make it work. I would be afraid that this could introduce unintended bugs. 

I suspect that these change will be easier once the array spec is more widely adopted.

KennethEnevoldsen:
I found a few places where we could probably remove the `np.asarray` conversion (e.g. in PairClassification)

PR #1329, opened 2024-10-26
WIP: Merge main into MIEB

KennethEnevoldsen:
Merge of main into MIEB. This includes
- [x] the initial merge
- [x] resolving conflict (minor)
- [x] formatting (minor)
- [x] Fix bugs occurring from `pytest --collect-only`, these were mainly import errors and updating metadata (removing socio eco., form, and standardizing license)
- [x] fix bugs due to `make test`, notably only wrap SentenceTransformers and crossEncoders with the SentenceTrf. wrapper. @Samoed I see that this also leads to a custom exception in `is_cross_encoder_compatible`, I think it would be best to avoid that. A solution might be to create CrossEncoderWrapper as a subclass of the existing wrapper. To do this I also added the `normalize_embeddings_to_numpy` function though #941 would be a much better fix.

From this, I see a few additional stuff that we need to do:

In this PR:

- missing license for CLEVR, SciMMIR (I have temp. set it to not specified)

not in this PR:

- Merge docs together and update main docs (it is important not to confuse users with the new additions)
- figure out what to do with folder structure. Should we create a folder like `evaluation/evaluators/{modality},` or do we want to do it in another way? I also have some concerns re. evaluators (as I would like them to be the same across text and images, but I see that it might not be possible)
- remove `results-mieb`
- Potentially add relevant tests
- Define an interface for MIEB encoders

## Checklist

Reviewers:
- @isaac-chung for integration
- @gowitheflow-1998 for ensuring that I don't break MIEB
- @Samoed for the changes listed above


mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -181,11 +182,13 @@ def _evaluate_subset(
...
 
-        X_test = model.encode(
-            test_text,
-            model=model,
-            task_name=self.metadata.name,
-            **encode_kwargs,
+        X_test = normalize_embeddings_to_numpy(
+            model.encode(
+                test_text,
+                model=model,
```
Samoed, line 188: I accidentally left `model=model` here, and it should be removed.

Samoed (approved):
Looks great! Maybe then remove from wrappers convert from tensors to numpy?

mteb/evaluation/MTEB.py
```diff
@@ -354,7 +353,7 @@ def run(
 
         meta = self.create_model_meta(model)
         output_path = self.create_output_folder(meta, output_folder)
-        if not isinstance(model, Wrapper):
+        if isinstance(model, (SentenceTransformer, CrossEncoder)):
```
Samoed, line 356: What will happen if model will be wrapped in `CachedEmbeddingWrapper`?

mteb/abstasks/AbsTaskMultilabelClassification.py
KennethEnevoldsen: removed

isaac-chung (approved):
I don't have a good way to review 541 changed files, other than to rely on tests, which needs expanding to cover the rest of the MIEB additions (e.g. models, abstasks). So for now I'm totally fine with this PR. 

> not in this PR:
> - Merge docs together and update main docs (it is important not to confuse users with the new additions)
> - figure out what to do with folder structure. Should we create a folder like evaluation/evaluators/{modality}, or do we want to do it in another way? I also have some concerns re. evaluators (as I would like them to be the same across text and images, but I see that it might not be possible)
> - remove results-mieb
> - Potentially add relevant tests
> - Define an interface for MIEB encoders

Good shout on these to-dos. We could put them in a separate issue for tracking / discussion?

KennethEnevoldsen:
> I don't have a good way to review 541 changed files, other than to rely on tests, which needs expanding to cover the rest of the MIEB additions (e.g. models, abstasks). So for now I'm totally fine with this PR.

Yep def. hard to do. It was also mostly to check if the steps I followed seems reasonable.

I assume it is fine to merge this in. Will create issues for the pointers (@gowitheflow-1998 will you have a look at the comment on licenses)

gowitheflow-1998:
@KennethEnevoldsen looks good to me! Def agree with the future tasks regarding merging docs, results, adding tests and encoder interfaces. I haven't thought of good ways to make evaluators the same across modalities, though we could potentially use the idea of `Any2AnyRetrieval` as a start to write unified evaluators. e.g., the current `Any2AnyRetrieval` in `mieb` can as well support text-to-text only retrieval. We just need to adjust the code a bit to support the encoding functions of text encoders. For now separate folders is likely the easiest!

License: will update in next commits. Think CLEVR is CC BY 4.0.

PR #1341, opened 2024-10-28
feat: update English benchmarks and mark MMTEB benchmarks as beta

KennethEnevoldsen:
Seems like the original english benchmark have been overwritten with mteb(eng), which seems to have been a mistake. I have instead renamed it MTEB(eng, classic) and reverted to the original list.

I have additionally added MTEB(eng, beta) for the new MMTEB benchmark and marked MMTEB benchmarks as "beta" as well. This makes it clear that they are under development (which I believe is the best way to do it now).

## Checklist

### DIFF
--- a/mteb/benchmarks/benchmarks.py
+++ b/mteb/benchmarks/benchmarks.py
@@ -62,8 +62,60 @@ class Benchmark:
         return base_results.select_tasks(self.tasks)
 
 
-MTEB_MAIN_EN = Benchmark(
-    name="MTEB(eng)",
+MTEB_EN = Benchmark(
+    name="MTEB(eng, beta)",
+    tasks=get_tasks(
+        tasks=[
+            "AmazonCounterfactualClassification",
+            "ArguAna",
+            "ArXivHierarchicalClusteringP2P",
+            "ArXivHierarchicalClusteringS2S",
+            "AskUbuntuDupQuestions",
+            "BIOSSES",
+            "Banking77Classification",
+            "BiorxivClusteringP2P.v2",
+            "CQADupstackGamingRetrieval",
+            "CQADupstackUnixRetrieval",
+            "ClimateFEVERHardNegatives",
+            "FEVERHardNegatives",
+            "FiQA2018",
+            "HotpotQAHardNegatives",
+            "ImdbClassification",
+            "MTOPDomainClassification",
+            "MassiveIntentClassification",
+            "MassiveScenarioClassification",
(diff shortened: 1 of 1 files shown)


mteb/benchmarks/benchmarks.py
```diff
@@ -556,7 +609,7 @@ def load_results(
 
 
 MTEB_multilingual = Benchmark(
-    name="MTEB(Multilingual)",
+    name="MTEB(Multilingual, beta)",
```
x-tabdeveloping, line 612: Make sure to change the `default_benchmark` to this in the leaderboard.

mteb/benchmarks/benchmarks.py
```diff
@@ -734,7 +787,7 @@ def load_results(
 
 
 MTEB_INDIC = Benchmark(
-    name="MTEB(indic)",
+    name="MTEB(indic, beta)",
```
x-tabdeveloping, line 790: Is there a reason, why indic shouldn't be capitalized? Do we have a unified scheme for naming benchmarks?

x-tabdeveloping (commented):
Looks good, left a few comments

mteb/benchmarks/benchmarks.py
isaac-chung: That was my bad, should have been `Indic`

mteb/benchmarks/benchmarks.py
```diff
@@ -62,8 +62,61 @@ def load_results(
         return base_results.select_tasks(self.tasks)
 
 
-MTEB_MAIN_EN = Benchmark(
```
isaac-chung, line 65: This var is referenced in 2 other files: `docs/adding_a_model.md` and `mteb/__init__.py`. We might want to reuse that for either the MMTEB one or the classic one.

isaac-chung (commented):
Looks good. One small thing in addition to Márton's comments.

mteb/benchmarks/benchmarks.py
KennethEnevoldsen: Fixed

mteb/benchmarks/benchmarks.py
KennethEnevoldsen: fixed

mteb/benchmarks/benchmarks.py
KennethEnevoldsen: Fixed

PR #1345, opened 2024-10-28
Leaderboard updates: Model meta + task and benchmark info

x-tabdeveloping:
Tasks:
 - [x] Add Benchmark description and citation
 - [x] Add model info to table
 - [x] Add model filtering based on model meta
 - [x] Add task information table
 - [x] Replace mean rank with Borda rank

### DIFF
--- a/mteb/leaderboard/table.py
+++ b/mteb/leaderboard/table.py
@@ -1,14 +1,25 @@
 from __future__ import annotations
 
+import math
+
 import gradio as gr
 import numpy as np
 import pandas as pd
 
+from mteb.models.overview import get_model_meta
 from mteb.overview import get_task
 
 
 def format_scores(score: float) -> float:
     return score * 100
+
+
+def format_n_parameters(n_parameters) -> str:
+    n_million = int(n_parameters) // 1e6
+    n_zeros = math.log10(n_million)
+    if n_zeros >= 3:
+        return str(n_million // (10**3)) + "B"
+    return str(n_million) + "M"
 
 
 def scores_to_tables(scores_long: list[dict]):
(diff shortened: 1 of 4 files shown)


x-tabdeveloping:
@isaac-chung yes!!

```
pip install mteb
git clone https://github.com/embeddings-benchmark/mteb.git
cd mteb
git checkout leaderboard_2
```
Then run this:
```python
from mteb.leaderboard import demo

demo.launch()
```

mteb/leaderboard/table.py
```diff
@@ -1,16 +1,42 @@
...
 
+from mteb.models.overview import get_model_meta
 from mteb.overview import get_task
 
 
+def borda_count(scores: pd.Series) -> pd.Series:
+    n = len(scores)
+    ranks = scores.rank(method="average", ascending=False)
+    counts = n - ranks
+    return counts
```
KennethEnevoldsen, line 17: Yep - The only problem now is ties, which we calculate slightly differently in the paper (but I don't believe there are practically any ties), so I think it is equivalent so let us keep it as is.

KennethEnevoldsen (approved):
Looks good!

mteb/leaderboard/table.py
x-tabdeveloping: I think I checked and this piece of code is equivalent to tournament borda. Is that not what we used in the paper?

KennethEnevoldsen: I don't think it is equivalent in the case of ties (but there are no ties so equivalent in practice). We should just remove the tournament borda description from the paper.

Issue #1066, opened 2024-07-10
Distinguishing instructions for RAR-b


Muennighoff:
Have been discussing with @gowitheflow-1998 how to distinguish results w/ & w/o instructions for RAR-b in the result files & the LB. Two ideas:

1. We differentiate at the model level. E.g. we have a `GritLM-7B` & a `GritLM-7B-noinstruct` folder in the [results](https://github.com/embeddings-benchmark/results) repo. For models that are no instruct by default, we'd have to add an instruct folder e.g. `text-embedding-3-large` & `text-embedding-3-large-instruct`.
  - Works out of the box when running mteb without overwriting results from the other setup by just specifying the `results_folder` via the kwargs. 
  - Generalizes nicely to other tasks - e.g. if you want to benchmark GritLM-7B on MTEB or FollowIR without instructions you just put those results in `GritLM-7B-noinstruct`.
  - For the LB we could consider adding a `Model types` filter that says `instructions` to filter for models that do/do not use instructions. Maybe that way we don't need two tabs anymore for `RAR-b` & `RAR-b No instructions` but just put all of them in the same `RAR-b` tab and then people can filter out for instruction usage similar to how one can filter out Cross Encoders for the FollowIR tab.
2. We differentiate at the task level. E.g. the files are named `RARbMath.json` for instructions & `RARbMathNoInstruct.json` for no instructions.
  - It would not work nicely out of the box as it would interfere with previously written results unless a new `results_folder` is specified. But then you still need to rename them and move them back into the same folder which makes it more complex than 1.
  - Can also generalize to other tasks but requires renaming json files which is a bit more complex than placing it in a different folder. We could have an automatic script for that though but still a bit clunky.
  - Adds some complexity for the LB because we'd break the behavior that the filename equals the taskname (e.g. the task would likely still be called `RARbMath`). Also does not work out of the box with our `Model types`-based filtering and would not allow showing e.g. GritLM-7B no instruction results in the main mteb tab.

So I suggest we go with 1. but lmk if you disagree! Curious about your thoughts @orionw @gowitheflow-1998! 😊

orionw:
Thanks for kicking off the discussion @Muennighoff @gowitheflow-1998!  I agree we should start updating the leaderboard!

[...]

Issue #1325, opened 2024-10-25
RetrivalEvaluator overrides encode kwargs


Samoed:
When developing the wrappers, I tried to remove the explicit conversion from tensor to NumPy, but I encountered errors due to the `convert_to_tensor` parameter in `encode_kwargs`. I later found that `RetrievalEvaluator` overrides these parameters and sets `batch_size`. Maybe this should be changed?

https://github.com/embeddings-benchmark/mteb/blob/650e8b89ba856b21869e3fa6d5921ed7d9ed8d07/mteb/evaluation/evaluators/RetrievalEvaluator.py#L73-L78

KennethEnevoldsen:
ahh, yes, these were added for backward compatibility when converting to encode_kwargs (#1030).

As you see in the diff, there are multiple of these, especially for batch size. For now, it is easy to just remove `convert_to_tensor`. However, it seems like the other two are being referred to in the code in multiple places. Not sure what the best choice is for these.

Samoed:
I think that `batch_size` and `convert_to_tensor` can be removed and `show_progress_bar` can be passed directly to `model.encode`. I don't think that removing these kwargs will break something

KennethEnevoldsen:
Removing batch size completely could lead to sudden OOM for many users, and it is also used in a few places in the evaluators.

Samoed:
The current implementation often results in OOM errors because it sets the batch size too high. By default (if not specified), SentenceTransformer uses a batch size of 32. Another solution would be to explicitly set a default batch size

KennethEnevoldsen:
I am all for removing it completely (it is really a model concern, not an evaluation concern)

Samoed:
Fixed in https://github.com/embeddings-benchmark/mteb/pull/2495

PR #1359, opened 2024-10-29
Consolidate Retrieval/Reranking/Instruction Variants

orionw:
This PR would consolidate much of the logic between reranking/retrieval/instruction variants. 

In principal, the only difference between these retrieval and reranking is that reranking is a bunch of "mini" retrieval tasks where the corpus is set by a different list. 

The only difference between reranking and instruction reranking is the presence of an instruction, so this PR also merges those together also (we will have to designate instruction tasks via some list, but we need to do that anyways for #1066 and others)

Benefits are:
- Enables reranking to use the full set of retrieval metrics, not just MAP and MRR. 
- Consolidates all the retrieval-style code (reranking, retrieval, instruction-based variants) hopefully making it easier to maintain and reducing duplicate code (e.g. we won't have three different places for retrieval eval code)
- Makes it possible to use cross-encoders on reranking tasks (#1213)
- Allows for instruction-based reranking tasks

Potential concerns/todos:
- Using the same implementation for both retrieval and reranking is less computationally efficient since we don't batch over all queries at once. It should be roughly similar though on the corpus side, since we're still batching. It depends on if we're okay with this though @KennethEnevoldsen @Muennighoff @imenelydiaker @samoed (and please tag others interested in reranking)? I don't remember reranking being particular compute intensive so perhaps this trade off is worth it? If not, I can revert some of this.
- I still need to remove AbsTaskInstructionRetrieval and put the extras into the AbsTaskRetrieval class (+ adding some new datasets), but I thought I would leave that for the next PR.

If people are okay with this approach, I'll go through and verify that all the Reranking tasks still work and reproduce score-wise.

## Checklist

- [] Run tests locally to make sure nothing is broken using `make test`. 


mteb/abstasks/AbsTaskReranking.py
```diff
@@ -1,98 +1,216 @@
...
-
-class AbsTaskReranking(AbsTask):
-    """Abstract class for re-ranking experiments.
-
+from ..load_results.task_results import ScoresDict
+from .AbsTaskRetrieval import AbsTaskRetrieval
+
+logger = logging.getLogger(__name__)
+
+OLD_FORMAT_RERANKING_TASKS = [
```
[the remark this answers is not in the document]
KennethEnevoldsen: I think for the commonly used reranking datasets converting the HF dataset is probably reasonable. Otherwise, I think this is fine.

mteb/abstasks/AbsTaskReranking.py
```diff
@@ -1,98 +1,216 @@
...
+        self.top_ranked: dict[str, dict[str, list[str]]] or dict[str, dict[str, dict[str, float]]]
+            Semantically, it should contain dict[split_name, dict[sample_id, list[doc_id]]] or dict[split_name, dict[sample_id, dict[doc_id, score]]]
+            E.g.: {"test": {"q1": ["document_one", "document_two"]}} or {"test": {"q1": {"document_one": 1, "document_two": 0.5}}}
+
+    Old Format:
+    -----------
     self.load_data() must generate a huggingface dataset with a split matching self.metadata_dict["eval_splits"], and assign it to self.dataset. It must contain the following columns:
         query: str
         positive: list[str]
         negative: list[str]
```
KennethEnevoldsen, line 64: Do we want people to use this? In which case I would probably just remove it.

mteb/abstasks/AbsTaskReranking.py
```diff
@@ -1,98 +1,216 @@
...
+    def load_data(self, **kwargs):
+        if self.data_loaded:
+            return
+
+        if self.metadata.name in OLD_FORMAT_RERANKING_TASKS:
+            self.dataset = datasets.load_dataset(**self.metadata_dict["dataset"])  # type: ignore
+            self.dataset_transform()
+        else:
+            # use AbsTaskRetrieval default to load the data
+            # TODO: need to make sure top_ranked comes back
```
KennethEnevoldsen, line 79: ?

mteb/abstasks/AbsTaskReranking.py
```diff
@@ -1,98 +1,216 @@
...
+
+        for i, neg_doc in enumerate(sorted(negative_docs)):
+            doc_id = f"{query_id}_negative_{i}"
+            example_data["doc_ids"].append(doc_id)
+            example_data["doc_texts"].append(neg_doc)
+            example_data["relevance_scores"].append(0)
+
+        return example_data
+
+    def dataset_transform(self):
```
KennethEnevoldsen, line 113: wouldn't this be overwritten by the tasks own dataset transform?

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
```diff
@@ -65,7 +65,7 @@ class MIRACLReranking(MultilingualTask, AbsTaskReranking):
         modalities=["text"],
         eval_splits=[_EVAL_SPLIT],
         eval_langs=_LANGUAGES,
-        main_score="NDCG@10(MIRACL)",
+        main_score="ndcg_cut_10",
```
KennethEnevoldsen, line 68: Question: What happened to the MIRACL retrieval and rerankers?

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
```diff
@@ -88,7 +88,8 @@ def _evaluate_subset(
         encode_kwargs: dict[str, Any] = {},
         **kwargs: Any,
     ) -> ScoresDict:
-        evaluator = RerankingEvaluator(
+        # TODO: this file will need to be fixed
```
KennethEnevoldsen, line 91: ?

KennethEnevoldsen (commented):
Love the PR. I think the time concern is a reasonable trade-off for simplicity and maintainability, but @Muennighoff might know more about this given he made the original choice to split the two.

Have we checked that the new rerankers produce similar scores?

Samoed:
Currently Retrieval tasks can't load multilingual datasets and I think this PR good place how to standardize this

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
orionw: Yes, still there - MIRACL needs a more significant change to convert to the new format, so I wanted to make sure we were okay with this direction before I spent more time changing everything.

`NDCG@10(MIRACL)` will be replaced with `ndcg_cut_10`, AFAIK they are the same but the previous reranking setup didn't compute NDCG.

mteb/abstasks/AbsTaskReranking.py
orionw: Ah shoot, good point. Okay, will need to rename this function, thanks!

mteb/abstasks/AbsTaskReranking.py
orionw: Sorry, left this as a note to myself but it brings up a good question for you. We currently have `HFDataLoader` which loads the Retrieval data.  But we could combine it with the InstructionHFDataLoader and also have it do the reranking loading.  It would just need to have options to (1) load an instruction config and (2) load a top_ranked config for reranking. 

Does that seem okay to you? It get a bit more complex.  I could also extend and make a new class, keeping as much as possible the same between them

orionw:
Thanks for the comments @KennethEnevoldsen and @Samoed! I'll go ahead and finish validating these / fixing MIRACL.

>  I think this PR is a good opportunity to standardize this and re-upload them, as retrieval datasets have different loaders. 

@Samoed, when you say standardize, what do you mean? I think each task usually defines their way of loading right? Unless you mean have a default loader for this in the main codebase, so each file doesn't have to define their own custom loader?

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
KennethEnevoldsen: We should probably test that to make sure.

mteb/abstasks/AbsTaskReranking.py
orionw: I was thinking something like checking if the dataset has a config called "instructions" or "top_ranked" and if it does to load them. Agree we don't want to use `trust_remote_code`!

mteb/tasks/Reranking/multilingual/MIRACLReranking.py
orionw: Absolutely! I plan to test all the reranking scores, want to make sure they're consistent.

mteb/abstasks/AbsTaskReranking.py
KennethEnevoldsen: Sounds reasonable

orionw:
I was finding a small delta between some results and finally narrowed it down. In the current code for reranking we have:
```
    self.samples = [
        sample
        for sample in self.samples
        if len(sample["positive"]) > 0 and len(sample["negative"]) > 0
    ]
```

But `pytrec_eval` allows it, and handles this differently - calculating it as either 1 or 0.  Thus there is a slight mismatch between the number of instances in the dataset and their scores.  

Since we already have many existing results for reranking, I have added this flag to the dataset transformation for the older datasets, but not for new datasets. 

-----------


I also confirmed that MIRACL nDCG@10 is the same, see the call is the same:
```
    ndcg, _map, recall, precision, naucs = RetrievalEvaluator.evaluate(
        qrels=qrels,
        results=results,
        k_values=self.k_values,
        ignore_identical_ids=False,
    )
```
----------------
The only weird and potentially unresolved thing, is that the existing code calls:

```

**Comment by orionw** [comment]
An update on this. Everything is done, except for MindSmall, but perhaps I need to make another large change to fit it.

The reranking code has:

```
    for instance in self.samples:
        num_subqueries = (
            len(instance["query"]) if isinstance(instance["query"], list) else 1
        )
        query_emb = all_query_embs[query_idx : query_idx + num_subqueries]
```

```
query_emb: Query embedding of shape `(num_queries, hidden_size)`)
                if `num_queries` > 0: we take the closest document to any of the queries
 ```
 
where these subqueries only applies to the MindSmall reranking task. 

The two scores (with or without subqueries) are actually pretty close if you ignore this step (only 0.1% off), but since it is in MTEB-original I wanted to make sure it hit parity.
 
The problem is that MindSmall has thousands of duplicate entries. So because my change makes it the "mini"-retrieval setting where each query sets up a mini retrieval task, it re-computes all the duplicates.  With all the duplicates (1000x), it's simply not feasible to run the task. 
 
 I considered a solution where (like we do currently) we compute a unique set of queries and documents, but then share the retrieval evaluation code. However, this will compute the dot product on a lot more query-doc pairs than we need to.
 
I'm not sure I see a solution, so I will probably re-add the reranking evaluator and try to have it share as much code as possible with the retrieval one. 

If anyone has other suggestions please let me know!

KennethEnevoldsen:
re. MindSmall:

I think the best solution is to duplicate the evaluation code over to mindsmall reranking (and then leave it as is). We can then create a v2 of the mindsmall both without the duplicates and using the standardized code. We can then update MTEB(eng, beta) to use v2.

orionw:
Okay, I think this is finally ready for review. Sorry it ended up so big, it was hard to break this into pieces.

A list of things changed:
- Deleted `InstructionRetrievalEvaluator`, `RerankingEvaluator`, `AbsTaskInstructionRetrieval` and severely reduced `AbstractTaskReranking`
- Made all retrieval-based tasks take the same main three configs: `corpus`, `queries`, `relevant_docs` with options for `instructions` (query-id->str) and `top_ranked` (which is reranking, query-id -> list of strings)
- Removed a lot of older code around instructions, such as the `keywords`, `short_instructions`, `save_corpus_embeddings`, etc.
- Made the reranking happen efficiently by changing the search function in `DenseRetrievalExactSearch` to handle reranking also, added torch compile to the similarity metrics if available (I saw a 2x speedup!), and added GPU-based cosine similarity (and dot product) to speed it up also. This makes MindSmall feasible. Without the GPU and compiling, MindSmall takes too long in the dot products - with it takes 30 minutes. 
- Allow for specialized metrics like p-MRR and Robustness@10 to be implemented in a way that doesn't affect all the other tasks, in `utils.py`
- Refactored a bit of the evaluation code so that other custom metrics could use it when they do aggregation over the scores.
- Added custom loaders for MIRACL and MindSmall so they load nicely until I have time to make them into clean datasets
- Updated the tests to add `InstructionReranking`
- Added a new InstructionRetrieval task, `InstructIR` and verified it matches the paper. Added `contriever` to the model section (including fixing a bug that didn't allow dot product similarity), so I could verify the InstructIR dataset.
- Added a new Reranking task (`NevIR`) with a custom metric (paired accuracy) to make sure custom metrics work.
- Moved FollowIR-based datasets into InstructionReranking which I think is a better fit
- Refactored `HFDataLoader` into its own file
- Refactored `DRESModel` and `DenseRetrievalExtactSearch` into a separate file.

I ran `sentence-transformers/paraphrase-multilingual-mpnet-base-v2` on all of the reranking tasks, the instruction-based tasks, and some retrieval tasks to be sure that the scores match and they all do. There is some minor variation in the 4th decimal place (off by 0.0001 ish) due to the previous reranking code not having query-ids and so the sorting is a bit different on ties and such.

Eventually, I'd like to create new datasets on HF for the 15 ish reranking tasks, but I don't think I'll have time for that in the next couple weeks. It takes a few seconds to convert for all tasks except `MIRACL`, `MindSmall`, and `WebLINXCandidatesReranking` which takes 5-10 minutes to convert.

mteb/abstasks/dataloaders.py
```diff
@@ -0,0 +1,322 @@
...
+            raise ValueError(f"File {fIn} must be present with extension {ext}")
+
+    def load(
+        self, split: str = "test"
+    ) -> tuple[
+        dict[str, dict[str, str]],  # corpus
+        dict[str, str | list[str]],  # queries
+        dict[str, dict[str, int]],  # qrels/relevant_docs
+        dict[str, str | list[str]],  # instructions (optional)
+        dict[str, list[str]] | dict[str, dict[str, float]],  # top_ranked (optional)
```
Samoed, line 102:
```suggestion
        dict[str, str | list[str]] | None,  # instructions
        dict[str, list[str]] | dict[str, dict[str, float]] | None,  # top_ranked
```

Samoed (approved):
Awesome changes!

mteb/evaluation/evaluators/model_classes.py
```diff
@@ -0,0 +1,562 @@
...
+        # Model is class that provides encode_corpus() and encode_queries()
+        self.model = model
+        self.encode_kwargs = encode_kwargs
+
+        if "batch_size" not in encode_kwargs:
+            encode_kwargs["batch_size"] = 128
+        if "show_progress_bar" not in encode_kwargs:
+            encode_kwargs["show_progress_bar"] = True
+        if "convert_to_tensor" not in encode_kwargs:
+            encode_kwargs["convert_to_tensor"] = True
```
Samoed, line 63: I think this can be removed and `convert_to_tensor` can be passed explicitly to `encode`  https://github.com/embeddings-benchmark/mteb/issues/1325

mteb/evaluation/evaluators/model_classes.py
```diff
@@ -0,0 +1,562 @@
...
+        all_doc_embeddings = self.model.encode(
+            unique_docs,
+            task_name=task_name,
+            prompt_type=PromptType.passage,
+            request_qid=request_qid,
+            **self.encode_kwargs,
+        )
+
+        # Let's make sure we don't get the warnings for the tokenizer here via torch.compile
+        if hasattr(torch, "compile"):
```
Samoed, line 246: It might be better to make compile as a flag or to check the CUDA version, as shown in [their tutorial](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html), since it can slow things down on certain architectures.

mteb/evaluation/evaluators/utils.py
```diff
@@ -1,56 +1,84 @@
...
 import torch
 import tqdm
+from datasets import load_dataset
 from packaging.version import Version
 from sklearn.metrics import auc
 
+try:
+    # speeds up computation if available
+    torch.set_float32_matmul_precision("high")
+except Exception:
```
Samoed, line 19: Maybe add log message

mteb/evaluation/evaluators/utils.py
```diff
@@ -266,9 +293,85 @@ def evaluate_change(
...
+        "Core17InstructionRetrieval": ("jhu-clsp/core17-instructions-mteb", False),
+        "Robust04InstructionRetrieval": ("jhu-clsp/robust04-instructions-mteb", False),
+        "News21InstructionRetrieval": ("jhu-clsp/news21-instructions-mteb", False),
+        "mFollowIR": ("jhu-clsp/mfollowir-parquet-mteb", True),
+        "mFollowIRCrossLingual": (
+            "jhu-clsp/mfollowir-cross-lingual-parquet-mteb",
+            True,
+        ),
     }
+    hf_path, is_multilingual = TASK_TO_HF_DATASET[task_name]
```
Samoed, line 320: To check if a task is multilingual, it may be better to use:

```python
task = mteb.get_task(task_name)
is_multilingual = isinstance(task, MultilingualTask)
```

mteb/evaluation/evaluators/utils.py
```diff
@@ -266,9 +293,85 @@ def evaluate_change(
...
+    if is_multilingual:
+        # figure out which of the languages this is: ["zho", "rus", "fas"]
+        # gather the changed_qrels for each, and store the keys as a check
+        for lang in ["zho", "rus", "fas"]:
```
Samoed, line 324: Can this be loaded in task class?

isaac-chung:
@orionw wow! Great PR. I'll take a more detailed look shortly. Right now I only have a small thought on the removal of the Evaluators and AbsTask: removing these classes would be considered a breaking change, which should trigger a major version bump. We might want to do that only when we group a few other [bigger impact changes](https://github.com/embeddings-benchmark/mteb/issues?q=is%3Aopen+is%3Aissue+milestone%3Av2.0.0) together. 

So my suggestion here is, instead of fully deleting them here and now, we could only add a warning to each of those classes that they are now deprecated, and will be removed in v2.0.0. That way if anyone had started using these will have a heads up, and we avoid 2 almost back-to-back major version bumps. Those classes will be fully removed at the v2.0.0 commit.

What do you think? also @KennethEnevoldsen @Samoed

KennethEnevoldsen:
> What do you think? also @KennethEnevoldsen @Samoed

We could include these in v2.0.0 along with MIEB

(maybe we should start the v2.0.0 branch to collect some changes together?

isaac-chung:
> We could include these in v2.0.0 along with MIEB
(maybe we should start the v2.0.0 branch to collect some changes together?

Yeah, good call!

KennethEnevoldsen:
Moved from main to v.2.0.0

mteb/tasks/InstructionRetrieval/eng/InstructIR.py
```diff
@@ -0,0 +1,49 @@
...
+        sample_creation="created",
+        bibtex_citation="""@article{oh2024instructir,
+      title={{INSTRUCTIR: A Benchmark for Instruction Following of Information Retrieval Models}}, 
+      author={{Hanseok Oh and Hyunji Lee and Seonghyeon Ye and Haebin Shin and Hansol Jang and Changwook Jun and Minjoon Seo}},
+      year={{2024}},
+      eprint={{2402.14334}},
+      archivePrefix={{arXiv}},
+      primaryClass={{cs.CL}}
+}""",
+        descriptive_stats={
```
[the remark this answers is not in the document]
orionw: Where do these go now? Thanks for letting me know!

orionw:
Thanks @isaac-chung and good point!

> So my suggestion here is, instead of fully deleting them here and now, we could only add a warning to each of those classes that they are now deprecated, and will be removed in v2.0.0

I agree with this.  Unless v2.0.0 is soon, I think this will make it much easier for people to add instruction-based tasks and we should merge it in sooner than later.  @KennethEnevoldsen do you know the timeline on that?

Happy to add back in some of these evaluators even if they are unused, just so we keep compatibility.

isaac-chung:
@orionw @KennethEnevoldsen maybe we can merge this into the 2.0.0 preview branch when ready? That way if users want to try it out before the actual release they still can, and they won't be blocked by our timeline.

mteb/tasks/InstructionRetrieval/eng/InstructIR.py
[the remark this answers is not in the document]
Samoed: In folder with [descriptive_stats](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/descriptive_stats). When you use `calculate_metadata_metrics` it will place it here automaticaly

mteb/evaluation/evaluators/utils.py
orionw: It could be but then I'd have to define a custom loader. I was leaning towards not, since this is just the info needed purely for eval and not for anything else. If you have a strong preference for putting it in the task class I can move it, but I figured the current setup leaves the `tasks` section of the code cleaner.

mteb/evaluation/evaluators/utils.py
orionw: I was hesitant to call mteb.get_tasks in the utils file because of import loops - trying to keep this file as import-free.  Perhaps I am wrong though?

orionw:
@isaac-chung we could do that too

I was mainly commenting from a "new-task-adder" perspective rather than a user perspective. New PR's like #1425 are adding new instruction-retrieval tasks and the current implementation of the InstructionRetrieval is quite confusing to add new tasks (which is totally my fault, as I added that setup earlier this year and didn't keep it broad enough). 

If v2.0.0 is coming out within a month or so then perhaps it's fine to make those PRs also merge in to v2.0.0, but I was trying to avoid making it difficult for others to add because of my poor decisions :)

KennethEnevoldsen:
I believe this is a good fit for v2. Feel free to merge in into v2 as it is now (resolving the merge conflict). Please add a release note in the description over at #1433

PR #1364, opened 2024-10-30
fix: Update KorSarcasm to avoid trust-remote code

KennethEnevoldsen:


bwanglzu:
nice! didn't realise there is a new revision, this indeed seems to be a nice way to fix.

PR #1371, opened 2024-10-31
[mieb] Remove null entries from corpus of ROxford, RParis

Jamie-Stirling:
Fix issue where images that should have been ignored were in the corpus of ROxford and RParis retrieval tasks, causing the MAP to be too low. These were removed from the new variation of the respective HF datasets and their corresponding revisions updated in task metadata.

## Checklist


gowitheflow-1998 (approved):

Muennighoff:
Since I'm already running models and dumping results here https://github.com/embeddings-benchmark/tmp, would be great if you could provide me a list at some point of all tasks where scores have changed so I can delete the old ones and rerun them 👍

Issue #1321, opened 2024-10-24
standardize descriptive stats


KennethEnevoldsen:
Currently, descriptive stats are quite inconsistent. This leads problems e.g. if we want to [calculate the number of characters pr. task](https://github.com/embeddings-benchmark/mteb/issues/896#issuecomment-2423416229) to estimate the number of compute tokens needed.

All of these calculations could be automated and are in the `_calculate_metrics_from_split`, however, it is not calculated for all datasets. It would be great to have a test that tests that these are calculated consistently across all tasks.

Additionally, this data is currently included in the metadata, which might not be ideal (often requiring copy-paste, which could lead to potential errors). A solution could be to write it to a json from which the data is fetched when needed. Tests can then fail if this cache is not full.

Samoed:
I've been considering improvements to `metadata_metrics` as well.

> A solution could be to write it to a JSON file from which the data is fetched when needed. Tests can then fail if this cache is not complete.

That's a great suggestion! Are you suggesting storing a JSON file with all the metadata directly in the `mteb` repository?

KennethEnevoldsen:
Yep - packaged into the package.

PR #1375, opened 2024-11-02
Standardize descriptive stats

Samoed:
## Checklist

Closes https://github.com/embeddings-benchmark/mteb/issues/1321


KennethEnevoldsen (commented):
Thanks for starting on this @Samoed. 

Seems n_samples is maintained with taskmetadata as well as in descriptive stats (this seems like a duplicate?)

How is task metadata added to descriptive stats? Can't seem to find that. How do we ensure that it is added for new datasets without making it too inconvenient for developers.

Is there a reason not to keep descriptive stats as @property on the taskmetadata?

As I understand it right now. If I want the metadata from one dataset I have to load the metadata for all datasets. That seems problematic. Might be better to simply add file pr. task (potentially located along with the file.

Samoed:
> Seems n_samples is maintained with taskmetadata as well as in descriptive stats (this seems like a duplicate?)

A bit yes, but some tasks (2 as I remember) using this information for generating splits. I can remove it if you want.

> How is task metadata added to descriptive stats?

When called [calculate_metadata_metrics](https://github.com/embeddings-benchmark/mteb/blob/2e22fab68024bfc99d686423dbfe8eca2645f380/mteb/abstasks/AbsTask.py#L199) metadata automatically will be added to `descriptive_stats.json`.

> How do we ensure that it is added for new datasets without making it too inconvenient for developers.

I don't have good ideas how to do this only for new tasks

> As I understand it right now. If I want the metadata from one dataset I have to load the metadata for all datasets. That seems problematic

Yes, just precalculated metadata, but I don't think that loading json will be too problematic, but if you want I can create directory, where metadata will be stored per task

KennethEnevoldsen:
> A bit yes, but some tasks (2 as I remember) using this information for generating splits. I can remove it if you want.

I would move it to a variable for those cases

> I don't have good ideas how to do this only for new tasks

Let us add a test that checks that it is computed.

> Yes, just precalculated metadata, but I don't think that loading json will be too problematic, but if you want I can create directory, where metadata will be stored per task

I think this is better. It is already causing issues when reviewing and it would be a very frustrating file to have merge conflict in.

> I'll try to add this, but this can be a bit wierd, because metadata calculated using AbsTask and descriptive stats can't be calculated using TaskMetadata

I think a solution where we simply throw an error or return an empty dict if it is not there is fine. After all calling descriptive_stats should not compute anything (it should just load it)

Samoed:
I've changed everything, as you said

mteb/abstasks/AbsTaskBitextMining.py
```diff
@@ -19,11 +20,13 @@ class BitextDescriptiveStatistics(DescriptiveStatistics):
 
     Attributes:
         num_samples: number of samples in the dataset.
+        total_symbols: Total number of symbols in the dataset.
         average_sentence1_length: Average length of sentence1
         average_sentence2_length: Average length of sentence2
     """
 
     num_samples: int
+    total_symbols: int
```
KennethEnevoldsen, line 29: number_of_characters. I initially misunderstood this as unique characters.

mteb/abstasks/AbsTask.py
```diff
@@ -197,35 +190,32 @@ def load_data(self, **kwargs):
...
-        if os.path.exists(descriptive_stats_file):
-            with open(descriptive_stats_file) as f:
-                existing_descriptive_stats = json.load(f)
-
-        if existing_descriptive_stats.get(self.metadata.type) is None:
-            existing_descriptive_stats[self.metadata.type] = {}
-
-        if self.metadata.name in existing_descriptive_stats[self.metadata.type]:
-            return existing_descriptive_stats[self.metadata.type][self.metadata.name]
+        if self.metadata.descriptive_stat_path.exists() and not overwrite_results:
```
KennethEnevoldsen, line 195: add logging message

mteb/abstasks/TaskMetadata.py
```diff
@@ -386,5 +390,38 @@ def intext_citation(self, include_cite: bool = True) -> str:
...
+        stats = self.descriptive_stats
+        if not stats:
+            return None
+
+        n_samples = {}
+        for subset, subset_value in stats.items():
+            if subset == "hf_subset_descriptive_stats":
+                continue
+            n_samples[subset] = subset_value["num_samples"]
+        return n_samples
```
KennethEnevoldsen, line 424: delete? Any reason to have it on its own?

KennethEnevoldsen (approved):
A few minor suggestions. Otherwise I believe it is good. @isaac-chung if you have the time to take a look at this that would be great (you can ignore everything not in the abstask folder) - I am especially wondering if there will be conflict with MIEB (I don't believe there will be)

mteb/abstasks/TaskMetadata.py
Samoed: I've added this to simplify usage of `n_samples` in `docs/create_tasks_table.py`. Maybe in other parts of code this can be helpful

isaac-chung:
@KennethEnevoldsen I only spotted 1 obvious conflict with MIEB: Only `AbsTaskVisualSTS` has `_calculate_metrics_from_split` implemented, and will likely need to follow some of the code changes in this PR (should be easy 👌 ). Otherwise, all other image-related AbsTasks has `pass` in their respective methods. So if we can accept empty descriptive stats for these tasks, then we should be ok.
CC @gowitheflow-1998 

@Samoed thanks so much for working on this. In future PRs, to help speed up the process, could you please include a small summary in the PR description? I would really appreciate it! 🙏

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -196,17 +197,21 @@ class RetrievalDescriptiveStatistics(DescriptiveStatistics):
...
+        num_queries: number of queries in the dataset
+        num_documents: Number of documents
+        number_of_characters: Total number of symbols in the dataset
         average_document_length: Average length of documents
         average_query_length: Average length of queries
-        num_documents: Number of documents
         average_relevant_docs_per_query: Average number of relevant documents per query
     """
 
+    num_samples: int
```
KennethEnevoldsen, line 209: what does this correspond to?

mteb/abstasks/TaskMetadata.py
KennethEnevoldsen: let us keep it

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -196,17 +197,21 @@ class RetrievalDescriptiveStatistics(DescriptiveStatistics):
     """Descriptive statistics for Retrieval
 
     Attributes:
-        num_queries: number of samples in the dataset
+        num_samples: Number of samples in the dataset
```
KennethEnevoldsen, line 200:
```suggestion
        num_samples: Number of queries and documents
```

mteb/abstasks/TaskMetadata.py
Samoed: Should I change something?

KennethEnevoldsen: No, I think your argument is good - let us keep it as is.

Samoed:
@KennethEnevoldsen, is this PR ready to be merged?

KennethEnevoldsen:
Yes merged in - great PR!

PR #1398, opened 2024-11-06
fix: add logging for RetrievalEvaluator NaN values for similarity scores

### LINKED ISSUE #1389: RetrievalEvaluator NaN values for similarity scores
I am looking at RetrievalEvaluator.py and I see the line   **cos_scores[torch.isnan(cos_scores)] = -1**

Basically, if a query-document similarity score has a NaN value, then it gets changed to -1. 

My question is how would a NaN score appear when using cosine similarity as the distance metric? Would the query and document embedding vectors contain very small values when a NaN value appears?

KennethEnevoldsen:
Fixes #1389

Ran through tests but couldn't find a case where it was actually used. It might be a thing we can simply remove.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,4 @@ tests/create_meta/model_card.md
 
 # removed results from mteb repo they are now available at: https://github.com/embeddings-benchmark/results
 results/
+uv.lock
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
@orionw or @Muennighoff will merge this in but would love to remove the -1 transformation if possible

orionw:
I’m not aware of when it would happen, I think it was there when I started using MTEB.  I assume like guessed in #1389 it’s for overflow or something. 

Perhaps @Muennighoff may know more.

Muennighoff:
Yeah I'm also not sure, maybe when embeddings are NaN, but then that's a problem with the model and we probably shouldn't work around that 🤔 We can also remove it and wait until someone reports an issue with it

KennethEnevoldsen:
removing them in #1444

Issue #1412, opened 2024-11-08
descriptive stats missing for multiple datasets


KennethEnevoldsen:
@Samoed it seems like I might have missed this in review, but seems like we don't have desc. stats for all datasets in MTEB.

Would be great to have this and add a test to force this on future datasets as well.

PR #1413, opened 2024-11-08
fix: Fix test for empty descriptive tasks 

Samoed:
## Checklist

It seems that when creating the test, I missed recreating the array in the test case, which caused the test to not work correctly. Fixes https://github.com/embeddings-benchmark/mteb/issues/1412

### DIFF
--- a/tests/test_TaskMetadata.py
+++ b/tests/test_TaskMetadata.py
@@ -1088,10 +1088,8 @@ def test_empy_descriptive_stat_in_new_datasets():
         553 == len(exceptions)
     ), "The number of exceptions has changed. Please do not add new datasets to this list."
 
-    exceptions = []
-
     for task in get_tasks():
         if task.metadata.descriptive_stats is None:
             assert (
-                task.metadata.name not in exceptions
+                task.metadata.name in exceptions
             ), f"Dataset {task.metadata.name} should have descriptive stats"


KennethEnevoldsen (commented):
This of course causes the tests to fail, if possible it would be great to run calculate descriptive stats for all datasets (maybe with a few exceptions).

Samoed:
Oh, I missed that test not passing. I'm preparing an update to descriptive stats, and after that I think I can try to calculate on some datasets

KennethEnevoldsen:
Moving this to v2.0.0, but we could move it back if it makes things easier

Samoed:
@KennethEnevoldsen I fixed tests. I think for now this can be merged to main. If somebody wanted to add new tasks this test can be helpful. Future changes to descriptive stats can be moved to 2.0.0

KennethEnevoldsen:
Shouldn't the test fail? GIven that most datasets don't have descriptive stats?

Samoed:
I've added current tasks without metadata to exceptions like in `test_disallow_trust_remote_code_in_new_datasets`

KennethEnevoldsen:
Hmm, we might be misunderstanding each other. A task like "TbilisiCityHallBitextMining" no longer has descriptive statistics, but did have it beforehand.

Samoed:
I updated the test. Now, if a task is in the exceptions list (tasks currently without descriptive stats) and has metadata, it will raise an error. Otherwise, descriptive stats are required.

Samoed:
@KennethEnevoldsen I've updated test. Can you review, please?

KennethEnevoldsen (approved):
Looks great!

PR #1425, opened 2024-11-10
Add new benchmark MAIR

### LINKED ISSUE #1426: Add new benchmark MAIR
Hi, following the discussion in https://github.com/embeddings-benchmark/mteb/issues/1358, I created a PR to add MAIR dataset (https://arxiv.org/abs/2410.10127) in https://github.com/embeddings-benchmark/mteb/pull/1425. This include 126 data classes. I've tested some models, and the results are correct.

Appreciate any feedback you have. Thanks!

sunnweiwei:
Fixes #1426

- Added MAIR (https://arxiv.org/abs/2410.10127, EMNLP 2024), a diverse benchmark for instructed IR.
- The data class is defined in mteb/tasks/MAIR/eng/MAIR.py, generating 126 data classes for the 126 tasks in MAIR on the fly.
- In benchmarks/benchmarks.py, the benchmark configuration has been added.
- Tested several models, and the results are consistent with those of the original repo: https://github.com/sunnweiwei/mair.

## Checklist

### Adding datasets checklist

The added data is introduced in https://arxiv.org/abs/2410.10127, which introduces a benchmark for instructable information retrieval. It contains 126 real-world retrieval tasks across 6 domains, with instructions manually annotated. And the data has been sampled to reduce evaluation costs.

### Adding a model checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mteb"
-version = "1.36.21"
+version = "1.36.22"
 description = "Massive Text Embedding Benchmark"
 readme = "README.md"
 authors = [


shizhl-code:
Following the above process, I am currently open a pull request in https://github.com/embeddings-benchmark/results to submit our evaluation results to the newly-added MAIR benchmark.
However, I am not very clear about the format of the result file.

Samoed:
When you run your tasks, MTEB will generate a folder with results from your runs, and you can submit that folder

mteb/tasks/MAIR/eng/MAIR.py
```diff
@@ -0,0 +1,88 @@
...
+        bibtex_citation=_MAIR_CITATION,
+    )
+
+
+def load_data(self, **kwargs):
+    if self.data_loaded:
+        return
+    self.corpus, self.queries, self.relevant_docs = {}, {}, {}
+    queries_path = self.metadata_dict["dataset"]["path"]
+    docs_path = self.metadata_dict["dataset"]["path"].replace("-Queries", "-Docs")
```
Samoed, line 60: Can you place queries and docs in same repo?

Samoed:
If you have results for both instruct and non-instruct, it might be better to create separate tasks, though @orionw might have a clearer perspective on this

orionw:
+1 to adding a duplicate task if you have a specific instruction you want them to use for each.  Otherwise models can define their own instructions. 

Once #1359 is in, you’d just need to add a version of those tasks with all the same attributes but also a config/attribute called “self.instruction” (query-id -> instruction_text) format

mteb/tasks/MAIR/eng/MAIR.py
sunnweiwei: Thanks for the feedback. To keep Q/D in one repo, I could create a separate repo for each task. 

But is that necessary? I think having two repo for Q and D would be easier to manage  than having over hundreds of repo for each task.

I can still proceed with the separate repo if required.

sunnweiwei:
Hi. If we have duplicate tasks with different instructions, will they appear in separate tables on the leaderboard? Like would there be a one called (XXX with instruction) and another called (XXX without instruction)?

orionw:
@sunnweiwei they would appear as different datasets yes.  So you could have one leaderboard with them and one without, if desired.  Or push it all together into one benchmark.

sunnweiwei:
Thanks for the answer! I was thinking to put them into one table for benchmarking purpose, maybe adding a column to indicate if instructions were used. Then people could compare models with and without instructions in the same table. Good to know we can do this then.

mteb/tasks/MAIR/eng/MAIR.py
Samoed: You can create different splits for queries and documents in the same repo

sunnweiwei: Thanks, I see.

One issue is that the data in MAIR has a two-level structure: task → subtasks, as some tasks contain multiple subtasks (e.g., IFEval, SWE-Bench). It’s tricky to maintain this structure without flattening it into a single repository.

And I think people may not need to download all the data if they’re only interested in evaluating a few specific tasks. So, if we still need to put Q and D in a single repo, the best way might be to generate 126 separate repo (for each task).

Or do you have any other suggestions?

Muennighoff:
Would be great to get this in @sunnweiwei in case you're still working on it; I think it'll be very useful to the community!

sunnweiwei:
@Muennighoff Thanks! I also hope that this can be merged. I have placed the queries and documents under one repo: MAIR-Bench/MAIR-QD. I’m not sure if the new version of MTEB includes any changes, so please let me know if any update needed.

KennethEnevoldsen:
Hi @sunnweiwei seems like we have dropped the ball on this one. Do you still have the time to finalize this PR? I sadly can't update your branch so you would have to resolve the conflict by merging main

sunnweiwei:
@KennethEnevoldsen Thanks! I can now work on it. I updated some files, and it now seems to pass the check. Let me know if anything needs to be updated.

mteb/benchmarks/benchmarks.py
Samoed, line 1: I don't think that benchmarks file should be changed that much

mteb/tasks/MAIR/eng/MAIR.py
```diff
@@ -0,0 +1,88 @@
...
+        reference="https://github.com/sunnweiwei/MAIR",
+        dataset={
+            "path": "MAIR-Bench/MAIR-QD",
+            "revision": "dad8aa74094cbc560b090a883977aa8db454a6cb",
+        },
+        type="Retrieval",
+        category="s2p",
+        modalities=["text"],
+        eval_splits=TASK2SPLIT.get(task_name, []),
```
Samoed, line 41:
```suggestion
        eval_splits=TASK2SPLIT[task_name],
```

Samoed (changes requested):
Can your update benchmarks file to have less changes?

Issue #1427, opened 2024-11-10
Enable Code Carbon by default in evaluation runs


isaac-chung:
[Edit] Was discussing with @KennethEnevoldsen about [this suggestion](https://github.com/embeddings-benchmark/results/pull/43#issuecomment-2465058800) as a way to combine run eval time + hardware info so we don't have to keep track of both. Enabling this would be a convenient step towards that before we decide to use something else that is more representative (such as FLOPS).

Muennighoff:
What's the motivation for enabling it by default? I think @imenelydiaker intentionally made it optional in https://github.com/embeddings-benchmark/mteb/pull/712#issuecomment-2110933622 🤔

isaac-chung:
Was discussing with @KennethEnevoldsen about [this suggestion](https://github.com/embeddings-benchmark/results/pull/43#issuecomment-2465058800) as a way to combine run eval time + hardware info so we don't have to keep track of both. Enabling this would be a convenient step towards that before we decide to use something else that is more representative (such as FLOPS).

Muennighoff:
Makes sense, thanks for the context!!

isaac-chung:
No worries, my bad for not filling out the description in the first place. Updated!

Samoed:
CO² added to open llm leaderboard https://huggingface.co/spaces/open-llm-leaderboard/open_llm_leaderboard

isaac-chung:
Merged into branch v2.0.0

PR #1428, opened 2024-11-10
feat: enable codecarbon by default

Samoed:
This PR enables code carbon by default. Closes https://github.com/embeddings-benchmark/mteb/issues/1427

## Checklist

### DIFF
--- a/mteb/cli.py
+++ b/mteb/cli.py
@@ -265,8 +265,8 @@ def add_run_parser(subparsers) -> None:
     parser.add_argument(
         "--co2_tracker",
         type=bool,
-        default=False,
-        help="Enable CO₂ tracker, disabled by default",
+        default=True,
+        help="Enable CO₂ tracker, enabled by default",
     )
     parser.add_argument(
         "--eval_splits",
--- a/mteb/cli.py
+++ b/mteb/cli.py
@@ -142,12 +142,14 @@ def run(args: argparse.Namespace) -> None:
         args.save_predictions if hasattr(args, "save_predictions") else False
     )
 
+    enable_co2_tracker = not args.disable_co2_tracker
+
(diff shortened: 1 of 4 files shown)


mteb/cli.py
```diff
@@ -265,8 +265,8 @@ def add_run_parser(subparsers) -> None:
     parser.add_argument(
         "--co2_tracker",
         type=bool,
-        default=False,
-        help="Enable CO₂ tracker, disabled by default",
+        default=True,
+        help="Enable CO₂ tracker, enabled by default",
     )
```
isaac-chung, line 270: Now that this option is true by default, we might want to offer a `disable_co2_tracker` instead.

isaac-chung (commented):
Nice, thanks! Let's get the tests passing as well.

Samoed:
Locally, the tests are passing, and I’m not sure why they're failing in the CI environment.

Samoed:
It seems that the emission tracker fails due to multiple workers, as it works fine when I run it without `-n auto`.

KennethEnevoldsen:
Since this is technically breaking I will move it to v2.0.0

isaac-chung:
Based one [this issue](https://github.com/mlco2/codecarbon/issues/681), it seems like we can* allow multiple runs, but the tracking will be somewhat skewed. It won't matter in the tests, but will for sure affect real runs if multiple runs are done on the same machine. I think we would also add a line in README and/or a warning to let users know.

KennethEnevoldsen:
Sounds like this can be merged into v2.0.0?

isaac-chung:
Yes, just missing a line in README and/or a warning in the comment above.

Samoed:
@isaac-chung I added logging message

PR #1444, opened 2024-11-13
fix: remove NaN handling

### LINKED ISSUE #1389: RetrievalEvaluator NaN values for similarity scores

KennethEnevoldsen:
As discussed in #1398 NaN should never occur. Removing handling of them.

Fixes #1389

## Checklist

### DIFF
--- a/mteb/evaluation/evaluators/RetrievalEvaluator.py
+++ b/mteb/evaluation/evaluators/RetrievalEvaluator.py
@@ -188,12 +188,6 @@ class DenseRetrievalExactSearch:
             cos_scores = self.score_functions[score_function](
                 query_embeddings, sub_corpus_embeddings
             )
-            is_nan = torch.isnan(cos_scores)
-            if is_nan.sum() > 0:
-                logger.warning(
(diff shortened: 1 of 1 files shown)


isaac-chung (approved):

KennethEnevoldsen:
@orionw not quite sure about the merge conflict here - seems like there is no conflict, but I might be missing something ?

Samoed:
Code was moved to https://github.com/embeddings-benchmark/mteb/blob/9c58518080f6640f1e644c18fe8f174cdb2530e6/mteb/evaluation/evaluators/model_classes.py#L341-L344

KennethEnevoldsen:
Perfect - redid the fix directly on v2.0.0

PR #1449, opened 2024-11-14
[MIEB] Change Flickr30k to test split

Jamie-Stirling:
Flickr30k was previously underperforming (CLIP) for Recall@5. Changing to the test split of only 5k images/25k captions moves performance to be much closer to [CLIP benchmark](https://github.com/LAION-AI/CLIP_benchmark/blob/main/benchmark/benchmark.csv).
## Checklist


isaac-chung (approved):
Thanks! I ran these locally and yes it's much closer now. cc @gowitheflow-1998 

CLIP benchmark: 0.834 / 0.949
Ours: 0.889 / 0.914

PR #1450, opened 2024-11-14
Fix: Made data parsing in the leaderboard figure more robust

x-tabdeveloping:
Fixes some validation bugs with the leaderboard's main figure.


KennethEnevoldsen:
Looks good - to avoid diverging main and v2.0.0 we should probably work from the v2.0.0 branch

PR #1459, opened 2024-11-14
feat: add CUREv1 retrieval dataset

dbuades:
---------
## CUREv1

Over the past year, we’ve worked closely with medical professionals to develop this dataset, which we’re now sharing with the community to support research in point-of-care information retrieval—a critical daily task for many practitioners.

The CURE is a cross-lingual retrieval dataset organized into:

- Ten splits covering different medical disciplines:
  - Dentistry and Oral Health
  - Dermatology
  - Gastroenterology
  - Genetics
  - Neuroscience and Neurology
  - Orthopedic Surgery
  - Otorhinolaryngology
  - Plastic Surgery
  - Psychiatry and Psychology
  - Pulmonology

- Three cross-lingual settings:
  - English-to-English
  - Spanish-to-English
  - French-to-English

Each split and cross-lingual setting is composed of **200 natural language queries formulated by health care professionals**, capturing their information needs when consulting academic literature during their daily work.

The corpus is constructed from an index of English passages extracted from biomedical academic articles. Passages are then marked as either _Highly Relevant_, _Partially Relevant_ or _Not Relevant_ with respect to each query.

For more details, please check the [Dataset Card in the Hub](https://huggingface.co/datasets/clinia/CUREv1) 🤗 
A preprint detailing the curation process and providing an extended rationale will soon be published on arXiv, along with pre-embedded indexes for several of these models!

---------
## MTEB(Medical)

At the same time, we take the opportunity to introduce a specialized benchmark that groups MTEB tasks relevant to the medical domain. Initially, we’ve included the following tasks, but we welcome any suggestions for additional tasks you think may be valuable:

- CUREv1
- NFCorpus
- SCIDOCS
- TRECCOVID
- SciFact
- MedicalQARetrieval
- PublicHealthQA

We have also computed results for these tasks across 18 open-source models. We can upload them to [the results repo](https://github.com/embeddings-benchmark/results) or somewhere else, please point us in the right direction as there seems to be lots of activity with the new leaderboard! 💪 

---------
### Adding datasets checklist


mteb/tasks/Retrieval/multilingual/CUREv1Retrieval.py
```diff
@@ -0,0 +1,151 @@
...
+        description="Collection of query-passage pairs curated by medical professionals, across 10 disciplines and 3 cross-lingual settings.",
+        type="Retrieval",
+        modalities=["text"],
+        category="s2p",
+        reference="https://huggingface.co/datasets/clinia/CUREv1",
+        eval_splits=CUREv1Splits.names(),
+        eval_langs=_LANGUAGES,
+        main_score="ndcg_at_10",
+        date=("2024-01-01", "2024-10-31"),
+        domains=["Medical", "Academic"],
```
KennethEnevoldsen, line 53:
```suggestion
        domains=["Medical", "Academic", "Written"],
```
Assumed it is not spoken

mteb/benchmarks/benchmarks.py
```diff
@@ -308,6 +308,24 @@ def load_results(
...
+            "CUREv1",
+            "NFCorpus",
+            "SCIDOCS",
+            "TRECCOVID",
+            "SciFact",
+            "MedicalQARetrieval",
+            "PublicHealthQA",
+        ]
+    ),
+    description="Medical benchmarks from MTEB",
```
KennethEnevoldsen, line 324: Maybe worth adding a selection procedure / curation rationale here. 

It seems like you haven't included e.g. MedrxivClusteringS2SFast, MedrxivClusteringP2PFast

It also seems to be somewhat multilingual, but does not include:

- CMedQAv1

There are also some datasets that do include medical, e.g. AutoRAGRetrieval, but do not target the medical domain per say. I also think these are fine to ignore.

You have also included e.g. scidocs even though it is not annotated for "medical"?

KennethEnevoldsen: It might also be worth differentiating between medical within a healthcare system (search on patient journals) and within e.g. medical journals or QA as these might be quite different. I would make it clear in the description as well.

I know other descriptions are quite small, but we can reasonably add a paragraph in the new leaderboard.

KennethEnevoldsen (commented):
Great PR!!

> We have also computed results for these tasks across 18 open-source models. We can upload them to [the results repo](https://github.com/embeddings-benchmark/results) or somewhere else, please point us in the right direction as there seems to be lots of activity with the new leaderboard!

Please do!

KennethEnevoldsen (approved):

mteb/benchmarks/benchmarks.py
dbuades: Great catch! We were too focused on strictly adding retrieval tasks and overlooked the clustering tasks. Thanks for flagging them @KennethEnevoldsen.

Here’s what we’re proposing:
	•	SCIDOCS:  Removed. While it’s an interesting dataset on scientific literature, we agree it’s too broad for inclusion in this specific benchmark.
	•	TRECCOVID-PL and SciFact-PL: Added. These fit the multilingual focus, and their English counterparts are already included. We also added CMedQAv2 and CmedqaRetrieval.
	•	MedrxivClusteringP2P and MedrxivClusteringS2S: Added as proposed, as they fit well with the type of content we’re targeting.
	•	VieMedEVBitextMining: Not included. Despite being medically oriented, it is a translation mining task and doesn’t quite fit the medical retrieval context.

Regarding the benchmark scope, while I agree the medical field is quite broad, all the current medical tasks in MTEB fall under the medical information retrieval category, rather than EMR search or its derivatives. My concern is that narrowing the scope too much could limit the benchmark’s growth potential, especially if we want to include tasks related to EMR search or patient education in the future since creating a second benchmark for those tasks would add unnecessary fragmentation. Of course, if the benchmark eventually grows to include a large number of tasks covering diverse use cases in the medical field, we can split it into more specific benchmarks down the line. I've clarified the description for now !

Also, if you agree with this line of thought, I have renamed the benchmark to `medMTEB`, since it sounds more catchy haha

If you agree, please go ahead and merge this PR. I’ll run the models on the newly added tasks tomorrow and open a PR with all the results so they’re ready to be displayed on the leaderboard 💪

dbuades: Thanks! I'm happy to wait for a second opinion, but I'm not strongly attached to the `medMTEB` naming either, it was just a suggestion since it sounded nice. So if that's the only thing left blocking this PR I don't have a problem changing it back.

gowitheflow-1998: thanks for the great PR! re. naming, I agree that following the naming convention might be better; also easier for users that want to evaluate on it to grab tasks in this benchmark with the correct name, since we have other domain specific ones like MTEB(law) etc.

dbuades: Thanks for chiming in! I've renamed it back to `MTEB(Medical)`

isaac-chung:
Thanks for the PR! The only thing left to point out is that it'd be great if the PR description can be updated to reflect the changes made above. Otherwise I think this is good to merge. If you end up promoting this on socials, let us know :)

dbuades:
> Thanks for the PR! The only thing left to point out is that it'd be great if the PR description can be updated to reflect the changes made above. Otherwise I think this is good to merge. If you end up promoting this on socials, let us know :)

Thanks, @isaac-chung ! Sorry I missed your last comment. I’ve updated the PR description retroactively and am currently running the 18 models on all the newly added tasks in the benchmark. Once that’s done, I’ll open a PR with the results.

As for promoting the work, do you have any specific ideas in mind? I was planning to post something on LinkedIn next week, which I can also share here. Additionally, we’re preparing a preprint that we’ll be uploading to arXiv soon. Maybe we could use that opportunity to co-write something for the HF blog? We can discuss the angle but I believe that it could be really interesting!

isaac-chung:
Thanks, @dbuades! Those all sound good, and I'm happy to share/repost what you have. An HF blog would be good as well - happy to collaborate there!

dbuades:
> Thanks, @dbuades! Those all sound good, and I'm happy to share/repost what you have. An HF blog would be good as well - happy to collaborate there!

Perfect! I'll keep you posted next week.

Issue #1465, opened 2024-11-16
Can't load some datasets because of datasets 3.0


Samoed:
When I try to download datasets, I get the error `ModuleNotFoundError: No module named 'datasets.tasks'` or a similar one, because in `datasets` version 3.0, `tasks` was deprecated
Tasks with errors:
- IndicQARetrieval
- HotelReviewSentimentClassification
- TweetEmotionClassification
- TenKGnadClassification
- ArxivClassification
- PatentClassification
- FilipinoHateSpeechClassification
- MyanmarNews
- DutchBookReviewSentimentClassification
- SwedishSentimentClassification
- WisesightSentimentClassification
- UrduRomanSentimentClassification
- JSTS

Simple solution would be fix datasets version `<3.0`, but I'm not sure if it good.

KennethEnevoldsen:
It is a good temp. fix. A better fix is to re-upload the datasets

KennethEnevoldsen:
> Simple solution would be fix datasets version <3.0, but I'm not sure if it good.

That is a good temp. solution, but not a good long-term one. I think it is the right move right now though

reneleonhardt:
@KennethEnevoldsen datasets 3.5 has been released, when is the [migration](https://github.com/huggingface/datasets/releases/tag/3.0.0) planned to apply all those improvements?
Time doesn't stand still, at least in the AI universe 😄

As a side note, it's sad that you have to use an unmaintained dependency like [salesforce-lavis](https://github.com/salesforce/LAVIS) and therefore had to [downgrade Python](https://github.com/embeddings-benchmark/mteb/pull/2323) 😞
Salesforce profits 2024: $26 billion

Samoed:
We have fixed these datasets in `v2` branch

reneleonhardt:
> We have fixed these datasets in `v2` branch

Nice!

PR #1470, opened 2024-11-17
fix: pinned datasets to <3.0.0

Napuh:
This PR pins `datasets` version to be less than 3.0.0 to avoid `datasets.tasks` errors.

This solves #1465 

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dependencies = [
-    "datasets>=2.19.0",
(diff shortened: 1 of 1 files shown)


Samoed:
I think it's better to make changes to `v2.0.0` branch

KennethEnevoldsen:
I think this is good. However as @Samoed argues I would move this PR to v2.0.0 (causes a lot of conflict so would just redo it from scratch)

Napuh:
I made #1471 for that. Feel free to close or merge this one if you want.

KennethEnevoldsen:
Thanks I have merged both

PR #1473, opened 2024-11-19
feat: Multilingual retrieval loader

Samoed:
## Checklist

I've added a multilingual retrieval loader with the following structure:  

- `{config}-corpus`  
- `{config}-queries`  
...  

However, currently, there are no datasets that fit this structure. Previously, when I uploaded datasets, I used `corpus-config`, but more other datasets start with `config`. These datasets have variations that make them incompatible with this loader for direct use, so this is more for the future


Samoed:
I tested this on `CodeSearchCCRetrieval`, which uses `corpus-{config}`. It worked when I swapped the name in the corpus for the test. However, I couldn't find datasets where the loader could be changed due to different naming schemes. This PR won't break anything; it's an attempt to standardize for future use

Issue #1491, opened 2024-11-23
Custom Models do not receive prompt_type or task_name


ArthurSBianchessi:
Specifically SentenceTransformerWrapper (from mteb/models/sentence_transformer_wrapper.py) does not forward these variables to Custom Model encode.

Samoed:
Can you provide code to check?

ArthurSBianchessi:
Sure. 
The code in the README.md raises error due to missing task_name
```python
from mteb.encoder_interface import PromptType
import mteb
from mteb import MTEB
import numpy as np

class CustomModel:
    def encode(
        self,
        sentences: list[str],
        task_name: str,
        prompt_type: PromptType | None = None,
        **kwargs,
    ) -> np.ndarray:
        """Encodes the given sentences using the encoder.
        
        Args:
            sentences: The sentences to encode.
            task_name: The name of the task.
            prompt_type: The prompt type to use.
            **kwargs: Additional arguments to pass to the encoder.
            
        Returns:
            The encoded sentences.
        """
        pass

model = CustomModel()
# tasks = mteb.get_task("Banking77Classification")              # Raises TypeError: 'Banking77Classification' object is not iterable
tasks = mteb.get_tasks(tasks=["Banking77Classification"])
evaluation = MTEB(tasks=tasks)
evaluation.run(model)
```

Printing the inputs of encode with:
```python
class CustomModel:
    def encode(self, sentences, *inputs, **kwargs):
        print(f'inputs: {inputs}')
        print(f'kwargs: {kwargs}')
``` 
outputs:
```
inputs: ()
kwargs: {'prompt_name': None, 'batch_size': 32}
```
There might be some other variable that is not being forwarded.

Samoed:
This is correct because `task_name` is a positional argument and is passed correctly to `CustomModel`. It would be better to update the example in the README. The `CustomModel` should function similarly to `SentenceTransformerWrapper`. You can find more implementations of custom models in the `models` directory.

ArthurSBianchessi:
Right now neither task_name nor prompt_type is passed to CustomModels, I did create a pull request, but sent task name as a keyword argument because any `encode' method that expects taskname as a positional argument does not work, and any other method that currently does not break in mteb amd does not receive `*args` will break (I expect it to be most models). In my understanding, `CustomModel.encode` should receive `task_name` either as a positional or as a keyword argument, but currently it does not receive either.
[...]

Issue #1490, opened 2024-11-23
Cannot reproduce results on text classification benchmark.


tsWen0309:
I am trying to reproduce the performance of the model "jina_v3" https://huggingface.co/jinaai/jina-embeddings-v3 on text classificiaton benchmark. 
And I am using the code below:

import mteb
from sentence_transformers import SentenceTransformer
from transformers import AutoTokenizer, AutoModel

model_name = "jinaai/jina-embeddings-v3"
model = SentenceTransformer('model_name',trust_remote_code=True)
tasks = mteb.get_tasks(tasks=['AmazonCounterfactualClassification',
                              'AmazonReviewsClassification',
                              "Banking77Classification",
                              'EmotionClassification',
                              'ImdbClassification',
                              'MTOPIntentClassification',
                              'ToxicConversationsClassification',
                              'TweetSentimentExtractionClassification'])
evaluation = mteb.MTEB(tasks=tasks)
results = evaluation.run(model, eval_splits=["test"],output_folder=f"results/{model_name}")

The results seem to differ significantly from the results in https://huggingface.co/spaces/mteb/leaderboard .
Any suggestion?

Samoed:
You should load model like this: 
```python
import mteb

model = mteb.load_model("jinaai/jina-embeddings-v3")
...
```

tsWen0309:
> You should load model like this:
> 
> ```python
> import mteb
> 
> model = mteb.load_model("jinaai/jina-embeddings-v3")
> ...
> ```

mteb has no attribute "load_model" ?  I am using mteb==1.20.0

Samoed:
Sorry, this should be 
```python
import mteb

model = mteb.get_model("jinaai/jina-embeddings-v3")
...

tsWen0309:
![image](https://github.com/user-attachments/assets/9c284e53-587e-4fd4-9476-da8943a4be03)
Then this bug occurs

tsWen0309:
File "D:\code\mteb-main\mteb\models\overview.py", line 126, in get_model
    model = meta.load_model(**kwargs)
  File "D:\code\mteb-main\mteb\model_meta.py", line 120, in load_model
    model: Encoder = loader(**kwargs)  # type: ignore
  File "D:\code\mteb-main\mteb\model_meta.py", line 37, in sentence_transformers_loader
    return SentenceTransformerWrapper(model=model_name, revision=revision, **kwargs)
[...]

PR #1493, opened 2024-11-24
fix: align readme with current mteb

Samoed:
## Checklist

The current README is a bit misleading and causing issues for newcomers. See #1490 and #1491 for reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ model_name = "average_word_embeddings_komninos"
 # model_name = "sentence-transformers/all-MiniLM-L6-v2"
 
 model = SentenceTransformer(model_name)
+# or directly from mteb:
+model = mteb.get_model(model_name)
 tasks = mteb.get_tasks(tasks=["Banking77Classification"])
 evaluation = mteb.MTEB(tasks=tasks)
 results = evaluation.run(model, output_folder=f"results/{model_name}")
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -50,6 +50,8 @@ model_name = "average_word_embeddings_komninos"
 # model_name = "sentence-transformers/all-MiniLM-L6-v2"
 
 model = SentenceTransformer(model_name)
+# or directly from mteb:
+model = mteb.get_model(model_name)
```
KennethEnevoldsen, line 54: Should we just recommend this one always? 

and maybe rephrase to:
```suggestion
model = mteb.get_model(model_name) # if the model is not implemented in MTEB it will be eq. to SentenceTransformer(model_name)
```

README.md
```diff
@@ -220,9 +222,13 @@ Note that the public leaderboard uses the test splits for all datasets except MS
...
 
 ```python
+import mteb
 from mteb.encoder_interface import PromptType
+from mteb.models.wrapper import Wrapper
+import numpy as np
+
 
-class CustomModel:
+class CustomModel(Wrapper):
```
KennethEnevoldsen, line 231: Why does it need to inherit from wrapper? I would instead inherit from the Encoder protocol

Samoed: MTEB class will wrap models that not inherit wrapper class https://github.com/embeddings-benchmark/mteb/blob/3ff38ec043f6ebc4e434c083cf96dc823439f8a3/mteb/evaluation/MTEB.py#L366-L367

README.md
Samoed: Yes, but I'm not sure removing it is the best approach. I can make changes to whatever you think is better

README.md
KennethEnevoldsen: Hmm, maybe it should just wrap SentenceTransformers?

Samoed: It can be like this. You changed it in mieb https://github.com/embeddings-benchmark/mteb/blob/fab0b82a5c2680ab4e4823641955cdc7b9bb9891/mteb/evaluation/MTEB.py#L356-L357 I can change there too

KennethEnevoldsen: Hmm then it will also be merged into v2.0.0 in which case we should probably just update the readme there (I plan to merge it during December)

Samoed: So, for now I think this PR can be merged, and I will update readme in the 2.0 branch

Samoed:
I made model check as in `mieb` branch

KennethEnevoldsen (approved):
Yep - pull from main should do

KennethEnevoldsen:
This seems ready to merge (for future reference feel free to just merge once you have gotten an approved review)

PR #1268, opened 2024-10-01
fix: evaluate missing splits

thivyanth:
Addresses #1260

## Checklist

@Muennighoff  @isaac-chung 
- [x] Run tests locally to make sure nothing is broken using `make test`. 
   -  `944 passed, 236 skipped, 55 warnings in 232.13s (0:03:52)`
- [x] Run the formatter to format the code using `make lint`.

PR #1525, opened 2024-11-28
feat: Evaluate missing splits

### LINKED ISSUE #1260: Only skip benchmarking if split results are the same too
We currently skip evaluation if the results file already exists, however, it may be that the file exists but not with the same splits/subsets as requested by the current evaluation. I think we may want to also check equivalence across splits & subsets and then run the ones still missing and edit the existing results file to put them back in.

isaac-chung:
Fixes #1260 

- Based on https://github.com/embeddings-benchmark/mteb/pull/1268 and addresses most of the comments
- Once run, the returned result will contain the missing splits as well as the already run splits
- Requires `overwrite=True` to work
- Changed tests to use the same MTEB object instead of a new one

Example Usage
```python
from mteb import MTEB, get_tasks, get_model

tasks = get_tasks(tasks=["Banking77Classification"])
model = get_model("sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2")

# Runs only one split
evaluation = MTEB(tasks=tasks)
results = evaluation.run(
    model,
    eval_splits=["val"],
    output_folder=str(tmp_path / "testcase2"),
    verbosity=2,
)

# Will run the missing split and return results of both splits.
results2 = evaluation.run(
    model,
    eval_splits=["val", "test"],
    output_folder=str(tmp_path / "testcase2"),
    verbosity=2,
    overwrite_results=True,
)
```

## Checklist

### DIFF
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -355,9 +355,9 @@ class MTEB:
         if existing_kg_co2_emissions and new_kg_co2_emissions:
             merged_kg_co2_emissions = existing_kg_co2_emissions + new_kg_co2_emissions
         merged_results = TaskResult(
-            dataset_revision=existing_results.dataset_revision,
-            task_name=existing_results.task_name,
-            mteb_version=existing_results.mteb_version,
+            dataset_revision=new_results.dataset_revision,
+            task_name=new_results.task_name,
+            mteb_version=new_results.mteb_version,
             scores=merged_scores,
             evaluation_time=existing_results.evaluation_time
             + new_results.evaluation_time,
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -350,7 +350,7 @@ class MTEB:
             scores=merged_scores,
             evaluation_time=existing_results.evaluation_time
             + new_results.evaluation_time,
-            kg_co2_emissions=existing_results.kg_co2_emissions,
+            kg_co2_emissions=existing_results.kg_co2_emissions + new_results.kg_co2_emissions,
         )
 
         return merged_results
@@ -437,7 +437,6 @@ class MTEB:
         )  # save them in case we re-use the object (e.g. for reranking)
 
         # To evaluate missing splits, we keep track of the task name and the corresponding splits.
-        # TODO: track languages too.
         self.last_evaluated_splits = {}
 
         while len(self.tasks) > 0:
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -5,7 +5,7 @@ import logging
 import os
 import traceback
 from collections.abc import Iterable
-from copy import copy
+from copy import copy, deepcopy
 from datetime import datetime
 from itertools import chain
 from pathlib import Path
@@ -15,6 +15,7 @@ from typing import Any
 import datasets
 from sentence_transformers import SentenceTransformer
 
+from mteb.abstasks.AbsTask import ScoresDict
 from mteb.encoder_interface import Encoder
 from mteb.model_meta import ModelMeta
 from mteb.models import model_meta_from_sentence_transformers
@@ -84,6 +85,8 @@ class MTEB:
 
         self._version = version
         self.err_logs_path = err_logs_path
+
+        self.last_evaluated_splits = {}
 
         self.select_tasks(**kwargs)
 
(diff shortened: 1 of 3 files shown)


mteb/evaluation/MTEB.py
```diff
@@ -308,6 +311,59 @@ def _run_eval(
...
+                merged_scores[split] = scores
+
+        merged_results = TaskResult(
+            dataset_revision=existing_results.dataset_revision,
+            task_name=existing_results.task_name,
+            mteb_version=existing_results.mteb_version,
+            scores=merged_scores,
+            evaluation_time=existing_results.evaluation_time
+            + new_results.evaluation_time,
+            kg_co2_emissions=existing_results.kg_co2_emissions,
```
Samoed, line 353: Should `kg_co2_emissions` sum too?

mteb/evaluation/MTEB.py
```diff
@@ -308,6 +311,59 @@ def _run_eval(
...
+                merged_scores[split] = MTEB._merge_split_scores(
+                    merged_scores[split], scores
+                )
+            else:
+                merged_scores[split] = scores
+
+        merged_results = TaskResult(
+            dataset_revision=existing_results.dataset_revision,
+            task_name=existing_results.task_name,
+            mteb_version=existing_results.mteb_version,
```
Samoed, line 349: What we should do if existing result and new result have different version?

Samoed (commented):
Overall looks great!

mteb/evaluation/MTEB.py
KennethEnevoldsen: One solution is only to extend results if the versions match.

mteb/evaluation/MTEB.py
```diff
@@ -379,38 +435,63 @@ def run(
         original_tasks = (
             self.tasks.copy()
         )  # save them in case we re-use the object (e.g. for reranking)
+
+        # To evaluate missing splits, we keep track of the task name and the corresponding splits.
+        # TODO: track languages too.
```
KennethEnevoldsen, line 440: by language here you mean the subset on HF?

mteb/evaluation/MTEB.py
```diff
@@ -379,38 +435,63 @@ def run(
...
                 evaluation_time = 0
                 kg_co2_emissions: int | None = 0 if co2_tracker else None
-                for split in task_eval_splits:
+
+                self.last_evaluated_splits[task.metadata.name] = []
+
+                for split in missing_splits:
                     if co2_tracker:
                         try:
                             from codecarbon import EmissionsTracker
```
KennethEnevoldsen, line 497: (not really for this PR but easy to add) codecarbon is a required dependency now so we could just move this to the top

KennethEnevoldsen (approved):
Looks great only a few minor things

mteb/evaluation/MTEB.py
Samoed: The target branch of this PR is `main`, and in this branch, CodeCarbon is still optional

isaac-chung: Ah yes, the required dependency is only in v2.0.0 for now, not `main` yet.

mteb/evaluation/MTEB.py
isaac-chung: Will add a warning if the versions do not match

mteb/evaluation/MTEB.py
Samoed: I think this is important, but I can work on after merge this PR

isaac-chung: Great!

mteb/evaluation/MTEB.py
isaac-chung: This was causing a test to fail, as this could be None, and needs to be handled. I think it's best for now to assign as None if either new or existing one is None, and only add them if both are non-None.

mteb/evaluation/MTEB.py
Samoed: Then maybe to use verion from new_results?

isaac-chung: Sure, will do.

Separately, I don't think this currently takes the difference in _dataset version_ into consideration. I think this is already an existing gap, where we only check whether the same model + model revision has been run, but not check for dataset version. We should probably address it here before merging. wdyt?

Samoed: Most datasets downloaded by revision, so I don't think we need more checks

isaac-chung: Ok, then this should be good to merge once the tests pass.

isaac-chung:
Thanks @Samoed and @KennethEnevoldsen for reviewing, and @thivyanth for the initial iteration! Merging now.

mteb/evaluation/MTEB.py
KennethEnevoldsen: I think it is good to merge. It might be nice to create a "deprecation version", e.g. before 1.11.0, we can then slowly outdated old results as needed. However, this is probably something for a more general discussion, before any implementation.

PR #1530, opened 2024-11-29
fix: Proprietary models now get correctly shown in leaderboard

x-tabdeveloping:
Proprietary models were not correctly shown due to model size filtering being incorrectly implemented.
I also fixed column wrapping and formatting, the per task performance table is also correctly ordered.

### DIFF
--- a/mteb/leaderboard/table.py
+++ b/mteb/leaderboard/table.py
@@ -163,9 +163,13 @@ def scores_to_tables(
         lambda name: name.split("/")[-1]
     )
     # Adding markdown link to model names
-    joint_table["model_name"] = (
-        "[" + joint_table["model_name"] + "](" + joint_table.pop("model_link") + ")"
+    name_w_link = (
+        "[" + joint_table["model_name"] + "](" + joint_table["model_link"] + ")"
     )
+    joint_table["model_name"] = joint_table["model_name"].mask(
+        joint_table["model_link"].notna(), name_w_link
+    )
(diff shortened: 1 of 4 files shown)


x-tabdeveloping:
Hmmm tests are failing for a seemingly unrelated reason...

isaac-chung:
Reruns are successful now 👍

KennethEnevoldsen (approved):

mteb/leaderboard/table.py
```diff
@@ -138,34 +139,43 @@ def scores_to_tables(
...
     # Removing HF organization from model
     joint_table["model_name"] = joint_table["model_name"].map(
         lambda name: name.split("/")[-1]
     )
     # Adding markdown link to model names
-    joint_table["model_name"] = (
-        "[" + joint_table["model_name"] + "](" + joint_table.pop("model_link") + ")"
+    name_w_link = (
+        "[" + joint_table["model_name"] + "](" + joint_table["model_link"] + ")"
     )
```
KennethEnevoldsen, line 174: Non-blocking: I would probably just create the column initially as model_name_w_link using a method on the model_meta object like model_meta.model_name_as_markdown(). 

(could imagine we could use this elsewhere and also removed completely from the leaderboard code)

KennethEnevoldsen (approved):

mteb/leaderboard/table.py
x-tabdeveloping: Yeah, seems reasonable, I think I will do this in a future PR

Issue #1515, opened 2024-11-27
leaderboard 2.0: Add missing models


KennethEnevoldsen:
Planned but still quite a few models missing; would be great to have them all! (e.g. only 42 models for MTEB classic vs 337 in the current leaderboard)

@x-tabdeveloping and I am currently working on resolving these by adding the metadata for the missing models.

related to #1317

isaac-chung:
What's the current diff / where to find that? Would love to help out.

KennethEnevoldsen:
This is my classification of the models:

```python
sorting = {
    "clear keep": [
        # probably all of these should be registrered as well
        "Alibaba-NLP/gte-Qwen1.5-7B-instruct",
        "Alibaba-NLP/gte-Qwen2-1.5B-instruct",
        "Alibaba-NLP/gte-base-en-v1.5",
        "Alibaba-NLP/gte-large-en-v1.5",
        "Alibaba-NLP/gte-multilingual-base",
        "BAAI/bge-en-icl",
        "BAAI/bge-m3",
        "BAAI/bge-multilingual-gemma2",
        "Cohere/Cohere-embed-english-light-v3.0",
        "Cohere/Cohere-embed-english-v3.0",
        "Cohere/Cohere-embed-multilingual-light-v3.0",
        "Cohere/Cohere-embed-multilingual-v3.0",
        "Linq-AI-Research/Linq-Embed-Mistral",
        "Muennighoff/SGPT-1.3B-weightedmean-msmarco-specb-bitfit",
        "Muennighoff/SGPT-125M-weightedmean-msmarco-specb-bitfit",
        "Muennighoff/SGPT-125M-weightedmean-nli-bitfit",
        "Muennighoff/SGPT-2.7B-weightedmean-msmarco-specb-bitfit",
        "Muennighoff/SGPT-5.8B-weightedmean-msmarco-specb-bitfit",
        "Muennighoff/SGPT-5.8B-weightedmean-nli-bitfit",
        "jinaai/jina-embedding-b-en-v1",
        "jinaai/jina-embedding-l-en-v1",
        "jinaai/jina-embedding-s-en-v1",
        "jinaai/jina-embeddings-v2-base-de",
        "jinaai/jina-embeddings-v2-base-en",
        "jinaai/jina-embeddings-v2-base-es",
        "jinaai/jina-embeddings-v2-small-en",
        "jxm/cde-small-v1",
        "intfloat/e5-base",
        "intfloat/e5-large",
[...]

PR #1537, opened 2024-12-02
docs: add more model meta (jina, e5)

isaac-chung:
For sentence transformers compatible models. Part of #1515 .

- [x]  e5 models
- [x]  jina models

## Checklist


mteb/models/sentence_transformers_models.py
```diff
@@ -165,3 +165,81 @@
     supersedes="sentence-transformers/all-mpnet-base-v1",
     adapted_from=None,
 )
+
+e5_base_v2 = ModelMeta(
+    name="intfloat/e5-base-v2",
```
Samoed, line 170: I think prompts should be added to the e5 models and moved to [e5_models.py](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/e5_models.py). Additionally, we could consider integrating a mechanism to fetch metadata from Hugging Face, similar to the script in the results repository for fetching external results.

KennethEnevoldsen: (@x-tabdeveloping did some work on the last part - we could use a process similar to the descriptive stats to avoid always downloading these). Let us make an issue on that part

KennethEnevoldsen (changes requested):
Thanks, but agree with @Samoed's suggestion on moving it to the e5_models.py and adding the prompt.

mteb/models/sentence_transformers_models.py
```diff
@@ -165,3 +165,81 @@
...
+    n_parameters=109_000_000,
+    memory_usage=None,
+    embed_dim=768,
+    license="apache-2.0",
+    max_tokens=512,
+    reference="https://huggingface.co/intfloat/e5-base-v2",
+    similarity_fn_name="cosine",
+    framework=["Sentence Transformers", "PyTorch"],
+    use_instructions=False,
+    supersedes="intfloat/e5-base",
```
isaac-chung, line 184: This might make more sense for filtering if it was `superseded_by`, like how we have it for AbsTasks?

KennethEnevoldsen: Here did consider the discrepancy between the two approaches. `superseded_by` is easier to filter supersedes is easier to add. We could make at property on the modelMeta to get `superseeded_by`. Fine with using superseeded_by instead (probably nicer for consistency).

mteb/models/sentence_transformers_models.py
isaac-chung: Added to e5_models.py

KennethEnevoldsen (approved):

PR #1543, opened 2024-12-03
fix: add nomic models

KennethEnevoldsen:
#1515

## Checklist

### DIFF
--- a/mteb/models/nomic_models.py
+++ b/mteb/models/nomic_models.py
@@ -87,6 +87,17 @@ nomic_embed_v1_5 = ModelMeta(
     open_weights=True,
     revision="b0753ae76394dd36bcfb912a46018088bca48be0",
     release_date="2024-02-10",  # first commit
+    n_parameters=137_000_000,
+    memory_usage=None,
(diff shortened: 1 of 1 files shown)


mteb/models/nomic_models.py
```diff
@@ -87,6 +87,17 @@ def encode(  # type: ignore
...
     release_date="2024-02-10",  # first commit
+    n_parameters=137_000_000,
+    memory_usage=None,
+    max_tokens=8192,
+    embed_dim=768,
+    license="apache-2.0",
+    reference="https://huggingface.co/nomic-ai/nomic-embed-text-v1.5",
+    similarity_fn_name="cosine",
+    framework=["Sentence Transformers", "PyTorch"],
+    use_instructions=True,
```
x-tabdeveloping, line 98: Does only specifying the task count as instruction?

x-tabdeveloping (approved):
Looks great!

mteb/models/nomic_models.py
KennethEnevoldsen: yes

PR #1544, opened 2024-12-04
[mieb] Fix numbers of CIRR, Fashion200k, FashionIQ, Flickr30k, MSCOCO data statistics

izhx:
When reviewing the mieb tasks, I found that the statistical information of some datasets did not match the actual data that was loaded, so I corrected it to the accurate numbers.

## Checklist

### DIFF
--- a/mteb/tasks/Image/Any2AnyRetrieval/eng/CIRRIT2IRetrieval.py
+++ b/mteb/tasks/Image/Any2AnyRetrieval/eng/CIRRIT2IRetrieval.py
@@ -35,13 +35,13 @@ class CIRRIT2IRetrieval(AbsTaskAny2AnyRetrieval):
         year={2021}
         }""",
         descriptive_stats={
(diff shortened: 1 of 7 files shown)


KennethEnevoldsen:
Thanks will let @gowitheflow-1998 review, but just want to mention that we over on v2.0.0 have added computation function such that the desc. stats are computed as saved to disk instead of having users manually input it. We will probably need to add this to MIEB as well

gowitheflow-1998 (approved):

gowitheflow-1998:
thanks for the fix!

gowitheflow-1998:
> Thanks will let @gowitheflow-1998 review, but just want to mention that we over on v2.0.0 have added computation function such that the desc. stats are computed as saved to disk instead of having users manually input it. We will probably need to add this to MIEB as well

that will be ideal. Will take a look after MIEB-lite.

Issue #1047, opened 2024-07-05
Implements a custom for documents to make data format consistent


KennethEnevoldsen:
This is not planned for MMTEB, but probably a good future addition.

Currently data formats are typically quite diverse we have:

```
corpus = list[str] # most cases
corpus = list[list[str]] # e.g. for conversations
corpus = list[dict[str, str]] # classic retrieval
```

These are to allow users to have custom ways to encode these format (which is ideal), however it does lead to quite a few issues (e.g. #1041) and handling. 

A solution is to convert everything into a Document:

```
# initialization
document = Document(["a", "b", "c"]) # works
document = Document([["a", "b"], ["c"]]) # works
document = Document([{"doc1": "a", ...}) # works

# if nothing else is assumed we have:
document: Sequence(str) # essentially a list

# but for models with encode_conversation:
document.to_conversiations()

# for models with custom encode for dict[str, str]
document.get_dict_repr()
```

This allows for highly custom data formats for certain tasks, but with the restriction that the format should always be convertable to list[str].

KennethEnevoldsen:
This we added this in v2, although in a slightly different shape, which also allows for other modalities

PR #1551, opened 2024-12-04
Discussing a standard for ImageEncoders

KennethEnevoldsen:
An intermediate proposed interface for the image encoder designed based on VLM2VecWrapper.

The intention here is to define the current interface and adapt to as close to MTEB as possible and align as much as possible with sentencetransformers.

My goal is to make the interface compatible with the current Encoder interface (or close to). Thus removing the modality specific `get_image_embeddings`, but leave it to be determined by model.

This is to start a discussion, let me know what you think.

### DIFF
--- a/mteb/encoder_interface.py
+++ b/mteb/encoder_interface.py
@@ -7,6 +7,9 @@ from typing import Any, Protocol, Union, runtime_checkable
 import numpy as np
 import torch
 
+from PIL import Image
+from torch.utils.data import DataLoader
+
 Corpus = Union[list[dict[str, str]], dict[str, list[str]]]
 
 
@@ -23,13 +26,15 @@ class Encoder(Protocol):
     In general the interface is kept aligned with sentence-transformers interface. In cases where exceptions occurs these are handled within MTEB.
     """
 
+    device: str | None
+
     def __init__(self, device: str | None = None) -> None:
         """The initialization function for the encoder. Used when calling it from the mteb run CLI.
 
         Args:
             device: The device to use for encoding. Can be ignored if the encoder is not using a device (e.g. for API)
         """
-        self.device = device
+        pass
 
     def encode(
         self,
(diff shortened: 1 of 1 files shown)


mteb/encoder_interface.py
```diff
@@ -151,3 +156,56 @@ def convert_conv_history_to_query(conversations: Sequence[Sequence[str]]) -> str
...
+        texts: list[str],
+        *, 
+        **kwargs, 
+    ) -> np.ndarray:
+        pass
+
+    def get_fused_embeddings( # hmm what if I have a document with images at specific positions?
+        self,
+        texts: list[str] | None = None,
+        images: list[Image.Image] | DataLoader | None = None, # the requirement for these two to be the same seems odd (docs without images, images without associated text, docs with multiple images)
```
KennethEnevoldsen, line 206: A solution here might to do something like:

documents: list[Document]

Where you could then do:

```
images = doc.images
text = doc.text

img_embbeddings = model.encode(images)
text_embeddings = model.encode(text)
# aggregate and return
```

This would also solve the weird cases we have for documents in the corpus with a "title". We could simply convert these to documents as well.

mteb/encoder_interface.py
```diff
@@ -151,3 +156,56 @@ def convert_conv_history_to_query(conversations: Sequence[Sequence[str]]) -> str
...
+        self,
+        sentences: Sequence[str],
+        *,
+        task_name: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> np.ndarray:
+        pass
+
+    def get_image_embeddings( # Seems like sentence transformers use a singular encode for both images and text. Not sure if we want to do the same.
```
KennethEnevoldsen, line 185: I have asked about the sentenceTransformer.encode interface for images here:

https://github.com/UKPLab/sentence-transformers/issues/3118

mteb/encoder_interface.py
```diff
@@ -151,3 +156,56 @@ def convert_conv_history_to_query(conversations: Sequence[Sequence[str]]) -> str
...
+
+    def get_text_embeddings( # any reason for this?
+        self,
+        texts: list[str],
+        *, 
+        **kwargs, 
+    ) -> np.ndarray:
+        pass
+
+    def get_fused_embeddings( # hmm what if I have a document with images at specific positions?
```
isaac-chung, line 203: Sort of like the way it's handled in vlm2vec, where the text would indicate where the image should be interleaved (via `<|image_1|>`)?

mteb/encoder_interface.py
isaac-chung: Hmm if a singular encode is used for both images and text, then type checking is needed? Would that be sort of like the "fused_embedding" method with a bunch of if-statements handling the presence of types?

isaac-chung (commented):
Sorry for the delay. I wonder if this can be easily replicated for other modalities like audio & video?

Pinging @gowitheflow-1998 as well.

mteb/encoder_interface.py
KennethEnevoldsen: yep exactly that - I would say if we want to load it in a low-memory fashion, we would have `Iterable[Document]` - I believe that works with dataloaders?

mteb/encoder_interface.py
KennethEnevoldsen: I think so yes - I also discussed it a bit with Tom Aarsen on slack, seems to be the direction (sentence) transformers are going as well with one "general processing" component (e.g. a tokenizer) followed by a transformer encoders.

The exact input format (document in this case) is a bit unclear.

mteb/encoder_interface.py
```diff
@@ -151,3 +156,56 @@ def convert_conv_history_to_query(conversations: Sequence[Sequence[str]]) -> str
...
+        # If not it might be ideal to redefine Encoder.encode
+        self,
+        images: list[Image.Image] | DataLoader,
+        *, 
+        **kwargs,
+        # removed batch_size, it is not required that it will accept kwargs
+    ) -> np.ndarray:  # added standard output (I believe we actually expect tensors in the code, but would like to be consistent)
+        pass
+
+    def get_text_embeddings( # any reason for this?
```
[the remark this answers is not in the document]
KennethEnevoldsen: I was thinking of defining the different modalities inputs using an [overload](https://peps.python.org/pep-3124/). Haven't used it much though

mteb/encoder_interface.py
KennethEnevoldsen: Yep exactly

mteb/encoder_interface.py
gowitheflow-1998: having a general `encode` sounds good. The encoding of different modalities can happen behind the scenes similar to `get_fused_embedding` where it can take in an image and a text or either.

Samoed:
Some ideas are implemented here https://github.com/embeddings-benchmark/mteb/pull/1583

KennethEnevoldsen:
I will merge this in and continue the work with the correction (there were some failing tests, I fixed those in #1596)

isaac-chung:
Addresses #1341

mteb/encoder_interface.py
isaac-chung: Related to https://github.com/embeddings-benchmark/mteb/issues/1047

PR #1554, opened 2024-12-05
Made result loading more permissive, changed eval splits for HotPotQA and DBPedia

x-tabdeveloping:
Some results were missing from the leaderboard, since task result validation removed them, because some languages were missing from the results.
Validation only throws warnings now, I also removed some splits from `eval_splits` on two tasks, since all models were missing these splits from the results repo.

### DIFF
--- a/mteb/load_results/task_results.py
+++ b/mteb/load_results/task_results.py
@@ -459,7 +459,7 @@ class TaskResult(BaseModel):
         values = []
         for split in splits:
             if split not in self.scores:
-                raise ValueError(f"Split {split} not found in scores")
+                logger.warning(f"Split {split} not found in scores")
 
             for scores in self.scores[split]:
                 eval_langs = scores["languages"]
@@ -511,12 +511,14 @@ class TaskResult(BaseModel):
                 new_scores[split].append(_scores)
                 seen_subsets.add(_scores["hf_subset"])
             if seen_subsets != hf_subsets:
-                raise ValueError(
-                    f"Missing subsets {hf_subsets - seen_subsets} for split {split}"
+                logger.warning(
+                    f"{task.metadata.name}: Missing subsets {hf_subsets - seen_subsets} for split {split}"
                 )
             seen_splits.add(split)
         if seen_splits != set(splits):
-            raise ValueError(f"Missing splits {set(splits) - seen_splits}")
+            logger.warning(
+                f"{task.metadata.name}: Missing splits {set(splits) - seen_splits}"
+            )
         new_res = {**self.to_dict(), "scores": new_scores}
         new_res = TaskResult.from_validated(**new_res)
         return new_res
--- a/mteb/load_results/task_results.py
+++ b/mteb/load_results/task_results.py
@@ -291,10 +291,13 @@ class TaskResult(BaseModel):
                 )
 
         pre_1_11_load = (
-            "mteb_version" in data
-            and data["mteb_version"] is not None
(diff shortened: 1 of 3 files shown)


Samoed (approved):

mteb/load_results/task_results.py
```diff
@@ -511,12 +511,14 @@ def validate_and_filter_scores(self, task: AbsTask | None = None) -> AbsTask:
...
-                    f"Missing subsets {hf_subsets - seen_subsets} for split {split}"
+                logger.warning(
+                    f"{task.metadata.name}: Missing subsets {hf_subsets - seen_subsets} for split {split}"
                 )
             seen_splits.add(split)
         if seen_splits != set(splits):
-            raise ValueError(f"Missing splits {set(splits) - seen_splits}")
+            logger.warning(
+                f"{task.metadata.name}: Missing splits {set(splits) - seen_splits}"
+            )
```
KennethEnevoldsen, line 521: This seems problematic to turn into a warning -  then it would get added to the leaderboard even though not all the results are there?

mteb/tasks/Retrieval/eng/HotpotQARetrieval.py
```diff
@@ -20,7 +20,7 @@ class HotpotQA(AbsTaskRetrieval):
         type="Retrieval",
         category="s2p",
         modalities=["text"],
-        eval_splits=["train", "dev", "test"],
+        eval_splits=["test"],
```
KennethEnevoldsen, line 23: I just double checked that this did not cause any problems with existing benchmark

KennethEnevoldsen (commented):
Would probably do a review on this next time - I do think the warning instead of error is a problem

mteb/load_results/task_results.py
x-tabdeveloping: It wouldn't because it fails at the get_score step. This just allows all splits to be loaded, which is beneficial when you have a benchmark, that only uses that split. (I hope I'm making sense)

x-tabdeveloping: So e.g. `results` repo only has results on the `test` split, but the task description says `eval_splits=["dev", "test"]`. If you throw an error instead of a warning, you don't load the task results, so it will be missing, even if you have a benchmark that only has `eval_splits=["test"]`

mteb/tasks/Retrieval/eng/HotpotQARetrieval.py
x-tabdeveloping: It seemed like an error to include the train split in evaluation by default

KennethEnevoldsen: Well some datasets do use the train split in the evaluation as it is the only one

mteb/load_results/task_results.py
KennethEnevoldsen: check!

mteb/tasks/Retrieval/eng/HotpotQARetrieval.py
x-tabdeveloping: right. As far as I know HotPotQA used the test set in the old benchmark though, right?

Samoed: Yes, you're right. HotPotQA is a retrieval dataset and doesn't require train split

KennethEnevoldsen: Yes no problems here

PR #1558, opened 2024-12-06
Correction of SICK-R metadata

rafalposwiata:
The SICK-R metadata included inappropriate references, namely those for the Polish version of this dataset.


isaac-chung (approved):
Thanks!

Issue #1556, opened 2024-12-05
Track contamination in new leaderboard


Muennighoff:
Just discussed this with @sivareddyg & others, but I think once we have this, many people would be willing to help us do all the annotations. Ideally it is super easy to add contamination annotations via a simple PR

- [x] Add easy way to add training data metadata of models to
- [ ] Modify UI to select whether contaminated models are allowed or not
- [ ] Modify UI to maybe always display a small symbol next to numbers which are contaminated based on the metadata
- [ ] Maybe add small section [here](https://github.com/embeddings-benchmark/mteb?tab=readme-ov-file#usage-documentation) on how to add contamination annotations

KennethEnevoldsen:
So you are suggesting to replace:
```
class ModelMeta(BaseModel):
    name: str | None
    revision: str | None
    ...
    zero_shot_benchmarks: list[str] | None = None
    citation: str | None = None
```

with 

```
class ModelMeta(BaseModel):
    name: str | None
    revision: str | None
    ...
    training_datasets: list[str] | None = None
    citation: str | None = None
```

And rely on users to annotate it? I would be up for that (@x-tabdeveloping this is also more in line with your idea)

We can extrapolate `zero_shot_benchmarks` from training datasets. What do we do for models where the training data is not known?

UI is pretty simple once we have it

x-tabdeveloping:
Yeah, once we've implemented the logic I can add it to the UI

isaac-chung:
Does that mean that #1372 is obsolete? If so we can close it.

KennethEnevoldsen:
Added a PR

Muennighoff:
Here's an example for displaying contamination: https://huggingface.co/spaces/allenai/reward-bench though we probably want to be more finegrained

x-tabdeveloping:
Thanks for implementing the metadata attribute for this.
I have a number of questions and concerns relating to this.

### 1. We have limited information
As it currently stands, only 18 of the 187 models that have associated metadata in MTEB have any annotation regarding what they were trained on, and I suspect that some of these are quite incomplete.

I personally don't think it makes a lot of sense to implement a filter in the leaderboard until we add this data on more models.

I'm also wondering who should be made responsible for doing this. What happens with models, for which this information is not published? Can we push people to give us information about their training data?

### 2. Indirect contamination
[...]

PR #1561, opened 2024-12-07
fix: Add training dataset to model meta

KennethEnevoldsen:
Adresses #1556

## Checklist

### DIFF
--- a/mteb/model_meta.py
+++ b/mteb/model_meta.py
@@ -72,7 +72,7 @@ class ModelMeta(BaseModel):
             in the Latin script.
         use_instructions: Whether the model uses instructions E.g. for prompt-based models. This also include models that require a specific format for
             input such as "query: {document}" or "passage: {document}".
-        training_datasets: A dictionary of datasets that the model was trained on. Names should be names as their appear in `mteb` for example 
+        training_datasets: A dictionary of datasets that the model was trained on. Names should be names as their appear in `mteb` for example
             {"ArguAna": ["test"]} if the model is trained on the ArguAna test set. This field is used to determine if a model generalizes zero-shot to
             a benchmark as well as mark dataset contaminations.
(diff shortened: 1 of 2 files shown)


mteb/model_meta.py
```diff
@@ -97,7 +98,7 @@ class ModelMeta(BaseModel):
     reference: STR_URL | None = None
     similarity_fn_name: DISTANCE_METRICS | None = None
     use_instructions: bool | None = None
-    zero_shot_benchmarks: list[str] | None = None
+    training_datasets: dict[str, list[str]] | None = None
```
isaac-chung, line 101: Will the dataset name be validated somehow, e.g. must match with one of the existing dataset names within MTEB?

If so, where should that validation be?

isaac-chung (approved):
Nice! Just one suggestion, and can be done in a separate PR.

mteb/model_meta.py
Samoed: I think this is hard to validate because there might be datasets intended only for training that overlap with some tasks

isaac-chung: @Samoed By validation, I'm referring whether the entered dataset name is from the list of tasks in MTEB. Right now, this field accepts any dataset name, e.g. "ABCDE", but such a dataset does not exist in MTEB, and I feel that should an invalid entry in the model meta. Could you elaborate what "that overlap with some tasks" means?

Samoed: For example, if a model is trained on the [NLLB](https://huggingface.co/datasets/allenai/nllb) dataset, some bitext mining datasets might potentially leak into the training data, but it's hard to determine which ones. Similarly, if a model is trained on Wikipedia, many datasets could have overlapping examples in the training data. Identifying these overlaps would require extensive testing to pinpoint where the leaks occur.

KennethEnevoldsen: Would be great to have validation, though I could imagine that people would also add datasets that are not in mteb here

PR #1562, opened 2024-12-07
feat(google_models): fix issues and add support for `text-embedding-005` and `text-multilingual-embedding-002`

dbuades:
Hello! Small PR focused on `google_models.py` since we noticed that they were not working properly on the latest versions of MTEB.

This PR introduces a few fixes to make them work again, as well as two new models: `text-embedding-005` and `text-multilingual-embedding-002`.

Fixes #1480

## Changelist

- Use batching to circumvent Vertex AI limits on number of instances and input tokens
- Optionally use `tqdm` to show batching progress
- Load correct `google_task_type`, since we were wrongly passing the `prompt_name` before
- Add support for `text-embedding-005` and `text-multilingual-embedding-002` models
- Fix linting issue in `model2vec_models.py` raised by `make lint`

## Checklist

- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 

### Adding a model checklist

 - [x] I have filled out the ModelMeta object to the extent possible
 - [x] I have ensured that my model can be loaded using
   - [x] `mteb.get_model(model_name, revision)` and
   - [x] `mteb.get_model_meta(model_name, revision)`
 - [x] I have tested the implementation works on a representative set of tasks. -> Results for `CUREv1` will be added to the results repo.

PR #1564, opened 2024-12-07
feat: (cohere_models) cohere_task_type issue, batch requests and tqdm for visualization

dbuades:
As a follow-up to https://github.com/embeddings-benchmark/mteb/pull/1562, Cohere models where lacking batching, which made it difficult to run them on a large corpus. This PR adds such batching as well as `tqdm` for tracking progress (also added to openai_models at the same time).

There is also a fix for loading the correct `cohere_task_type`, since we were wrongly passing the `prompt_name` before as with `google_models`.

## Checklist

### Adding a model checklist

 - [x] I have tested the implementation works on a representative set of tasks. -> Results for CUREv1 will be added to the results repo.

### DIFF
--- a/mteb/models/cohere_models.py
+++ b/mteb/models/cohere_models.py
@@ -144,7 +144,7 @@ class CohereTextEmbeddingModel(Wrapper):
         self,
         sentences: list[str],
         cohere_task_type: str,
-        show_progress_bar: bool,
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen (approved):
great fix - thanks!

PR #1566, opened 2024-12-07
fix(bm25s): search implementation

dbuades:
The `bm25s` implementation for retrieval tasks does not currently work with the latest versions of MTEB. This PR introduces two fixes to restore its functionality:

- Update the argument check in `RetrievalEvaluator` from `self.retriever.model` to `self.retriever.model.model`.
- Adjust the `search` method to account for the fact that `corpus[cid]` may now be a dictionary, removing the need to always wrap it as `{"text": corpus[cid]}`.

These changes were already applied in the results presented in https://github.com/embeddings-benchmark/results/pull/55 to verify that they worked properly. They will not impact older results.

## Checklist


isaac-chung (approved):
Thanks for fixing this!

Issue #1463, opened 2024-11-16
Importing mteb is slow


Muennighoff:
mteb takes 30s while transformers takes 6s (https://colab.research.google.com/drive/166j9HRQF8hAqwiiYpMDvYoznm1v-veOC?usp=sharing). Probably because of the many `*` imports 🤔

KennethEnevoldsen:
Yeah we can def. reduce the number of `*` imports. I could also imagine that it might be the many TaskMetadata objects that take some time as well - we should probably profile it to see where the problem is.

KennethEnevoldsen:
Yes v2.0.0 is the best place to start

Napuh:
At first glance the star imports are not taking that much time to load. Using `py-spy` I can see that `datasets` and `sentence_transformers` are taking the mayority of the time spent importing:

![profile](https://github.com/user-attachments/assets/27d3ff98-f1f0-4c8c-b325-fbd25f3f68c1)

I can get rid of some star imports, but this will not make any effect on import speed :(

Muennighoff:
Hm as mentioned here https://github.com/embeddings-benchmark/mteb/issues/102 we can make sentence-transformers (and also torch) an optional dependency?

Napuh:
Also, on `v2.0.0` i hace not experienced the 30s import time. For me it just took 5-ish seconds.

KennethEnevoldsen:
For me it is still slow on v2.0.0 - So it might depend on versions or python instance -- @Napuh are you using python 3.9?

An optional thing we could do is to move the datasets import into the load_dataset function, but I think a PR adding a new dataset will likely not do this so I don't think it is a viable solution.

We could make sentence transformers an optional dependency but the error message would have to be good to avoid confusion. v2.0.0 is probably the version to do it in

Samoed:
When working on https://github.com/embeddings-benchmark/mteb/pull/1362, I encountered weird issues with the `logging` module due to circular imports. I believe removing `*` imports would resolve the problem. @Napuh

ayush1298:
@Samoed , should we mark this as complete?

Samoed:
It still kind slow, so I don't know. @KennethEnevoldsen What do you think?

KennethEnevoldsen:
So def. possible to speed it 

Adding this [rule](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) could likely help removing a lot of unec. import during runtime.

[...]

PR #1567, opened 2024-12-07
fix: Add documentation

KennethEnevoldsen:
I had some time on a plane so I made a documentation page!! I initially was thinking of moving everything there, but there were too many problems (especially with circular imports encountered when using the autodoc function) so I decided to work only on the API documentation which we currently don't have.

Therefore there are a lot of changes that are currently not used in the docs. The hope is that after this PR we can resolve some of these issues.

Here are some examples of the docs: 

Screenshot 2024-12-07 at 13 07 36

Screenshot 2024-12-07 at 13 06 45

PR #1569, opened 2024-12-08
fix: remove `*` imports

Samoed:
## Checklist

Ref https://github.com/embeddings-benchmark/mteb/issues/1463
Also merged changes from `main` and found 3 datasets that previously never imported:
- Ddisco
- GeorgianSentimentClassification
- WongnaiReviewsClassification


isaac-chung:
Nice! Was wondering actually, if this could be merged to main instead? I don't think there were any major compatible changes? (as in, everything should run the same as before)

mteb/__init__.py
```diff
@@ -2,6 +2,21 @@
...
+    AbsTaskMultilabelClassification,
+    AbsTaskPairClassification,
+    AbsTaskReranking,
+    AbsTaskRetrieval,
+    AbsTaskSpeedTask,
+    AbsTaskSTS,
+    AbsTaskSummarization,
+    MultilingualTask,
+)
 from mteb.benchmarks.benchmarks import (
```
KennethEnevoldsen, line 20: We could remove these

KennethEnevoldsen (commented):
Looks great. I'm very happy about this!

It was a big frustration for me in #1567 so I am very happy to see it. Did you do it all manually?

> Nice! Was wondering actually, if this could be merged to main instead? I don't think there were any major compatible changes? (as in, everything should run the same as before)

Before you could e.g. do 

```
from mteb import load_datasets
```

I believe this will no longer be possible

isaac-chung:
I see a script being used to generate these.  To see changes from this PR, I was switching the commits one at a time. 

Might be nice if the merge from main was separated.

Samoed:
Yes, as @isaac-chung mentioned, I wrote [a script](https://github.com/embeddings-benchmark/mteb/blob/07f139159a19bd10c78b484e5550e8d624deca49/scripts/generate_imports.py) to generate imports for tasks, but for other directories, I did it manually.

isaac-chung (approved):
Thanks for tackling this :) 
Might be worth running the same benchmarks in https://github.com/embeddings-benchmark/mteb/issues/1463 again as a comparison.

Samoed:
Here is profile results of profiling, but it is hard to tell difference 
profile

Issue #1443, opened 2024-11-13
Leaderboard: Add radar chart 


KennethEnevoldsen:
Not planned for the current release of the leaderboard, but it would be nice to have a radio chart for the plots. To make it clear where models excel. The radar plot could include task types, languages and/or domains.

It could either be placed where the description is now (moving the description up) or we could have it below the table.

Example of a radar chart
![radar_chart](https://github.com/user-attachments/assets/4a4d5735-9f4c-4e94-bdf1-7feb6dbe4be8)

A critique of radar chart:
https://blog.scottlogic.com/2011/09/23/a-critique-of-radar-charts.html

There might be better plots available to access model capabilities quickly (though I think radar is a good fit on our case).

isaac-chung:
Would this chart be per-model? I'm trying to imagine where it would fit in the current layout. Maybe a separate tab as a single-model-view would fit this better

KennethEnevoldsen:
Yea, I also considered a tab where you compare models (e.g. 1-4) - otherwise it would get two crowded

x-tabdeveloping:
I'm working on a prototype, will send screenshots once it gets neat enough

x-tabdeveloping:
![image](https://github.com/user-attachments/assets/af613c23-05db-478c-be4a-e7634bc1b257)

Whatcha think about this?

KennethEnevoldsen:
Looks great @x-tabdeveloping - is it possible to remove the shading i think that would make the plot more readable.

I would also change the tabs headers to:

"Performance by Size", "Performance by Task Category"

And then in the plot add "Performance by Task Category for the top-5 models."

Otherwise Wonderful addition.

PR #1570, opened 2024-12-09
fix: Added radar chart displaying capabilities on task types

x-tabdeveloping:
I added a radar chart to a tab you can switch to from the other plot.
It only displays the top 5 models that have been run on all task types to avoid clutter.
I also decided not to display instruction retrieval, since it can be negative and it looks weird.
Here's how it looks:
image

### DIFF
--- a/mteb/leaderboard/figures.py
+++ b/mteb/leaderboard/figures.py
@@ -111,6 +111,8 @@ task_types = [
     "STS",
     "Summarization",
     # "InstructionRetrieval",
+    # Not displayed, because the scores are negative,
+    # doesn't work well with the radar chart.
     "Speed",
 ]
 
--- a/mteb/leaderboard/figures.py
+++ b/mteb/leaderboard/figures.py
@@ -97,3 +97,90 @@ def performance_size_plot(df: pd.DataFrame) -> go.Figure:
         margin=dict(b=20, t=10, l=20, r=10),  # noqa
     )
     return fig
+
+
+TOP_N = 5
+task_types = [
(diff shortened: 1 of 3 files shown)


x-tabdeveloping:
#1443

mteb/leaderboard/figures.py
```diff
@@ -97,3 +97,90 @@ def performance_size_plot(df: pd.DataFrame) -> go.Figure:
...
+    "BitextMining",
+    "Classification",
+    "MultilabelClassification",
+    "Clustering",
+    "PairClassification",
+    "Reranking",
+    "Retrieval",
+    "STS",
+    "Summarization",
+    # "InstructionRetrieval",
```
isaac-chung, line 113: Could include the reason you shared here.

mteb/leaderboard/figures.py
```diff
@@ -97,3 +97,90 @@ def performance_size_plot(df: pd.DataFrame) -> go.Figure:
         margin=dict(b=20, t=10, l=20, r=10),  # noqa
     )
     return fig
+
+
+TOP_N = 5
```
isaac-chung, line 102: May be moved lower and make it depend on something we'll change if we were to increase top N?
```suggestion
TOP_N = len(line_colors)
```

isaac-chung (approved):
Looks great! Added 2 suggestions.

mteb/leaderboard/figures.py
x-tabdeveloping: Sure thing

PR #1572, opened 2024-12-09
fix: no longer using same query text for all of BLINKIT2TMultiChoice

Jamie-Stirling:
## Checklist


gowitheflow-1998:
The only test error here is perhaps due to some unexpected version mismatch. Tagging @isaac-chung for a further look but it’s merged now for running final results.

PR #1576, opened 2024-12-10
[mieb] update script for final re-run

gowitheflow-1998:
## Checklist

re-run script for datasets that needed to be fixed (flickr, r-oxford, r-paris, blink's 4 settings). We are changing r-oxford and r-paris to use the MultiChoice evaluator instead of the retrieval evaluator to match its original setting in the paper, which will be merged later in #1572


gowitheflow-1998:
Just a comment that the test error is perhaps due to some unexpected version mismatch, also seen in #1572. Am merging now and will revisit after rerun.

Issue #1589, opened 2024-12-14
The score_function is cos_sim() by default, when evaluating ColBert model in Retrieval task 


CZH-THU:
In mteb\evaluation\evaluators\RetrievalEvaluator.py:450. The score_function is cos_sim by default which will cause 
"File d:\ProgramData\Anaconda3\envs\embedding\lib\site-packages\mteb\evaluation\evaluators\utils.py:34, in cos_sim(a, b)      32 a_norm = torch.nn.functional.normalize(a, p=2, dim=1)      33 b_norm = torch.nn.functional.normalize(b, p=2, dim=1) ---> 34 return torch.mm(a_norm, b_norm.transpose(0, 1))  RuntimeError: self must be a matrix" when evaluating ColBert model in Retrieval task.
  The score_function should be  max_sim when the model is ColBert. Maybe score_function need be choosed by users.

Samoed:
@sam-hey

sam-hey:
Hello @CZH-THU, passing a ColBERT model directly is not supported and will default to cosine similarity, which results in an error.

You can refer to this example for guidance:
[Using Late Interaction Models for Retrieval](https://github.com/embeddings-benchmark/mteb?tab=readme-ov-file#using-late-interaction-models-for-retrieval)

CZH-THU:
> Hello @CZH-THU, passing a ColBERT model directly is not supported and will default to cosine similarity, which results in an error.
> 
> You can refer to this example for guidance: [Using Late Interaction Models for Retrieval](https://github.com/embeddings-benchmark/mteb?tab=readme-ov-file#using-late-interaction-models-for-retrieval)
I use the demo （https://github.com/embeddings-benchmark/mteb?tab=readme-ov-file#using-late-interaction-models-for-retrieval） code:
'''
from mteb import MTEB
import mteb


colbert = mteb.get_model("colbert-ir/colbertv2.0")
tasks = mteb.get_tasks(tasks=["NFCorpus"], languages=["eng"])

eval_splits = ["test"]

evaluation = MTEB(tasks=tasks)

evaluation.run(
    colbert,
    eval_splits=eval_splits,
    corpus_chunk_size=500,
)
'''
but still result in error,I have figure it out by changing score_function.Maybe the demo code need be  modified

PR #1592, opened 2024-12-14
doc: colbert add score_function & doc section

sam-hey:
Closes: https://github.com/embeddings-benchmark/mteb/issues/1589

Improve the documentation and add information about the PLAID Index in PyLate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,7 +317,7 @@ evaluation.run(
 </details>
 
 <details>
-  <summary>  Saving retrieval task predictions </summary>
+  <summary> Late Interaction (ColBERT) </summary>
 
 ### Using Late Interaction models for retrieval
 
@@ -336,10 +336,16 @@ evaluation = MTEB(tasks=tasks)
 evaluation.run(
     colbert,
     eval_splits=eval_splits,
+    score_function="max_sim", 
     corpus_chunk_size=500,
 )
 ```
-This implementation uses the MaxSim operation to calculate the similarity between the sentences. Because of this, the number of embeddings processed is higher, which might lead to higher resource usage. Therefore, you can lower the `corpus_chunk_size` to reduce the load on resources.
+This implementation employs the MaxSim operation to compute the similarity between sentences. While MaxSim provides high-quality results, it processes a larger number of embeddings, potentially leading to increased resource usage. To manage resource consumption, consider lowering the `corpus_chunk_size` parameter.
+
+Currently, this implementation does not utilize an index. However, support for indexing will be introduced once PyLate adds compatibility with the PLAID Index. You can track the progress of this feature on the PyLate GitHub repository: [Issue #72](https://github.com/lightonai/pylate/issues/72).
+
+<details>
+  <summary>  Saving retrieval task predictions </summary>
 
 ### Saving retrieval task predictions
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -336,10 +336,16 @@ evaluation = MTEB(tasks=tasks)
...
     colbert,
     eval_splits=eval_splits,
+    score_function="max_sim", 
     corpus_chunk_size=500,
 )
 ```
-This implementation uses the MaxSim operation to calculate the similarity between the sentences. Because of this, the number of embeddings processed is higher, which might lead to higher resource usage. Therefore, you can lower the `corpus_chunk_size` to reduce the load on resources.
+This implementation employs the MaxSim operation to compute the similarity between sentences. While MaxSim provides high-quality results, it processes a larger number of embeddings, potentially leading to increased resource usage. To manage resource consumption, consider lowering the `corpus_chunk_size` parameter.
+
+Currently, this implementation does not utilize an index. However, support for indexing will be introduced once PyLate adds compatibility with the PLAID Index. You can track the progress of this feature on the PyLate GitHub repository: [Issue #72](https://github.com/lightonai/pylate/issues/72).
```
KennethEnevoldsen, line 345: I would not add this in the docs - but rather track this in an issue

```suggestion
```

KennethEnevoldsen (approved):
Minor changes - otherwise good

README.md
```diff
@@ -336,10 +336,16 @@ evaluation = MTEB(tasks=tasks)
 evaluation.run(
     colbert,
     eval_splits=eval_splits,
+    score_function="max_sim",
```
Samoed, line 339: This would work, but I think we need a more general solution to specify `score_function` to prevent potential future issues with scores

KennethEnevoldsen: Agree -- but probably outside the scope of this PR

sam-hey: I completely agree with you. In my opinion, it's a bit surprising that `ModelMeta.similarity_fn_name` isn't being utilized.

Priorities:
1. Pass `score_function` directly to `run()`
2. Utilize `ModelMeta.similarity_fn_name`
3. Default to `RetrievalEvaluator.score_function: str = "cos_sim"`

However, this is currently outside the scope for me...

Samoed: Parameters in `ModelMeta` are only for metadata and don't play a role in processing. I think the solution would be to pass `score_function` to the wrapper, and then `RetrievalEvaluator` can use the parameter from the wrapper

README.md
```diff
@@ -336,10 +336,15 @@ evaluation = MTEB(tasks=tasks)
...
     eval_splits=eval_splits,
+    score_function="max_sim", 
     corpus_chunk_size=500,
 )
 ```
-This implementation uses the MaxSim operation to calculate the similarity between the sentences. Because of this, the number of embeddings processed is higher, which might lead to higher resource usage. Therefore, you can lower the `corpus_chunk_size` to reduce the load on resources.
+This implementation employs the MaxSim operation to compute the similarity between sentences. While MaxSim provides high-quality results, it processes a larger number of embeddings, potentially leading to increased resource usage. To manage resource consumption, consider lowering the `corpus_chunk_size` parameter.
+
+
+<details>
```
isaac-chung, line 346: Need to close it before opening a new one.

```suggestion
</details>

<details>
```

isaac-chung (commented):
One small thing. Otherwise it looks good!

README.md
KennethEnevoldsen: > it's a bit surprising that ModelMeta.similarity_fn_name isn't being utilized.

We would love to switch to that one and would encourage a PR for this. It was only recently added to haven't been integrated throughout.

isaac-chung (approved):

PR #1594, opened 2024-12-14
Feat: add support for scoring function

Samoed:
## Checklist

Currently, there is no way to specify the scoring function except by passing it as `kwargs`, but `ColBERT` models require it.

Ref https://github.com/embeddings-benchmark/mteb/pull/1592#discussion_r1885256342
Closes https://github.com/embeddings-benchmark/mteb/issues/1589

@sam-hey

### DIFF
--- a/mteb/evaluation/evaluators/RetrievalEvaluator.py
+++ b/mteb/evaluation/evaluators/RetrievalEvaluator.py
@@ -15,7 +15,7 @@ import tqdm
 from sentence_transformers import CrossEncoder, SentenceTransformer
 
 from mteb.encoder_interface import Encoder, PromptType
-from mteb.model_meta import ModelMeta
+from mteb.model_meta import DISTANCE_METRICS, ModelMeta
 
 from .Evaluator import Evaluator
 from .utils import (
@@ -79,12 +79,12 @@ class DenseRetrievalExactSearch:
             encode_kwargs["convert_to_tensor"] = True
 
         self.score_functions = {
-            "cos_sim": cos_sim,
+            "cosine": cos_sim,
             "dot": dot_score,
             "max_sim": max_sim,
         }
         self.score_function_desc = {
-            "cos_sim": "Cosine Similarity",
+            "cosine": "Cosine Similarity",
             "dot": "Dot Product",
             "max_sim": "Max Similarity",
         }
@@ -112,7 +112,7 @@ class DenseRetrievalExactSearch:
         corpus: dict[str, dict[str, str]],
         queries: dict[str, str | list[str]],
         top_k: int,
-        score_function: str,
+        score_function: DISTANCE_METRICS,
         task_name: str,
         instructions: dict[str, str] | None = None,
         request_qid: str | None = None,
@@ -124,7 +124,7 @@ class DenseRetrievalExactSearch:
         # Returns a ranked list with the corpus ids
         if score_function not in self.score_functions:
             raise ValueError(
-                f"score function: {score_function} must be either (cos_sim) for cosine similarity or (dot) for dot product"
+                f"score function: {score_function} must be either (cosine) for cosine similarity, (dot) for dot product or (max_sim) for max similarity"
             )
 
         logger.info("Encoding Queries.")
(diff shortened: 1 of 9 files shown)


KennethEnevoldsen (commented):
I really appreciate this PR, but I believe this should be done using:

https://github.com/embeddings-benchmark/mteb/blob/b466051b952ade5bb60e0f25321fed0ebb73cb50/mteb/encoder_interface.py#L71

Samoed:
I've aligned `RetrievalEvaluator` with the other evaluators. It now uses `model.similarity`.

mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -191,34 +167,42 @@ def search(
                     self.corpus_embeddings[request_qid].append(sub_corpus_embeddings)
 
             # Compute similarites using either cosine-similarity or dot product
```
KennethEnevoldsen, line 169:
```suggestion
            # Compute similarites using self defined similarity otherwise default to cosine-similarity
```

mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -191,34 +167,42 @@ def search(
...
+                    float(self.model.similarity(e1, e2))
+                    for e1, e2 in zip(query_embeddings, sub_corpus_embeddings)
+                )
+            is_nan = torch.isnan(similarity_scores)
             if is_nan.sum() > 0:
                 logger.warning(
                     f"Found {is_nan.sum()} NaN values in the similarity scores. Replacing NaN values with -1."
                 )
-            cos_scores[is_nan] = -1
+            similarity_scores[is_nan] = -1
```
KennethEnevoldsen, line 181: I might have actually removed this over in v2.0.0 (it is not needed)

mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -112,21 +100,12 @@ def search(
         corpus: dict[str, dict[str, str]],
         queries: dict[str, str | list[str]],
         top_k: int,
-        score_function: str,
```
KennethEnevoldsen, line 115: breaking? should we move to v2.0.0?

KennethEnevoldsen (approved):
I like this a lot better. Should we move it to v2.0.0?

mteb/evaluation/evaluators/RetrievalEvaluator.py
Samoed: I don't think this is very breaking because, in the implementation from the main branch, `score_function` can only be passed with `evaluator.run`, and I doubt many users have specified this.

Samoed:
This PR can be moved to 2.0, but this change mainly benefits `ColBert` and I don't think it will break anything. However, if you believe it's breaking, I can move it to 2.0.

KennethEnevoldsen:
So technically breaking, in practice not breaking. I agree with that. Let's keep it as it is and merge this in.

Samoed:
Also found that `BitextMiningEvaluator` uses `cosine` as score function. I think this also should be changed
https://github.com/embeddings-benchmark/mteb/blob/992b20b7079bba96fee913181e28292757c0087b/mteb/evaluation/evaluators/BitextMiningEvaluator.py#L112
Because `PairClassification` , `STSEvaluator` and `SummarizatonEvaluator` already can use `model.similarity`

KennethEnevoldsen:
Would love to add that as well. Feel free to add it in a separate PR and merge this one though

sam-hey:
@Samoed, if I’m not mistaken, the ability to use the `Dot product` seems to have been lost with this PR. Could you please confirm?

Samoed:
Yes, but no model was using it, and to use it required passing it as a parameter, which wasn’t clear and would make it hard to reevaluate results. Currently, only Contriever uses `dot` similarity, but that's in the `v2.0.0` branch. https://github.com/embeddings-benchmark/mteb/pull/1749

PR #1601, opened 2024-12-16
Leaderboard: Refined plots

x-tabdeveloping:
I added a guide for embedding size as requested, along with smaller adjustments to plotting behaviour.
image


isaac-chung (approved):
🚀

PR #1607, opened 2024-12-17
[MIEB] Switch to ViDoRe BEIR version

izhx:
In ViDoRe code, they have some corpus deduplication operation (by image_filename).
https://github.com/illuin-tech/vidore-benchmark/blob/469665d6286f04a4ae09ed7c31dc66dfb44e727a/src/vidore_benchmark/retrievers/vision_retriever.py#L119

However, this deduplication was not implemented in MIEB's ViDoRe data loading process, resulting in inconsistent corpus size. This inconsistency could significantly affect the accuracy of evaluation results.

I switched to ViDoRe BEIR version.
https://huggingface.co/collections/vidore/vidore-benchmark-beir-672fd04012c14cc4e162f2bc

~~Therefore, I added deduplication code to ensure consistency with the official ViDoRe release.~~

~~For example, the corpus size of `VidoreTatdqaRetrieval` should be `277`, not `1,663` as previously shown.
Refer to https://huggingface.co/datasets/vidore/tatdqa_test_beir~~

## Checklist

### DIFF
--- a/mteb/tasks/Image/Any2AnyRetrieval/eng/VidoreBenchRetrieval.py
+++ b/mteb/tasks/Image/Any2AnyRetrieval/eng/VidoreBenchRetrieval.py
@@ -1,54 +1,13 @@
 from __future__ import annotations
 
-from datasets import load_dataset
+from datasets import load_dataset, Dataset
 
 from mteb.abstasks.Image.AbsTaskAny2AnyRetrieval import AbsTaskAny2AnyRetrieval
 from mteb.abstasks.TaskMetadata import TaskMetadata
 
 
-def _load_data(path: str, splits: str, cache_dir: str = None, revision: str = None):
-    corpus = {}
-    queries = {}
-    relevant_docs = {}
-
-    dataset = load_dataset(
-        path,
-        cache_dir=cache_dir,
-        revision=revision,
-    )
-
-    for split in splits:
-        split_dataset = dataset[split]
-        split_dataset = split_dataset.rename_column("query", "text")
-        corpus[split] = split_dataset.map(
-            lambda x, idx: {
-                "id": f"corpus-{split}-{idx}",
-                "modality": "image",
-                "text": None,
-            },
-            with_indices=True,
-        )
-
-        queries[split] = split_dataset.map(
-            lambda x, idx: {
-                "id": f"query-{split}-{idx}",
-                "image": None,
-                "modality": "text",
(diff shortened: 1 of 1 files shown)


isaac-chung (commented):
> I would also like to flag that the query texts in VidoreTabfquadRetrieval and VidoreShiftProjectRetrieval are in French rather than English. But the document screenshot are in English.
So eval_langs=["eng-Latn"], appears to be inappropriate.

In that case, we should move those tasks separately into the multilingual folder within Any2AnyRetrieval. Also, those tasks can also inherit from `MultilingualTask` as they are cross lingual tasks. [This task](https://github.com/embeddings-benchmark/mteb/blob/b81b584ceb1bd8a42a676482edcc19c90de75cb1/mteb/tasks/Retrieval/multilingual/CrossLingualSemanticDiscriminationWMT19.py#L33) is a good example.

izhx:
> > I would also like to flag that the query texts in VidoreTabfquadRetrieval and VidoreShiftProjectRetrieval are in French rather than English. But the document screenshot are in English.
> > So eval_langs=["eng-Latn"], appears to be inappropriate.
> 
> In that case, we should move those tasks separately into the multilingual folder within Any2AnyRetrieval. Also, those tasks can also inherit from `MultilingualTask` as they are cross lingual tasks. [This task](https://github.com/embeddings-benchmark/mteb/blob/b81b584ceb1bd8a42a676482edcc19c90de75cb1/mteb/tasks/Retrieval/multilingual/CrossLingualSemanticDiscriminationWMT19.py#L33) is a good example.

Thanks for the suggestion.

It contains 8 English subsets and 2 French subsets, I'm not sure if the "one subset per language" approach for multilingual tasks would be suitable here.

As for the French subsets, I'm not sure if they can be considered truly monolingual, since the text in the images is in English.

gowitheflow-1998 (approved):

gowitheflow-1998:
thanks for pointing the corpus problem out! Switching to their deduplicated BEIR version makes a lot of sense. For the language part, we have noticed and corrected in the paper to release and haven't got to update. Looks like for shift project, most images in the corpus are French texts; Tabfquad contains a mixture of English texts and French texts in the images. In general I think it makes sense to put these two in a separate French folder and adjust the metadata?

gowitheflow-1998:
will merge this and adjust metadata in a separate PR

PR #1613, opened 2024-12-18
Add IBM Granite Embedding Models

aashka-trivedi:
## Checklist

### Adding a model checklist


mteb/models/ibm_granite_models.py
```diff
@@ -0,0 +1,106 @@
...
+    "jpn_Latn",
+    "kor_Latn",
+    "nld_Latn",
+    "por_Latn",
+    "zho_Hant",
+    "zho_Hans",
+]
+
+
+granite_107m_multilingual = ModelMeta(
```
isaac-chung, line 24: Could you please fill out adapted_from and superseded_by in the [ModelMeta](https://github.com/embeddings-benchmark/mteb/blob/48cb97d1e75cdf8260283d6628ac515141ed1e92/mteb/model_meta.py#L81) as well? (same for all below).
```
adapted_from: Name of the model from which this model is adapted from. For quantizations, fine-tunes, long doc extensions, etc.
```

isaac-chung (commented):
Thanks for adding this! Just a small comment. Looks good otherwise.

mteb/models/ibm_granite_models.py
aashka-trivedi: Thanks for the review! The base model is not a public model (they are RoBERTa-like models we trained internally at IBM Research), which is why I left that field empty. Do you have a recommendation of what I should add here, considering it cannot be accessed? Ty!

isaac-chung: `None` is fine here for both fields :)

aashka-trivedi: Makes sense, Done!

isaac-chung (approved):
Thanks again! Will merge.

PR #1616, opened 2024-12-20
fix: set `use_instructions` to True in models using prompts

dbuades:
This PR sets the `use_instructions` parameter to True for models that use prompts.

I've also included Arctic models because, while they don’t explicitly mention prompts in their MTEB implementation, they add the prompts directly in their `config_sentence_transformers.json`, so SentenceTransformers uses them automatically (example [here](https://huggingface.co/Snowflake/snowflake-arctic-embed-m-v2.0/blob/main/config_sentence_transformers.json))

Additional context for this PR can be found in [this thread](https://github.com/embeddings-benchmark/results/pull/66#discussion_r1879811928)

A follow up PR in the results repo updates the corresponding `model_meta.json` values for these specific models and revisions.

## Checklist

### DIFF
--- a/mteb/models/arctic_models.py
+++ b/mteb/models/arctic_models.py
@@ -100,7 +100,7 @@ arctic_embed_xs = ModelMeta(
     license="apache-2.0",
     reference="https://huggingface.co/Snowflake/snowflake-arctic-embed-xs",
     similarity_fn_name="cosine",
-    use_instructions=False,
+    use_instructions=True,
     adapted_from="sentence-transformers/all-MiniLM-L6-v2",
     superseded_by=None,
 )
@@ -125,7 +125,7 @@ arctic_embed_s = ModelMeta(
     license="apache-2.0",
     reference="https://huggingface.co/Snowflake/snowflake-arctic-embed-s",
     similarity_fn_name="cosine",
-    use_instructions=False,
+    use_instructions=True,
     adapted_from="intfloat/e5-small-unsupervised",
     superseded_by=None,
 )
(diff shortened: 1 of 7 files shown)


mteb/models/arctic_models.py
```diff
@@ -100,7 +100,7 @@
     license="apache-2.0",
     reference="https://huggingface.co/Snowflake/snowflake-arctic-embed-xs",
     similarity_fn_name="cosine",
-    use_instructions=False,
+    use_instructions=True,
```
Samoed, line 103: I don't see prompts being passed to this model

mteb/models/voyage_models.py
Samoed, line 1: `voyage_large_2_instruct` is using prompts too

mteb/models/arctic_models.py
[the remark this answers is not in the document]
dbuades: They do use them, it's just that they define the prompts directly in the `config_sentence_transformers.json` and SentenceTransformers reads it.

An example for `arctic-embed-m-v2.0` is [here](https://huggingface.co/Snowflake/snowflake-arctic-embed-m-v2.0/blob/main/config_sentence_transformers.json) and for `arctic-embed-m` is [here](https://huggingface.co/Snowflake/snowflake-arctic-embed-m/blob/main/config_sentence_transformers.json).

It's a pretty handy feature of SentenceTransformers !

mteb/models/voyage_models.py
dbuades: I agree, but it was already set to True, so no change was required.

Samoed: Missed somehow

mteb/models/arctic_models.py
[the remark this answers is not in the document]
Samoed: Yes, I know. This is very helpful, but most models use different names for prompts, which don’t match. However, in this case, it works.

mteb/models/voyage_models.py
dbuades: No worries, happened to me too when doing the PR, since it was the only model for which it was already set to True

dbuades:
Thank you for the quick review @Samoed !

Samoed (approved):

PR #1584, opened 2024-12-12
Feat: Evaluate missing languages

Samoed:
## Checklist
- [X] Run tests locally to make sure nothing is broken using `make test`. 
- [X] Run the formatter to format the code using `make lint`. 

Continue ideas from https://github.com/embeddings-benchmark/mteb/pull/1525

If a missing split is present in the existing results, return its results without running the evaluation.

PR #1617, opened 2024-12-20
fix: override existing results

Samoed:
## Checklist

Currently, if `overwrite` is passed with existing results, it won't overwrite them, and everything will be skipped.

This was the reason of https://github.com/embeddings-benchmark/mteb/pull/1584#issuecomment-2541310911

### DIFF
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -470,22 +470,7 @@ class MTEB:
 
             existing_results = None
             save_path = None
-
-            if output_path:
-                save_path = output_path / f"{task.metadata.name}{task.save_suffix}.json"
-                if save_path.exists():
-                    existing_results = TaskResult.from_disk(save_path)
-
-                    if not overwrite_results:
-                        logger.info(
-                            f"{task.metadata.name} results already exists. Loading results from disk."
-                            f" Set overwrite_results=True to overwrite or `--overwrite`."
-                        )
-                        evaluation_results.append(existing_results)
-                        del self.tasks[0]  # empty memory
-                        continue
-
-            # Unified call to get missing splits and subsets
+            final_splits_to_run = task_eval_splits
             missing_evaluations = self._get_missing_evaluations(
                 existing_results,
                 task_eval_splits,
@@ -493,16 +478,40 @@ class MTEB:
                 eval_subsets,
             )
 
-            # Determine final splits to run
-            final_splits_to_run = []
-            # We need to run any split that is fully missing or has missing subsets
-            for sp, info in missing_evaluations.items():
-                if (
-                    info["whole_split_missing"]
-                    or info["missing_subsets"]
-                    or overwrite_results
-                ):
-                    final_splits_to_run.append(sp)
+            if output_path:
+                save_path = output_path / f"{task.metadata.name}{task.save_suffix}.json"
+                if save_path.exists():
+                    existing_results = TaskResult.from_disk(save_path)
+
+                    # Unified call to get missing splits and subsets
+                    missing_evaluations = self._get_missing_evaluations(
+                        existing_results,
+                        task_eval_splits,
+                        task_subsets,
+                        eval_subsets,
+                    )
+
+                    # Determine final splits to run
+                    final_splits_to_run = []
+                    # We need to run any split that is fully missing or has missing subsets
+                    for sp, info in missing_evaluations.items():
+                        if (
+                                info["whole_split_missing"]
+                                or info["missing_subsets"]
+                        ):
+                            final_splits_to_run.append(sp)
+
+                    if overwrite_results:
+                        final_splits_to_run = task_eval_splits
+
+                    if not overwrite_results and len(final_splits_to_run) == 0:
+                        logger.info(
+                            f"{task.metadata.name} results already exists. Loading results from disk."
+                            f" Set overwrite_results=True to overwrite or `--overwrite`."
+                        )
+                        evaluation_results.append(existing_results)
+                        del self.tasks[0]  # empty memory
+                        continue
 
             # If no splits need to be run and results exist, skip
             if not final_splits_to_run:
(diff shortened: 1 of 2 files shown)


mteb/evaluation/MTEB.py
```diff
@@ -527,8 +532,15 @@ def run(
                     # Determine subsets to run for this split
                     # If the whole split is missing, run all required subsets
                     # If only some subsets are missing, run only those
-                    subsets_to_run = info["missing_subsets"]
-                    if info["whole_split_missing"] and task_subsets is None:
+                    subsets_to_run = (
+                        info["missing_subsets"]
+                        if not overwrite_results
+                        else task_subsets
```
isaac-chung, line 538: Judging from the test failures, maybe this condition needs to be reverted?

Samoed: It seems, that test were incorrect a bit, because they use `overwrite_results=True`

isaac-chung: `overwrite_results` shouldn't affect how splits are determined, so those tests should pass whether `overwrite_results` is True or False, right?

tests/test_evaluation/test_split_evaluation.py
```diff
@@ -245,7 +238,6 @@ def test_multilingual_one_missing_split_no_missing_lang(
         output_folder=str(tmp_path / "partial_langs_partial_splits"),
         verbosity=2,
         eval_subsets=["eng", "fra"],
-        overwrite_results=True,
```
isaac-chung, line 248: This should not affect how missing splits or langs are selected.

isaac-chung (changes requested):
I'd suggest adding a test case in missing splits/langs where `overwrite_results=True`, as that should not affect how splits and langs are selected.

Samoed:
I've updated the test as you suggested. Previously, the problem was that when tasks were run a second time with the `overwrite` option, it only checked if one of the selected splits had been run. However, this behavior was incorrect, because all splits should be run with `overwrite`.

mteb/evaluation/MTEB.py
```diff
@@ -470,35 +470,46 @@ def run(
...
+
+                    # Determine final splits to run
+                    final_splits_to_run = []
+                    # We need to run any split that is fully missing or has missing subsets
+                    for sp, info in missing_evaluations.items():
+                        if info["whole_split_missing"] or info["missing_subsets"]:
+                            final_splits_to_run.append(sp)
+
+                    if overwrite_results:
+                        final_splits_to_run = task_eval_splits
```
isaac-chung, line 502: `final_splits_to_run = task_eval_splits` is already called on line 473. Maybe only run this part (L494-L499) when `if not overwrite_results`?

mteb/evaluation/MTEB.py
```diff
@@ -513,7 +524,7 @@ def run(
 
             try:
                 task.check_if_dataset_is_superseded()
-                task.load_data(eval_splits=task_eval_splits, **kwargs)
+                task.load_data(**kwargs)
```
isaac-chung, line 527: Could you please help me understand why this is not needed anymore?

tests/test_evaluation/test_split_evaluation.py
```diff
@@ -305,3 +295,70 @@ def test_multilingual_one_missing_lang_in_one_split(
...
+        model,
+        eval_splits=["val"],
+        output_folder=str(tmp_path / "all_splits_evaluated_with_overwrite"),
+        verbosity=2,
+        overwrite_results=True,
+    )
+    assert "MockRetrievalTask" == results2[0].task_name
+    last_evaluated_splits = evaluation.get_last_evaluated_splits()
+    assert set(last_evaluated_splits["MockRetrievalTask"]) == {"val"}
+    assert len(last_evaluated_splits["MockRetrievalTask"]) == 1
```
isaac-chung, line 325: Let's swap the order here - check len first before asserting values. Same above.

tests/test_evaluation/test_split_evaluation.py
```diff
@@ -305,3 +295,70 @@ def test_multilingual_one_missing_lang_in_one_split(
     # output merged result with previous results
     assert results[0].scores.keys() == {"test", "val"}
     assert len(results[0].scores["test"]) == 2
+
+
+def test_all_splits_evaluated_with_overwrite(model, tasks, tmp_path):
```
isaac-chung, line 300: What does "all splits" mean here? There are test and val but only val is run here. 
>because all splits should be run with overwrite

Based on this comment, is it correct that "all" means all specified splits in the run method, and not all splits available in the task metadata's eval_splits?

Perhaps this can be spelled out, or preferably shown in additional test cases, where the 1st and 2nd runs have different `eval_splits`

isaac-chung (commented):
Thanks for iterating! Got a few comments and questions.

mteb/evaluation/MTEB.py
Samoed: It is not used in dataset loading

tests/test_evaluation/test_split_evaluation.py
Samoed: > means all specified splits in the run method,

Yes, all splits specified in `evaluation.run`

mteb/evaluation/MTEB.py
Samoed: Yes, but in the test, `last_evaluated_splits` was being checked, and with `overwrite_results`, all selected splits and subsets should be reevaluated. This caused the test to fail. For example, instead of just `val` in `last_evaluated_splits`, it should include both `val` and `test`. However, the splits were selected correctly.

mteb/evaluation/MTEB.py
```diff
@@ -470,35 +470,46 @@ def run(
 
             existing_results = None
             save_path = None
+            final_splits_to_run = task_eval_splits
+            missing_evaluations = self._get_missing_evaluations(
+                existing_results,
+                task_eval_splits,
+                task_subsets,
+                eval_subsets,
+            )
```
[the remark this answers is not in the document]
Samoed: If both `task_eval_splits` and `task_subsets` are `None`, all splits will be selected, but filtering for splits and subsets is still necessary. I think calling the same function is easier to understand than adding more complex logic.

isaac-chung (approved):
Thanks for taking the time for my questions. I feel we're in a good place and can leave the rest for future improvement. Let's merge :)

Samoed:
Thank you for review!

PR #1621, opened 2024-12-22
fix: GermanDPR Dataset Causes Cross-Encoder Failure Due to Unexpected dict

### LINKED ISSUE #1609: GermanDPR Dataset Causes Cross-Encoder Failure Due to Unexpected dict
When using the GermanDPR dataset with a CrossEncoder, the dataset is returning a dict instead of a str. This results in an error because the CrossEncoder expects text data as a string.

The following error is raised when processing the dataset:
```
/sentence_transformers/cross_encoder/CrossEncoder.py", line 170, in smart_batching_collate_text_only
    texts[idx].append(text.strip())
AttributeError: 'dict' object has no attribute 'strip' 
```

I see two potential ways to address this issue:

1. Modify the GermanDPR Dataset:
Combine title and formatted_content fields from the dict into a single str before passing it to the CrossEncoder.
Example: `f"{title} {formatted_content}"`
2. Update CrossEncoder Logic:
Add a check in CrossEncoder to handle cases where a dict is passed instead of a str.
If a dict is detected, convert it to a str within the CrossEncoder.

https://github.com/embeddings-benchmark/mteb/blob/b81b584ceb1bd8a42a676482edcc19c90de75cb1/mteb/tasks/Retrieval/deu/GermanDPRRetrieval.py#L57

-- 

https://github.com/embeddings-benchmark/mteb/blob/b81b584ceb1bd8a42a676482edcc19c90de75cb1/mteb/evaluation/evaluators/RetrievalEvaluator.py#L311:L320

KennethEnevoldsen:
Fixes #1609

## Checklist

### DIFF
--- a/mteb/tasks/Retrieval/deu/GermanDPRRetrieval.py
+++ b/mteb/tasks/Retrieval/deu/GermanDPRRetrieval.py
@@ -83,6 +83,7 @@ class GermanDPR(AbsTaskRetrieval):
             )
             corpus.update(neg_docs)
             relevant_docs[q_id] = {k: 1 for k in pos_docs}
+        corpus = {doc["id"]: doc.get("title", "") + " " + doc["text"] for doc in corpus}
         self.queries = {self._EVAL_SPLIT: queries}
         self.corpus = {self._EVAL_SPLIT: corpus}
         self.relevant_docs = {self._EVAL_SPLIT: relevant_docs}


Samoed (approved):
For now I think this is good, but I think this should be added to `AbsRetrievalTask` to avoid future problems

Issue #1600, opened 2024-12-16
Investigate performance discrepancies in gte-Qwen and NV-embed models


isaac-chung:
From https://github.com/embeddings-benchmark/mteb/pull/1436

AlexeyVatolin:
Hello,

I conducted a comparison of the models using the examples provided in the `readme.md` file for each model. Here's a summary of my findings:
- Alibaba-NLP/gte-Qwen2-7B-instruct
- Alibaba-NLP/gte-Qwen1.5-7B-instruct
- Alibaba-NLP/gte-Qwen2-1.5B-instruct
- Linq-AI-Research/Linq-Embed-Mistral

   For these models, I found that all three implementations (i.e., Transformers AutoModel, sentence_transformers, and mteb) are exactly the same. This consistency is great to see.

   - nvidia/NV-Embed-v2
   - nvidia/NV-Embed-v1
   In these cases, the official implementation of Transformers AutoModel differs from the official sentence_transformers implementation, which is unexpected. The implementation in mteb aligns completely with sentence_transformers.

I also wanted to share the code I used for this comparison: [View the Gist](https://gist.github.com/AlexeyVatolin/1569b47e186d40bac2514f82a6ee8925)

Please note that questions regarding the correctness of prompt usage were not within the scope of this study. However, it does highlight that the models added to mteb are correctly implemented.

AlexeyVatolin:
Qwen model repository includes a [script](https://huggingface.co/Alibaba-NLP/gte-Qwen2-7B-instruct/blob/main/scripts/eval_mteb.py) to calculate scores for their models on the MTEB benchmark. I ran this script on the same tasks covered in my pull request.

The results from the original script are, in most cases, worse than those reported on the leaderboard and also fall short when compared to results obtained using the code from the MTEB models.
Here is my command to run this script
```
OPENBLAS_NUM_THREADS=8 python scripts/eval_mteb.py -m Alibaba-NLP/gte-Qwen2-1.5B-instruct --output_dir results_qwen_2_1.5b_eval_mteb --task mteb
```

Additionally, there is a open [discussion](https://huggingface.co/Alibaba-NLP/gte-Qwen2-7B-instruct/discussions/35) about this on the Qwen model page.

## Classification
|                         | source          |   AmazonCounterfactualClassification |   EmotionClassification |   ToxicConversationsClassification |
|:------------------------|:----------------|-------------------------------------:|------------------------:|-----------------------------------:|
[...]

PR #1637, opened 2024-12-27
Correction of discrepancies for gte-Qweb model

AlexeyVatolin:
#1600

## Checklist

### DIFF
--- a/mteb/models/gte_models.py
+++ b/mteb/models/gte_models.py
@@ -2,12 +2,21 @@ from __future__ import annotations
 
 from functools import partial
 
+import torch
+
+from mteb.encoder_interface import PromptType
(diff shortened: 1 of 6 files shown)


KennethEnevoldsen (approved):
Thanks for taking the time on this. The new results looks promising. Do we want to rerun this to update the results on the leaderboard?

AlexeyVatolin:
@KennethEnevoldsen, I'm not sure if a rerun is required. The current results from MTEB are almost identical to the leaderboard. It is unlikely that the model will change its position in the leaderboard after the rerun. Recalculating the scores makes sense if we want the implementation in MTEB to match the leaderboard 100%

KennethEnevoldsen:
alright then I would just leave it.

PR #1633, opened 2024-12-26
[mieb] Task filtering by modality supported by models

isaac-chung:
Fixes #1632 

- Added ModelMeta.modalities, which defaults to `["text"]` to be backwards compatible. VLMs will have `["image","text"]`.
- Added logic in `MTEB.run` for exact match between task and model modalities
- Added modality model meta to relevant models

Here is the new log, showing that moco only supports images.
```
## Evaluating 1 tasks:
Any2AnyRetrieval
    - VidoreSyntheticDocQAAIRetrieval, t2i


INFO:mteb.evaluation.MTEB:

********************** Evaluating VidoreSyntheticDocQAAIRetrieval **********************
INFO:mteb.evaluation.MTEB:nyu-visionx/moco-v3-vit-b only supports ['image'], but the task modalities are ['text', 'image'].
```

## Checklist

- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the formatter to format the code using `make lint`. 

cc @Muennighoff @gowitheflow-1998

PR #1640, opened 2024-12-29
[mieb] model task modalities matching logic

gowitheflow-1998:
In #1633, the skipping of image&text tasks for image-only models was added but image-only tasks are now being skipped for image-text models as well. Fixing this with a subset logic.

## Checklist


isaac-chung (approved):
Good catch! Sorry about that.

PR #1642, opened 2024-12-30
Fix output_folder for co2 evaluation

Muennighoff:
lol surprised we had this bug; it literally errors out for me 😁


Samoed:
Previously, I tried to remove the `output_folder` parameter and pass it through `kwargs`, since it’s only needed for retrieval. However, it seems that during refactoring, I accidentally removed it.

Muennighoff:
No worries! Thanks for merging 🚀

Issue #1652, opened 2025-01-01
voyageai/voyage-3 fails on RuBQReranking


Muennighoff:
2025-01-01 03:02:01.019970: E external/local_xla/xla/stream_executor/cuda/cuda_fft.cc:485] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2025-01-01 03:02:01.034005: E external/local_xla/xla/stream_executor/cuda/cuda_dnn.cc:8454] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2025-01-01 03:02:01.037988: E external/local_xla/xla/stream_executor/cuda/cuda_blas.cc:1452] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
INFO:mteb.cli:Running with parameters: Namespace(model='voyageai/voyage-3', task_types=None, categories=None, tasks=['RuBQReranking'], languages=None, benchmarks=None, device=None, output_folder='/data/niklas/results/results', verbosity=2, co2_tracker=True, eval_splits=None, model_revision=None, batch_size=64, overwrite=False, save_predictions=False, func=<function run at 0x7f912e3d9630>)
INFO:mteb.evaluation.MTEB:

## Evaluating 1 tasks:
─────────────────────────────── Selected tasks  ────────────────────────────────
Reranking
    - RuBQReranking, s2p

INFO:mteb.evaluation.MTEB:

********************** Evaluating RuBQReranking **********************
No config specified, defaulting to the single config: rubq-reranking/default
INFO:datasets.builder:No config specified, defaulting to the single config: rubq-reranking/default
Loading Dataset Infos from /env/lib/conda/gritkto/lib/python3.10/site-packages/datasets/packaged_modules/json
INFO:datasets.info:Loading Dataset Infos from /env/lib/conda/gritkto/lib/python3.10/site-packages/datasets/packaged_modules/json
Overwrite dataset info from restored data version if exists.
INFO:datasets.builder:Overwrite dataset info from restored data version if exists.
Loading Dataset info from /data/huggingface/datasets/ai-forever___rubq-reranking/default/0.0.0/2e96b8f098fa4b0950fc58eacadeb31c0d0c7fa2
[...]

Issue #1649, opened 2025-01-01
error when running voyageai/voyage-3-lite on DBPedia


Muennighoff:
2025-01-01 02:32:26.267786: E external/local_xla/xla/stream_executor/cuda/cuda_fft.cc:485] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2025-01-01 02:32:26.281088: E external/local_xla/xla/stream_executor/cuda/cuda_dnn.cc:8454] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2025-01-01 02:32:26.284877: E external/local_xla/xla/stream_executor/cuda/cuda_blas.cc:1452] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
INFO:mteb.cli:Running with parameters: Namespace(model='voyageai/voyage-3-lite', task_types=None, categories=None, tasks=['DBPedia'], languages=None, benchmarks=None, device=None, output_folder='/data/niklas/results/results', verbosity=2, co2_tracker=True, eval_splits=None, model_revision=None, batch_size=64, overwrite=False, save_predictions=False, func=<function run at 0x7fba41e2d630>)
INFO:mteb.evaluation.MTEB:

## Evaluating 1 tasks:
─────────────────────────────── Selected tasks  ────────────────────────────────
Retrieval
    - DBPedia, s2p

INFO:mteb.evaluation.MTEB:

********************** Evaluating DBPedia **********************
INFO:mteb.abstasks.AbsTaskRetrieval:Loading Corpus...
Overwrite dataset info from restored data version if exists.
INFO:datasets.builder:Overwrite dataset info from restored data version if exists.
Loading Dataset info from /data/huggingface/datasets/mteb___dbpedia/corpus/0.0.0/c0f706b76e590d620bd6618b3ca8efdd34e2d659
INFO:datasets.info:Loading Dataset info from /data/huggingface/datasets/mteb___dbpedia/corpus/0.0.0/c0f706b76e590d620bd6618b3ca8efdd34e2d659
Found cached dataset dbpedia (/data/huggingface/datasets/mteb___dbpedia/corpus/0.0.0/c0f706b76e590d620bd6618b3ca8efdd34e2d659)
[...]

PR #1665, opened 2025-01-01
fix: Use prompts instead of prompt names for voyage

Samoed:
## Checklist

Closes https://github.com/embeddings-benchmark/mteb/issues/1652 
Closes https://github.com/embeddings-benchmark/mteb/issues/1649

### DIFF
--- a/mteb/models/voyage_models.py
+++ b/mteb/models/voyage_models.py
@@ -94,10 +94,7 @@ class VoyageWrapper(Wrapper):
         prompt_type: PromptType | None = None,
         **kwargs: Any,
     ) -> np.ndarray:
-        input_type = (
-            self.get_prompt_name(self.model_prompts, task_name, prompt_type)
-            or "document"
-        )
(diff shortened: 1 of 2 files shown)


mteb/models/voyage_models.py
```diff
@@ -95,8 +95,7 @@ def encode(
         **kwargs: Any,
     ) -> np.ndarray:
         input_type = (
-            self.get_prompt_name(self.model_prompts, task_name, prompt_type)
-            or "document"
+            self.get_prompt(self.model_prompts, task_name, prompt_type) or "document"
         )
```
isaac-chung, line 99: It's known from `self.model_prompts` that voyage models only accept "query" and "document" as input types. Maybe this can be further simplified to the following? That way, the `get_prompt` method is not needed.
```suggestion
input_type = self.model_prompts.get(prompt_type.value)
```

isaac-chung (approved):

Issue #1651, opened 2025-01-01
cli does not allow passing trust_remote_code=True to get_model


Muennighoff:
...which is needed for some models like cpm

https://github.com/embeddings-benchmark/mteb/blob/e4edb66ef1b4c867e7f57f888f8db93c88361739/mteb/cli.py#L123

KennethEnevoldsen:
I believe the best way to do it is like this:

```py
    loader=partial(  # type: ignore
        sentence_transformers_loader,
        name="openbmb/MiniCPM-Embedding",
        revision="c0cb2de33fb366e17c30f9d53142ff11bc18e049",
        trust_remote_code=True,
    ),
```

(for model that we actually trust)

See PR for the specific model (if you have a list I can add them all.

PR #1669, opened 2025-01-01
fix: Pass trust_remote_code=True to CPM model

KennethEnevoldsen:
Adresses #1651

## Checklist


isaac-chung:
Should we add ST to its framework too?

Issue #1659, opened 2025-01-01
openbmb/MiniCPM-Embedding fails


Muennighoff:
```
2025-01-01 04:50:33.532246: E external/local_xla/xla/stream_executor/cuda/cuda_fft.cc:485] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2025-01-01 04:50:33.546080: E external/local_xla/xla/stream_executor/cuda/cuda_dnn.cc:8454] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2025-01-01 04:50:33.549929: E external/local_xla/xla/stream_executor/cuda/cuda_blas.cc:1452] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
INFO:mteb.cli:Running with parameters: Namespace(model='openbmb/MiniCPM-Embedding', task_types=None, categories=None, tasks=['MassiveIntentClassification'], languages=None, benchmarks=None, device=None, output_folder='/data/niklas/results/results', verbosity=2, co2_tracker=True, eval_splits=None, model_revision=None, batch_size=32, overwrite=False, save_predictions=False, func=<function run at 0x7f8054790f70>)
WARNING:mteb.model_meta:Loader not specified for model openbmb/MiniCPM-Embedding, loading using sentence transformers.
You are attempting to use Flash Attention 2.0 without specifying a torch dtype. This might lead to unexpected behaviour
You are attempting to use Flash Attention 2.0 with a model not initialized on GPU. Make sure to move the model to GPU after initializing it on CPU with `model.to('cuda')`.

Loading checkpoint shards:   0%|          | 0/2 [00:00<?, ?it/s]
Loading checkpoint shards:  50%|█████     | 1/2 [00:01<00:01,  1.11s/it]
Loading checkpoint shards: 100%|██████████| 2/2 [00:01<00:00,  1.61it/s]
Loading checkpoint shards: 100%|██████████| 2/2 [00:01<00:00,  1.44it/s]
Traceback (most recent call last):
  File "/env/lib/conda/gritkto/bin/mteb", line 8, in <module>
    sys.exit(main())
  File "/data/niklas/mteb/mteb/cli.py", line 387, in main
    args.func(args)
  File "/data/niklas/mteb/mteb/cli.py", line 123, in run
    model = mteb.get_model(args.model, args.model_revision, device=device, trust_remote_code=True)
[...]

Issue #1657, opened 2025-01-01
HIT-TMG/KaLM-embedding-multilingual-mini-instruct-v1 fails


Muennighoff:
```
2025-01-01 04:07:12.442556: E external/local_xla/xla/stream_executor/cuda/cuda_fft.cc:485] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2025-01-01 04:07:12.456047: E external/local_xla/xla/stream_executor/cuda/cuda_dnn.cc:8454] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2025-01-01 04:07:12.459781: E external/local_xla/xla/stream_executor/cuda/cuda_blas.cc:1452] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
INFO:mteb.cli:Running with parameters: Namespace(model='HIT-TMG/KaLM-embedding-multilingual-mini-instruct-v1', task_types=None, categories=None, tasks=['BiorxivClusteringP2P.v2'], languages=None, benchmarks=None, device=None, output_folder='/data/niklas/results/results', verbosity=2, co2_tracker=True, eval_splits=None, model_revision=None, batch_size=32, overwrite=False, save_predictions=False, func=<function run at 0x7fbe7ae68f70>)
WARNING:mteb.model_meta:Loader not specified for model HIT-TMG/KaLM-embedding-multilingual-mini-instruct-v1, loading using sentence transformers.
Traceback (most recent call last):
  File "/env/lib/conda/gritkto/bin/mteb", line 8, in <module>
    sys.exit(main())
  File "/data/niklas/mteb/mteb/cli.py", line 387, in main
    args.func(args)
  File "/data/niklas/mteb/mteb/cli.py", line 123, in run
    model = mteb.get_model(args.model, args.model_revision, device=device, trust_remote_code=True)
  File "/data/niklas/mteb/mteb/models/overview.py", line 150, in get_model
    model = meta.load_model(**kwargs)
  File "/data/niklas/mteb/mteb/model_meta.py", line 128, in load_model
    model: Encoder = loader(**kwargs)  # type: ignore
  File "/data/niklas/mteb/mteb/model_meta.py", line 40, in sentence_transformers_loader
    return SentenceTransformerWrapper(model=model_name, revision=revision, **kwargs)
[...]

PR #1675, opened 2025-01-01
fix: add check for key error in loader

Samoed:
Incorrect task names would raise `KeyError`, but previously the check was only for `ValueError`.

Closes https://github.com/embeddings-benchmark/mteb/issues/1657
Closes https://github.com/embeddings-benchmark/mteb/issues/1659

## Checklist


Muennighoff:
Nice! Should we maybe also raise the KeyError here: https://github.com/embeddings-benchmark/mteb/blob/84f3f419e806cff575947a81a6fc1a5426b95d1b/mteb/models/wrapper.py#L83 which seems to be the problematic line like the ValueErrors there? (and maybe log them or sth such that the user is aware)

Muennighoff (approved):
Nice!

Issue #1627, opened 2024-12-24
Abnormally low values for NanoBEIR benchmark


minsik-ai:
Continuing from #1588 

NanoBEIR performance on Touche2020 and NFCorpus is too low compared to [reported values](https://www.zeta-alpha.com/post/fine-tuning-an-llm-for-state-of-the-art-retrieval-zeta-alpha-s-top-10-submission-to-the-the-mteb-be). 

You can check out some of the values here: https://github.com/embeddings-benchmark/results/pull/72

isaac-chung:
@minsik-ai could you please specify: 
1. which model you tried, the script and/or commands you used,
2. the corresponding results file in that PR you linked, and 
3. what values (metrics) you're comparing 
 
Thanks in advance!

Samoed:
The original blog only presents results for `e5-mistral`, and it's hard to evaluate because we don't know which prompts were used during testing. I think @ArthurCamara might be able to share some insights on how they evaluated models on `NanoBEIR`.

Samoed:
I've evaluated `multilingual-e5-small` on mteb NanoBEIR and sentence transformers. [Code](https://www.kaggle.com/code/samoed/nanobeir-eval/notebook?scriptVersionId=214634512). Here scores is `ndcg@10`

| Task | MTEB | Sentece Transformers |
| - | - | - |
| ArguAna | 0.44536 | 0.444486 |
| ClimateFever | 0.2222 | 0.30642 |
| DBPedia | 0.17534 | 0.6053 |
| Fever | 0.80845 | 0.30642 |
| FiQA2018 | 0.34363 | 0.4430 |
| HotpotQA | 0.56911 | 0.81012 |
| MSMARCO | 0.62091 | 0.62091 |
| NFCorpus  | 0.05535 | 0.2885 |
| NQ | 0.67664 | 0.68618 |
| Quora | 0.90621 | 0.97279 |
| SCIDOCS | 0.20826 | 0.34377 |
| SciFact | 0.71129 | 0.72457 |
| Touche2020 | 0.19598 | 0.49540 |

Not matching results:
- ArguAna
- ClimateFever
- DBPedia
- Fever
- FiQA2018
- HotpotQA
- NFCorpus
- Quora
- SCIDOCS
- Touche2020

minsik-ai:
@Samoed 's findings is the main difference I've seen! 
[...]

PR #1687, opened 2025-01-02
fix: NanoBeir

Samoed:
## Checklist

In the previous implementation, each query could only have one relevant document, but in most datasets, queries can have multiple relevant documents.

Results https://github.com/embeddings-benchmark/mteb/issues/1627#issuecomment-2568373454, https://github.com/embeddings-benchmark/results/pull/84
Closes https://github.com/embeddings-benchmark/mteb/issues/1627


isaac-chung (approved):
Good catch!

PR #1698, opened 2025-01-03
test: Add script to test model loading below n_parameters threshold

### LINKED ISSUE #1690: Unittests to check model loading?
There are lots of models that fail at the basic step of loading (e.g. https://github.com/embeddings-benchmark/mteb/issues/1689). Maybe it would be good to have a unit test that checks that all models can be loaded? One problem may the size of some large models, but maybe there is a way to work around that / we only test models that can be loaded in a reasonable amount of time.

Non-exhaustive list:
- #1652
- #1653
- #1657
- #1658
- #1659
- #1661
- #1679 
- #1680 
- #1681 
- #1689 
- #1693 
- #1696

isaac-chung:
Fixes #1690 

- [x] first get the model meta from model registry and check for n_parameters, and omit API models.
- [x] if it is below threshold (2B for now), then run `mteb.get_model` and output the exception (None or raw text)
- [x] write these to a results.json file in the repo (in the scripts folder)
- [x] ability to continue running from where the script left off (args to specify running missing models, unsuccessful models, or by model names)
- [x] Trigger script to run when there are changes to the model files, then write the results out to the json file.

## Checklist


Samoed:
I think we could create a file to record whether a model load was successful, to avoid repeated loading attempts.

isaac-chung:
Right now the main issue is disk running out of space. Looking into how not to save the weights into cache folder when loading a model.

Samoed:
Also, test should skip API based models

scripts/failures.json
```diff
@@ -0,0 +1,31 @@
...
+    "Snowflake/snowflake-arctic-embed-m-v1.5": "None",
+    "Snowflake/snowflake-arctic-embed-m-v2.0": "None",
+    "Snowflake/snowflake-arctic-embed-s": "None",
+    "Snowflake/snowflake-arctic-embed-xs": "None",
+    "ai-forever/sbert_large_mt_nlu_ru": "None",
+    "bm25s": null,
+    "colbert-ir/colbertv2.0": "To use the ColBERT models `pylate` is required. Please install it with `pip install mteb[pylate]`.",
+    "google/flan-t5-large": null,
+    "intfloat/e5-mistral-7b-instruct": "Over threshold. Not tested.",
+    "intfloat/multilingual-e5-large-instruct": "XLMRobertaModel.__init__() got an unexpected keyword argument 'cache_folder'",
```
Samoed, line 24: It seems that `cache_folder=CACHE_FOLDER` breaks loading. Maybe use default HF cache folder with `scan_cache_dir().delete_revisions`? https://huggingface.co/docs/huggingface_hub/en/guides/manage-cache#clean-cache-from-python

isaac-chung: Will give it a try.

KennethEnevoldsen (commented):
Looking good so far. Just to clarify the intention is to move this into the tests suite?

isaac-chung:
@KennethEnevoldsen thanks, and yes.

isaac-chung:
@KennethEnevoldsen @Samoed just pushed the latest changes, and updated the description. Would you mind reviewing and seeing if anything is unclear?

I'll try to push a "test" commit to see how it works.

.github/workflows/model_loading.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install mteb[pylate] gritlm mteb[model2vec] xformers
```
KennethEnevoldsen, line 25:
```suggestion
        pip install mteb[pylate, model2vec, gritlm, xformers]
```
assuming it should be specified as so (otherwise it could break unintentionally)

scripts/model_load_failures.json
KennethEnevoldsen, line 1: would move these into `tests/*` instead of scripts

.github/workflows/model_loading.yml
```diff
@@ -0,0 +1,53 @@
+name: Model Loading
+
+on:
+  pull_request:
+    paths:
+      - 'mteb/models/**.py'
```
KennethEnevoldsen, line 6: Assuming this indicate "on changes to paths"?

.github/workflows/model_loading.yml
```diff
@@ -0,0 +1,53 @@
...
+      with:
+        python-version: '3.10'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install mteb[pylate] gritlm mteb[model2vec] xformers
+
+    - name: Extract model names from changed files
```
KennethEnevoldsen, line 27: I would actually move this to a bash file in tests (potentially put it all in a folder within tests), and then add a command to the makefile something like `make test-model-loading`.

That way one can reproduce the command without sending it to github.

.github/workflows/model_loading.yml
```diff
@@ -0,0 +1,53 @@
...
+      id: extract_models
+      run: |
+        changed_files=$(git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep 'mteb/models/.*.py')
```
KennethEnevoldsen, line 30: Could be done entirely in python if you wish and then add it to the pytests using:

```
@pytest.mark.skipif(updates_to_models, reason="There have been no updated to models.")
```

I have previously made some utilities for this here:

https://huggingface.co/datasets/danish-foundation-models/danish-dynaword/blob/main/src/git_utilities.py

Which uses gitpython.

(however realize that this is quite a bit of extra work, so I am also more than happy to accept this as is)

KennethEnevoldsen (changes requested):
Generally looks good, very happy to accept this with just the changes the dependecy installations.

.github/workflows/model_loading.yml
Samoed: Also, repo doesn't have `requirements.txt` and this should be `pip install .[pylate, model2vec, gritlm, xformers]`

isaac-chung: Fill to be removed.

.github/workflows/model_loading.yml
isaac-chung: Fill to be removed.

.github/workflows/model_loading.yml
isaac-chung: When refactoring this into Python, where do you see the pip install happening? So far I see 2 main options:
- update `make install-for-tests` with all model dependencies
- call pip install for all model dependencies in a [bash script](https://github.com/embeddings-benchmark/mteb/pull/1698#discussion_r1905028123). Test duration will count the time taken for installs

isaac-chung: How do those utilities functions fit here, i.e. can they be used for finding the diff between the PR and the target branch? I tried something quick but with no success yet.

Samoed: Could we run all models, and if a model is already in the cache, skip it; otherwise, run it?

scripts/model_load_failures.json
isaac-chung: Done.

.github/workflows/model_loading.yml
isaac-chung: Related comment: https://github.com/embeddings-benchmark/mteb/pull/1698#discussion_r1905174199

.github/workflows/model_loading.yml
isaac-chung: Yes. see the `omit_previous_success` arg in `scripts/model_loading.py`. But I don't see how this is relevant to this discussion?

Samoed: I think you can convert script to test and run models. If model name not appear in model_load_failures.json then run it, otherwise skip it

Samoed: You could also compute the hash of the file and check if the hashes or the Git revision of the file matches

isaac-chung:
Thanks both for your input. Right now the test fails at the merge base step (empty list is returned). If you have some time, it would be great if you could help take a quick look there. Otherwise I plan to return to this later tomorrow.

scripts/extract_model_names.py
```diff
@@ -11,7 +11,7 @@ def get_changed_files(base_branch="main"):
     repo_path = Path(__file__).parent.parent
     repo = Repo(repo_path)
     repo.remotes.origin.fetch(base_branch)
-    repo.git.rebase(f"origin/{base_branch}")
+    repo.git.merge(f"origin/{base_branch}")
```
Samoed, line 14: It seems like you're trying to merge the current branch with the existing branch during CI, but GitHub Actions does this automatically, so it’s unnecessary. Instead, you could check the base branch's commit and find diffs by comparing commit revisions.

isaac-chung: The odd thing is that without the merge or rebase, or fetch, this works fine locally, I.e. able to find correct diff, and trigger the checks. This leads me to believe that the runner could not fetch properly somehow, or it's something else.

isaac-chung: The failed run shows "fatal: refusing to merge unrelated histories". It seems like the HEAD commit and origin/main have no common base / ancestry. Are you able to reproduce this error locally? (without the merge/rebase/fetch step)

isaac-chung: > Instead, you could check the commit of the base branch and find diffs using commit revisions. This approach would also simplify running it locally.

@Samoed your suggestion worked. Thanks!!

Makefile
```diff
@@ -35,4 +35,12 @@ pr:
...
 	# since we do not have a documentation site, this just build tables for the .md files
-	python docs/create_tasks_table.py
\ No newline at end of file
+	python docs/create_tasks_table.py
+
+
+model-load-test:
+	@echo "--- 🚀 Running model load test ---"
+	make install-for-tests
+	pip install ".[pylate,gritlm,xformers,model2vec]"
```
KennethEnevoldsen, line 44:
```suggestion
	pip install ".[dev, speedtask, pylate,gritlm,xformers,model2vec]"
```

KennethEnevoldsen: otherwise you can't make sure that dependencies are compatible.

.github/workflows/model_loading.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Install dependencies and run tests
+      run: |
+        make install-for-tests
+        pip install ".[pylate,gritlm,xformers,model2vec]"
+    - name: Extract model names
+      run: |
+        python scripts/extract_model_names.py
+    - name: Run tests
+      run: |
+        python tests/test_models/test_model_loading.py
```
KennethEnevoldsen, line 31:
```suggestion
    - name: Install dependencies and run tests
      run: |
        make model-load-test
```

to keep them compatible.

mteb/models/bge_models.py
```diff
@@ -4,6 +4,7 @@
 
 from mteb.model_meta import ModelMeta, sentence_transformers_loader
 
+# testing changes to file.
```
KennethEnevoldsen, line 7:
```suggestion
```

scripts/extract_model_names.py
```diff
@@ -0,0 +1,63 @@
...
+                                None,
+                            )
+                            if model_name:
+                                model_names.append(model_name)
+    return model_names
+
+
+if __name__ == "__main__":
+    """
+    Can pass in base brnach as an argument. Defaults to 'main'.
```
KennethEnevoldsen, line 55:
```suggestion
    Can pass in base branch as an argument. Defaults to 'main'.
```

tests/test_models/model_load_failures.json
```diff
@@ -0,0 +1,197 @@
...
+    "samaya-ai/promptriever-mistral-v0.1-7b-v1": "You are trying to access a gated repo.\nMake sure to have access to it at https://huggingface.co/mistralai/Mistral-7B-v0.1.\n401 Client Error. (Request ID: Root=1-67794457-688a6d9c24a9e8f15cf70d28;da3a233f-7c7c-4919-9cee-72a1d66acdb6)\n\nCannot access gated repo for url https://huggingface.co/mistralai/Mistral-7B-v0.1/resolve/main/config.json.\nAccess to model mistralai/Mistral-7B-v0.1 is restricted. You must have access to it and be authenticated to access it. Please log in.",
+    "sdadas/mmlw-e5-base": "None",
+    "sdadas/mmlw-e5-large": "None",
+    "sdadas/mmlw-e5-small": "None",
+    "sdadas/mmlw-roberta-base": "None",
+    "sdadas/mmlw-roberta-large": "None",
+    "sentence-transformer/multi-qa-MiniLM-L6-cos-v1": "sentence-transformer/multi-qa-MiniLM-L6-cos-v1 is not a local folder and is not a valid model identifier listed on 'https://huggingface.co/models'\nIf this is a private repository, make sure to pass a token having permission to this repo either by logging in with `huggingface-cli login` or by passing `token=<your_token>`",
```
KennethEnevoldsen, line 171: what is happening here?

tests/test_models/model_load_failures.json
```diff
@@ -0,0 +1,197 @@
...
+    "samaya-ai/promptriever-llama3.1-8b-v1": "You are trying to access a gated repo.\nMake sure to have access to it at https://huggingface.co/meta-llama/Meta-Llama-3.1-8B.\n401 Client Error. (Request ID: Root=1-677bba8f-608cf825273d8d2b0670b5ad;066bb2fa-3bef-4fb9-b3cb-4c5ffee41047)\n\nCannot access gated repo for url https://huggingface.co/meta-llama/Meta-Llama-3.1-8B/resolve/main/config.json.\nAccess to model meta-llama/Llama-3.1-8B is restricted. You must have access to it and be authenticated to access it. Please log in.",
+    "samaya-ai/promptriever-mistral-v0.1-7b-v1": "You are trying to access a gated repo.\nMake sure to have access to it at https://huggingface.co/mistralai/Mistral-7B-v0.1.\n401 Client Error. (Request ID: Root=1-67794457-688a6d9c24a9e8f15cf70d28;da3a233f-7c7c-4919-9cee-72a1d66acdb6)\n\nCannot access gated repo for url https://huggingface.co/mistralai/Mistral-7B-v0.1/resolve/main/config.json.\nAccess to model mistralai/Mistral-7B-v0.1 is restricted. You must have access to it and be authenticated to access it. Please log in.",
```
KennethEnevoldsen, line 165: We have quite a few gated models. Should we add a token for these for the test or ignore gated errors?

tests/test_models/test_model_loading.py
```diff
@@ -0,0 +1,33 @@
...
+def test_model_loading():
+    model_loading_file_path = (
+        Path(__file__).parent.parent.parent / "scripts" / "model_loading.py"
+    )
+    result = subprocess.run(
+        ["python", model_loading_file_path, "--model_name", model_names],
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, result.stderr
```
KennethEnevoldsen, line 33: Why not just import the function and call it?

KennethEnevoldsen (commented):
Great work! I think we are almost there

.github/workflows/model_loading.yml
isaac-chung: I separated them for things to be cached and for better debugging experience. Do you think that it's a fair tradeoff?

tests/test_models/model_load_failures.json
isaac-chung: I'd prefer to leave any actual work on the load failures to be separate from this PR. Let's open an issue for this.

tests/test_models/model_load_failures.json
isaac-chung: Sure. I'd prefer to leave any actual work on the load failures to be separate from this PR. Let's open an issue for this?

tests/test_models/test_model_loading.py
isaac-chung: Rrunning the script would also write the results in the json file. To keep things compatible, this way the script can just be rerun if an error is encountered?

.github/workflows/model_loading.yml
Samoed, line 1: For now, it seems that test_model_loading is being called twice—once during model loading and once during testing. I think we can keep just one call

Samoed, line 1: For now, it seems that test_model_loading is being called twice—once during model loading and once during testing. I think we can keep just one call

Samoed (commented):
Great work!

.github/workflows/model_loading.yml
isaac-chung: Ah yes, good catch. At some point, the model loading script was in the scripts folder only. We could move it from the tests folder to the scripts folder instead, and keep the CI workflow (so that we can install needed dependencies). wdyt? 

cc @KennethEnevoldsen

KennethEnevoldsen: I would normally think that scripts folder is a more lenient folder (we can edit and delete stuff and not care much about it). I would move model_loading to tests/test_models and remove the test_model_loading file. 

That would mean that we have test code in tests that is not run by pytest. I think that is fine.

KennethEnevoldsen (approved):

.github/workflows/model_loading.yml
isaac-chung: Thanks both. Just pushed a commit to address this thread. Once CI passes, we should be good to merge.

isaac-chung:
Thanks again. Merging now. Note that the model loading "test" will likely require merging the target branch to get the accurate diffs.

PR #1703, opened 2025-01-04
fix: GermanDPR

Samoed:
## Checklist

### Adding datasets checklist
- [X] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb -m {model_name} -t {task_name}` command.

Closes https://github.com/embeddings-benchmark/mteb/issues/1700


isaac-chung (approved):
Nice. Good to merge. 

In addition, I wonder if this is a specific case, or if we need to consider updating the AbsTask as well? https://github.com/embeddings-benchmark/mteb/blob/0753abaacb31989e00b8e00322254cce85903639/mteb/abstasks/AbsTaskRetrieval.py#L293

Samoed:
This is a specific case, but in this place should be modified `corpus_to_str` I think

PR #1707, opened 2025-01-04
[v2] Refactor evaluators and Abstasks

Samoed:
I've made some refactoring for the task evaluators, except for Retrieval, which still requires a significant overhaul.  

Additionally, I have some suggestions for classification:  
- Should we limit it to a single scoring method to avoid reproduction issues, since scoring methods are currently passed in the constructor?  
- Should we hardcode split names to avoid reproduction issues, as they are also passed in the constructor?  
- Should `n_experiments` and `k` become part of the task definitions without requiring changes, to avoid reproduction issues since these are currently passed?

## Checklist


mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -276,6 +261,7 @@ def clustering_downsample(
     dataset: DatasetDict, seed: int, max_samples_in_cluster: int = 2048
 ) -> DatasetDict:
     """In cases where it is not possible to convert the dataset to a fast version, we can downsample the dataset to speed up the evaluation.
+    Only used in ArXivHierarchicalClusteringP2P
```
KennethEnevoldsen, line 264: we could probably just reupload it and remove this part then

mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -97,25 +97,13 @@ class AbsTaskMultilabelClassification(AbsTask):
...
         self,
         n_experiments=None,
-        batch_size=32,
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self.batch_size = batch_size
 
         # Bootstrap parameters
         self.n_experiments = n_experiments or getattr(self, "n_experiments", 10)
```
KennethEnevoldsen, line 105: re n. experiment. You are thinking of adding like:


```
class BengaliDocumentClassification(AbsTaskClassification):
    n_experiments = 10 
    metadata = TaskMetadata(
        name="BengaliDocumentClassification",
        ...
```

I agree with this

mteb/abstasks/AbsTaskReranking.py
```diff
@@ -33,9 +33,6 @@
 class AbsTaskReranking(AbsTaskRetrieval):
     """Abstract class for re-ranking experiments. This is mostly the same as the RetrievalEvaluator, but here to adapt the old format to the new format. TODO: update these tasks to the new format and delete this class."""
 
-    def __init__(self, **kwargs):
-        super(AbsTaskRetrieval, self).__init__(**kwargs)
```
KennethEnevoldsen, line 37: isn't this needed?

mteb/abstasks/MultilingualTask.py
```diff
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
+import datasets
+
 from .AbsTask import AbsTask
-from .MultiSubsetLoader import MultiSubsetLoader
 
 
-class MultilingualTask(MultiSubsetLoader, AbsTask):
+class MultilingualTask(AbsTask):
```
KennethEnevoldsen, line 8: much cleaner

mteb/evaluation/evaluators/BitextMiningEvaluator.py
```diff
@@ -31,6 +32,7 @@ def __init__(
         self.pairs = pair_columns
         self.n = len(sentences)
         self.sentences = sentences
+        # TODO used only by BUCC
```
KennethEnevoldsen, line 35: you are probably already thinking this, but let us just re-upload it

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -34,30 +34,21 @@ def __init__(
...
-        self.encode_kwargs = encode_kwargs
-
-        if "batch_size" not in self.encode_kwargs:
-            self.encode_kwargs["batch_size"] = 32
 
         self.k = k
 
-    def __call__(self, model, test_cache=None):
+    def __call__(
+        self, model: Encoder, *, encode_kwargs: dict[str, Any] = {}, test_cache=None
```
KennethEnevoldsen, line 50: is test_cache a Path | None?

mteb/evaluation/evaluators/SummarizationEvaluator.py
```diff
@@ -26,24 +26,20 @@
...
+        gold_scores: list[list[float]],
+        task_name: str,
         **kwargs,
     ):
-        # human_summaries shape: (None, num_human_summaries)
-        # machine_summaries shape: (None, num_machine_summaries)
-        # gold scores shape: (None, num_machine_summaries)
-        # texts: (None,)
+        """Args:
+        human_summaries: shape: (-1, num_human_summaries)
```
KennethEnevoldsen, line 37: indent?

KennethEnevoldsen (commented):
Great additions!

> Should we hardcode split names to avoid reproduction issues, as they are also passed in the constructor?

Yes

> Should we limit it to a single scoring method to avoid reproduction issues, since scoring methods are currently passed in the constructor?

Hmm not sure what you refer to here

mteb/abstasks/AbsTaskMultilabelClassification.py
Samoed: Yes, I've changed like this

Samoed:
> Should we limit it to a single scoring method to avoid reproduction issues, since scoring methods are currently passed in the constructor?

For now in classification can be used KNN, PytorchKNN and LogReg (default) and I think we can leave only LogReg

mteb/abstasks/AbsTaskReranking.py
Samoed: No, if I remove it would call retrieval and this task doesn't require init

orionw: I think is it required, but only 80% sure. I think I remember fighting with this to correctly load the data. 

If you change this, would you mind checking scores for both a reranking and an instruction reranking before/after?

Samoed: Yes, I'll check scores

mteb/abstasks/MultilingualTask.py
orionw, line 8: I may have missed the recent updates, but where did the subset loader go? I remember this was very painful for MIRACL type datasets and the order of those two classes in the inheritance made a big difference

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -107,46 +98,36 @@ def __init__(
...
-            sentences_test = sentences_test[:limit]
-            y_test = y_test[:limit]
 
         self.sentences_train = sentences_train
         self.y_train = y_train
         self.sentences_test = sentences_test
         self.y_test = y_test
 
         self.task_name = task_name
-        self.encode_kwargs = encode_kwargs
```
orionw, line 127: This is a naive question (and not a comment on this PR), but what happened to encode_kwargs? I assume we moved away from it, but last I recall that was the preferred method of passing args.

mteb/abstasks/MultilingualTask.py
Samoed: I've integrated it's code to MultilingualTask

orionw: Awesome, just making sure it works. That's a great refactor!

mteb/evaluation/evaluators/ClassificationEvaluator.py
Samoed: In Evaluator interface `encode_kwargs` are passing to `__call__`.  I've aligned implementation with interface (moved from init to call)

KennethEnevoldsen:
> I think we can leave only LogReg

Agree - Set it as a class attribute, so tasks can redefine it if needed (similar to evaluator for summeval)

Then people can just redefine the class if needed.

mteb/evaluation/evaluators/ClassificationEvaluator.py
Samoed: No, this is embeddings of test split https://github.com/embeddings-benchmark/mteb/blob/8d033f39415cd00840a1e0f6305d453ca6032abf/mteb/evaluation/evaluators/ClassificationEvaluator.py#L145-L153 
For each experiment there we sample N examples per label for train and use full split for testing

mteb/abstasks/AbsTaskReranking.py
Samoed: I've tested it and it seems to work fine

Samoed:
I've combined `AbsTaskClassification` and `AbsTaskMultilabelClassification` because they have `evaluate`, `_calculate_metrics_from_split` functions

mteb/abstasks/AbsTaskClusteringFast.py
Samoed: Moved this function to `ArXivHierarchicalClusteringP2P.v2`, because `ArXivHierarchicalClusteringP2P` uses same dataset

KennethEnevoldsen:
> I’ve made AbsTaskClassification the parent class for AbsTaskMultilabelClassification because they share the evaluate and _calculate_metrics_from_split functions.

Yea I have actually considered just making them into one class (`AbsTaskClassification` is just a simplified `AbsTaskMultilabelClassification`)

KennethEnevoldsen (approved):
Looks - feel free to merge this

Samoed:
Tested some tasks on `intfloat/multilingual-e5-small`

| Task name | main | v2.0.0 | PR |
| --------------- |  ------- | --------- | ---- |
| AmazonCounterfactualClassification | 0.715075 | 0.729701 | 0.729701 |
| EmotionClassification | 0.4064 | 0.3868 | 0.3868 |
| ToxicConversationsClassification | 0.621289 |  0.643701 | 0.643701 |
| ArxivClusteringS2S | 0.33326 | 0.33326 | 0.33326 |
| RedditClustering | 0.423392 | 0.423392 | 0.423392 |
| SprintDuplicateQuestions | 0.935791 | 0.935791 | 0.935791 |
| TwitterSemEval2015 | 0.728198 | 0.728198 | 0.728198 |
| STS22 | 0.642763 | 0.642763 | 0.642763 |
| STSBenchmark | 0.835894 | 0.642763 | 0.835894 |
| STS16 | 0.834906 | 0.834906 | 0.834906 |
| SummEval | 0.293653 | 0.293653 | 0.29219 |

Summarization results are slightly different; I'll look into that. Classification results seem flaky, but I think it's due to RNG.

KennethEnevoldsen:
Didn't we change seeds in v2?

Samoed:
Yes, we changed them. Results for classification are matching between different branches of `v2`

PR #1710, opened 2025-01-05
Zero shot

x-tabdeveloping:
Added zero shot model filtering to the leaderboard along with an error message if all models have been filtered out.


mteb/model_meta.py
```diff
@@ -133,3 +134,16 @@ def model_name_as_path(self) -> str:
...
+    def is_zero_shot_on(self, tasks: list[AbsTask]) -> bool:
+        """Indicates whether the given model can be considered
+        zero-shot or not on the given tasks.
+        """
+        if not self.training_datasets:
+            return False
+        benchmark_datasets = set()
+        for task in tasks:
+            benchmark_datasets.add(task.metadata.dataset.get("path"))
+        model_datasets = {ds_name for ds_name, splits in self.training_datasets.items()}
```
orionw, line 147: Do we need to go through and add this metadata for all the models still?

I like the design though, especially that you can tell what is zero shot even on custom selected benchmarks!

x-tabdeveloping: Yes we will need to do that for this to work. We can try fetching it automatically for as many models as we can but it will have to be done

KennethEnevoldsen, line 147:
```suggestion
        model_datasets = {ds_name for ds_name, splits in self.training_datasets}
```

x-tabdeveloping:
As per a discussion with @KennethEnevoldsen we will not remove models for which the training data is unknown but rather flag them ⚠️ . I will implement this tomorrow.

orionw:
I would strongly recommend we remove them.  I think even just seeing them defeats the purpose and continues to incentivize keeping data hidden.

Is there some reason to keep them?

KennethEnevoldsen:
Hmm so more than ready to open this up for discussion, but generally I see three categories:

assuming we have three models: 

OpenAI (public reference for many models, dataset is known)
OpenModel (data is fully open)
FineTuneOnTrain (trained on the train set -> inflated performance) 


Zero-shot (hard):
- only models with annotated data are here (i.e. only OpenModel)

Zero-shot (soft) 
- same as hard but we allow for unknown private APIs but mark them using ⚠️

Non-zero shot
- all models, but we use ⚠️ to denote non-zero-shot and ❌ to denote not FineTuneOnTrain

What I am afraid of if we set "Zero-shot (hard)" as the default we will encourage people to go to non-zero-shot (max information) since they don't see what they are looking for.

However, after the discussion and writing this out. I could actually see us simply adding the zero-shot column with the symbols and I then would be happy to by have it as "zero-shot (hard)", since the symbols seems be reasonable readable (assuming we have some explainer at the buttom).

Note that we could extend "zero-shot (soft)" in the future with estimates of leakage.

This solution does not have a spot for model trained on the test set. Not sure how to handle those.

(another option is also to do: ❌: trained on train, ⚠️: test set leakage,❓: Training set is unknown)

generally the goal is to make the leaderboard attractive also to commercial parties (we have no influence if people simply go to another leaderboard)

orionw:
Perhaps worth asking people whose models are most affected (can also post here: https://github.com/embeddings-benchmark/leaderboard/issues/41).

I agree with the three categories as options. I personally think this is our chance to drive some change in the community, since the leaderboard is so used. I do see the concern that if someone is expecting to see OpenAI on the leaderboard and it's not on the leaderboard, they could be confused. But perhaps we could display the "zero-shot" prominently.

> OpenAI (public reference for many models, dataset is known)

I did not think the dataset was known, please let me know if I am wrong! 

-----------

Perhaps it would be possible for orgs to certify that their model didn't train on them, without releasing the data? I know most can't admit what they did train on, but perhaps they can acknowledge that they didn't train/validate on the MTEB set (or X benchmark).  For all the major orgs, we could trust their word.

It seems that even people who build embeddings in industry agree that we are all overfitting on MTEB and this could perhaps allow for them to gracefully move to zero-shot training in the future rather than feeling forced to be on the top of the (overfit) leaderboard since it's the default screen people land on currently.

Tagging some people who may have thoughts on the above paragraph @bwanglzu @spacemanidol @lukemerrick @vaibhavad @bclavie

orionw:
Thanks @x-tabdeveloping! Looks nice. 

I think I would still prefer zero-shot as the leaderboard default to encourage it, since that is the main leaderboard that industry folk will have to be on top of. But we can discuss that in an issue (and hopefully get broader feedback) instead of this PR, since it would be a minor change to move the default.

lukemerrick:
Giving my 2c as an author of Arctic Embed:
- I agree that it is frustrating to see a bunch of retrieval models topping the leaderboard disclosing that they trained on the training set of nearly half the datasets in the benchmark and have no way of knowing if they're actually better or if they're just overfitting to the evaluation
  - We have seen pretty compelling signs of benchmark overfitting in our work on multilingual embedding models, so it's definitely a very real issue to run truly out-of-distribution evaluation -- we saw a lot of potential overfitting to the MIRACL benchmark in our [Arctic Embed 2.0 paper](https://arxiv.org/abs/2412.04506), which contains evaluations in which competitive open-weights models perform much less competitively on the little-known CLEF benchmark which was not used in their training
  - Along this line, I really liked what the [Google Gecko paper did](https://arxiv.org/abs/2403.20327) by testing a "zero shot" variant and including those results in their paper
- The high volune of changes to the `mteb` Python package has also been endlessly frustrating, though, and I would argue somewhat unjustified
  - My team can run the same evals we set up in May on our current internal codebase just fine, because we care about this kind of thing, and it makes our lives better. Every couple months we have had to start over basically from zero when it comes to understanding and running the official `mteb` evalaution scripts, despite the datasets and metrics staying the same. This feels quite unnecessary.
- I would really appreciate a stronger commitment to backwards compatibility and stability across the mteb benchmark
  - If you want a new certified-zero-shot benchmark, one option is to *create a new benchmark* (e.g. a "zero shot" tab, just like you have for non-English MTEB).
  - If you really want to avoid overfitting, the best solution is to not publicly disclose or make available the evaluation data. This is a much more reliable strategy used by Kaggle competitions, ARC-AGI, and other competitive benchmarks

x-tabdeveloping:
@orionw I agree, I think we should take this discussion out of the PR itself :D Should I merge, or do we have any other changes in mind that should be implemented in this PR?

KennethEnevoldsen:
It is great that you say this @lukemerrick. I agree with almost all the points that you made.

After the release of v2.0.0 I believe we should be at a stable state for text-only models. If you find that this remains an issue you are more than welcome to open an issue

I will merge this now. It seems like the point of discussion is the default setting. We are more than willing to change that so do feel free to open an issue on it.

Issue #1605, opened 2024-12-16
leaderboard v2 - add autoupdate


KennethEnevoldsen:
Seems like we currently need to manually update the MTEB leaderboard (@orionw I believe you fixed the old one. I believe we really only need to restart the space daily)

orionw:
Yes, definitely! I'm trying to figure out how it get's it data, seems like there's a function in MTEB now that provides the results, is that right? (also cc'ing @x-tabdeveloping). The way I restarted it daily before was by updating the results file in the repo daily and then the results repo would re-start.

If I understand correctly, would the best way to do this be to auto-restart the leaderboard on MTEB merges? Or perhaps MTEB results github merges? I don't see the connection to the `results` github, but I haven't traced it all the way back yet.

orionw:
Totally works great :) Do you mind pointing me to that step? I can work out how to get the leaderboard to refresh after it.

orionw:
@x-tabdeveloping if you have any pointers :)

PR #1711, opened 2025-01-06
Refresh the v2 leaderboard daily

orionw:
This calls to refresh the leaderboard with a factory rebuild daily. 

If the space name changes, we will need to change this and the HF_TOKEN secret in the space (I created a new one with only permissions for the leaderboard repos just in case).


KennethEnevoldsen (approved):
Perfect - we can simply give the token access to both the new and old versions, such that when we update the link we don't have to fix the token.

orionw:
thanks @KennethEnevoldsen!

closes #1605

Issue #1724, opened 2025-01-08
Leaderboard demo down: Results are too straining to load


x-tabdeveloping:
The leaderboard is currently down. I can't access the build logs, but my guess would be that it's because the results take too much memory to load...
I ran out of RAM trying to load them on my own computer, and the build process on the HF Space also died abruptly, so I think this might be behind it.

x-tabdeveloping:
I'm wondering if this could be fixed by streaming results instead of loading them all at the same time.

Samoed:
How much it requires if you run locally?

x-tabdeveloping:
I have no idea, my PC has 16 gigs of ram and runs out of all of it when I tried to call `mteb.load_results()`

KennethEnevoldsen:
Hmm

```py
import mteb

res = mteb.load_results()
```

works fine for me. Working on 16g MacBook, running Python 3.9. Takes about 3 mins to run.

Max GPU load is around 7.5 GB, but not at all due to the script. (e.g zotero takes up a whopping 2GB) 

(the error on the HF leaderboard does look a lot like OOM though)

x-tabdeveloping:
It would be nice if we could find a more elegant solution to this, but if need be I can create a `stream_results()` utility that has the same interface as `load_results()` but treats it as a stream rather than loading them all at once.

KennethEnevoldsen:
@x-tabdeveloping and I dicusssed it. We decided that as a default we can load only the main scores in the scoresDict, but there will be a flag to enable / disable this feature.

This will drastically reduce the memory.

PR #1730, opened 2025-01-08
feat: reduce logging for load_results()

KennethEnevoldsen:
potentially related to #1724

- redacts missing subsets to avoid 100+ subsets printed to a max of 2 (see below)
- reduce to logging.info instead of logging.warning
- removed splits that are commonly never evaluated on and thus also the errors for them being missing

The second part removed quite a few warnings (4930 to 3865)

It also seems like these train splits were accidentally included in some of the MMTEB benchmark.
This will remove those splits from those benchmarks (which are all in beta, so I believe that is fine). We will have to recompute the tables for the paper though (which we should probably do anyway).

I also found what I believe to be an error: Scifact (train) was is included in MTEB(Medical). I have removed the "train" split (alspochecked other tasks in benchmark)

Here is a count of the current top errors (along with reason for not removing them):
```py
{
    "MassiveScenarioClassification: Missing splits {'validation'}": 238,  # included in e.g. mteb(fra)
    "MassiveIntentClassification: Missing splits {'validation'}": 237, # included in e.g. mteb(fra)
    "MassiveScenarioClassification: Missing subsets {'af', 'da', ...} for split test": 230,
    "AmazonReviewsClassification: Missing splits {'validation'}": 229, # included in e.g. mteb(deu)
    "MassiveIntentClassification: Missing subsets {'af', 'da', ...} for split test": 228,
    "STS22: Missing subsets {'fr-pl', 'de-en', ...} for split test": 223,
    "AmazonReviewsClassification: Missing subsets {'es', 'ja', ...} for split test": 196,
    "MTOPDomainClassification: Missing splits {'validation'}": 195, # included in mteb(fra)
    "MTOPIntentClassification: Missing splits {'validation'}": 194, # included in mteb(fra)
    "AmazonCounterfactualClassification: Missing splits {'validation'}": 189, # included in mteb(deu)
    "MTOPDomainClassification: Missing subsets {'es', 'th', ...} for split test": 165,
    "STS17: Missing subsets {'en-ar', 'es-es', ...} for split test": 164,
    "MTOPIntentClassification: Missing subsets {'es', 'th', ...} for split test": 164,
    "AmazonCounterfactualClassification: Missing subsets {'de', 'ja', ...} for split test": 148,
}
```

One could argue that we should just remove this logging as it might hide potentially more severe logging.

## Checklist

### DIFF
--- a/mteb/load_results/task_results.py
+++ b/mteb/load_results/task_results.py
@@ -511,8 +511,15 @@ class TaskResult(BaseModel):
                 new_scores[split].append(_scores)
                 seen_subsets.add(_scores["hf_subset"])
             if seen_subsets != hf_subsets:
+                missing_subsets = hf_subsets - seen_subsets
+                if len(missing_subsets) > 2:
+                    subset1, subset2 = list(missing_subsets)[:2]
+                    missing_subsets_str = f"{{'{subset1}', '{subset2}', ...}}"
+                else:
+                    missing_subsets_str = str(missing_subsets)
+
                 logger.warning(
-                    f"{task.metadata.name}: Missing subsets {hf_subsets - seen_subsets} for split {split}"
+                    f"{task.metadata.name}: Missing subsets {missing_subsets_str} for split {split}"
                 )
             seen_splits.add(split)
         if seen_splits != set(splits):
--- a/mteb/load_results/load_results.py
+++ b/mteb/load_results/load_results.py
@@ -163,7 +163,7 @@ def load_results(
                         r = r.validate_and_filter_scores(task=task)
                         filtered_results.append(r)
                     except Exception as e:
-                        logger.warning(
+                        logger.info(
                             f"Validation failed for {r.task_name} in {model_name} {revision}: {e}"
                         )
                 _results = filtered_results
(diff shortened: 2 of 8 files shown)


KennethEnevoldsen:
Seems like I accidentally pushed this to main. @Samoed and @isaac-chung can you do a post-hoc review on this?

Samoed:
I think everything is great!

mteb/load_results/task_results.py
```diff
@@ -511,8 +511,15 @@ def validate_and_filter_scores(self, task: AbsTask | None = None) -> AbsTask:
                 new_scores[split].append(_scores)
                 seen_subsets.add(_scores["hf_subset"])
             if seen_subsets != hf_subsets:
+                missing_subsets = hf_subsets - seen_subsets
+                if len(missing_subsets) > 2:
+                    subset1, subset2 = list(missing_subsets)[:2]
+                    missing_subsets_str = f"{{'{subset1}', '{subset2}', ...}}"
```
isaac-chung, line 517: [Minor] I worry that users may not notice that we had limited to the 1st 2 missing subsets, and assume that there were only 2 from the current log message. I wonder if mentioning this fact explicitly would help. (We only knew because of the PR description and had read the relevant code here)

```suggestion
                    missing_subsets_str = f"First two: {{'{subset1}', '{subset2}', ...}}"
```

isaac-chung (commented):
Great stuff! One small suggestion.

PR #1733, opened 2025-01-08
Fix RepLLaMA-based models and Instructions for Cross-Encoders

orionw:
Someone emailed me to report really low scores for Promptriever. I ended up fixing that and several other things:
- The bug they emailed me about was that the prompt name was being used as the prefix e.g. "query" instead of "query:  ". The Promptriever models also used the wrong Wrapper class and were loading from RepLLaMA. 
- Models were not allowed to define their own custom way of placing the instructions, which are used for Promptriever. I added this also. 
- While fixing the above two, I also noticed that instructions wouldn't get passed to cross-encoder models due to an update (which seems to be due to inclusion of ST cross-encoders which can't take instructions). I changed it to always pass the instructions and wrap ST models to skip them.
- I added InstructIR a while ago and forgot myself that we don't match their paper out of the box, so I added another reminder to myself and others :)

## Checklist


orionw:
Overall, while making this PR I realized we have some confusing naming (and perhaps as a community also). We have "prompts" which are dataset-level prefixes before the query (as similarly implemented by ST) but we also have "instructions" which are query-specific, and can be added in addition to the "prompt". 

I think this should be in documentation somewhere, at least. I'm not sure there's much we can do to change it at this point but thought I'd mention it.

mteb/models/promptriever_models.py
```diff
@@ -19,18 +19,19 @@ class PromptrieverWrapper(RepLLaMAWrapper, Wrapper):
...
-                for i, query in enumerate(queries)
-            ]
-        return self.encode(queries, **kwargs)
+    def encode(
+        self,
+        sentences: list[str],
+        *,
+        task_name: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,  # noqa
```
KennethEnevoldsen, line 28:
```suggestion
        **kwargs: Any,
```

KennethEnevoldsen: What is the problem here?

mteb/models/promptriever_models.py
```diff
@@ -19,18 +19,19 @@ class PromptrieverWrapper(RepLLaMAWrapper, Wrapper):
...
+    ) -> np.ndarray:
+        kwargs["is_promptriever"] = True
+        # pass task_name and prompt_type along with other kwargs
```
KennethEnevoldsen, line 31:
```suggestion
```

mteb/models/repllama_models.py
```diff
@@ -85,9 +89,17 @@ def encode(
     ) -> np.ndarray:
         batch_size = 16 if "batch_size" not in kwargs else kwargs.pop("batch_size")
         all_embeddings = []
-        prompt = self.get_prompt_name(self.model_prompts, task_name, prompt_type)
+        prompt_name = self.get_prompt_name(self.model_prompts, task_name, prompt_type)
+        prompt = self.model_prompts.get(prompt_name)
+
         if prompt:
-            sentences = [f"{prompt}{sentence}".strip() for sentence in sentences]
+            if prompt_name == "queries":
```
KennethEnevoldsen, line 96: shouldn't this check for prompt_type?

KennethEnevoldsen (commented):
Hmm my understanding of the problem is not a 100% and I think that might be a general problem (some of us are assuming encoder everywhere).

One way that we currently document the `Encoder` is using the Encoder protocol/interface (`mteb.encoder_interface.Encoder`). This documents what we expect of an encoder. 

There are a few variants of it for special use-cases (we should seek to combine these #1735), but generally the interface allows the user to quickly understand what it can or can't do.

We should create a similar thing for the CrossEncoder. I think that is the best place to start here (otherwise it would be impossible to know that the CrossEncoder can have `combine_query_and_instruction` method).

mteb/models/repllama_models.py
orionw: Good question. I am not as familiar with this part of the prompting code. Is prompt_type either query or passage? If so, I can switch it.

mteb/models/promptriever_models.py
orionw: I copied this from the RepLLaMA, at this point I am not sure. Will check and remove if not needed.

orionw:
> We should create a similar thing for the CrossEncoder. I think that is the best place to start here (otherwise it would be impossible to know that the CrossEncoder can have combine_query_and_instruction method).

Happy to also add a CrossEncoder interface. I think it would have the base class (with just predict and combine_query_and_instruction) and could optionally have the Conversational part that the Encoder has. 

Since CrossEncoders get query, passage, and instruction, there is just more complexity/combinations in how they fit it into a prompt. Whereas for Encoders, nearly all of them prefix the "prompt". The MonoT5/FollowIR-style rerankers have specific prompts they use that we have to fit these parts into.

+1 to having it in an interface and documentation so that people can figure this out :)

orionw:
@KennethEnevoldsen Added it to the interface (and added a CrossEncoder one)!

orionw:
(tests failure is for encoder class, I don't know enough about them to define an optional method)

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -125,3 +125,14 @@ def _predict(
             convert_to_numpy=True,
             **kwargs,
         )
+
+    def handle_instructions_predict(self, sentences, **kwargs):
```
Samoed, line 129: Will it work for reranking tasks?

mteb/cross_encoder_interface.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@runtime_checkable
+class CrossEncoder(Protocol):
+    """The interface for a cross-encoder in MTEB.
+
+    In general the interface is kept aligned with sentence-transformers interface. In cases where exceptions occurs these are handled within MTEB.
+    """
+
+    def __init__(self, device: str | None = None) -> None:
```
KennethEnevoldsen, line 19: should we allow for kwargs here (also for the encoder)

mteb/cross_encoder_interface.py
```diff
@@ -0,0 +1,46 @@
...
+    def predict(
+        self,
+        queries: Sequence[str],
+        passages: Sequence[str],
+        *,
+        task_name: str | None = None,
+        instruction: str | None = None,
+        **kwargs: Any,
+    ) -> np.ndarray | torch.Tensor:
+        """Predicts scores for query-passage pairs. Note that unlike the encoder, the prompt is specified in the model as it is typically more complex and model-specific.
```
KennethEnevoldsen, line 35:
```suggestion
        """Predicts relevance scores for query-passage pairs. Note that, unlike the encoder, the cross-encoder can compare across queries and passages.
        """
```

> the prompt is specified in the model as it is typically more complex and model-specific.

Hmm, I am not sure I follow. You can implement arbitrarily complex prompts using just the task name.

mteb/cross_encoder_interface.py
```diff
@@ -0,0 +1,46 @@
...
+
+        Args:
+            queries: The queries to score.
+            passages: The passages to score.
+            instruction: Optional instruction text to combine with the query.
+            **kwargs: Additional arguments to pass to the cross-encoder.
+
+        Returns:
+            The predicted scores for each query-passage pair.
```
KennethEnevoldsen, line 44:
```suggestion
            The predicted relevance scores for each query-passage pair.
```

mteb/cross_encoder_interface.py
orionw: I may be misunderstanding the prompts then. Does the task name allow for slotting arguments?

e.g. 
prompt = """{query}\n\nSome text here: {passage}\n\n{instruction} closing text here"""? This is what most rerankers use something like.

mteb/encoder_interface.py
```diff
@@ -60,6 +60,22 @@ def encode(
         """
         ...
 
+    def combine_query_and_instruction(
```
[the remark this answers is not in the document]
orionw: It is already in the model implementation class and is needed for encoders. 

Happy to move it out of the interface class though, will admit I am 100% sure what goes in an interface class. I just thought @KennethEnevoldsen was asking for it in the interface above but could be wrong.

mteb/models/sentence_transformer_wrapper.py
orionw: I thought the `predict` function only gets used for cross-encoders. In that case, yes it works for reranking tasks. Does it get used for encoder reranking tasks also?

Samoed: In your implementation sentences consist of queries, corpus and instructions, but in reranking there is only queries and corpus

orionw: Thanks for the clarification. Yes, currently in the code for all retrieval/reranking tasks the cross-encoders [instructions are a list of None](https://github.com/embeddings-benchmark/mteb/blob/4144ea54a782757f6401727ad0ab1db50b791749/mteb/evaluation/evaluators/model_classes.py#L432) so it expects the three lists.

mteb/cross_encoder_interface.py
orionw: @Samoed if you know as well?

Samoed: I'm not sure what you mean, but you can specify instructions for tasks. However, I don't think this applies to templates.

KennethEnevoldsen: The model will get the task_name and from there they can essentially implement their own mapping from task to prompt/template.

KennethEnevoldsen: > Got it, so you're saying define a model_prompts which has the main prompt and then get it back via the task_name?

yes

> Is it okay if we leave it as is right now, and push this to a separate PR? Perhaps there are other use cases but I can't think of them currently.

Hmm would much rather add the complexity when needed. Removing instructions here would allow us to have the CrossEncoder and Encoder's encode interface be exactly the same (which would greatly simplify things)

orionw: Okay, will remove the crossencoder stuff from this PR then and we can handle it in another.  That’s a much bigger change and I have others waiting on the Promptriever fix.

KennethEnevoldsen: Hmm, sorry you are right, leave it as is and get this fix merged in, we can then update it in an upcoming (easier that splitting it up)

orionw:
Thanks @KennethEnevoldsen! I made an issue in https://github.com/embeddings-benchmark/mteb/issues/1776 and will work on making a separate PR to address it. 

I think this is done then (although it has a new test which is failing, but looks like the test is in development? Perhaps you know more).

Samoed:
If you're about model load it now working with v2 branch

orionw:
Yes that one, thanks @Samoed.  I reran that test, seemed to fail by taking too much disk space when trying to load results.

KennethEnevoldsen:
Yea I think this is good to merge (model loading can be ignored for now)

orionw:
Thanks @KennethEnevoldsen and @Samoed for the help!

PR #1739, opened 2025-01-09
Fixed result loading on leaderboard

x-tabdeveloping:
When using the leaderboard, only main scores get loaded thereby preventing OOM errors.


orionw (approved):
LGTM and makes sense for the leaderboard!

Issue #1720, opened 2025-01-07
Leaderboard: Add training data annotations for models


x-tabdeveloping:
In my PR #1710 I have implemented zero-shot filtering for the leaderboard, as well as added training data annotations for 24 models that had it in their metadata on HuggingFace.
We are missing annotations for 150 models.

Here's an exhaustive list:
```python
models_missing_training_data_annotations = [
    "Snowflake/snowflake-arctic-embed-xs",
    "Snowflake/snowflake-arctic-embed-s",
    "Snowflake/snowflake-arctic-embed-m",
    "Snowflake/snowflake-arctic-embed-m-long",
    "Snowflake/snowflake-arctic-embed-l",
    "Snowflake/snowflake-arctic-embed-m-v1.5",
    "Snowflake/snowflake-arctic-embed-m-v2.0",
    "Snowflake/snowflake-arctic-embed-l-v2.0",
    "BAAI/bge-small-en-v1.5",
    "BAAI/bge-base-en-v1.5",
    "BAAI/bge-large-en-v1.5",
    "bm25s",
    "Cohere/Cohere-embed-multilingual-v3.0",
    "Cohere/Cohere-embed-english-v3.0",
    "Cohere/Cohere-embed-multilingual-light-v3.0",
    "Cohere/Cohere-embed-english-light-v3.0",
    "colbert-ir/colbertv2.0",
    "jinaai/jina-colbert-v2",
    "intfloat/multilingual-e5-large-instruct",
    "intfloat/e5-mistral-7b-instruct",
    "intfloat/multilingual-e5-small",
    "intfloat/multilingual-e5-base",
    "intfloat/multilingual-e5-large",
    "intfloat/e5-small-v2",
    "intfloat/e5-small",
    "intfloat/e5-base-v2",
    "intfloat/e5-large-v2",
    "intfloat/e5-large",
    "intfloat/e5-base",
    "google/text-embedding-004",
    "google/text-embedding-005",
    "google/text-multilingual-embedding-002",
    "Alibaba-NLP/gte-Qwen2-7B-instruct",
    "Alibaba-NLP/gte-Qwen1.5-7B-instruct",
    "Alibaba-NLP/gte-Qwen2-1.5B-instruct",
    "ibm-granite/granite-embedding-107m-multilingual",
    "ibm-granite/granite-embedding-278m-multilingual",
    "ibm-granite/granite-embedding-30m-english",
    "ibm-granite/granite-embedding-125m-english",
[...]

PR #1742, opened 2025-01-09
fix: added annotations for training data

KennethEnevoldsen:
Added training data annotations for a variety of models. 

This was quite hard to do I must say so something might be wrong. It would be great to get a review especially on the sentence embeddings training data that will solve a lot of downstream cases.

We should probably tag relevant model authors here as well.

@Muennighoff unsure if the NQ test split on mteb corresponds to train/dev split on natural questions? Can you also take a look at the stackexchange cases.

If the split wasn't annotated I assumed that they trained on the full dataset (including test).

Adressed #1720

## Checklist


mteb/models/sentence_transformers_models.py
```diff
@@ -78,9 +78,17 @@
     superseded_by=None,
     adapted_from=None,
     training_datasets={
+        # trained on stack exchange, unsure if sources match
+        "StackExchangeClusteringP2P": ["test"],
```
Samoed, line 82: I can't find mention of StackExchange in https://arxiv.org/abs/1908.10084. Is it part of SNLI?

mteb/models/openai_models.py
```diff
@@ -176,4 +182,7 @@ def _to_numpy(self, embedding_response) -> np.ndarray:
     use_instructions=False,
     n_parameters=None,
     memory_usage=None,
+    public_training_data=False,  # assumed
+    public_training_code=False,  # assumed
+    training_datasets=None,
```
Samoed, line 187: From https://arxiv.org/abs/2201.10005 ([post from their blog](https://openai.com/index/introducing-text-and-code-embeddings/))
They trained on SNLI, MultiNLI, MS-MARCO, but this for the first models generation (text-similarity-{ada, babbage, curie, davinci}-001)

Samoed (commented):
Awesome!

Muennighoff:
Re: NQ - I think it is probably a subset of the dev split (filtered out: queries without an answer, or having a table as an answer, or with conflicting Wikipedia pages)

mteb/models/sentence_transformers_models.py
KennethEnevoldsen: Good point - @x-tabdeveloping where you find these?

KennethEnevoldsen: it is based on the datasheet (see yaml header): https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2

mteb/models/openai_models.py
KennethEnevoldsen: Will leave it unknown then.

mteb/models/sentence_transformers_models.py
Samoed: I thought all datasets were displayed, but they don’t appear in the interface.

KennethEnevoldsen:
Updated a few cases. I am unsure if "StackExchangeClusteringP2P" is included in the training data if you train on:
"flax-sentence-embeddings/stackexchange_xm"l

mteb/models/sentence_transformers_models.py
x-tabdeveloping: Yes, they are in the model card

KennethEnevoldsen:
Will merge this in for now and keep adding annotations

PR #1745, opened 2025-01-10
fix: Leaderboard Speedup

x-tabdeveloping:
Due to the huge increase in data, I needed to optimize the leaderboard again.

#### `get_scores_fast`
Since one of the main issues was the amount of time it would take to call `get_scores()` I wrote a faster, more limited version plus I made sure scores only get calculated when absolutely necessary. 
I made the function itself faster by avoiding memory allocations wherever possible. On a larger scale this corresponds to a roughly 2.5x speed increase to `get_scores`

#### Dependency graph and events
I also added smarter managing of user-induced and callback-induced state changes, thereby drastically reducing the number of unnecessary callbacks.

The difference is night and day, the leaderboard went from being barely usable to quite fast :zap:

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+import itertools
 import json
+import logging
 import tempfile
+import time
 from collections import defaultdict
 from pathlib import Path
 from urllib.parse import urlencode
@@ -15,6 +18,8 @@ from mteb.caching import json_cache
 from mteb.leaderboard.figures import performance_size_plot, radar_chart
 from mteb.leaderboard.table import scores_to_tables
 from mteb.models.overview import get_model_meta
+
+logger = logging.getLogger(__name__)
 
 
 def load_results():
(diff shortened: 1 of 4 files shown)


mteb/leaderboard/app.py
```diff
@@ -112,15 +117,20 @@ def update_task_info(task_names: str) -> gr.DataFrame:
     return gr.DataFrame(df, datatype=["markdown"] + ["str"] * (len(df.columns) - 1))
 
 
+print("Loading all benchmark results")
```
KennethEnevoldsen, line 120:
```suggestion
```
or change to logging

KennethEnevoldsen: (general comment)

mteb/load_results/benchmark_results.py
```diff
@@ -87,21 +87,35 @@ def get_scores(
...
+        getter: Callable[[ScoresDict], Score] | None = None,
+        aggregation: Callable[[list[Score]], Any] | None = None,
         format: Literal["wide", "long"] = "wide",
     ) -> dict | list:
+        if (getter is not None) or (aggregation is not None) or (scripts is not None):
+            use_fast = False
+            getter = (
+                getter if getter is not None else lambda scores: scores["main_score"]
+            )
+            aggregation = aggregation if aggregation is not None else np.mean
```
KennethEnevoldsen, line 99: Move this to TaskResult.get_score?

(no need to have the complexity here)

x-tabdeveloping: I'd argue that this is better. The reason being that if we check this here, we won't have to check over and over again in `get_score` which would then get called a lot.

mteb/leaderboard/app.py
x-tabdeveloping: sure thing

PR #1748, opened 2025-01-10
fix: fixes implementation of similarity()

sam-hey:
## Checklist

Fixes ColBERT similarity()
https://github.com/embeddings-benchmark/mteb/pull/1594

CC @Samoed

### DIFF
--- a/mteb/evaluation/evaluators/RetrievalEvaluator.py
+++ b/mteb/evaluation/evaluators/RetrievalEvaluator.py
@@ -167,12 +167,12 @@ class DenseRetrievalExactSearch:
                     self.corpus_embeddings[request_qid].append(sub_corpus_embeddings)
 
             # Compute similarites using self defined similarity otherwise default to cosine-similarity
-            similarity_scores = cos_sim(query_embeddings, sub_corpus_embeddings)
             if hasattr(self.model, "similarity"):
                 similarity_scores = self.model.similarity(
-                    float(self.model.similarity(e1, e2))
-                    for e1, e2 in zip(query_embeddings, sub_corpus_embeddings)
+                    query_embeddings, sub_corpus_embeddings
                 )
+            else:
+                similarity_scores = cos_sim(query_embeddings, sub_corpus_embeddings)
             is_nan = torch.isnan(similarity_scores)
             if is_nan.sum() > 0:
                 logger.warning(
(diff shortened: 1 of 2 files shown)


mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -376,6 +376,9 @@ def __init__(self, model, **kwargs):
         self.save_corpus_embeddings = kwargs.get("save_corpus_embeddings", False)
         self.corpus_embeddings = {}
 
+        if hasattr(self.model, "similarity") and callable(self.model.similarity):
```
Samoed, line 379: Why do you need this?

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -59,9 +59,6 @@ def __init__(
         if isinstance(self.model, CrossEncoder):
             self.predict = self._predict
 
-        if hasattr(self.model, "similarity"):
```
Samoed, line 62: This shouldn't be removed because you'll need to use `model.model.similarity`.

Samoed (changes requested):
Good catch!

mteb/models/sentence_transformer_wrapper.py
Samoed: This is inside model class. Yes, you're right

mteb/evaluation/evaluators/RetrievalEvaluator.py
Samoed: This is inside model class. Yes, you're right

mteb/models/sentence_transformer_wrapper.py
sam-hey: ColBERT does not implement SentenceTransformerWrapper, so it is not required at this point. The object mentioned above is of type DRESModel, which can handle this functionality.

Samoed: Yes, you're right about colber, but other models would benefit from this function

Samoed: To `DRESModel` would be passed `SentenceTransformerWrapper`, which doesn't have similarity function. To call it would be required to call `model.model.similarity`

sam-hey: Yes, this is correct, but I changed the code in `DRESModel` to handle this:

[https://github.com/sam-hey/mteb/blob/main/mteb/evaluation/evaluators/RetrievalEvaluator.py#L379](https://github.com/sam-hey/mteb/blob/main/mteb/evaluation/evaluators/RetrievalEvaluator.py#L379)

Samoed: But in that case, `SentenceTransformerWrapper` (model that passed to the evaluator) wouldn't have `model.similarity`

sam-hey: You are right, I hadn't considered this case. I've added a commit to address it

Samoed (approved):

Issue #1731, opened 2025-01-08
Non Cosine-Sim Similarity functions for Sentence Transformer models are broken in v2.0.0


orionw:
I tested using  model `facebook/contriever-msmarco` (which should use `dot`) and dataset `InstructIR` (but TBH any dataset works).

If you stick a breakpoint in the sim function after [this line](https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/evaluation/evaluators/model_classes.py#L324) you can check, it is always cos_sim. This might be because the SentenceTransformer models are only defined via ModelMeta and the default sim function is not stored in ST (AFAIK).

Samoed:
Default similarity function is stored in `sentence_transformer_config.json`. Example https://huggingface.co/infgrad/jasper_en_vision_language_v1/blob/c6991f5a54805d15ea1af42f9db2eee4cca96edc/config_sentence_transformers.json#L12

orionw:
Ah, awesome @Samoed! But this is probably a newer update for ST then.  I don't see that in [contriever-msmarco's files]. So then I guess this issue is for older models where that isn't defined and is only defined in the model_meta.

Samoed:
We need to add the model meta for this model with a wrapper that uses the dot function. From what I found when adding a similar function, all models used cosine similarity, but I might be wrong.

orionw:
The model meta says [dot](https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/models/sentence_transformers_models.py#L305) - I don't think I know the wrapper you are talking about though. 

It does seem to be the only dot model in the file though!

Samoed:
I'll add wrapper for this

sam-hey:
@orionw `ModelMeta.similarity_fn_name` is currently ignored and does not affect which similarity function is used. However, [#1749](https://github.com/embeddings-benchmark/mteb/pull/1749) has temporarily addressed this issue.

It seems that [#1759](https://github.com/embeddings-benchmark/mteb/pull/1759) might provide a more permanent fix for this behavior.

Samoed:
I think this can be closed for now, because issue was fixed for now

PR #1749, opened 2025-01-10
[v2] fix contriever (add similarity_fn_name to ST wrapper)

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/1731

## Checklist

### DIFF
--- a/mteb/models/sentence_transformer_wrapper.py
+++ b/mteb/models/sentence_transformer_wrapper.py
@@ -11,6 +11,7 @@ from sentence_transformers import CrossEncoder, SentenceTransformer
 from mteb.encoder_interface import PromptType
 
 from .wrapper import Wrapper
+from ..evaluation import dot_distance
 
 logger = logging.getLogger(__name__)
 
@@ -21,6 +22,7 @@ class SentenceTransformerWrapper(Wrapper):
         model: str | SentenceTransformer | CrossEncoder,
         revision: str | None = None,
         model_prompts: dict[str, str] | None = None,
+        use_model_similarity: bool = True,
         **kwargs,
     ) -> None:
         """Wrapper for SentenceTransformer models.
@@ -32,6 +34,7 @@ class SentenceTransformerWrapper(Wrapper):
                 First priority is given to the composed prompt of task name + prompt type (query or passage), then to the specific task prompt,
                 then to the composed prompt of task type + prompt type, then to the specific task type prompt,
                 and finally to the specific prompt type.
+            use_model_similarity: Whether to use the model's similarity method.
             **kwargs: Additional arguments to pass to the SentenceTransformer model.
         """
         if isinstance(model, str):
@@ -59,7 +62,7 @@ class SentenceTransformerWrapper(Wrapper):
         if isinstance(self.model, CrossEncoder):
             self.predict = self._predict
 
-        if hasattr(self.model, "similarity"):
+        if hasattr(self.model, "similarity") and use_model_similarity:
             self.similarity = self.model.similarity
 
     def encode(
@@ -125,3 +128,8 @@ class SentenceTransformerWrapper(Wrapper):
             convert_to_numpy=True,
             **kwargs,
         )
+
+
+class SentenceTransformerWrapperDotSimilarity(SentenceTransformerWrapper):
+    def similarity(self, embedding1: np.ndarray, embedding2: np.ndarray) -> float:
+        return dot_distance(embedding1, embedding2)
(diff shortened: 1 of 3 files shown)


Samoed:
@isaac-chung Model loading gives 
```
System.IO.IOException: No space left on device : '/home/runner/runners/2.321.0/_diag/Worker_20250110-152956-utc.log' ...
```

sam-hey:
Just as an idea: I believe the `ModelMeta` object already has this information `similarity_fn_name="cosine"`. Why not add an Enum type to specify the evaluation function? This would make it much more straightforward to add new models. There's no need to implement a new wrapper class—just one Enum pointing to the fn

```
class EvaluationFunction(Enum):
    DOT_PRODUCT = dot_product
    MAX_SIM = max_similarity
    COSINE = cosine_similarity
```

sam-hey:
image

Please take a look at the screenshot. The MTEB `ModelMeta` is already available.

https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/evaluation/evaluators/model_classes.py#L324:L326

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -125,3 +126,8 @@ def _predict(
             convert_to_numpy=True,
             **kwargs,
         )
+
+
+class SentenceTransformerWrapperDotSimilarity(SentenceTransformerWrapper):
+    def similarity(self, embedding1: np.ndarray, embedding2: np.ndarray) -> float:
+        return dot_distance(embedding1, embedding2)
```
KennethEnevoldsen, line 133: Isn't it possible to have the model self define their similarity function? Any reason for us to create a custom wrapper?

KennethEnevoldsen: (if this is a one-of I would just move it next to the special case)

KennethEnevoldsen (approved):

mteb/models/sentence_transformer_wrapper.py
Samoed: Model that trained with sentence transformers have ability to provide it, but this model doesn't have config for this as was discussed in https://github.com/embeddings-benchmark/mteb/issues/1731. I think there can be more models with different similarity functions. Maybe we can use approach from https://github.com/embeddings-benchmark/mteb/pull/1759

isaac-chung: Have we tried defining a Model Meta for the contriever model with the dot similarity function? I thought that was the same way we implemented Colbert with @sam-hey but with max-sim.

Samoed: For colbert models created different wrapper with max-sim

sam-hey: > it's a bit surprising that ModelMeta.similarity_fn_name isn't being utilized.

We would love to switch to that one and would encourage a PR for this. It was only recently added to haven't been integrated throughout.

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/pull/1592#discussion_r1885422883_

---
I believe `ModelMeta` was never fully utilized. Instead of reverting to the previous version—which I agree was a better approach—we might be able to resolve these issues by properly implementing `ModelMeta` now.

[@isaac-chung's comment](https://github.com/embeddings-benchmark/mteb/pull/1759#discussion_r1912103624) raises a valid point: 

> I'd like to understand why, and perhaps convince us to move/revert back to what we had before, as to me, it worked better than needing to define separate wrappers per scoring function, which isn't scalable.

Samoed: I've updated the approach so that the similarity function can now be directly passed to the wrapper.  

Previously, users had to specify which similarity function to use, which was a hidden feature and independent of the models.

```python
evaluation.run(
    model,
    score_function="max_sim",
)
```

Samoed: I’ve updated it again, and now it uses a string `similarity_fn_name` like it stated in `ModelMeta`. I suggest leaving it as is for now and integrate `ModelMeta` parameters in a separate PR, as we’re currently duplicating model names and revisions for every model.

isaac-chung: Thanks all. I felt that we could have improved docs for the previous 'hidden' feature, rather than removing it. 

I agree that we can improve the ModelMeta param integration in a separate PR.

Samoed: The previous implementation required users to explicitly pass the scoring function, which could lead to issues with results that were hard to reproduce or match. This new approach reduces user input.

isaac-chung (approved):
Thanks for iterating! Feel free to skip the model loading test while I work on a fix. 

Perhaps we could repurpose https://github.com/embeddings-benchmark/mteb/pull/1759 a bit now for the model meta integration.

PR #1759, opened 2025-01-10
[v2] add similarity_fn in ModelMeta

sam-hey:
## Checklist

- **fix:** Retrieve `name` and `revision` from SentenceTransformers metadata.
- **mv:** Move the `get_similarity_function` method to `ModelMeta`.
- **fix:** Resolve issues with `bm25s`.
- **feat:** Add a `max_similarity` function to `utils.py`.
- **feat:** Map and load SentenceTransformers similarity functions to MTEB similarity functions.
- **ref**: Changed `max_sim` to `MaxSim` to align with the `pylate` implementation of `similarity_fn_name` https://github.com/lightonai/pylate/issues/77
- **test:** add test for ColBERT and bm25s


mteb/evaluation/evaluators/model_classes.py
```diff
@@ -321,8 +321,11 @@ def _full_corpus_search(
             query_embeddings = torch.as_tensor(query_embeddings).to(device)
             sub_corpus_embeddings = torch.as_tensor(sub_corpus_embeddings).to(device)
 
+            # TODO: If the model is not DRESModel
```
Samoed, line 324: In this branch there won't be DRESModel, it will be removed in https://github.com/embeddings-benchmark/mteb/pull/1750

mteb/model_meta.py
```diff
@@ -101,13 +108,27 @@ class ModelMeta(BaseModel):
     public_training_code: bool | None = None
     framework: list[FRAMEWORKS] = []
     reference: STR_URL | None = None
-    similarity_fn_name: DISTANCE_METRICS | None = None
+    similarity_fn_name: DISTANCE_METRICS | EvaluationFunction | None = None
```
Samoed, line 111: I think it's better to be left similarity_fn_name as is and check only in `get_evaluation_function`

mteb/evaluation/evaluators/model_classes.py
```diff
@@ -321,8 +321,11 @@ def _full_corpus_search(
...
             score_function = (
-                self.model.similarity if hasattr(self.model, "similarity") else cos_sim
+                self.model.similarity
+                if hasattr(self.model, "similarity")
+                else self.model.model.mteb_model_meta.get_evaluation_function()
```
Samoed, line 328: I think we should directly modify the similarity function in the model itself and keep `ModelMeta` only for descriptive information about the model. It might be possible to pass this parameter in `MTEB.py`, but I’m not sure how to do that effectively.

Samoed (commented):
Overall, I really appreciate your suggestions, but I’m not sure if we should rely on the `ModelMeta` object for anything during evaluation.

mteb/evaluation/evaluators/model_classes.py
sam-hey: I see your point, but I personally find it unintuitive to update the information in multiple places. In my opinion, there should be a single source of truth. Since ModelMeta already has the appropriate access and information available, it seems like a suitable candidate to serve as that single source :)

Samoed: I totally agree

Samoed: Overall I think we should modify `similarity` function somehow

Samoed: What do you mean by `global similarity functions`?

mteb/model_meta.py
```diff
@@ -51,6 +53,11 @@ def get_loader_name(
     return loader.__name__
 
 
+class EvaluationFunction(str, Enum):
```
isaac-chung, line 56: I believe that's how we've been calling these in AbsTasks?

```suggestion
class ScoringFunction(str, Enum):
```

mteb/evaluation/evaluators/model_classes.py
isaac-chung: @sam-hey I agree with your point here about defining scoring functions separately, and wrappers can specify them in places like ModelMeta. I thought we were already doing that. If we've deviated from that, I'd like to understand why, and perhaps convince us to move/revert back to what we had before as to me, it worked better than needing to define separate wrappers per scoring function, which isn't scalable.

isaac-chung (commented):
Added 2 comments as I see that this is meant to be partially complete at this time.

mteb/evaluation/evaluators/model_classes.py
sam-hey: > What do you mean by `global similarity functions`?

Consolidate all similarity functions into a single **new** file or retain them in `/mteb/evaluation/evaluators/utils.py` for better organization.

sam-hey:
I made some additional changes, so the PR ended up being a bit larger than originally planned. Please take a look at what I’ve done and share your feedback—I’m looking forward to hearing your thoughts! @Samoed @isaac-chung 

Just a heads-up, there’s one failing test: https://github.com/embeddings-benchmark/mteb/issues/1777.

Thanks!

Samoed (commented):
This currently doesn’t fix the issues with the similarity function, as it focuses only on the retrieval model wrapper. I think the changes should be made in `get_model`. The model name and revision can be used there, and the type of similarity function can be passed to the loader.

https://github.com/embeddings-benchmark/mteb/blob/c3b46b7b25de642b40a0020c192a7c4e24530bb2/mteb/models/overview.py#L156-L157

sam-hey:
> This currently doesn’t fix the issues with the similarity function, as it focuses only on the retrieval model wrapper. I think the changes should be made in `get_model`. The model name and revision can be used there, and the type of similarity function can be passed to the loader.
> 
> https://github.com/embeddings-benchmark/mteb/blob/c3b46b7b25de642b40a0020c192a7c4e24530bb2/mteb/models/overview.py#L156-L157
> 
> Also `ModelMeta` can be changed to separate `loader` and additional `loader_kwargs`

Yes, you're correct about the changes. Did you notice these changes? I’m using the st similarity function names  and map them to the Enum.
https://github.com/embeddings-benchmark/mteb/pull/1759/files#diff-629817399c17e22b713b95ac5146dc9ddd81b3977bef3165c23fa6166534224b

Samoed:
No, I missed, thanks! But still I think this should be changed to not duplicate `model_name` and `revision`

sam-hey:
What do you mean by duplicate? The code already existed, more or less. I just needed to fix the implementation because it was non-functional and assign the `similarity_fn_name` from st to the `ModelMeta`.

sam-hey:
By the way, this implementation also resolves the issue where MTEB fails to extract metadata from SentenceTransformers :)

Samoed:
I think that we should change not only how `similarity` working, but totally how integrated `ModelMeta` with `get_model`. For now, all models duplicating `model_name` and `revision`. E.g. https://github.com/embeddings-benchmark/mteb/blob/c3b46b7b25de642b40a0020c192a7c4e24530bb2/mteb/models/e5_models.py#L372-L381
And I think this can be changed in this PR

sam-hey:
I completely agree with you that this is something worth addressing! However, I’d prefer to keep the scope of this PR as it is for now. At the moment, I don’t have a clear approach in mind for tackling this, and I feel the PR is already quite substantial and feature-complete.

That said, I fully agree that the duplicates should be removed. Perhaps we could get this PR merged if everyone is fine with it as it stands. If you have a good idea on how to remove the duplicates, feel free to take it over, @Samoed.

KennethEnevoldsen:
I am note quite sure of the context of this PR. It would great with a short explanation of what the PR is trying to solve. Below are some questions to help getting that process started.

> fix: Retrieve name and revision from SentenceTransformers metadata.

Can't see where this happens? Why do we want this?

> feat: Introduce an Enum to represent all implementations of similarity functions.
> feat: Refactor all models to utilize the ScoringFunction Enum.
> rm: Eliminate literals used to describe scoring functions.

We generally havenøt used enum's anywhere else in the repo. Any strong reason to introduce it over the literal?

> mv: Move the get_similarity_function method to ModelMeta.
> feat: Map and load SentenceTransformers similarity functions to MTEB similarity functions.

Hmm, but why do we need a get_similarity function in the first place? The interface specified that the model should specify the similarity function. Is there anything that I am missing here?

> fix: Resolve issues with bm25s.

What issue?

> but totally how integrated `ModelMeta` with `get_model`.  For now, all models duplicating model_name and revision. E.g.

I agree. I don't like this as well. However that is quite a large overhaul. We could consider reformatting models similar to tasks (so that we have a model where we can call `model.metadata`). However, the reason why we have the current system is that we don't want to load the model when only dealing with the metadata. This is solved in tasks using the `load_data` method. We could redo models in the same way, but currently I don't see a strong reason to do so. Thogh I would love to be convinced otherwise (however this is not for this PR)

Samoed:
Ok, I'll create separate PR for updating work with `ModelMeta`

sam-hey:
> **fix:** Retrieve `name` and `revision` from SentenceTransformers metadata.  
> **Can't see where this happens? Why do we want this?**  

Currently, when passing a model, the old way MTEB creates two folders with missing model name and revision. This happens because a type check in `mteb/models/overview.py` is on `model`, not `model.model`, which is of type `SentenceTransformers`.

> **We generally haven't used enums anywhere else in the repo. Any strong reason to introduce it over the literal?**  

This might be a subjective preference, but using an Enum reduces the code size and, in my opinion, makes it easier to work with when implementing a new model, as it provides type suggestions from the IDE.

I also believe it makes development in other parts of the project much easier, as the available types are clearly defined and suggested by the IDE.  
Personally, I'd also love to see all the different language codes as an `Enum`.

> **Hmm, but why do we need a `get_similarity_function` in the first place? 
> The interface specifies that the model should specify the similarity function. Is there anything that I’m missing here?**  

The `get_similarity_function` maps the `Enum` in `ModelMeta` to the actual function that gets called. It’s not practical to define a function directly in `ModelMeta`, as it would require searching for every function.

> **What issue?**  

The BM25 function call does not match the defined `search` function.

KennethEnevoldsen:
> Ok, I'll create separate PR for updating work with ModelMeta

Maybe good to outline it in an issue first before you spend a lot of time on the refactor

KennethEnevoldsen:
> Currently, when passing a model, the old way MTEB creates two folders with missing model name and revision. This happens because a type check in mteb/models/overview.py is on model, not model.model, which is of type SentenceTransformers.

Not sure what you mean here. What is model in this case? When I run:

```py
import mteb
task = mteb.get_task("BornholmBitextMining") 
model = mteb.get_model("sentence-transformers/all-MiniLM-L6-v2")
bench = mteb.MTEB(tasks=[task])
bench.run(model) # only created one folder
```

I do not get two folders.

> re Enum

I understand where you are coming from. I think this is a more general change across the repo and not one that should happen just here. I would suggest that we stick with Literals as this will be a breaking change so no reason to add it unless we have a strong reason to. However do open an issue on Literals vs. Enums (most IDEs, e.g. vscode also allow suggestions when using literals).

> The get_similarity_function maps the Enum in ModelMeta to the actual function that gets called. It’s not practical to define a function directly in ModelMeta, as it would require searching for every function.

But why not use the `model.similarity` interface:

https://github.com/embeddings-benchmark/mteb/blob/0c5c3a544bea7dcb4c6e6d75d612638171cf0332/mteb/encoder_interface.py#L71

This is modelled after sentence-transformers and freely allow the model to implement their own similarity function. Why is there a reason to fetch it?

sam-hey:
> Not sure what you mean here. What is model in this case? When I run:

When I run the following code:
```

from mteb import MTEB

de_name = "average_word_embeddings_komninos"
revision = "21eec43590414cb8e3a6f654857abed0483ae36e"
de = SentenceTransformer(de_name, revision=revision)
ce = CrossEncoder("cross-encoder/ms-marco-TinyBERT-L-2-v2")
ce_revision = "e9ea2688951463fc2791a2ea2ddfce6762900675"

eval = MTEB(tasks=["SciFact"])
eval.run(
    de,
    output_folder="tests/results/stage1",
    overwrite_results=True,
    save_predictions=True,
    eval_splits=["test"],
)
eval.run(
    ce,
    eval_splits=["test"],
    output_folder="tests/results/stage2",
    save_predictions=True,
    top_k=10,
    previous_results="tests/results/stage1",
    )
```

### Enums
VSCode does not provide autocompletion for Literal types by default when using the Python extension.
Example (no autocompletion for DISTANCE_METRICS): No Autocomplete No Autocomplete Example

Accessing Available Types:
To find the available types, you need to manually explore ModelMeta and check the DISTANCE_METRICS attribute.
This process is inconvenient and hampers productivity. Transitioning to Enums would:

- Drastically simplify type inspection
- Improve the ease of adding new models


> But why not use the model.similarity interface:

The current implementation uses the `similarity()` method.
By aligning the approach with `SentenceTransformers` and incorporating ModelMeta, the implementation becomes more streamlined and consistent.

KennethEnevoldsen:
Thanks for the clarification. Besides the Enums (which is being discussed in an #1784) I believe this is ready to merge as soon as tests pass.

sam-hey:
This PR originally had a completely different scope, so you're absolutely correct—it ended up getting quite tangled with multiple different PRs and issues addressing the same topic. My apologies for the confusion.

Only https://github.com/embeddings-benchmark/mteb/issues/1777 is failing, but that's expected

isaac-chung:
> Only https://github.com/embeddings-benchmark/mteb/issues/1777 is failing, but that's expected

That's fixed in the latest commit in main. Feel free to copy that small commit over.

sam-hey:
@KennethEnevoldsen, what do you think about merging it in its current state? I’ve kept the Enum but reverted the changes in all the models.
This should minimize the effort required to merge the new changes while still supporting enums.

Personally, I think this is a good compromise. Backward compatibility is fully maintained.

mteb/models/overview.py
```diff
@@ -157,9 +157,12 @@ def get_model(model_name: str, revision: str | None = None, **kwargs: Any) -> En
     model = meta.load_model(**kwargs)
 
     # If revision not available in the modelmeta, try to extract it from sentence-transformers
-    if meta.revision is None and isinstance(model, SentenceTransformer):
-        _meta = model_meta_from_sentence_transformers(model)
-        meta.revision = _meta.revision if _meta.revision else meta.revision
+    if isinstance(model.model, SentenceTransformer):
+        _meta = model_meta_from_sentence_transformers(model.model)
+        if meta.revision is None:
+            meta.revision = _meta.revision if _meta.revision else meta.revision
```
Samoed, line 163:
```suggestion
        meta.revision = _meta.revision if _meta.revision else meta.revision
```

mteb/model_meta.py
```diff
@@ -51,6 +53,15 @@ def get_loader_name(
     return loader.__name__
 
 
+DISTANCE_METRICS = Literal["cosine", "MaxSim", "dot"]
+
+
+class ScoringFunction(str, Enum):
```
Samoed, line 59: For now, you haven’t used the enum, so it can be removed and introduced in a different PR if needed.

mteb/models/overview.py
sam-hey: Yes, you're right that the logic can be condensed into one line, but the example you provided changes the logic slightly:
```
meta.revision = (
    _meta.revision
    if meta.revision is None and _meta.revision
    else meta.revision
)

```
I think this version makes the code harder to read. However, I'm open to hearing other opinions.

isaac-chung:
Model loading test has been fixed here in main as well: https://github.com/embeddings-benchmark/mteb/pull/1775

Samoed:
@isaac-chung I've merged your updates to `v2` branch and on next CI trigger everything should work

mteb/model_meta.py
sam-hey: > @KennethEnevoldsen, what do you think about merging it in its current state? I’ve kept the Enum but reverted the > changes in all the models.
> This should minimize the effort required to merge the new changes while still supporting enums.
>
> Personally, I think this is a good compromise. Backward compatibility is fully maintained.

Let's discuss this with @KennethEnevoldsen

mteb/models/overview.py
sam-hey: @Samoed, what do you think about this? Do you agree, or should I change it to my version?

sam-hey:
All `Enums` have been removed. The changes should be ready for merging now.

tests/test_benchmark/test_models.py
Samoed, line 1: Sorry, last changes! Maybe apply here `pytest.importskip`?

sam-hey: I'm totally here to improve my coding, so I really appreciate a thorough review, even if it means I have to redo parts! @Samoed, thanks for the feedback!

And you're absolutely right it is not so nice —`pytest.importskip` doesn't exist, as far as I can tell. However, `pytest.importorskip` is a similar function, but it’s not a marker.

We could go with something like this, or even write a custom decorator:

```python
try:
    import pylate  # noqa
except ImportError:
    pylate_installed = False

@pytest.mark.skipif(not pylate_installed, reason="PyLate not installed")
```
https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip
https://github.com/pytest-dev/pytest/issues/9548
https://stackoverflow.com/questions/73750328/make-pytest-ignore-gathering-a-file-if-a-module-is-not-installed

Samoed: Yes, I meaned importorskip

sam-hey: Have you reviewed the documentation and the related issue? You can find it here: https://github.com/pytest-dev/pytest/issues/9548.

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip

Using `importorskip` will work, but only if we split each component into its own file, which isn't very practical.

Samoed: You can use it inside test function

Samoed (approved):
Great work!

sam-hey:
Thanks a lot to all of you, @KennethEnevoldsen, @isaac-chung, and @Samoed! 

I have one more question: I believe it would be beneficial to add BM25 and ColBERT tests to the pipeline. I’ve read that the pipeline is currently taking too long, and I’d like to know your thoughts on incorporating the BM25 and ColBERT requirements into the `install-for-tests` section of the Makefile. 

What do you think?

Samoed:
I think they can be integrated, but let's make separate PR

KennethEnevoldsen:
Great work on this @sam-hey! Thanks for taking the time

PR #1768, opened 2025-01-11
feat: add instruct wrapper

Samoed:
## Checklist

Add `InstructWrapper` that uses `SentenceTransformer` models


mteb/models/instruct_wrapper.py
```diff
@@ -78,3 +80,78 @@ def encode(
...
+        **kwargs: Any,
+    ):
+        """Args:
+        model_name: model name
+        revision: revision
+        instruction_template: Model template. Should contain the string '{instruction}'.
+        max_seq_length: maximum sequence length. If None, the maximum sequence length won't be changed.
+        apply_instruction_to_passages: Whether to apply the instruction template to the passages.
+        **kwargs:
+        """
```
KennethEnevoldsen, line 102: format is wrong

It would be nice to add an explanation on when to use this as model models with instructions that will not need this wrapper (e.g. e5 models)

mteb/models/instruct_wrapper.py
```diff
@@ -78,3 +80,78 @@ def encode(
...
+        instruction = self.get_task_instruction(task_name, prompt_type)
+
+        # to passage prompts won't be applied to passages
+        if not self.apply_instruction_to_passages:
+            if task.metadata.type == "p2p":
+                instruction = None
+                logger.info(
+                    f"No instruction used, because task retrieve type {task.metadata.type}"
+                )
+            elif prompt_type == PromptType.passage and task.metadata.type == "s2p":
```
KennethEnevoldsen, line 140: I am not sure why "p2p" and "s2p" are required here. I might be missing something

Samoed: I think most retrieval models are trained with passages being longer than queries (asymmetrical), and in our cases, paragraphs are longer than queries (sentences). I thought it would be better to:  

- Apply prompts for both query and passage in symmetrical `s2s`.  
- Avoid applying prompts to passages in asymmetrical `s2p` for models that don't recommend prompts for passages.  
- Completely skip applying prompts for `p2p` models that don't recommend prompts for passages, though I don't have real-world tests for this scenario.  

In my work, when I applied query prefixes for symmetrical retrieval, I achieved higher scores compared to using standard asymmetrical prompts. Perhaps prompt application strategies could also be adjusted for other models.

If you think that not good idea, then I can change prompt applying based on just `PromptType`

KennethEnevoldsen: Hmm, but s2s can still be `asymetric`:

"how tall is the Eiffel tower?" <-(symmetric)-> "What is the height of the Eiffel tower?"
"how tall is the Eiffel tower?" <-(asymmetric)-> "The Eiffel tower is 200 meters"

I am perfectly fine with adding an annotation to the metadata for symmetric vs asymmetric and using that. However as I understand the MTEB paper s and p refers to length

Samoed: Hm. So for now it would be better to apply prompt based on `PromptType`?

KennethEnevoldsen: Yes indeed

isaac-chung:
Thanks for the effort! 

Following up on [this comment] (https://github.com/embeddings-benchmark/mteb/pull/1768#discussion_r1912530836) from @KennethEnevoldsen, it's not quite clear in this PR why this new wrapper is needed, and why we cannot support prompts in the existing ST wrapper. If there was an existing discussion, it would be great to link here.

Could we please add an explanation clearly with code examples in the PR description (since there's no issue reference), and also a short line in the docstrings?

Samoed:
@isaac-chung This wrapper is designed for models that use prompts to generate embeddings. Previously, this was handled by `InstructWrapper`, but it relied on `gritlm`, which can be a bit difficult for newcomers to understand. A similar wrapper was used with the Jasper model, which was initially included in this PR. However, I removed it because it used `s2p` to decide whether to apply a passage prompt or not, and I wanted to keep this PR more reproducible.

I didn't add this functionality to `SentenceTransformerWrapper` because I believe it’s easier to have a separate wrapper specifically for models using instructions and prefixes. I'm not sure what kind of code example you're looking for, as this wrapper works similarly to all other wrappers.

isaac-chung:
Thanks. I would have liked to discuss something like this in an issue instead of a PR in the future.

1. So how are users supposed to know which wrapper to use, between the gritlm-instruct wrapper and this instruct wrapper? To me, this and having 2 separate ST wrappers are the major points of confusion. E.g. other than the perceived "difficulty to understand" (are there direct quotes from people saying so?", is the gritlm wrapper not usable for supporting instructions with ST models?
2. There is only 1 changed file in this PR and no linked results. What do you mean by "reproducible" here?
3. Why and how is it easier to have a separate one, compared to extending the existing ST wrapper?

isaac-chung:
Thanks for the explanation and the effort you put into this! I think this is a good step forward, but I have a few follow-up questions to help clarify things further:

1. Could we provide more guidance on how users should choose between the gritlm-instruct wrapper and the new instruct wrapper? Having two separate wrappers might introduce some confusion, so it would help to clarify their use cases. For example, are there any user feedback or specific scenarios that prompted this change?


2. I noticed there is only one file changed in this PR. Could you elaborate on what you mean by 'reproducible' here? A bit more context would help me understand how this is addressed.


3. Lastly, could you share more details on why this approach (a separate wrapper) is easier or preferable compared to extending the existing SentenceTransformerWrapper? If there are challenges in doing so, it would be helpful to document them for context.

mteb/models/instruct_wrapper.py
```diff
@@ -78,3 +79,88 @@ def encode(
...
+    def encode(
+        self,
+        sentences: Sequence[str],
+        *,
+        task_name: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> np.ndarray:
+        if self.add_eos_token:
+            # for nv-embed, we add eos token to each input example
```
isaac-chung, line 140: Might be more suitable to move this comment to the nv-embed file, beside the parameter.

mteb/models/instruct_wrapper.py
```diff
@@ -78,3 +79,88 @@ def encode(
...
+        **kwargs: Any,
+    ):
+        """Instruct Sentence Transformer Wrapper. Wrapper that passes instructions to the Sentence Transformer model.
+        Applied for models like NV-Embed, gte-Qwen, e5-mistral, etc.
+
+        Arguments:
+            model_name: Model name
+            revision: Revision
+            instruction_template: Model template. Should contain the string '{instruction}'.
+            max_seq_length: Maximum sequence length. If None, the maximum sequence length won't be changed.
```
isaac-chung, line 103: Could we be more specific about _what_ it might be changed from, e.g.
```suggestion
            max_seq_length: Maximum sequence length. If None, the maximum sequence length will be read from the model config.
```

mteb/models/instruct_wrapper.py
```diff
@@ -78,3 +79,88 @@ def encode(
...
+            apply_instruction_to_passages: Whether to apply the instruction template to the passages.
+            padding_side: Padding side. If None, the padding side won't be changed.
```
isaac-chung, line 105: Same here re: "changed".

mteb/models/instruct_wrapper.py
```diff
@@ -78,3 +79,88 @@ def encode(
...
+        padding_side: str | None = None,
+        add_eos_token: bool = False,
+        **kwargs: Any,
+    ):
+        """Instruct Sentence Transformer Wrapper. Wrapper that passes instructions to the Sentence Transformer model.
+        Applied for models like NV-Embed, gte-Qwen, e5-mistral, etc.
+
+        Arguments:
+            model_name: Model name
+            revision: Revision
```
isaac-chung, line 101:
```suggestion
            model_name: Model name of the sentence transformers model.
            revision: Revision of the sentence transformers model.
```

isaac-chung (commented):
Thanks for the detailed explanation, this makes complete sense now, and I agree with your approach. 

I think adding a section to the README file to document the purpose and usage of the new wrapper would be a great final step. This will help users quickly understand how and when to use it compared to other wrappers.

Great work!

Samoed:
I think we shouldn’t add this to the README because I don’t believe wrappers should be included there. Instead, it’s better to add it to `docs/adding_a_model`, as this change is more developer-focused than user-focused.

docs/adding_a_model.md
```diff
@@ -71,3 +71,18 @@ The leaderboard [automatically refreshes daily](https://github.com/embeddings-be
     ###### Instantiating the Model with Prompts
 
     If you are unable to directly add the prompts in the model configuration, you can instantiate the model using the `sentence_transformers_loader` and pass `prompts` as an argument. For more details, see the `mteb/models/bge_models.py` file.
+
+##### Adding instruction models
+
+Model that using instructions during, then it can use [`InstructSentenceTransformerWrapper`](../mteb/models/instruct_wrapper.py).
```
isaac-chung, line 77:
```suggestion
Models that use instructions can use the [`InstructSentenceTransformerWrapper`](../mteb/models/instruct_wrapper.py). For example:
```

mteb/models/instruct_wrapper.py
```diff
@@ -78,3 +79,86 @@ def encode(
...
+        **kwargs: Any,
+    ):
+        """Instruct Sentence Transformer Wrapper. Wrapper that passes instructions to the Sentence Transformer model.
+        Applied for models like NV-Embed, gte-Qwen, e5-mistral, etc.
+
+        Arguments:
+            model_name: Model name of the sentence transformers model.
+            revision: Revision of the sentence transformers model.
+            instruction_template: Model template. Should contain the string '{instruction}'.
+            max_seq_length: Maximum sequence length. If None, the maximum sequence length will be read from the model config.            apply_instruction_to_passages: Whether to apply the instruction template to the passages.
```
isaac-chung, line 103:
```suggestion
            max_seq_length: Maximum sequence length. If None, the maximum sequence length will be read from the model config. 
            apply_instruction_to_passages: Whether to apply the instruction template to the passages.
```

isaac-chung (approved):
Makes sense. 2 small suggestions and I think we're set here. Thanks again!

isaac-chung:
Good to merge 🙌

Issue #1763, opened 2025-01-11
Missing results for CQDupstack*Retrieval


KennethEnevoldsen:
Results are still missing for CQDupstack*Retrieval:

<img width="1677" alt="Screenshot 2025-01-11 at 17 22 45" src="https://github.com/user-attachments/assets/3f510cef-c69a-45f6-a08c-e23df4508cd0" />

```
NV-Embed-v2
Zeta-Alpha-E5-Mistral
NV-Embed-v1
voyage-large-2-instruct
voyage-multilingual-2
voyage-3
voyage-3-lite
MiniCPM-Embedding
e5-R-mistral-7b
gemma-2b-embeddings
MicroLlama-text-embedding
```

I would rerun on the entire MTEB(eng, classic) (skipping where already run).

Related to #1571

Samoed:
I think this was fixed

PR #1771, opened 2025-01-11
fix: Allow aggregated tasks within benchmarks

### LINKED ISSUE #1231: Allow aggregated tasks within benchmarks
We currently have only one aggregated task (CQGDupstack), however, we can def. imagien more in the future (e.g. for CoIR in https://github.com/embeddings-benchmark/leaderboard/pull/27).

A proposed solution is to use the benchmark (they are already a group of tasks) and then allow a benchmark to be a `list[task | benchmark]`

This will require updated to the `MTEB.MTEB`, as well as the `create_meta` and potentially for CLI.k

This approach should also solve: https://github.com/embeddings-benchmark/mteb/issues/1171

KennethEnevoldsen:
Fixes #1231
Adresses #1763 (though we need a fix for the results repo)

- [x] added AbsTaskAggregated
- [x] added CQADupstackRetrieval
- [x] updated mteb(eng, classic) to use  CQADupstackRetrieval instead of its subtasks
- [x] refactor

Did quite a few refactors here. Not at all settled that this is the right representation, but it is at least much better than where we started. 

We will still need to combine the score of CQGA scores on `embedding-benchmark/results`.

@x-tabdeveloping let me know if this works on the leaderboard end (I believe it should)

## Checklist


mteb/abstasks/aggregated_task.py
```diff
@@ -0,0 +1,128 @@
...
+        self.tasks = self.metadata.tasks
+        self.save_suffix = kwargs.get("save_suffix", "")
+
+        self.seed = seed
+        random.seed(self.seed)
+        np.random.seed(self.seed)
+        torch.manual_seed(self.seed)
+        torch.cuda.manual_seed_all(self.seed)
+
+    def evaluate(
```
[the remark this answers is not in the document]
KennethEnevoldsen: I did quite a large refactor where I just ignored the evaluation method. We can still make it valid, but for now, I don't think it is needed (though this kinda showed me that some sort of `mteb.run_task(model, task)` would be great to have)

mteb/abstasks/AbsTask.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, seed: int = 42, **kwargs: Any):
         np.random.seed(self.seed)
         torch.manual_seed(self.seed)
         torch.cuda.manual_seed_all(self.seed)
+        self.hf_subsets = list(self.metadata.hf_subsets_to_langscripts.keys())
```
KennethEnevoldsen, line 76: Avoiding None here makes a lot of things simpler for me.

mteb/abstasks/aggregate_task_metadata.py
```diff
@@ -0,0 +1,171 @@
...
+    reference: str | None = None
+    bibtex_citation: str | None = None
+
+    @property
+    def hf_subsets_to_langscripts(self) -> dict[HFSubset, list[ISO_LANGUAGE_SCRIPT]]:
+        """Return a dictionary mapping huggingface subsets to languages."""
+        return {"default": self.eval_langs}  # type: ignore
+
+    @model_validator(mode="after")  # type: ignore
+    def compute_unfilled_cases(self) -> AggregateTaskMetadata:
```
KennethEnevoldsen, line 65: A user can still specify these but they can also be computed automatically if left as default.

mteb/evaluation/MTEB.py
```diff
@@ -463,6 +466,29 @@ def run(
                 f"\n\n********************** Evaluating {task.metadata.name} **********************"
             )
 
+            if isinstance(task, AbsTaskAggregate):
```
KennethEnevoldsen, line 469: Recursively resolving Aggregated tasks allows arbitrary nesting of AbsTaskAggregate.

mteb/evaluation/MTEB.py
```diff
@@ -463,6 +466,29 @@ def run(
...
+                self_ = MTEB(tasks=task.metadata.tasks)
+                task_results = self_.run(
+                    model,
+                    verbosity=verbosity - 1,
```
KennethEnevoldsen, line 473: (not sure what the best decision is for verbosity here)

mteb/load_results/task_results.py
```diff
@@ -501,8 +482,8 @@ def get_score_fast(self, splits: str | None, languages: str | None) -> float:
         return val_sum / n_val
 
     @classmethod
-    def from_validated(cls, **data) -> TaskResult:
-        return cls.model_construct(**data)
+     def from_validated(cls, **data) -> TaskResult:
```
x-tabdeveloping, line 485: we have an indentation error here

mteb/load_results/task_results.py
```diff
@@ -536,14 +517,12 @@ def validate_and_filter_scores(self, task: AbsTask | None = None) -> TaskResult:
 
         if task is None:
             task = get_task(self.task_name)
+
         splits = task.metadata.eval_splits
-        if task.is_multilingual:
```
x-tabdeveloping, line 540: I'm not sure I understand this change. Can you explain why this was done?

mteb/models/overview.py
```diff
@@ -30,7 +30,6 @@
     llm2vec_models,
     misc_models,
     model2vec_models,
-    moka_models,
```
x-tabdeveloping, line 33: This might have been an error in resolving merge conflicts, but this definitely shouldn't be removed

mteb/abstasks/AbsTask.py
x-tabdeveloping: Are we sure this won't break other things?

mteb/models/overview.py
Samoed: I got same removals with `make lint`

mteb/abstasks/aggregate_task_metadata.py
```diff
@@ -0,0 +1,171 @@
...
+    name: str
+    description: str
+    dataset: dict[str, Any] = {
+        "path": "aggregate tasks do not have a path",  # just a place holder
+        "revision": "1",
+    }
+
+    tasks: list[AbsTask]
+    main_score: str
+    type: Literal["aggregate-task"] = "aggregate-task"
```
x-tabdeveloping, line 52: This breaks task types on the leaderboard, and also the `TASK_TYPES` type definition. Can't we either make it a property or force people to specify a task type when they introduce an aggregate task?

mteb/models/overview.py
Samoed: I think that in https://github.com/embeddings-benchmark/mteb/commit/748955c367b5c549f4b8d54945361f5bbc7184f6 you didn't add `moka_models` to list of models, but later that was fixed

mteb/abstasks/aggregated_task.py
```diff
@@ -0,0 +1,149 @@
...
+    def task_results_to_scores(
+        self, task_results: list[TaskResult]
+    ) -> dict[str, dict[HFSubset, ScoresDict]]:
+        """The function that aggregated scores. Can be redefined to allow for custom aggregations."""
+        scores = {}
+        for split in self.metadata.eval_splits:
+            main_scores = []
+            for task_res in task_results:
+                main_scores.append(
+                    task_res.get_score(
```
x-tabdeveloping, line 42: You could use _get_scores_fast here, since it matches the default behaviour.

mteb/abstasks/TaskMetadata.py
```diff
@@ -334,6 +336,15 @@ def _check_language_code(code):
                 f"Invalid script code: {script}, you can find valid ISO 15924 codes in {path_to_lang_scripts}"
             )
 
+    @property
+    def bcp47_codes(self) -> list[ISO_LANGUAGE_SCRIPT]:
```
Samoed, line 340: Why did you introduce a new method for filtering languages?

mteb/benchmarks/benchmarks.py
```diff
@@ -148,18 +148,7 @@ def load_results(
...
-                "CQADupstackGisRetrieval",
-                "CQADupstackMathematicaRetrieval",
-                "CQADupstackPhysicsRetrieval",
-                "CQADupstackProgrammersRetrieval",
-                "CQADupstackStatsRetrieval",
-                "CQADupstackTexRetrieval",
-                "CQADupstackUnixRetrieval",
-                "CQADupstackWebmastersRetrieval",
-                "CQADupstackWordpressRetrieval",
+                "CQADupstackRetrieval",
```
Samoed, line 151: Maybe `BEIR` can be added now?

mteb/evaluation/MTEB.py
```diff
@@ -66,7 +69,7 @@ def __init__(
             self._tasks = tasks
             if isinstance(tasks[0], Benchmark):
                 self.benchmarks = tasks
-                self._tasks = list(chain.from_iterable(tasks))
```
x-tabdeveloping, line 69: I'm unsure about this. The idea here is that you would flatten the incoming list, which is a benchmark, which is in turn a list of tasks. If you remove the `chain.from_iterable()` then this will become a list of Benchmark objects, not a list of AbsTask objects. Won't that cause a problem?

x-tabdeveloping (commented):
Thanks for the good work! I left a couple of comments on some things that cause or might cause errors.

mteb/abstasks/AbsTask.py
KennethEnevoldsen: 99% and 50% on that in might actually fix bugs that were there previously (as the None case wasn't handled)

mteb/load_results/task_results.py
KennethEnevoldsen: if it is multilingual then it would have more than 1 subset on huggingface, requiring you to fetch these. However since the format for `task.metadata.hf_subsets_to_langscripts` is always consistent you don't need this handling.

it is a simplification (the cast is actually not nec. I have removed that)

mteb/abstasks/TaskMetadata.py
KennethEnevoldsen: It is not a new method it is a method for fetching languages in the bcp47 format (eng-Latn as opposed to eng). It is used to compute eval langs for the aggregated task (using just language code breaks the tests)

mteb/abstasks/aggregated_task.py
KennethEnevoldsen: fixed. Had to fix typing for get_scores_fast and added default to it as well.

mteb/benchmarks/benchmarks.py
KennethEnevoldsen: Isn't BEIR just a subset of MTEB(eng, classic)? Any reason to not simply use the retrieval score for MTEB(eng, classic)

mteb/evaluation/MTEB.py
KennethEnevoldsen: Hmm you are right, but the current is also problematic:

```py
from itertools import chain

# all benchmarks (works)
list(chain.from_iterable([[1], [2], [3, 4]]))
# [1, 2, 3, 4]

# benchmark and tasks (breaks)
list(chain.from_iterable([1, 2, [3, 4]]))
# TypeError: 'int' object is not iterable

# benchmark and strings (inintended behavior)
list(chain.from_iterable(["111", [2], [3, 4]]))
# ['1', '1', '1', 2, 3, 4]
```

break is the list is more than just benchmarks

I will revert it as @Samoed have already updated the function in v2.

mteb/models/overview.py
KennethEnevoldsen: Indeed it was removed in a lint (not sure why it didn't give an error in 748955c). The version of ruff haven't changed so it should have raised an error? (not the first time I have seen this, might be worth looking into)

KennethEnevoldsen: I have pulled from main and fixed it

mteb/benchmarks/benchmarks.py
Samoed: Yes, but some research still evaluates on `BEIR`, such as `ModernBert`. To simplify things, we could add it, as it only requires the benchmark object and could be helpful for (re)evaluating results.

mteb/abstasks/TaskMetadata.py
Samoed: Maybe we need to standardize how we specify languages https://github.com/embeddings-benchmark/mteb/issues/1822, as the current approach is a bit problematic https://github.com/embeddings-benchmark/mteb/pull/1821#discussion_r1918181081

mteb/load_results/task_results.py
x-tabdeveloping: ahh alright, thanks for explaining

mteb/benchmarks/benchmarks.py
x-tabdeveloping: Yea I think @Samoed has a point here. It would also make researchers' work easier if they had it available out of the box

mteb/abstasks/aggregate_task_metadata.py
KennethEnevoldsen: resolved

mteb/benchmarks/benchmarks.py
KennethEnevoldsen: We can discuss this in an issue - but generally I don't want to overdo it with the benchmarks I would instead add to the mteb(eng, classic) that the retrieval score is eq. to BEIR (cite).

mteb/abstasks/TaskMetadata.py
KennethEnevoldsen: Agree. Let us take the discussion in the issue

KennethEnevoldsen:
Leaderboard seemingly works - @Samoed is it possible that we did not scrape the CQADupstack scores (if so can we do that?)

I will also go over the results repo to and manually create a merged dict for all model which have the scores

Samoed:
No, I think we scrape all of them, e.g. e5-small has CQADupstackWordpressRetrieval https://github.com/embeddings-benchmark/results/blob/6f32ccc978e2d84b905f10b2848479881f43ab02/results/intfloat__e5-small/external/CQADupstackWordpressRetrieval.json#L4

x-tabdeveloping:
I have added aggregate results on CQADupstack for all models in [this PR](https://github.com/embeddings-benchmark/results/pull/101)

PR #1773, opened 2025-01-12
[v2] Remove deprecated parameters from `MTEB` and cli

Samoed:
Now `MTEB` only gets the list of tasks (AbsTask) or `Benchmark` and it won't be possible to run `mteb -m model_name ...`.

## Checklist


isaac-chung:
The [fix](https://github.com/embeddings-benchmark/mteb/pull/1785) for broken test has been merged into main.

mteb/evaluation/MTEB.py
```diff
@@ -5,153 +5,72 @@
...
-                    if "-" not in y:
-                        pair = f"{x}-{y}"
-                        if pair not in langs:
-                            self._task_langs.append(pair)
-        return
-
-    def _display_tasks(self, task_list, name=None):
+        return {x.metadata.category for x in self.tasks}
+
+    def _display_tasks(self, task_list: Iterable[AbsTask], name: str | None = None):
```
isaac-chung, line 73: maybe we can be consistent on whether we expect "task_list" to be a list or Iterable?

mteb/evaluation/MTEB.py
```diff
@@ -5,153 +5,72 @@
...
-            self._task_categories = task_categories
-            self._tasks = None
-
-        self._task_langs = task_langs if task_langs is not None else []
-        if isinstance(self._task_langs, str):
-            self._task_langs = [self._task_langs]
-
-        self._extend_lang_code()
-        self._extend_lang_pairs()  # add all possible pairs
+        self.tasks = deepcopy(tasks)
```
isaac-chung, line 52: Can you help me understand why deepcopy is needed here please?

isaac-chung (commented):
Nice! Added a few suggestions.

mteb/overview.py
```diff
@@ -119,7 +119,7 @@ def filter_task_by_categories(
     return [t for t in tasks if t.metadata.category in _categories]
 
 
-class MTEBTasks(tuple):
+class MTEBTasks(list):
```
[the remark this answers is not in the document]
Samoed: I've got problems with tuple when deleting task from memory `del self.tasks[0]`

mteb/evaluation/MTEB.py
Samoed: In some tests, the results of the run are compared to the original tasks list. However, since running `del self.tasks[0]` here would modify the original list, maybe it might be better to convert the passed tuple of tasks to a list.

tests/test_benchmark/task_grid.py
```diff
@@ -39,31 +33,25 @@
     MockSummarizationTask,
 )
 
-twenty_news = TwentyNewsgroupsClusteringFast()
-
-# downsample to speed up tests
-twenty_news.max_document_to_embed = 1000
-twenty_news.n_clusters = 2
-twenty_news.max_fraction_of_documents_to_embed = None
```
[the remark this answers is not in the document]
Samoed: I’ve updated the tasks grid to use get_tasks instead of directly passing a list of strings and objects. However, this change doesn’t add significant speedups, as tests in this branch still run in 11–14 minutes, which is the same as in https://github.com/embeddings-benchmark/mteb/pull/1774/commits/216aa7cd14ca741c6d950af76dc8d9a1b3a50fba, where tests also run in 11–14 minutes.

mteb/evaluation/MTEB.py
```diff
@@ -5,153 +5,72 @@
...
-        task_categories: list[str] | None = None,
-        task_langs: list[str] | None = None,
-        version=None,
         err_logs_path: str = "error_logs.txt",
         **kwargs,
     ):
         """Create an Evaluation pipeline, based on the provided tasks.
 
         Args:
             tasks: List of tasks to be evaluated.
```
isaac-chung, line 46: bonus: Since we've removed `str` as a possible type to be accepted as `tasks`, we could once more guide users to use `mteb.get_tasks()`, which does accept `str` for tasks. It's already in the docs, but better be safe than sorry)

isaac-chung (approved):
Looks good! With one bonus suggestion (non-blocking).

mteb/evaluation/MTEB.py
Samoed: Are you suggesting adding type checking for the input tasks?

KennethEnevoldsen (approved):
Nice clean-up not a lot to add. Will cause a merge conflict with #1771, but I think it is minor

Issue #1766, opened 2025-01-11
[v2] Getting password prompt for task and tests


sam-hey:
Branch: v2.0.0

Steps to Reproduce:
```
import mteb
from mteb import MTEB

model = mteb.get_model("intfloat/multilingual-e5-small")
tasks = mteb.get_tasks(tasks=["NFCorpus"], languages=["eng"])

eval_splits = ["test"]

evaluation = MTEB(tasks=tasks)

evaluation.run(
    model,
    eval_splits=eval_splits,
    corpus_chunk_size=500,
)
```

Error:
```
INFO:sentence_transformers.SentenceTransformer:Use pytorch device_name: mps
INFO:sentence_transformers.SentenceTransformer:Load pretrained SentenceTransformer: intfloat/multilingual-e5-small
INFO:mteb.models.sentence_transformer_wrapper:Model prompts will be overwritten with {'query': 'query: ', 'passage': 'passage: '}
INFO:mteb.evaluation.MTEB:

## Evaluating 1 tasks:
────────────────────────────────────────────────────────────────────── Selected tasks  ───────────────────────────────────────────────────────────────────────
Retrieval
    - NFCorpus, s2p


INFO:mteb.evaluation.MTEB:

********************** Evaluating NFCorpus **********************
INFO:mteb.abstasks.dataloaders:Loading Corpus...
INFO:mteb.abstasks.dataloaders:Loaded 3633 TEST Documents.
[...]

PR #1774, opened 2025-01-12
fix: [v2] _run_eval() for case: co2_tracker False & add test

sam-hey:
Fixes `_run_eval()` for `co2_tracker=False` and adds a corresponding test

## Checklist
- [x] Run tests locally to make sure nothing is broken using `make test`.  Not all -> https://github.com/embeddings-benchmark/mteb/issues/1766

### DIFF
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -589,7 +589,6 @@ class MTEB:
                             task,
                             model,
                             split,
-                            output_folder,
                             subsets_to_run=subsets_to_run,
                             encode_kwargs=encode_kwargs,
                             **kwargs,
--- a/tests/test_benchmark/test_benchmark.py
+++ b/tests/test_benchmark/test_benchmark.py
@@ -67,6 +67,24 @@ def test_benchmark_encoders_on_task(task: str | AbsTask, model: mteb.Encoder):
     eval.run(model, output_folder="tests/results", overwrite_results=True)
 
 
+@pytest.mark.parametrize("task", [MockMultilingualRetrievalTask])
+@pytest.mark.parametrize(
+    "model",
+    [MockSentenceTransformer()],
+)
+def test_run_eval_without_co2_tracking(task: str | AbsTask, model: mteb.Encoder):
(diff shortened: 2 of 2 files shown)


mteb/evaluation/MTEB.py
```diff
@@ -589,7 +589,6 @@ def run(
                             task,
                             model,
                             split,
-                            output_folder,
```
isaac-chung, line 592: Was this intentional?

isaac-chung (commented):
I'd suggest updating the title to something like "Add test case for running eval without CO2 tracking" to align with the actual changes.

mteb/evaluation/MTEB.py
sam-hey: Yes,` _run_eval()` no longer takes in `output_folder`. However, since `co2_tracker=True` is now the default, no tests failed when updating the `_run_eval` function signature. 

Therefore, this also serves as a fix for cases where `co2_tracker` is not used.
https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/evaluation/MTEB.py#L565:L596
```
 if co2_tracker:
                        logger.warning(
                            "Evaluating multiple MTEB runs simultaniously will produce incorrect CO₂ results"
                        )
                        with EmissionsTracker(
                            save_to_file=False,
                            save_to_api=False,
                            logging_logger=logger,
                            allow_multiple_runs=True,
                        ) as tracker:
                                task,
                                model,
                                split,
                                encode_kwargs=encode_kwargs,
                                subsets_to_run=subsets_to_run,
                                **kwargs,
                            )

                        kg_co2_emissions += (
                            tracker.final_emissions
                        )  # expressed as kilograms of CO₂-equivalents
                    else:
                            task,
                            model,
                            split,
                            output_folder,
                            subsets_to_run=subsets_to_run,
                            encode_kwargs=encode_kwargs,
                            **kwargs,
                        )

```

isaac-chung: Makes sense. Thanks for confirming.

Issue #1783, opened 2025-01-13
Add C-MTEB to benchmarks.py


x-tabdeveloping:
It just occurred to me that we are missing a Chinese benchmark from the new leaderboard that is present in the old one.
Is there a good reason for this, or did we just forget and should it be included?

KennethEnevoldsen:
It should simply be included in benchmarks.py

x-tabdeveloping:
what worries me slightly is that C-MTEB is a different package. Have we even implemented their tasks?

KennethEnevoldsen:
You do but they are submitted to MTEB

x-tabdeveloping:
okay great

PR #1786, opened 2025-01-13
fix: Added C-MTEB

x-tabdeveloping:
Based on issue: #1783 

It does seem however that we are missing quite a few results on the Chinese benchmark.
@Samoed were these not scraped from the old results?

image


x-tabdeveloping:
cc @staoxiao

x-tabdeveloping:
Well, we are missing the mean on quite a few (E5 models, LLM2Vec, GritLM etc.). I can compile a list of tasks and models that are missing if need be.

Samoed:
You can write some tasks that are missing

x-tabdeveloping:
Almost all are missing :D I just compiled a list of all models that are missing for each task, and it's enormous:
[missing_chinese.json](https://github.com/user-attachments/files/18398303/missing_chinese.json)

There could be some weird stuff going on with the evaluation splits, as I got a few errors saying "evaluation split validation is missing"

x-tabdeveloping:
There might be mismatches between task names, but I'm not certain. Something is definitely fishy since we should, in theory have all of this.

Samoed:
I'll into that

x-tabdeveloping:
@Samoed I think I'm onto something. I think there might be an issue with the HF subsets. On the task `AFQMC` we are missing results for `gte-Qwen2-7B-instruct`, but I checked and they are clearly present in the old leaderboard.
It is also there in the results repository:
```json
{
    "dataset_revision": "b44c3b011063adb25877c13823db83bb193913c4",
    "task_name": "AFQMC",
    "evaluation_time": null,
    "mteb_version": null,
    "scores": {
        "validation": [
            {
                "hf_subset": "default",
                "languages": [
                    "cmn-Hans"
                ],
                "cos_sim_pearson": 0.6558442135663871,
                "cos_sim_spearman": 0.722538631361313,
                "euclidean_pearson": 0.7097255486607429,
                "euclidean_spearman": 0.7225374250228647,
                "manhattan_pearson": 0.7083250199989911,
                "manhattan_spearman": 0.7214819496536272,
                "cosine_pearson": 0.6558442135663871,
                "cosine_spearman": 0.722538631361313,
                "main_score": 0.722538631361313
            }
        ]
    }
}
```
On the other hand when I run:
```python
mteb.get_task("AFQMC").hf_subsets
```
I get: `["cmn-Hans"]` not `["default"]`.
I don't know yet how widespread this is, but could be the root cause of our troubles.

x-tabdeveloping:
@KennethEnevoldsen is working on it in a PR rn.

KennethEnevoldsen:
The PR: #1787

PR #1793, opened 2025-01-13
fix: add even more training dataset annotations

### LINKED ISSUE #1556: Track contamination in new leaderboard

KennethEnevoldsen:
Added more annotations on top of the existing

I believe this 
fixes #1556

(at lest for now)

## Checklist

### DIFF
--- a/.github/workflows/model_loading.yml
+++ b/.github/workflows/model_loading.yml
@@ -21,4 +21,4 @@ jobs:
 
     - name: Install dependencies and run tests
       run: |
-        make model-load-test
+        make model-load-test BASE_BRANCH=${{ github.event.pull_request.base.ref }}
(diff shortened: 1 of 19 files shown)


x-tabdeveloping (approved):
Looks good!

PR #1795, opened 2025-01-13
fix: remove default params `TaskMetadata`

Samoed:
Removed default parameters value for `TaskMetadata` to make sure that everything is filled. I thought it would be more problematic.

## Checklist


KennethEnevoldsen:
used to be that way, was changed in #1121. Tests should enforce filled metadata, defaults allow for flexible forking of MTEB for internal company use.

PR #1804, opened 2025-01-14
fix: Fixed eval split for MultilingualSentiment in C-MTEB

x-tabdeveloping:


mteb/benchmarks/benchmarks.py
```diff
@@ -1150,40 +1150,42 @@ def load_results(
...
+                "Cmnli",
+                "CLSClusteringS2S",
+                "CLSClusteringP2P",
+                "ThuNewsClusteringS2S",
+                "ThuNewsClusteringP2P",
+                "ATEC",
+                "BQ",
+                "LCQMC",
+                "PAWSX",
+                "STSB",
```
Samoed, line 1178: Ateq, BQ, STSB have multiple splits too

x-tabdeveloping: and they should only use the test split then?

Samoed: It seems that in previous leaderboard validation split was used https://github.com/embeddings-benchmark/results/blob/4f6a9fc42a2e67922d36ad3d41baf3aa9c969e6e/results.py#L236-L246
But I can't find information in their repo/paper

KennethEnevoldsen: Let us just do whatever they do on the leaderboard.

x-tabdeveloping:
@Samoed I removed the test splits from those tasks, you believe this can be merged yet?

Samoed (approved):
Great! Yes, I think this can be merged

Issue #1809, opened 2025-01-15
Leaderboard: Mean results scored incorrectly


Samoed:
Example: on `MTEB(multilingual, beta)` `infgrad/jasper_en_vision_language_v1` missing results for `PawsXPairClassification`, but have score for `Mean (TaskType)` and `PairClassification`
![image](https://github.com/user-attachments/assets/d3f93737-e179-4763-b575-7adf85c9ee0a)

x-tabdeveloping:
hmm curious. Must be something in `leaderboard/table.py`. I have a piece of code for the aggregation of these results, it might be incorrect.

x-tabdeveloping:
@Samoed I'm on it, will put it in a PR with other fixes.

x-tabdeveloping:
It's because in `pandas.Series.mean()` the default option for `skipna` is `False`. I'll never get over why they decided to do this xd

x-tabdeveloping:
Fixed in #1810

Issue #1807, opened 2025-01-15
Leaderboard: Zero-shot filter is buggy


x-tabdeveloping:
Zero-shot button only filters out models with unknown training data when asked to but not ones that are known not to be zeroshot.

PR #1810, opened 2025-01-15
fix: Zero shot and aggregation on Leaderboard

x-tabdeveloping:
Added a number of minor fixes:
 - `join_revision` now prefers when revision is specified as opposed to `"no_revision_available"`
 - Fixed problems with zero-shot filtering #1807 
 - Fixed aggregation issues on a task type level #1809

### DIFF
--- a/mteb/load_results/benchmark_results.py
+++ b/mteb/load_results/benchmark_results.py
@@ -13,8 +13,7 @@ from packaging.version import InvalidVersion, Version
 from pydantic import BaseModel, ConfigDict
 
 from mteb.abstasks.AbsTask import AbsTask, ScoresDict
-from mteb.abstasks.TaskMetadata import (ISO_LANGUAGE_SCRIPT, TASK_DOMAIN,
-                                        TASK_TYPE)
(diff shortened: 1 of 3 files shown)


mteb/load_results/benchmark_results.py
```diff
@@ -260,8 +260,16 @@ def parse_version(version_str: str) -> Version | None:
...
             is_main_revision = group["revision"] == group["main_revision"]
-            if is_main_revision.sum() == 1:
-                return group[is_main_revision]
+            # If the main revision is present we select that
+            if is_main_revision.sum() > 0:
+                return group[is_main_revision].head(n=1)
+            unique_revisions = group["revision"].unique()
+            # Filtering out no_revision_available if other revisions are present
+            if (len(unique_revisions) > 1) and (
+                "no_revision_available" in unique_revisions
```
Samoed, line 269: Maybe we can ignore `no_revision_available` and just rerun old results?

x-tabdeveloping: I don't think we can, at least for now. It would be lovely to rerun all old results, but I would stick with this for now. I think the plan is that we will release the leaderboard soon, so we need all the results we can get.

Issue #1803, opened 2025-01-14
Missing `ModelMeta` for Chinese models


x-tabdeveloping:
Many of the models that have been run on the original C-MTEB and we have results on are currently missing `ModelMeta` objects in the library.

Here's a list of **Chinese-specific** models that have yet to be added to MTEB:
```python
missing_meta_chinese = [
    "BAAI/bge-base-zh", # not planned, outdated
    "BAAI/bge-base-zh-v1.5",
    "BAAI/bge-large-zh", # not planned, outdated
    "BAAI/bge-large-zh-noinstruct",  # not planned, outdated
    "BAAI/bge-large-zh-v1.5",
    "BAAI/bge-small-zh",  # not planned, outdated
    "BAAI/bge-small-zh-v1.5",
    "DMetaSoul/Dmeta-embedding-zh-small",
    "DMetaSoul/sbert-chinese-general-v1",
    "Erin/IYun-large-zh",
    "Erin/mist-zh",
    "Pristinenlp/alime-embedding-large-zh",
    "Pristinenlp/alime-reranker-large-zh",
    "RookieHX/bge_m3e_stella",
    "akarum/cloudy-large-zh",
    "arkohut/jina-embeddings-v2-base-zh",
    "dunzhang/stella-large-zh-v3-1792d",
    "dunzhang/stella-mrl-large-zh-v3.5-1792d",
    "fangxq/XYZ-embedding-zh",
    "fangxq/XYZ-embedding-zh-v2",
    "iampanda/zpoint_large_embedding_zh",
    "infgrad/stella-base-zh",
    "infgrad/stella-base-zh-v2",
    "infgrad/stella-base-zh-v3-1792d",
    "infgrad/stella-large-zh",
    "infgrad/stella-large-zh-v2",
    "jinaai/jina-embeddings-v2-base-zh",
    "moka-ai/m3e-base",
    "moka-ai/m3e-large",
    "neofung/m3e-ernie-xbase-zh",
    "sensenova/piccolo-base-zh",
    "sensenova/piccolo-large-zh",
    "sensenova/piccolo-large-zh-v2",
    "shanghung/stella-base-zh-v3-1792d",
    "shibing624/text2vec-base-chinese",  # Needs custom implementation
    "shibing624/text2vec-large-chinese",  # Needs custom implementation
    "silverjam/jina-embeddings-v2-base-zh",
    "thenlper/gte-base-zh",
    "thenlper/gte-large-zh",
[...]

PR #1814, opened 2025-01-15
fix: Added more Chinese models' `ModelMeta`

x-tabdeveloping:
Added the following model families:
 - SBERT multilingual USE models
 - Moka
 - DMeta
 - Jina-zh
 - Piccolo


x-tabdeveloping:
#1803

Samoed (approved):
Great!

PR #1818, opened 2025-01-15
ci: only return 1 model_name per file

### LINKED ISSUE #1756: Error Model Loading / extract-and-run
When doing commits, it appears
```
System.IO.IOException: No space left on device : '/home/runner/runners/2.321.0/_diag/Worker_20250110-152956-utc.log' ...
```
example run https://github.com/embeddings-benchmark/mteb/actions/runs/12712643686/job/35438686654?pr=1749
@isaac-chung

isaac-chung:
Fixes #1756 

A change to `mteb/models/sentence_transformers_models.py` only results in `sentence-transformers/all-MiniLM-L6-v2` written to `scripts/model_names.txt` now using the `make model-load-test` command. The scripts can still be used separately to run existing options.

## How this affects implementing new models
It shouldn't and it doesn't. When models are added from scratch, the PRs are to follow the [relevant docs](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_model.md) and PR checklist. We saw an influx of model meta because of the leaderboard backfill. Every change to a model file now will trigger a model load test that takes 1 model from each file. The rest (loading, running, getting results) does not change and will remain the same as before.

This CI workflow is mainly for validating any updates made to model wrappers. I'd see it as a sanity check, because we should expect model wrappers to be tested (run on datasets to confirm changes are correct) before merging.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,5 +41,5 @@ build-docs:
 model-load-test:
 	@echo "--- 🚀 Running model load test ---"
 	pip install ".[dev, speedtask, pylate,gritlm,xformers,model2vec]"
-	python scripts/extract_model_names.py $(BASE_BRANCH)
+	python scripts/extract_model_names.py $(BASE_BRANCH) --return_one_model_name_per_file
 	python tests/test_models/model_loading.py --model_name_file scripts/model_names.txt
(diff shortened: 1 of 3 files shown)


Samoed (approved):
Great work!

PR #1821, opened 2025-01-15
fix: add bge-m3 `ModelMeta`

Samoed:
ref https://github.com/embeddings-benchmark/mteb/issues/1803

## Checklist

### DIFF
--- a/mteb/models/bge_models.py
+++ b/mteb/models/bge_models.py
@@ -89,6 +89,215 @@ bge_chinese_training_data = {
     # "s2orc": [],  # (title, abstract) (title, citation title) (abstract, citation abstract)
 }
 
+bgem3_training_data = {
+    # source https://arxiv.org/abs/2402.03216
+    "T2Retrieval": ["train"],
+    "DuReader": ["train"],
+    "MMarcoReranking": ["train"],
+    "CMedQAv2-reranking": ["train"],
+    "HotpotQA": ["train"],
+    "NQ": ["train"],
+    "MSMARCO": ["train"],
+    "MrTidyRetrieval": ["train"],
+    "MIRACLRetrieval": ["train"],
+    "CodeSearchNet": ["train"],
+    # not in mteb
+    # "s2orc"
+    # Wikipedia
+    # "xP3"
+    # "mC4"
+    # "CC-News"
+    # "MTP"
+    # "NLLB"
+    # "CCMatrix"
+    # TriviaQA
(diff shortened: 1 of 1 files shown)


mteb/models/bge_models.py
```diff
@@ -89,6 +89,215 @@
...
+    # mMARCO-ZH
+    # LawGPT
+    # NLI-zh2, LeCaRDv2,
+    # NLI, MultiLongDoc (their syntetic)
+}
+
+# https://huggingface.co/BAAI/bge-m3/discussions/29
+bgem3_languages = [
+    "afr_Latn",  # af
+    # als
```
x-tabdeveloping, line 127: I'm a bit unsure why these are commented out?

x-tabdeveloping:
Looks great otherwise, feel free to merge!

mteb/models/bge_models.py
Samoed: I've taken these language codes from the discussion, but I can't find them in the language mapping or I'm not sure which ones they correspond to.

x-tabdeveloping: oooh okay ChatGPT usually does a remarkable job at matching these, there is also a Python library that can do this for you, wait a sec I'll find it

x-tabdeveloping: @Samoed It's called [ISO639](https://pypi.org/project/python-iso639/) and it feels like magic

Samoed: I think that we have not all langs in LANG_MAPPING https://github.com/embeddings-benchmark/mteb/blob/d7a77918cc0e8b8f03cbbe5199e8a0fe58e429d9/mteb/evaluation/LangMapping.py#L5

Samoed: Hm. Semes that `LANG_MAPPING` used only in MTEB class. I think this should be removed in v2

x-tabdeveloping: hmm yea interesting

Samoed: I can't find an easy way to get script from the language, so I'll leave it as is for now.

x-tabdeveloping: Hmm, again, I think LLMs can be a good friend in doing that. If you have the name of the language you're also probably a google search away from the solution. And most languages use Latin, Arab or Cyrillic script anyway, so there are some sensible defaults to go with.

PR #1787, opened 2025-01-13
feat: Update task filtering, fixing bug which included cross-lingual tasks in overly many benchmarks

KennethEnevoldsen:
- Updated task filtering adding exclusive_language_filter and hf_subset
- fix bug in MTEB where cross-lingual splits were included
- also fix bug where "en-ext" was included in AmazonCounterfactualClassification
- added missing language filtering to MTEB(europe, beta) and MTEB(indic, beta)

fixes #1757
fixes https://github.com/embeddings-benchmark/mteb/issues/1757

The following code outlines the problems:

```py
import mteb
from mteb.benchmarks import MTEB_ENG_CLASSIC

task = [t for t in MTEB_ENG_CLASSIC.tasks if t.metadata.name == "STS22"][0]
# was eq. to:
task = mteb.get_task("STS22", languages=["eng"])
task.hf_subsets
# correct filtering to English datasets:
# ['en', 'de-en', 'es-en', 'pl-en', 'zh-en']
# However it should be:
# ['en']

# with the changes it is:
task = [t for t in MTEB_ENG_CLASSIC.tasks if t.metadata.name == "STS22"][0]
task.hf_subsets
# ['en']
# eq. to
task = mteb.get_task("STS22", hf_subsets=["en"])
# which you can also obtain using the exclusive_language_filter (though not if there was multiple english splits):
task = mteb.get_task("STS22", languages=["eng"], exclusive_language_filter=True)
```

edit: This also fixes another bug, where ´"default"´ hf_subset was specified as its language code. In practice this does not influence how models are run, but it influences how results are loaded in the leaderboard (see https://github.com/embeddings-benchmark/mteb/pull/1786).

```py
for task in MTEB_ENG_CLASSIC.tasks:
    print(task.metadata.name, "-", task.hf_subsets)


# in main
# AmazonCounterfactualClassification - ['en-ext', 'en']
# AmazonPolarityClassification - ['eng-Latn']
# AmazonReviewsClassification - ['en']
# ArguAna - ['eng-Latn']
# ArxivClusteringP2P - ['eng-Latn']
# ArxivClusteringS2S - ['eng-Latn']
# AskUbuntuDupQuestions - ['eng-Latn']
# BIOSSES - ['eng-Latn']
# ...

# in branch:
# AmazonCounterfactualClassification - ['en']
# AmazonPolarityClassification - ['default']
# AmazonReviewsClassification - ['en']
# ArguAna - ['default']
# ArxivClusteringP2P - ['default']
[...]

PR #1830, opened 2025-01-17
fix: subsets to run

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/pull/1787#issuecomment-2598205049

Now if run 
```bash
mteb run -m sentence-transformers/all-MiniLM-L6-v2 -t STS17 -l eng
```
It would give error
```
INFO:mteb.abstasks.AbsTask:
Task: STS17, split: test, subset: ko-ko. Running...
ERROR:mteb.evaluation.MTEB:Error while evaluating STS17: 'ko-ko'
  File "/root/conda/envs/debug/bin/mteb", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/root/conda/envs/debug/lib/python3.11/site-packages/mteb/cli.py", line 387, in main
    args.func(args)
  File "/root/conda/envs/debug/lib/python3.11/site-packages/mteb/cli.py", line 145, in run
    eval.run(
  File "/root/conda/envs/debug/lib/python3.11/site-packages/mteb/evaluation/MTEB.py", line 630, in run
    raise e
  File "/root/conda/envs/debug/lib/python3.11/site-packages/mteb/evaluation/MTEB.py", line 583, in run
                          ^^^^^^^^^^^^^^^
  File "/root/conda/envs/debug/lib/python3.11/site-packages/mteb/evaluation/MTEB.py", line 304, in _run_eval
    results = task.evaluate(
              ^^^^^^^^^^^^^^
  File "/root/conda/envs/debug/lib/python3.11/site-packages/mteb/abstasks/AbsTask.py", line 130, in evaluate
    data_split = self.dataset[hf_subset][split]
                 ~~~~~~~~~~~~^^^^^^^^^^^
KeyError: 'ko-ko'
```
CC @izhx

## Checklist


tests/test_evaluation/test_split_evaluation.py
```diff
@@ -362,3 +363,12 @@ def test_all_splits_subsets_evaluated_with_overwrite(
...
         assert sorted(results2[0].languages) == ["eng", "fra"]
+
+
+def test_splits_evaluated_with_prefiltering():
+    """Test that the evaluation only runs on the specified languages. Issue https://github.com/embeddings-benchmark/mteb/pull/1787#issuecomment-2598205049"""
+    task = MockMultilingualSTSTask().filter_languages(languages=["fra"])
+
+    evaluation = MTEB(tasks=[task])
+
+    evaluation.run(MockSentenceTransformer(), overwrite_results=True)
```
isaac-chung, line 374: Could we also assert the desired subsets are run, like we do in the test above?

Samoed: Added!

isaac-chung (commented):
Thanks! I realized that I missed something in the previous round. My apologies. 

The fix is reverting a change made to "allow overwrite of pre-filtering". However, the added test case seem to cover only pre-filtering. 

If tests were passing in the last PR where this line was introduced (which caused the issue), and the added test case isn't addressing the failure case, then we're still missing that test case - overwriting pre-filtering - in this PR, right? Am I missing something else?

tests/test_benchmark/mock_tasks.py
```diff
@@ -927,12 +927,10 @@ def load_data(self, **kwargs):
...
         }
 
-        self.dataset = DatasetDict(
-            {
-                "eng": data,
-                "fra": data,
-            }
-        )
+        self.dataset = {}
+        for lang in self.hf_subsets:
```
isaac-chung, line 931: When I added a break point here, `self.hf_subsets=['fra']`, but I think we still want `['eng','fra']` here, right?

Samoed: I don't think that we need to overwrite subsets if user limit them in other part

Samoed:
No, I’ve added a test case that reproduces the failing scenario described in the comment, and without this fix, it wouldn’t pass

tests/test_benchmark/mock_tasks.py
Samoed: Issue caused by https://github.com/embeddings-benchmark/mteb/blob/762f729b70cdad2e9137a68af4b2693ca96dd3b7/mteb/abstasks/MultiSubsetLoader.py#L40-L47

isaac-chung: Oh YES, my mistake. CLI args like lang will be passed into get_tasks, which does the pre-filtering. Got it. we're good 👍

isaac-chung (approved):
I overlooked a small fact. Looks good here!

Issue #1792, opened 2025-01-13
[MIEB] performance of VOC2007


gowitheflow-1998:
am just finalizing MIEB results for overall table in the paper and find VOC2007 results a bit strange https://github.com/embeddings-benchmark/tmp/tree/master (e.g., E5-v gets 70%+ `lrap` and Voyage multi-modal gets like 20% - these two have very similar performance trend on all other 120+ tasks (voyage multi-modal typically is slightly better); VLM2Vec-full gets 70%+ and VLM2vec-lora gets ~20% as well).

Since this is the only multi-label classification task, I am trying to merge it into the regular linear-probe task type in the leaderboard table. This is pretty much the last task that needs to take a look if anyone has bandwidth!

cc @isaac-chung

isaac-chung:
I'll take a look. Since I haven't found scores for these models on VOC2007, I'm going to check whether this dataset is absolutely correct and first see if the current VOC2007 matches CLIP benchmark numbers on OpenAI CLIP models. 

There are also 2 different multilabel datasets:
- https://huggingface.co/datasets/HuggingFaceM4/pascal_voc (currently using)
- https://huggingface.co/datasets/clip-benchmark/wds_voc2007_multilabel

Do you see any major discrepancies for VLM2vec-lora or voyage multi-modal on other tasks?

isaac-chung:
This could be related to https://github.com/embeddings-benchmark/mteb/pull/1420, where the `samples_per_label` were too low to observe scaling laws. On `openai/clip-vit-base-patch32`:

Here is a markdown table based on your specifications:

```markdown
| samples_per_label | 8     | 16    | 32    |
|-------------------|-------|-------|-------|
| map               | 0.660 | 0.757 | 0.801 |
| lrap              | 0.660 | 0.757 | 0.801 |
```

It's not entirely 1:1 with CLIP benchmark since it's not zero-shot. 

If VLM2Vec-full gets 70%+, does that mean samples_per_label=64 or something?

gowitheflow-1998:
> I'll take a look. Since I haven't found scores for these models on VOC2007, I'm going to check whether this dataset is absolutely correct and first see if the current VOC2007 matches CLIP benchmark numbers on OpenAI CLIP models.

scores are here:
Voyage: https://github.com/embeddings-benchmark/tmp/blob/master/voyage-multimodal-3/1/VOC2007.json
[...]

PR #1834, opened 2025-01-19
[mieb] fixing lrap computation for multi-label classification

gowitheflow-1998:
Fixes https://github.com/embeddings-benchmark/mteb/issues/1792

Fixing the computation of the `lrap` metric for multi-label classification. `lrap` is supposed to [operate on continuous scores](https://scikit-learn.org/1.5/modules/generated/sklearn.metrics.label_ranking_average_precision_score.html), instead of predicted labels. This change largely smooths the large performance range in `lrap` across models brought by change in `samples_per_label`.

Was trying to fix the same for main but got a chain of failed tests (mainly because `main` is using `KNeighborsClassifier()` by itself - which I'm not sure would work in the expected way either; `mieb` is currently using `MultiOutputClassifer(estimator=LogisicRegression())`, which makes each label a separate binary classification problem).

## Checklist


isaac-chung (approved):
Great catch!

Issue #1842, opened 2025-01-20
add zero-shot definition to new LB


Muennighoff:
I think it is important the definition is shown on the LB; maybe an unfoldable box at the bottom like for borda count? (and then remove the current explanation of the signs at the top)

isaac-chung:
Related to https://github.com/embeddings-benchmark/mteb/issues/1760

x-tabdeveloping:
I added this in #1849

Issue #1839, opened 2025-01-20
nan in leaderboard


Muennighoff:
Still nans in the leaderboard - if it is missing scores, should we just make the field be empty instead of nan?

<img width="1372" alt="Image" src="https://github.com/user-attachments/assets/0ec08cb8-05db-4a30-a73e-3010c4f9b385" />

x-tabdeveloping:
Fixed in #1849

Issue #1838, opened 2025-01-20
Leaderboard: Make benchmark not `beta` and give them better descriptions


x-tabdeveloping:
Currently, many of the benchmarks in the new leaderboard have a `beta` label beside them. We should probably change this, and in the descriptions explain it clearer what the benchmarks are for.
For instance I can imagine that a new user would be confused as to what the difference is between MTEB(eng, classic) and MTEB(eng, beta) or why we made a new English benchmark in the first place.

x-tabdeveloping:
Fixed in #1849

Issue #1760, opened 2025-01-11
Defining zero-shot for MTEB 


KennethEnevoldsen:
The next version of the MTEB leaderboard will soon be released and with it a new zero-shot filter. However, we are currently planning to use the following definition of zero-shot. This issue is to open that discussion up to the community to ensure that we consider all relevant views.  

> **Zero Shot**
A model is considered zero-shot if it is not trained on other splits of the dataset used to derive the task.
E.g., if a model is trained on Natural Questions, it cannot be considered zero-shot on benchmarks containing the task “NQ” which is derived from Natural Questions.
This definition creates a few edge cases. For instance, multiple models are typically trained on Wikipedia title and body pairs, but we do not define this as leakage on, e.g., “WikipediaRetrievalMultilingual” and “WikiClusteringP2P” as these datasets are not based on title-body pairs.
Distilled, further fine-tunes or in other ways, derivative models inherit the datasets of their parent models.
Based on community feedback and research findings, This definition could change in the future.

williambarberjr:
How about a "thought experiment":
I'm an enterprise picking which embedding model to use. I have two options:

1. A retrieval model trained by billion dollars of funding AI lab #3. It's reportedly been trained on every piece of human text ever written, including all the text in the MTEB training set. It is on the top of the MTEB leaderboard and reports from other enterprises using this model are that it's working quite well for their use case.
2. A late interaction model exclusively trained on MS Marco that generalizes much more effectively to MTEB data not seen during training than other models that I can also isolate with a zero shot filter. This model represents a major achievement in cutting edge techniques for sample efficiency, better architecture etc, as demonstrated in it's performance against models only trained on MS Marco that are then tested on MTEB.

Assume the inference cost, storage costs of the embeddings, and complexity of the retrieval implementations are the same for the sake of the thought experiment. Which model do you want to use?

My primary point is that - if the goal is for MTEB to be more useful for researchers trying to determine what training techniques are more sample efficient and produce a model that generalizes better, then this zero shot filter would be helpful. But for businesses trying to decide which model to implement in production, the waters are permanently muddied by competition in model training pivoting to centering around scaling up self-supervised and supervised learning to the largest, cleanest datasets possible.

[...]

PR #1849, opened 2025-01-21
fix: Leaderboard Refinements

x-tabdeveloping:
Couple of fixes:
 - removed Beta and added better descriptions to benchmarks per #1838 
 - Filled NaN values with blank in table per #1839 
 - Added zero-shot definition per #1842 from #1760 
 - Zero-shot filtering didn't work properly on start up, so I fixed that

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -6,6 +6,7 @@ import logging
 import tempfile
 import time
 from pathlib import Path
+from typing import Literal
 from urllib.parse import urlencode
 
 import gradio as gr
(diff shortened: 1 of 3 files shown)


mteb/leaderboard/app.py
```diff
@@ -117,23 +120,75 @@ def update_task_info(task_names: str) -> gr.DataFrame:
     return gr.DataFrame(df, datatype=["markdown"] + ["str"] * (len(df.columns) - 1))
 
 
+# Model sizes in million parameters
+MIN_MODEL_SIZE, MAX_MODEL_SIZE = 0, 10_000
+
+
+def filter_models(
+    model_names,
```
Samoed, line 128: Maybe it would be better to add a method to MTEB that filters model metas?

Samoed (approved):
Great!

mteb/leaderboard/app.py
Samoed, line 128: Can you add type hints?

Samoed:
Will the models' ranks still be computed if not all benchmark results are available, especially now that you've fixed the NaN issues? https://github.com/embeddings-benchmark/mteb/issues/1813#issuecomment-2595841432

x-tabdeveloping:
Yea, so, NaN accounts to zero votes in the ranks, and the blanking out only happens in formatting, this shouldn't cause any difference in behaviour.

mteb/leaderboard/app.py
x-tabdeveloping: Dun

PR #1852, opened 2025-01-21
fix: Fixed leaderboard search bar

x-tabdeveloping:
Fixed issues with table not updating when searched, and also when no results are found.
Additionally, now the callback only gets called once you press enter, instead of with every change.


isaac-chung (approved):
Looks good. I did not verify locally.

Issue #1848, opened 2025-01-21
Missing models on leaderboards [WIP]


KennethEnevoldsen:
I checked to top 10 models for each leaderboard. They seem to be missing the following scores

MTEB(eng, classic):
- NV-Embed-v2
  - CQADupstackRetrieval
  - STS22
- NV-Embed-v1
  - MSMARCO
  - CQADupstackRetrieval
  - STS22
- Missing models:
  - BAAI/bge-en-icl (no meta, no results)
  - yibinlei/LENS-d8000 (no meta, no results)
  - yibinlei/LENS-d4000 (no meta, no results)

MTEB(chinese)
- missing models:
  - [Conan-embedding-v1](https://huggingface.co/TencentBAC/Conan-embedding-v1) (no meta, no results)

@Samoed can I ask you to add the missing models (results to the results repo + model meta). Feel free to add a filler class for "modelnotimplemented" in the loader (otherwise we will never catch up with model releases).

PR #1856, opened 2025-01-22
fix: Adding missing model meta

x-tabdeveloping:
Working on this issue #1848 .
So far added annotations and meta to CDE and bge-en-icl.


Samoed (approved):

x-tabdeveloping:
Though curiously enough they do not show up in the leaderboard. Might be something with the results repo or loading

mteb/models/gme_models.py
```diff
@@ -0,0 +1,255 @@
...
+from transformers import AutoModelForVision2Seq, AutoProcessor
+
+from mteb.model_meta import ModelMeta
+
+from .wrapper import Wrapper
+
+logger = logging.getLogger(__name__)
+
+
+class GMEWrapper(Wrapper):
```
Samoed, line 20: Maybe we shouldn't add wrappers from `mieb` for now. I think you could handle it like `cde` by filling only the metadata.

Samoed (approved):
Great work!

mteb/models/gme_models.py
x-tabdeveloping: Alright I will delete the wrapper then.

PR #1794, opened 2025-01-13
fix: Remove default params, `public_training_data` and `memory usage` in `ModelMeta`

Samoed:
Removed default params in `ModelMeta` to make sure that authors filled this information. For now, I fill most of missing values with `None`, but this needs to be filled more relevant to the models. 

Removed `public_training_data` and `memory usage` from `ModelMeta`.

## Checklist

- [X] Run tests locally to make sure nothing is broken using `make test`. 
- [X] Run the formatter to format the code using `make lint`.

PR #1857, opened 2025-01-22
fix: Hotfixed public_training_data type annotation

x-tabdeveloping:
It was removed then readded with the incorrect type and now every test is failing. This should fix it, but we should also probably add some docs and be on the same page on what this property is actually for.

### DIFF
--- a/mteb/model_meta.py
+++ b/mteb/model_meta.py
@@ -95,7 +95,7 @@ class ModelMeta(BaseModel):
     license: str | None
     open_weights: bool | None
     public_training_code: str | None
-    public_training_data: str | None
+    public_training_data: str | bool | None
     framework: list[FRAMEWORKS]
     reference: STR_URL | None = None
     similarity_fn_name: DISTANCE_METRICS | None


Samoed (commented):
It was changed in https://github.com/embeddings-benchmark/mteb/pull/1846 and I don't think that `bool` provide a lot of information.

mteb/model_meta.py
```diff
@@ -95,7 +95,7 @@ class ModelMeta(BaseModel):
     license: str | None
     open_weights: bool | None
     public_training_code: str | None
-    public_training_data: str | None
+    public_training_data: str | bool | None
```
isaac-chung, line 98: Looking at https://github.com/embeddings-benchmark/mteb/pull/1794 where this was first removed, do we want to revert these back to `bool` for both "public" fields, instead of `str`?

isaac-chung (commented):
More of a question:

isaac-chung:
@x-tabdeveloping just wondering if we should do the same for both "public_" fields. I don't mind deferring the discussion with you/Kenneth till later on how it can be best refactored, which it seemed needed but missing in the linked PR.

x-tabdeveloping:
Well, I also think it's a bad solution, but it keeps breaking things, so let's allow bool for now, and then we can iron this out later.
Debating this should be the least of our concerns now, especially considering that the property is not currently being used in the leaderboard.

mteb/model_meta.py
x-tabdeveloping: Since public_training_code hasn't caused any problems anywhere to my knowledge, I would keep it as is

PR #1882, opened 2025-01-27
feat: Added Hindi benchmark.

SaileshP97:
### Code Quality

- [ ] **Code Formatted**: Format the code using `make lint` to maintain consistent style.

### Documentation

### Testing

### Adding datasets checklist

### Adding a model checklist


mteb/benchmarks/benchmarks.py
```diff
@@ -1276,3 +1276,45 @@ def load_results(
     year={2024}
 }""",
 )
+
+MTEB_INDIC = Benchmark(
```
x-tabdeveloping, line 1280: Would this not overwrite the Hindic benchmark above? I suppose this should have a different variable name.

mteb/benchmarks/benchmarks.py
```diff
@@ -1276,3 +1276,45 @@ def load_results(
...
+    name="MTEB(Hindi)",
+    tasks=get_tasks(
+        languages=["hin"],
+        tasks=[
```
x-tabdeveloping, line 1284: Are you not adding any novel tasks to the benchmark? How will this benchmark be different from selecting Hindi as the language on `MTEB(Indic)`?

mteb/benchmarks/benchmarks.py
```diff
@@ -1276,3 +1276,45 @@ def load_results(
...
+            "MintakaRetrieval",
+            "MIRACLRetrievalHardNegatives",
+            "MultiLongDocRetrieval",
+            "WikipediaRetrievalMultilingual",
+            # reranking
+            "MIRACLReranking",
+            "WikipediaRerankingMultilingual",
+        ],
+        exclusive_language_filter=True,
+        eval_splits=["test", "validation", "dev"],
```
x-tabdeveloping, line 1314: Are you sure you want to use all of these splits for all tasks? Perhaps narrowing it down might be beneficial for certain tasks, if you think model trainers might use the dev and validation splits for validating their models.

mteb/benchmarks/benchmarks.py
x-tabdeveloping: I would also advise you to think about making the benchmark as zero-shot as possible. Meaning if many models train on a certain dataset, you should probably avoid adding it to the benchmark, as the performance might not be representative of the model's ability to generalize.
You can get these for a given task by saying:
```python
from collections import defaultdict
import mteb

model_metas = mteb.get_model_metas()
tasks = mteb.get_benchmark("MTEB(Hindi)").tasks
task_names = [task.metadata.name for task in tasks]
models_trained_on_tasks = defaultdict(list)
for model_meta in model_metas:
    if (model_meta.training_datasets is not None):
          for training_dataset in model_meta.training_datasets:
                 if training_dataset in task_names:
                     models_trained_on_tasks[model_meta.name].append(training_dataset)

print(models_trained_on_tasks)
# And then this would print something like:
{
     "Model1": ["task_1", "task_2", ...],
     ...
}
```

mteb/benchmarks/benchmarks.py
```diff
@@ -1276,3 +1276,45 @@ def load_results(
...
+    ),
+    description="The Hindi Leaderboard benchmark extends the MTEB framework by incorporating Hindi-specific datasets and tasks derived from existing MTEB data. It evaluates text embedding models on a variety of tasks, including text classification, semantic similarity, and information retrieval, with a focus on Hindi language performance. This benchmark aims to provide a standardized evaluation platform to advance research and innovation in Hindi NLP, leveraging pre-existing datasets to ensure consistency and comparability across models.",
+    reference=None,
```
x-tabdeveloping, line 1317: Do you not have a technical report yet? How are you progressing with the paper?

x-tabdeveloping (commented):
I've added a couple of questions, and things you should consider, but this is a good start :D

mteb/benchmarks/benchmarks.py
SaileshP97: Yes, I didn't saw this. I will update it.

mteb/benchmarks/benchmarks.py
SaileshP97: Right now I don't have any technical report because we have not added any new model or dataset for any task. But we are planning to work on new model. So we decide to add about this benchmark in that paper.

So is it possible to add citation later?

mteb/benchmarks/benchmarks.py
SaileshP97: I thought each dataset has either test, validation or dev split.

mteb/benchmarks/benchmarks.py
x-tabdeveloping: If the benchmark is subject to change in the near future, I'm wondering if it's worth it to merge it now. In any case I would add a `beta` marker in the name to suggest that it will change in the future. (`name="MTEB(Hindi, beta)"`).
What do you think @KennethEnevoldsen ?

mteb/benchmarks/benchmarks.py
Samoed: Yes, and for each task you should select needed splits

mteb/benchmarks/benchmarks.py
```diff
@@ -1276,3 +1276,41 @@ def load_results(
...
+            "IndicQARetrieval",
+            "MintakaRetrieval",
+            "MultiLongDocRetrieval",
+            # reranking
+            "WikipediaRerankingMultilingual",
+        ],
+        exclusive_language_filter=True,
+        eval_splits=["test"],
+    ) + get_tasks(tasks=["HindiDiscourseClassification",
+                        "SentimentAnalysisHindi"], eval_splits=["train"]),
```
Samoed, line 1310: Tasks should not use train splits for evaluation. They're automatically used during training for classification

SaileshP97: When I checked the eval_splits for the task it only had 'train' in `/tasks/Classification/hin/HindiDiscourseClassification.py`. 
So If I set eval_splits='test' it will still used train split? or validation is not possible on this task?

Samoed: I think, yes train split is used, because it seems that this dataset have only one split

mteb/benchmarks/benchmarks.py
KennethEnevoldsen: def. add the beta marker

KennethEnevoldsen:
@SaileshP97 seems like this PR might have gotten a bit stale - are you still working on this?

isaac-chung:
Feel free to reopen if you decide to pick this back up.

PR #1843, opened 2025-01-20
Add FaMTEB (Farsi/Persian Text Embedding Benchmark)

mehran-sarmadi:
We are a research team from Sharif University of Technology and [MCINext Company](https://www.linkedin.com/company/mcinext/posts/?feedView=all) developing a text embedding benchmark for the Persian language based on MTEB. So far, we have gathered around 63 datasets spanning 7 tasks (Classification, Clustering, Pair Classification, Reranking, Retrieval, STS, and Summary Retrieval), including a mix of existing, translated, and newly generated datasets. Notably, we are introducing the Summary Retrieval task for the first time, which focuses on identifying the correct summary of a paragraph from a set of candidates. We have also evaluated several Persian language models and text embeddings that support Persian for this benchmark.

We also open related PR for the [results](https://github.com/embeddings-benchmark/results/pull/92) and [leaderboard tab](https://github.com/embeddings-benchmark/leaderboard/pull/67), and we are finalizing a paper on this work, which will be published in the near future.

## Checklist

- [X] Run tests locally to make sure nothing is broken using `make test`. 
- [X] Run the formatter to format the code using `make lint`. 

### Adding datasets checklist

**Reason for dataset addition**: ... 

- [X] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb -m {model_name} -t {task_name}` command.
  - [X] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [X] `intfloat/multilingual-e5-small`
- [X] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [X] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [X] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [X] Run tests locally to make sure nothing is broken using `make test`. 
- [X] Run the formatter to format the code using `make lint`.

PR #1915, opened 2025-02-01
fix: remove SummaryRetrieval as a type

### LINKED ISSUE #1913: [Leaderboard] Can't select `FaMTEB`
When I'm selecting `FaMTEB` I've got error 
```
Value: 'SummaryRetrieval' (type: ) is not in the list of choices: ['Summarization', 'STS', 'Speed', 'Reranking', 'BitextMining', 'Retrieval', 'InstructionRetrieval', 'Clustering', 'MultilabelClassification', 'PairClassification', 'Classification']
```

KennethEnevoldsen:
fixes #1913 

### Code Quality

### Documentation

### Testing

### DIFF
--- a/mteb/abstasks/TaskMetadata.py
+++ b/mteb/abstasks/TaskMetadata.py
@@ -97,7 +97,6 @@ TASK_TYPE = Literal[
     "Summarization",
     "InstructionRetrieval",
     "Speed",
-    "SummaryRetrieval",
 ]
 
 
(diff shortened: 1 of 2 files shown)


isaac-chung:
Renaming the tasks breaks the Farsi benchmark introduced in https://github.com/embeddings-benchmark/mteb/pull/1843. I don't have a preference on naming, but we should include the change in benchmark.py

Perhaps we can also add in the description that these are "summary retrieval" tasks, just using the bitext mining abstask. I can add these.

KennethEnevoldsen:
I don't think it actually breaks it (the task still has the same name, it is only the object that it renamed)

Issue #1916, opened 2025-02-01
CI: Update task table generated too many commits


isaac-chung:
On the main branch after the commit 8fb59a49b00e7932abec42c045c1cc068c7eba41 (~ 3 days ago), the job "Update task table" normally should run only once. Now it often run 10+ times, generating a big amount of commits that are not related to PRs.

PR #1922, opened 2025-02-01
docs: Add sort to domains for task metadata

KennethEnevoldsen:
Tests currently go into an infinite loop. This should prevent that.


isaac-chung:
Closes https://github.com/embeddings-benchmark/mteb/issues/1916.

PR #1925, opened 2025-02-02
docs: Add instruction for running leaderboard

### LINKED ISSUE #1923: Leaderboard - Instructions for running locally somewhere in README?
Could be useful/intereting to some people maybe

KennethEnevoldsen:
@x-tabdeveloping should we rename it from demo to app?

fixes #1923

### Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,26 @@ model_w_contamination = ModelMeta(
 </details>
 
 <details>
+  <summary>  Running the leaderboard locally </summary>
+
+
+### Running the Leaderboard
+
+It is possible to completely deploy the leaderboard locally or self-host it. This can e.g. be relevant for companies that might want to
+integrate build their own benchmarks or integrate custom tasks into existing benchmarks. 
+
+Running the leaderboard is quite easy. Simply run:
+```py
+from mteb.leaderboard import demo
+
+demo.launch()
+```
+
+The leaderboard requires gradio install, which can be installed using `pip install mteb[gradio]` and requires python >3.10.
+
+</details>
+
+<details>
   <summary>  Caching Embeddings To Re-Use Them </summary>
 
 
--- a/README.md
+++ b/README.md
@@ -457,10 +457,7 @@ integrate build their own benchmarks or integrate custom tasks into existing ben
 
 Running the leaderboard is quite easy. Simply run:
 ```py
-from mteb.leaderboard import demo
-
-demo.launch()
-```
+python -m mteb.leaderboard.app
 
 The leaderboard requires gradio install, which can be installed using `pip install mteb[gradio]` and requires python >3.10.
 


Samoed:
Maybe we could integrate a command into the CLI for running the leaderboard?

KennethEnevoldsen:
> Maybe we could integrate a command into the CLI for running the leaderboard?

Agree. (@x-tabdeveloping might have some thoughts on this?)

with MIEB I (if we end up with two separate apps) we might also choose to do something like:

```bash
mteb leaderboard --benchmarks MTEB(eng), MTEB(eng, classic), ...  # limit selection to only text task for MMTEB leaderboard
mteb leaderboard --benchmarks MIEB,MIEB lite # only image+text benchmarks
```

README.md
```diff
@@ -444,6 +444,26 @@ model_w_contamination = ModelMeta(
...
+integrate build their own benchmarks or integrate custom tasks into existing benchmarks. 
+
+Running the leaderboard is quite easy. Simply run:
+```py
+from mteb.leaderboard import demo
+
+demo.launch()
+```
+
+The leaderboard requires gradio install, which can be installed using `pip install mteb[gradio]` and requires python >3.10.
```
isaac-chung, line 465: Note: We might want to consider dropping support for Python 3.9 in tests? 
cc @imenelydiaker as we were discussing that offline.

isaac-chung (approved):
The add instruction part is complete. Maybe we could separate out the rename bit and the command line bit to separate issues/PRs.

README.md
```diff
@@ -444,6 +444,26 @@ model_w_contamination = ModelMeta(
...
+<details>
+  <summary>  Running the leaderboard locally </summary>
+
+
+### Running the Leaderboard
+
+It is possible to completely deploy the leaderboard locally or self-host it. This can e.g. be relevant for companies that might want to
+integrate build their own benchmarks or integrate custom tasks into existing benchmarks. 
+
+Running the leaderboard is quite easy. Simply run:
```
x-tabdeveloping, line 458: You can actually just run it like this: `python3 -m mteb.leaderboard.app` Then you don't have to interact with the `demo` thing at all.

x-tabdeveloping:
@KennethEnevoldsen The reason it's still called demo, is because, when you debug, Gradio actually looks for the name `demo`, which is a bit weird, but I chose to live with it

README.md
KennethEnevoldsen: We can def. consider it, but what would the argument be?

README.md
```diff
@@ -444,6 +444,26 @@ model_w_contamination = ModelMeta(
...
+```py
+from mteb.leaderboard import demo
+
+demo.launch()
+```
```
KennethEnevoldsen, line 463:
```suggestion
```py
python -m mteb.leaderboard.app
```
```

README.md
imenelydiaker: I guess because 3.9 retirement is scheduled on next October. Tests take a lot of time to run (more than 20min and sometimes they timeout), we are running 4 versions of Python with 2 versions of OS for 3.9. I'm wondering if the minimal requirement for mteb shouldn't be 3.10 now

Issue #1832, opened 2025-01-18
[v2] Remove MultilingualTask


Samoed:
I think we can integrate the `MultilingualTask` class into the `AbsTask` class. The most useful function is already there, and it can be easily integrated into `load_data`.  

https://github.com/embeddings-benchmark/mteb/blob/ce5cb3e859322d77f32019434f3d7cf9d72b02f5/mteb/abstasks/MultilingualTask.py#L56-L61

Additionally, we should reupload datasets with `fast_loading`, as using `parquet` files will make them much faster.
https://github.com/embeddings-benchmark/mteb/blob/ce5cb3e859322d77f32019434f3d7cf9d72b02f5/mteb/abstasks/MultilingualTask.py#L39-L52

KennethEnevoldsen:
completely agree with this

PR #1926, opened 2025-02-02
[v2] Remove multilingual task

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/1832

I've removed `MultilingualTask` and explicitly added `is_multilingual` to tasks, which was previously handled by `MultilingualTask`. I also moved the `fast_loading` method from `MultilingualTask` to `AbsTask`, but I think we should reupload these datasets in the correct format.  

Additionally, we might want to determine `is_multilingual` automatically based on `eval_langs`, but I’m not entirely in favor of this since it feels a bit like magic. Another option is to integrate `is_multilingual` into `TaskMetadata`.

### Code Quality

### Documentation

### Testing


mteb/abstasks/AbsTask.py
```diff
@@ -84,6 +86,7 @@ class AbsTask(ABC):
     data_loaded: bool = False
     is_multilingual: bool = False
```
KennethEnevoldsen, line 87: can't this simply be derived from eval_langs? (if it is a dictionary with more than one/default key?)

mteb/abstasks/AbsTask.py
```diff
@@ -74,6 +74,8 @@ class AbsTask(ABC):
         abstask_prompt: The potential prompt of the abstask
         superseded_by: Denotes the task that this task is superseeded by. Used to issue warning to users of outdated datasets, while maintaining
             reproducibility of existing benchmarks.
+        is_multilingual: Denotes if the task is multilingual. If True, the task will be evaluated on all languages in the metadata.eval_langs.
+        fast_loading: (Deprecated) Denotes if the task should be loaded using the fast loading method. This is only possible if the dataset have a "default" config
```
KennethEnevoldsen, line 78: Why is this deprecated?

mteb/abstasks/AbsTask.py
```diff
@@ -224,10 +228,42 @@ def load_data(self, **kwargs):
...
-        self.dataset = datasets.load_dataset(**self.metadata.dataset)  # type: ignore
+        if self.is_multilingual:
+            if self.fast_loading:
+                self.fast_load()
+            else:
+                self.dataset = {}
+                for lang in self.hf_subsets:
+                    self.dataset[lang] = datasets.load_dataset(
+                        name=lang,
+                        **self.metadata.dataset,
```
KennethEnevoldsen, line 239: I believe we can use this in all cases. (multilingual or not) (is the is_multilingual even required?)

```suggestion
                for hf_subset in self.hf_subsets:
                    self.dataset[hf_subset] = datasets.load_dataset(
                        name= hf_subset,
                        **self.metadata.dataset,
```
As it can also be something other than language (see brightretrieval)

mteb/abstasks/AbsTask.py
```diff
@@ -224,10 +228,42 @@ def load_data(self, **kwargs):
...
+                    )
+        else:
+            self.dataset = datasets.load_dataset(**self.metadata.dataset)  # type: ignore
         self.dataset_transform()
         self.data_loaded = True
 
+    def fast_load(self, **kwargs):
+        # todo remove
```
KennethEnevoldsen, line 247: ?

KennethEnevoldsen:
I have also made some changes on this over on main. I would probably do a merge before (my fix removed `.is_multilingual`). It was required to allow for the aggregate task I believe.

mteb/abstasks/AbsTask.py
Samoed: I think we shouldn't add new datasets with `fast_loading` because it downloads the full dataset split and after that splitting by language split, and these datasets are not very large. I checked, and the largest one is around 10MB, but they use `jsonl.gz`. We could convert them to `parquet` and load splits the same way as for other tasks.

mteb/abstasks/AbsTask.py
Samoed: https://github.com/embeddings-benchmark/mteb/pull/1926#discussion_r1938471947

mteb/abstasks/AbsTask.py
KennethEnevoldsen: So it is more a "don't use this feature" - got it.

KennethEnevoldsen: Might be worth writing that out. We don't recommend using this feature, instead we ...

Samoed:
No, `is_multilingual` is still in `AbsTask`.
https://github.com/embeddings-benchmark/mteb/blob/dba7a952e39a8d4a53aab967d761354ae030ca54/mteb/abstasks/AbsTask.py#L65

KennethEnevoldsen:
(we could also consider merging this and resolving it in a merge/partial merge)

mteb/abstasks/AbsTask.py
Samoed: I have written in the PR description. I can change it like that, but not totally sure
> Additionally, we might want to determine is_multilingual automatically based on eval_langs, but I’m not entirely in favor of this since it feels a bit like magic. Another option is to integrate is_multilingual into TaskMetadata.

KennethEnevoldsen:
> No, is_multilingual is still in AbsTask.

Yes, but it is not in `.evaluate` (which I see it also isn't here, that is my mistake).

So yea, no merge is needed

mteb/abstasks/AbsTask.py
KennethEnevoldsen: Sorry missed that. let us keep it as is

mteb/abstasks/AbsTask.py
```diff
@@ -224,10 +228,42 @@ def load_data(self, **kwargs):
...
+                self.fast_load()
+            else:
+                self.dataset = {}
+                for hf_subset in self.hf_subsets:
+                    self.dataset[hf_subset] = datasets.load_dataset(
+                        name= hf_subset,
+                        **self.metadata.dataset,
+                    )
+        else:
+            self.dataset = datasets.load_dataset(**self.metadata.dataset)  # type: ignore
```
KennethEnevoldsen, line 242:
```suggestion
            if self.fast_loading:
                self.fast_load()
            else:
                self.dataset = {}
                for hf_subset in self.hf_subsets:
                    self.dataset[hf_subset] = datasets.load_dataset( 
                        name= hf_subset,
                        **self.metadata.dataset,
                    )
```
I assume this should work in all cases?

KennethEnevoldsen (commented):
Alright, looking a bit more into this. Are there cases where we can't replace `.is_multilingual`?

mteb/abstasks/AbsTask.py
Samoed: Yes, I think so

Samoed:
No, we can totally remove `is_multilingual`. If you think we should do this, I will remove it

mteb/abstasks/AbsTask.py
Samoed: Some monolingual datasets include the split name in their `metadata.dataset`, which causes this change to break them since `name=hf_subset` is passed. I'll roll back this change.

mteb/abstasks/AbsTask.py
```diff
@@ -74,6 +74,10 @@ class AbsTask(ABC):
         abstask_prompt: The potential prompt of the abstask
         superseded_by: Denotes the task that this task is superseeded by. Used to issue warning to users of outdated datasets, while maintaining
             reproducibility of existing benchmarks.
+        is_multilingual: Denotes if the task is multilingual. If True, the task will be evaluated on all languages in the metadata.eval_langs.
+        fast_loading: (Not recommended to use) Denotes if the task should be loaded using the fast loading method.
```
isaac-chung, line 78: Isn't v2 for breaking changes? Just thinking of when this would be removed, and whether we can simply remove it in this PR if it's not recommended.

mteb/abstasks/MultilingualTask.py
```diff
@@ -1,61 +0,0 @@
-from __future__ import annotations
-
-import datasets
-
-from .AbsTask import AbsTask
-
-
-class MultilingualTask(AbsTask):
```
isaac-chung, line 8: We should add a warning in `main` for all the things we plan to remove in V2.

mteb/tasks/BitextMining/kat/TbilisiCityHallBitextMining.py
```diff
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
-from datasets import DatasetDict, load_dataset
+import datasets
```
isaac-chung, line 3: This seems like a no-op change. Suggest to revert this (and the related change below in the same file).

docs/adding_a_dataset.md
```diff
@@ -76,7 +76,7 @@ evaluation = MTEB(tasks=[SciDocsReranking()])
 evaluation.run(model)
 ```
 
-> **Note:** for multilingual / crosslingual tasks, make sure your class also inherits from the `MultilingualTask` class like in [this](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Classification/multilingual/MTOPIntentClassification.py) example.
+> **Note:** for multilingual / crosslingual tasks, make sure your class added `is_multiligual=True` to class attributes like in [this](../mteb/tasks/Classification/multilingual/MTOPIntentClassification.py) example.
```
isaac-chung, line 79:
```suggestion
> **Note:** for multilingual / crosslingual tasks, make sure you've added `is_multiligual=True` to the class attributes like in [this](../mteb/tasks/Classification/multilingual/MTOPIntentClassification.py) example.
```

isaac-chung (commented):
Maybe I'm missing something: how do we determine which subsets to load? Does `is_multilingual` play a role or do we just check for the presence of a `default` subset?

mteb/tasks/BitextMining/kat/TbilisiCityHallBitextMining.py
Samoed: This breaking test with test that `datasets.load_data` is called

mteb/abstasks/AbsTask.py
Samoed: Yes, but I'll leave it as is for now. I'll create a new PR to completely remove this since it's a bit out of scope for this PR.

isaac-chung: Okay sounds good.

Samoed:
> how do we determine which subsets to load?

By `hf_subsets`. If user want to run some part of them, then they require to run `mteb.get_tasks` that will filter required subsets. For example, for language filter.
https://github.com/embeddings-benchmark/mteb/blob/dba7a952e39a8d4a53aab967d761354ae030ca54/mteb/abstasks/AbsTask.py#L315

> Does is_multilingual play a role or do we just check for the presence of a default subset?

Previously yes, but now we can get subsets with `hf_subsets_to_langscripts` (maybe we can create separate method just for getting splits), that return them correctly. For now, I think we can totally remove it

mteb/tasks/BitextMining/kat/TbilisiCityHallBitextMining.py
Samoed: Or maybe this problem with `datasets`, because I have same error in https://github.com/embeddings-benchmark/mteb/pull/1927

mteb/abstasks/AbsTask.py
Samoed: Created issue https://github.com/embeddings-benchmark/mteb/issues/1929

mteb/abstasks/MultilingualTask.py
isaac-chung: @Samoed could you please add a PR for this?

isaac-chung:
Sorry, I didn't phrase the question right. What I'm curious about is the evidence supporting this statement:
>  we can totally remove is_multilingual.

tests/test_tasks/test_all_abstasks.py
```diff
@@ -32,7 +32,6 @@ def test_load_data(
         isinstance(task, AbsTaskRetrieval)
         or isinstance(task, AbsTaskReranking)
         or isinstance(task, AbsTaskSpeedTask)
-        or isinstance(task, MultilingualTask)
```
isaac-chung, line 35: Maybe a check for `task.is_multilingual` is needed here to account of this. Removing this likely cause the test failure.

mteb/tasks/BitextMining/kat/TbilisiCityHallBitextMining.py
isaac-chung: See https://github.com/embeddings-benchmark/mteb/pull/1926#discussion_r1938749660

mteb/abstasks/MultilingualTask.py
Samoed: Yes, I'll create

tests/test_tasks/test_all_abstasks.py
Samoed: But I didn't change this in #1927, but tests still failing there

Samoed: It seems I've deleted it there too somehow. Will revert there

isaac-chung: I think I've found the fix in this commit 7a96b37

Samoed: Yes, I forgot to readd it. Will revert more 2 tasks changes

Samoed:
> Sorry, I didn't phrase the question right. What I'm curious about is the evidence supporting this statement:
> 
> > we can totally remove is_multilingual.

Yes, for now, `is_multilingual` is only used to check the default split.

```python
hf_subsets = list(self.dataset) if self.is_multilingual else ["default"]
```

isaac-chung:
> Yes, for now, is_multilingual is only used to check the default split.

Hmm it seems like there are a few more references, like
- In descriptive stats: https://github.com/embeddings-benchmark/mteb/blob/d0eb782c88d6652a3976cf00fece7ad7d2fe8d33/mteb/abstasks/AbsTask.py#L290
- In upload dataset to hub: https://github.com/embeddings-benchmark/mteb/blob/d0eb782c88d6652a3976cf00fece7ad7d2fe8d33/mteb/abstasks/AbsTask.py#L386

Do they seem easy to refactor?

Samoed:
I think we can make is_multilingual just property of TaskMetadata that would be calculated based on `eval_splits`

docs/adding_a_dataset.md
```diff
@@ -76,7 +76,7 @@ evaluation = MTEB(tasks=[SciDocsReranking()])
 evaluation.run(model)
 ```
 
-> **Note:** for multilingual / crosslingual tasks, make sure you've added `is_multiligual=True` to the class attributes like in [this](../mteb/tasks/Classification/multilingual/MTOPIntentClassification.py) example.
+> **Note:** for multilingual / crosslingual tasks, make sure you've specified `eval_langs` as dictionary. [Example](../mteb/tasks/Classification/multilingual/MTOPIntentClassification.py).
```
isaac-chung, line 79: Making the link part of the sentence.
```suggestion
> **Note:** for multilingual / crosslingual tasks, make sure you've specified `eval_langs` as a dictionary, as shown in [this example](../mteb/tasks/Classification/multilingual/MTOPIntentClassification.py).
```

isaac-chung (approved):
Looks good! Just got one small suggestion.

Issue #1729, opened 2025-01-08
Remove "memory_usage" from ModelMeta


KennethEnevoldsen:
I don't believe it is specified by any model and previously is was computed based on the parameters anyway. I would just remove it since we are currently not using it.

PR #1934, opened 2025-02-02
add model memory usage

Samoed:
To make fully compatible old leaderboard with new, we can add model memory usage. Add property to `ModelMeta`.

### Code Quality

### Documentation

### Testing


isaac-chung:
Did you ask @KennethEnevoldsen ? I'd prefer not to have such back-and-forth: to add then remove then add back things. https://github.com/embeddings-benchmark/mteb/issues/1729

Samoed:
Opened issue for this https://github.com/embeddings-benchmark/mteb/issues/1935. Previously it was removed, because it was not filled and should be autocalculated based on parameters

mteb/model_meta.py
```diff
@@ -146,3 +146,15 @@ def is_zero_shot_on(self, tasks: list[AbsTask]) -> bool | None:
         model_datasets = {ds_name for ds_name, splits in self.training_datasets.items()}
         intersection = model_datasets & benchmark_datasets
         return len(intersection) == 0
+
+    @property
+    def memory_usage(self) -> int | None:
+        """Calculates the memory usage (in FP32) of the model in GB."""
```
isaac-chung, line 152: docstrings says GB but the return value is named mb? 
It seems the calculation matches GB, only the naming is off.

mteb/model_meta.py
```diff
@@ -146,3 +146,15 @@ def is_zero_shot_on(self, tasks: list[AbsTask]) -> bool | None:
...
+        if self.n_parameters is None or self.embed_dim is None:
```
isaac-chung, line 153: How is embed_dim used? If unused, can we remove the condition?

isaac-chung (commented):
I understand now (we're not adding this back to be filled in - we are auto-calculating it here. Great stuff!). Thanks for the initiative! Just 2 clarifications then I think we're ready.

mteb/model_meta.py
Samoed: Yes, I can change this. I switched from `MB` to `GB` to match the old leaderboard, but I think the leaderboard could handle this dynamically—for example, if the model size is greater than 1024 MB, it could automatically change the suffix to `GB`.

isaac-chung (approved):
Sweet!

mteb/model_meta.py
```diff
@@ -146,3 +146,15 @@ def is_zero_shot_on(self, tasks: list[AbsTask]) -> bool | None:
         model_datasets = {ds_name for ds_name, splits in self.training_datasets.items()}
         intersection = model_datasets & benchmark_datasets
         return len(intersection) == 0
+
+    @property
+    def memory_usage(self) -> float | None:
```
x-tabdeveloping, line 151: we could perhaps add the unit in the name to avoid confusion (`memory_usage_mb`)

mteb/model_meta.py
```diff
@@ -146,3 +146,15 @@ def is_zero_shot_on(self, tasks: list[AbsTask]) -> bool | None:
...
+        """Calculates the memory usage (in FP32) of the model in MB."""
+        if self.n_parameters is None:
+            return None
+        # Model memory in bytes. For FP32 each parameter is 4 bytes.
+        model_memory_bytes = self.num_params * 4
```
x-tabdeveloping, line 156: Is this a good assumption to make? Do all models have FP32 parameters?

x-tabdeveloping (commented):
Left a couple of minor comments. Thanks for adding this

mteb/model_meta.py
Samoed: Large models (>1B) params usally loaded with fp16/bp16, but I don't know how to handle this automatically

x-tabdeveloping: And then do a `cached_property` so that it doesn't have to be fetched every time

x-tabdeveloping: I agree. We could fetch all of these in a script and manually keep count of them in ModelMeta perhaps?

Samoed: Calculated them

Samoed:
@x-tabdeveloping Can this PR be merged?

KennethEnevoldsen (approved):
Looks good on my end - great addition. Do we want to create a column for it on the leaderboard (if so let us make an issue on that)

Samoed:
@KennethEnevoldsen I've already created issue

Issue #1941, opened 2025-02-03
[v2] Add warning for any breaking changes in `main`


isaac-chung:
For any breaking changes, e.g. removing parameters, changes in interface, we should add a warning to warn users so that they can expect such significant changes. Here is an example: https://github.com/embeddings-benchmark/mteb/pull/1940

@Samoed as you've been very actively contributing to v2, I hope you don't mind being assigned to this for now 🙏 We can start accumulating a list of changes before adding the warning PR for previous changes in one go. For future changes, we should probably have one accompanying PR along with the one to the v2 branch.

cc @KennethEnevoldsen

PR #1924, opened 2025-02-02
[v2] remove `save_suffix`

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/1770

### Code Quality

- [X] **Code Formatted**: Format the code using `make lint` to maintain consistent style.

PR #1820, opened 2025-01-15
[v2] remove metadata_dict

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/issues/1791

## Checklist

- [X] Run tests locally to make sure nothing is broken using `make test`. 
- [X] Run the formatter to format the code using `make lint`.

PR #1945, opened 2025-02-03
misc: add warnings

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/1941

### Code Quality

### Documentation

### Testing

### DIFF
--- a/mteb/abstasks/AbsTaskClassification.py
+++ b/mteb/abstasks/AbsTaskClassification.py
@@ -78,14 +78,14 @@ class AbsTaskClassification(AbsTask):
         super().__init__(**kwargs)
         if method != "logReg":
             warnings.warn(
-                "Passing method to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
+                "Passing `method` to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
                 DeprecationWarning,
             )
         self.method = method
 
         if n_experiments:
             warnings.warn(
-                "Passing n_experiments to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
+                "Passing `n_experiments` to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
                 DeprecationWarning,
             )
 
@@ -98,7 +98,7 @@ class AbsTaskClassification(AbsTask):
 
         if k != 3:
             warnings.warn(
-                "Passing k to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
+                "Passing `k` to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
                 DeprecationWarning,
             )
         # kNN parameters
@@ -122,7 +122,7 @@ class AbsTaskClassification(AbsTask):
 
         if train_split != "train":
             warnings.warn(
-                "Passing train_split to AbsTaskClassification.evaluate is deprecated and will be removed in v2.0.0.",
+                "Passing `train_split` to AbsTaskClassification.evaluate is deprecated and will be removed in v2.0.0.",
                 DeprecationWarning,
             )
 
--- a/mteb/abstasks/AbsTaskClassification.py
+++ b/mteb/abstasks/AbsTaskClassification.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import warnings
 from collections import Counter, defaultdict
 from typing import Any
 
@@ -75,7 +76,18 @@ class AbsTaskClassification(AbsTask):
         **kwargs,
     ):
         super().__init__(**kwargs)
+        if method != "logReg":
+            warnings.warn(
+                "Passing method to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
+                DeprecationWarning,
+            )
         self.method = method
+
+        if n_experiments:
+            warnings.warn(
+                "Passing n_experiments to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
+                DeprecationWarning,
+            )
 
         # Bootstrap parameters
         self.n_experiments: int = (  # type: ignore
@@ -84,6 +96,11 @@ class AbsTaskClassification(AbsTask):
             else self.metadata_dict.get("n_experiments", 10)
         )
 
+        if k != 3:
+            warnings.warn(
+                "Passing k to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
+                DeprecationWarning,
+            )
         # kNN parameters
         self.k = k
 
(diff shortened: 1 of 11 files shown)


mteb/abstasks/AbsTaskClassification.py
```diff
@@ -75,15 +76,31 @@ def __init__(
         **kwargs,
     ):
         super().__init__(**kwargs)
+        if method != "logReg":
+            warnings.warn(
+                "Passing method to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
```
isaac-chung, line 81: Using this quotation to denote parameters. Can apply to the rest.
```suggestion
                "Passing `method` to AbsTaskClassification is deprecated and will be removed in v2.0.0.",
```

mteb/abstasks/MultilingualTask.py
```diff
@@ -1,12 +1,17 @@
...
+
 from .AbsTask import AbsTask
 from .MultiSubsetLoader import MultiSubsetLoader
 
 
 class MultilingualTask(MultiSubsetLoader, AbsTask):
     def __init__(self, hf_subsets: list[str] | None = None, **kwargs):
         super().__init__(**kwargs)
+        warnings.warn(
+            "`MultilingualTask` will be removed in v2.0.0.", DeprecationWarning
```
isaac-chung, line 13: Could briefly mention whether or what contributors are expected to do differently when they have multilingual tasks.

isaac-chung (commented):
Wow so fast, thanks! Could you mind listing out the PR links that this PR covers? This way we'll be able to know that we've covered the previous ones. I counted 19 `warnings.warn` here, but I understand there might not be 19 PRs to v2. The future warnings would be 1-to-1 so we're good there.

e.g.
- https://github.com/embeddings-benchmark/mteb/pull/1924: Remove `save_suffix` from AbsTask

mteb/abstasks/MultilingualTask.py
Samoed: Added more information

Samoed:
- Multilingual Task was removed in https://github.com/embeddings-benchmark/mteb/pull/1926
- https://github.com/embeddings-benchmark/mteb/pull/1820
- task_types, task_categories, task_langs, version were removed from `MTEB` in https://github.com/embeddings-benchmark/mteb/pull/1773
- Loading local files in retrieval loader were removed in https://github.com/embeddings-benchmark/mteb/pull/1750
- Passing parameters to tasks (e.g. n_experiments and k) and evaluators were removed in https://github.com/embeddings-benchmark/mteb/pull/1707

mteb/abstasks/AbsTaskReranking.py
```diff
@@ -70,6 +71,10 @@ class AbsTaskReranking(AbsTask):
     abstask_prompt = "Retrieve text based on user query."
 
     def __init__(self, **kwargs):
+        warnings.warn(
+            "`AbsTaskReranking` will be merged with AbsTaskRetrieval in v2.0.0.",
```
isaac-chung, line 75: I wasn't able to connect this to either #1750 or #1707 . Was there something I missed?

mteb/evaluation/MTEB.py
```diff
@@ -96,29 +97,29 @@ def deprecation_warning(
...
-                "The `task_langs` argument is deprecated and will be removed in the next release. "
+            warnings.warn(
+                "The `task_langs` argument is deprecated and will be removed in the 2.0 release. "
                 + "Please use `tasks = mteb.get_tasks(... languages = [...])` to filter tasks instead. "
                 + "Note that this uses 3 letter language codes (ISO 639-3)."
             )
         if version is not None:
-            logger.warning(
+            warnings.warn(
                 "The `version` argument is deprecated and will be removed in the next release."
```
isaac-chung, line 117: Not this one?
```suggestion
                "The `version` argument is deprecated and will be removed in the 2.0 release."
```

isaac-chung (commented):
Looks good! Just 2 small questions.

mteb/abstasks/AbsTaskReranking.py
Samoed: https://github.com/embeddings-benchmark/mteb/pull/1359, I'll update comment

isaac-chung (approved):
🎉

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -44,6 +45,10 @@ def __init__(
             # By default fetch qrels from same repo not a second repo with "-qrels" like in original
             self.hf_repo_qrels = hf_repo_qrels if hf_repo_qrels else hf_repo
         else:
+            warnings.warn(
```
ashokrajab, line 48: @Samoed May i know why are planning to restrict Loading from local files?

Samoed: Currently, we don’t use it in MTEB and I think this was left from BEIR as is, and the code for the retrieval dataloader is already quite complicated. You can override `load_data` if you want to load local data

ashokrajab: Even if we were to override `load_data`, we will be instantiating a `HFDataLoader` object within the overriden `load_data` method.
So having a check within `HFDataLoader` `__init__()` to throw exception (in furture) when loading a local file seems not quite right.

Ideally, in this [line](https://github.com/embeddings-benchmark/mteb/blob/67f7ad92d2d8db8e9ee6332e542f27a038ccb232/mteb/abstasks/AbsTaskRetrieval.py#L301) mteb/abstasks/AbsTaskRetrieval.py

` 
corpus, queries, qrels = HFDataLoader(
                hf_repo=dataset_path,
                hf_repo_qrels=hf_repo_qrels,
                streaming=False,
                keep_in_memory=False,
                trust_remote_code=self.metadata_dict["dataset"].get(
                    "trust_remote_code", False
                ),
`
 we need to check whether the dataset_path is a local file or a hf dataset repo name and then pass dataset_path as a value for data_folder param for the former case and for hf_repo param for the latter case.

Please share your thoughts on this @Samoed.

Samoed: > Even if we were to override load_data, we will be instantiating a HFDataLoader object within the overriden load_data method.

I don't clearly understand. If you overload `load_data` then you can load dataset in any format as you want.

>  we need to check whether the dataset_path is a local file or a hf dataset repo name and then pass dataset_path as a value for data_folder param for the former case and for hf_repo param for the latter case.

This case would download from local files using datasets, which is fine I think. To download from local `jsonl` files, you need to pass `dataset_dir` which we don't pass here you can check current implementation https://github.com/embeddings-benchmark/mteb/blob/c34ea3b432393bc82fefddc45e294c645c1a64ca/mteb/abstasks/retrieval_dataset_loaders.py#L42, which will be released soon

Samoed: I see that you've added `dataset_dir` argument to the dataloader [in your fork](https://github.com/ashokrajab/mteb/blob/2eb462ce76e9d87a731f86c20050e8b9b87f7b12/mteb/abstasks/AbsTaskInstructionRetrieval.py#L352), but because we don't pass it in `main` we deprecate it and removed in `v2`

Issue #1920, opened 2025-02-01
Leaderboard: Link for text-embedding-005


Muennighoff:
Currently seems like it has no links - maybe we could link to the Google docs; https://huggingface.co/spaces/mteb/leaderboard

<img width="571" alt="Image" src="https://github.com/user-attachments/assets/c3782ea6-3030-4208-92a8-a681128c47c2" />

imadtyx:
Does this link work? https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-text-embeddings

Muennighoff:
Yes!

PR #1962, opened 2025-02-05
Fix: added link for text-embedding-005 and other google models #1920

imadtyx:


imadtyx:
#1920

Muennighoff:
LGTM! I think this would render it correctly in the LB (maybe cc @x-tabdeveloping)

x-tabdeveloping:
Yeah it looks good!

x-tabdeveloping:
@isaac-chung Any idea about the tests? Seems like some dataset is not available on HF.

isaac-chung:
> @isaac-chung Any idea about the tests? Seems like some dataset is not available on HF.

Just hit rerun since tests were passing in other python versions and [the dataset commit exists](https://huggingface.co/datasets/mteb/webis-touche2020-v3/commits/main).

x-tabdeveloping (approved):

Issue #1954, opened 2025-02-04
Leaderboard model size toggle bug


Muennighoff:
When moving the `Model Size (#M Parameters)` toggle to the left and back to the maximum on the right GritLM-8x7B will not reappear cuz the maximum defaults to 10B and it is >10B. So it seems like one cannot make it reappear without refreshing.

Relatedly, when just moving the toggle a tiny bit at the beginning the model won't disappear

x-tabdeveloping:
That does seem weird, I'm guessing it could be some sort of floating point error...
I will look into it.

x-tabdeveloping:
Okay so: Gradio is weird - if you set a custom value, it will send `None` to the callback, that's why it doesn't work with arbitrarily small numbers.
I don't how you're experiencing that issue with GritLM on your end. If I bump the size back up to max, then it shows up again.

Muennighoff:
Attached a quick screen recording; It seems like it is not just the max error but generally models disappear that shouldn't?
[tmp.mov.zip](https://github.com/user-attachments/files/18659984/tmp.mov.zip)

x-tabdeveloping:
This is incredibly odd. This error doesn't replicate locally at all, but I'm also experiencing it at the HF Space. My guess would be that the server gets overloaded, and the order of the callback gets messed up in the queue (which is again, not really an issue with our code). There are two things we can try: a) Adding a submit button to the slider, (it could help, but I'm not sure) b) moving the callback to the client end, this would require a lot of code though...

This could also be something with the double-ended slider component

x-tabdeveloping:
There might be something weird going on with when the callbacks get triggered, I'm taking a deeper look right now.

x-tabdeveloping:
I have submitted a potential fix here #1969 

This is not an optimal solution though. I think we should push the Gradio team to add a default double-ended slider component to their standard components.
Currently the situation is that `gradio_rangeslider` has known issues, which have been reported but not fixed and the library has not seen an update for the last 4 months.
I also tried [another custom component](https://github.com/argmaxinc/gradio-components), where you can't actually slide the slider and it hasn't been updated for the last 7 months.
Both of these are maintained by single individuals.

Considering that this is the right component to use, and that we seem not to be the only ones who need it, it would be nice if the Gradio team created and maintained a range slider.

@Muennighoff

KennethEnevoldsen:
It seems to work now (though it is a bit laggy). I will close this for now

[...]

PR #1969, opened 2025-02-05
fix: Changed callback for slider, accounted for None input

x-tabdeveloping:
@Muennighoff this hopefully fixes some of our issues with the slider.
I will say though, that this solution is suboptimal, and I would still keep the issue open.


x-tabdeveloping:
#1954

Samoed (approved):

Issue #1784, opened 2025-01-13
[v2] Refactor to Use Enums Instead of Literals


sam-hey:
## Description:
Currently, the codebase relies on literals to describe scoring functions and other constants (magic literals), which can make the code harder to understand and maintain. Using literals for values like scoring functions  instead of Enums makes the codebase less clear and reduces the benefits of modern IDE features such as autocompletion and error detection.

In line with the refactoring principle outlined in [Refactoring Guru's "Replace Magic Number with Symbolic Constant"](https://refactoring.guru/replace-magic-number-with-symbolic-constant), I propose refactoring the code to replace these magic literals with Enums. This change will enhance code clarity, facilitate easier development, and prevent potential errors caused by the incorrect use of raw literals.

For example, using literals like `cosine` or `dot` as scoring function identifiers makes the code harder to inspect and add new functions to, especially as the project grows. The same applies to other literals used in the project, such as `ISO_LANGUAGE_SCRIPT`.

VSCode does not provide autocompletion for Literal types by default when using the Python extension.
 <img width="455" alt="No Autocomplete" src="https://github.com/user-attachments/assets/038ec8b9-0efc-40cd-999a-2e111b0f6fe4" />
<img width="737" alt="image" src="https://github.com/user-attachments/assets/0b29f226-349a-4180-bac1-e35de1a56fd3" />

## Proposed Solution:
1. **Introduce an Enum** to represent all possible scoring functions.
2. **Refactor all models and methods** that currently use raw literals to instead use Enums (refer to [PR #1759](https://github.com/embeddings-benchmark/mteb/pull/1759)).

## Benefits:
- **Increased Readability:** Using Enums or symbolic constants will make it immediately clear what values are being used and their purpose.
- **Easier Maintenance:** Modifying or extending the codebase becomes simpler when Enums are used, as they centralize value management and eliminate the risk of incorrect usage of raw literals.
- **Better IDE Support:** IDEs will provide autocompletion and error checking when using Enums, leading to fewer bugs and faster development.
- **Consistent Usage:** With named constants or Enums, the same value will always be referred to by the same name, reducing the chance of errors caused by inconsistent usage of literals.

Please provide any feedback or thoughts on this proposed change.

Thx a lot!

Start of the Discussion in: https://github.com/embeddings-benchmark/mteb/pull/1759

KennethEnevoldsen:
[...]

PR #1975, opened 2025-02-05
feat: add enums for ModelMeta similarity_fn_name

sam-hey:
ref: https://github.com/embeddings-benchmark/mteb/issues/1784

ref: add enum `ScoringFunction` for `similarity_fn_name`

### Code Quality

### Documentation

### Testing


mteb/model_meta.py
```diff
@@ -101,19 +107,36 @@ class ModelMeta(BaseModel):
...
+        if type(value) is ScoringFunction or value is None:
+            return value
+        mapping = {
+            "cosine": ScoringFunction.COSINE,
+            "dot": ScoringFunction.DOT_PRODUCT,
+            "MaxSim": ScoringFunction.MAX_SIM,
+        }
+        if value in mapping:
+            return mapping[value]
+        raise ValueError(f"Invalid similarity function name: {value}")
```
Samoed, line 132: Does enums need additional validation?

Samoed:
You didn't change MaxSim functions

mteb/model_meta.py
sam-hey: This section is responsible for mapping the similarity function from `sentence_transformers` to enums when a model without a `ModelMeta` is provided.

mteb/model_meta.py
```diff
@@ -101,19 +107,36 @@ class ModelMeta(BaseModel):
...
     training_datasets: dict[str, list[str]] | None
     adapted_from: str | None = None
     superseded_by: str | None = None
     citation: str | None = None
 
+    @field_validator("similarity_fn_name", mode="before")
+    @classmethod
+    def validate_similarity_fn_name(cls, value):
+        """Converts the similarity function name to the corresponding enum value.
+        sentence_transformers uses Literal['cosine', 'dot', 'euclidean', 'manhattan'] for similarity_fn_name
```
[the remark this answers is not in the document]
sam-hey: Thx, I added a comment with extra information, clarifying that it is only PyLate that uses MaxSim.

mteb/model_meta.py
```diff
@@ -101,19 +107,37 @@ class ModelMeta(BaseModel):
...
     use_instructions: bool | None
     training_datasets: dict[str, list[str]] | None
     adapted_from: str | None = None
     superseded_by: str | None = None
     citation: str | None = None
 
+    @field_validator("similarity_fn_name", mode="before")
+    @classmethod
+    def validate_similarity_fn_name(cls, value):
+        """Converts the similarity function name to the corresponding enum value.
```
isaac-chung, line 120: Love that this is backwards compatible! 

Just wanted to point out that at the moment, the target branch is v2, which means in v2, this field actually accepts both string and enum. I'd originally expect v2 to only accept enum, and this specific method would only be applied to `main` with a warning, saying that this field type will be switched to enum. But I'm okay with this as well 👍

isaac-chung (approved):
Thanks for adding this!

KennethEnevoldsen:
Just a post-hoc review here: This looks good. I think it is a step in the right direction

Issue #1987, opened 2025-02-05
Leaderboard: FAQ md file on transition


Muennighoff:
Maybe having a quick FAQ section for people wondering what changed/where their scores went and put that at the top of the leaderboard?

Sth like

FAQ:

Why is a model from the old leaderboard no longer showing up?

Because of XXX; You can do YYY to bring it back.

Why did the scores of models change?

ZZZ

Where did the main MTEB leaderboard go?

Renamed to MTEB(eng, classic) etc etc

....

KennethEnevoldsen:
I think this issue is solved or represented in other issues so will close this one

PR #1990, opened 2025-02-06
fix: Added description and resolved bug in rangeslider

### LINKED ISSUE #1954: Leaderboard model size toggle bug

KennethEnevoldsen:
The leaderboard now looks like

Screenshot 2025-02-06 at 11 54 39

addresses #1987
fixes #1954


.gradio/certificate.pem
Samoed, line 1: I think this sould be deleted

x-tabdeveloping (approved):
Looks good! Please remove the certificate

PR #1993, opened 2025-02-06
Sort benchmarks and add temp. reference to old leaderboard

### LINKED ISSUE #1984: Leaderboard: Order benchmarks alphabetically

### LINKED ISSUE #1987: Leaderboard: FAQ md file on transition

KennethEnevoldsen:
fixes #1984
adresses #1987

@x-tabdeveloping just so that you are aware, but will just auto merge this.


Muennighoff:
Hmm do you not want actual FAQ items like in #1987 ?

I think it could be worth explaining that the new leaderboard only displays models with implementations which leads to many models missing. And asking people to open a PR if they want a model to reappear. And the other items I mentioned in the issue

Muennighoff:
Sorry had missed that you already added it; Looks great to me 🚀

PR #2004, opened 2025-02-07
Simplify

Muennighoff:
Also why can we not allow derivatives? I think we should so removed it for now


KennethEnevoldsen (approved):

KennethEnevoldsen:
> Also maybe we can allow derivatives (removed the line for now)

Agree that we should, currently, it would be just like another model (#1211). Let us remove it for now and see if someone submits a quantized model and take it from there.

Ideally, I would want something like:

```
all-MiniLM-L6-v2      | scores |
  + quantization         | scores |
  + emb. dim. 256     | scores |
```

Shouldn't be too hard to add with a bit of sorting and string manipulation (and the possibility to filter it out)

PR #2006, opened 2025-02-07
docs: update MTEB eng classic benchmark description 

Muennighoff:
I'm not sure the generalization argument makes a lot of sense because the other version uses the ~same tasks so any model tuned the first version is also tuned to that one I think? Maybe I am wrong 🤔

### DIFF
--- a/mteb/benchmarks/benchmarks.py
+++ b/mteb/benchmarks/benchmarks.py
@@ -213,9 +213,7 @@ MTEB_ENG_CLASSIC = Benchmark(
     ),
     description="""The original English benchmark by Muennighoff et al., (2023).
 This page is an adaptation of the [old MTEB leaderboard](https://huggingface.co/spaces/mteb/leaderboard_legacy).
-
-> We recommend that you use [MTEB(eng)](https://huggingface.co/spaces/mteb/leaderboard?benchmark_name=MTEB%28eng%29) instead,
-as many models have been tuned on MTEB(eng, classic) datasets, and MTEB(eng) might give a more accurate representation of models' generalization performance.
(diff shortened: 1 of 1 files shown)


mteb/benchmarks/benchmarks.py
```diff
@@ -213,9 +213,7 @@ def load_results(
     ),
     description="""The original English benchmark by Muennighoff et al., (2023).
 This page is an adaptation of the [old MTEB leaderboard](https://huggingface.co/spaces/mteb/leaderboard_legacy).
-
-> We recommend that you use [MTEB(eng)](https://huggingface.co/spaces/mteb/leaderboard?benchmark_name=MTEB%28eng%29) instead,
-as many models have been tuned on MTEB(eng, classic) datasets, and MTEB(eng) might give a more accurate representation of models' generalization performance.
+Also, see [MTEB(eng)](https://huggingface.co/spaces/mteb/leaderboard?benchmark_name=MTEB%28eng%29), which is a version that is cheaper to run with a high correlation to this one.
```
KennethEnevoldsen, line 216: I would probably recommend it (but I also made it so 🤷‍♂️), but I agree the argument is not quite outlined. How does this work instead:

```suggestion
We recommend that you use [MTEB(eng)](https://huggingface.co/spaces/mteb/leaderboard?benchmark_name=MTEB%28eng%29) instead as it uses updated versions of the task making it notably faster to run and resolves [a known bug](https://github.com/embeddings-benchmark/mteb/issues/1156) in existing tasks. This benchmark also removes datasets common for fine-tuning such as MSMARCO, which makes model performance scores more comparable. However, generally, both benchmarks provide similar estimates.
```

Muennighoff: Nice this reads really well!

Issue #1979, opened 2025-02-05
Leaderboard: voyage-law gone


Muennighoff:
Why is this model gone? 🤔 voyage contributed most of this benchmark I think

<img width="1433" alt="Image" src="https://github.com/user-attachments/assets/4d09b835-cfe4-433f-8d90-ed6ac3fa8278" />

<img width="1255" alt="Image" src="https://github.com/user-attachments/assets/e9529e6a-58d4-45db-bd56-cce1036a13ef" />

Samoed:
I can't find any of voyage models

Muennighoff:
Yeah me neither; even though we have their implementation?

Samoed:
Yes, we have implementations https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/voyage_models.py

KennethEnevoldsen:
APIs was accidentally filtered, but this was fixed in #2014

Issue #1976, opened 2025-02-05
Leaderboard: OpenAI results gone?


Muennighoff:
The results from running OpenAI text-embedding-3-large and other models seem gone in the new leaderboard?

Samoed:
@x-tabdeveloping

x-tabdeveloping:
There was a typo in the leaderboard code, hopefully it's resolved with Kenneth's PR

x-tabdeveloping:
Yup they are there now, sorry for the troubles

Issue #1946, opened 2025-02-03
[v2] Integrate cde models


Samoed:
https://github.com/embeddings-benchmark/mteb/issues/1851#issuecomment-2605077294

jxmorris12:
@Samoed @x-tabdeveloping @Muennighoff @orionw 

Hi everyone. Congrats on the new leaderboard!

I logged on to take a look and noticed that neither of the CDE models ([[1]](https://huggingface.co/jxm/cde-small-v1), [[2]](https://huggingface.co/jxm/cde-small-v2)) are on there, even though they were first for their model class on the old leaderboard. Is there something else that I need to do? At least the second one, I think, should be on there, since it really is the best small-size embedding model right now.

Samoed:
I see results of the model in [results dir](https://github.com/embeddings-benchmark/results/tree/main/results/jxm__cde-small-v2/external), but can't find the models on the leaderboard. Maybe some logic was changed with handling `external` dir in https://github.com/embeddings-benchmark/mteb/pull/1893.

KennethEnevoldsen:
#2014 might resolve this. I think it might be due to the missing implementation.

jxmorris12:
@Samoed  @KennethEnevoldsen 
I don't think all of the models on the leaderboard have an MTEB implementation. Is that really the case? If so, can we please get one up there, so the leaderboard is actually _correct_? I'm happy to help if a working MTEB implementation is truly necessary.

Samoed:
Previously, your models were listed on the leaderboard, but after PRs, they have disappeared. I will try to start integrating your models over the weekend.

x-tabdeveloping:
They are there again, though they are listed as non-zero-shot on the MTEB(classic) leaderboard, since they have been trained on a number of datasets in the benchmark.

Sorry for the inconvenience.

Samoed:
Currently, I am experiencing a minor issue where I am unable to use these models with revision specified. Opened issue on HF https://huggingface.co/jxm/cde-small-v2/discussions/9

KennethEnevoldsen:
Seems like this issue is mostly done? We still need to add these two, though:

- [jxm/cde-small-v3](https://huggingface.co/jxm/cde-small-v3)
- [jxm/cde-small-v3-zero](https://huggingface.co/jxm/cde-small-v3-zero)

Samoed:
I think we can close this for now and open different issue for `v3`

[...]

PR #2014, opened 2025-02-07
fix: Add SONAR metadata and resolve missing models

KennethEnevoldsen:
Add ModelMeta for SONAR it also ensure that an "Unknown" number of parameters is not filtered.

Should resolve:
#1979
#1976

This seems to have been caused by the restructering of calls on the leaderboard.

This however leads to the side-effect that selecting number of parameters can't remove "Unknown". A solution is to add a selection box.

@x-tabdeveloping I still can't get the SONAR model to show. I was hoping you had an idea for this?

### Code Quality

### Documentation

### Testing

### Adding a model checklist

   - [ ] ~~`mteb.get_model(model_name, revision)`~~ and
 - [ ] ~~I have tested the implementation works on a representative set of tasks.~~


mteb/models/overview.py
```diff
@@ -147,10 +149,22 @@ def get_model_metas(
     open_weights: bool | None = None,
     frameworks: Iterable[str] | None = None,
     n_parameters_range: tuple[int | None, int | None] = (None, None),
+    allow_no_n_parameters: bool = True,
```
x-tabdeveloping, line 152: I still don't like this solution. `n_parameters_range` already allows for not filtering models out, by setting both settings to `None`. This flag shouldn't be there in my opinion.

mteb/models/sonar_models.py
```diff
@@ -0,0 +1,234 @@
...
+
+sonar = ModelMeta(
+    loader=no_model_implementation_available,
+    name="facebook/SONAR",
+    languages=sonar_langs,
+    open_weights=True,
+    use_instructions=False,  # it does take a language code as input
+    revision="a551c586dcf4a49c8fd847de369412d556a7f2f2",
+    release_date="2021-05-21",
+    n_parameters=None,  # it is really multiple models so not sure how to calculate this
```
x-tabdeveloping, line 221: I checked and the old leaderboard doesn't have these numbers either. Perhaps we could do number of active parameters?

mteb/models/sonar_models.py
```diff
@@ -0,0 +1,234 @@
...
+    max_tokens=None,  # couldn't find this
```
x-tabdeveloping, line 222: As far as I can tell the [default is 512](https://github.com/facebookresearch/SONAR/blob/549d287466443bd8720f938047882630c1c5c3f7/sonar/models/sonar_text/builder.py#L139).

mteb/load_results/load_results.py
```diff
@@ -68,7 +68,7 @@ def _model_name_and_revision(
     model_meta = revision_path / "model_meta.json"
     model_path = revision_path.parent
     if not model_meta.exists() and fallback_to_path:
-        logger.warning(
+        logger.info(
```
x-tabdeveloping, line 71: I'm not entirely sure why this is justified? Doesn't this warrant a warning?

x-tabdeveloping (commented):
I would definitely not do that workaround with the model sizes. Added a couple of comments.

Samoed:
Do cde models appear in this version of leaderboard https://github.com/embeddings-benchmark/mteb/issues/1946#issuecomment-2649326118?

mteb/load_results/load_results.py
KennethEnevoldsen: I would say that if you raise a warning you expect the user to consider it. We raise >100 warning. I don't think that is reasonable here.

mteb/models/overview.py
KennethEnevoldsen: Sorry I see that I hadn't pushed by change here. I removed this yesterday (agree with you)

mteb/models/sonar_models.py
KennethEnevoldsen: Well it is really multiple models (one for each language). I think for now it is fine to leave it None until we have an implementation (then I would be happy to go for active parameters)

KennethEnevoldsen:
> Do cde models appear in this version of leaderboard https://github.com/embeddings-benchmark/mteb/issues/1946#issuecomment-2649326118?

No they do not actually. @x-tabdeveloping this might be because we try to load it but can't? (it gives an error) and then it is simply discarded?

x-tabdeveloping:
No, they should appear. I remember that CDE models were appearing a week or two ago, something got messed up in the process.

Samoed:
Maybe we should rename this pull request to `add SONAR metadata`, as we haven't yet implemented the actual model.

KennethEnevoldsen:
@x-tabdeveloping can I merge this? It woud be great to have the APIs on the leaderboard. We can then debug CDE + SOLAR

x-tabdeveloping:
Yes sure! I was just waiting for the tests to pass. I can also try having a look at CDE and SONAR later today. Ping me if you find anything

PR #2015, opened 2025-02-07
fix: Added script for generating and saving a local leaderboard

Pringled:
## Description

Added a script to create and save a leaderboard locally for a given benchmark as discussed in https://github.com/embeddings-benchmark/mteb/discussions/2011.

The script retrieves results, parses them to a Pandas DF, prints them, and (optionally) saves them to a CSV. It uses the same functionality used for generating the live leaderboard. This makes it much easier to generate results locally and check aggregate scores before submitting them to MTEB.

NOTE: I named the script `make_leaderboard` but open to other suggestions.

## Usage

The script can be used as follows:
`python scripts/make_leaderboard --benchmark "MTEB(eng)" --save-path "leaderboard_results"`

Optionally, a custom results repo can be used (default is https://github.com/embeddings-benchmark/results), and specific models can be requested, for example:
`python scripts/make_leaderboard --benchmark "MTEB(eng)" --results_repo "https://github.com/Pringled/mteb-results" --models "potion-base-8M" --save-path "results"`

### Code Quality

### DIFF
--- a/scripts/make_leaderboard.py
+++ b/scripts/make_leaderboard.py
@@ -52,6 +52,7 @@ def load_leaderboard(
     benchmark_results = load_results(
         results_repo=results_repo,
         only_main_score=True,
+        require_model_meta=False,
     )
 
     # Filter results for the selected benchmark
--- a/scripts/make_leaderboard.py
+++ b/scripts/make_leaderboard.py
@@ -50,7 +50,7 @@ def load_leaderboard(
 
     # Load all results from the specified repository
     benchmark_results = load_results(
-        results_repo=results_repo or "https://github.com/Pringled/mteb-results",
+        results_repo=results_repo,
         only_main_score=True,
     )
 
--- a/scripts/make_leaderboard.py
+++ b/scripts/make_leaderboard.py
@@ -32,7 +32,7 @@ def load_leaderboard(
         benchmark_name: Name of the benchmark (e.g., "MTEB(eng)").
         models: List of model names to include. Default is None (all models).
         results_repo: Path to results repo (local or remote). Default is None (default repo).
-        save_path: (Optional) Path to save the leaderboard as CSV.
+        save_path: Path to save the leaderboard as CSV.
 
     Returns:
         summary_df: Leaderboard summary table.
--- a/scripts/make_leaderboard.py
+++ b/scripts/make_leaderboard.py
@@ -1,0 +1,151 @@
+from __future__ import annotations
+
+import argparse
+import logging
+import os
+
+import pandas as pd
+
+import mteb
+from mteb.leaderboard.table import scores_to_tables
+from mteb.load_results import load_results
+
+logging.basicConfig(level=logging.INFO)
+
+logger = logging.getLogger(__name__)
+
+
+def get_available_benchmarks():
+    """Get all available benchmark names."""
+    return [b.name for b in mteb.get_benchmarks()]
+
+
+def load_leaderboard(
+    benchmark_name: str,
+    results_repo: str,
+    models: list[str] | None = None,
+    save_path: str | None = None,
+) -> tuple[pd.DataFrame, pd.DataFrame]:
+    """Load the MTEB leaderboard for a specific benchmark, results repo, and models.
+
+    Args:
+        benchmark_name: Name of the benchmark (e.g., "MTEB(eng)").
+        models: List of model names to include. Default is None (all models).
+        results_repo: Path to results repo (local or remote). Default is None (default repo).
+        save_path: (Optional) Path to save the leaderboard as CSV.
+
+    Returns:
+        summary_df: Leaderboard summary table.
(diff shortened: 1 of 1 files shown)


scripts/make_leaderboard.py
```diff
@@ -0,0 +1,151 @@
...
+    models: list[str] | None = None,
+    save_path: str | None = None,
+) -> tuple[pd.DataFrame, pd.DataFrame]:
+    """Load the MTEB leaderboard for a specific benchmark, results repo, and models.
+
+    Args:
+        benchmark_name: Name of the benchmark (e.g., "MTEB(eng)").
+        models: List of model names to include. Default is None (all models).
+        results_repo: Path to results repo (local or remote). Default is None (default repo).
+        save_path: (Optional) Path to save the leaderboard as CSV.
```
KennethEnevoldsen, line 35:
```suggestion
        save_path: Path to save the leaderboard as CSV.
```

scripts/make_leaderboard.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def get_available_benchmarks():
+    """Get all available benchmark names."""
+    return [b.name for b in mteb.get_benchmarks()]
+
+
+def load_leaderboard(
+    benchmark_name: str,
+    results_repo: str,
```
KennethEnevoldsen, line 25: should just default to cache?

scripts/make_leaderboard.py
```diff
@@ -0,0 +1,151 @@
...
+        summary_df.to_csv(summary_file, index=False)
+        per_task_df.to_csv(per_task_file, index=False)
+
+        logger.info(f"Leaderboard saved to {summary_file}")
+        logger.info(f"Per-task results saved to {per_task_file}")
+
+    return summary_df, per_task_df
+
+
+if __name__ == "__main__":
```
KennethEnevoldsen, line 105: it is totally fine that this is in scripts, but that also means that it will not be maintained. A solution would be to add it to the CLI along with a few tests.

I would be happy to accept this PR either way.

I also added @x-tabdeveloping as a reviewer as he is one of the main developers on the leaderboard.

x-tabdeveloping: I'd argue as it stands it's definitely better in `scripts`.
If we want to add this to the CLI, then I don't feel comfortable relying on implementation details of the leaderboard table, and would rather copy the entire table generation code. I wouldn't feel comfortable with leaderboard-specific code breaking other things.

scripts/make_leaderboard.py
```diff
@@ -0,0 +1,151 @@
...
+        summary_gr_df.value["data"], columns=summary_gr_df.value["headers"]
+    )
+    per_task_df = pd.DataFrame(
+        per_task_gr_df.value["data"], columns=per_task_gr_df.value["headers"]
+    )
+
+    # Filter models if specified
+    if models:
+        summary_df = summary_df[
+            summary_df["Model"].apply(lambda x: any(m in x for m in models))
```
x-tabdeveloping, line 79: I believe using
```python
summary_df = summary_df["Model"].isin(models)
```
would be a more readable and faster solution.

scripts/make_leaderboard.py
```diff
@@ -0,0 +1,151 @@
...
+    # Get the selected benchmark
+    benchmarks = mteb.get_benchmarks()
+    benchmark = next((b for b in benchmarks if b.name == benchmark_name), None)
+    if not benchmark:
+        raise ValueError(
+            f"Benchmark '{benchmark_name}' not found. Available: {get_available_benchmarks()}"
+        )
+
+    # Load all results from the specified repository
+    benchmark_results = load_results(
```
x-tabdeveloping, line 52: Perhaps also `require_model_meta=False`? Otherwise some results will not be loaded from the old leaderboard, and we won't have feature parity.

x-tabdeveloping (commented):
Added a couple of comments, looks okay otherwise.

scripts/make_leaderboard.py
Pringled: It's not the same behavior though; my approach uses substring matching while this one uses exact string matching. So in this case, you can pass "potion-base" and it will match "potion-base-8M", "potion-base-4M", etc, which I thought would be a bit nicer. I'm open to your suggestion though, just wanted to point out the difference, let me know which one you prefer.

scripts/make_leaderboard.py
Pringled: Updated

scripts/make_leaderboard.py
Pringled: Isn't the cache system dependent, and based on an environment variable? Or do you mean the MTEB results repo as a default here as well?

KennethEnevoldsen: Yep I was reffering to the MTEB cache path

PR #2024, opened 2025-02-10
Add versioning to MTEB benchmarks

### LINKED ISSUE #2001: Rename MTEB(eng) & MTEB(eng,classic)
I think we need a name that scales better with future versions e.g. if the new MTEB(eng) deprecates, it may not make sense to call it MTEB(eng,classic2)

Proposals:
(A)
- MTEB(eng,classic) -> MTEB(eng)
- MTEB(eng) -> MTEB(eng,v2) and then v3, v4 etc if we change it again

(B)
- MTEB(eng,classic) -> MTEB(eng,2022)
- MTEB(eng) -> MTEB(eng,2025) ; Don't really like this one cuz it makes it weird if others do not have a year like MTEB(rus)

let me know if any other ideas else I would suggest going with (A); wdyt?

KennethEnevoldsen:
- Following suggestion made in #2001 I added version to MTEB benchmarks
- changed the name of the MTEB(Chinese) to MTEB(cmn, v1). Though we could go for MTEB(Chinese, v1) assuming it is a group (also covering chinese other than mandarin)
- change the name of FaMTEB(fas, beta) to MTEB(fas)
- did a minor formatting of imports due to a circular import error
- moved the Benchmark object out of the file with the benchmarks
  - this is still >1000 lines so we could split it up to "external_benchmarks", "monolingual_benchmarks", "domain_specific_benchmarks", "multilingual_benchmarks"

Fixes #2001

### Code Quality

### Documentation

- [ ] ~~**Updated Documentation**: Add or update documentation to reflect the changes introduced in this PR.~~

### Testing

### DIFF
--- a/mteb/benchmarks/benchmarks.py
+++ b/mteb/benchmarks/benchmarks.py
@@ -1,15 +1,14 @@
 from __future__ import annotations
 
-from collections.abc import Sequence
-from dataclasses import dataclass
-from typing import Annotated
+from typing import TYPE_CHECKING, Annotated
 
 from pydantic import AnyUrl, BeforeValidator, TypeAdapter
 
-from mteb.abstasks.AbsTask import AbsTask
-from mteb.load_results.benchmark_results import BenchmarkResults
-from mteb.load_results.load_results import load_results
+from mteb.benchmarks.benchmark import Benchmark
 from mteb.overview import MTEBTasks, get_task, get_tasks
+
+if TYPE_CHECKING:
+    pass
 
 http_url_adapter = TypeAdapter(AnyUrl)
 UrlString = Annotated[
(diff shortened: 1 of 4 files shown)


mteb/benchmarks/benchmarks.py
```diff
@@ -1219,7 +1159,7 @@ def load_results(
 )
 
 C_MTEB = Benchmark(
-    name="MTEB(Chinese)",
+    name="MTEB(cmn, v1)",
```
Muennighoff, line 1162: Do we know all these tasks are Mandarin? If so then should we also rename the folders to `cmn`? 
[zho](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/tasks/Retrieval/zho)

Muennighoff (commented):
Nice! To avoid breaking backward compat would it make sense to temporarily allow the old names by just mapping to the new benchmark carrying that name and raising a warning that the old name will be deprecated soon?

mteb/benchmarks/benchmarks.py
Samoed: All tasks in metadata have `cmn` as thier language https://github.com/embeddings-benchmark/mteb/blob/1588b9ac31b7a84d1e34ef0325c0c10b0b0e8db5/mteb/tasks/Retrieval/zho/CMTEBRetrieval.py#L96C22-L96C25

Muennighoff: I guess we can also keep the folder as zho and then use it for tasks with other dialects too

x-tabdeveloping (commented):
I agree with Niklas, Also make sure this doesn't break the leaderboard (I think it probably will since the multilingual benchmark is referred to by name)

KennethEnevoldsen:
I have added changed for backward compatibility. I have also updated the leaderboard code to refer to the `MTEB_multlingual.name`. This should prevent inconsistencies.

mteb/benchmarks/get_benchmark.py
```diff
@@ -1,20 +1,68 @@
...
+}
+
 
 def get_benchmark(
     benchmark_name: str,
 ) -> Benchmark:
+    if benchmark_name in PREVIOUS_BENCHMARK_NAMES:
+        logger.warning(
+            f"Using the previous benchmark name '{benchmark_name}' is deprecated. Please use '{PREVIOUS_BENCHMARK_NAMES[benchmark_name]}' instead."
+        )
```
Samoed, line 65: I think now, exception would be raised.
```suggestion
        )
        benchmark_name = PREVIOUS_BENCHMARK_NAMES[benchmark_name]
```

Samoed (approved):

Issue #2012, opened 2025-02-07
CI: Seperate out tests that fail due to HTTP calls


KennethEnevoldsen:
Currently, the test suite sometimes fails due to an error outside the scope of mteb (e.g. on HF). This makes it hard to trust when the test fails. A solution to this would be to simply factor out those tests in a separate CI (such that if the main tests fail it is clear easy to trust the error)

@Samoed and @isaac-chung WDYT about this?

[thread](https://stackoverflow.com/questions/54898578/how-to-keep-unit-tests-and-integrations-tests-separate-in-pytest) on how to do this in pytest for reference

Samoed:
Yes, we can separete integration tests

sam-hey:
This might also be a solution: https://github.com/pytest-dev/pytest-rerunfailures

One argument in favor of this fix is that cases not covered otherwise might be missed.
```
pytest --reruns 5 --only-rerun AssertionError --only-rerun ValueError --reruns-delay 1
```

KennethEnevoldsen:
```
pytest --reruns 5 --only-rerun AssertionError --only-rerun ValueError --reruns-delay 1
```

Hmm, but this could also hide an error that happens by chance.

sam-hey:
I don't think that `AssertionError` and `ValueError` are merely placeholders for a `ConnectionError` or any other temporary issue. Since this error should disappear after a short time, rerunning the test with `--reruns-delay xxxx` should resolve the issue. However, if the problem persists, it might indicate a deeper issue that should be addressed by a developer as it will fail also after the reruns.

PR #2029, opened 2025-02-11
ci: Rerun tests that fail due to networking issues.

sam-hey:
#2012: Rerun test cases that frequently fail in the pipeline. After 3 unsuccessful reruns, the test cases will fail.

### Code Quality

### Documentation

### Testing

Closes: https://github.com/embeddings-benchmark/mteb/issues/2012
cc: @KennethEnevoldsen @Samoed @isaac-chung

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,11 @@ lint-check:
 
 test:
 	@echo "--- 🧪 Running tests ---"
-	pytest -n auto --durations=5 --reruns 3 --only-rerun requests.exceptions.ReadTimeout 
+	pytest -n auto 
 
 test-with-coverage:
 	@echo "--- 🧪 Running tests with coverage ---"
-	pytest -n auto --durations=5 --cov-report=term-missing --cov-config=pyproject.toml --cov=mteb --reruns 3 --only-rerun requests.exceptions.ReadTimeout 
+	pytest -n auto --cov-report=term-missing --cov-config=pyproject.toml --cov=mteb 
 
 pr:
 	@echo "--- 🚀 Running requirements for a PR ---"
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,11 @@ lint-check:
 
 test:
 	@echo "--- 🧪 Running tests ---"
-	pytest -n auto --durations=5
+	pytest -n auto --durations=5 --reruns 3 --only-rerun requests.exceptions.ReadTimeout 
 
 test-with-coverage:
 	@echo "--- 🧪 Running tests with coverage ---"
-	pytest -n auto --durations=5 --cov-report=term-missing --cov-config=pyproject.toml --cov=mteb
+	pytest -n auto --durations=5 --cov-report=term-missing --cov-config=pyproject.toml --cov=mteb --reruns 3 --only-rerun requests.exceptions.ReadTimeout 
 
 pr:
 	@echo "--- 🚀 Running requirements for a PR ---"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,5 +161,6 @@ addopts = """
             --only-rerun requests.exceptions.ReadTimeout 
             --only-rerun huggingface_hub.errors.HfHubHTTPError 
             --only-rerun huggingface_hub.errors.LocalEntryNotFoundError
+            --only-rerun FileNotFoundError
             --durations=5
         """
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,4 +156,4 @@ patch_types = ["fix", "perf"]
 
 
 [tool.pytest.ini_options]
-addopts = "--reruns 3 --only-rerun requests.exceptions.ReadTimeout --durations=5"
+addopts = "--reruns 3 --only-rerun requests.exceptions.ReadTimeout --only-rerun huggingface_hub.errors.HfHubHTTPError --durations=5"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,3 +153,7 @@ tag_format = "{version}"
 major_types = ["breaking"]
 minor_types = ["feat"]
 patch_types = ["fix", "perf"]
+
+
+[tool.pytest.ini_options]
+addopts = "--reruns 3 --only-rerun requests.exceptions.ReadTimeout --durations=5"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,4 +156,10 @@ patch_types = ["fix", "perf"]
 
 
 [tool.pytest.ini_options]
-addopts = "--reruns 3 --only-rerun requests.exceptions.ReadTimeout --only-rerun huggingface_hub.errors.HfHubHTTPError --durations=5"
+addopts = """
+            --reruns 3 
+            --only-rerun requests.exceptions.ReadTimeout 
+            --only-rerun huggingface_hub.errors.HfHubHTTPError 
+            --only-rerun huggingface_hub.errors.LocalEntryNotFoundError
+            --durations=5
+        """
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,4 +156,4 @@ patch_types = ["fix", "perf"]
 
 
 [tool.pytest.ini_options]
-addopts = "--reruns 3 --only-rerun requests.exceptions.ReadTimeout --only-rerun json.decoder.JSONDecodeError --durations=5"
+addopts = "--reruns 3 --only-rerun requests.exceptions.ReadTimeout --durations=5"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,7 @@ mteb = "mteb.cli:main"
 
 [project.optional-dependencies]
 dev = ["ruff==0.6.4", # locked so we don't get PRs which fail only due to a lint update
-"pytest", "pytest-xdist", "pytest-coverage"]
+"pytest", "pytest-xdist", "pytest-coverage", "pytest-rerunfailures"]
 codecarbon = ["codecarbon"]
 speedtask = ["GPUtil>=1.4.0", "psutil>=5.9.8"]
 peft = ["peft>=0.11.0"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,4 +156,4 @@ patch_types = ["fix", "perf"]
 
 
 [tool.pytest.ini_options]
-addopts = "--reruns 3 --only-rerun requests.exceptions.ReadTimeout --durations=5"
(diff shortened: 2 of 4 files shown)


sam-hey:
I observed a `requests.exceptions.ReadTimeout` and

**Failed test:**  
`tests/test_tasks/test_all_abstasks.py::test_dataset_availability`  
**Error:**  
`AssertionError: Datasets not available on Hugging Face.`

If there are any additional issues, let me know, and I'll add them to the PR.

Samoed:
There is some more integrational tests
- [test_benchmark_integration_with_sentencetransformers](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_benchmark/test_benchmark_integration_with_sentencetransformers.py)
- [test_benchmark_integration_with_datasets](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_benchmark/test_benchmark_integration_with_datasets.py)

Samoed:
It seems that we need to rerun tasks with `JSONDecodeError`, because sometimes results not loading correctly, like in last CI

sam-hey:
> There is some more integrational tests
> 
>     * [test_benchmark_integration_with_sentencetransformers](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_benchmark/test_benchmark_integration_with_sentencetransformers.py)
> 
>     * [test_benchmark_integration_with_datasets](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_benchmark/test_benchmark_integration_with_datasets.py)

@Samoed What expectations do you have for these tests? I’d like to cover them using the CLI call since there will be more than one test, ensuring that all errors are caught.

Samoed:
I thought similar like `pytest.mark.flaky`, but if this would cover with cli, then everything is fine

sam-hey:
Currently, I am handling the requests.exceptions.ReadTimeout error through the CLI call, as I observed this issue in the actions. If there are additional exceptions occurring, they will also need to be identified and added to the list for rerun

Makefile
```diff
@@ -20,11 +20,11 @@ lint-check:
 
 test:
 	@echo "--- 🧪 Running tests ---"
-	pytest -n auto --durations=5
+	pytest -n auto --durations=5 --reruns 3 --only-rerun requests.exceptions.ReadTimeout 
 
 test-with-coverage:
 	@echo "--- 🧪 Running tests with coverage ---"
-	pytest -n auto --durations=5 --cov-report=term-missing --cov-config=pyproject.toml --cov=mteb
+	pytest -n auto --durations=5 --cov-report=term-missing --cov-config=pyproject.toml --cov=mteb --reruns 3 --only-rerun requests.exceptions.ReadTimeout
```
KennethEnevoldsen, line 27:
```suggestion
	pytest -n auto --durations=5 --cov-report=term-missing --cov-config=pyproject.toml --cov=mteb
```
I don't think I would add it here

pyproject.toml
```diff
@@ -53,7 +53,7 @@ mteb = "mteb.cli:main"
 
 [project.optional-dependencies]
 dev = ["ruff==0.6.4", # locked so we don't get PRs which fail only due to a lint update
-"pytest", "pytest-xdist", "pytest-coverage"]
+"pytest", "pytest-xdist", "pytest-coverage", "pytest-rerunfailures"]
```
KennethEnevoldsen, line 56: add min version (we should probably do this for all of these)

tests/test_benchmark/test_benchmark.py
```diff
@@ -42,7 +42,7 @@ def test_mulitple_mteb_tasks(
 ):
     """Test that multiple tasks can be run"""
     eval = mteb.MTEB(tasks=tasks)
-    eval.run(model, output_folder=str(tmp_path), overwrite_results=True)
+    eval.run(model, output_folder=tmp_path.as_posix(), overwrite_results=True)
```
KennethEnevoldsen, line 45: Hmm this will break on windows? `str(path)` uses system native syntax.

tests/test_tasks/test_all_abstasks.py
```diff
@@ -84,6 +84,12 @@ async def check_datasets_are_available_on_hf(tasks):
         assert False, f"Datasets not available on Hugging Face:\n{pretty_print}"
 
 
+@pytest.mark.flaky(
```
KennethEnevoldsen, line 87: Any way why we simply can't use this only? I would def. prefer that

tests/test_benchmark/test_benchmark.py
Samoed: No, the winodows runner is working fine. These changes have already been added in v2.

KennethEnevoldsen: Alright would have suspect that `"c:/windows"` would break on window, but if it is fine I have note worries.

KennethEnevoldsen: Was looking if we could add a check for this but seems like there isn't one: https://github.com/astral-sh/ruff/issues/3675

tests/test_tasks/test_all_abstasks.py
sam-hey: I would advocate doing this only for specific cases; otherwise, we would need to add the marker to many functions, and everyone adding a new test would have to consider it as well. Instead, I’ll add it as an argument in `pyproject.toml`. This should make it cleaner and apply to all cases automatically.

isaac-chung:
I think there are enough eyes on this PR. I'll sit this one out.

tests/test_tasks/test_all_abstasks.py
KennethEnevoldsen: I think we only have 2 integration tests, the cli test, the reproducible workflow I believe. I would much rather add specific handling to these rather than add overall generic handling this also adds documentation to when it fails locally with e.g. HTTPConnectionErrror  that this is an expected error for this test. If we add it in pyproject.toml most people will not see it.

pyproject.toml
```diff
@@ -153,3 +153,7 @@ tag_format = "{version}"
 major_types = ["breaking"]
 minor_types = ["feat"]
 patch_types = ["fix", "perf"]
+
+
+[tool.pytest.ini_options]
+addopts = "--reruns 3 --only-rerun requests.exceptions.ReadTimeout --only-rerun json.decoder.JSONDecodeError --durations=5"
```
KennethEnevoldsen, line 159: I am not sure JSONDecodeError is actually a error that we want to retry.

tests/test_tasks/test_all_abstasks.py
sam-hey: I would argue that this is precisely the approach we should be pursuing. Since the issue appears to be related to networking on a part of the system that we do not control, I believe this is an error that new contributors should not need to account for when adding new tests. Additionally, long-time contributors might overlook this when reviewing the PR.

Therefore, I strongly recommend adding the relevant configurations to the pyproject.toml file. This would allow maintainers to easily document errors that affect all tests. Our goal should be to ensure that contributors, especially those who may not anticipate such issues, can contribute without unnecessary complexity.

Moreover, the test results indicate that the test was executed multiple times (the reruns) if it fails. While I do not expect the test to fail after this change, should it happen, we should consider increasing the timeouts to prevent future issues.

Samoed: We have some tests that download data from hf, or github. Obviously, [test_benchmark_integration_with_sentencetransformers](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_benchmark/test_benchmark_integration_with_sentencetransformers.py) and [test_benchmark_integration_with_datasets](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_benchmark/test_benchmark_integration_with_datasets.py). Then 
- [test_model_memory_usage](https://github.com/embeddings-benchmark/mteb/blob/3e3af6a7e81abf4022e797303f7c17c0e423aa1a/tests/test_models/test_model_meta.py#L17-L21)
- [test_mteb_rerank](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_tasks/test_mteb_rerank.py)
- [test_cli](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_cli.py)
- [test_encoder_interfaces](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_encoder_interfaces.py)
- [test_reproducible_workflow](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_reproducible_workflow.py)

I used the files instead of the test functions, and I think there are almost 10 (or maybe more) tests that involve a network connection. I don't think we should add a decorator to each of these tests and can use some global config (cli or pyproject)

sam-hey:
We should probably track this for a while and document the errors as they occur, as there seems to be a larger issue on the HF side, with different errors arising due to problems on their end.

tests/test_tasks/test_all_abstasks.py
KennethEnevoldsen: Let us go for the general solution then

pyproject.toml
```diff
@@ -153,3 +153,14 @@ tag_format = "{version}"
...
+
+
+[tool.pytest.ini_options]
+addopts = """
+            --reruns 3 
+            --only-rerun requests.exceptions.ReadTimeout 
+            --only-rerun huggingface_hub.errors.HfHubHTTPError 
+            --only-rerun huggingface_hub.errors.LocalEntryNotFoundError
+            --only-rerun FileNotFoundError
+            --durations=5
```
KennethEnevoldsen, line 165: It would be nice to have some documentation on these. Could we add a comment. E.g. not sure why FileNotFoundError is there?

E.g. thinking:
```suggestion
            --reruns 3 
            --only-rerun requests.exceptions.ReadTimeout 
            --only-rerun huggingface_hub.errors.HfHubHTTPError # HF is unavailable, e.g. seen here: https://github.com/embeddings-benchmark/mteb/actions/runs/13275350693/job/37093688544
            --only-rerun huggingface_hub.errors.LocalEntryNotFoundError
            --only-rerun FileNotFoundError
            --durations=5
```

KennethEnevoldsen (approved):
I think this is essentially ready to merge. Comments would be nice, but it is not blocking

pyproject.toml
sam-hey: https://github.com/embeddings-benchmark/mteb/actions/runs/13302915091/job/37147507251?pr=2029
`FileNotFoundError` it is also a error from HF Lib.

KennethEnevoldsen: perfect, great to have these when we go over the CI again next time

sam-hey:
This can be merged if someone has the capacity.

PR #2042, opened 2025-02-12
Leaderboard integration test

x-tabdeveloping:
Added a test to see if a set of models are present in the leaderboard table.

### DIFF
--- a/mteb/leaderboard/utils.py
+++ b/mteb/leaderboard/utils.py
@@ -1,0 +1,64 @@
+import json
+from pathlib import Path
+from typing import Literal
+
+import mteb
+
+# Model sizes in million parameters
+MIN_MODEL_SIZE, MAX_MODEL_SIZE = 0, 100_000
+
+
+def filter_models(
+    model_names: list[str],
+    task_select: list[str],
+    availability: bool | None,
+    compatibility: list[str],
+    instructions: bool | None,
+    model_size: tuple[int | None, int | None],
+    zero_shot_setting: Literal["hard", "soft", "off"],
+):
+    lower, upper = model_size
+    # Setting to None, when the user doesn't specify anything
+    if (lower == MIN_MODEL_SIZE) or (lower is None):
+        lower = None
+    else:
+        # Multiplying by millions
+        lower = lower * 1e6
+    if (upper == MAX_MODEL_SIZE) or (upper is None):
+        upper = None
+    else:
+        upper = upper * 1e6
+    model_metas = mteb.get_model_metas(
+        model_names=model_names,
+        open_weights=availability,
+        use_instructions=instructions,
+        frameworks=compatibility,
+        n_parameters_range=(lower, upper),
+    )
+    models_to_keep = set()
+    for model_meta in model_metas:
+        is_model_zero_shot = model_meta.is_zero_shot_on(task_select)
+        if is_model_zero_shot is None:
+            if zero_shot_setting == "hard":
+                continue
+        elif not is_model_zero_shot:
+            if zero_shot_setting != "off":
+                continue
+        models_to_keep.add(model_meta.name)
+    return list(models_to_keep)
+
+
+ALL_MODELS = {meta.name for meta in mteb.get_model_metas()}
+
+
+def load_results():
+    results_cache_path = Path(__file__).parent.joinpath("__cached_results.json")
+    if not results_cache_path.exists():
+        all_results = mteb.load_results(
+            only_main_score=True, require_model_meta=False, models=ALL_MODELS
+        ).filter_models()
+        all_results.to_disk(results_cache_path)
+        return all_results
+    else:
+        with results_cache_path.open() as cache_file:
+            return mteb.BenchmarkResults.from_validated(**json.load(cache_file))
--- a/tests/test_leaderboard.py
+++ b/tests/test_leaderboard.py
@@ -1,0 +1,57 @@
+from __future__ import annotations
+
+import pandas as pd
+import pytest
+
+from mteb.benchmarks import (MTEB_EN, MTEB_ENG_CLASSIC, Benchmark,
+                             MTEB_multilingual)
+from mteb.leaderboard.table import scores_to_tables
+from mteb.leaderboard.utils import filter_models, load_results
+
+# A couple of models to look for.
+SHOULD_BE_THERE = [
+    "NV-Embed-v2",
+    "e5-mistral-7b-instruct",
+    "text-embedding-005",
+    "voyage-large-2-instruct",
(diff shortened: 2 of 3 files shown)


x-tabdeveloping:
hohooo we have a problem. Gradio 5 requires Python 3.10 at minimum, so the leaderboard can't run on 3.9. What do we do? @KennethEnevoldsen @isaac-chung @Muennighoff @Samoed

KennethEnevoldsen:
I think this should solve it:
```
@pytest.mark.skipif(sys.version_info < (3,3), reason="requires python3.3")
def test():
  pass
```

Makefile
```diff
@@ -5,7 +5,7 @@ install:
 install-for-tests:
 	@echo "--- 🚀 Installing project dependencies for test ---"
 	@echo "This ensures that the project is not installed in editable mode"
-	pip install ".[dev,speedtask]"
+	pip install ".[dev,speedtask,leaderboard]"
```
KennethEnevoldsen, line 8: Ah this install will also break for v3.9

We could remove v3.9 from tests, but then I would drop support for 3.9 completely

mteb/leaderboard/utils.py
```diff
@@ -0,0 +1,64 @@
+import json
+from pathlib import Path
+from typing import Literal
+
+import mteb
+
+# Model sizes in million parameters
+MIN_MODEL_SIZE, MAX_MODEL_SIZE = 0, 100_000
```
KennethEnevoldsen, line 8: These seems like leaderboard configs. I would probably move to config.py (potentially together with main leaderboard etc.)

mteb/leaderboard/utils.py
```diff
@@ -0,0 +1,64 @@
...
+            if zero_shot_setting == "hard":
+                continue
+        elif not is_model_zero_shot:
+            if zero_shot_setting != "off":
+                continue
+        models_to_keep.add(model_meta.name)
+    return list(models_to_keep)
+
+
+ALL_MODELS = {meta.name for meta in mteb.get_model_metas()}
```
KennethEnevoldsen, line 51: move to top and probably also to config.py

mteb/leaderboard/utils.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def load_results():
```
KennethEnevoldsen, line 54:
```suggestion
def load_cached_results():
```
to differentiate it from `mteb.load_results`

mteb/leaderboard/utils.py
KennethEnevoldsen, line 1: I would probably split this up into three scripts. `config.py`, `filter_models.py` and `load_cached_results.py`.

tests/test_leaderboard.py
```diff
@@ -0,0 +1,56 @@
...
+    "sentence-t5-xxl",
+    "M2V_base_glove",
+]
+BENCHMARKS_TO_CHECK = [
+    MTEB_EN,
+    MTEB_ENG_CLASSIC,
+    MTEB_multilingual,
+]
+
+all_results = load_results()
```
KennethEnevoldsen, line 29: This should be moved to a fixture (otherwise it is loaded even on `pytest --collect-only`

isaac-chung:
Might be worth dropping 3.9 support all together: https://github.com/embeddings-benchmark/mteb/pull/1925#discussion_r1942650712 (separate thread)

tests/test_leaderboard.py
KennethEnevoldsen: We load just the target model and tasks to speed this up.

tests/test_leaderboard.py
```diff
@@ -0,0 +1,56 @@
...
+    "text-embedding-3-small",
+    "GritLM-7B",
+    "sentence-t5-xxl",
+    "M2V_base_glove",
+]
+BENCHMARKS_TO_CHECK = [
+    MTEB_EN,
+    MTEB_ENG_CLASSIC,
+    MTEB_multilingual,
+]
```
KennethEnevoldsen, line 27: I would cluster these in what they try to test:

```suggestion
SHOULD_BE_THERE = [
    # generally important
    "NV-Embed-v2",
    "e5-mistral-7b-instruct",   # e5
     "GritLM-7B",                       # gritlm
    "sentence-t5-xxl",
    "M2V_base_glove",             # minishlab
    # APIs
    "text-embedding-005",       # google
    "voyage-large-2-instruct",  # voyage
    "text-embedding-3-small", # openai
    # have previously been missing: https://github.com/embeddings-benchmark/mteb/issues/1946
    "cde-small-v2",
    "SONAR",
]
BENCHMARKS_TO_CHECK = [
    MTEB_EN,
    MTEB_ENG_CLASSIC,
    MTEB_multilingual,
]
```

Should we add some sentence transformers models as well?

tests/test_leaderboard.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@pytest.mark.parametrize("benchmark", BENCHMARKS_TO_CHECK)
+def test_leaderboard_tables(benchmark: Benchmark):
+    benchmark_results = benchmark.load_results(
+        base_results=all_results
+    ).join_revisions()
+    scores = benchmark_results.get_scores(format="long")
+    all_models = list({entry["model_name"] for entry in scores})
+    filtered_models = filter_models(
```
KennethEnevoldsen, line 39: Let us make a test for filter models and what models should be filtered in what cases.

tests/test_leaderboard.py
```diff
@@ -0,0 +1,56 @@
...
+    MTEB_ENG_CLASSIC,
+    MTEB_multilingual,
+]
+
+all_results = load_results()
+
+
+@pytest.mark.parametrize("benchmark", BENCHMARKS_TO_CHECK)
+def test_leaderboard_tables(benchmark: Benchmark):
+    benchmark_results = benchmark.load_results(
```
KennethEnevoldsen, line 34: If we just want to ensure a tests for functionality we could create  `tests/dummy_results_repo/{model_name}/*.json`.

This would probably speed this up quite a bit.

Makefile
Samoed: Maybe leaderboard integration test would skip python3.9?

tests/test_leaderboard.py
Samoed: Maybe change `have previously been missing` to `without implementation`?

Makefile
KennethEnevoldsen: Ahh but even trying to install the gradio version would break

KennethEnevoldsen:
Will close this as it seems to have gotten stale.

Issue #2033, opened 2025-02-11
Freeze columns for better leaderboard UI


Muennighoff:
https://github.com/gradio-app/gradio/issues/2407 got solved so we can freeze e.g. the rank & model columns

shikhar1729:
I'm working on this

Muennighoff:
Feel free to suggest what you think looks best via PR & we can iterate there!

shikhar1729:
Raised PR #2044 for the fix. @Muennighoff Please review.

PR #2044, opened 2025-02-13
fix: Freeze model/rank columns in leaderboard

shikhar1729:
This PR enhances the leaderboard UI by freezing `Rank` and `Model` columns on the `Summary` table and the `Model` column on the `Performance per task` table. With this change, users can scroll horizontally without losing track of model.

Addresses #2033 

### Checklist

https://github.com/user-attachments/assets/7a62b0fd-b581-43fb-8922-6a26b6cab4ee

### DIFF
--- a/mteb/leaderboard/table.py
+++ b/mteb/leaderboard/table.py
@@ -49,23 +49,6 @@ def split_on_capital(s: str) -> str:
     return " ".join(re.findall(r"[A-Z]?[a-z]+|[A-Z]+(?=[A-Z]|$)", s))
 
 
-def get_column_widths(df: pd.DataFrame) -> list[str]:
-    widths = []
-    for column_name in df.columns:
-        column_word_lengths = [len(word) for word in column_name.split()]
-        if is_numeric_dtype(df[column_name]):
-            value_lengths = [len(f"{value:.2f}") for value in df[column_name]]
-        else:
-            value_lengths = [len(str(value)) for value in df[column_name]]
-        try:
-            max_length = max(max(column_word_lengths), max(value_lengths))
-            n_pixels = 35 + (max_length * 12.5)
-            widths.append(f"{n_pixels}px")
-        except Exception:
-            widths.append("50px")
-    return widths
-
-
 def get_column_types(df: pd.DataFrame) -> list[str]:
     types = []
     for column_name in df.columns:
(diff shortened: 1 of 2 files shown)


Samoed:
I think you should update min gradio version in pyproject

x-tabdeveloping:
ahhh that looks cool and is actually pretty useful, thanks!

mteb/leaderboard/table.py
```diff
@@ -240,12 +219,9 @@ def scores_to_tables(
     return (
         gr.DataFrame(
             joint_table_style,
-            column_widths=column_widths,
```
x-tabdeveloping, line 243: Are you sure getting rid of those was the right way to deal with this? Now it seems to me that both the rank and model columns are too wide, and it's hard to see anything else because of them.

x-tabdeveloping: What happens if you leave the column widths in?

x-tabdeveloping (commented):
I left one question about the column widths, otherwise it looks good!

mteb/leaderboard/table.py
shikhar1729: The `column_widths` interfere with the `pinned_columns` attribute, causing only the first column to be pinned when column widths are applied.

Additionally, I explored a few alternative approaches:

- Keeping the existing columns and adding custom sticky CSS in `app.py` to render a new table, but the result wasn’t visually appealing.

- Passing the raw DataFrame (to check if `.style` was causing issues), but this didn’t resolve the problem either.

shikhar1729: Here you go : https://github.com/gradio-app/gradio/issues/10584

shikhar1729:
> I think you should update min gradio version in pyproject

Pushed commit

mteb/leaderboard/table.py
KennethEnevoldsen: @shikhar1729 will you add a new issue on this so that we can track when it is resolved within MTEB. 

Great work on this PR

shikhar1729: Here you go: https://github.com/embeddings-benchmark/mteb/issues/2060
thanks!

KennethEnevoldsen (approved):
I think this is good to merge @x-tabdeveloping what is your thoughts here?

PR #2045, opened 2025-02-13
misc: Add image classification descriptive stats implementation

isaac-chung:
Add image classification descriptive stats implementation
- Added MNIST as an example. Running on all tasks will be separate.
- Renamed TASK_TYPE to MMTEB_TASK_TYPE, and defined MIEB_TASK_TYPE to make it easier to check conditions

### Code Quality

### Documentation

### Testing

### DIFF
--- a/mteb/abstasks/Image/AbsTaskImageClassification.py
+++ b/mteb/abstasks/Image/AbsTaskImageClassification.py
@@ -16,10 +16,39 @@ from ...evaluation.evaluators import (
     ImagelogRegClassificationEvaluator,
 )
 from ..AbsTask import AbsTask, ScoresDict
+from ..TaskMetadata import DescriptiveStatistics
 
 ImageFile.LOAD_TRUNCATED_IMAGES = True
 
 logger = logging.getLogger(__name__)
+
+
+class ImageClassificationDescriptiveStatistics(DescriptiveStatistics):
+    """Descriptive statistics for ImageClassification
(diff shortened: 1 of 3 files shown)


Samoed (commented):
Can you add labels count? Also I think target branch should be `v2`

isaac-chung:
Thanks!

Label count = num samples here. In multilabel classification, we'll add label count there 👍

Btw do you have a preference on this?
> Would you prefer a PR per abstask? Or a PR for all abstask?

isaac-chung:
I'd like the target branch to remain as `main` if that's ok. There are existing descriptive stats there as well.

Samoed:
By "label count", I mean the number of times each label occurs, similar to how it is done in the current AbsTaskClassification.

https://github.com/embeddings-benchmark/mteb/blob/fddd898bbd555ed6d856bdcc4878670b01c1b0f8/mteb/abstasks/AbsTaskClassification.py#L281-L283

Example
https://github.com/embeddings-benchmark/mteb/blob/fddd898bbd555ed6d856bdcc4878670b01c1b0f8/mteb/descriptive_stats/Classification/SlovakHateSpeechClassification.json#L29-L37

I think we can do per task.

isaac-chung:
Oh then absolutely! Thanks for clarifying. Will add those here as well.

Samoed (approved):

PR #2057, opened 2025-02-13
misc: Add image clustering descriptive stats implementation

isaac-chung:
- Add image clustering descriptive stats implementation
- Run on all 5 image clustering tasks

### Code Quality

### Documentation

### Testing


mteb/abstasks/Image/AbsTaskImageClustering.py
```diff
@@ -10,10 +11,43 @@
 from ...encoder_interface import Encoder
 from ...evaluation.evaluators import ImageClusteringEvaluator
 from ..AbsTask import AbsTask, ScoresDict
+from ..TaskMetadata import DescriptiveStatistics
 
 logger = logging.getLogger(__name__)
 
 
+class ImageClusteringDescriptiveStatistics(DescriptiveStatistics):
```
Samoed, line 19: Can you also add? https://github.com/embeddings-benchmark/mteb/blob/68ff565004f39a6b91c1a3e0c9c1789b4d6ceec9/mteb/abstasks/AbsTaskClustering.py#L48-L50

isaac-chung: The paper only uses single label per sample, so these would just be 1. Maybe we can omit these?

Samoed: However, in the future, there may be tasks with an unbalanced distribution, and this could help to identify them

isaac-chung: I added that for multilabel classification though, so good shout 👍

isaac-chung: I'd prefer to align these to the existing tasks in the paper. If there are new tasks, these can be introduced then.

Samoed: Then can we add as you want to `main`, and add min/max to `v2`?

isaac-chung: Can do yeah, though I wouldn't put it as high priority. Plus if there will be multiple labels per sample, at that point other parts will need to be updated anyway, e.g. `unique_num_labels = len(set(labels))`

Issue #1974, opened 2025-02-05
MIEB classes with duplicate names


Samoed:
These classes have AbsTaskImageClassification and AbsTaskZeroshotClassification variants 
- BirdsnapClassification
- Caltech101Classification,
- Country211Classification,
- DTDClassification,
- EuroSATClassification,
- FER2013Classification,
- FGVCAircraftClassification,
- Food101Classification,
- GTSRBClassification,
- Imagenet1kClassification,
- MNISTClassification,
- OxfordPetsClassification,
- PatchCamelyonClassification,
- RESISC45Classification,
- StanfordCarsClassification,
- STL10Classification,
- SUN397Classification,
- UCF101Classification,

gowitheflow-1998:
in current version, they can run without problems in evaluation because of different metadata names (e.g., `name="Birdsnap"` and `name="BirdsnapZeroShot"`) but might worth renaming the classes?

isaac-chung:
Ya, maybe could add "ZeroShot" at the end of the ZS class names?

PR #2061, opened 2025-02-13
[MIEB] Rename classes with duplicated names

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/1974

### Code Quality


isaac-chung (approved):
Thanks!

I was worried that the renaming would affect importing, but it's okay. Import * is still used here, so I guess we could fix that next .

PR #2063, opened 2025-02-14
fix: generate metadata

sam-hey:
- fix script to create metadata 

### Code Quality

### Documentation

### Testing


scripts/generate_metadata.py
```diff
@@ -168,12 +168,18 @@ def get_max_token(model_name: str) -> int | None:
...
+
+
 def get_base_model(model_name: str) -> str | None:
     try:
         file_path = hf_hub_download(repo_id=model_name, filename="config.json")
         with open(file_path) as in_file:
             config = json.loads(in_file.read())
             base_model = config.get("_name_or_path", None)
+            if base_model in BASE_MODEL_ERRORS:
+                print(f"Base model error for {model_name} with base model {base_model}")
```
Samoed, line 181:
```suggestion
                logger.log(f"Base model error for {model_name} with base model {base_model}")
```

sam-hey: Switched to `logging` for the complete script from `print`.

sam-hey:
@Samoed Any idea about this error? For me, it seems to be an issue with the runners.
`memory_usage_mb` does not exist in this branch.
```
E           pydantic_core._pydantic_core.ValidationError: 1 validation error for ModelMeta
E           memory_usage_mb
E             Field required [type=missing, input_value={'name': 'intfloat/multil...ic_training_code': None}, input_type=dict]
E               For further information visit https://errors.pydantic.dev/2.10/v/missing
/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/scripts/generate_metadata.py:257: ValidationError
```

Samoed:
During CI for pull_request merges your branch with main and in the main this field is required

sam-hey:
Thanks a lot for the hint! It seems like I got confused with my upstreams...

pyproject.toml
```diff
@@ -53,7 +53,7 @@ mteb = "mteb.cli:main"
 
 [project.optional-dependencies]
 dev = ["ruff==0.6.4", # locked so we don't get PRs which fail only due to a lint update
-"pytest", "pytest-xdist", "pytest-coverage"]
+"pytest", "pytest-xdist", "pytest-coverage", "iso639"]
```
KennethEnevoldsen, line 56: I would add this to an optional dependency (+ add lower version boundary)

KennethEnevoldsen (commented):
I am unsure if we actually want to continually test stuff in scripts. Scripts are useful documentary, however if it is something that we continually want to re-use then I would probably add the functionality to the package itself.

E.g. something like:

```py
get_metadata_from_model_name(name: str) -> ModelMeta:
  ...
```

We already have something like this used in mteb.MTEB to estimate ModelMeta from an Encoder.

sam-hey:
> I am unsure if we actually want to continually test stuff in scripts. Scripts are useful documentary, however if it is something that we continually want to re-use then I would probably add the functionality to the package itself.
> 
> E.g. something like:
> 
> ```python
> get_metadata_from_model_name(name: str) -> ModelMeta:
>   ...
> ```
> 
> We already have something like this used in mteb.MTEB to estimate ModelMeta from an Encoder.

@KennethEnevoldsen 
I see three possible options going forward:

   1.  Remove all scripts, as non-functional scripts don’t provide value to anyone.
   2.  Migrate all scripts into the MTEB package.
   3.  Add testing to ensure all scripts continue to work.

I would argue that option 3 is the best approach, as including some of those scripts functionality in the pip package doesn’t make much sense. However, I’m open to other opinions.

KennethEnevoldsen:
I am not quite sure I agree with non-functional script not providing value. They are both documentary and acts as starting points is you want to do something similar. We could remove some of the MMTEB stuff to a seperate repository (though I would wait until after the release with that). E.g. think that some of the table creation was re-used for MIEB without me telling explicitly reporting it to anyone.

Some can be deleted. E.g.`run_mteb_bright.py`.

It is also a perfectly viable solution to just add tests for the ones that we plan to re-use. I don't however think we should support all scripts in the scripts folder (we have better things to do).

sam-hey:
OK, that sounds like a reasonable compromise to ensure tests are added to the scripts that will be used.

pyproject.toml
sam-hey: I would leave this as is to ensure that the tests run in the pipeline; otherwise, they must be skipped. Since it is just a development dependency, there should be no issues for the production use case.

KennethEnevoldsen, line 56: Yea but if it breaks it still documents a version of the dependency that likely worked.

```suggestion
"pytest", "pytest-xdist", "pytest-coverage", 
"iso639>=0.1.4" # used for tests/scripts/test_generate_model_meta.py
]
```

KennethEnevoldsen (approved):
I think we can probably merge this with the change

sam-hey:
This can be merged if someone has the capacity.

Issue #1953, opened 2025-02-04
Zero-shot default


Muennighoff:
I worry that the `Zero-shot` default of `Allow Unknown` in the leaderboard will incentivize people not to share/talk about their training data if it includes any benchmark data. I think maybe just doing `Allow all` but making the cross a bit clearer? Can we somehow make it red like ❌

x-tabdeveloping:
As it is now, we use a UTF-8 emoji, and there is no cross that renders red in Gradio :(

I have also tried a markdown field with emoji names between colons (`:cross:`) but it doesn't work in the table.

I agree with the incentivization part, but I would prefer not giving up on the default filter. I have heard some people have even stronger opinions about it. It's an easy fix though if we collectively decide that something else is a better move.

KennethEnevoldsen:
We could do❓: Not known and ⚠️: Not zero.-shot

x-tabdeveloping:
I'd prefer the cross, but we can change it if that's the vibe

Muennighoff:
Noting that the Open LLM leaderboard also allows contaminated models by default (https://huggingface.co/spaces/open-llm-leaderboard/open_llm_leaderboard#/); Maybe we do a community vote on this?

Muennighoff:
The leaderboard also looks much more alive with `Allow all` as it is way more models 🤔

KennethEnevoldsen:
> Maybe we do a community vote on this?

I believe we already had this discussion on Slack

> Noting that the Open LLM leaderboard also allows contaminated models by default 

Can't seem to find any contaminated models on their leaderboard, only:

<img width="269" alt="Image" src="https://github.com/user-attachments/assets/0535cdd9-b0b9-4e8d-8efa-f29763f4cf5c" />

---

We have also had specific complaints from model developers about known zero-shot models being on the leaderboard (reducing trust in the scores).

E.g. the NV-Embed-v2 scores aren't comparable to e5. The voyage-3-exp is a quite clear result of this (not recommended for use but is at the top). I would say the ranking should (to the extent possible) reflect something that we would recommend.

I think `Allow Unknown` makes a reasonable compromise between encouraging good practices, fair comparison and building trust, while not punishing APIs.

However, from this, it might be too harsh with the ⚠️ and instead using the ❓would be more appropriate.

Let me know what you think @Muennighoff

KennethEnevoldsen:
[...]

PR #2076, opened 2025-02-16
Integrate cde models

Samoed:
### Code Quality

### Documentation

### Testing

Closes https://github.com/embeddings-benchmark/mteb/issues/1946

Added `load_task_sample` to `Encoder` interface to load some part of dataset.

| Task | Leaderboard | PR |
| - | - | - |
| AmazonCounterfactualClassification |  |  |
| EmotionClassification |  |  |
| ToxicConversationsClassification | | 0.790088 |
| ArxivClusteringS2S |  |  |
| RedditClustering |  |  | 
| SprintDuplicateQuestions |  | 0.960375 |
| TwitterSemEval2015 |  | 0.624438 |
| SciDocsRR |  |  |
| AskUbuntuDupQuestions | |  |
| SCIDOCS | |  |
| SciFact |  | 0.70848 |
| STSBenchmark |  | 0.830939 |
| STS16 |  | 0.814253 |
| SummEval |  | 0.264545 |

### DIFF
--- a/mteb/abstasks/AbsTaskBitextMining.py
+++ b/mteb/abstasks/AbsTaskBitextMining.py
@@ -130,6 +130,13 @@ class AbsTaskBitextMining(AbsTask):
             pair_columns=pairs,  # type: ignore
             **kwargs,
         )
+        sentences = [row[col] for col in pairs[0] for row in data_split]
+
+        model.load_task_sample(
+            sentences,
+            task_name=self.metadata.name,
+            **encode_kwargs,
+        )
         metrics = evaluator(model, encode_kwargs=encode_kwargs)
         if parallel:
             for v in metrics.values():
--- a/mteb/encoder_interface.py
+++ b/mteb/encoder_interface.py
@@ -63,7 +63,13 @@ class Encoder(Protocol):
         """
         ...
 
-    def load_task_sample(self, sentences: Sequence[str], task_name: str, prompt_type: PromptType | None = None, **kwargs: Any) -> None:
+    def load_task_sample(
+        self,
+        sentences: Sequence[str],
+        task_name: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> None:
         """Load a sample of the task to be used for encoding. This is used for cde models.
 
         Args:
@@ -71,6 +77,7 @@ class Encoder(Protocol):
             task_name: The name of the task. Sentence-transformers uses this to
                 determine which prompt to use from a specified dictionary.
             prompt_type: The name type of prompt. (query or passage)
+            kwargs: Additional arguments to pass to the encoder.
         """
         ...
 
--- a/mteb/encoder_interface.py
+++ b/mteb/encoder_interface.py
@@ -63,6 +63,17 @@ class Encoder(Protocol):
         """
         ...
 
+    def load_task_sample(self, sentences: Sequence[str], task_name: str, prompt_type: PromptType | None = None, **kwargs: Any) -> None:
+        """Load a sample of the task to be used for encoding. This is used for cde models.
+
+        Args:
+            sentences: The sentences to encode.
+            task_name: The name of the task. Sentence-transformers uses this to
+                determine which prompt to use from a specified dictionary.
+            prompt_type: The name type of prompt. (query or passage)
+        """
+        ...
+
 
 class EncoderWithQueryInstructionFormatting(Protocol):
     """Optional protocol for encoders that support combining queries with instructions in a model-specific way. If not implemented, MTEB will use the default query instruction formatting ({query} {instruction})."""
--- a/mteb/models/cde_models.py
+++ b/mteb/models/cde_models.py
@@ -1,18 +1,18 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Sequence
 from functools import partial
 from math import ceil
-from typing import Any, Sequence
+from typing import Any
 
 import numpy as np
 import torch
 
-from mteb.models.sentence_transformer_wrapper import SentenceTransformerWrapper
 from mteb.encoder_interface import PromptType
-from mteb.model_meta import ModelMeta
 from mteb.model_meta import ModelMeta, ScoringFunction
 from mteb.models.bge_models import bge_full_data
+from mteb.models.sentence_transformer_wrapper import SentenceTransformerWrapper
 
 logger = logging.getLogger(__name__)
 
@@ -23,16 +23,14 @@ class CDEWrapper(SentenceTransformerWrapper):
     max_sentences: int = 1000
 
     def encode(
-            self,
-            sentences: Sequence[str],
-            *,
-            task_name: str,
-            prompt_type: PromptType | None = None,
-            **kwargs: Any,
+        self,
+        sentences: Sequence[str],
+        *,
+        task_name: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
     ) -> np.ndarray:
-        prompt_name = self.get_prompt_name(
-            self.model_prompts, task_name, prompt_type
-        )
+        prompt_name = self.get_prompt_name(self.model_prompts, task_name, prompt_type)
         if prompt_name:
             logger.info(
                 f"Using prompt_name={prompt_name} for task={task_name} prompt_type={prompt_type}"
@@ -56,16 +54,27 @@ class CDEWrapper(SentenceTransformerWrapper):
             embeddings = embeddings.cpu().detach().float().numpy()
         return embeddings
 
-    def load_task_sample(self, sentences: Sequence[str], task_name: str, prompt_type: PromptType | None = None, **kwargs: Any) -> None:
-        logger.info(f"Loading dataset embeddings for task {task_name}. Prompt type: {prompt_type}")
+    def load_task_sample(
+        self,
+        sentences: Sequence[str],
+        task_name: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> None:
+        logger.info(
+            f"Loading dataset embeddings for task {task_name}. Prompt type: {prompt_type}"
+        )
+        if isinstance(sentences[0], list):
+            sentences = [s for sentences_list in sentences for s in sentences_list]
         used_sentences = np.random.choice(
-            sentences, size=min(ceil(len(sentences) * self.sentence_fraction), self.max_sentences)
+            sentences,
+            size=min(ceil(len(sentences) * self.sentence_fraction), self.max_sentences),
         )
         self.dataset_embeddings = self.model.encode(
             used_sentences,
-            # prompt_name=self.get_prompt_name(
-            #     self.model_prompts, task_name, prompt_type
-            # ),
+            prompt_name=self.get_prompt_name(
+                self.model_prompts, task_name, prompt_type
+            ),
             convert_to_tensor=True,
             **kwargs,
         )
--- a/mteb/models/cde_models.py
+++ b/mteb/models/cde_models.py
@@ -1,19 +1,93 @@
 from __future__ import annotations
 
 import logging
+from functools import partial
+from math import ceil
+from typing import Any, Sequence
 
+import numpy as np
+import torch
+
+from mteb.models.sentence_transformer_wrapper import SentenceTransformerWrapper
+from mteb.encoder_interface import PromptType
+from mteb.model_meta import ModelMeta
 from mteb.model_meta import ModelMeta, ScoringFunction
 from mteb.models.bge_models import bge_full_data
 
 logger = logging.getLogger(__name__)
 
 
+class CDEWrapper(SentenceTransformerWrapper):
+    dataset_embeddings: torch.Tensor = None
+    sentence_fraction: float = 0.1
+    max_sentences: int = 1000
+
+    def encode(
+            self,
+            sentences: Sequence[str],
+            *,
+            task_name: str,
+            prompt_type: PromptType | None = None,
+            **kwargs: Any,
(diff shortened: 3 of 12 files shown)


jxmorris12:
Hi @Samoed - thanks so much for working on this :) I see the benchmark numbers are slightly lower; trying to figure out what's going on. here are some questions I'm trying to answer:
- what's the size of the minicorpus used to create dataset embeddings?
- where are the minicorpus docs coming from?
- are dataset embeddings computed using the first-stage model and the corpus prompt?
- are identical dataset embeddings used for corpus and query embeddings?
- are there any other obvious differences from the [official CDE eval script](https://github.com/jxmorris12/cde/blob/main/evaluate_mteb.py)?

jxmorris12:
Awesome. Thanks for checking and running the benchmarks. Yeah, the dataset embedding should always be computed with the documents' prefix, and the same dataset embeddings should be used for both queries & documents.

Samoed:
Here is updated scores
| Task                              | Leaderboard      | PR      |
|:-----------------------------------|:------------------|:------------------|
| AmazonCounterfactualClassification | 86.0149 | 0.848358 |
| AskUbuntuDupQuestions              | 61.7903           | 62.198 |
| EmotionClassification              | 63.235            | 61.38 |
| RedditClustering                   | 58.7946           | 57.4909 |
| SCIDOCS                            | 22.847            | 22.353 |
| STS16                              | 86.0136           |  84.2347 |
| STSBenchmark                       | 87.7139           | 85.7  |
| SciDocsRR                          | 84.3736 | 82.882  |
| SciFact                            | 73.476            | 70.725 |
| SprintDuplicateQuestions           | 97.2412           | 96.6403 |
| SummEval                           | 31.5967           | 33.202 |
| ToxicConversationsClassification   | 81.4014           |  80.5811  |
| TwitterSemEval2015                 | 72.0492           | 70.6891 |

Samoed:
The main difference for `SciDocsRR` and `SciFact` is that I've used full corpus for indexing instead of `positives`. And for classification, I've used sentences after resampling, which can cause the difference. @jxmorris12 What do you think?

jxmorris12:
I think it's hard to say where the performance degradation is coming from but these results seem good enough to me.

Samoed:
I think you didn't use filter for zero-shot
image

jxmorris12:
Ok you're right – I see it. Thanks!
Oh, so the default is to allow "unknown training data", which probably includes many models that train on the benchmark datasets but don't report it, but it blocks open models which train on one or more of the MTEB datasets. That's odd too.

Muennighoff:
Yeah I agree and am also worried about the default (https://github.com/embeddings-benchmark/mteb/issues/1953); I think changing it to Allow All is much clearer

tomaarsen:
I agree that the default of "Allow Unknown" has some pretty negative effects:
1. Much increased focus on poorer and unknown models where we simply don't know much - but they were most likely trained on at least one of the datasets. For example, in the top 15 on MMTEB, only 1 model (!) is marked as zero-shot and all 14 others are Unknown:
image

2. Extremely competent models are seemingly gone from the leaderboard because they overlap with e.g. 1% of the test datasets. This will result in people being less informed!

In my opinion, the emoji immediately draws attention, so if we update the default to "Allow all", then people will still be informed to be wary of certain models. See for example here, the Zero-shot column draws the eye pretty quickly:
image

Beyond that, the "Zero-shot" option is among all kinds of model options that don't need updating much, so it's very easy to miss.

In short: I think that users will be more than sufficiently informed about the zero-shot-ness of a model even if we use "Allow all" as the default, and they will be more informed than if we keep using "Allow unknown".

- Tom Aarsen

Samoed:
@KennethEnevoldsen @isaac-chung Can you review, please?

jxmorris12:
@Samoed I'm pretty sure something is still wrong. I filtered models for below 1000M params and CDE doesn't show up for either MTEB "v1" or "v2" (also what's the difference?).
By the way, the slider seems pretty biased against larger models, since you can't select any number below 1000M (1 billion) params.

Samoed:
On what benchmark are you watching?

Samoed:
@KennethEnevoldsen @isaac-chung Can you review, please?

jxmorris12:
> On what benchmark are you watching?

As far as I can tell – they don't show up anywhere. The most conventional workflow is MTEB>eng{v1, v2}. Then if you want you can decrease the number of parameters. Here's a screenshot with the max params set at 1B. CDE isn't there: 
Screenshot 2025-02-24 at 12 08 13 PM


My two cents: it feels this benchmark launch was quite premature; there are many issues, and no parity with the original leaderboard, which a lot of people relied on. Also again CDE is the best small model and doesn't display, which really sends the wrong message.

Samoed:
Yes, I can't find too. I've added an issue about that your model is missing. You can continue commenting there https://github.com/embeddings-benchmark/mteb/issues/2147

Samoed:
@jxmorris12 Can you solve this issue https://huggingface.co/jxm/cde-small-v2/discussions/9?

mteb/abstasks/AbsTaskBitextMining.py
```diff
@@ -130,6 +130,13 @@ def _evaluate_subset(
             pair_columns=pairs,  # type: ignore
             **kwargs,
         )
+        sentences = [row[col] for col in pairs[0] for row in data_split]
+
+        model.load_task_sample(
+            sentences,
+            task_name=self.metadata.name,
+            **encode_kwargs,
+        )
```
KennethEnevoldsen, line 139: Won't this break with other models?

KennethEnevoldsen: Couldn't this be moved to the model's encode function? No need to make changes in the abstasks

mteb/encoder_interface.py
```diff
@@ -63,6 +63,24 @@ def encode(
...
+        """Load a sample of the task to be used for encoding. This is used for cde models.
+
+        Args:
+            sentences: The sentences to encode.
+            task_name: The name of the task. Sentence-transformers uses this to
+                determine which prompt to use from a specified dictionary.
+            prompt_type: The name type of prompt. (query or passage)
+            kwargs: Additional arguments to pass to the encoder.
+        """
+        ...
```
KennethEnevoldsen, line 82: I don't think this is needed to do what you want, so to keep the implementation simple and to avoid breaking backward compatibility I would prefer if we didn't do this

KennethEnevoldsen (changes requested):
(sorry for the late review on this - we had some work to do on the leaderboard). This looks quite reasonable, but I would say that the `load_task_sample` can be avoided

mteb/models/cde_models.py
```diff
@@ -1,20 +1,119 @@
...
+    PromptType.query.value: "search_query: ",
+    PromptType.passage.value: "search_document: ",
+}
+
 cde_small_v1 = ModelMeta(
-    loader=no_model_implementation_available,
+    loader=partial(
+        CDEWrapper,
+        model="jxm/cde-small-v1",
+        # revision="7017cdcb2abeccc8e9abd1c2379eb0e05121eec8",
```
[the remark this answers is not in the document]
Samoed: I can't download model with revision, because it's downloading tokenizer from another repository https://huggingface.co/jxm/cde-small-v2/discussions/9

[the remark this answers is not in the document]
KennethEnevoldsen: Ahh will you link the issue after the comment

mteb/models/cde_models.py
```diff
@@ -1,20 +1,119 @@
...
+    def __init__(self, model: str, *args, **kwargs: Any) -> None:
+        super().__init__(model, *args, **kwargs)
+        model_config = AutoConfig.from_pretrained(model, trust_remote_code=True)
+        self.max_sentences = model_config.transductive_corpus_size
+
+    def encode(
+        self,
+        sentences: Sequence[str],
+        *,
+        task_name: str,
```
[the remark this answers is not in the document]
Samoed: Yes, that's good solution. I will change to that

Samoed:
I've modified the `encode` function slightly by adding `task_sample_data` for Retrieval (Instruct, Reranking, etc.), (Multilabel) Classification, and Summarization to ensure the correct data is passed to `encode`. I made this change because it would be difficult to determine which data to include in the context, as there can be multiple splits or subsets.

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -274,6 +274,7 @@ def __call__(
         X_train = model.encode(
             self.sentences_train,
             task_name=self.task_name,
+            task_sample_data=self.sentences_train,
```
KennethEnevoldsen, line 277: This still seems like a thing that is not needed and only customized to one specific model. I would prefer it if we didn't do it.

KennethEnevoldsen (changes requested):

KennethEnevoldsen:
> I didn't use the approach with load data, because there is no method to deterimine current split/subset

Hmm yeah, we have seen this problem before when we have been missing the language during the encode step (https://github.com/embeddings-benchmark/mteb/issues/215). A solution would be to simply supply the (split, hf_subset) (would give you both of these).

Train/test/dev shouldn't really influence it I would say, so leaving out splits might be an option, but I believe some dataset use splits to denote e.g. domain (I might be wrong here)

Samoed:
We can update the encode function to pass spits, subset and `TaskMetadata` instead of task name

KennethEnevoldsen:
> We can update the encode function to pass spits, subset and TaskMetadata instead of task name

I wouldn't remove task name or add task metadata as we can simply:

```py
task_meta = mteb.get_task(task_name).metadata
```

This also keeps the interface backward compatibility (unless someone doesn't handle arbitrary kwargs)

jxmorris12:
Hi @Samoed -- are you planning to get back to this? Is there any way I can help? I want to upload the scores for my new v3 model on the new benchmark too, and this seemed like the easiest way.

Samoed:
Hi! Yes, I'll return to it next week

Samoed:
Restuls for `cde-small-v2`. @jxmorris12 I've updated implementation
| Task                              | Leaderboard      | PR      |
|:-----------------------------------|------------------:|------------------:|
| AmazonCounterfactualClassification | 86.0149 | 84.5672 |
| AskUbuntuDupQuestions              | 61.7903           | 61.525 |
| EmotionClassification              | 63.235            | 61.375 |
| SCIDOCS                            | 22.847            | 22.459 |
| STS16                              | 86.0136           |  83.2319 |
| STSBenchmark                       | 87.7139           | 84.828  |
| SciDocsRR                          | 84.3736 | 82.828  |
| SciFact                            | 73.476            | 70.753 |
| SprintDuplicateQuestions           | 97.2412           | 96.5982 |
| SummEval                           | 31.5967           | 33.4948 |
| ToxicConversationsClassification   | 81.4014           |  80.5811  |
| TwitterSemEval2015                 | 72.0492           | 70.6891 |

jxmorris12:
These results seem ok for v2 to me. I'll compare our implementations and fix any discrepancy when I release v3 in the next couple weeks.

Samoed:
@jxmorris12 I've opened PR to fix issue with tokenizer loading https://huggingface.co/jxm/cde-small-v2/discussions/12

Samoed:
Also opened PR for `cde-small-v1` https://huggingface.co/jxm/cde-small-v1/discussions/8

KennethEnevoldsen:
@Samoed, sorry for not responding here. Feel free to ping me next time.

mteb/models/cde_models.py
```diff
@@ -1,19 +1,188 @@
...
+        "InstructionRetrieval",
+        "InstructionReranking",
+    )
+
+    def __init__(self, model: str, *args, **kwargs: Any) -> None:
+        super().__init__(model, *args, **kwargs)
+        model_config = AutoConfig.from_pretrained(model, trust_remote_code=True)
+        self.max_sentences = model_config.transductive_corpus_size
+
+    def create_embeddings_key(
```
KennethEnevoldsen:
```suggestion
    def _create_embeddings_key(
```

mteb/models/cde_models.py
```diff
@@ -1,19 +1,188 @@
...
+            **kwargs,
+        )
+        if isinstance(embeddings, torch.Tensor):
+            # sometimes in kwargs can be return_tensors=True
+            embeddings = embeddings.cpu().detach().float().numpy()
+
+        self.prev_embeddings_key = self.create_embeddings_key(task_metadata, hf_subset)
+        return embeddings
+
+    def load_task_sample(
```
KennethEnevoldsen: I would love some documentation on this since it is fairly non-standard.

I would also like it to be possible to run it like so:

```
task = mteb.get_task(...)
mdl = ...
mdl.create_corpus_embedding(task)
```

Though it would be great if `mdl.create_corpus_embedding(task)` is not required.

Samoed: > I would also like it to be possible to run it like so:
>
>```
>task = mteb.get_task(...)
>mdl = ...
>mdl.create_corpus_embedding(task)
>```

I don't think that we want to support passing `task` to it since this function is used only for this model and tasks would need to store different embeddings for different subsets and this will be hard and won't work with multiple tasks

KennethEnevoldsen: Alright, then maybe worth making the method private?

Samoed: Yes, I will change it then

KennethEnevoldsen:
@Samoed should we just merge this?

Samoed:
If it looks good for you now, then yes

jxmorris12:
Amazing!!!

Issue #2079, opened 2025-02-16
Integrate mmE5


Samoed:
https://huggingface.co/intfloat/mmE5-mllama-11b-instruct

Samoed:
Created issue to make repo installable https://github.com/haon-chen/mmE5/issues/1

Samoed:
@Aradhye2002 They've updated demo script without usage of functions from their repo https://github.com/haon-chen/mmE5/blob/eacae6b576ba379ae48baa9384fc07ebad547720/demo.py

Samoed:
I've created PR for proper integration with sentence transformers https://huggingface.co/intfloat/mmE5-mllama-11b-instruct/discussions/2

Aradhye2002:
> [@Aradhye2002](https://github.com/Aradhye2002) They've updated demo script without usage of functions from their repo https://github.com/haon-chen/mmE5/blob/eacae6b576ba379ae48baa9384fc07ebad547720/demo.py

Noted.

PR #2080, opened 2025-02-17
fix: Add back task filtering by modalities

### LINKED ISSUE #2075: Add back Task filtering by modality supported by models
We can add a check to see if the model supports running the task. For example, if the task has `text` and `image` modalities and the model has `text`, then the task should be skipped.

isaac-chung:
Fixes #2075 
- Added back task filtering by modalities. 
- Added unit test to cover cases of running models with different modalities as tasks
 
When running `mteb run -m average_word_embeddings_komninos -t MNIST`, we now get (instead of an exception / error):

```
## Evaluating 1 tasks:
────────────────────────────────────────────────────────────────────────────── Selected tasks  ───────────────────────────────────────────────────────────────────────────────
ImageClassification
    - MNIST, i2i


INFO:mteb.evaluation.MTEB:

********************** Evaluating MNIST **********************
INFO:mteb.evaluation.MTEB:average_word_embeddings_komninos only supports ['text'], but the task modalities are ['image'].
```

### Code Quality

### Documentation

### Testing

### DIFF
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -497,11 +497,14 @@ class MTEB:
                 del self.tasks[0]  # empty memory
                 continue
 
-            # skip evaluation if the model does not support the task modalities.
-            task_modalities = "".join(sorted(task.metadata.modalities))
-            if "".join(sorted(meta.modalities)) != task_modalities:
+            # NOTE: skip evaluation if the model does not support all of the task's modalities.
+            # If the model covers more than the task's modalities, evaluation will still be run.
+            sorted_task_modalities = sorted(task.metadata.modalities)
+            if meta.modalities is not None and any(
+                m not in meta.modalities for m in sorted_task_modalities
+            ):
                 logger.info(
-                    f"{meta.name} only supports {meta.modalities}, but the task modalities are {task.metadata.modalities}."
+                    f"{meta.name} only supports {meta.modalities}, but the task modalities are {sorted_task_modalities}."
                 )
                 del self.tasks[0]  # empty memory
                 continue
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -497,6 +497,15 @@ class MTEB:
                 del self.tasks[0]  # empty memory
                 continue
 
+            # skip evaluation if the model does not support the task modalities.
+            task_modalities = "".join(sorted(task.metadata.modalities))
+            if "".join(sorted(meta.modalities)) != task_modalities:
+                logger.info(
+                    f"{meta.name} only supports {meta.modalities}, but the task modalities are {task.metadata.modalities}."
+                )
+                del self.tasks[0]  # empty memory
+                continue
+
             task_eval_splits = (
                 eval_splits if eval_splits is not None else task.eval_splits
             )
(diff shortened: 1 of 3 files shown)


mteb/evaluation/MTEB.py
```diff
@@ -497,6 +497,15 @@ def run(
                 del self.tasks[0]  # empty memory
                 continue
 
+            # skip evaluation if the model does not support the task modalities.
+            task_modalities = "".join(sorted(task.metadata.modalities))
+            if "".join(sorted(meta.modalities)) != task_modalities:
```
Samoed, line 502: If model support image and text, I think we should run it
```suggestion
            if any([m not in task_modalities for m in meta.modalities])::
```

isaac-chung: Good call! Will add that.

KennethEnevoldsen (approved):
This looks great, not a lot to add here

edit: Seems like tests needs a fix though

tests/test_benchmark/test_benchmark.py
```diff
@@ -330,3 +334,27 @@ def encode(self, sentences, prompt_name: str | None = None, *args, **kwargs):
...
         overwrite_results=True,
     )
+
+
+# NOTE: Covers image and image-text tasks. Can be extended to cover new mixed-modality task types.
+@pytest.mark.parametrize(
+    "task", [MockImageTextPairClassificationTask(), MockRetrievalTask()]
+)
+@patch.object(logger, "info")
+def test_task_modality_filtering(mock_logger, task):
```
Samoed, line 344: Can you please add a test for the model with `modalities=["image", "text"]` on text tasks?

isaac-chung: For image only tasks, it makes sense. I can add that.

But for text only tasks, conceptually this can be done, but I don't think it should be. Often such models like CLIP have considerably smaller context limits (77 tokens) compared to the common 512 in ST models. Also, current CLIP model implementations do not have `encode` method, so this will throw an exception and I think we can omit this the "text" case.

Samoed: Yes, but if there are cases like 'gme', which is a pure encoder and can handle images, it should be able to perform text-related tasks.

isaac-chung: Hmm I see. We can create a MockGMEModel just for this case, as this test is not generalizable at this time. The majority of the multimodal models are not like GME. Would that work?

Samoed: Yes. I think it would better to call something like `MockTextImageModel`

isaac-chung: MockTextImageModel would be misleading as CLIP also support text and image. I'd prefer to go with `MockGMEModel` for now.

isaac-chung: Actually, if it's just one case, I'd prefer not to add it at this point. Let's circle back when this is more prominent.

Samoed: Jasper and mme5 (https://github.com/embeddings-benchmark/mteb/issues/2079) have can encode images too, but for now they're not integrated in mieb properly

isaac-chung: hmm yeah good find. Let's open an issue for this so it's not hidden in a PR? It seems like it's more than adding test cases and about properly integrating these models in mieb + adding `encode` methods to some existing models.

Issue #2005, opened 2025-02-07
Clicking model names does not open new window


Muennighoff:
https://github.com/user-attachments/assets/2b996686-3315-434d-b527-a3517099dc84

KennethEnevoldsen:
Hmm, odd. Seems like the specification of the data frame is quite similar:

[legacy](https://huggingface.co/spaces/mteb/leaderboard_legacy/blob/8c2a406d6df6644d40d6fb1ebd089fd3b80cceda/app.py#L343):
```
dataframe = gr.Dataframe(item["data"], datatype=datatype, type="pandas", height=500)
```

new:
```
        gr.DataFrame(
            joint_table_style,
            column_widths=column_widths,
            datatype=column_types,
            interactive=False,
            wrap=True,
        ),
        gr.DataFrame(
            per_task_style, column_widths=task_column_widths, interactive=False
        ),
```

The interactive=False is probably the culprit (@x-tabdeveloping I know there was a reason, but can't remember why) - setting it to True solves the issue.

x-tabdeveloping:
This is a bug with Gradio. I have also experienced it a couple of times, luckily you can right click and open in new tab, but it's still pretty annoying.
The reason the table is set not to be interactive is because:
a) It would allow users to add rows and columns, which is non-sensical
b) Gradio doesn't support styling on tables, where interactive=True. Don't ask why

KennethEnevoldsen:
> a) It would allow users to add rows and columns, which is non-sensical

Example:

<img width="458" alt="Image" src="https://github.com/user-attachments/assets/8eaee579-c929-4df1-bd61-b28d249f28f8" />


> b) Gradio doesn't support styling on tables, where interactive=True. Don't ask why

Example:
<img width="1437" alt="Image" src="https://github.com/user-attachments/assets/b391ea4e-2aaf-4eaa-a2ff-d5a86d26e79c" />

@x-tabdeveloping is there a gradio issue on this (then we should link it) otherwise let us create it so that we can keep track on what we can and can't solve.

x-tabdeveloping:
Uhm, I opened an issue a while back, I don't think there has been any activity on it for a while.

x-tabdeveloping:
Lol, they marked it as `pending clarification` and told me that it's intended behaviour: https://github.com/gradio-app/gradio/issues/9938

x-tabdeveloping:
I have not opened one on the link thing though, we could do that

[...]

Issue #1972, opened 2025-02-05
Reduce built time for the leaderboard


KennethEnevoldsen:
So I have at least a few times gone to the leaderboard and seen that it was rebuilding. 

This takes quite a while, but once it is done the leaderboard is reasonably fast. 

There are two solutions to this problem:
- [ ] limit rebuilds (23 people can now rebuild and someone might rebuild the space if they see some oddities not knowing how long it takes)
- [ ] reduce built time, there are multiple ways to do this
  - **speed up result loading** (`mteb.load_results()`)
    - 1) reduce the results that are loaded by deleting unused files. E.g. if model 1 is run with `ArxivClustering` both with and without a revision, we might choose to keep only the one with a revision
    - 2) another solution is the speed up the loading of the results.
  - **Avoid recomputing the cache**: 
    - Once the cache is created loading the data is quite quick. However, the cache is recomputed on rebuilt. We could have the cache updated daily instead.

@x-tabdeveloping has been the main person working on this. @x-tabdeveloping feel free to edit this issue if there are solutions you would like to add or remove.

Muennighoff:
What is happening when it shows this? I've never seen this on the old leaderboard 🤔

<img width="566" alt="Image" src="https://github.com/user-attachments/assets/fbc600e2-2edf-4151-8a8d-3ba3a56a6eaf" />

Muennighoff:
This is when it is stuck for a very long time for me

Muennighoff:
Also sometimes this load time is extremely slow; this is just when opening it; I think we need some form of caching

<img width="1421" alt="Image" src="https://github.com/user-attachments/assets/2ad94c02-30eb-449b-8713-3db3c0a09e42" />

KennethEnevoldsen:
> This is when it is stuck for a very long time for me

It could be that someone rebuilt the space (though I agree that it is not quite clear).

> Also sometimes this load time is extremely slow; this is just when opening it

This is really wierd to me. I don't have any issues loading the space (@x-tabdeveloping have you experience >80s load times on the tables?)

x-tabdeveloping:
My guess would be that the long loading times on the tables happen because of long queues on the server.
Since we only have one relatively small server to run everything, and all callbacks are executed on the server side (since we didn't write any Javascript), this is kind of expected. I'm not sure if we can optimize this away, since I have already tried to get callback runtime down to an absolute minimum.

[...]

PR #2085, opened 2025-02-17
fix: Smarter leaderboard caching with cachetools

x-tabdeveloping:
I've improved the leaderboard caching by preventing as many taxing calculations as humanly possible.
There's a slight performance improvement on my machine, but I'd imagine this makes a bigger difference in production.
Additionally, all callbacks are run with all benchmarks on startup, so that most things would already be in the cache when the leaderboard is running.

Unfortunately I have no way of completely preventing callbacks and communication with the server, so it's not as significant an improvement as previous ones.

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -9,13 +9,13 @@ from pathlib import Path
 from typing import Literal
 from urllib.parse import urlencode
 
+import cachetools
 import gradio as gr
 import pandas as pd
 from gradio_rangeslider import RangeSlider
 
 import mteb
 from mteb.benchmarks.benchmarks import MTEB_multilingual
-from mteb.caching import json_cache
 from mteb.leaderboard.figures import performance_size_plot, radar_chart
 from mteb.leaderboard.table import scores_to_tables
 
@@ -458,7 +458,10 @@ see existing implementations [here](https://github.com/embeddings-benchmark/mteb
     # This sets the benchmark from the URL query parameters
     demo.load(set_benchmark_on_load, inputs=[], outputs=[benchmark_select])
 
-    @json_cache
+    @cachetools.cached(
+        cache={},
+        key=lambda benchmark_name: hash(benchmark_name),
+    )
     def on_benchmark_select(benchmark_name):
         start_time = time.time()
         benchmark = mteb.get_benchmark(benchmark_name)
@@ -483,7 +486,7 @@ see existing implementations [here](https://github.com/embeddings-benchmark/mteb
             languages,
             domains,
             types,
-            [task.metadata.name for task in benchmark.tasks],
+            sorted([task.metadata.name for task in benchmark.tasks]),
             scores,
         )
 
@@ -493,7 +496,12 @@ see existing implementations [here](https://github.com/embeddings-benchmark/mteb
         outputs=[lang_select, domain_select, type_select, task_select, scores],
     )
 
-    @json_cache
+    @cachetools.cached(
+        cache={},
+        key=lambda benchmark_name, languages: hash(
+            (hash(benchmark_name), hash(tuple(languages)))
+        ),
+    )
     def update_scores_on_lang_change(benchmark_name, languages):
         start_time = time.time()
         benchmark_results = all_benchmark_results[benchmark_name]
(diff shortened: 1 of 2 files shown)


x-tabdeveloping:
#1972

x-tabdeveloping:
Note that this does not reduce startup time, in fact it increases it a bit. It only makes the application faster once it's running.

pyproject.toml
```diff
@@ -58,7 +59,7 @@ dev = ["ruff==0.6.4", # locked so we don't get PRs which fail only due to a lint
 codecarbon = ["codecarbon"]
 speedtask = ["GPUtil>=1.4.0", "psutil>=5.9.8"]
 peft = ["peft>=0.11.0"]
-leaderboard = ["gradio>=5.7.1", "gradio_rangeslider>=0.0.8", "plotly>=5.24.0,<6.0.0"]
+leaderboard = ["gradio>=5.5.0,<6.0.0", "gradio_rangeslider>=0.0.8", "plotly>=5.24.0,<6.0.0"]
```
KennethEnevoldsen, line 62: I believe `>=5.7.1` was required for frozen, no?

mteb/leaderboard/app.py
```diff
@@ -458,7 +458,10 @@ def filter_models(
     # This sets the benchmark from the URL query parameters
     demo.load(set_benchmark_on_load, inputs=[], outputs=[benchmark_select])
 
-    @json_cache
+    @cachetools.cached(
```
KennethEnevoldsen, line 461: I not sure why the [standard implementation](https://docs.python.org/3/library/functools.html) of cache wouldn't work here

KennethEnevoldsen (commented):
This seems like it would make it faster. It is quite hard to test in practice though. Any chance we can push this to the Demo leaderboard to allow us to compare?

mteb/leaderboard/app.py
```diff
@@ -655,49 +697,93 @@ def update_models(
...
+        ),
+    )
     def update_tables(
         scores,
         search_query: str,
         tasks,
         models_to_keep,
+        benchmark_name: str,
     ):
+        print("Search query: ", str(search_query))
```
KennethEnevoldsen, line 719:
```suggestion
```

pyproject.toml
Samoed: It wasn't merged yet

KennethEnevoldsen: Yea waiting for @x-tabdeveloping. 

Actually I would set lower bound to 5.12.0 as this also solved #2005

mteb/leaderboard/app.py
x-tabdeveloping: In this specific callback it would, but I intentionally wanted to keep it explicit to match the other ones.
In case the callback gets extended, people will have to specify how to cache the additional parameters.

In most other callbacks you either need to turn lists into tuples or use IDs instead of hashes.

x-tabdeveloping:
@KennethEnevoldsen I don't think the difference would show, since not many are using the demo space. We'll only see how well it works in practice once a bunch of people are using it.

pyproject.toml
```diff
@@ -58,7 +59,7 @@ dev = ["ruff==0.6.4", # locked so we don't get PRs which fail only due to a lint
 codecarbon = ["codecarbon"]
 speedtask = ["GPUtil>=1.4.0", "psutil>=5.9.8"]
 peft = ["peft>=0.11.0"]
-leaderboard = ["gradio>=5.7.1", "gradio_rangeslider>=0.0.8", "plotly>=5.24.0,<6.0.0"]
+leaderboard = ["gradio>=5.7.1,<6.0.0", "gradio_rangeslider>=0.0.8", "plotly>=5.24.0,<6.0.0"]
```
KennethEnevoldsen, line 62:
```suggestion
leaderboard = ["gradio>=5.12.0,<6.0.0", "gradio_rangeslider>=0.0.8", "plotly>=5.24.0,<6.0.0"]
```
You don't want to enforce this to resolve the link clicking?

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
     "eval_type_backport>=0.0.0",
     "polars>=0.20.22",
     "torchvision>0.0.0",
+    "cachetools>=5.2.0",
```
KennethEnevoldsen, line 44: Move to leaderboard dependencies

pyproject.toml
x-tabdeveloping: Yeye, I'm doing that rn. Just testing to see if all works.

pyproject.toml
x-tabdeveloping: will do

KennethEnevoldsen (approved):

KennethEnevoldsen:
All good on my end

PR #2097, opened 2025-02-18
[v2] reupload reranking datasets in old format

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/1878

### Code Quality

- [ ] **Code Formatted**: Format the code using `make lint` to maintain consistent style.

### Documentation

- [ ] **Updated Documentation**: Add or update documentation to reflect the changes introduced in this PR.

### Testing

- [ ] **New Tests Added**: Write tests to cover new functionality. Validate with `make test-with-coverage`.
- [ ] **Tests Passed**: Run tests locally using `make test` or `make test-with-coverage` to ensure no existing functionality is broken.

### Adding datasets checklist

**Reason for dataset addition**: ... 

- [ ] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb -m {model_name} -t {task_name}` command.
  - [ ] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [ ] `intfloat/multilingual-e5-small`
- [ ] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [ ] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [ ] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [ ] Run tests locally to make sure nothing is broken using `make test`. 
- [ ] Run the formatter to format the code using `make lint`. 

### Adding a model checklist

 - [ ] I have filled out the ModelMeta object to the extent possible
 - [ ] I have ensured that my model can be loaded using
   - [ ] `mteb.get_model(model_name, revision)` and
   - [ ] `mteb.get_model_meta(model_name, revision)`
 - [ ] I have tested the implementation works on a representative set of tasks.

PR #2116, opened 2025-02-20
WIP: Merge main into v2

KennethEnevoldsen:
- There were some changes to taskmetadata's  `descriptive_stat_path`, I am not entirely sure if the merge is correct here.
- why the "InstructionReranking"? (probably missed the reason somewhere?)
- @sam-hey they couldn't get the iso library to work. Seems it could be `python-iso639`? (seems like the test passed because try except loop). 
- @Samoed I tried to calculate desc stats, in v2, but got this error, which I believe you might have fixed (my merge could have undone it)

```
from mteb.tasks.Reranking.eng import BuiltBenchReranking

task = BuiltBenchReranking()
task.load_data() # fails (but only on v2) as "default" is not available?
```

for now I just calculated the desc. stats over on main and moved it over.

~~Note: we are likely missing some task imports in v2 (it is impossible to know with the star imports) - any good way to check?~~ I used the now outdated category to check.

It might be nice to start doing these merges more often?

### Code Quality

### Documentation

### Testing


Samoed:
Since there no more duplicate class names you can run [`scripts/generate_imports.py`](https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/scripts/generate_imports.py)

Samoed:
> why the "InstructionReranking"? (probably missed the reason somewhere?)

Instruction reranking were added in https://github.com/embeddings-benchmark/mteb/pull/1359 after unifing retrieval and reranking

> There were some changes to taskmetadata's descriptive_stat_path, I am not entirely sure if the merge is correct here.

Yes, that's correct 

> I tried to calculate desc stats, in v2, but get this quite error, which I believe you might have fixed (a merge could have undone it)

I think you've fixed it for now.

sam-hey:
Just had a look at the original pr - back then it was `iso639`, I think `python-iso639` is also fine to use. Not sure when the dependency was removed.

Samoed:
To load BuiltBenchReranking correctly you need to put it in https://github.com/embeddings-benchmark/mteb/blob/056c09ad254826f7afaa5899f2eec08b0cb12232/mteb/abstasks/AbsTaskReranking.py#L13-L30

I'm removing this in https://github.com/embeddings-benchmark/mteb/pull/2097

KennethEnevoldsen:
Failures seems to be HF failures. Will try to rerun to see if it resolves it

Samoed:
I think we should modify `test_dataset_availability` to test datasets independently. Currently, if one test fails, all datasets are rechecked due to `pytest.mark.flaky`, which results in around 4,000 requests after retries. This could lead to Hugging Face blocking our requests.

KennethEnevoldsen:
I am working on refactoring that test

Samoed:
I think we can merge as is and you will create separete PR for test refactoring

PR #2153, opened 2025-02-25
Added zero-shot percentages and different filtering scheme

x-tabdeveloping:
This PR includes a proposed solution to the issues we've been having with our previous way of filtering for zero-shot models.
Zero-shot-ness is now of a gradual nature and shows up in the main table as a percentage along with accompanying colours.
I have also changed the zero-shot filtering scheme to better fit the new system.
Now all models will be visible by default, but one can choose to hide the ones that have been trained on any of the datasets, or ones, whose makers did not disclose their training data.
image
image

### DIFF
--- a/mteb/model_meta.py
+++ b/mteb/model_meta.py
@@ -176,9 +176,7 @@ class ModelMeta(BaseModel):
             benchmark_datasets = set(tasks)
         else:
             tasks = cast(Sequence[AbsTask], tasks)
-            benchmark_datasets = set()
-            for task in tasks:
-                benchmark_datasets.add(task.metadata.name)
+            benchmark_datasets = {task.metadata.name for task in tasks}
         overlap = model_datasets & benchmark_datasets
         perc_overlap = 100 * (len(overlap) / len(benchmark_datasets))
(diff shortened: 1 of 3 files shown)


mteb/model_meta.py
```diff
@@ -165,6 +165,24 @@ def is_zero_shot_on(self, tasks: Sequence[AbsTask] | Sequence[str]) -> bool | No
...
+        if (self.training_datasets is None) or (not tasks):
+            return None
+        model_datasets = {ds_name for ds_name, splits in self.training_datasets.items()}
+        if isinstance(tasks[0], str):
+            benchmark_datasets = set(tasks)
+        else:
+            tasks = cast(Sequence[AbsTask], tasks)
+            benchmark_datasets = set()
+            for task in tasks:
+                benchmark_datasets.add(task.metadata.name)
```
Samoed, line 181:
```suggestion
            benchmark_datasets = {task.metadata.name for task in tasks}
```

Samoed (approved):
Great changes!

KennethEnevoldsen (approved):
Looks great

Issue #2167, opened 2025-02-26
Add training data annotations for Bilingual Embeddings


x-tabdeveloping:
According to their [HF page](https://huggingface.co/Lajavaness/bilingual-embedding-large) this is their finetuning procedure:

#### Stage 1: NLI Training
Dataset: [(SNLI+XNLI) for english+french]
Method: Training using Multi-Negative Ranking Loss. This stage focused on improving the model's ability to discern and rank nuanced differences in sentence semantics.
#### Stage 3: Continued Fine-tuning for Semantic Textual Similarity on STS Benchmark
Dataset: [STSB-fr and en]
Method: Fine-tuning specifically for the semantic textual similarity benchmark using Siamese BERT-Networks configured with the 'sentence-transformers' library.
#### Stage 4: Advanced Augmentation Fine-tuning
Dataset: STSB with generate [silver sample from gold sample](https://www.sbert.net/examples/training/data_augmentation/README.html)
Method: Employed an advanced strategy using [Augmented SBERT](https://arxiv.org/abs/2010.08240) with Pair Sampling Strategies, integrating both Cross-Encoder and Bi-Encoder models. This stage further refined the embeddings by enriching the training data dynamically, enhancing the model's robustness and accuracy.

KennethEnevoldsen:
So what is this - something like?

```
{
"STSBenchmark": ["train"],
"STSBenchmarkMultilingualSTS": ["train"], 
"XNLI": ["train"]
# not in mteb
# SNLI
}
```

x-tabdeveloping:
Yeah, I think that looks reasonable

PR #2173, opened 2025-02-26
Add Training data annotations

KennethEnevoldsen:
fixes #2168
fixes #2164 
fixes #2166 
fixes #2162

Removed fiqa pl and arxivclusteringS2s.v2 which both does not exist.

PR #2178, opened 2025-02-27
fix: Add more training data annotations

KennethEnevoldsen:
fixes https://github.com/embeddings-benchmark/mteb/issues/2167

and adds a few missing fixes from #2173


Samoed (approved):
Great!

PR #2182, opened 2025-02-27
add labse annotation

Samoed:
They don't have much information in paper, so I don't know what to add

### Code Quality

### DIFF
--- a/mteb/models/sentence_transformers_models.py
+++ b/mteb/models/sentence_transformers_models.py
@@ -215,7 +215,13 @@ labse = ModelMeta(
     use_instructions=False,
     superseded_by=None,
     adapted_from=None,
-    training_datasets=None,
+    training_datasets={
+        # CommonCrawl
+        # wiki  05-21-2020 dump
+        "WikipediaRetrievalMultilingual": [],
+        "WikipediaRerankingMultilingual": [],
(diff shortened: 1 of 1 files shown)


x-tabdeveloping:
Yea, it's a bit cryptic honestly.
Though I would guess that it is more or less zero-shot, since LaBSE predates MTEB.

mteb/models/sentence_transformers_models.py
```diff
@@ -215,7 +215,13 @@
     use_instructions=False,
     superseded_by=None,
     adapted_from=None,
-    training_datasets=None,
+    training_datasets={
+        # CommonCrawl
+        # wiki  05-21-2020 dump
+        "WikipediaRetrievalMultilingual": [],
+        "WikipediaRerankingMultilingual": [],
```
KennethEnevoldsen, line 222: Since these are partly synthetic data I would not add them here (they are also not constructed the same way). We have generally not added it unless it is the same dataset construction (but I agree that that boundary is somewhat arbitrary).

KennethEnevoldsen:
```suggestion
```

KennethEnevoldsen (approved):

PR #2194, opened 2025-02-28
Dev: improve local dev env

sam-hey:
The goal of this PR is to enhance the local development environment and minimize resource consumption by ensuring the pipeline runs only on linted code.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,20 @@
+fail_fast: true
+
+repos:
+  - repo: https://github.com/abravalheri/validate-pyproject
+    rev: v0.23
+    hooks:
+      - id: validate-pyproject
+
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v3.1.0
+    hooks:
+      - id: prettier
+        types_or: [yaml, json5]
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.9.6
+    hooks:
+      - id: ruff-format
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Create table
         run: |
           make build-docs
-      
+
       - name: Push table
         run: |
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
@@ -60,4 +60,3 @@ jobs:
             git commit -m "Update tasks table"
             git push
           fi
-          
--- a/.github/workflows/leaderboard_refresh.yaml
+++ b/.github/workflows/leaderboard_refresh.yaml
@@ -2,8 +2,8 @@ name: Daily Space Rebuild
 on:
   schedule:
     # Runs at midnight Pacific Time (8 AM UTC)
-    - cron: '0 8 * * *'
-  workflow_dispatch:  # Allows manual triggering
+    - cron: "0 8 * * *"
+  workflow_dispatch: # Allows manual triggering
 
 jobs:
   rebuild:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,4 +25,3 @@ jobs:
         id: lint
         run: |
           make lint-check
-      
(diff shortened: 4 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,20 @@
+fail_fast: true
+
+repos:
+  - repo: https://github.com/abravalheri/validate-pyproject
+    rev: v0.23
+    hooks:
+      - id: validate-pyproject
+
+  - repo: https://github.com/pre-commit/mirrors-prettier
```
Samoed, line 9: Maybe it would be easier to use default pre-commit formatter for yaml and json?

sam-hey: I believe that the defaults do not provide any automatic linting and only check for syntax. Therefore, even if archived, I believe that using prettier is the best option.

Samoed: I tried it now, and it worked. Can you change something like
```yaml
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v2.3.0
    hooks:
    -   id: check-yaml
    -   id: check-json
    -   id: end-of-file-fixer
    -   id: trailing-whitespace
    -   id: pretty-format-json
    -   id: check-toml
```

sam-hey: @Samoed are you ok with the changes?

Samoed: Yes! Thank, you. Let's wait for feedback from others

KennethEnevoldsen:
So, is the addition only for the pre-commit hooks? (hard to see with the large diff)

Personally, I am fine with adding an _optional_ pre-commit hooks. However, I do not want to force people to use it, so if we add it, it should be compatible with `make lint`.

sam-hey:
> So, is the addition only for the pre-commit hooks? (hard to see with the large diff)

Yes, it is. 
> Personally, I am fine with adding an optional pre-commit hooks. However, I do not want to force people to use it, so if we add it, it should be compatible with make lint.

The changes are fully optional—if somebody prefers not to use them, they can still install the packages without using the pre-commit hooks. The Ruff linter is used with all configs applied, so no. Calling make lint does not make too much sense to me, as this would not allow to fail if changes are applied (I would personally prefer this as it makes the Person aware of the changes that have been applied). It would be possible to call the local Ruff installation directly, eliminating the need for `chor`. However, this would prevent us from using the Ruff repository that provides additional configurations.

```yaml

  - repo: local
    hooks:
      - id: ruff
        name: ruff
        description: "Run 'ruff' for extremely fast Python linting"
        entry: ruff check --force-exclude
        language: python
        types_or: [python]
        args: [--fix, --exit-non-zero-on-fix]
        require_serial: true
        additional_dependencies: []
        minimum_pre_commit_version: "2.9.2"
      - id: ruff-format
        name: ruff-format
        description: "Run 'ruff format' for extremely fast Python formatting"
        entry: ruff format --force-exclude
        language: python
        types_or: [python]
        args: []
        require_serial: true
        additional_dependencies: []
        minimum_pre_commit_version: "2.9.2"

```

Let me know what you think is best @KennethEnevoldsen

Samoed:
I think that you can change to 
```yaml
  - repo: local
    hooks:
      - id: lint
        name: lint
        description: "Run 'make lint'"
        entry: make lint
        language: python
        types_or: [python]
        minimum_pre_commit_version: "2.9.2"
```

sam-hey:
@Samoed Yes, this would be an option, but it would mean adding `--exit-non-zero-on-fix` to the Makefile.  
My impression is that this could be rigid, as someone might remove it without knowing why it is there, since it is not obvious.

KennethEnevoldsen:
This looks reasonable to me. 

My only concern is that this will cause a lot of merge conflicts. It might be ideal to remove the end of line fixes + similar until after the v2 merge?

KennethEnevoldsen:
Assuming that there are no other edits, of course. 

I am fine with merging if you do the merge to v2 afterward. There is an ongoing merge here (#2204)

sam-hey:
https://github.com/embeddings-benchmark/mteb/pull/2235
This enables merging with just the normal issues.

sam-hey:
So what do you think, can we merge this?

KennethEnevoldsen:
Yes all good here - we just need to merge to v2 as well

sam-hey:
@KennethEnevoldsen I've resolved all conflicts and completed a lint check. It's ready to be merged. Thanks a lot for your help!

sam-hey:
Thanks a lot, as always, for your great work and help! @KennethEnevoldsen @Samoed

Issue #2156, opened 2025-02-25
Remove overlapping legends from radar chart


KennethEnevoldsen:
Currently, the legend of radar chart is a bit hard to read:

<img width="793" alt="Image" src="https://github.com/user-attachments/assets/63f1ef81-2fc3-4a68-bdf7-90e3bebf0595" />

Moving these to the side would solve the issue.

Kes104:
Hello @KennethEnevoldsen  I am interested in contributing to this issue, Can you help me with file name where changes has to be made please ...

KennethEnevoldsen:
I would start by looking at the `radar_chart()`function in `src/mteb/leaderboard/figures.py`

x-tabdeveloping:
Shouldn't be too hard to fix. I think it's probably because I forced the legend to be in the bottom horizontally. If you remove that piece of code it might just work.

ayush1298:
Added PR #2195, Can anyone have a look at it?

ayush1298:
Whenever I make certain changes in code related to the leaderboard, do I need to run `python -m mteb.leaderboard.app` again each time, or is there any other way to see the changes without running again?

KennethEnevoldsen:
As it is now you have to rerun. We could have it re-use the cache though

PR #2195, opened 2025-02-28
Remove overlapping legends from radar chart

ayush1298:
This PR is related to the issue #2156 to remove overlapping legends from the radar chart

### Code Quality

The Radar Plot now looks like below where the legend is shifted to the right.
newplot (4)

### DIFF
--- a/mteb/leaderboard/figures.py
+++ b/mteb/leaderboard/figures.py
@@ -243,13 +243,12 @@ def radar_chart(df: pd.DataFrame) -> go.Figure:
             ),
         ),
         legend=dict(
-            orientation="h",
-            yanchor="bottom",
-            y=-0.6,
+            orientation="v",
+            yanchor="middle",
+            y=0.5,
             xanchor="left",
-            x=-0.05,
-            entrywidthmode="fraction",
-            entrywidth=1 / 5,
+            x=1.1,
         ),
+        margin=dict(l=200, r=150, t=50, b=50),
     )
     return fig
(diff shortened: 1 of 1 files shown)


isaac-chung:
@ayush1298 thanks for working on this! Could we make it such that it's fully in the margin and not blocking the graph at all?

ayush1298:
> @ayush1298 thanks for working on this! Could we make it such that it's fully in the margin and not blocking the graph at all?
>

@isaac-chung, I mistakenly didn't see that some part of the graph was still blocking previosuly. I have corrected it and added the updated image in the description. Can you have a look at it now?

isaac-chung:
@x-tabdeveloping I wonder if the original placement at the bottom worked better? I do like the vertical orientation.

ayush1298:
@x-tabdeveloping, can you have a look at this?

KennethEnevoldsen:
@x-tabdeveloping is at a conference, but it sounds from @isaac-chung that we still have issues with overlap? (the picture seems fine to me)

> I wonder if the original placement at the bottom worked better?

From the issue https://github.com/embeddings-benchmark/mteb/issues/2156 it made it very hard to know which model was which.

It might be possible to do a vertical list at the buttom, but I would prefer this solution (assuming the overlap isn't a problem)

x-tabdeveloping:
hmm it takes a lot of space up this way though, can you play around with it a bit more to make it look a bit more balanced?
I think the horizontal positioning would also be nice if we could get it to break into multiple lines

ayush1298:
@KennethEnevoldsen @x-tabdeveloping , does the following plot look good? Tried many things, but none of them worked. Also, there is no option to break legend into multiple lines. If this looks good, then I will push a commit.

newplot (3)

ayush1298:
@isaac-chung I have pushed the commit and added an update plot in the description of the PR.

isaac-chung (approved):
Great, thank you!

Issue #1606, opened 2024-12-17
Normalize image encoders interface to the text encoder interface


KennethEnevoldsen:
As discussed in this [PR](https://github.com/embeddings-benchmark/mteb/pull/1551) we would like to normalize the interfaces to allow for:

```py
model.encode(sentences)
model.encode(images)
model.encode(documents) # can be image and text
```

isaac-chung:
Adding to [this comment](https://github.com/embeddings-benchmark/mteb/pull/1551#discussion_r1876142413), the encoder can then be something like:

```python
class MultiModalEncoder:
    """For MIEB, MOEB, etc.
    """

    def __init__(self, device: str | None,**kwargs: Any):
        pass

    def encode(
        self,
        inputs: Sequence[str] | Sequence[Image.Image] | DataLoader | Sequence[Any],
        *,
        task_name: str,
        prompt_type: PromptType | None = None,
        **kwargs: Any,
    ) -> np.ndarray:
        """
        inputs: Handles uni-modal or multi-modal inputs. For example:
            - text only: Sequence[str]
            - image and text: Sequence[Any] -> Sequence[tuple[Image.Image, str]]
            - image dataloader and text: Sequence[Any] -> tuple[DataLoader, list[str]]
            - images and text: Sequence[Any] -> Sequence[tuple[list[Image.Image], str]]
        
        This can potentially wrap existing get_fused_embedding methods as well.
        """
        pass
```

isaac-chung:
This may be applicable to audio + video as well. Tagging @Muennighoff too for MOEB.

KennethEnevoldsen:
Hmm maybe:
```
inputs: Iterable[mteb.document] | Iterable[str] | Iterable[image] | Iterable[audio]

"""
        inputs: Handles uni-modal or multi-modal inputs. For example:
            - text only: Iterable[str]
            - images: Iterable[image]
            - ...
            - Multimodal: Iterable[mteb.document]. A document is a mteb object that contain text, image, audio. It can even contain text metadata such as title. These attributes can be acceced using document.body, document.title, document.images, document.audio. You can also get the relative position of the modalities in the text using document.text which is formatting as follows: "We can see in the picture <image-1>", where "<image-1>" refers to document.images[1].
"""
```

Currently documents are not needed for anything in MTEB, but it would simplify a lot of cases (retrieval + joint image + text cases). We don't need to add the ".text" attribute to start with.

@isaac-chung wdyt?

[...]

Issue #1330, opened 2024-10-26
Add promptType Conversation


KennethEnevoldsen:
@Samoed, it seems like we might accidentally have removed the encode_conversations without a replacement. This seems like a bug.

(Checking in with @vaibhavad here as well, ensuring that encode_conversations() is required - we now denote these using the promptType e.g. "query" and "corpus".

Samoed:
I tried to leave it as it, but in models folder there weren't models that implement this function to test it and look for implementation. 
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/evaluation/evaluators/RetrievalEvaluator.py#L346-L349

KennethEnevoldsen:
@vaibhavad do you have a model that implements an encode_conversations() method?

KennethEnevoldsen:
@vaibhavad we would love to refactor this part of the code. We sadly can't refactor it if we don't know how it works

vaibhavad:
Hi @KennethEnevoldsen, apologies for the delay. I will push a resolution by the end of this week.

Samoed:
Do we need to do this because we have updated the inputs to the `encode` function and the models can now receive a full conversation?

KennethEnevoldsen:
@Samoed, so how do you imagine the input here?

Samoed:
This is currenly done in `v2` branch. https://github.com/embeddings-benchmark/mteb/blob/559dcce710914c1f50924fca464bb4b263a5387d/mteb/types.py#L56

```json
{
  "text": "user: text 1\nassistant: text 2",
  "query": [{"message": "text 1", "role": "user"}, {"message": "text 2", "role": "asistant"}]
}
```
I can try to find conversational datasets to test

KennethEnevoldsen:
I think this would be totally fine

KennethEnevoldsen:
How about:
```
{
  "text": "user: text 1\nassistant: text 2",
  "query": "user: text 1\nassistant: text 2",
   "conversation": [{"message": "text 1", "role": "user"}, {"message": "text 2", "role": "asistant"}]
}
```
To prevent too many types of query?

Samoed:
I can change that to this format

KennethEnevoldsen:
We also have the list[list[str]]. When is that used? Can we have multiple queries?

Samoed:
I think this is for conversations too, where replics without `roles`

KennethEnevoldsen:
Ahh great so then we can greatly simplify query

PR #2198, opened 2025-02-28
[v2] Refactor text tasks to use DataLoader

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/issues/1606

Now models will receive `encode` function Dataloader
```json
{
   "text": [...],  # default text
   "image": [...], 
   "audio": [...], 
   "body: [...], # models are allowed to construct the text from the body + title if they wish
   "title: [...],
}
```

### Code Quality


Samoed:
> Right now it is very much a quick wrapper. Wouldn’t we prefer directly working with the dataset for datasets? (I know that this is more code to write)

It's not easy because datasets have different column names and most datasets require encoding two columns, and I don’t have a clear solution for handling that. Also in most tasks list of sentences passed to evaluators and there datasets can't be used for now, but we can change that. Additionally, some datasets return a dictionary instead of a dataset, and Pair classification expects all data to be in the first row (as I recall). I could pass the dataset directly and select columns, but that would be a similar approach to using a wrapper. (edited)

KennethEnevoldsen (commented):
So I would really like to see how a Dataloader native abstask would look like. Can we try to do it with just Classification?

I am also afraid of how much this influences throughput - can we do a quick test e.g. using minishlab models?

It is a bit annoying that we have to convert everything in the encode functions (it might be the right solution). We could consider whether it better to just hand of the Dataset object to the model? (but I assume that does not work for images?)

mteb/evaluation/evaluators/model_classes.py
```diff
@@ -110,48 +93,66 @@ def search(
...
+        unique_queries = [pair[0] for pair in unique_pairs]
+        unique_instructions = (
+            [pair[1] for pair in unique_pairs] if instructions else None
+        )
 
-        # Encode only unique queries
-        if isinstance(queries[0], list):
+        # Encode only unique queries using the dataloader
+        if isinstance(query_list[0], list):
+            # For conversations, still use the original encode_conversations method
```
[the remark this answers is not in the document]
Samoed: We want, but I still don't know what to do with them, because we don't have implementation for any model https://github.com/embeddings-benchmark/mteb/issues/1330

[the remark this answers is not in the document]
KennethEnevoldsen: I pinged him. Can't we just convert it to text and keep the "conversation in a column as well??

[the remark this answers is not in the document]
Samoed: Yes, can change like that

Samoed:
I've updated clustering and classification tasks to use `Dataset` more natively

mteb/create_dataloaders.py
```diff
@@ -0,0 +1,117 @@
...
+from typing import Callable
+
+import torch
+from datasets import Dataset
+from torch.utils.data import DataLoader
+
+from mteb.encoder_interface import BatchedInput
+
+
+def create_dataloader(
```
KennethEnevoldsen, line 12: is this needed?

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -58,13 +61,13 @@ def __call__(
         max_f1 = 0
         max_ap = 0
         X_train = model.encode(
-            self.sentences_train,
+            create_dataloader_from_texts(self.sentences_train),
```
KennethEnevoldsen, line 64: We are still using the old format here?

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -272,32 +271,31 @@ def __call__(
...
-            self.sentences_train,
+            DataLoader(self.train_dataset),
             task_name=self.task_name,
             **encode_kwargs,
         )
         if test_cache is None:
-            X_test = model.encode(
-                self.sentences_test,
+            test_cache = model.encode(
+                DataLoader(self.eval_dataset),
```
KennethEnevoldsen, line 280:
```suggestion
                DataLoader(self.eval_dataset.with_format("torch"), **dataloader_kwargs),
```

Not sure we need `with_format("torch")`

mteb/abstasks/AbsTaskClassification.py
```diff
@@ -143,11 +143,13 @@ def _evaluate_subset(
                 idxs,
             )
 
+            train_dataset = Dataset.from_dict(
+                {"text": X_sampled, "label": y_sampled}
+            )
```
KennethEnevoldsen, line 148: Couldn't we just sample directly from the dataset using the indices (not need to recreate it)

Also doesn't:

```py
    def _undersample_data(self, X, y, samples_per_label: int, idxs=None):
        """Undersample data to have samples_per_label samples of each label"""
        # ...
        rng_state = np.random.default_rng(self.seed)
        rng_state.shuffle(idxs)
        # ....
```
result in exactly the same shuffle every time? (seems problematic?)

KennethEnevoldsen (commented):
Looking better. 

I added a few comments on the classification.

we should also update the documentation to match (how to implement a custom encoder).

mteb/create_dataloaders.py
Samoed: Yes, I'm using this function in other functions in this file

mteb/evaluation/evaluators/ClassificationEvaluator.py
Samoed: I think so, because with it, the dataloader will output `Tensors` instead of lists. This will be helpful for image tasks.
https://huggingface.co/docs/datasets/en/use_with_pytorch#dataset-format

mteb/evaluation/evaluators/ClassificationEvaluator.py
Samoed: I've only updated the `logreg` classifier for now. I haven't updated `kNNClassificationEvaluator` or `kNNClassificationEvaluatorPytorch` yet.

mteb/evaluation/evaluators/ClassificationEvaluator.py
KennethEnevoldsen: but it does not seem to be required for text:

```py
import torch
ds = load_dataset('DDSC/lcc', split='train', streaming=True) # small ds
ds["text"]
dataloader = torch.utils.data.DataLoader(ds.with_format("torch"), num_workers=0)
sample = next(iter(dataloader))
sample # <-- same regardless of with_format
```

mteb/create_dataloaders.py
KennethEnevoldsen: Hmm but maybe just use Dataloader direct (it is a one-liner anyway)

mteb/evaluation/evaluators/ClassificationEvaluator.py
Samoed: For texts yes, but image would require this
```python
from datasets import load_dataset
import torch
dataloader = torch.utils.data.DataLoader(ds.with_format("torch"))
dataset = load_dataset("ylecun/mnist")["test"]
dataloader = torch.utils.data.DataLoader(ds.with_format("torch"), num_workers=0)
sample = next(iter(dataloader))
# {'image': tensor([[[[  0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,
#             0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0],
```

mteb/abstasks/AbsTaskClassification.py
Samoed: I tried to run it and it sampled different indexes
```python
        rng_state = np.random.default_rng(self.seed)
        rng_state.shuffle(idxs)
        print(idxs)
```
```
[  712 24774 45142 ...  5668 22113  9864]
[11345 32908 14658 ... 13719 39909 28784]
[ 8445 45360 19144 ...  3998 20588 32862]
[ 4024 15204  4345 ...  4456 31586 14108]
[24812  4537 17666 ... 20520 44755 21161]
[24220 27535 20122 ... 35776 45309 45714]
[22663 14095 17823 ...  1575 48150 36188]
[ 4985 14854 12100 ... 32535 38903 46329]
[34107 37558 15857 ...  1858 30374  6695]
[43526 27456 42653 ...  5937 12086 16500]
```

mteb/evaluation/evaluators/ClassificationEvaluator.py
KennethEnevoldsen: Agree, but for text tasks we can remove it (but for anything general it we can keep it there)

mteb/abstasks/AbsTaskClassification.py
KennethEnevoldsen: Ahh yea right because the state in the list changes, but if we keep the list static it does not:

```py
import numpy as np
idxs = list(range(100))
rng_state = np.random.default_rng(42)
rng_state.shuffle(idxs)
print(idxs) # [59, 21, 56, 18, 33, 42, 50, ...
rng_state = np.random.default_rng(42)
rng_state.shuffle(list(range(100)))
print(idxs) # [59, 21, 56, 18, 33, 42, ...
```

Which it seems like we do (recreated if idxs is None)

A solution would be to grab the `self.rng_state`.

Samoed: We shuffle idxs every experiment and pass them to next experiment

Samoed: Updated to use dataset directly when shuffling

mteb/abstasks/AbsTaskClassification.py
```diff
@@ -162,21 +159,36 @@ def _evaluate_subset(
...
+        """Undersample data to have `samples_per_label` samples of each label.
+
+        Args:
+            dataset (Dataset): Hugging Face `datasets.Dataset` containing "text" and "label".
+            samples_per_label (int): Number of samples per label to retain.
+            idxs (list[int] | None): Optional indices to shuffle and sample from.
+
+        Returns:
+            Dataset: A new Dataset containing undersampled examples.
+            list[int]: The shuffled indices used for sampling.
```
KennethEnevoldsen, line 174:
```suggestion
            dataset: Hugging Face `datasets.Dataset` containing "text" and "label".
            samples_per_label: Number of samples per label to retain.
            idxs: Optional indices to shuffle and sample from.

        Returns:
            A new Dataset containing undersampled examples.
            The shuffled indices used for sampling.
```

mteb/abstasks/AbsTaskClassification.py
KennethEnevoldsen: Ahh no because we pass the indexes back in

Samoed:
I've updated Classification evaluator and removed `create_dataloader`. What else do you want to change?

mteb/abstasks/AbsTaskClassification.py
```diff
@@ -162,21 +159,36 @@ def _evaluate_subset(
...
+        Returns:
+            A new Dataset containing undersampled examples.
+            The shuffled indices used for sampling.
+        """
         if idxs is None:
-            idxs = np.arange(len(y))
+            idxs = list(range(len(dataset)))
+
         rng_state = np.random.default_rng(self.seed)
         rng_state.shuffle(idxs)
```
KennethEnevoldsen, line 180:
```suggestion
         self.rng_state.shuffle(idxs)
```
test and believe they should be eq.

mteb/abstasks/AbsTaskClassification.py
```diff
@@ -240,3 +252,15 @@ def _calculate_metrics_from_split(
...
+# [  712 24774 45142 ...  5668 22113  9864]
+# [11345 32908 14658 ... 13719 39909 28784]
+# [ 8445 45360 19144 ...  3998 20588 32862]
+# [ 4024 15204  4345 ...  4456 31586 14108]
+# [24812  4537 17666 ... 20520 44755 21161]
+# [24220 27535 20122 ... 35776 45309 45714]
+# [22663 14095 17823 ...  1575 48150 36188]
+# [ 4985 14854 12100 ... 32535 38903 46329]
+# [34107 37558 15857 ...  1858 30374  6695]
+# [43526 27456 42653 ...  5937 12086 16500]
```
KennethEnevoldsen, line 266:
```suggestion
```

mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -177,7 +178,7 @@ def _evaluate_subset(
             downsampled_dataset = dataset.select(example_indices)  # type: ignore
 
         embeddings = model.encode(
-            downsampled_dataset["sentences"],  # type: ignore
+            create_dataloader_from_texts(downsampled_dataset["sentences"]),
```
KennethEnevoldsen, line 181: we can do a direct conversion

mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -81,7 +82,7 @@ def _evaluate_subset(
         unique_train_sentences = train_split.select(unique_train_indices)["text"]
 
         _unique_train_embeddings = model.encode(
-            unique_train_sentences,
+            create_dataloader_from_texts(unique_train_sentences),
```
KennethEnevoldsen, line 85: We can directly work with the dataset here as well

mteb/encoder_interface.py
```diff
@@ -17,6 +17,27 @@ class PromptType(str, Enum):
...
 
+class BatchedInput(TypedDict, total=False):
+    """The input to the encoder. This is the input to the encoder when using the encode_batch function.
+
+
+    Args:
+        text: The text to encode.
+        image: The image to encode. Can be a list of images or a list of lists of images.
+        audio: The audio to encode. Can be a list of audio files or a list of lists of audio files.
+    """
```
KennethEnevoldsen, line 28: missing descriptions for args

mteb/encoder_interface.py
```diff
@@ -17,6 +17,27 @@ class PromptType(str, Enum):
     passage = "passage"
 
 
+class BatchedInput(TypedDict, total=False):
+    """The input to the encoder. This is the input to the encoder when using the encode_batch function.
```
KennethEnevoldsen, line 21: encode_batch -> encode?

mteb/encoder_interface.py
```diff
@@ -17,6 +17,27 @@ class PromptType(str, Enum):
...
+        audio: The audio to encode. Can be a list of audio files or a list of lists of audio files.
+    """
+
+    text: list[str]
+    image: list[list[Image.Image]]
+    audio: list[list[bytes]]
+    # Retrieval corpus
+    title: list[str]
+    body: list[str]
+    # Retrieval query
```
KennethEnevoldsen, line 36: ea are only for instruction retrieval right?

KennethEnevoldsen (approved):
A few more minor things.

Would love @isaac-chung s opinion on this as well

(would love to see more adaption of tasks to avoid the many dataset transformation)

mteb/encoder_interface.py
Samoed: Yes, instruction used only in instruction retrieval/reranking

mteb/abstasks/AbsTaskClassification.py
Samoed: On the first experiment they're equal and on others they're different

mteb/evaluation/evaluators/model_classes.py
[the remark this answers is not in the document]
Samoed: I've tried to standardize it in https://github.com/embeddings-benchmark/mteb/pull/2198/commits/bb2a897a642a1fb3222f95cb23e1315482c1e7be, but it is hard to tell if it correct, because I don't know conversational datasets to check results

Samoed:
@orionw It would be great if you could review this PR!

orionw:
Dataloader seems fine to me if it helps make things easier. I am not sure what the benefits are offhand, but I am not opposed. 

Re: allowing to choose how to combine passage and title, I like the motivation but this is a fairly large change (every single model that anyone ever uses). 

Could we instead define a custom function that can be overridden, something like "combine_passage_and_title" and have a default? I am hesitant to make such a large API change. We already have a custom function that can be overridden for `combine_query_and_instruction`

Samoed:
Now, the title and passage are computed the same way as before in `text` section, so this won't break anything. However, we could add a function to allow overriding if needed. 

The main benefit of dataloaders is standardizing input, especially since we now have images and audio, which are difficult to handle otherwise. You can check the discussion in [this thread](https://github.com/embeddings-benchmark/mteb/issues/1606)

orionw:
That makes sense and seems good to have the input change happen with v2 then. It is a lot of changes but shouldn't change anything of substance.

> Now, the title and passage are computed the same way as before in text field, so this won't break anything. However, we could add a function to allow overriding if needed.

Seems great then. We can add it as an extension if we want but not high priority.

Samoed:
@KennethEnevoldsen We will wait for more reviews, or can we merge this?

KennethEnevoldsen:
Good to merge!

Issue #2200, opened 2025-02-28
Frequent 403 Errors in GitHub Actions – Possible Solutions?


sam-hey:
Hello,

I am currently encountering a high number of 403 status codes in GitHub Actions. The jobs take an extremely long time and ultimately fail:

https://github.com/embeddings-benchmark/mteb/actions/runs/13586589565/job/37982824854
https://github.com/embeddings-benchmark/mteb/actions/runs/13591675569/job/37999052862

This is quite frustrating, and finding a solution would be greatly appreciated.

If someone is in contact with Hugging Face and could help address this issue, that would be fantastic. Otherwise, do we have alternative options? Is there general access to the infrastructure of the runners that we could leverage? For example, could deploying a reverse proxy to cache requests help reduce the load on Hugging Face?

Looking forward to any insights.

More: 
307 error
https://github.com/embeddings-benchmark/mteb/actions/runs/13587992217/job/37987196418#step:5:183

@isaac-chung @KennethEnevoldsen @Samoed

Samoed:
Maybe if a test failed, it tried to rerun all tests, and the anti-spam system rejected the requests.

sam-hey:
Yes, that is the main reason, but I still encountered so many of them today... It's just really annoying. And without maintainer privileges, you have to push a new commit just to retry a single failed job-only to have a high probability that another job will fail on the next attempt.

isaac-chung:
> Yes, that is the main reason, but I still encountered so many of them today... It's just really annoying. And without maintainer privileges, you have to push a new commit just to retry a single failed job-only to have a high probability that another job will fail on the next attempt.

And that's okay 😃 Then it's a maintainer's responsibility to help with those when the PR is ready. Most, if not all, tests can be run locally anyway.

Separately, I feel that this might be a signal to update the tests. Perhaps dataset availability of 1000 other datasets do not need to be checked at every PR, after every commit. I think these can be separated out, and once a day (or even a few days) would suffice. If new datasets were added, then only those should be tested in the PR.

isaac-chung:
Closing this now. Feel free to reopen this if the issue persists.

PR #2201, opened 2025-02-28
ci: dont run test_dataset_on_hf in every pr

sam-hey:
ref: https://github.com/embeddings-benchmark/mteb/issues/2200
test: test_dataset_on_hf

It is just running in pr with changes in `'mteb/tasks/**.py'`
or every 2 days 

### Code Quality

### Documentation

### Testing

### DIFF
--- a/.github/workflows/dataset_loading.yml
+++ b/.github/workflows/dataset_loading.yml
@@ -1,0 +1,26 @@
+name: Dataset Loading
+
+on:
+  pull_request:
+    paths:
+      - 'mteb/tasks/**.py'
+  schedule:
+    - cron: '0 0 */2 * *' # every 2 days
+
+jobs:
+  extract-and-run:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+        cache: 'pip'
+
+    - name: Install dependencies and run test
+      run: |
+        make dataset-load-test BASE_BRANCH=${{ github.event.pull_request.base.ref }}
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ lint-check:
 
 test:
 	@echo "--- 🧪 Running tests ---"
-	pytest -n auto 
+	pytest -n auto
 
 test-with-coverage:
 	@echo "--- 🧪 Running tests with coverage ---"
@@ -43,6 +43,13 @@ model-load-test:
 	pip install ".[dev, speedtask, pylate,gritlm,xformers,model2vec]"
 	python scripts/extract_model_names.py $(BASE_BRANCH) --return_one_model_name_per_file
 	python tests/test_models/model_loading.py --model_name_file scripts/model_names.txt
+
+
+dataset-load-test:
+	@echo "--- 🚀 Running dataset load test ---"
+	pip install ".[dev, speedtask]"
+	NO_SKIP=1 pytest -n auto tests/test_tasks/test_all_abstasks.py::test_dataset_on_hf
+
 
 run-leaderboard:
 	@echo "--- 🚀 Running leaderboard locally ---"
--- a/tests/test_tasks/test_all_abstasks.py
+++ b/tests/test_tasks/test_all_abstasks.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from unittest.mock import Mock, patch
-
+import os
 import pytest
 import requests
 
@@ -65,7 +65,7 @@ def test_load_data(
         if not task.is_multilingual:
             mock_dataset_transform.assert_called_once()
 
-
+@pytest.mark.skipif(condition=os.getenv("NO_SKIP", 0) == 1, reason="Just in separate CI Job")
 @pytest.mark.flaky(
     reruns=3,
     reruns_delay=5,


.github/workflows/dataset_loading.yml
Samoed, line 1: I don't think this is a good solution because we wouldn't get any useful information, because not many people check the Actions tab for the tasks.

Samoed: We can try to move retries directly to the test function with delays, but this is a bit not what you want

isaac-chung (commented):
@sam-hey thanks for putting this together. It looks good to me so far. 

I was wondering if it would make more sense _in the next iteration_ to run
- everything once every 2 days (ok)
- just the changed / added tasks in the PR (👈)

Would love to get @KennethEnevoldsen 's take here and on the original issue as well.

.github/workflows/dataset_loading.yml
isaac-chung: @Samoed I get what you're proposing. However, @sam-hey already mentioned that the changed files will be run in each PR, so contributors don't need to fish in the actions tab. Also, as mentioned in the issue, failures in checking unrelated datasets often block development. I think these suggestions are not mutually exclusive, so feel free to add delays as well.

sam-hey: All of the requested features are already implemented. I believe that some IPs may be getting blacklisted temporarily. Therefore, if you receive a 403 error, the request will certainly fail, even with longer delays.

https://github.com/embeddings-benchmark/mteb/blob/29464aca853c14a1eec66b3f553d1dafde8e7218/tests/test_tasks/test_all_abstasks.py#L69-L76

Samoed: I'm not sure if one test of `test_dataset_on_hf` is failed will `flaky` rerun all parameters or just one.

We can run this test on merge to main. This would give better visibility.

sam-hey: Just the one revision will be re run!

```
    @pytest.mark.flaky(reruns=2, condition=sys.platform.startswith("win32"))
    @pytest.mark.parametrize("x", [1, 2, 3])
    def test_example(x):
        execution_count[x] = execution_count.get(x, 0) + 1  # Track execution count
    
        print(f"Running test with x={x}, attempt {execution_count[x]}")
    
>       assert x != 2  # Only x=2 should fail and be retried
E       assert 2 != 2

tmp_2.py:14: AssertionError
--------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------
Running test with x=2, attempt 1
=============================================================== slowest 5 durations ================================================================

(5 durations < 0.005s hidden.  Use -vv to show these durations.)
```

tests/test_tasks/test_all_abstasks.py
```diff
@@ -66,6 +67,9 @@ def test_load_data(
             mock_dataset_transform.assert_called_once()
 
 
+@pytest.mark.skipif(
+    condition=os.getenv("NO_SKIP", 0) == 1, reason="Just in separate CI Job"
+)
```
KennethEnevoldsen, line 72: Hmm, I would probably use:


```suggestion
@pytest.mark.test_datasets
```

which is then run using:
```
pytest -m pytest.mark.test_datasets
```

[docs](https://docs.pytest.org/en/stable/example/markers.html)

Makefile
```diff
@@ -44,6 +44,13 @@ model-load-test:
 	python scripts/extract_model_names.py $(BASE_BRANCH) --return_one_model_name_per_file
 	python tests/test_models/model_loading.py --model_name_file scripts/model_names.txt
 
+
+dataset-load-test:
+	@echo "--- 🚀 Running dataset load test ---"
+	pip install ".[dev, speedtask]"
+	NO_SKIP=1 pytest -n auto tests/test_tasks/test_all_abstasks.py::test_dataset_on_hf
```
KennethEnevoldsen, line 51:
```suggestion
	pip install ".[dev, speedtask]"
	pytest -n auto -m test_datasets
```

KennethEnevoldsen (changes requested):
This looks good - added a few suggestions, but nothing that changes behaviour.

sam-hey:
@Samoed What do you think? I would still split this up, even though we have your changes, as it makes it clearer that this might not be an error related to the code but rather some third party.

Samoed:
Since we changed checking to the official API it seems that it's not a problem for now. We can split this, but run on merge or commits for clarification

.github/workflows/dataset_loading.yml
```diff
@@ -0,0 +1,26 @@
+name: Dataset Loading
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  schedule:
+    - cron: '0 0 */2 * *' # every 2 days
```
Samoed, line 8: I think we can remove the schedule, because we run this frequently

tests/test_tasks/test_all_abstasks.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import os
```
Samoed, line 4:
```suggestion
```

KennethEnevoldsen (approved):
Great!

.github/workflows/dataset_loading.yml
```diff
@@ -0,0 +1,27 @@
+name: Datasets available on HuggingFace
+
+on:
+  pull_request:
```
isaac-chung, line 4: > We can split tests, but run on merge or commits for visibility

@Samoed should we remove `pull_request` then?se

Samoed: I think yes, we can remove it

Issue #2094, opened 2025-02-18
Add tasks relations


Samoed:
For now, we need to annotate models with all translation tasks
```
    "MSMARCO": ["train"],
    "MSMARCOHardNegatives": ["train"],
    "NanoMSMARCORetrieval": ["train"],
    "MSMARCO-PL": ["train"],  # translation not trained on
    "mMARCO-NL": ["train"],  # translation not trained on
    "NQ": ["train"],
    "NQ-NL": ["train"],  # translation not trained on
    "NQHardNegatives": ["train"],
    "NanoNQRetrieval": ["train"],
    "NQ-PL": ["train"],  # translation not trained on
```
We can add to `ModelMeta` field for translated datasets `translated_from` and auto check in leaderboard

KennethEnevoldsen:
Assuming you are referring to `TaskMetadata`?

PR #2228, opened 2025-03-03
Automatically add similar tasks to training_tasks

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/2094

This PR adds the `adapted` dataset to the training dataset automatically.

For example, if annotation looks like
```python
ModelMeta(
    training_datasets={"Touche2020": []},  # or {"Touche2020-NL": []}
)
```
All `Touche2020` datasets will be added with `get_training_tasks`
```
[
        "NanoTouche2020Retrieval",
        "Touche2020",
        "Touche2020-Fa",
        "Touche2020-NL",
        "Touche2020Retrieval.v3",
]
```
Also, if model specifies `adapted_from` all training datasets from `parent` model will be added too. 
### Code Quality

### Documentation

### Testing


KennethEnevoldsen (commented):
We should change the leaderboard to use `.get_training_tasks`.

You can also see if there are other places where we call the attribute.

(will you do a merge to v2 after this one)

Otherwise I think everything looks good

Samoed:
Leaderboard using `zero_shot_percentage` from `ModelMeta` and I've updated it

KennethEnevoldsen:
Alright then, I don't have any issues. feeel free to merge this

KennethEnevoldsen (approved):

Samoed:
Leaderboard from this branch
image

Leaderboard on spaces
image

mteb/models/overview.py
```diff
@@ -248,6 +249,10 @@ def get_model_meta(model_name: str, revision: str | None = None) -> ModelMeta:
             )
         return MODEL_REGISTRY[model_name]
     else:  # assume it is a sentence-transformers model
+        if mteb_only:
```
Samoed, line 254: I've added this because otherwise, it would try to download metadata from Hugging Face. What do you think, @KennethEnevoldsen?

x-tabdeveloping:
Huh cool! While we're at it we could also add the training models of the parent model in this method. That would make a lot of things easier

mteb/models/overview.py
KennethEnevoldsen: Hmm, I would probably rename it to:

```
fetch_from_hf: bool = True
```

KennethEnevoldsen: otherwise I think it is a good thing to add

Samoed:
I think I've already added it https://github.com/embeddings-benchmark/mteb/blob/293f7bc3a93a7b74592bc8bce9580231f968033c/mteb/model_meta.py#L181-L190

You can look in test. `deepvk/USER-bge-m3` dosen't have `MIRACLRetrieval`, but "parent" model `bge-m3` has it. 
https://github.com/embeddings-benchmark/mteb/blob/293f7bc3a93a7b74592bc8bce9580231f968033c/tests/test_models/test_model_meta.py#L70-L75

x-tabdeveloping:
riight sorry, didn't notice, but that's great

PR #2234, opened 2025-03-04
fix: Allow model to output torch.tensor

KennethEnevoldsen:
Adressed aspects of https://github.com/embeddings-benchmark/mteb/issues/941 by allowing model.encode to return a torch.tensor.

Also does a lot of fixes to array typing.

# further problems
1) Unsqueeze not in array API spec, we should use expand_dims, but torch does not implement this (https://github.com/pytorch/pytorch/issues/56774). We can use reshape instead.
2) What to do with torch.functional? We would have to reimplent these to make it work.

We have a few cases of torch.compile. Which are called every time the function is called. They seem to have been added by @orionw, do we expect that these speed this up (they are compiled every time the function is called)

### Code Quality

### Documentation

### Testing

- [ NA ] **New Tests Added**: Write tests to cover new functionality. Validate with `make test-with-coverage`.


mteb/encoder_interface.py
```diff
@@ -40,7 +40,7 @@ def encode(
         task_name: str,
         prompt_type: PromptType | None = None,
         **kwargs: Any,
-    ) -> np.ndarray:
+    ) -> np.ndarray | torch.Tensor:
```
Samoed, line 43: Maybe model should output only one type of array for simplification?

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -172,7 +171,9 @@ def __call__(
         return scores, test_cache
 
     @staticmethod
-    def _cos_sim(a: Tensor, b: Tensor):
```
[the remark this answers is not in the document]
KennethEnevoldsen: Removed it. KNN is supported by sklearn (bumped version to >=1.4.0)

```py
import torch
X = [[0], [1], [2], [3]]
y = [0, 0, 1, 1]
from sklearn.neighbors import KNeighborsClassifier
neigh = KNeighborsClassifier(n_neighbors=3)
neigh.fit(torch.tensor(X), torch.tensor(y))
# works
```

mteb/encoder_interface.py
KennethEnevoldsen: Currently are valid for the encoder interface. Isn't this what we want?

KennethEnevoldsen: But isn't that what https://github.com/embeddings-benchmark/mteb/issues/941 is all about?

Samoed: I don't know well array-api spec. We can use something that not in in and this can cause problems. For now in code some things done using `np.something`, so I don't know how it would handle tensors

KennethEnevoldsen: We have tests for that - they are already implemented in `MockTorchEncoder`.

Samoed: Then this is great!

KennethEnevoldsen:
@orionw would love your opinion here on the compiled functions. This would, e.g., allow us to e.g. replace

`torch.nn.functional.normalize` -> `sklearn.preprocessing.normalize` (~eq.)

which allow for both torch tensor and numpy arrays

orionw:
I am not sure why they are compiled every time, they should be compiled once. I added it because on tasks with lots to normalize it was about a 10x speedup. 

We can remove it for simplicity, but is is **much much** faster.

KennethEnevoldsen:
Hmm, that is good to know. I will keep it, at least for now

PR #2277, opened 2025-03-07
fix: add annotation models for stella zh

KennethEnevoldsen:
Additionally fixed a few annotation errors

### DIFF
--- a/mteb/models/stella_models.py
+++ b/mteb/models/stella_models.py
@@ -6,6 +6,41 @@ from mteb.model_meta import ModelMeta
 from mteb.models.instruct_wrapper import instruct_wrapper
 from mteb.models.nvidia_models import nvidia_training_datasets
 
+stella_zh_datasets = {
+    "BQ": [],
+    "LCQMC": [],
+    "PAWSX": [],
+    "STS-B": [],
+    "DuRetrieval": [],
+    "AFQMC": [],
+    "Cmnli": [],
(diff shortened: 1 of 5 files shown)


mteb/models/stella_models.py
```diff
@@ -135,7 +172,7 @@
     adapted_from="dunzhang/stella-large-zh-v3-1792d",
     public_training_code=None,
     public_training_data=None,
-    training_datasets=None,  # Not specified
+    training_datasets=stella_zh_datasets,  # Not specified
```
isaac-chung, line 175:
```suggestion
    training_datasets=stella_zh_datasets,  
```

mteb/models/stella_models.py
```diff
@@ -6,6 +6,41 @@
...
+    "LCQMC": [],
+    "PAWSX": [],
+    "STS-B": [],
+    "DuRetrieval": [],
+    "AFQMC": [],
+    "Cmnli": [],
+    "Ocnli": [],
+}
+
+# Derived from conversation:
```
isaac-chung, line 20: Might be worth linking the convo if public?

isaac-chung (approved):
Got some non-blocking suggestions.

mteb/models/stella_models.py
KennethEnevoldsen: Not public - mail

Issue #1769, opened 2025-01-11
Integrate vdr-2b-multilingual


Muennighoff:
Integrate https://huggingface.co/blog/vdr-2b-multilingual

Both models & data

ayush1298:
@isaac-chung @gowitheflow-1998, Does this model have the same implementation as other multilingual models? Can you give me a reference to any model and data to which I can see?

isaac-chung:
@ayush1298 the key here is that it's a model that supports images and texts. So feel free to reference any relevant ones in the models folder.

ayush1298:
@isaac-chung @Muennighoff, do we have to add the dataset also, because it does not have a validation or test split, it has train split only?

ayush1298:
Can anyone clarify this one whether the dataset has to add or not?

PR #2290, opened 2025-03-08
Added VDR Model

ayush1298:
This PR is related to addition of VDR model and Training Dataset given in issue #1769 
- [x] **Model Addition**:  llamaindex/vdr-2b-multi-v1 model added

### Code Quality

### Documentation

### Testing

### Adding a model checklist

### DIFF
--- a/mteb/models/vdr_models.py
+++ b/mteb/models/vdr_models.py
@@ -1,0 +1,267 @@
+from __future__ import annotations
+
+from functools import partial
+from typing import Any
+
+import torch
+from PIL import Image
+from torch.utils.data import DataLoader
+from transformers import AutoProcessor, Qwen2VLForConditionalGeneration
+
+from mteb.encoder_interface import PromptType
+from mteb.model_meta import ModelMeta
+from mteb.models.wrapper import Wrapper
+
+
+class VDRWrapper(Wrapper):
+    def __init__(
+        self,
+        model_name: str = "llamaindex/vdr-2b-multi-v1",
+        device: str = "cuda:0" if torch.cuda.is_available() else "cpu",
+        **kwargs: Any,
+    ):
+        """Wrapper for VDR multimodal model that supports both image and text.
+
+        Args:
+            model_name: Path to model or model name
+            device: Device to use (cuda, cpu)
+            **kwargs: Additional arguments to pass to the model
+        """
+        super().__init__()
+        self.model_name = model_name
+        self.max_pixels = kwargs.get("max_pixels", 768 * 28 * 28)
+        self.min_pixels = kwargs.get("min_pixels", 1 * 28 * 28)
+
+        # Load processor with image size constraints
+        self.processor = AutoProcessor.from_pretrained(
+            model_name, min_pixels=self.min_pixels, max_pixels=self.max_pixels
+        )
+        use_flash_attention = (
+            "cuda" in device
+            and kwargs.get("attn_implementation", "flash_attention_2")
+            == "flash_attention_2"
+        )
+        self.model = Qwen2VLForConditionalGeneration.from_pretrained(
+            model_name,
+            attn_implementation="eager"
+            if not use_flash_attention
+            else "flash_attention_2",
+            torch_dtype=torch.bfloat16,
+            device_map=device,
+        )
+
+        # Set model configuration
+        self.model.eval()
+        self.model.padding_side = "left"
+        self.processor.tokenizer.padding_side = "left"
+        self.device = device
+        self.embed_dim = kwargs.get("embed_dim")
+
+        # Define standard prompts for text and queries
+        self.document_prompt = "<|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\n<|vision_start|><|image_pad|><|vision_end|>What is shown in this image?<|im_end|>\n<|endoftext|>"
+        self.query_prompt = "<|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\n<|vision_start|><|image_pad|><|vision_end|>Query: %s<|im_end|>\n<|endoftext|>"
+        self.text_only_prompt = "<|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\n%s<|im_end|>\n<|endoftext|>"
+
+    def encode(
+        self,
+        sentences: list[str],
+        *,
+        task_name: str | None = None,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> torch.Tensor:
+        """Returns embeddings for text inputs."""
+        return self.get_text_embeddings(sentences, **kwargs)
+
+    def get_text_embeddings(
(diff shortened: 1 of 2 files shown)


ayush1298:
@Samoed, can you please review this?
I have some doubts:
1. When I tried to load the model using `mteb.get_model(model_name = "llamaindex/vdr-2b-multi-v1"
revision = "2c4e54c8db4071cc61fc3c62f4490124e40c37db")`, then it gives me the following warning:
```
WARNING:mteb.model_meta: Loader not specified for model llamaindex/vdr-2b-multi-v1, loading using sentence transformers.
```
I don't know why it is using sentence transformers and not the Custom one.
2. In code given at [hugging face](https://huggingface.co/llamaindex/vdr-2b-multi-v1), there is a function name `encode_documents`, under which it is using following code for preparing inputs:
 ```
cache_position = torch.arange(0, len(queries))
inputs = model.prepare_inputs_for_generation(
    **inputs, cache_position=cache_position, use_cache=False)
```
But, this was giving a following dimension mismatch error to me:
```
RuntimeError: The size of tensor a (43) must match the size of tensor b (512) at non-singleton dimension 0
```
So, I change the above piece of code to:
```
seq_length = inputs["input_ids"].shape[1]
            cache_position = torch.arange(0, seq_length).to(self.device)
            inputs = self.model.prepare_inputs_for_generation(
                **inputs, cache_position=cache_position, use_cache=False
            )
```
It's running fine after this, can you confirm if this is correct?

File: mteb/models/vdr_models.py [hunk 1]
```diff
@@ -0,0 +1,267 @@
...
+            == "flash_attention_2"
+        )
+        self.model = Qwen2VLForConditionalGeneration.from_pretrained(
+            model_name,
+            attn_implementation="eager"
+            if not use_flash_attention
+            else "flash_attention_2",
+            torch_dtype=torch.bfloat16,
+            device_map=device,
+        )
```
Samoed, line 51: I don't think that you need this, because model supporting SentenceTransformers https://huggingface.co/llamaindex/vdr-2b-multi-v1/blob/main/custom_st.py

Samoed:
> When I tried to load the model using mteb.get_model(model_name = "llamaindex/vdr-2b-multi-v1" revision = "2c4e54c8db4071cc61fc3c62f4490124e40c37db"), then it gives me the following warning

I think you installed the library incorrectly. It looks like you ran `pip install .` instead of `pip install -e .`, which means you're running an `mteb` version where your VDR model doesn't exist.

> Where exactly the dataset needs to be added, should I add it under mteb/tasks/Reranking/multilingual with a new file name vdrmultilingual.py

If you're about training datasets, then you need to add it to model meta like
```python
model = ModelMeta(
    ....,  
    training_datasets={
      #   llamaindex/vdr-multilingual-train
   },
)
```

ayush1298:
> I think you installed the library incorrectly. It looks like you ran `pip install .` instead of `pip install -e .`, which means you're running an `mteb` version where your VDR model doesn't exist.
> 
Do I need to do `pip install -e .` every time I make changes in the model file?


> > Where exactly the dataset needs to be added, should I add it under mteb/tasks/Reranking/multilingual with a new file name vdrmultilingual.py
> 
> If you're about training datasets, then you need to add it to the model meta like
> 
> ```python
> model = ModelMeta(
>     ....,  
>     training_datasets={
>       #   llamaindex/vdr-multilingual-train
>    },
> )
> ```
>
Can you once see the issue #1769, I think it's not about adding it to the model meta, it's about adding it as a task also.

mteb/models/vdr_models.py
Samoed: Yes, but this would support only text and for image this would need new wrapper, but I'll do it later

Samoed:
> Do I need to do pip install -e . every time I make changes in the model file?

No. `-e` means `--editable`, so you will run library as now in your code

> Can you once see the issue https://github.com/embeddings-benchmark/mteb/issues/1769, I think it's not only about adding it to the model meta, it's about adding it as a task also.

We can, but I think you can add model in this PR, and dataset in another

mteb/models/vdr_models.py
ayush1298: Can you give me some idea about how to do it, and I can have a try

Samoed: I'm doing refactor how we handle multimodal input in https://github.com/embeddings-benchmark/mteb/pull/2294, so I don't think you should do this, because input will be changed

ayush1298: Got it

mteb/models/vdr_models.py
```diff
@@ -0,0 +1,61 @@
...
+
+from mteb.encoder_interface import PromptType
+from mteb.model_meta import ModelMeta
+from mteb.models.instruct_wrapper import InstructSentenceTransformerWrapper
+
+
+def instruction_template(
+    instruction: str, prompt_type: PromptType | None = None
+) -> str:
+    return "<|im_start|>system\nYou are a helpful assistant.<|im_end|>\n<|im_start|>user\n%s<|im_end|>\n<|endoftext|>"
```
Samoed, line 15: I don't think you need to specify this, because this will be applied in their implementation https://huggingface.co/llamaindex/vdr-2b-multi-v1/blob/2c4e54c8db4071cc61fc3c62f4490124e40c37db/custom_st.py#L234
But in that case instructions will be applied in user prompt, which can be incorrect, but I'm not sure how to do it right. Maybe @logan-markewich or @marplex can suggest something

ayush1298: cool, I will wait for their reply.

mteb/models/vdr_models.py
```diff
@@ -0,0 +1,61 @@
...
+            "torch_dtype": torch.bfloat16,
+            "normalized": True,
+        },
+    ),
+    name="llamaindex/vdr-2b-multi-v1",
+    languages=languages,
+    open_weights=True,
+    revision="2c4e54c8db4071cc61fc3c62f4490124e40c37db",
+    release_date="2024-01-08",
+    modalities=["image", "text"],  # Do we have to change this to only text for now?
```
ayush1298, line 46: @Samoed , what should be here?

ayush1298:
> We can, but I think you can add the model in this PR, and the dataset in another
>
 Should I add it under mteb/tasks/Reranking/multilingual with a new file name vdrmultilingual.py in another PR?

Samoed:
> Should I add it under mteb/tasks/Reranking/multilingual with a new file name vdrmultilingual.py in another PR?
Yes

mteb/models/vdr_models.py
ayush1298: @Samoed, should we close this now with this instruction_template? Then you can look at it once again when you implement the multimodal input for it.

Samoed: You should change template to apply instruction. Other template will be added inside sentence transformers automatically
```suggestion
    return "{instruction}"
```

ayush1298: @Samoed done, this one

Samoed (approved):
Great work!

Issue #2078, opened 2025-02-16
[v2] Merge MIEB and MTEB tasks


Samoed:
- Since in `v2`, we merged `Retrieval` and `Reranking` tasks, and `MultiChoice` appears to be a `Reranking` task with a single correct answer, can we merge it with the `Retrieval` task?
- If we have `MultiChoiceAny2Any`, do we still need `MultiChoiceAny2Text`?  
- Can we create `AbsTaskClassification` and split it into `AbsTaskTextClassification` and `AbsTaskImageClassification`, since they share the same logic in the `evaluate` function?  
- Zero-shot classification seems more similar to `ImageTextSimilarity`—should we just rename it?  
- I think we could introduce `Any2AnyPairClassification` and `Any2AnySTS`.  

Ref https://github.com/embeddings-benchmark/mteb/issues/1337

isaac-chung:
Thanks for starting this! Updated to use question numberings. Tagging @gowitheflow-1998 while I take a deeper look into the AbsTasks. A few points to begin with:

1. I don't think so. See its docstrings. 
2. Yes. See their respective docstrings. 
3. It's a good first suggestion. I'd prefer not to add another layer of abstraction, and I'd like to work towards the proposal originally [in this thread](https://github.com/embeddings-benchmark/mteb/discussions/1950#discussioncomment-12112662, where we create e.g. an `AbsTaskAnyClassification` that can handle any modalities, while reusing the evaluate parts.
4. I was not able to find `ImageTextSimilarity`. Do you have a link, or do you mean `ImageTextPairClassification`?
5. Agreed!

Samoed:
1. I see the docstring, but I can't find any differences in the code, except for one relevant document that should be for the query.
2. I'm not sure I understand the difference between docstrings. Additionally, it's a bit confusing that we have both `AbsTaskAny2TextMultipleChoice` and `AbsTaskAny2AnyMultiChoice`. Why do we need `Any2Text` when we have `Any2Any`?
3.  
4. Yes, `ImageTextPairClassification`

isaac-chung:
[...]

PR #2396, opened 2025-03-19
fix: Minor style changes

KennethEnevoldsen:
Adresses #2078


isaac-chung:
Would you mind merging master? That way we can see the changes for just this PR ☺️

isaac-chung (approved):

PR #2404, opened 2025-03-20
Add validation to model_name in `ModelMeta`

Samoed:
Add validation to model_name in `ModelMeta`.

### Code Quality

### Documentation

### Testing

### Adding datasets checklist

### Adding a model checklist

### DIFF
--- a/mteb/models/cohere_v.py
+++ b/mteb/models/cohere_v.py
@@ -182,7 +182,7 @@ def cohere_v_loader(**kwargs):
 
 cohere_mult_3 = ModelMeta(
     loader=partial(cohere_v_loader, model_name="embed-multilingual-v3.0"),
-    name="embed-multilingual-v3.0-v",
+    name="cohere/embed-multilingual-v3.0",
     languages=[],  # Unknown, but support >100 languages
     revision="1",
     release_date="2024-10-24",
@@ -204,7 +204,7 @@ cohere_mult_3 = ModelMeta(
 
 cohere_eng_3 = ModelMeta(
     loader=partial(cohere_v_loader, model_name="embed-english-v3.0"),
-    name="embed-english-v3.0-v",
+    name="cohere/embed-english-v3.0",
     languages=["eng-Latn"],
     revision="1",
     release_date="2024-10-24",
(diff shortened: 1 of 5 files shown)


mteb/models/cohere_v.py
```diff
@@ -204,7 +204,7 @@ def get_fused_embeddings(
 
 cohere_eng_3 = ModelMeta(
     loader=partial(cohere_v_loader, model_name="embed-english-v3.0"),
-    name="embed-english-v3.0-v",
```
Samoed, line 207: Same as in `v2`

mteb/models/cohere_v.py
```diff
@@ -204,7 +204,7 @@ def get_fused_embeddings(
...
+    name="Cohere/Cohere-embed-english-v3.0",
```
Samoed, line 207: To make compatible with current results dir https://github.com/embeddings-benchmark/results/tree/main/results/Cohere__Cohere-embed-english-v3.0

Samoed: It seems we have same models in https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/cohere_models.py, but they not support vision. Maybe we should remove text only implementation and left only multimodal in v2?

isaac-chung: We can, as long as we move the ModelMeta.

Hmm, there don't seem to be any image tasks in the linked folder. @gowitheflow-1998 maybe we haven't run results for this model?

Samoed: I believe we should combine two implementations, but the `cohere_models` module uses `cohere.Client`, while `cohere_v` utilizes `cohere.ClientV2`. Therefore, we need to verify if the results for text tasks are the smae with `cohere.ClientV2`.

gowitheflow-1998: > Hmm, there don't seem to be any image tasks in the linked folder. @gowitheflow-1998 maybe we haven't run results for this model?

yes, we weren't able to run it months back because the rate limit was 5 images per minute.

Samoed: So, we can merge implementation and integrate them properly after this? 5 image per minute is for free trial. I think we can request them for credits

gowitheflow-1998: found that it is now raised to 400 for production key but back in last Nov it was 5 for both trial and production when it was just released. We can run it if we want

isaac-chung (approved):

PR #2392, opened 2025-03-18
Add Background Gradients in Summary and Task Table

ayush1298:
This PR is related to applying background gradients for all columns in summary table performance per task table on leaderboard for better visibility as described in issue #2382 

Closes #2382 
<details>
<summary> Summary Table </summary>
Screenshot 2025-03-26 at 10 37 57 PM
</details>

<details>
<summary> Performance Per Task Table </summary>
Screenshot 2025-03-26 at 8 19 47 PM

</details>

### Code Quality

- [x] **Code Formatted**: Format the code using `make lint` to maintain consistent style.

### Documentation

- [ ] **Updated Documentation**: Add or update documentation to reflect the changes introduced in this PR.

### Testing

- [ ] **New Tests Added**: Write tests to cover new functionality. Validate with `make test-with-coverage`.
- [x] **Tests Passed**: Run tests locally using `make test` or `make test-with-coverage` to ensure no existing functionality is broken.

PR #2406, opened 2025-03-20
suppress logging warnings on leaderboard

Samoed:
It seems that the previous logging configuration was not properly set up, as there were multiple basicConfigs in place. I have moved the `demo` object into a function to suppress the logger inside the `__name__ == "__main__"` module so that it does not affect other modules. It would break current leaderboard, because I removed `demo` object. I've also created PR to simplify build https://huggingface.co/spaces/mteb/leaderboard/discussions/160

### Code Quality

### Documentation

### Testing

### Adding datasets checklist

### Adding a model checklist


mteb/leaderboard/app.py
```diff
@@ -23,13 +23,6 @@
 from mteb.leaderboard.figures import performance_size_plot, radar_chart
 from mteb.leaderboard.table import scores_to_tables
```
isaac-chung, line 25: If the changes were only the following, i.e. 1) change to ERROR and 2) specify model_meta logger, the same effect can be achieved. See printout below.

```suggestion

logging.getLogger("mteb.load_results.task_results").setLevel(
    logging.ERROR
)  # Warnings related to task split
logging.getLogger("mteb.model_meta").setLevel(
    logging.ERROR
)
logging.getLogger("mteb.models.overview").setLevel(
    logging.ERROR
)  # Warning related to model metadata (fetch_from_hf=False)
```

Here is the printout:
```
(mteb) isaac@macbook mteb % python -m mteb.leaderboard.app
<frozen runpy>:128: RuntimeWarning: 'mteb.leaderboard.app' found in sys.modules after import of package 'mteb.leaderboard', but prior to execution of 'mteb.leaderboard.app'; this may result in unpredictable behaviour
* Running on local URL:  http://127.0.0.1:7860
* Running on public URL: https://e0b4a9baf5be43a504.gradio.live

This share link expires in 72 hours. For free permanent hosting and GPU upgrades, run `gradio deploy` from the terminal in the working directory to deploy to Hugging Face Spaces (https://huggingface.co/spaces)
```

isaac-chung (commented):
Maybe we can achieve the same effect with minimal changes to the code. See comment.

mteb/leaderboard/app.py
isaac-chung: Makes sense.

isaac-chung (approved):

mteb/leaderboard/__init__.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import annotations
 
-from mteb.leaderboard.app import demo
+from mteb.leaderboard.app import run_leaderboard
```
KennethEnevoldsen, line 3: Hmm but I would want to access the demo object e.g. for debugging.

I also think this breaks the make run-leaderboard command

KennethEnevoldsen (changes requested):
Logging changes look good. I don't think the wrapper function is needed.

mteb/leaderboard/__init__.py
Samoed: No, this wouldn't break it. I tried to run it like this. I can change function to output demo object instead of running directly

Samoed:
The `app` file has `all_results = load_results()` at the top, which loads all results immediately. I don’t think we should leave it like this, especially if we plan to include the leaderboard in the main `__init__`, as it would trigger the loading. Because of that, I’ve moved it into a function.  

I also tried moving only the logging level configuration inside the `if __name__ == "__main__"` block, but since results are loaded before logging is suppressed, the logs still appear.

KennethEnevoldsen:
> The app file has all_results = load_results() at the top, which loads all results ...

I don't think we should import the leaderboard in the main init.

> I also tried moving only the logging level configuration inside ...

Ahh right this is indeed correct. I would have it return the demo object then (we should probably rename it then)

Samoed:
I've reused old space to test leaderboard from this branch. I think the problem with previous leaderboard was that I haven't passed `server_name` to launch. Now it working from this branch https://huggingface.co/spaces/mteb/leaderboard_2_demo

KennethEnevoldsen:
I think this looks good. It will clash with #2392 -- should we merge the two branches?

Samoed:
Maybe. I'll try to update build on space now and after that this can be merged

Samoed:
I've updated leaderboard deployment https://huggingface.co/spaces/mteb/leaderboard and it's working. This PR can be merged

KennethEnevoldsen:
Should we create a separate branch `leaderboard` and merge this on in and the move https://github.com/embeddings-benchmark/mteb/pull/2392 over that branch and then we can also add #2438 (I think all of these together is quite a good leaderboard release).

Marton and I can look at #2438

Samoed:
@KennethEnevoldsen Can this PR be merged?

KennethEnevoldsen:
Yep, it's all good to merge

Samoed:
Tried to run this leaderboard version in `2_demo` space and got error. Merged main, because I think issue related to https://github.com/embeddings-benchmark/mteb/pull/2456

Samoed:
Yes, it's working fine now. Will merge

PR #2454, opened 2025-03-28
CLI Tool for results dataframe on leaderboard

### LINKED ISSUE #2174: Add CLI tool for inspecting and creating dataframe of the results on a give leaderboard
Currently it is quite hard to "just" get a data frame for inspecting the results. 

We could add a CLI like:
```
mteb create-table 
  --results results/  # where to fetch the results from
  --models model1,model2,model3 # what models to include in the table
  --bechmark "mteb(eng, v1)" # what benchmark
```

This would also be useful when comparing models on the results repo as well.

ayush1298:
This PR is related to addition of CLI Tool for inspection and creation of dataframe of results as suggested in issue #2174 .
closes #2174 .

To use it command is: 
```
python scripts/create_table.py --results <results_path> --models <comma_separated_models> --benchmark <benchmark_name> --aggregation-level <aggregation-level > --output <output_file>
```

### Code Quality

### Documentation

### Testing

### DIFF
--- a/scripts/create_table.py
+++ b/scripts/create_table.py
@@ -1,0 +1,306 @@
+from __future__ import annotations
+
+import argparse
+import logging
+import os
+from pathlib import Path
+from typing import Literal
+
+import numpy as np
+import pandas as pd
+
+import mteb
+from mteb.load_results import load_results
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def get_available_benchmarks():
+    """Get all available benchmark names."""
+    return [b.name for b in mteb.get_benchmarks()]
+
+
+def save_dataframe(
+    df: pd.DataFrame,
+    output_path: str,
+):
+    """Save a DataFrame to the specified format based on file extension.
+
+    Args:
+        df: The DataFrame to save
+        output_path: Path for the output file, extension determines format
+
+    Returns:
+        str: The full path to the saved file
+    """
+    ext = Path(output_path).suffix.lower()
+    fallback_path = str(Path(output_path).with_suffix(".csv"))
+
+    def warn_and_fallback(reason: str):
+        """Logs a warning and saves the DataFrame as CSV instead."""
+        logger.warning(f"{reason}. Defaulting to CSV format: {fallback_path}")
+        df.to_csv(fallback_path, index=False)
+        return fallback_path
+
+    if ext == ".csv":
+        df.to_csv(output_path, index=False)
+    elif ext == ".xlsx":
+        try:
+            df.to_excel(output_path, index=False)
+        except ImportError:
+            return warn_and_fallback(
+                "openpyxl not installed. Please install with 'pip install mteb[xlsx]' to save as Excel."
+            )
+    elif ext == ".md":
+        try:
+            with open(output_path, "w") as f:
+                f.write(df.to_markdown(index=False))
+        except ImportError:
+            return warn_and_fallback(
+                "tabulate not installed. Please install with 'pip install mteb[markdown]' to save as Markdown."
+            )
+    else:
+        return warn_and_fallback(
+            f"Unsupported file extension: {ext}, defaulting to CSV"
+        )
+
+    return output_path
+
+
+def create_comparison_table(
+    results_folder: str,
+    model_names: list[str],
+    benchmark_name: str | None = None,
+    output_path: str | None = None,
+    aggregation_level: Literal["subset", "split", "task"] = "task",
+) -> pd.DataFrame:
+    """Create comparison tables for MTEB models.
+
+    Args:
+        results_folder: Path to the results folder
+        model_names: List of model names to include
+        benchmark_name: Name of the benchmark (optional)
+        output_path: Path to save the output tables
+        aggregation_level: Level of aggregation for results ('subset', 'split', or 'task')
+                          - 'subset': Results for each subset within each split for each task
+                          - 'split': Results aggregated over subsets for each split for each task
+                          - 'task': Results aggregated over subsets and splits for each task
+
+    Returns:
+        result_df: DataFrame with aggregated results
+    """
+    logger.info(f"Creating comparison table for models: {', '.join(model_names)}")
+    logger.info(f"Using aggregation level: {aggregation_level}")
+
+    # Load results
+    benchmark_results = load_results(
+        results_repo=results_folder,
+        only_main_score=True,
+        require_model_meta=False,
+        models=model_names,
+    )
+
+    # Filter by benchmark if specified
+    if benchmark_name:
+        logger.info(f"Filtering tasks for benchmark: {benchmark_name}")
+        benchmark = next(
+            (b for b in mteb.get_benchmarks() if b.name == benchmark_name), None
+        )
+        if not benchmark:
+            raise ValueError(
+                f"Benchmark '{benchmark_name}' not found. Available: {get_available_benchmarks()}"
+            )
+
+        benchmark_results_filtered = benchmark.load_results(
+            base_results=benchmark_results
+        ).join_revisions()
+    else:
+        logger.info("Using all available tasks for the specified models")
+        benchmark_results_filtered = benchmark_results.join_revisions()
+
+    # Check if we have any results
+    if not benchmark_results_filtered.model_results or not any(
+        model_result.task_results
+        for model_result in benchmark_results_filtered.model_results
+    ):
+        logger.warning("No results found for the specified models and benchmark")
+        return pd.DataFrame()
+
+    # Get detailed scores
+    scores_data = []
+    for model_result in benchmark_results_filtered.model_results:
+        model_name = model_result.model_name
+        for task_result in model_result.task_results:
+            task_name = task_result.task_name
+            for split, scores_list in task_result.scores.items():
+                for score_item in scores_list:
+                    scores_data.append(
+                        {
+                            "model_name": model_name,
+                            "task_name": task_name,
+                            "split": split,
+                            "subset": score_item.get("hf_subset", "default"),
+                            "score": score_item.get("main_score", 0.0) * 100,
+                        }
+                    )
+
+    if not scores_data:
+        logger.warning("No scores found for the specified models and benchmark")
+        return pd.DataFrame()
+
+    scores_df = pd.DataFrame(scores_data)
+
+    # Create the appropriate table based on aggregation level
+    if aggregation_level == "subset":
+        # For subset level, show raw data at task/split/subset level (no aggregation)
+        pivot_df = scores_df.pivot_table(
+            index=["task_name", "split", "subset"],
+            columns="model_name",
+            values="score",
+            aggfunc="mean",
+        ).reset_index()
+
+    elif aggregation_level == "split":
+        # For split level, aggregate across subsets for each task/split combination
+        agg_df = (
+            scores_df.groupby(["model_name", "task_name", "split"])["score"]
+            .mean()
+            .reset_index()
+        )
+        pivot_df = agg_df.pivot_table(
+            index=["task_name", "split"],
+            columns="model_name",
+            values="score",
+            aggfunc="mean",
+        ).reset_index()
+
+    elif aggregation_level == "task":
+        # For task level, aggregate across both subsets and splits for each task
+        agg_df = (
+            scores_df.groupby(["model_name", "task_name"])["score"].mean().reset_index()
+        )
+        pivot_df = agg_df.pivot_table(
+            index=["task_name"],
+            columns="model_name",
+            values="score",
+            aggfunc="mean",
+        ).reset_index()
+
+    pivot_df.columns.name = None
+    model_cols = [
+        col for col in pivot_df.columns if col not in ["task_name", "split", "subset"]
+    ]
+    if model_cols:
+        # Create mean row based on aggregation level
+        if aggregation_level == "subset":
+            # Add an empty row for overall mean
+            overall_mean_row = {"task_name": "mean_score", "split": "", "subset": ""}
+            for model in model_cols:
+                overall_mean_row[model] = pivot_df[model].mean()
+            pivot_df = pd.concat(
+                [pivot_df, pd.DataFrame([overall_mean_row])], ignore_index=True
+            )
+
+        elif aggregation_level == "split":
+            overall_mean_row = {"task_name": "mean_score", "split": ""}
+            for model in model_cols:
+                overall_mean_row[model] = pivot_df[model].mean()
+            pivot_df = pd.concat(
+                [pivot_df, pd.DataFrame([overall_mean_row])], ignore_index=True
+            )
+
+        elif aggregation_level == "task":
+            # Add overall mean row
+            overall_mean_row = {"task_name": "mean_score"}
+            for model in model_cols:
+                overall_mean_row[model] = pivot_df[model].mean()
+            pivot_df = pd.concat(
+                [pivot_df, pd.DataFrame([overall_mean_row])], ignore_index=True
+            )
+
+    # Round scores to 2 decimal places
+    numeric_columns = pivot_df.select_dtypes(include=np.number).columns
+    pivot_df[numeric_columns] = pivot_df[numeric_columns].round(2)
+
+    # Save output if path is provided
+    if output_path:
+        output_dir = Path(output_path).parent
+        os.makedirs(output_dir, exist_ok=True)
+
+        save_dataframe(pivot_df, output_path)
+        logger.info(f"Comparison table saved to {output_path}")
+
+    return pivot_df
+
+
+def format_table_for_display(df: pd.DataFrame) -> str:
+    """Format a DataFrame for terminal display."""
+    max_rows = 10
+    if len(df) > max_rows:
+        display_df = df.head(max_rows)
+        return f"{display_df.to_string()}\n... {len(df) - max_rows} more rows"
+    return df.to_string()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Create comparison tables for MTEB models."
+    )
+
+    parser.add_argument(
+        "--results",
+        type=str,
+        default="results/",
+        help="Path to fetch results from (local folder or GitHub repo URL)",
+    )
+
+    parser.add_argument(
+        "--models",
+        type=str,
+        required=True,
+        help="Comma-separated list of models to include in the table",
+    )
+
+    parser.add_argument(
+        "--benchmark",
+        type=str,
+        default=None,
+        help=f"Benchmark to use (optional). Available: {get_available_benchmarks()}",
+    )
+
+    parser.add_argument(
+        "--aggregation-level",
+        type=str,
+        choices=["subset", "split", "task"],
+        default="task",
+        help="Level of aggregation for results (subset, split, or task)",
+    )
+
+    parser.add_argument(
+        "--output",
+        type=str,
+        default="comparison_table.csv",
+        help="Output path for the generated table (include extension: .csv, .xlsx, or .md)",
+    )
+
+    args = parser.parse_args()
+
+    models = [model.strip() for model in args.models.split(",")]
+
+    result_df = create_comparison_table(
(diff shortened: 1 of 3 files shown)


scripts/create_table.py
Samoed, line 1: I think we thought about integrating this to MTEB CLI, rather than separete script

scripts/create_table.py
```diff
@@ -0,0 +1,306 @@
...
+        default="results/",
+        help="Path to fetch results from (local folder or GitHub repo URL)",
+    )
+
+    parser.add_argument(
+        "--models",
+        type=str,
+        required=True,
+        help="Comma-separated list of models to include in the table",
+    )
```
Samoed, line 263: Can we change default to None and create table with scores for all models?

Samoed (commented):
Great addition!

ayush1298:
Update PR description, with command to run, sample example and results and their verifications

scripts/create_table.py
ayush1298: from issue [description](https://github.com/embeddings-benchmark/mteb/issues/2174#issue-2882739878), I thought that these will be case similar to running `make_leaderboard.py`.

Samoed: No, I thought that this should be run with `mteb ...` as said in the description

scripts/create_table.py
ayush1298: So, do you want that these feature is also optional, and if not specified then we should use all models?

scripts/create_table.py
ayush1298: sorry, I misunderstood it. How to change these script to CLI then to run with `mteb `?

scripts/create_table.py
Samoed: Yes!

scripts/create_table.py
Samoed: I don't think that you need to change much. You can just integrate your script into https://github.com/embeddings-benchmark/mteb/blob/main/mteb/cli.py

ayush1298: Okay, I will look and add that. Should I keep these script also for backward compatibility?

ayush1298: I am saying to keep it just because not to move all helper functions there. Or, should I move them also in `cli.py`

Samoed: You can't import functions from script folder inside mteb repo. You can create separate file to store functions

ayush1298: I have in integrated this one. Should we have a mean_score row also at end which will show average results for a model. It is there in current code if you see the function `create_comparison_table`?

Samoed: Yes, good idea!

mteb/cli.py
```diff
@@ -354,6 +368,50 @@ def add_create_meta_parser(subparsers) -> None:
...
+        default="results/",
+        help="Path to fetch results from (local folder or GitHub repo URL)",
+    )
+
+    parser.add_argument(
+        "--models",
+        type=str,
+        default=None,
+        help="Comma-separated list of models to include in the table (default: all models)",
+    )
```
Samoed, line 388: I think it would be better to add `nargs` and use like `--models model1 model2` 
```suggestion
    parser.add_argument(
        "--models",
        type=str,
        nargs='*',
        default=None,
        help="Models to include in the table (default: all models from results dir)",
    )
```

mteb/cli.py
```diff
@@ -354,6 +368,50 @@ def add_create_meta_parser(subparsers) -> None:
...
+
+    parser.add_argument(
+        "--benchmark",
+        type=str,
+        default=None,
+        help="Benchmark to use (optional). Available benchmarks can be listed with 'mteb available_benchmarks'",
+    )
```
Samoed, line 395: I think benchmark can be also a list
```suggestion
    parser.add_argument(
        "--benchmark",
        type=str,
        nargs='*',
        choices=mteb.get_bencharls(),
        default=None,
        help="Benchmark to use (optional). Available benchmarks can be listed with 'mteb available_benchmarks'",
    )
```

mteb/create_table.py
Samoed, line 1: I think this could be renamed to `create_results_table` to make its purpose a bit clearer.

mteb/create_table.py
```diff
@@ -0,0 +1,273 @@
...
+from pathlib import Path
+from typing import Literal
+
+import numpy as np
+import pandas as pd
+
+import mteb
+from mteb.load_results import load_results
+
+logging.basicConfig(level=logging.INFO)
```
Samoed, line 15:
```suggestion
logging.basicConfig(level=logging.INFO)
```

mteb/create_table.py
```diff
@@ -0,0 +1,273 @@
...
+        )
+
+    return output_path
+
+
+def create_comparison_table(
+    results_folder: str,
+    model_names: list[str] | None = None,
+    benchmark_name: str | None = None,
+    output_path: str | None = None,
```
Samoed, line 75: Output_path will always be in your cli
```suggestion
    output_path: str,
```

Samoed (commented):
Great! Can you create test for this like `test_create_meta_from_existing`?

https://github.com/embeddings-benchmark/mteb/blob/43adb0cd2a0cc611edc85c11df120629a7963de1/tests/test_cli.py#L137-L138

mteb/create_table.py
ayush1298: So, is it that we don't have to use these argument then?

mteb/create_table.py
ayush1298: I am seeing invalid suggestion here. Are you suggesting to remove empty line before `logging.basicConfig(level=logging.INFO)`?

Samoed: No, I'm suggesting to remove `logging.basicConfig` because it would change logging level for all mteb

ayush1298: okay, will remove this

ayush1298: I can keep the below line , 
`logger = logging.getLogger(__name__)`, right? As logger is used in that file.

mteb/cli.py
ayush1298: @Samoed current implementation was to support single benchmarks only. Does different benchmarks can have same tasks also or not?

Samoed: > Does different benchmarks can have same tasks also or not?.

Yes, they can.

> current implementation was to support single benchmarks only.

You can leave it as one benchmark, but it would be better to add choices.

ayush1298: > > Does different benchmarks can have same tasks also or not?.
> 
> Yes, they can.
> 
So, if a particular task is present in 2 benchmarks, then does it have different implementation and results also?
If not, then I thing I can have a 1 more column in table with benchmark names. And for task present under different benchmarks, the column will list out all of them like:

| task    | benchmark | split   | subset | model1 | model2 |
|----------|----------|-------|--------|--------|
| task1   | 'MTEB(Multilingual, v1)', 'MTEB(jpn, v1)| test  | subset1  | 50.55  | 50.55  |
|         | |          | subset2  | 50.55  | 50.55  |
|          | 'MTEB(Multilingual, v1)', 'MTEB(jpn, v1)| dev  | subset1  | 50.55  | 50.55  |
|         | |          | subset2   | 50.55  | 50.55  |

Samoed: Benchmarks mean scores on tasks, so it won't have `split` and `subset`. I think you can make to select only one benchmark at a time

Samoed:
I tried to run this command, but it is not working with local dirs. I tried to run with `mteb create-table --results v2/ --output example.md`, but it tried to load all results from results repo
[v2.zip](https://github.com/user-attachments/files/19522849/v2.zip)

ayush1298:
> I tried to run this command, but it is not working with local dirs. I tried to run with `mteb create-table --results v2/ --output example.md`, but it tried to load all results from results repo [v2.zip](https://github.com/user-attachments/files/19522849/v2.zip)
>
With local dir, there is some issue ig. As I have also got the same thing. I will try to find it out. You can try with github url as given in PR description. That was working as expected.

ayush1298:
@Samoed, Its working now with local path as well. Also, in tests. for testing it with local results repo, should I do results of both these models under `tests/results`?

mteb/create_results_table.py
```diff
@@ -144,7 +144,9 @@ def create_comparison_table(
                             "task_name": task_name,
                             "split": split,
                             "subset": score_item.get("hf_subset", "default"),
-                            "score": score_item.get("main_score", 0.0) * 100,
+                            "score": score_item.get("main_score", 0.0) * 100
+                            if score_item.get("main_score", 0.0) is not None
+                            else 0.0,
```
Samoed, line 149: Why this needed?

tests/test_cli.py
```diff
@@ -188,6 +188,80 @@ def test_create_meta_from_existing(existing_readme_name: str, gold_readme_name:
     assert result.returncode == 0, "Command failed"
 
 
+@pytest.mark.parametrize(
+    "results_folder",
+    [
+        # "tests/results",  # Local results folder
+        "https://github.com/embeddings-benchmark/results",  # Remote results repository
```
Samoed, line 195: I don't think that we want to download full results repo. You can use `tests/historic_results` in test

tests/test_cli.py
```diff
@@ -188,6 +188,80 @@ def test_create_meta_from_existing(existing_readme_name: str, gold_readme_name:
...
+    elif output_file.endswith(".xlsx"):
+        try:
+            import pandas as pd
+
+            df = pd.read_excel(output_path)
+            assert sorted(df.columns) == sorted(expected_headers), (
+                f"Excel headers do not match: {list(df.columns)}"
+            )
+        except ImportError:
+            pytest.fail("pandas or openpyxl is not installed for reading Excel files")
```
Samoed, line 253: Can you change to split to make test pass?

tests/test_cli.py
ayush1298: Here, I just want to check that both local path and remote url is working so added this. For local one, under `tests/historic_results`, there are no results of specific models. I think thet are just results on some tasks

mteb/create_results_table.py
ayush1298: When I run the command for "openai/text-embedding-3-large-256" "openai/text-embedding-3-small-instruct" models, specifically below command:
```
 mteb create-table --results /Users/ayushmunot/Desktop/OpenSource/results/ \
                 --models "openai/text-embedding-3-large-256" "openai/text-embedding-3-small-instruct" \
                 --benchmark "MTEB(Multilingual, v1)" \
                 --aggregation-level subset \
                 --output comparison_table5.csv
```
I got the error as: 
```
  File "/Users/ayushmunot/miniconda3/envs/ayushOS/bin/mteb", line 8, in <module>
    sys.exit(main())
  File "/Users/ayushmunot/Desktop/OpenSource/mteb/mteb/cli.py", line 448, in main
    args.func(args)
  File "/Users/ayushmunot/Desktop/OpenSource/mteb/mteb/create_results_table.py", line 255, in create_table_cli
    result_df = create_comparison_table(
  File "/Users/ayushmunot/Desktop/OpenSource/mteb/mteb/create_results_table.py", line 147, in create_comparison_table
    "score": score_item.get("main_score", 0.0) * 100
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```
So, I think somewhere None is present in main_score, so added that check

tests/test_cli.py
ayush1298: 1 thing is, do we need to check the headers of the files, or is file exists is sufficient?

mteb/cli.py
```diff
@@ -354,6 +368,51 @@ def add_create_meta_parser(subparsers) -> None:
...
+        type=str,
+        nargs="*",
+        default=None,
+        help="Models to include in the table (default: all models from results dir)",
+    )
+
+    parser.add_argument(
+        "--benchmark",
+        type=str,
+        default=None,
```
Samoed, line 394: Can you add choices here too? This would simplify validation.

ayush1298: Okay, will add it. Can you check comments for previous reviews also?
1 more thing are checks still failing because of the line: 
```
 if output_path.exists():
        output_path.unlink()
``

**Comment by ayush1298** [comment]
@KennethEnevoldsen Can you please review this PR?

File: mteb/cli.py (hunk 9 above)
Comment (ayush1298) [reply mteb/cli.py:394]: Sorry, I forgot this one. Added this now.

Comment (Samoed) [reply mteb/cli.py:394]: Great!

**Review by Samoed — APPROVED** [review]
Great work!

File: tests/test_cli.py (hunk 7 above)
Comment (KennethEnevoldsen) [reply tests/test_cli.py:195]: But we still don't want to download the whole repo - you can make a dummy folder for testing

File: tests/test_cli.py [hunk 10]
```diff
@@ -188,6 +188,54 @@ def test_create_meta_from_existing(existing_readme_name: str, gold_readme_name:
...
+    "results_folder",
+    [
+        # "tests/results",  # Local results folder
+        "https://github.com/embeddings-benchmark/results",  # Remote results repository
+    ],
+)
+@pytest.mark.parametrize("benchmark", ["MTEB(eng, v1)", "MTEB(Multilingual, v1)"])
+@pytest.mark.parametrize("aggregation_level", ["subset", "split", "task"])
+@pytest.mark.parametrize("output_format", ["csv", "md", "xlsx"])
+def test_create_table(
```
KennethEnevoldsen, line 201: I would love an additional test that doesn't go through the CLI.

KennethEnevoldsen (commented):
Thanks for this PR.

Currently it seems like we haven't really added a meaningful interface in python.

I think that there is a few use cases that we can see:

1) I have run a model and want to see how well it compared to another model
2) I make a PR to embeddings-benchmark/results and want to create a markdown table of performances

There is probably more, but I think it is worth spending some more time on getting this one right.

Feel free to just outline how you think it might look like in a test.

KennethEnevoldsen:
But there is no documentation on the python interface and no tests for functionality

Samoed:
> Taking a user perspective: I just evaluated my model (let's say a fine-tune of e5) and now want to compare the it with the original model. I don't care what directory something is in.

I think this can be done with additional command (or flag to current command) to compare current results with leaderboard or tasks from folder, but maybe this can be done in other PR

tests/test_cli.py
ayush1298: So, are you saying that, we should check only with results of model which will be added under `tests/historic_results` and not check whether it is able to pull that from github url?

tests/test_cli.py
ayush1298: @KennethEnevoldsen, without CLI how to test, because we are mainly dependent on output file that CLI is producing.

KennethEnevoldsen:
```
df = convert_to_df(results)
# do test on the df ?
```

KennethEnevoldsen:
Hmm, I will skip many of the comments here. This is not to be rude, but rather to focus on what I think is the main problem.

While I agree that this PR targets the Issue, it does it in a very direct way. This is a problem because then the CLI dictates the Python interface. This makes it inflexible and hard to test (we can only test the input/output). 

Here is a few example of why I don't think `create_comparison_table` is a good python interface:
- if I already have a `mteb.load_results(...)` I would have to recreate that to get to a dataframe
- it is a different interface that load_results and it is not compatible (e.g. exclude_meta is forced) 
- If I want to check something in the results before I make the table that is impossible
- If something is missing in the table, I have no idea why/when it was filtered out

So here is an alternative suggestion:

```py
models = ... # desired models
tasks = mteb.get_benchmark("MTEB(eng, v2)")

benchmark_results = benchmark.load_results(models = models, tasks = tasks)
df = benchmark_results.to_dataframe(aggregate = "task")
```

This might not be perfect, but I think it is definitely true. more flexible than the current implementation.

ayush1298:
@KennethEnevoldsen, sorry for having too low comments previously. 
>if I already have a mteb.load_results(...) I would have to recreate that to get to a dataframe
>
Using implementation that you suggested, still user has to run mteb.load_results(...) atleast 1 time. And using below code,
```
results_path = Path(results_repo)
    if results_path.exists() and results_path.is_dir():
        logger.info(f"Using local results repository at {results_path}")
        return results_path
```
if it runs 1 time, then it will use that only and not load again when we use `create_comparison_table` again. 
> If I want to check something in the results before I make the table that is impossible
>
What kind of checking you are referring here(any example), I think our main aim behind this was, just to provide easy way of comparing results. 
> If something is missing in the table, I have no idea why/when it was filtered out
>
Am I missing something here, because in whole code we are not doing any specific filtering.

Samoed:
I've used this command for creating tables in https://github.com/embeddings-benchmark/mteb/pull/2415 and I think there can be some improvements:
1. Add flag for transposing tables, because when I have 2-3 tasks and 5 models, it would be simpler to have tasks as headers.
2. `load_results` is expecting all data in results folder, but I can have different names and this is a bit confusing to have `custom_name/results` just to correctly create table https://github.com/embeddings-benchmark/mteb/blob/89db9ac28b16a0701646a9730732d1eb5419e50c/mteb/load_results/load_results.py#L114-L115

KennethEnevoldsen:
> Using implementation that you suggested...

> if it run it once, then it will use that only and not load again

Hmm, but yes, but if I already have it in memory, then I wouldn't. E.g.:

```
# analysis script
results = mteb.load_results(...)

# Some checks and analyses (e.g. task X is included, all tasks are run etc.)
# potentially custom filertering e.g. mteb version higher than X

# create table
create_comparison_table(...)
# requires respecification of argument
# does not guarantee that the results are the same as what is checked
# has to reload the results
```

ayush1298:
@KennethEnevoldsen I understand some of the things, but it will be good if you can write suggested solution whenever you become free after thursday. I will work on this then.

KennethEnevoldsen:
Perfect, I have set a reminder :)

ayush1298:
@KennethEnevoldsen After merge of #2542, Is their any work left that is to be done in this particular PR as I have saw that you have added `Prerequisite for https://github.com/embeddings-benchmark/mteb/pull/2454` in the description of that PR.
Sorry for replying after many days, I was busy because of exams.

Samoed:
1. I think you can solve these problems firstly https://github.com/embeddings-benchmark/mteb/pull/2454#issuecomment-2779598964
2. And these comments left unresolved https://github.com/embeddings-benchmark/mteb/pull/2454#pullrequestreview-2735550866

ayush1298:
> 1. I think you can solve these problems firstly [CLI Tool for results dataframe on leaderboard #2454 (comment)](https://github.com/embeddings-benchmark/mteb/pull/2454#issuecomment-2779598964)
> 2. And these comments left unresolved [CLI Tool for results dataframe on leaderboard #2454 (review)](https://github.com/embeddings-benchmark/mteb/pull/2454#pullrequestreview-2735550866)

aren't all these gets solved after #2542? As kenneth wants instead of CLI, that should be a function like `to_dataframe()` which he implemented

Samoed:
No, they're not solved. https://github.com/embeddings-benchmark/mteb/pull/2542 is added some functions to simplify working with `BenchmarkResults` and not address these issues

ayush1298:
> No, they're not solved. #2542 is added some functions to simplify working with `BenchmarkResults` and not address these issues
>
So, will still want these CLI Tool, right?
And using those updated functionality from #2542, we want to simplify these CLI tool correct?
Should I continue in this PR only, or close and create new one?

Samoed:
> So, will still want these CLI Tool, right?

Yes

> And using those updated functionality from https://github.com/embeddings-benchmark/mteb/pull/2542, we want to simplify these CLI tool correct?

Yes

> Should I continue in this PR only, or close and create new one?

You can continue this PR

KennethEnevoldsen:
Yes as @Samoed says, the PR that I made only adds a lot of dataframe utilities required for this PR.

You would have to refactor this PR to rely on the dataframe utilities (to ensure that we consistently do it the same way)

KennethEnevoldsen:
@ayush1298 just pinging you here

KennethEnevoldsen:
@ayush1298 seems like we never finished up this PR, do you still have time to work on it?

PR #2465, opened 2025-03-31
test: cache datasets check to avoid failing ci 

sam-hey:
Enhance the dataset availability test by caching the result when the dataset is available. The test should be skipped if it has already been performed by another CI on the same day.

Example: 
https://github.com/embeddings-benchmark/mteb/actions/runs/14171665536/job/39696653523

### Code Quality

### Testing

### DIFF
--- a/.github/workflows/dataset_loading.yml
+++ b/.github/workflows/dataset_loading.yml
@@ -13,6 +13,13 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v3
 
+    - name: Cache Hugging Face
+      id: cache-hf
+      uses: actions/cache@v4
+      with:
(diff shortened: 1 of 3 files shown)


Samoed (commented):
I don't think we want this, because we won't be able to check if some datasets were removed (like currently https://huggingface.co/datasets/navjordj/SNL_summarization)

sam-hey:
It will check daily for any missing datasets, reducing the number of false positives caused by `failed with 429 Client Error: Too Many Requests for URL`. Additionally, newly added datasets will always be checked.

KennethEnevoldsen:
Hmm, it seems like this PR has gotten stale - will close it

PR #2439, opened 2025-03-26
add ops_moa_models

ahxgw:
### Code Quality

- [x] **Code Formatted**: Format the code using `make lint` to maintain consistent style.

### Documentation

- [ ] **Updated Documentation**: Add or update documentation to reflect the changes introduced in this PR.

### Testing

- [ ] **New Tests Added**: Write tests to cover new functionality. Validate with `make test-with-coverage`.
- [ ] **Tests Passed**: Run tests locally using `make test` or `make test-with-coverage` to ensure no existing functionality is broken.

### Adding datasets checklist

**Reason for dataset addition**: ... 

- [ ] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb -m {model_name} -t {task_name}` command.
  - [ ] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [ ] `intfloat/multilingual-e5-small`
- [ ] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [ ] If the dataset is too big (e.g. >2048 examples), considering using `self.stratified_subsampling() under dataset_transform()`
- [ ] I have filled out the metadata object in the dataset file (find documentation on it [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#2-creating-the-metadata-object)).
- [ ] Run tests locally to make sure nothing is broken using `make test`.
- [ ] Run the formatter to format the code using `make lint`.

### Adding a model checklist

 - [x] I have filled out the ModelMeta object to the extent possible
 - [x] I have ensured that my model can be loaded using
   - [x] `mteb.get_model(model_name, revision)` and
   - [x] `mteb.get_model_meta(model_name, revision)`
 - [x] I have tested the implementation works on a representative set of tasks.

PR #2467, opened 2025-03-31
fix: CacheWrapper per task

flogrammer:
### Code Quality

### Documentation

### Testing

### DIFF
--- a/docs/usage/usage.md
+++ b/docs/usage/usage.md
@@ -438,14 +438,14 @@ mteb run -t NFCorpus -m all-MiniLM-L6-v2 --output_folder results --save_predicti
 
 ### Caching Embeddings To Re-Use Them
 
-There are times you may want to cache the embeddings so you can re-use them. This may be true if you have multiple query sets for the same corpus (e.g. Wikipedia) or are doing some optimization over the queries (e.g. prompting, other experiments). You can setup a cache by using a simple wrapper, which will save the cache per task in the `<path_to_cache_dir>/{task_name}` folder:
+There are times you may want to cache the embeddings so you can re-use them. This may be true if you have multiple query sets for the same corpus (e.g. Wikipedia) or are doing some optimization over the queries (e.g. prompting, other experiments). You can setup a cache by using a simple wrapper, which will save the cache per task in the `<path_to_cache_dir>/<task_name>` folder:
 
 ```python
 # define your task(s) and model above as normal
 ...
 # wrap the model with the cache wrapper
 from mteb.models.cache_wrapper import CachedEmbeddingWrapper
-model_with_cached_emb = CachedEmbeddingWrapper(model, tasks, cache_path='<path_to_cache_dir>')
+model_with_cached_emb = CachedEmbeddingWrapper(model, cache_path='<path_to_cache_dir>')
 # run as normal
 evaluation.run(model, ...)
 ```
--- a/docs/usage/usage.md
+++ b/docs/usage/usage.md
@@ -438,16 +438,32 @@ mteb run -t NFCorpus -m all-MiniLM-L6-v2 --output_folder results --save_predicti
 
 ### Caching Embeddings To Re-Use Them
 
-There are times you may want to cache the embeddings so you can re-use them. This may be true if you have multiple query sets for the same corpus (e.g. Wikipedia) or are doing some optimization over the queries (e.g. prompting, other experiments). You can setup a cache by using a simple wrapper, which will save the cache per task in the `cache_embeddings/{task_name}` folder:
+There are times you may want to cache the embeddings so you can re-use them. This may be true if you have multiple query sets for the same corpus (e.g. Wikipedia) or are doing some optimization over the queries (e.g. prompting, other experiments). You can setup a cache by using a simple wrapper, which will save the cache per task in the `<path_to_cache_dir>/{task_name}` folder:
 
 ```python
-# define your task and model above as normal
+# define your task(s) and model above as normal
 ...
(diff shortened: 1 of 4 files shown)


mteb/models/cache_wrapper.py
```diff
@@ -203,21 +203,26 @@ def close(self):
 
 
 class CachedEmbeddingWrapper(Wrapper, Encoder):
-    def __init__(self, model: Encoder, cache_path: str | Path):
+    def __init__(self, model: Encoder, tasks: list[str], cache_path: str | Path):
```
Samoed, line 206: Maybe add option to pass `AbsTask`?

Samoed: Yes, like `task.metadata.name`

Samoed:
@orionw Can you review this PR?

flogrammer:
Hi, thanks for the feedback. 

As you outlined, it is indeed possible to cache only specific tasks. The key change here is however to have a separate cache per task, which is currently not supported. Everything is cached into the same memmap. 

I wanted to do some analyses on the embeddings, but couldn't track back which embedding belongs to which task. Let me know if this makes sense.

Also, I wanted to push my reviewed changes, but somehow the linter complains about / fixes files I haven't touched:

```
ruff check . --fix --exit-non-zero-on-fix       # running ruff linting # --exit-non-zero-on-fix is used for the pre-commit hook to work
mteb/models/ops_moa_models.py:19:57: F821 Undefined name `np`
```

Any idea why this is happening?

Samoed:
Somehow ruff didn't showed this error during merge https://github.com/embeddings-benchmark/mteb/pull/2439. You can add this import to this file

mteb/models/cache_wrapper.py
KennethEnevoldsen: No need to specify the tasks in the init. Let us simply cache all tasks the model is run on.

mteb/models/cache_wrapper.py
```diff
@@ -203,21 +203,26 @@ def close(self):
...
         self.cache_path.mkdir(parents=True, exist_ok=True)
+        self.cache_dict = {}
+        self.tasks = tasks
 
         if hasattr(model, "encode"):
-            self.cache = TextVectorMap(self.cache_path / "cache")
-            self.cache.load(name="cache")
+            for task in tasks:
+                self.cache_dict[task] = TextVectorMap(self.cache_path / task)
+                self.cache_dict[task].load(name=task)
```
KennethEnevoldsen, line 216: I would create this in the encode

tests/test_embedding_caching.py
```diff
@@ -87,7 +87,7 @@ def test_other_functions_still_work(self, cache_dir):
         dummy_model = DummyModel()
 
         # Create the wrapper
-        wrapped_model = CachedEmbeddingWrapper(dummy_model, cache_dir)
+        wrapped_model = CachedEmbeddingWrapper(dummy_model, ["DummyTask"], cache_dir)
```
KennethEnevoldsen, line 90:
```suggestion
        wrapped_model = CachedEmbeddingWrapper(dummy_model, cache_dir)
```

KennethEnevoldsen (changes requested):
Thanks for the clarification. I think this is a reasonable change, but if that being the problem I have suggested a few changes. Notably removing task_names from the init.

mteb/models/cache_wrapper.py
flogrammer: Good idea. I adjusted the code accordingly.

mteb/models/ops_moa_models.py
```diff
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
-from mteb.model_meta import ModelMeta
-from mteb.models.wrapper import Wrapper
 from functools import partial
+
+import numpy as np
 from sentence_transformers import SentenceTransformer
-import torch
```
flogrammer, line 7: Not sure why this happened, but this was auto-fixed by  ruff

docs/usage/usage.md
```diff
@@ -438,18 +438,34 @@ mteb run -t NFCorpus -m all-MiniLM-L6-v2 --output_folder results --save_predicti
...
+from mteb.models.cache_wrapper import TextVectorMap
+
+# Access the memory-mapped file and convert to array
+vector_map = TextVectorMap("<path_to_cache_dir>/AppsRetrieval")
+vector_map.load(name="AppsRetrieval")
+vectors = np.asarray(vector_map.vectors)
+
+# Remove all "placeholders" in the embedding cache
+zero_mask = (vectors == 0).all(axis=1)
+vectors = vectors[~zero_mask]
```
KennethEnevoldsen, line 466: It is not entirely clear what the "placeholder" is here

flogrammer: The memmap is initalized with 100k zero-vectors and increased by another 100k whenever the cache is full. This happens in TextVectorMap in cache_wrapper.py.

KennethEnevoldsen: Ohh right. From a user perspective, it is a bit weird. Could we fix this in the `close()` call? (probably not needed for this PR, though)

KennethEnevoldsen (approved):
I think this is ready to merge - thanks for taking the time. I would be happy with another PR cleaning up the cache at the end of the PR, but I don't think we need to block this PR anymore

docs/usage/usage.md
flogrammer: That's a very good idea. Will open another PR for that :)

Issue #2230, opened 2025-03-03
Simplify UI


Muennighoff:
https://huggingface.co/spaces/mteb/leaderboard/discussions/158

I think there is a valid point being raised in the above that the UI is not as simple anymore? Maybe we can hide all the advanced filtering by default and instead have some quick presets like in https://huggingface.co/spaces/open-llm-leaderboard/open_llm_leaderboard#/ ?

KennethEnevoldsen:
The issue is also mirrored here: #2224

I see two sources of confusion here 1) the user might be expecting the English leaderboard 2) the voyage models do not have results. 

I see two solutions. I would do both:

1) **Simplify UI**:

I propose this:

<img width="951" alt="Image" src="https://github.com/user-attachments/assets/87b4393c-0fbf-4afc-9d25-d8cb696d4d8a" />


With it unfolded:
<img width="952" alt="Image" src="https://github.com/user-attachments/assets/5bd3dd01-3514-4e3e-b86b-f3e40e1fa59b" />

2) Make an overview leaderboard:

See #2038 

Let me know what you think

x-tabdeveloping:
I think this makes quite a bit of sense! Though we could choose a more descriptive name than `advanced filters` and maybe add a description so people understand what it means.

Perhaps something like **Customize Benchmark** or **Filter Task Selection**

ayush1298:
I have added 1 of 2 things which are suggested. And leaderboard looks like:

<img width="1479" alt="Image" src="https://github.com/user-attachments/assets/66516048-3539-45fa-9fda-c3356e92946a" />

I am working on 2nd one which is Make an overview leaderboard. For the above things, do we have any other suggestion? Should I try to do things mentioned in #2438?

x-tabdeveloping:
Hey @ayush1298 ! Thanks for the good work! I think hiding the advanced filters is a great idea and makes the UI way more slick :))

One thing we should definitely watch out for is not to hide the benchmark selection dropdown, since this is what most people will be reaching for.
Additionally, I think if we just put it right above the table we might make the filters a bit visually disconnected from what they actually control. We will have to figure out a smart way to organize the leaderboard if we hide so much of what was previously there.

Another concern of mine is that putting benchmark filters and model filters together might not make sense anymore.
[...]

PR #2469, opened 2025-03-31
UI Simplify

ayush1298:
This PR is related to simplifiaction of UI as suggested in issue #2230 .

https://github.com/user-attachments/assets/54ed1128-eefa-41d7-8614-2b067268de9f

### Code Quality

### Documentation

### Testing

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -298,19 +298,41 @@ with gr.Blocks(fill_width=True, theme=gr.themes.Base(), head=head) as demo:
     > Looking for the previous MTEB leaderboard? We have made it available [here](https://huggingface.co/spaces/mteb/leaderboard_legacy) but it will no longer be updated.
     """
     )
-
     with gr.Row():
-        with gr.Column(scale=5):
-            gr.Markdown(
-                """
+        with gr.Column():
+            benchmark_select.render()
+            description = gr.Markdown(
+                update_description,
+                inputs=[benchmark_select, lang_select, type_select, domain_select],
+            )
+            citation = gr.Markdown(update_citation, inputs=[benchmark_select])
+            with gr.Accordion("Share this benchmark:", open=False):
+                gr.Markdown(produce_benchmark_link, inputs=[benchmark_select])
+        with gr.Column():
+            with gr.Tab("Performance per Model Size"):
+                plot = gr.Plot(performance_size_plot, inputs=[summary_table])
+                gr.Markdown(
+                    "*We only display models that have been run on all tasks in the benchmark*"
+                )
+            with gr.Tab("Performance per Task Type (Radar Chart)"):
+                radar_plot = gr.Plot(radar_chart, inputs=[summary_table])
+                gr.Markdown(
+                    "*We only display models that have been run on all task types in the benchmark*"
+                )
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
Hi @ayush1298 thanks for this PR. 

I don't think the UI quite does it for me. I spent some time talking with @x-tabdeveloping about it yesterday, and we are afraid that this might lead to further confusion.

I took a look at the thread and I do really like the suggestion that you made here:
https://github.com/embeddings-benchmark/mteb/issues/2230#issuecomment-2764033044

However, I would move the benchmark selection out of the "advanced filters" box.

ayush1298:
> However, I would move the benchmark selection out of the "advanced filters" box.
>
@KennethEnevoldsen, I have made couple of changes after that. Benchmark selection was also removed from "advanced filters" box. Can you once have a look at the video in description of this PR.

KennethEnevoldsen:
I had a look at the video, but I strongly prefer the look in the linked version. The suggestion is too much of a departure from the current one with no clear upside gained from it.

ayush1298:
> I had a look at the video, but I strongly prefer the look in the linked version. 
>
Are you referring to this [#2230 (comment)](https://github.com/embeddings-benchmark/mteb/issues/2230#issuecomment-2764033044)

>The suggestion is too much of a departure from the current one with no clear upside gained from it.
>
I think gain was just to make it less clutter and more clean. As I think the components like benchmarks details, citation are less relevant, so we can avoid spaces by adding them under toggled open/close layout. 
Can you clarify what you have suggestions/expectations from simplify UI?

KennethEnevoldsen:
As also said previously, I discussed this with @x-tabdeveloping, who also agreed that this is not the direction we want to go.

1) There can be important information in the description, e.g. "see the newer version of this benchmark..." (so hiding that is not ideal)
2) figure should be linked to table (the suggested UI splits these up (the linked suggestion also does this, I would have advanced filters above - with the benchmark selection outside of that and on the top)
3) filter in sidebar is not great (it is not a standard design element so it breaks user expectations)


Good elements:
- Acknowledgement should probably be moved to fold down

ayush1298:
> 1. There can be important information in the description, e.g. "see the newer version of this benchmark..." (so hiding that is not ideal)
>
Got your point, will do this. 
> 2. figure should be linked to table (the suggested UI splits these up (the linked suggestion also does this, I would have advanced filters above - with the benchmark selection outside of that and on the top)
>
I didn't get this exactly.  Are you saying that to put in this order. 
| Benchmark selection | Figure | 
| Advance Filters |  | 

> 3. filter in sidebar is not great (it is not a standard design element so it breaks user expectations)
> 
So, are you suggesting to completely remove sidebars then? Or having model filters there, which are generally less usable as suggested [here](https://github.com/embeddings-benchmark/mteb/issues/2230#issuecomment-2765932363)

@KennethEnevoldsen Once, I got answer to this, I will share the UI with this changes, and then can you can have a look to suggest further changes.

KennethEnevoldsen:
Talked with @x-tabdeveloping, and he is currently reworking the UI. To prevent double work and since Marton has more experience with the leaderboard and its design I will close this PR, but @ayush1298 would love your review on the upcoming PR.

Issue #2476, opened 2025-04-02
[v2] Define encoding kwargs


Samoed:
Currently in code a lot of checks for `batch_size` like
https://github.com/embeddings-benchmark/mteb/blob/04d0cf423d178273b881fce12a500f2fde507c4a/mteb/evaluation/evaluators/Image/Any2AnyMultiChoiceEvaluator.py#L51-L52

Maybe we can define base `batch_size` in cli or python interface.

Currently, it is not possible to pass any encoding kwargs other than "batch_size" using the CLI, for example, to enable/disable the progress bar.

PR #2495, opened 2025-04-05
[v2] set default batch_size

Samoed:
closes https://github.com/embeddings-benchmark/mteb/issues/2476

Set default batch_size `MTEB` class and removed from other places

### Code Quality

### Documentation

### Testing

### Adding datasets checklist

### Adding a model checklist

### DIFF
--- a/mteb/evaluation/MTEB.py
+++ b/mteb/evaluation/MTEB.py
@@ -250,7 +250,7 @@ class MTEB:
         overwrite_results: bool = False,
         raise_error: bool = True,
         co2_tracker: bool = True,
-        encode_kwargs: dict[str, Any] = {},
+        encode_kwargs: dict[str, Any] | None = None,
         **kwargs,
     ) -> list[TaskResult]:
         """Run the evaluation pipeline on the selected tasks.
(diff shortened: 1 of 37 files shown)


mteb/evaluation/MTEB.py
```diff
@@ -277,6 +277,11 @@ def run(
         """
         # update logging to account for different levels of Verbosity (similar to the command line)
 
+        if encode_kwargs is None:
+            encode_kwargs = {}
+
+        encode_kwargs["batch_size"] = encode_kwargs.get("batch_size", 32)
```
KennethEnevoldsen, line 283: Add logging message

KennethEnevoldsen (commented):
I would even consider not adding default batch size anywhere?

Samoed:
We need batch_size to make dataloaders for image tasks to work correctly. I don't know solution how we can avoid it

KennethEnevoldsen:
Hmm, maybe we should then only set the default at the very end? (just before the dataloader?)

Otherwise user might get wierd errors if they try to call a function further up the stack?

Samoed:
> just before the dataloader

I don't like it, because it would requrie changes in a lot of places. I think this is simiplier solution

> Otherwise user might get wierd errors if they try to call a function further up the stack?

Can you give example?

Samoed:
Do we excpecting that users can call evaluators directly?

KennethEnevoldsen:
Yea so this is why I made #2515. I am fine with assuming that they can't

PR #2499, opened 2025-04-06
fix: validate lang code in ModelMeta

### LINKED ISSUE #2013: Validate language codes for ModelMeta Languages
Currently, some models (e.g. BGE) specify the language code as "eng_Latn" instead of "eng-Latn" as specified in the tasks. We should ensure that this is consistent.

ayush1298:
closes #2013 

### Code Quality

### Documentation

### Testing


ayush1298:
@isaac-chung Can you please check this? On doing `make test` all tests are passing, but I don't know what exactly will be way to check that this fix the issue

mteb/model_meta.py
```diff
@@ -123,6 +126,19 @@ def to_dict(self):
...
+        if self.languages is not None:
+            self.languages_are_valid(self.languages)
+
+    def languages_are_valid(self, languages: list[ISO_LANGUAGE_SCRIPT] | None) -> None:
+        """This method checks that each language is valid."""
+        if languages is None:
+            return
+
+        for code in languages:
+            TaskMetadata._check_language_code(code)
```
Samoed, line 140: I'm not sure if we want to use TaskMetadata here. Maybe tonmove to separate function?

isaac-chung:
I don't think the validation is done anywhere. See the `check_name` function that uses `@field_validator("name")`. Let's add a similar function but for the `languages` parameter.

mteb/model_meta.py
isaac-chung: Might be a good idea to put that separate function in `mteb/languages.py`

ayush1298: Should I add _check_language_code as a separate function on under LanguageScripts class in `mteb/languages.py`?

ayush1298:
@isaac-chung On adding Field Validator for `languages` similar to `name`,  `make test` is failing and giving error at many places, where `eng_Latn` is defined instead of `eng-Latn`. So, should I correct all these manually ?

isaac-chung:
> @isaac-chung On adding Field Validator for `languages` similar to `name`,  `make test` is failing and giving error at many places, where `eng_Latn` is defined instead of `eng-Latn`. So, should I correct all these manually ?

Yes please. This means the check is working as intended :)

ayush1298:
@isaac-chung, I have corrected lang-code for all models and `make test` is also passing all tests without any errors

isaac-chung (approved):
Nice! Thanks for working on this.

PR #2501, opened 2025-04-06
ci: refactor TaskMetadata eval langs test

### LINKED ISSUE #2500: TaskMetadata eval_langs validation robustness
1. Currently, a TaskMetadata with `eval_langs=["eng_Latn"]` will pass `tests/test_TaskMetadata.py::test_all_metadata_is_filled_and_valid` (when it should fail). The test should be refactored such that this case can be caught
2. Since we currently check lang and script with this line: `lang, script = code.split("-")`, we should likely check if `"-"` is in the code to avoid the uncaught error.
3. Consider using a field_validator instead of running this in pytest? (Cannot when we still have HISTORIC_DATASETS)

isaac-chung:
Fixes #2500

- Currently, a TaskMetadata with eval_langs=["eng_Latn"] will pass tests/test_TaskMetadata.py::test_all_metadata_is_filled_and_valid (when it should fail)
- unnest the if statements in test_all_metadata_is_filled_and_valid
- add check if "-" is in the code to avoid the uncaught error.
- skip AbsTaskAggregate instead of manually adding each instance in `_HISTORIC_DATASETS`

### Code Quality

### Documentation

### Testing

### DIFF
--- a/tests/test_TaskMetadata.py
+++ b/tests/test_TaskMetadata.py
@@ -5,6 +5,7 @@ import logging
 import pytest
 
 from mteb import AbsTask
+from mteb.abstasks.aggregated_task import AbsTaskAggregate
 from mteb.abstasks.TaskMetadata import TaskMetadata
 from mteb.overview import get_tasks
 
@@ -179,10 +180,6 @@ _HISTORIC_DATASETS = [
     "TamilNewsClassification",
     "TenKGnadClusteringP2P.v2",
     "TenKGnadClusteringS2S.v2",
-    "SynPerChatbotConvSAClassification",
-    "CQADupstackRetrieval-Fa",
-    "VisualSTS17Eng",
-    "VisualSTS17Multilingual",
 ]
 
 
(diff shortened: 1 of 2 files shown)


tests/test_TaskMetadata.py
```diff
@@ -357,23 +354,56 @@ def test_filled_metadata_is_filled():
...
 
     unfilled_metadata = []
+    invalid_metadata = []
     for task in all_tasks:
         if (
-            task.metadata.name not in _HISTORIC_DATASETS
-            and task.metadata.name.replace("HardNegatives", "")
-            not in _HISTORIC_DATASETS
+            task.metadata.name in _HISTORIC_DATASETS
+            or task.metadata.name.replace("HardNegatives", "") in _HISTORIC_DATASETS
```
Samoed, line 392: Maybe we should add also hardbegatives datasets to list directly?

isaac-chung: Good call, added

Issue #2517, opened 2025-04-07
RTEB in MTEB


fzliu:
Creating a checklist to help track the RTEB migration into MTEB. (This is tentative and will likely require changes as we progress.)

- [ ] Migrate code from RTEB into MTEB
- [ ] Deprecate the use of RTEB-specific `model_meta`; use MTEB instead
- [ ] Create hook for RTEB evaluator inside MTEB repository
- [ ] Migrate RTEB results into https://github.com/embeddings-benchmark/results

Some tasks for the leaderboard:

- [ ] Point the RTEB leaderboard updator to use the MTEB results repo
- [ ] Embed the RTEB leaderboard as an iframe within MTEB

Samoed:
What is `RTEB`? Is this about this repo https://github.com/embedding-benchmark/rteb?

isaac-chung:
@Samoed yes. There's a plan to integrate RTEB into MTEB, which would leverage its UI's simplicity and optimized code for retrieval. Would love your help during this integration process as well 🤗

Samoed:
I don't think we'll be making many more changes to the retrieval tasks beyond what's in https://github.com/embeddings-benchmark/mteb/pull/2496. Could someone create a list of datasets from that repo? I quickly looked through it, and it seems like we already have most of the tasks and all the models.

KennethEnevoldsen:
Models: Yep

Datasets: We should take a look at the ones not implemented. It might be easier to implement them in MTEB rather than wrapping the evaluator (thanks for pointing this out @Samoed)

PR #2529, opened 2025-04-10
[WIP] Add RTEB retrieval code

fzliu:
Related: #2517 

### Code Quality

- [ ] **Code Formatted**: Format the code using `make lint` to maintain consistent style.

### Documentation

### Testing


Samoed (changes requested):
Please, don't add directly rteb repository here. You should integrate your datasets into mteb. We standartized loading for retrieval tasks in `v2` branch

fzliu:
Still WIP but I think this is ready for a first pass.

@Samoed @KennethEnevoldsen to your earlier points, we changed the format of all the datasets to match those of existing retrieval datasets; however, the task is still separate since we're tracking multiple metrics and the format of the output results is a bit different.

We can merge it into the broader retrieval task down the road.

mteb/model_meta.py
```diff
@@ -263,6 +297,15 @@ def calculate_memory_usage_mb(self) -> int | None:
         model_memory_mb = model_memory_bytes / MB
         return round(model_memory_mb)
 
+    @property
+    def _id(self) -> str:
```
Samoed, line 301: Why do you need this?

mteb/models/voyage_models.py
```diff
@@ -368,6 +368,31 @@ def _batched_encode(
...
+    name="voyageai/voyage-3-large",  # Use the identifier the user provided
+    revision="1",  # Assuming revision 1
+    release_date="2024-09-18",  # Assuming same release as voyage-3
+    languages=None,
+    loader=partial(  # type: ignore
+        VoyageWrapper,
+        model_name="voyage-3-large",  # Match the API model name
+        model_prompts=model_prompts,
+    ),
+    max_tokens=32000,  # Assuming same as voyage-3
```
Samoed, line 381:
```suggestion
    max_tokens=32768,
```

mteb/models/voyage_models.py
```diff
@@ -368,6 +368,31 @@ def _batched_encode(
...
+    embed_dim=1024,  # Assuming same as voyage-3
```
Samoed, line 382: Can you check?

mteb/abstasks/AbsTaskRTEB.py
```diff
@@ -0,0 +1,1110 @@
...
+            model,
+            task_name=self.metadata.name,
+            model_name=model_name,
+            save_embds=save_embds_flag,
+            load_embds=load_embds_flag,
+            batch_size=batch_size,
+        )
+        rteb_encoder._trainer = trainer
+
+        args = argparse.Namespace(
```
Samoed, line 717: Why do you use `argparse.Namespace`?

mteb/abstasks/AbsTaskRTEB.py
```diff
@@ -0,0 +1,1110 @@
...
+            "accelerator": kwargs.get("accelerator", "auto"),
+            "devices": kwargs.get("devices", "auto"),
+            "num_nodes": kwargs.get("num_nodes", 1),
+            "strategy": kwargs.get("strategy", "auto"),
+            "precision": kwargs.get("precision", "32-true"),
+            "logger": False,  # Disable default logger
+            "enable_checkpointing": False,
+            "enable_progress_bar": True,
+        }
+        trainer = pl.Trainer(**trainer_kwargs)
```
Samoed, line 702: Why do you need Trainer if you're only encoding?

mteb/abstasks/AbsTaskRTEB.py
```diff
@@ -0,0 +1,1110 @@
...
+        Warning:
+            This method is deprecated and will be removed in future versions.
+            Use RTEBTaskRunner.run_rteb_evaluation for evaluation logic.
+
+        Delegates to the parent class implementation while issuing a deprecation warning.
+        """
+        import warnings
+
+        warnings.warn(
+            "_evaluate_subset is deprecated for RTEB tasks. Use RTEBTaskRunner.run_rteb_evaluation instead.",
```
Samoed, line 1082: Why?

Samoed (changes requested):
Your implementation is quite different from how other tasks are handled. I checked a few datasets and couldn't find `zenodo/4063986` used in `RTEBAILACasedocsTask`, and `lavita/ChatDoctor-HealthCareMagic-100k` doesn't load with your current loader.

If you need just to add scores to Retrieval to correctly work then, you can use this https://github.com/embeddings-benchmark/mteb/pull/2600 for that (this is `v2` branch)

KennethEnevoldsen:
Hi @fzliu. This PR has currently introduced:

1) Multiple tasks 
2) A new abstask (the core) 
3) A model
4) an aggregated task (not needed, should be a benchmark)

I would actually suggest closing this PR and opening two new ones:

1) with the model
2) with the RTEB abstask and one sample task

This allows us to focus on the main points. On those main points:

> the task is still separate since we're tracking multiple metrics 

What metrics are different, why do we add them

> and the format of the output results is a bit different.

I prefer keeping the current results format. This works with the current leaderboard, review pipeline, utility functions etc.

Changing the result format would require some justification.

PR #2531, opened 2025-04-10
[mieb] fix hatefulmemes

### LINKED ISSUE #2510: Problems with HatefulMemesRetrieval
1. The dataset has many duplicates e.g. https://huggingface.co/datasets/Ahren09/MMSoc_HatefulMemes/viewer/default/train?q=meanwhile+at+the+isis+strip+club which are currently not accounted for so we waste compute encoding the same images and texts I think; Can we just dedup based on the exact images? (else there's also code at https://github.com/Muennighoff/vilio/blob/50eb7cc9c901795c394070f7705fc0c5a7667bd2/utils/pandas_scripts.py#L61 using image hashes). We could reupload a clean version under mteb/ on HF
2. IIUC the [current code assumes that there is one match per image or text](https://github.com/embeddings-benchmark/mteb/blob/77bef06186c365ee6945fe234663439c3c984f77/mteb/tasks/Image/Any2AnyRetrieval/eng/HatefulMemesT2IRetrieval.py#L56) but this is not the case because of (a) duplicates as mentioned above and (b) HatefulMemes contains the same caption and the same image multiple times but with different image/caption combinations. This adversarial construction is one of the main selling points of the difficulty of the dataset. Below is a simple example of the same caption with two different images (from [here](https://arxiv.org/abs/2012.07788))

Image

So I guess two options are:
a - Dedup all captions only (this will leave some duplicate images but with different captions but should remove all images which are complete dups)
b - Have multiple labels for all images where the caption appears (in this case probably worth dedupping duplicate images)

gowitheflow-1998:
deduplicate both tasks for HatefulMemes. Fix #2510.

### Code Quality

### Testing

### DIFF
--- a/mteb/tasks/Image/Any2AnyRetrieval/eng/HatefulMemesI2TRetrieval.py
+++ b/mteb/tasks/Image/Any2AnyRetrieval/eng/HatefulMemesI2TRetrieval.py
@@ -4,6 +4,7 @@ from datasets import concatenate_datasets, load_dataset
 
 from mteb.abstasks.Image.AbsTaskAny2AnyRetrieval import AbsTaskAny2AnyRetrieval
 from mteb.abstasks.TaskMetadata import TaskMetadata
+import pandas as pd
 
 
 def _load_data(path: str, splits: str, cache_dir: str = None, revision: str = None):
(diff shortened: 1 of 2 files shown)


gowitheflow-1998:
for `clip-vit-base-patch32`, `I2T`: `0.433` -> `0.393`; `T2I`: `0.465` -> `0.429`.
the corpus (pooled across all splits) is 10k -> ~8.2k after deduplication, we'll need to rerun but looks rank-preserving to me.

isaac-chung (approved):
Nice! Added what we discussed re: task description and polars.

PR #2536, opened 2025-04-11
fix: Update mteb.get_tasks with an exclude_aggregate parameter to exclude aggregate tasks

### LINKED ISSUE #2176: Exclude aggregated tasks from `get_task`
> hmm, what a hacky way of using the aggregate task ;)
> 
> Like it though. We could also do this for the BRIGHT tasks (on the old leaderboard the split appears as different tasks)
> 
> Should we by default exclude aggregate tasks for `mteb.get_tasks()`?

Thanks! 

And I don't have a strong preference. I guess if they are excluded by default, a parameter would added to "enable" them? e.g. `include_aggregate=True`

_Originally posted by @isaac-chung in https://github.com/embeddings-benchmark/results/issues/133#issuecomment-2684964625_

Sid-MB:
Resolves #2176

### Code Quality

### Documentation

### Testing

### DIFF
--- a/mteb/overview.py
+++ b/mteb/overview.py
@@ -133,6 +133,15 @@ def filter_tasks_by_modalities(
         return [t for t in tasks if _modalities.intersection(t.modalities)]
 
 
+def filter_aggregate_tasks(
+    tasks: list[AbsTask], include_aggregate: bool = True
+) -> list[AbsTask]:
+    """Filters tasks by whether they are aggregate or not."""
+    if include_aggregate:
+        return tasks
+    return [t for t in tasks if not t.is_aggregate]
+
+
 class MTEBTasks(tuple):
     def __repr__(self) -> str:
         return "MTEBTasks" + super().__repr__()
@@ -278,6 +287,7 @@ def get_tasks(
     exclusive_language_filter: bool = False,
     modalities: list[MODALITIES] | None = None,
     exclusive_modality_filter: bool = False,
+    include_aggregate: bool = True,
 ) -> MTEBTasks:
     """Get a list of tasks based on the specified filters.
 
@@ -300,6 +310,7 @@ def get_tasks(
         exclusive_modality_filter: If True, only keep tasks where _all_ filter modalities are included in the
             task's modalities and ALL task modalities are in filter modalities (exact match).
             If False, keep tasks if _any_ of the task's modalities match the filter modalities.
+        include_aggregate: If True, include aggregate tasks. If False, aggregate tasks are filtered out.
 
     Returns:
         A list of all initialized tasks objects which pass all of the filters (AND operation).
@@ -350,6 +361,8 @@ def get_tasks(
         _tasks = filter_tasks_by_modalities(
             _tasks, modalities, exclusive_modality_filter
         )
+    if include_aggregate:
+        _tasks = filter_aggregate_tasks(_tasks, include_aggregate)
 
     return MTEBTasks(_tasks)
 
--- a/mteb/overview.py
+++ b/mteb/overview.py
@@ -134,12 +134,15 @@ def filter_tasks_by_modalities(
 
 
 def filter_aggregate_tasks(
-    tasks: list[AbsTask], include_aggregate: bool = True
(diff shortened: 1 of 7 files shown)


Sid-MB:
@KennethEnevoldsen @Samoed let me know if this functions how you want. 
I used exclude_aggregate instead of include_aggregate because it matches the format of other `get_tasks()` parameters, like `exclude_superseded`, and the logic is simpler. I can change it to `include_aggregate` if it makes more sense, though.

mteb/overview.py
```diff
@@ -133,6 +133,18 @@ def filter_tasks_by_modalities(
...
+def filter_aggregate_tasks(
+    tasks: list[AbsTask], exclude_aggregate: bool = True
+) -> list[AbsTask]:
+    """Args:
+    tasks: A list of tasks to filter.
+    exclude_aggregate: Is True, filter out aggregate tasks. If False, return all the original tasks.
+    """
+    if exclude_aggregate:
+        return [t for t in tasks if not t.is_aggregate]
+    return tasks
```
isaac-chung, line 145: The `exclude_aggregate` flag has to be True before this function is called. Let's omit the same check and directly return the filtered list.

I see that filter_tasks_by_modalities should be similarly updated as well. It is separate and I will raise a new issue for that.

```suggestion
    return [t for t in tasks if not t.is_aggregate]
```

mteb/overview.py
```diff
@@ -278,6 +290,7 @@ def get_tasks(
     exclusive_language_filter: bool = False,
     modalities: list[MODALITIES] | None = None,
     exclusive_modality_filter: bool = False,
+    exclude_aggregate: bool = False,
```
isaac-chung, line 293: Wanted true by default
```suggestion
    exclude_aggregate: bool = True,
```

mteb/overview.py
```diff
@@ -350,6 +364,8 @@ def get_tasks(
         _tasks = filter_tasks_by_modalities(
             _tasks, modalities, exclusive_modality_filter
         )
+    if exclude_aggregate:
+        _tasks = filter_aggregate_tasks(_tasks, exclude_aggregate)
```
isaac-chung, line 368: Follow up on the comment above, we can omit passing the flag in.
```suggestion
    if exclude_aggregate:
        _tasks = filter_aggregate_tasks(_tasks)
```

tests/test_TaskMetadata.py
```diff
@@ -1161,3 +1161,10 @@ def test_eval_langs_correctly_specified(task: AbsTask):
         assert isinstance(task.metadata.eval_langs, list), (
             f"{task.metadata.name} should have eval_langs as a list"
         )
+
+
+def test_is_aggregate_property_correct():
```
isaac-chung, line 1166: Nice! Maybe this test fits better in https://github.com/embeddings-benchmark/mteb/blob/main/tests%2Ftest_tasks%2Ftest_all_abstasks.py, as it's not part of TaskMetadata?

isaac-chung (commented):
This is very clean, thank you! ☺️ I've only got a few suggestions. The rest looks good.

tests/test_TaskMetadata.py
Sid-MB: Moved!

mteb/overview.py
Sid-MB: Done! This may change the behavior of get_tasks for existing callers, as aggregate tasks will no longer be returned by default.

Sid-MB:
@isaac-chung Thank you for the comments! I made changes.

mteb/overview.py
isaac-chung: You're right. I misread the issue comment - we do want default to include all tasks.

isaac-chung: Let's revert back to False by default then. Thanks!

Sid-MB: Reverted!

Sid-MB:
@isaac-chung made exclude_aggregate false by default again. Lmk if there are any other changes needed

isaac-chung (approved):
Thanks so much!

PR #2537, opened 2025-04-11
[v2] introduce AbsAnyClassification and refactor mieb classification

Samoed:
I've created `AbsAnyClassification` for classification tasks, by merging `AbsTaskClassification` and `AbsTaskImageClassification`. Also updated evaluator to support both modalities

Closes https://github.com/embeddings-benchmark/mteb/issues/2432

Results for `openai/clip-vit-base-patch16`. There is a small mismatch, because I've changed seeds a bit.
| task_name   |   main_score (PR) |   main_score (main) |   eval_time (main) |
|:------------|------------------:|--------------------:|-------------------:|
| CIFAR10     |           0.92078 |             0.91886 |            151.892 |
| MNIST       |           0.92706 |             0.9183  |            143.583 |

Results for  `minishlab/potion-base-2M`
| task_name                          |   main_score (PR) |   main_score (main) |
|:-----------------------------------|------------------:|--------------------:|
| AmazonCounterfactualClassification |          0.643178 |            0.643178 |
| Banking77Classification            |          0.652857 |            0.652857 |
| EmotionClassification              |          0.40185  |            0.40185  |
| ImdbClassification                 |          0.715456 |            0.715456 |
| ToxicConversationsClassification   |          0.650537 |            0.650537 |

### Code Quality

### Documentation

### Testing

### Adding datasets checklist

### Adding a model checklist


isaac-chung (changes requested):
Thanks for the effort. However, this is not what we discussed in https://github.com/embeddings-benchmark/mteb/issues/2078#issuecomment-2667833913. I'd prefer not to have _more_ layers of abstraction in a refactor, to make it easier to maintain and newcomers to understand. I believe @KennethEnevoldsen [shared the same view as well](https://github.com/embeddings-benchmark/mteb/discussions/1950#discussioncomment-12120867).

Samoed:
But you're also said `In my head, e.g. a common AbsTaskAnyClassification could support both text and image columns, AND any other modalities that we would introduce, like audio.`. I can merge `AbsTaskImageClassification` with `AbsTaskClassification` and left only one `AbsTaskAnyClassifiction`

isaac-chung:
Good find. Maybe we can go the following,
- for the "Any" default, follow that of the text tasks (as there are a lot of them)
- for the MIEB classification tasks, we patch at the task level to use n_experiments: int = 5 and samples_per_label: int = 16

Samoed:
@isaac-chung I've changed to `AbsAnyClassification` task

isaac-chung (commented):
Amazing work so far, thanks for putting this together. Got a few questions. The rest looks good 🚀

mteb/abstasks/AbsTaskMultilabelClassification.py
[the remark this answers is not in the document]
Samoed: Yes, but I thought it would be better to merge MultilabelClassification tasks after this PR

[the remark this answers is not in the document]
isaac-chung: Okay. In that case, let's move these changes related to multilabel out of this PR.

tests/test_tasks/test_metadata.py
```diff
@@ -2,10 +2,18 @@
...
+    MockImageClassificationTask,
+)
 from tests.test_benchmark.task_grid import MOCK_TASK_TEST_GRID
 
 
-@pytest.mark.parametrize("task", MOCK_TASK_TEST_GRID)
+@pytest.mark.parametrize(
+    "task",
+    MOCK_TASK_TEST_GRID
+    + [MockImageClassificationTask(), MockImageClassificationKNNTask()],
```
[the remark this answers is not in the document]
Samoed: Added all meib tasks

mteb/abstasks/AbsTaskMultilabelClassification.py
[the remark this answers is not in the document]
Samoed: I've changed just inherited class name since `AbsTaskClassification` was renamed. I haven't changed anything else, and I can't move out this change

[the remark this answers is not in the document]
isaac-chung: I see, then let's leave it as is.

isaac-chung (approved):
Looks good! I feel it's good to merge. @KennethEnevoldsen feel free to still review post-hoc

mteb/abstasks/AbsTaskAnyClassification.py
```diff
@@ -35,27 +38,48 @@ class ClassificationDescriptiveStatistics(DescriptiveStatistics):
...
     min_labels_per_text: int
     average_label_per_text: float
     max_labels_per_text: int
+
     unique_labels: int
     labels: dict[str, dict[str, int]]
 
 
-class AbsTaskClassification(AbsTask):
+class AbsTaskAnyClassification(AbsTask):
```
KennethEnevoldsen, line 82: For now it is fine to keep it as is, but I would probably prefer that we do the following:

AbsTaskClassification # all modalities
AbsTaskTextClassification # only text

But it is a minor preference

mteb/abstasks/AbsTaskAnyClassification.py
```diff
@@ -69,12 +93,14 @@ class AbsTaskClassification(AbsTask):
...
-    abstask_prompt = "Classify user passages."
+    evaluator: type[ClassificationEvaluator] = logRegClassificationEvaluator
     samples_per_label: int = 8
     n_experiments: int = 10
     k: int = 3
-    train_split = "train"
+    train_split: str = "train"
+    label_column_name: str = "label"
+    input_column_name: str = "text"
+    abstask_prompt = "Classify user passages."
```
KennethEnevoldsen, line 103: should this be changed for images?

mteb/abstasks/AbsTaskAnyClassification.py
```diff
@@ -230,18 +280,32 @@ def _calculate_metrics_from_split(
...
+            min_image_width=min(img_widths) if img_widths else None,
+            average_image_width=sum(img_widths) / len(img_widths)
+            if img_widths
+            else None,
+            max_image_width=max(img_widths) if img_widths else None,
+            min_image_height=min(img_heights) if img_heights else None,
+            average_image_height=sum(img_heights) / len(img_heights)
+            if img_heights
+            else None,
+            max_image_height=max(img_heights) if img_heights else None,
```
KennethEnevoldsen, line 307: do we want to nest these?

mteb/evaluation/evaluators/ClassificationEvaluator.py
```diff
@@ -27,27 +31,86 @@ def dot_distance(a: np.ndarray, b: np.ndarray) -> float:
     return -np.dot(a, b)
 
 
-class kNNClassificationEvaluator(Evaluator):
+class ClassificationEvaluator(Evaluator, ABC):
```
KennethEnevoldsen, line 34: Hmm what the the difference between the two? can't we just merge them into one ClassificationEvaluator where we supply a clf on init?

KennethEnevoldsen (commented):
This looks great!

I few things, I would like to clarify, but overall I think it looks very good.

mteb/abstasks/AbsTaskAnyClassification.py
Samoed: I've change them for image directly in their tasks. Originally I had `AbsTaskClassification`, `AbsTaskImageClassificatin` and `AbsTaskTextClassification`, but @isaac-chung said that this is new layer to abstraction that we should avoid https://github.com/embeddings-benchmark/mteb/pull/2537#pullrequestreview-2762146235

mteb/evaluation/evaluators/ClassificationEvaluator.py
Samoed: Do you want something like this?
```python
class ClassificationEvaluator(Evaluator, ABC):
    def __init__(..., clf: Literal["knn", "logreg"]):
        if clf == "logreg":
           self.clf = logerg
    
    def __call__(...):
          return self.clf(...)
````
If so, I think the current method with `self.evaluator` provides support for customization of evaluator for different tasks without affecting other tasks

File: mteb/abstasks/AbsTaskAnyClassification.py (hunk 2 above)
Comment (Samoed) [reply mteb/abstasks/AbsTaskAnyClassification.py:82]: Originally, this was like this, but I have since changed it after review https://github.com/embeddings-benchmark/mteb/pull/2537#discussion_r2044511971

File: mteb/abstasks/AbsTaskAnyClassification.py (hunk 3 above)
Comment (KennethEnevoldsen) [reply mteb/abstasks/AbsTaskAnyClassification.py:103]: check - I think this is a good choice

File: mteb/evaluation/evaluators/ClassificationEvaluator.py (hunk 5 above)
Comment (KennethEnevoldsen) [reply mteb/evaluation/evaluators/ClassificationEvaluator.py:34]: What about?
```
class ClassificationEvaluator(Evaluator, ABC):
    def __init__(..., clf: SklearnEstimator = LogisticRegression): 
```
I would like to avoid the inheritance if possible

File: mteb/abstasks/AbsTaskAnyClassification.py (hunk 2 above)
Comment (Samoed) [reply mteb/abstasks/AbsTaskAnyClassification.py:82]: Do I need to rename it or change it somehow?

Comment (KennethEnevoldsen) [reply mteb/abstasks/AbsTaskAnyClassification.py:82]: No requirements yet, just wanted to align on direction. 

The rename is probably better done in a separate PR

File: mteb/evaluation/evaluators/ClassificationEvaluator.py (hunk 5 above)
Comment (Samoed) [reply mteb/evaluation/evaluators/ClassificationEvaluator.py:34]: Like this?
```python
class AbsClassifcation:
    self.evaluator = partial(ClassificationEvaluator, clf=LogisticRegression())
    
    def _evaluate_subset(...):
         evaluator = self.evaluator(dataset)
         results = evaluator(model)

class ClassificationEvaluator(Evaluator):
    def __init__(..., clf: SklearnEstimator = LogisticRegression): 
        self.clf = clf
    
   def __call__(...)
       scores = self.clf(...)
       return scores
```

This would work for LogisticRegression, but for KNN, we need to check multiple distance metrics. However, there are no tasks that use this, so I think we can remove this and leave support for any Sklearn scorer.

Samoed: @KennethEnevoldsen What do you think?

KennethEnevoldsen: I would be perfectly fine with this

KennethEnevoldsen: potentially:
```py
class AbsClassifcation:
    self.evalatution_clf = LogisticRegression
    self.evaluator = ClassificationEvaluator
    
    def _evaluate_subset(...):
         evaluator = self.evaluator(dataset, clf=self.evaluation_clf = LogisticRegression())
         results = evaluator(model)

class ClassificationEvaluator(Evaluator):
    def __init__(..., clf: SklearnEstimator = LogisticRegression): 
        self.clf = clf
    
   def __call__(...)
       scores = self.clf(...)
       return scores
```

If we want people to customize it more for their tasks, but I think what you suggest is fine as well.

Samoed: I've updated evaluator. Scores haven't changed

mteb/abstasks/AbsTaskAnyClassification.py
KennethEnevoldsen: like nest modality specific attributes for images and text.

Samoed: I've upated them

KennethEnevoldsen:
I am totally fine with merging this - tests seems to fail though,

PR #2542, opened 2025-04-14
feat: Added dataframe utilities to BenchmarkResults

KennethEnevoldsen:
- Added `get_results_table`. I was considering renaming it to `to_dataframe` to align with `tasks.to_dataframe`. WDYT?
- Added a tests for ModelResults and BenchmarksResults
- Added a few utility functions where needed
- Added docstring throughout ModelResults and BenchmarksResults
- Added todo comment for missing aspects - mostly v2 - but we join_revisions seems like it could use an update before then.

Prerequisite for #2454:

@ayush1298 can I ask you to review this PR as well? I hope this give an idea of what I was hinting at. Sorry that it took a while. I wanted to make sure to get it right.

### Code Quality

### Documentation

### Testing


Samoed (commented):
Great changes!

docs/usage/results.md
```diff
@@ -0,0 +1,123 @@
...
+### Filtering Results
+
+There is also utility function that allows you to select certain models or tasks:
+```py
+# select only gritLM
+results = results.select_models(["GritLM/GritLM-7B"])
+
+# select only retrieval tasks
+tasks = mteb.get_tasks(tasks=task_names)
+retrieval_tasks = [task for task in tasks if task.metadata.type == "Retrieval"]
```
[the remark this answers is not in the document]
KennethEnevoldsen: It doesn't work. get_tasks assumes that you specify a task in tasks that you want, so it doesn't filter the list. 

We could change that, but it is the current intended behaviour.

docs/usage/results.md
```diff
@@ -0,0 +1,123 @@
...
+results = results.select_tasks(retrieval_tasks)
+```
+
+### Creating a Dataframe
+
+```py
+df = results.get_results_table()
+
+print(df)
+# model_name                        task_name  GritLM/GritLM-7B
```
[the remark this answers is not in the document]
KennethEnevoldsen: It is the name of the columns. This is the correct print - though I get the confusion, I would rather make sure that it matches

KennethEnevoldsen:
Went through the changes:
- after conversation wth x-tabdeveloping I set `include_revisions=False` as default, which now call `.join_revisions()` beforehand
  - this also removed one of the TODO's from join revisions
- I also corrected the comment + renamed it to `to_dataframe()`

mteb/load_results/benchmark_results.py
```diff
@@ -207,9 +368,17 @@ def modalities(self) -> list[str]:
...
+        model_results: List of ModelResult objects.
+    """
+
     model_results: list[ModelResult]
-    model_config = ConfigDict(
-        protected_namespaces=(),
+    model_config = (
+        ConfigDict(  # to free up the name model_results which is otherwise protected
+            protected_namespaces=(),
+        )
```
Samoed, line 381: Since `pydantic v2.10` protected_namespaces protect only to `('model_validate', 'model_dump',)` do we still need this? https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.validate_default

Samoed (approved):
Great!

mteb/load_results/benchmark_results.py
```diff
@@ -206,7 +241,7 @@ def _get_score_for_table(self) -> list[dict[str, str | float]]:
 
         return scores_data
 
-    def get_results_table(
```
x-tabdeveloping: This is a breaking change. Can't we just deprecate this function instead of removing it?

x-tabdeveloping: Then it can internally call the new one

mteb/load_results/benchmark_results.py
x-tabdeveloping: I think it throws a warning if you don't have this in there

mteb/load_results/benchmark_results.py
Samoed: This function is not presented in main branch and was created and renamed in this PR

x-tabdeveloping: aaah okay I see

mteb/load_results/benchmark_results.py
KennethEnevoldsen: No warning:
Screenshot 2025-04-15 at 14 28 41

But this would force a version bump to 2.10.0, which I am not sure that we want to do.

PR #2549, opened 2025-04-15
UI Overhaul

x-tabdeveloping:
I reworked the MTEB UI using the latest new features in Gradio.
Here are some of the main changes:
 - Benchmark selection in side bar with benchmark categories and colorful icons for easier navigation.
 - Faster model search with table filtering, full screen tables and copying the table to the clipboard in CSV format.
 - Custom settings are now hidden by default but more accessible by giving them more space.
 - New look including a new font and color scheme.

Here's a video of the UI after my changes:

https://github.com/user-attachments/assets/0daf9e98-a14b-4c4e-8c37-377a78a573c5

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -156,7 +156,13 @@ def update_task_info(task_names: str) -> gr.DataFrame:
         }
     )
     df = df.drop(columns="reference")
-    return gr.DataFrame(df, datatype=["markdown"] + ["str"] * (len(df.columns) - 1))
+    return gr.DataFrame(
+        df,
+        datatype=["markdown"] + ["str"] * (len(df.columns) - 1),
+        show_copy_button=True,
+        show_fullscreen_button=True,
+        show_search="filter",
+    )
 
 
 # Model sizes in million parameters
@@ -390,17 +396,23 @@ def get_leaderboard_app() -> gr.Blocks:
                     update_description,
                     inputs=[benchmark_select, lang_select, type_select, domain_select],
                 )
-                citation = gr.Markdown(update_citation, inputs=[benchmark_select])  # noqa: F841
+                citation = gr.Markdown(
+                    update_citation, inputs=[benchmark_select]
+                )  # noqa: F841
                 with gr.Accordion("Share this benchmark:", open=False):
                     gr.Markdown(produce_benchmark_link, inputs=[benchmark_select])
             with gr.Column():
                 with gr.Tab("Performance per Model Size"):
-                    plot = gr.Plot(performance_size_plot, inputs=[summary_table])  # noqa: F841
+                    plot = gr.Plot(
+                        performance_size_plot, inputs=[summary_table]
+                    )  # noqa: F841
                     gr.Markdown(
                         "*We only display models that have been run on all tasks in the benchmark*"
                     )
                 with gr.Tab("Performance per Task Type (Radar Chart)"):
-                    radar_plot = gr.Plot(radar_chart, inputs=[summary_table])  # noqa: F841
+                    radar_plot = gr.Plot(
+                        radar_chart, inputs=[summary_table]
+                    )  # noqa: F841
                     gr.Markdown(
                         "*We only display models that have been run on all task types in the benchmark*"
                     )
@@ -476,7 +488,9 @@ def get_leaderboard_app() -> gr.Blocks:
                 download_table, inputs=[per_task_table], outputs=[download_per_task]
             )
         with gr.Tab("Task information"):
-            task_info_table = gr.DataFrame(update_task_info, inputs=[task_select])  # noqa: F841
+            task_info_table = gr.DataFrame(
+                update_task_info, inputs=[task_select]
+            )  # noqa: F841
 
         # This sets the benchmark from the URL query parameters
         demo.load(set_benchmark_on_load, inputs=[], outputs=[benchmark_select])
@@ -557,11 +571,7 @@ def get_leaderboard_app() -> gr.Blocks:
 
         @cachetools.cached(
             cache={},
-            key=lambda benchmark_name,
-            type_select,
-            domain_select,
-            lang_select,
-            modality_select: hash(
+            key=lambda benchmark_name, type_select, domain_select, lang_select, modality_select: hash(
                 (
                     hash(benchmark_name),
                     hash(tuple(type_select)),
@@ -637,13 +647,7 @@ def get_leaderboard_app() -> gr.Blocks:
 
         @cachetools.cached(
             cache={},
-            key=lambda scores,
-            tasks,
-            availability,
-            compatibility,
-            instructions,
-            model_size,
-            zero_shot: hash(
+            key=lambda scores, tasks, availability, compatibility, instructions, model_size, zero_shot: hash(
                 (
                     id(scores),
                     hash(tuple(tasks)),
@@ -776,11 +780,7 @@ def get_leaderboard_app() -> gr.Blocks:
 
         @cachetools.cached(
             cache={},
-            key=lambda scores,
-            search_query,
-            tasks,
-            models_to_keep,
-            benchmark_name: hash(
+            key=lambda scores, search_query, tasks, models_to_keep, benchmark_name: hash(
                 (
                     id(scores),
                     hash(search_query),
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -20,6 +20,8 @@ from mteb.abstasks.TaskMetadata import TASK_DOMAIN, TASK_TYPE
 from mteb.benchmarks.benchmarks import MTEB_multilingual
 from mteb.custom_validators import MODALITIES
 from mteb.languages import ISO_TO_LANGUAGE
+from mteb.leaderboard.benchmark_selector import (BENCHMARK_ENTRIES,
+                                                 make_selector)
 from mteb.leaderboard.figures import performance_size_plot, radar_chart
 from mteb.leaderboard.table import create_tables
 
@@ -241,13 +243,6 @@ def get_leaderboard_app() -> gr.Blocks:
     summary_table, per_task_table = create_tables(
         [entry for entry in default_scores if entry["model_name"] in filtered_models]
     )
-
-    benchmark_select = gr.Dropdown(
-        [bench.name for bench in benchmarks],
-        value=default_benchmark.name,
-        label="Prebuilt Benchmarks",
-        info="Select one of our expert-selected benchmarks from MTEB publications.",
-    )
     lang_select = gr.Dropdown(
         ISO_TO_LANGUAGE,
         value=sorted(default_results.languages),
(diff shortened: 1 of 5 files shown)


x-tabdeveloping:
I'm also eager to hear your opinion @ayush1298

Muennighoff (commented):
It looks so cool 🤯

mteb/leaderboard/table.py
```diff
@@ -278,22 +292,38 @@ def apply_styling(
...
     ).highlight_max(subset=task_score_columns, props="font-weight: bold")
-    for col in task_score_columns:
-        if col != "Model":
-            mask = per_task[col].notna()
-            per_task_style = per_task_style.background_gradient(
-                cmap=light_green_cmap,
-                subset=pd.IndexSlice[mask, col],
-                gmap=per_task[col].loc[mask],
-            )
+    # TODO: uncomment this when Gradio fixes it
```
isaac-chung, line 295: Could add an issue number/link?

mteb/leaderboard/table.py
```diff
@@ -237,7 +251,7 @@ def apply_styling(
     ]
     light_green_cmap = create_light_green_cmap()
     numeric_data = joint_table.copy()
-    numeric_data["Zero-shot"] = numeric_data["Zero-shot"].replace(-1, np.nan)
+    # numeric_data["Zero-shot"] = numeric_data["Zero-shot"].replace(-1, np.nan)
```
isaac-chung, line 254: Remove commented out code?

mteb/leaderboard/app.py
```diff
@@ -512,7 +524,7 @@ def on_benchmark_select(benchmark_name):
             elapsed = time.time() - start_time
             benchmark_results = all_benchmark_results[benchmark_name]
             scores = benchmark_results.get_scores(format="long")
-            logger.info(f"on_benchmark_select callback: {elapsed}s")
+            print(f"on_benchmark_select callback: {elapsed}s")
```
isaac-chung, line 527: Are these intentional? Could we use e.g. logger.debug?

mteb/leaderboard/table.py
```diff
@@ -278,22 +292,38 @@ def apply_styling(
...
+    #     if col != "Model":
+    #         mask = per_task[col].notna()
+    #         per_task_style = per_task_style.background_gradient(
+    #             cmap=light_green_cmap,
+    #             subset=pd.IndexSlice[mask, col],
+    #             gmap=per_task[col].loc[mask],
+    #         )
+    column_widths = get_column_widths(joint_table_style.data)
+    column_widths[0] = "100px"
+    column_widths[1] = "250px"
```
isaac-chung, line 306: Are these for PC screens? Could these be relative to accommodate for different devices/screen sizes?

isaac-chung (commented):
This looks very cool! Just got a few small, non-design-related questions.

mteb/leaderboard/benchmark_selector.py
```diff
@@ -0,0 +1,254 @@
...
+    ("First Benchmark", dict(value="MTEB(something)", icon="icon_url")),
+    ("Group of Benchmarks", 
+        [
+            ("Second Benchmark", dict(value="MTEB(something)", icon="icon_url")),
+            ("Third Benchmark", dict(value="MTEB(something)", icon="icon_url")),
+        ],
+    ),
+]
+"""
+BENCHMARK_ENTRIES = [
```
Samoed, line 17: Can we ensure that all benchmarks are included?

pyproject.toml
```diff
@@ -70,7 +70,7 @@ speedtask = [
 ]
 peft = ["peft>=0.11.0"]
 leaderboard = [
-    "gradio==5.16.0; python_version > '3.9'", # 3.10 is required for gradio
+    "gradio==5.17.1; python_version > '3.9'", # 3.10 is required for gradio
     "pydantic<2.11", # remove with gradio bump https://github.com/embeddings-benchmark/mteb/issues/2523
```
Samoed, line 74:
```suggestion
```

Samoed (commented):
Awesome changes!

mteb/leaderboard/table.py
x-tabdeveloping: Yes will do! There's already a fix on the way

mteb/leaderboard/table.py
x-tabdeveloping: Will remove

mteb/leaderboard/table.py
x-tabdeveloping: Contrary to popular belief, px in CSS does not actually correspond to pixels, and its size is highly dependent on the device and screen size. Here's [an article discussing this](https://www.w3.org/Style/Examples/007/units.en.html). This will actually probably look just fine on mobile devices too. Using percentages would probably not work since the columns need at least a certain amount of space.

mteb/leaderboard/benchmark_selector.py
x-tabdeveloping: Do we want that? I would imagine that once somebody adds a benchmark to MTEB, we would prefer if they sorted it into one of the categories in the leaderboard instead of landing in the Misc category.

I would love to hear opinions about this, but I'm not 100% certain that it's what we need.

x-tabdeveloping:
Looks even cooler in white if you ask me:
image

mteb/leaderboard/benchmark_selector.py
KennethEnevoldsen: We don't want all benchmarks displayed by default.

We should probably remove the current fix for this (I think this solution is better).

https://github.com/embeddings-benchmark/mteb/blob/3ff993d35d856c4ddf74532b2e156cf6f08852ba/mteb/benchmarks/benchmarks.py#L1133C5-L1133C34

mteb/leaderboard/app.py
```diff
@@ -104,7 +106,8 @@ def update_description(
     benchmark_name: str, languages: list[str], task_types: list[str], domains: list[str]
 ) -> str:
     benchmark = mteb.get_benchmark(benchmark_name)
-    description = f"## {benchmark.name}\n{benchmark.description}\n"
+    # description = f"## {benchmark.name}\n{benchmark.description}\n"
```
KennethEnevoldsen, line 108:
```suggestion
```

KennethEnevoldsen (approved):
A few minor comments on the code, but the new look is great.

x-tabdeveloping:
@KennethEnevoldsen 

1. Yes, it is folded out by default, I think it's good if it's immediately obvious to our users how to change between them.
2. I'm not sure I would remove the reference to the old leaderboard, some people will still be looking for it. And frankly, until we fix some annoying bugs, I don't think people will be completely satisfied with the new leaderboard.
3.  I think leaving share this benchmark folded out is okay, it takes up little space, and makes it more visually present. It would be a shame if people tried copying the URL from the browser bar and then discovered that it doesn't actually lead to the benchmark they want.
4. The radar chart is not shown by default, you have to click on the tab to show it. I have no idea what you mean by moving the description up.
5. MTEB(eng, v1) is listed in the language-specific benchmarks section as English, Legacy.

x-tabdeveloping:
Maybe we can move the reference to the legacy leaderboard to the bottom, that way it doesn't look like we are apologizing for having made a new one

mteb/leaderboard/benchmark_selector.py
```diff
@@ -0,0 +1,254 @@
...
+                dict(
+                    value="MTEB(fra, v1)",
+                    icon="https://github.com/lipis/flag-icons/raw/260c91531be024944c6514130c5defb2ebb02b7d/flags/4x3/fr.svg",
+                ),
+            ),
+            (
+                "Japanese",
+                dict(
+                    value="MTEB(jpn, v1)",
+                    icon="https://github.com/lipis/flag-icons/raw/260c91531be024944c6514130c5defb2ebb02b7d/flags/4x3/jp.svg",
```
Samoed, line 152: Can we move these values to Benchmark object?

mteb/leaderboard/benchmark_selector.py
Samoed: Can we move these values to Benchmark object? I don't like that we have multiple definition of benchmark object
And we can change this definition like
```python
(
        "Multilingual", ["MTEB_Multilingual"],
       "Language-specific Benchmarks", get.benchmarks(["MTEB(cmn, v1)", MTEB(eng, v1)], )
),
```

```python
class Benchmark:
    ...
    display_name="Japanese"
    icon="https://github.com/lipis/flag-icons/raw/260c91531be024944c6514130c5defb2ebb02b7d/flags/4x3/jp.svg"
```

x-tabdeveloping:
I might be able to fix the language selection thing too, I'm looking into it right now

KennethEnevoldsen:
> Yes, it is folded out by default, I think it's good if it's immediately obvious to our users how to change between them.

Agree

> I'm not sure I would remove the reference to the old leaderboard, some people will still be looking for it. And frankly, until we fix some annoying bugs, I don't think people will be completely satisfied with the new leaderboard.

>Maybe we can move the reference to the legacy leaderboard to the bottom, that way it doesn't look like we are apologizing for having made a new one

Buttom, it is then 

> I think leaving share this benchmark folded out is okay, it takes up little space, and makes it more visually present. It would be a shame if people tried copying the URL from the browser bar and then discovered that it doesn't actually lead to the benchmark they want.

It is only a slight preference, because it look a bit jarring so would fold it down (but not a showstopper). I don't think people will miss it

> The radar chart is not shown by default, you have to click on the tab to show it. I have no idea what you mean by moving the description up.

Something like this:

Screenshot 2025-04-16 at 11 52 59


> MTEB(eng, v1) is listed in the language-specific benchmarks section as English, Legacy.

Is it also linked in MTEB(eng, v2). Is it possible to have it not on the list and still linkable? otherwise let us put it in a legacy section and call it "English, v1"

x-tabdeveloping:
@KennethEnevoldsen What chart would you put to the right? I personally think the layout is fine as is to be honest.
I can make the "share this benchmark" not visible by default, and we can also hide MTEB(eng, v1) and make it linkable.
My issue is that if we do that, then more people would reach for the old leaderboard space to see that benchmark. I would prefer keeping it visible but making it very obvious that we do not want people to use it.

KennethEnevoldsen:
> My issue is that if we do that, then more people would reach for the old leaderboard space to see that benchmark. I would prefer keeping it visible but making it very obvious that we do not want people to use it.

Yep let us do that. We could:

> put it in a "Legacy" section and call it "English, v1"

mteb/leaderboard/benchmark_selector.py
Samoed: Yes, but with this we will have multiple configs for the same benchmarks and I don't think we should do that

mteb/leaderboard/table.py
isaac-chung: Awesome. Thanks for checking!

x-tabdeveloping:
> I might be able to fix the language selection thing too, I'm looking into it right now
I can't. Not even a submit button fixes it, or removing callbacks altogether...
I will try to recreate the problem in an isolated context for the Gradio team.

KennethEnevoldsen:
Sounds like a fix that is not required for this PR to go through.

Should we fix the rest and get this merged?

x-tabdeveloping:
Ye will do tmrw

tomaarsen (approved):
I'm seeing a lot of welcome improvements on top of an already solid interface, nice work here!

PR #2552, opened 2025-04-16
fix: me5 trainind data config to include xquad dataset

### LINKED ISSUE #2551: Add SquAD dataset to e5 training dataset config
Hi MTEB maintainers @Samoed 

While the [multilingual-e5 paper](https://arxiv.org/html/2402.05672v1) contains the SQuAD dataset, the [code](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/e5_models.py) does not seem to include it.

This might be problematic, especially considering that the leaderboard evaluates whether a model has been trained on in-domain data. Since xQuAD is derived from SQuAD, prior exposure could unfairly influence performance on certain benchmarks. 

What are your thoughts?

torchtorchkimtorch:
Closes #2551

### DIFF
--- a/mteb/models/e5_models.py
+++ b/mteb/models/e5_models.py
@@ -130,6 +130,7 @@ E5_TRAINING_DATA = {
 
 ME5_TRAINING_DATA = {
     **E5_TRAINING_DATA,
+    "XQuAD": ["train"], # trained on SQuAD train dataset
     "FEVER": ["train"],
     "FEVERHardNegatives": ["train"],
     "FEVER-NL": ["train"],  # translation not trained on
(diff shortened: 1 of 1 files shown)


mteb/models/e5_models.py
```diff
@@ -130,6 +130,7 @@
 
 ME5_TRAINING_DATA = {
     **E5_TRAINING_DATA,
+    "XQuAD": ["train"], # trained on SQuAD train dataset
```
Samoed, line 133: In mteb it defined as `XQuADRetrieval`
```suggestion
    "XQuADRetrieval": ["train"], # trained on SQuAD train dataset
```

Samoed (approved):

Issue #2016, opened 2025-02-07
[v2] Refactor task specific scores


Samoed:
https://github.com/embeddings-benchmark/mteb/blob/724f5538dcf0cc65982122fd151719d2385429d7/mteb/evaluation/evaluators/utils.py#L356-L372

https://github.com/embeddings-benchmark/mteb/blob/724f5538dcf0cc65982122fd151719d2385429d7/mteb/evaluation/evaluators/utils.py#L587-L610

Samoed:
@orionw Can these metrics be added to the results of all datasets?

orionw:
They really are task-specific though sadly -- there isn't overlap to combine them as it depends on the dataset's relevant judgements and structure.

Why is it complicated to maintain? I expect more datasets in the future will have task-specific metrics as well and this seems like a fairly straightforward way of adding them without forcing them always. Personally, I think our files already have too much (nauc_* and many more that you have to search through) and this approach allows a lot of flexibility for future growth (multimodal or otherwise).

Samoed:
I don't like that we change logic of scores based on task name in different parts of evaluation pipeline. If new tasks want to add their specific metrics, these lists of exceptions will raise.

 I think we can change like this:

```python
class AbsTaskRetrieval:
    ...

    def _evauate_subset(self, ...):
       ...
       scores.update(**self.cutom_metrics())
       return scores

    def custom_metrics(self, ....) -> dict[str, float]:
       return {}

class CutomTask(AbsTaskRetrieval):
      ...

       def custom_metrics(self, ....) -> dict[str, float]:
            return {"custom score": 1}
```

orionw:
>  If new tasks want to add their specific metrics, these lists of exceptions will raise.

Agree that this will continue to increase.

>  def custom_metrics(self, ....) -> dict[str, float]:

I like this direction! I forget, do the tasks already subclass the `AbsTask*`? If so then it seems like we could have the main class call `custom_metrics` to compute it, and it will have the same result as this custom function. The only potential wrinkle I could think of is to be sure that all the results are there for the `custom_metric` to process as IIRC it is a bit of a pain to thread all the relevant variables there (e.g. qrels, scores, etc.) but hopefully that is not too hard

Samoed:
> I forget, do the tasks already subclass the AbsTask*?

Yes

> AbstractTaskRetrieval, so those might need to change to account for the difference between AbsTaskInstructionRetrieval

[...]

Issue #2030, opened 2025-02-11
load_data() does not load self.dataset for all classes


jhyuklee:
Hi, I'm trying to load MTEB(multilingual) datasets locally by

```
tasks = mteb.get_tasks(...)

for task in tqdm(tasks):
  task.load_data()
  # do something with task.dataset

```

But I found out that not all `task` contain the `dataset` object (e.g. mteb/abstasks/AbsTaskRetrieval.py), which makes it difficult to download and process each dataset.

Was this intended? or Is there any plan to have a more consistent structure so that `load_data` always loads `self.dataset`?

Samoed:
Yes, retrieval tasks (and reranking) have different structure. Maybe this can be changed in `v2`.

orionw:
Thanks for raising @jhyuklee! I don't frequently use the non-Retrieval datasets and I missed this. +1 to making it standardized.

I think we could just move the fields into a dict like it is for STS:
```python
self.dataset[hf_subset][split]["sentence2"]
```

But it would be for retrieval fields:
```python
self.dataset[hf_subset][split]["corpus"]
self.dataset[hf_subset][split]["qrels"]
...
```

KennethEnevoldsen:
@Samoed I believe this is fixed? Any datasets that we are missing?

Samoed:
Yes, I think mieb retrieval using this format

Samoed:
I think we can close this issue after merge https://github.com/embeddings-benchmark/mteb/pull/3125, but retrieval tasks would still require additional conversion

PR #2600, opened 2025-04-30
[v2] Add task specific scores to retrieval

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/2016

| task_name                    |   main_score (PR) |   main_score (main) |
|:-----------------------------|------------------:|--------------------:|
| Core17InstructionRetrieval   |          0.020852 |            0.020852 |
| InstructIR                   |          0.107589 |            0.107589 |
| NevIR                        |          0.111352 |            0.111352 |
| News21InstructionRetrieval   |          0.036815 |            0.036815 |
| Robust04InstructionRetrieval |         -0.039893 |           -0.039893 |
| mFollowIR                    |          0.048701 |            0.048701 |
| mFollowIRCrossLingual        |         -0.020629 |           -0.020629 |

CC @orionw 

### Code Quality

### Documentation

### Testing

### Adding datasets checklist

### Adding a model checklist


mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -121,27 +121,31 @@ class AbsTaskRetrieval(AbsTask):
                     'top_ranked': Optional[dict[str, list[str]]]  # query_id -> doc_ids
                         Semantically, it should contain dict[split_name, dict[sample_id, str]]. If there are multiple instructions per query, please duplicate the queries and give them unique ids for consolidation.
                         E.g. {"test": {"query-id1": "instruction text"}}
+                    'qrels_diff': Optional[dict[str, list[str]]]
```
orionw, line 124: Is this adding the qrels diff to the main `AbsTaskRetrieval`? I think only 3 ish datasets use that, so I'm not sure if it's worth adding to the main class...

orionw (commented):
Thanks @Samoed!  I think this resolve the task specific scores we were talking about.

I'm worried that this is too tailored to how some of my datasets are. Is there a way we could define a `kwargs` like way of adding new attributes, so that if someone else has a new data attribute it can be easily added? I don't love having `qrel_diff` be in the main class and it seems like perhaps we could do something where each task can define a new attribute in `additional_features: dict` or something. 

Would love to hear @KennethEnevoldsen's thoughts as well

mteb/abstasks/AbsTaskRetrieval.py
Samoed: Yes, but I think all the data we're using should be downloaded from a one place

Samoed:
> I'm worried that this is too tailored to how some of my datasets are.  

If you're referring to `qrels_diff`, then yes, it is — but I’m not sure how to avoid that.

> Is there a way we could define a kwargs-like method of adding new attributes, so that if someone else has a new data attribute, it can be easily added?  

I’m not entirely sure that I understand the use case. I think most attributes can just be added directly to the class and accessed from there. `qrels_diff` is a bit different, though, since it involves loading data, and I'm not sure of the best way to handle that generically.

orionw:
Could we have a somewhat generic function where it will load keys passed from the task name?

Something like "additional_loading_kwargs" and then it loads those. 

Generally trying to find a solution that doesn't involve hardcoding the "qrel_diffs"

Samoed:
Yes, we can add a flag, but I think it would be better if the dataset object would have constant keys

mteb/abstasks/dataset_loaders.py
```diff
@@ -26,16 +26,18 @@ class RetrievalSplitData(TypedDict):
     - `relevant_docs`: A mapping of query IDs to a mapping of document IDs and their relevance scores.
     - `instructions`: A mapping of query IDs to their instructions (if applicable).
     - `top_ranked`: A mapping of query IDs to a list of top-ranked document IDs (if applicable).
+    - `qrels_diff`: A mapping of query IDs to a list of document IDs that have changed relevance scores (if applicable).
```
KennethEnevoldsen, line 29: This description makes no sense to me.

KennethEnevoldsen (commented):
I must say I am not at all sure what is going on in this PR - can I get an intro?

I get that we are trying to remove some custom metrics from the retrieval evaluation to then add them to the specific subtasks that need them.

However I am not sure exactly how this PR implements this and how one could add a custom metric.

I am also not quite sure what qrels_diff is used for.

Samoed:
> I must say I am not at all sure what is going on in this PR - can I get an intro?

For now some metrics for retrieval tasks are hidden in utils, and it's a hard to work with them 
https://github.com/embeddings-benchmark/mteb/blob/724f5538dcf0cc65982122fd151719d2385429d7/mteb/evaluation/evaluators/utils.py#L356-L372

https://github.com/embeddings-benchmark/mteb/blob/724f5538dcf0cc65982122fd151719d2385429d7/mteb/evaluation/evaluators/utils.py#L587-L610

> However I am not sure exactly how this PR implements this and how one could add a custom metric.

For now, I’ve moved the logic for determining whether a task needs custom metrics from a standalone function into a class method. These metrics are required for certain retrieval and reranking tasks, and can now be added by defining a `task_specific_scores` function within the task class.

> I am also not quite sure what qrels_diff is used for.

@orionw Can you give some description?

KennethEnevoldsen:
Got it - @orionw would love some more description on the qrels?

orionw:
Yes, I had set it up so that individual tasks could add extra custom metrics at the end without having to add them for all tasks.  @Samoed disagreed and wanted them to be defined in the task rather than in a catch all function at the end. My opinions on the matter are slight so I'm fine with whatever.

> qrel_diff

For some tasks (that are mostly mine) I have the model do a pair wise comparison with two very similar queries.  

This is really specific to mostly just some of my benchmarks. I would probably prefer this weird quirk not end up being hardcoded.  And what if someone else comes along and wants to compare more than pairs (says 3+) or has some other unique data they need to load.

So I was suggesting in this PR that we provide a function to allow each task to load some custom data that can then get stored and later used in their custom task scoring function.

KennethEnevoldsen:
Thanks @orionw!


> So I was suggesting in this PR that we provide a function to allow each task to load some custom data that can then get stored and later used in their custom task scoring function.

I am afraid that this will simply be an undocumented interface that no-one will use. @Samoed would it be possible to move the complexity completely to the tasks? 

I agree with @orionw point on:

> This is really specific to mostly just some of my benchmarks. I would probably prefer this weird quirk not end up being hardcoded.

Samoed:
> would it be possible to move the complexity completely to the tasks?

Yes, I've made it like this in this PR https://github.com/embeddings-benchmark/mteb/blob/40f547438ad74869d33174ce32c1a29c510a8454/mteb/abstasks/AbsTaskRetrieval.py#L124

Samoed:
> would it be possible to move the complexity completely to the tasks?

Do you mean to create a new function that would download qrels_diff for tasks that are needed? This approach will be similar with custom function

KennethEnevoldsen:
Yea so treat it as a custom metrics, which fetches the required information.
 
 We could allow for additional columns in the dataset and just fetch it from there (so we don't need to download steps).
 
 I.e. let us say that for a classification dataset we have some labels that are considered correct, but also have other that are acceptable. 

``` 
| text                               | label | viable-labels |
| ---------------------  | ----- | ------------- |
|  "Right love the idea!" | pos   | pos, neg        |

(text could be sarcastic)
```

There custom metric can then fetch data from self to compute the new metric.

I imagine that something like this would work?

Samoed:
I don't want to add this function for other tasks, because it is retrieval specific and it will be hard to standardize for all tasks. The problem with `qrels_diff` is that it's not an additional column, but a different subset that requires additional downloading. I can move downloading `qrels_diff` to `task_specific_scores`, but it would cause problems if we want to reupload dataset

KennethEnevoldsen:
Hmm, what are you suggesting then?

Samoed:
To left it as is, with `qrels_diff` as part of downloaded dataset in `self.dataset`

KennethEnevoldsen:
Alright, let us go with that

Samoed:
So can I merge this or you will do review of PR?

KennethEnevoldsen:
Yeah, fine to merge

orionw:
I think I'm confused why the custom data loading per task is undesirable here? I think this solution here is quite brittle and not extendable.

KennethEnevoldsen:
Should we try to make a PR to this branch with the attempted solution? I could actually imagine that it might be a good fix. I would really like to remove some of the complexity from the retrieval (especially because that is the part that most people will use)

Samoed:
I moved `qrels_diff` to task files

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -150,13 +154,14 @@ def convert_v1_dataset_format_to_v2(self):
...
-                    "instructions": None,
-                    "top_ranked": None,
-                }
+                lambda: RetrievalSplitData(
+                    corpus={},
+                    queries={},
+                    relevant_docs={},
+                    instructions=None,
+                    top_ranked=None,
+                    qrels_diff=None,
```
KennethEnevoldsen, line 163: can't we remove qrels here?

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -176,6 +181,10 @@ def convert_v1_dataset_format_to_v2(self):
                         self.dataset[subset][split]["top_ranked"] = self.top_ranked[
                             subset
                         ][split]
+                    if hasattr(self, "qrels_diff"):
```
KennethEnevoldsen, line 184: shouldn't be needed, right? we pass it directly to the compute?

KennethEnevoldsen (commented):
Alright thanks for the work on this, I believe we can remove the qrels from a few places, but might be wrong

Samoed:
Removed rest of `qrels_diff` mention

KennethEnevoldsen (approved):
I am very happy with how this turned out. I think it is good that we spent some time on it. 

@orionw do you have the chance to give this another look?

orionw (approved):
Yes, this is awesome - sorry I explained it poorly before. This helps keep it modular while also making it cleaner.  Thanks so much @Samoed!

Does this also resolve #2030? Seems like it does also which is great!

Samoed:
No, this doesn't resolve #2030. I’ve updated the retrieval task to store all data in `self.dataset`, but we still need to apply this change across all `load_data` methods in the tasks. I'm also in the process of reuploading the datasets to MTEB, and I plan to remove the `load_data` logic once we switch over to the new dataset format.

orionw:
Ah my bad.  But of course not a blocker for this, it looks great as is!

Issue #1650, opened 2025-01-01
openai/text-embedding-3-large does not allow for emptry string


Muennighoff:
2025-01-01 02:59:59.297956: E external/local_xla/xla/stream_executor/cuda/cuda_fft.cc:485] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2025-01-01 02:59:59.311878: E external/local_xla/xla/stream_executor/cuda/cuda_dnn.cc:8454] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
2025-01-01 02:59:59.315723: E external/local_xla/xla/stream_executor/cuda/cuda_blas.cc:1452] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
INFO:mteb.cli:Running with parameters: Namespace(model='openai/text-embedding-3-large', task_types=None, categories=None, tasks=['STS22'], languages=None, benchmarks=None, device=None, output_folder='/data/niklas/results/results', verbosity=2, co2_tracker=True, eval_splits=None, model_revision=None, batch_size=64, overwrite=False, save_predictions=False, func=<function run at 0x7ff797b0d630>)
INFO:mteb.evaluation.MTEB:

## Evaluating 1 tasks:
─────────────────────────────── Selected tasks  ────────────────────────────────
STS
    - STS22, p2p, multilingual 18 / 18 Subsets

INFO:mteb.evaluation.MTEB:

********************** Evaluating STS22 **********************
WARNING:mteb.abstasks.AbsTask:Dataset 'STS22' is superseded by 'STS22.v2', you might consider using the newer version of the dataset.
Loading Dataset Infos from /env/lib/conda/gritkto/lib/python3.10/site-packages/datasets/packaged_modules/json
INFO:datasets.info:Loading Dataset Infos from /env/lib/conda/gritkto/lib/python3.10/site-packages/datasets/packaged_modules/json
Overwrite dataset info from restored data version if exists.
INFO:datasets.builder:Overwrite dataset info from restored data version if exists.
Loading Dataset info from /data/huggingface/datasets/mteb___sts22-crosslingual-sts/default/0.0.0/de9d86b3b84231dc21f76c7b7af1f28e2f57f6e3
[...]

PR #2634, opened 2025-05-03
fix for empty string input to openai/text-embedding-3-large

ayush1298:
fix for #1650.
I dont have OpenAI API key to test whether its working correctly or not. @KennethEnevoldsen Can you please test it, when you get time.

### Code Quality

### Documentation

### Testing

### DIFF
--- a/mteb/models/openai_models.py
+++ b/mteb/models/openai_models.py
@@ -63,6 +63,10 @@ class OpenAIWrapper(Wrapper):
                 "Reducing embedding size available only for text-embedding-3-* models"
             )
 
+        if self._embed_dim and all(not s.strip() for s in sentences):
+            logger.warning("Empty input detected - returning zero embeddings.")
+            return np.zeros((len(sentences), self._embed_dim), dtype=np.float32)
+
         trimmed_sentences = []
         for sentence in sentences:
             encoded_sentence = self._encoding.encode(sentence)
--- a/mteb/models/openai_models.py
+++ b/mteb/models/openai_models.py
@@ -55,14 +55,10 @@ class OpenAIWrapper(Wrapper):
         return self._encoding.decode(truncated_sentence)
 
     def encode(self, sentences: list[str], **kwargs: Any) -> np.ndarray:
-        requires_package(self, "openai", "Openai text embedding")
+        print("Starting OpenAIWrapper encode method")
         from openai import NotGiven
 
         empty_mask = [not s.strip() for s in sentences]
-        if all(empty_mask):
-            logger.warning("Empty input detected - returning zero embeddings.")
-            return np.zeros((len(sentences), self._embed_dim), dtype=np.float32)
-
         if self._model_name == "text-embedding-ada-002" and self._embed_dim is not None:
             logger.warning(
                 "Reducing embedding size available only for text-embedding-3-* models"
--- a/mteb/models/openai_models.py
+++ b/mteb/models/openai_models.py
@@ -19,8 +19,8 @@ class OpenAIWrapper(Wrapper):
         self,
         model_name: str,
         max_tokens: int,
+        embed_dim: int,
         tokenizer_name: str = "cl100k_base",  # since all models use this tokenizer now
-        embed_dim: int | None = None,
         **kwargs,
     ) -> None:
         """Wrapper for OpenAIs embedding API.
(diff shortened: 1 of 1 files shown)


mteb/models/openai_models.py
```diff
@@ -63,6 +63,10 @@ def encode(self, sentences: list[str], **kwargs: Any) -> np.ndarray:
                 "Reducing embedding size available only for text-embedding-3-* models"
             )
 
+        if self._embed_dim and all(not s.strip() for s in sentences):
```
KennethEnevoldsen, line 66: Since this doesn't actually call openai you could have just tested, but:

```py
import mteb

model = mteb.get_model("openai/text-embedding-3-small")

arr = model.encode([""])
# error
```

It seems like `self._embed_dim` is None, so this statement is skipped.

We can simply require the embed dim to be specified, by changing the input to:
```
    def __init__(
        self,
        model_name: str,
        *,
        max_tokens: int,
        tokenizer_name: str,
        embed_dim: int,
        **kwargs,
    ) -> None:
```

KennethEnevoldsen: However, event with this fix it still fails on cases like:

```
arr = model.encode(["", "hello world"])
```

This is because the conditions require all strings to be empty, which is rarely the case.

What you need to do here is take out the empty string, encode the rest, and add it back in.

KennethEnevoldsen (changes requested):

mteb/models/openai_models.py
```diff
@@ -54,23 +54,35 @@ def truncate_text_tokens(self, text):
...
-        for sentence in sentences:
+        for sentence in non_empty_sentences:
             encoded_sentence = self._encoding.encode(sentence)
             if len(encoded_sentence) > self._max_tokens:
                 truncated_sentence = self.truncate_text_tokens(sentence)
                 trimmed_sentences.append(truncated_sentence)
             else:
                 trimmed_sentences.append(sentence)
+        if not trimmed_sentences:
+            return np.zeros((len(sentences), self._embed_dim), dtype=np.float32)
```
KennethEnevoldsen, line 85: not needed

mteb/models/openai_models.py
```diff
@@ -54,23 +54,35 @@ def truncate_text_tokens(self, text):
         return self._encoding.decode(truncated_sentence)
 
     def encode(self, sentences: list[str], **kwargs: Any) -> np.ndarray:
-        requires_package(self, "openai", "Openai text embedding")
+        empty_mask = [not s.strip() for s in sentences]
+        if all(empty_mask):
+            logger.warning("Empty input detected - returning zero embeddings.")
+            return np.zeros((len(sentences), self._embed_dim), dtype=np.float32)
```
KennethEnevoldsen, line 60: not needed

mteb/models/openai_models.py
```diff
@@ -54,23 +54,35 @@ def truncate_text_tokens(self, text):
...
 
+        requires_package(self, "openai", "Openai text embedding")
```
KennethEnevoldsen, line 62: require pacakge and import should be in the start of the function

mteb/models/openai_models.py
```diff
@@ -118,7 +130,11 @@ def encode(self, sentences: list[str], **kwargs: Any) -> np.ndarray:
                     )
             all_embeddings.extend(self._to_numpy(response))
 
-        return np.array(all_embeddings)
+        all_embeddings = np.array(all_embeddings)
+        final_embeddings = np.zeros((len(sentences), self._embed_dim), dtype=np.float32)
```
KennethEnevoldsen, line 134: this already fill out the empty texts.

You can then simply fill in the text_embeddings using:

```py
mask = [i for i, t in enumerate(text) if t.strip()]
final_emb[mask, :] = text_emb
``` 

Here is a small sample to show the idea:

```py
import numpy as np

matrix1 = np.zeros((5, 4))
matrix2 = np.random.rand(3, 4)
matrix1[[0, 2, 4],:] = matrix2
```

mteb/models/openai_models.py
```diff
@@ -19,8 +19,8 @@ def __init__(
         self,
         model_name: str,
         max_tokens: int,
```
KennethEnevoldsen, line 21:
```suggestion
        model_name: str,
        *,
        max_tokens: int,
```

I would do this to ensure that every argument is a keyword argument.

KennethEnevoldsen (commented):
I addded some suggestions. Do give it another try and try to give me a debug trace if it doesn't work.

mteb/models/openai_models.py
ayush1298: What if whole input is empty or it's not possible?

mteb/models/openai_models.py
ayush1298: Didn't get these one. Current implementation was 1st creating whole vector of zeros, and then only setting text embeddings, at positions where non-empty text is present. I think you are suggesting same only.

mteb/models/openai_models.py
KennethEnevoldsen: I suppose it could be possible, but that should be handled later in the workflow

if sentences:
  embed

KennethEnevoldsen:
Thanks for taking the time on this @ayush1298. I built upon you earlier pr. and made a fix: #2676

ayush1298:
> Thanks for taking the time on this @ayush1298. I built upon you earlier pr. and made a fix: #2676
>
Thanks @KennethEnevoldsen

PR #2664, opened 2025-05-06
Share benchmark

ayush1298:
This PR is related to creating a link for sharing a benchmark for custom benchmark means when some custom filters are applied, apart from the default one. 

@x-tabdeveloping @KennethEnevoldsen In this PR, the idea that I used was simple, like whenever we apply any filters other than the default values of each filter for that particular benchmark (which we see 1st time on the leaderboard), based on this filter state construction is done. Filter state is encoded to JSON, JSON is compressed, and converted to a base64 encoded string, and then converted to a URL. The URL will be long, still I don't think it's a problem as there is a copy button already and the user has to copy url from there and paste it, which is not a difficult thing. But, still, I have added a code to convert these long URLs to short URLs, but I don't know whether we need it or not. 

From these long URLs, it is 1st analyzed whether it's a default or custom benchmark, and then url is decoded and then tries to produce the exact same filter state as the user used at the time of sharing the benchmark.  So, overall flow is as follows: 
Original Filters → encode_filter_state → URL → decode_filter_state → Restored Filters

### Code Quality

### Documentation

- [] **Updated Documentation**: Add or update documentation to reflect the changes introduced in this PR.

### Testing

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -8,7 +8,6 @@ import time
 import warnings
 from pathlib import Path
 from typing import Literal, get_args
-from urllib.parse import urlencode
 
 import cachetools
 import gradio as gr
@@ -23,6 +22,11 @@ from mteb.languages import ISO_TO_LANGUAGE
 from mteb.leaderboard.benchmark_selector import BENCHMARK_ENTRIES, make_selector
 from mteb.leaderboard.figures import performance_size_plot, radar_chart
 from mteb.leaderboard.table import create_tables
+from mteb.leaderboard.url_utils import (
+    decode_filter_state,
+    encode_filter_state,
+    generate_short_url,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


mteb/models/openai_models.py
```diff
@@ -55,16 +56,26 @@ def truncate_text_tokens(self, text):
 
     def encode(self, sentences: list[str], **kwargs: Any) -> np.ndarray:
         requires_package(self, "openai", "Openai text embedding")
-
         from openai import NotGiven
 
+        empty_mask = [not s.strip() for s in sentences]
```
Samoed, line 61: I don't think this is relevant to this PR

ayush1298: Sorry, I am deleting this PR. I mistakenly made it from different branch instead of main

mteb/leaderboard/app.py
```diff
@@ -65,25 +69,98 @@ def load_results():
...
+        filter_state["d"] = domains
+
+    if task_types:
+        filter_state["tt"] = task_types
+
+    if modalities:
+        filter_state["m"] = modalities
+
+    if tasks:
+        filter_state["t"] = tasks
```
Samoed, line 100: I think full names will be easier to understand (instead of `t` tasks)

mteb/models/openai_models.py
ayush1298: @Samoed I made these branch from openai_models branch, should I just remove that file?

ayush1298:
Sorry, But I have made a mistake so I am closing this PR and created a new PR #2666

Issue #1755, opened 2025-01-10
Leaderboard: Share this benchmark does not work for custom benchmarks


Muennighoff:
E.g. when removing / adding custom tasks / languages; ideally the link would change and one could link to that I think

KennethEnevoldsen:
For now let us add a note that states that it does not work for custom benchmarks

Muennighoff:
Makes sense; alternative to long URLs could be hashing it into a short string I think

Muennighoff:
The new multi-pages feature in Gradio may be relevant here: https://github.com/gradio-app/gradio/pull/10433

Thanks @abidlabs for the pointer!

Muennighoff:
Other markers like e.g. the zero-shot filtering setting could also be added to the URL as Tengyu suggested!

x-tabdeveloping:
Certainly! We can expand the functionality as we see fit.
I have looked at the Gradio feature you linked, and it does not seem like it's exactly what we are looking for.
In order for the URL in the browser bar to change according to the filters applied, we would need some way of programmatically altering the URL, which to my knowledge is still not supported in Gradio. Perhaps some Javascript hack could make it work, but I would prefer to avoid such solutions.

KennethEnevoldsen:
I took a closer look at this over at #2666

> We have ~1000 languages, 131 tasks, and probably like 100 options in the other categories. Even if we just use three characters pr. item we will get >3000 characters, which is the URL limit. This does not even include additional characters. It might be doable with a compression on the text, but it would still be very close to 2k characters to share a url...
> 
> Here is the following option as I see it:
> 
> 1. dictionary method (current): The problem is that is resets whenever the leaderboard resets. I would rather not have the feature than see people get frustrated when it doesn't work
> 2. some sort of database: We store the setting for a specific ID (close to the current solution). This is quite a dependency to manage for a very small feature
> 3. Reduce the number of languages: We could group low-resource languages together (~700 languages only appear in BibleNLP). One option is simply to group these together into some kind of "Other". This gives a lot of problems with the UI (What is other? How do we convey that such that people understand)
> 4. Reduce the number of filter that can be applied (I don't know which ones I would remove)
> 5. We drop the feature
> 
> I am totally open to ideas here, but currently I am leaning towards 3 or 5 (maybe four because we have other things to do)

@Samoed responded:

> I think we can lower amount of languages in url generation (e.g. top 50 most common languages), and can leave other parts as is


[...]

PR #2666, opened 2025-05-06
Share Benchmark Link for Custom Benchmarks

ayush1298:
This PR is related to creating a link for sharing a benchmark for custom benchmark means when some custom filters are applied, apart from the default one as suggested in issue #1755

@x-tabdeveloping @KennethEnevoldsen In this PR, the idea that I used was simple, like whenever we apply any filters other than the default values of each filter for that particular benchmark (which we see 1st time on the leaderboard), based on this filter state construction is done. Filter state is encoded to JSON, JSON is compressed, and converted to a base64 encoded string, and then converted to a URL. The URL will be long, still I don't think it's a problem as there is a copy button already and the user has to copy url from there and paste it, which is not a difficult thing. But, still, I have added a code to convert these long URLs to short URLs, but I don't know whether we need it or not.

From these long URLs, it is 1st analyzed whether it's a default or custom benchmark, and then url is decoded and then tries to produce the exact same filter state as the user used at the time of sharing the benchmark. So, overall flow is as follows:
Original Filters → encode_filter_state → URL → decode_filter_state → Restored Filters

### Code Quality

### Documentation

### Testing

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -8,7 +8,6 @@ import time
 import warnings
 from pathlib import Path
 from typing import Literal, get_args
-from urllib.parse import urlencode
 
 import cachetools
 import gradio as gr
@@ -23,6 +22,11 @@ from mteb.languages import ISO_TO_LANGUAGE
 from mteb.leaderboard.benchmark_selector import BENCHMARK_ENTRIES, make_selector
 from mteb.leaderboard.figures import performance_size_plot, radar_chart
 from mteb.leaderboard.table import create_tables
+from mteb.leaderboard.url_utils import (
+    decode_filter_state,
+    encode_filter_state,
+    generate_short_url,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -65,16 +69,43 @@ def load_results():
             return mteb.BenchmarkResults.from_validated(**json.load(cache_file))
 
 
-def produce_benchmark_link(benchmark_name: str, request: gr.Request) -> str:
-    """Produces a URL for the selected benchmark."""
-    params = urlencode(
-        {
-            "benchmark_name": benchmark_name,
-        }
-    )
+def produce_benchmark_link(
+    benchmark_name: str,
+    languages: list[str],
+    domains: list[str],
+    task_types: list[str],
+    modalities: list[str],
+    tasks: list[str],
+    request: gr.Request,
+) -> str:
+    """Produces a URL for the selected benchmark with all applied filters in compact form."""
+    # Create filter state dictionary with non-empty/non-default values
+    filter_state = {
+        "b": benchmark_name,
+    }
+
+    if languages:
+        filter_state["l"] = languages
+
+    if domains:
+        filter_state["d"] = domains
+
+    if task_types:
+        filter_state["tt"] = task_types
+
+    if modalities:
+        filter_state["m"] = modalities
+
+    if tasks:
+        filter_state["t"] = tasks
+
+    encoded = encode_filter_state(filter_state)
+
     base_url = request.request.base_url
-    url = f"{base_url}?{params}"
-    md = f"```\n{url}\n```"
+    long_url = f"{base_url}?s={encoded}"
+    short_id = generate_short_url(long_url)
+    short_url = f"{base_url}?id={short_id}"
+    md = f"```\n{short_url}\n```"
     return md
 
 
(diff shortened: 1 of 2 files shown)


mteb/leaderboard/app.py
```diff
@@ -65,25 +69,98 @@ def load_results():
...
+    encoded = encode_filter_state(filter_state)
+
     base_url = request.request.base_url
-    url = f"{base_url}?{params}"
-    md = f"```\n{url}\n```"
+    long_url = f"{base_url}?s={encoded}"
+    short_id = generate_short_url(long_url)
+    short_url = f"{base_url}?id={short_id}"
+    md = f"```\n{short_url}\n```"
     return md
```
ayush1298, line 109: @Samoed didn't get what return types you are suggesting to add here?

mteb/leaderboard/app.py
```diff
@@ -65,25 +69,98 @@ def load_results():
...
+        filter_state["d"] = domains
+
+    if task_types:
+        filter_state["tt"] = task_types
+
+    if modalities:
+        filter_state["m"] = modalities
+
+    if tasks:
+        filter_state["t"] = tasks
```
ayush1298, line 100: @Samoed, I have used short names instead of full names, mainly to minimize URL lengths.  One of the things is many browsers have 2048 character limits, so I was just trying to optimize string length as much as possible. As stated in description, I am not sure whether we have to use short_url form of long_url or not, as 2 different long url can get converted to same short url, causing confusions. So, we have to think of length of url also.

mteb/leaderboard/app.py
Samoed: Now, `set_benchmark_on_load` returns a lot of values, and it would be nice if there were type annotations for them.

ayush1298: Added this

ayush1298:
@KennethEnevoldsen @x-tabdeveloping  Can you review this one, when you get time.

KennethEnevoldsen:
@ayush1298 can I ask you to push this to a dummy space on the hub, otherwise it is quite hard to validate if it works

ayush1298:
> @ayush1298 can I ask you to push this to a dummy space on the hub, otherwise it is quite hard to validate if it works
>
Sorry for late reply. Can you check this now?
https://huggingface.co/spaces/AyushM6/leaderboard.
Also, do we want short url or not as it is not actual short_url, its obtained from long_url only and 2 different long url can get converted to same short url, causing confusions, though chances will be very less of having that. Or we can use long url only as we are already having copy button.

KennethEnevoldsen:
great! 

Just to test, can you show me a screenshot of the filtering for:

http://ayushm6-leaderboard.hf.space/?id=bf4702e3

ayush1298:
> great!
> 
> Just to test, can you show me a screenshot of the filtering for:
> 
> http://ayushm6-leaderboard.hf.space/?id=bf4702e3
>
I think only No of task types are changed in this link from 7 to 9, rest other filters are set to default of `MTEB(Multilingual, v2)`
image

KennethEnevoldsen:
Hmm, the selection was languages = "dan" and then remove bitext mining

ayush1298:
> Hmm, the selection was languages = "dan" and then remove bitext mining
>
Can you confirm whether you have same filter setting as of below images. I am getting exactly same link as yours(you can see it in 1st image).
image
image

I think problem is somewhere in decoding code, maybe it was not able to decode the url correctly to filters.

KennethEnevoldsen:
Alright, new test:

http://ayushm6-leaderboard.hf.space/?id=d8070841

Once we have it working I would also like to keep the original share link as is, but create a share with filters button at the very end of the dropdown.

Samoed:
I think it's working 
image

ayush1298:
> http://ayushm6-leaderboard.hf.space/?id=d8070841
>
Got the same as given by @Samoed. Only single language `Dan` and and 8 task types, `BitexMinining` is removed.

> Once we have it working I would also like to keep the original share link as is, but create a share with filters button at the very end of the dropdown.
>
Didn't get why we need 2 different buttons, as even when no filters are applied these will work.

mteb/leaderboard/app.py
```diff
@@ -23,12 +22,18 @@
...
+    encode_filter_state,
+    generate_short_url,
+)
 
 logger = logging.getLogger(__name__)
 
 
 LANGUAGE: list[str] = list({l for t in mteb.get_tasks() for l in t.metadata.languages})
 ALL_MODELS = {meta.name for meta in mteb.get_model_metas()}
+SHORT_URL_MAP: dict[str, str] = {}  # Map Short Url id to its encoded state
```
KennethEnevoldsen, line 36: Hmm, but won't this be reset if you restart the space (which we do every day)?

KennethEnevoldsen: Hmm that doesn't work. We need to be able to update the leaderboard daily, and we don't want to have a saved JSON (too many backward compatibility issues).

Samoed: Can you decode strings without storing?

KennethEnevoldsen:
Took a closer look at this. We have ~1000 languages, 131 tasks, and probably like 100 options in the other categories.
Even if we just use three characters pr. item we will get >3000 characters, which is the URL limit. This does not even include additional characters. It might be doable with a compression on the text, but it would still be very close to 2k characters to share a url...

Here is the following option as I see it:

1) dictionary method (current): The problem is that is resets whenever the leaderboard resets. I would rather not have the feature than see people get frustrated when it doesn't work
2) some sort of database: We store the setting for a specific ID (close to the current solution). This is quite a dependency to manage for a very small feature
3) Reduce the number of languages: We could group low-resource languages together (~700 languages only appear in BibleNLP). One option is simply to group these together into some kind of "Other". This gives a lot of problems with the UI (What is other? How do we convey that such that people understand)
4) We drop the feature

I am totally open to ideas here, but currently I am leaning towards 3 or 4 (maybe four because we have other things to do)

@Samoed would love your opinion here as well?

Samoed:
I think we can lower amount of languages in url generation (e.g. top 50 most common languages), and can leave other parts as is

KennethEnevoldsen:
Let is take this discussion in the issue. I will close this one for now

ayush1298:
@KennethEnevoldsen @Samoed I can still try to remove dependency on map. For grouping large no of languages, instead of defining other we can name them with task only, so people can identify them.  Maybe somewhere we can add list of this type of groups

KennethEnevoldsen:
Naming them with only a task won't work, as language also filters within the task. 

Let us take the time to reach a good solution in the discussion; no need to rush to find a solution

ayush1298:
> Naming them with only a task won't work, as language also filters within the task.
> 
> Let us take the time to reach a good solution in the discussion; no need to rush to find a solution
>
Okay, will wait untill we get a good solution.

Issue #2556, opened 2025-04-17
LB freezes


Muennighoff:
Maybe this is just me but the LB freezes for me when trying to customize languages, e.g. below an image of me trying to add Zho to the Japanese benchmark selectable on the left-hand-side

<img width="1467" alt="Image" src="https://github.com/user-attachments/assets/ace20722-1e3f-4641-bc04-d03cb41efcef" />

KennethEnevoldsen:
I believe this is a known problem that @x-tabdeveloping is currently working with the gradio team to debug.

KennethEnevoldsen:
Just tried to reproduce this:

- Safari: It works without issue (no bug)
- Chrome: I get the same error
  - tried to figure out where it happens:
    - 1) doesn't happen with just a language deselect 
    - 2) doesn't happen with a "deselect all" in domain or task (which are both quite large)
    - Conclusion it is probably because we have a dropdown of +1000 languages.
- firefox: Works as well


A potential solution might be to move a bunch of languages into the "other" category (e.g., all languages only present in the < <3 task). This would probably move it down to ~100-250 languages.

PR #2672, opened 2025-05-07
Leaderboard: UI simplifications for menus

KennethEnevoldsen:
Did a few things to improve the simplify the leaderboard UI.

Changes:
- Combined FAQ entries
- Created dropdowns in the select benchmark menu sidebar
- Removed reference to arena
- Removed reference to old leaderboard
- reduced size of the select menu
- reduced the size of acknowledgements
- removed farsi from the selection (as it is a beta)

bugfixes:
- Fixed the bug that happens due to language selection: #2556. 
  - Turns out that it happened due to the size of the list. Reducing it to "only" the language in MTEB resolved the issue.

refactors:
- Refactored to use a class for menu items
- refactored text segments out of app.py

<details>
<summary> Images </summary>

Overview
Screenshot 2025-05-07 at 13 03 09

FAQ Unfolded:
Screenshot 2025-05-07 at 13 03 26

FAQ Folded (default) + Ack. 
Screenshot 2025-05-07 at 13 03 16

</details>

### DIFF
--- a/mteb/leaderboard/benchmark_selector.py
+++ b/mteb/leaderboard/benchmark_selector.py
@@ -7,35 +7,24 @@ import gradio as gr
 import mteb
 from mteb import Benchmark
 
-"""
-Each entry is a tuple, where the first element is a label, and the second is either a single benchmark or a group of benchmarks.
-
-Example:
-[
-    ("First Benchmark", dict(value="MTEB(something)", icon="icon_url")),
-    ("Group of Benchmarks",
-        [
-            ("Second Benchmark", dict(value="MTEB(something)", icon="icon_url")),
-            ("Third Benchmark", dict(value="MTEB(something)", icon="icon_url")),
-        ],
-    ),
-]
-"""
-
 
 @dataclass
 class MenuEntry:
     name: str | None
     benchmarks: list[Benchmark]
     open: bool = False
+    size: str = "sm"
 
 
 BENCHMARK_ENTRIES = [
     MenuEntry(
-        None, mteb.get_benchmarks(["MTEB(Multilingual, v2)", "MTEB(eng, v2)"]), False
+        None,
+        mteb.get_benchmarks(["MTEB(Multilingual, v2)", "MTEB(eng, v2)"]),
+        False,
+        size="md",
     ),
     MenuEntry(
-        "Image Benchmarks",
+        "Image",
         mteb.get_benchmarks(
             [
                 "MIEB(Multilingual)",
@@ -47,7 +36,7 @@ BENCHMARK_ENTRIES = [
         True,
     ),
     MenuEntry(
-        "Regional Benchmarks",
+        "Regional",
         mteb.get_benchmarks(
             [
                 "MTEB(Europe, v1)",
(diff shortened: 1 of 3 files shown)


mteb/leaderboard/benchmark_selector.py
```diff
@@ -19,9 +21,20 @@
     ),
 ]
 """
+
+
+@dataclass
+class MenuEntry:
+    name: str | None
+    benchmarks: list[Benchmark]
+    open: bool
```
Samoed, line 30: Good addition. Maybe add default value to open to write it only once?

```suggestion
class MenuEntry:
    name: str | None
    benchmarks: list[Benchmark]
    open: bool = False
```

Samoed (approved):

mteb/leaderboard/benchmark_selector.py
```diff
@@ -19,9 +21,20 @@
...
+
+
 BENCHMARK_ENTRIES = [
-    mteb.get_benchmarks(["MTEB(Multilingual, v2)", "MTEB(eng, v2)"]),
-    (
+    MenuEntry(
+        None, mteb.get_benchmarks(["MTEB(Multilingual, v2)", "MTEB(eng, v2)"]), False
```
x-tabdeveloping, line 35: This `None` thing is a bit weird don't you think? I think if you make a datastructure it should preferably account for both top level benchmarks and groups of benchmarks.

mteb/leaderboard/benchmark_selector.py
```diff
@@ -133,38 +151,29 @@ def _update_value(label: str) -> str:
...
 ) -> tuple[gr.State, gr.Column]:
     if not entries:
         raise ValueError("No entries were specified, can't build selector.")
     label_to_value = {}
-    state = None
+
     with gr.Column() as column:
+        state = gr.State(entries[0].benchmarks[0].name)
         i = 0
         for entry in entries:
```
x-tabdeveloping, line 163: Yeah, removing this and not accounting for hierarchy is going exactly in the opposite direction of what we want, no? I mean we would prefer to be able to do the thing in the future that the RTEB guys want where you can have sub-benchmarks.

If you remove this it also makes it less visually apparent, which benchmarks we consider to be top-level (of high importance) and which we consider to be a part of a group of benchmarks. (though this might be something we want idk)

x-tabdeveloping (commented):
Agree with most of the changes, I have some worries about the new `MenuEntry` datatype though. I have a feeling that it is a bit premature to introduce a data type here, and you already have to make an odd exception (something with a `None` name).
I also feel like this will not be able to account for the hierarchical structure we would like to introduce to this UI element in the future.

mteb/leaderboard/benchmark_selector.py
KennethEnevoldsen: None is basically stating that the header does not have name. I think that is as intended. 

The only reason for the datatype was because I spent a while figuring out the previous list. This says that the content of the list is and defined the items within the tuple. It could be a named tuple as well.

mteb/leaderboard/benchmark_selector.py
KennethEnevoldsen: but it looks exactly the same as before? I see no differences - did the previous code allow for hierarchies?

mteb/leaderboard/benchmark_selector.py
KennethEnevoldsen: I don't think it is a big deal, but will change it (depends on comments below)

KennethEnevoldsen:
Discussed with @x-tabdeveloping, added some size fixes, but will then merge

PR #2424, opened 2025-03-24
Add model specific dependencies in pyproject.toml

ayush1298:
This PR is related to addition of model specific dependencies in pyproject.toml as well as giving detailed and clear warning using` requires_package`.
Closes #2398 .

### Code Quality

- [x] **Code Formatted**: Format the code using `make lint` to maintain consistent style.

### Documentation

- [x] **Updated Documentation**: Add or update documentation to reflect the changes introduced in this PR.

### Testing

- [ ] **New Tests Added**: Write tests to cover new functionality. Validate with `make test-with-coverage`.
- [x] **Tests Passed**: Run tests locally using `make test` or `make test-with-coverage` to ensure no existing functionality is broken.

PR #2706, opened 2025-05-19
fix: Ensure that optional dependencies work

KennethEnevoldsen:
Fixes mistakes introduced in https://github.com/embeddings-benchmark/mteb/pull/2424

It seems like many of these requirements doesn't exist (voyageai>=1.0.0). @ayush1298 I am hoping you could clear up how this happened?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,12 +87,12 @@ bm25s = ["bm25s>=0.2.6", "PyStemmer>=2.2.0.3"]
 gritlm = ["gritlm>=1.0.2"]
 xformers = ["xformers>=0.0.29"]
 blip2 = ["salesforce-lavis>=1.0.2"]
-voyageai = ["voyageai>=1.0.0,<2.0.0"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -203,3 +203,8 @@ addopts = """
 override-dependencies = [
     "salesforce-lavis", # salesforce-lavis is not valid with sentence transformers >=3.0.0
 ]
+conflicts = [
+    [{ extra = "timm" }, { extra = "blip2" }],
+    [{ extra = "llm2vec" }, { extra = "pylate" }],
+    [{ extra = "llm2vec" }, { extra = "model2vec" }]
+]
```
ayush1298, line 210: How did you define this exactly? Will there some other conflicts also that we have to add?

KennethEnevoldsen: I used [UV](https://docs.astral.sh/uv/) to install `mteb`, it gives you all the problems

ayush1298 (approved):

PR #2707, opened 2025-05-19
fix: Promote Persian benchmark to v1

mehran-sarmadi:
1-Promoted Persian benchmark to **v1**: renamed to **"Farsi"**, updated description, citation, and reference.
2-Renamed task `PersianTextTone` to `SynPerTextToneClassification`.
3-Updated `tooka_sbert` metadata.
4-Added `MTEB(fas, v1)` to benchmark selector.

### Code Quality

### Testing

### DIFF
--- a/mteb/overview.py
+++ b/mteb/overview.py
@@ -367,6 +367,9 @@ def get_tasks(
     return MTEBTasks(_tasks)
 
 
+_TASK_RENAMES = {"PersianTextTone": "SynPerTextToneClassification"}
+
+
 def get_task(
     task_name: str,
     languages: list[str] | None = None,
@@ -400,6 +403,12 @@ def get_task(
     Examples:
         >>> get_task("BornholmBitextMining")
     """
+    if task_name in _TASK_RENAMES:
+        _task_name = _TASK_RENAMES[task_name]
+        logger.warning(
+            f"The task with the given name '{task_name}' has been renamed to '{_task_name}'. To prevent this warning use the new name."
+        )
+
     if task_name not in TASKS_REGISTRY:
         close_matches = difflib.get_close_matches(task_name, TASKS_REGISTRY.keys())
         if close_matches:


mteb/tasks/Classification/fas/FaMTEBClassification.py
```diff
@@ -436,13 +436,13 @@ class SynPerChatbotToneUserClassification(AbsTaskClassification):
...
-class PersianTextTone(AbsTaskClassification):
+class SynPerTextToneClassification(AbsTaskClassification):
     metadata = TaskMetadata(
-        name="PersianTextTone",
+        name="SynPerTextToneClassification",
         description="Persian Text Tone",
         reference="https://mcinext.com/",
         dataset={
-            "path": "MCINext/persian-text-tone",
+            "path": "MCINext/synthetic-persian-text-tone-classification",
```
Samoed, line 445: Data have been changed? If yes, I think you need rerun models. Also you need to rename files in results repo

mehran-sarmadi: No, only its name was changed to follow the naming convention used for all our datasets

Samoed: Ah, yes. Forgot about https://github.com/embeddings-benchmark/results/pull/199

mteb/models/fa_models.py
```diff
@@ -168,3 +166,43 @@
...
+    similarity_fn_name="cosine",
+    framework=[],
+    use_instructions=False,
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
+)
+
+hakim_small = ModelMeta(
+    name="MCINext/Hakim-small",
```
Samoed, line 191: I can't find these models on HF

mteb/tasks/Classification/fas/FaMTEBClassification.py
```diff
@@ -436,13 +436,13 @@ class SynPerChatbotToneUserClassification(AbsTaskClassification):
     samples_per_label = 32
 
 
-class PersianTextTone(AbsTaskClassification):
+class SynPerTextToneClassification(AbsTaskClassification):
     metadata = TaskMetadata(
-        name="PersianTextTone",
+        name="SynPerTextToneClassification",
```
KennethEnevoldsen, line 441: Hmm, this will break backward compatibility if someone is using this task. I am not sure what the best thing to do here.

Did parts of the data change? If so then I would just superseede it.

btw. I think you also need to change the revision.

KennethEnevoldsen (changes requested):

mteb/models/fa_models.py
Samoed: Maybe it would be easier to move out to separate PR?

mteb/tasks/Classification/fas/FaMTEBClassification.py
mehran-sarmadi: The data hasn’t changed at all, just the task name. 
I checked on Hugging Face and since it’s only a name change, the revision stayed the same.

mteb/models/fa_models.py
```diff
@@ -141,9 +141,7 @@
     use_instructions=False,
     public_training_code=None,
     public_training_data=None,
-    training_datasets={
-        # This model is finetuned from PartAI/TookaBERT-Large
-    },
+    training_datasets=None,
```
Samoed, line 144: With this change `ZeroShot` will be 0

mehran-sarmadi: I checked with existing models and ran it locally. It will appear as N/A, and this is  true for this model because they don't mention anything about the datasets.

mteb/tasks/Classification/fas/FaMTEBClassification.py
KennethEnevoldsen: @Samoed, @isaac-chung what do you think are good ways to handle the name changes? We could have a mapping get_tasks, such that if the `PersianTextTone` is called we simply replace it with `SynPerTextToneClassification` and throw a deprecation warning?

Samoed: I think yes, that's fine

isaac-chung: Does this mean that the old name is no longer accessible? If so, then I think it's ok just to replace (as we did for a few datasets that got moved into diff orgs/usernames).

KennethEnevoldsen: I added the fix

mehran-sarmadi:
Hi @KennethEnevoldsen, just wanted to check in — all requested changes are resolved now, including the commit you pushed.

Since all checks have passed and auto-merge is enabled, would you mind updating your review to “Approve” so it can be merged?

Thanks a lot!

isaac-chung (approved):

PR #2737, opened 2025-05-29
fix: raise error when model not found

KennethEnevoldsen:
1) Raise an error when model not found using get_model_meta instead of returning empty model_meta
2) Added more helpful error messages

```py
import mteb

meta = mteb.get_model_meta("BAAI/bge-m3")
# Before fix: returns empty model meta and raises a warning that it was not found on HF
# expected behaviour: Raise an error

# After fix: Raises the following error:
# KeyError: "Model 'BAI/bge-m3' not found in MTEB registry nor on the Huggingface Hub. Did you mean: 'BAAI/bge-m3' or BAAI/bge-small-zh?"
```

This is technically a breaking change, I can move it to v2?

@x-tabdeveloping was initially developed by you as a part of the leaderboard. I couldn't find a place in the leaderboard where it is called, but I want to double check with you.

### Checklist

- [x] I did not add a dataset, or if I did, I added the [dataset checklist](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#submit-a-pr) to the PR and completed it.
- [x] I did not add a model, or if I did, I added the [model checklist](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_model.md#submitting-your-model-as-a-pr) to the PR and completed it.

### DIFF
--- a/mteb/models/overview.py
+++ b/mteb/models/overview.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
+import difflib
 import logging
 from collections.abc import Iterable
-from functools import lru_cache
 from typing import Any
 
 from huggingface_hub import ModelCard
+from huggingface_hub.errors import RepositoryNotFoundError
 from sentence_transformers import SentenceTransformer
 
 from mteb.abstasks.AbsTask import AbsTask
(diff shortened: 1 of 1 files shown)


x-tabdeveloping:
Seems like a reasonable idea to me. If it works and doesn't break anything then go ahead. I might stick to v2 though

Samoed:
Yes, I think this is breaking. Even our tests can't handle it correctly

KennethEnevoldsen:
Will move it to v2

PR #2745, opened 2025-05-31
docs: Updated description of FEVER

KennethEnevoldsen:
Update the description to state that the corpus is the same as fever as we have have [multiple questions on it](https://huggingface.co/datasets/mteb/climate-fever/discussions/2)


david-koleckar:
typo on dataset card... "The underlying corpus is the same as FVER."

PR #2748, opened 2025-05-31
fix: Specify dependencies for nomic models and ensure that they can load

KennethEnevoldsen:
- added dependencies to .toml
- added require_package for error message
- reformatted the function to ensure the interface

@Samoed considered making all models implementations private (this would make a change such as the one I just made non-breaking). WDYT?


Samoed:
> considered making all models implementations private

What do you mean by it? That users shouldn't import them directly? If so, I think that's good

KennethEnevoldsen:
Yea that is exactly what I meant. 

After having a second thought on it, I actually don't think it is needed (we haven't seen a direct import), so it might not be worth the change, but if it starts happening, then we might consider changing it.

PR #2760, opened 2025-06-02
Merge branch 'main' into v2.0.0

KennethEnevoldsen:
@Samoed I have left two files unresolved (Any2AnyRetrieval.py and colbert_models.py). Will you have a look at these?

Also I believe we might be missing some task imports for retrieval

- lint and tests will fails due to the missing conflicts. I can fix the tests afterwards


Samoed:
Yes, I will look. Also, `colpali` needed to be refactored, but I think we can do it later

KennethEnevoldsen:
I will merge this

PR #2763, opened 2025-06-03
fix: Add IFIR relevant tasks

SighingSnow:
- [x] I have tested that the dataset runs with the mteb package.

- [x] I have run the following models on the task (adding the results to the pr). These can be run using the mteb run -m {model_name} -t {task_name} command.
  - [x] sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2
  - [x] intfloat/multilingual-e5-smal

### DIFF
--- a/mteb/benchmarks/benchmarks.py
+++ b/mteb/benchmarks/benchmarks.py
@@ -274,6 +274,33 @@ MTEB_RETRIEVAL_WITH_INSTRUCTIONS = Benchmark(
 """,
 )
 
+MTEB_RETRIEVAL_WITH_DOMAIN_INSTRUCTIONS = Benchmark(
+    name="IFIR",
+    display_name="IFIR",
+    tasks=get_tasks(
+        tasks=[
+            "IFIRAila",
+            "IFIRCds",
+            "IFIRFiQA",
+            "IFIRFire",
+            "IFIRNFCorpus",
+            "IFIRPm",
+            "IFIRScifact",
+        ]
+    ),
+    description="Benchmark for Evaluating Instruction-Following in Expert-Domain Information Retrieval",
+    reference="https://arxiv.org/abs/2503.04644",
+    citation=r"""
+@inproceedings{song2025ifir,
+  title={IFIR: A Comprehensive Benchmark for Evaluating Instruction-Following in Expert-Domain Information Retrieval},
+  author={Song, Tingyu and Gan, Guo and Shang, Mingsheng and Zhao, Yilun},
+  booktitle={Proceedings of the 2025 Conference of the Nations of the Americas Chapter of the Association for Computational Linguistics: Human Language Technologies (Volume 1: Long Papers)},
+  pages={10186--10204},
+  year={2025}
+}
+""",
+)
+
 MTEB_RETRIEVAL_LAW = Benchmark(
     name="MTEB(Law, v1)",  # This benchmark is likely in the need of an update
     display_name="Legal",
--- a/mteb/tasks/Retrieval/eng/IFIRAilaRetrieval.py
+++ b/mteb/tasks/Retrieval/eng/IFIRAilaRetrieval.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+from ....abstasks.AbsTaskRetrieval import AbsTaskRetrieval
+
+
+class IFIRAila(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="IFIRAila",
+        dataset={
+            "path": "if-ir/aila",
+            "revision": "6bce742",
+        },
+        description="Benchmark IFIR aila subset within instruction following abilities.",
+        reference="https://arxiv.org/abs/2503.04644",
+        type="Retrieval",
+        category="s2p",
+        modalities=["text"],
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="ndcg_at_20",
+        date=None,
+        # domains=["Medical", "Academic", "Written"],
+        task_subtypes=None,
+        license=None,
+        annotations_creators=None,
+        dialect=None,
+        sample_creation=None,
+        bibtex_citation=r"""
+@inproceedings{song2025ifir,
+  title={IFIR: A Comprehensive Benchmark for Evaluating Instruction-Following in Expert-Domain Information Retrieval},
+  author={Song, Tingyu and Gan, Guo and Shang, Mingsheng and Zhao, Yilun},
+  booktitle={Proceedings of the 2025 Conference of the Nations of the Americas Chapter of the Association for Computational Linguistics: Human Language Technologies (Volume 1: Long Papers)},
+  pages={10186--10204},
+  year={2025}
+}
+""",
+    )
--- a/mteb/tasks/Retrieval/eng/IFIRCdsRetrieval.py
+++ b/mteb/tasks/Retrieval/eng/IFIRCdsRetrieval.py
@@ -1,0 +1,38 @@
+from __future__ import annotations
+
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+from ....abstasks.AbsTaskRetrieval import AbsTaskRetrieval
+
+
+class IFIRCds(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="IFIRCds",
+        dataset={
+            "path": "if-ir/cds",
+            "revision": "c406767",
+        },
+        description="BBenchmark IFIR cds subset within instruction following abilities.",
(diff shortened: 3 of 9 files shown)


mteb/tasks/Retrieval/eng/IFIRCdsRetrieval.py
```diff
@@ -0,0 +1,38 @@
...
+        },
+        description="BBenchmark IFIR cds subset within instruction following abilities.",
+        reference="https://arxiv.org/abs/2503.04644",
+        type="Retrieval",
+        category="s2p",
+        modalities=["text"],
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="ndcg_at_20",
+        date=None,
```
KennethEnevoldsen, line 23: There are still a bunch of annotations missing. These will need to be filled out

mteb/tasks/Retrieval/eng/IFIRCdsRetrieval.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class IFIRCds(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="IFIRCds",
+        dataset={
+            "path": "if-ir/cds",
+            "revision": "c406767",
+        },
+        description="BBenchmark IFIR cds subset within instruction following abilities.",
```
KennethEnevoldsen, line 15: The description has to be detailed enough for me to get a gist of what the dataset contains.

mteb/benchmarks/benchmarks.py
```diff
@@ -274,6 +274,33 @@
 """,
 )
 
+MTEB_RETRIEVAL_WITH_DOMAIN_INSTRUCTIONS = Benchmark(
```
KennethEnevoldsen, line 277: If we create the benchmark you will have to submit some models that have been evaluated on it as well

mteb/tasks/Retrieval/eng/IFIRAilaRetrieval.py
```diff
@@ -0,0 +1,39 @@
...
+        description="Benchmark IFIR aila subset within instruction following abilities.",
+        reference="https://arxiv.org/abs/2503.04644",
+        type="Retrieval",
+        category="s2p",
+        modalities=["text"],
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="ndcg_at_20",
+        date=None,
+        # domains=["Medical", "Academic", "Written"],
```
KennethEnevoldsen, line 24: Please, do annotate domains

KennethEnevoldsen (commented):
Congratulations on the paper and thanks for the PR! There seems to be a few missing things but nothing major. Running a couple of models on the benchmark would be great.

I examined a bit of the data to figure out how you handled the instruction without using the instruction retrieval task, but found a few odd cases:

> How Beans Help Our Bones I am a student researching the impact [...]

I suspect it is:

> Query: How Beans Help Our Bones 
> Instruction: I am a student researching the impact [...]

The casing also seems like a (consistent) formatting issue.

@orionw since it is instruction following I just want to get your eyes on this as well.

mteb/tasks/Retrieval/eng/IFIRCdsRetrieval.py
```diff
@@ -0,0 +1,38 @@
...
+        reference="https://arxiv.org/abs/2503.04644",
+        type="Retrieval",
```
orionw, line 17: This should be changed to InstructionRetrieval

SighingSnow:
> My understanding of IFIR is that there is one instruction per query but that instruction was done at three hierarchies, is that right

Yes, but we only provide three hierarchies in four subset(fiqa, aila, scifact_open and pm). 

> but I think it would make more sense as InstructionRetrieval of course.

Yes I will check it in more detail. Should I use make all these 7 datasets to InstructionRetrieval.

mteb/tasks/Retrieval/eng/IFIRAilaRetrieval.py
SighingSnow: Yes, no problem.

mteb/benchmarks/benchmarks.py
SighingSnow: Yes, sure. I will further run some models on it.

mteb/tasks/Retrieval/eng/IFIRCdsRetrieval.py
SighingSnow: I saw there is a [comment](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/abstasks/AbsTaskInstructionRetrieval.py#L324), that ``AbsTaskInstructionRetrieval`` will be merged with Retrieval in v2.0.0. So do I still need move it to the ``InstructRetrieval`` folder.

orionw: Ah I see - I thought this was already going into v2. We have all the updates to instruction retrieval there, which could be causing some confusion. 

@KennethEnevoldsen does it make sense to put this in v2? We are switching soon anyways IIUC.

KennethEnevoldsen:
Yeah, so I think it would be best to move this to v2. I plan to have it merged during the summer holidays. @Samoed would any issues be transferring this that I don't know of?

Samoed:
I don't think there are any issues. The dataset is in the correct format and should be easy to integrate.

KennethEnevoldsen:
Would moving it over work for you @SighingSnow?

SighingSnow:
> Would moving it over work for you

Yes, no problem. I will try to get some results and fix my dataset format as @orionw mentioned.

SighingSnow:
Hey @orionw , should I follow the [InstructIR](https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/tasks/InstructionRetrieval/eng/InstructIR.py) in the v2.0.0 branch to format my data. 
For 3 sub-tasks, I have only one level. For the other remaining 4 sub-tasks, I have 3 level of instructions. So may I include all the instructions in this dataset ?

orionw:
> For 3 sub-tasks, I have only one level. For the other remaining 4 sub-tasks, I have 3 level of instructions. So may I include all the instructions in this dataset ?

Just to confirm: you want separate evaluation numbers for each instruction setting?  E.g. level 1 has a score of X, level 2 has a score of Y, etc. If this is case I think the easiest way to do:

A) to create three datasets on HF when there are three versions of the instructions (with same everything except different instructions) and put three versions of it in the task file that link to those three datasets (like this one that has two [versions](https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/tasks/InstructionReranking/multilingual/mFollowIR.py#L156)). The downside is you have to do a bit of duplication on HF since the corpora/qrels/queries are the same and only the instructions differ.

You could also do (more complicated but useful if you have mixed metrics):

B) If you want mixed numbers (like the microaverage of two versions of the instructions) you can do it by duplicating the queries with some identifier (I.e. `query_23_instruction_v1` and `query_23_instruction_v2`) and then do some aggregation in the end with the `task_specific_scores` function (e.g. if "_v1" in query calculate v1 score, OR if "_v1" or "_v2" in queries and "_v3" not in queries etc.). This is a bit more complicated, but can let you do arbitrary things with the scores.

SighingSnow:
Got it. I will take the second choice. Hope I can finish this job this weekend. Thank you for your timely reply and help!

mteb/tasks/Retrieval/eng/IFIRCdsRetrieval.py
SighingSnow: > to create a separate split of the dataset mapping query_id to instruction 

Should it be similar to [InstructIR](https://huggingface.co/datasets/mteb/InstructIR-mteb/viewer/instruction?views%5B%5D=instruction) in InstructIR-mteb.

orionw: Yes, same format (two columns) although it will differ according to your setup. They have 10 instructions for each query they aggregate into one score per query. So you could aggregate yours however you'd prefer. 

Really what's important is that the number you get at the end is the score you want people using for your benchmark

SighingSnow:
@KennethEnevoldsen Sorry for the bothering, should I rebase my code based on the 2.0.0 branch ? 
I mean, if I use the main branch code. My class will derived from AbsInstructionRetrieval class. This class is designed speficially for FollowIR.

KennethEnevoldsen:
@SighingSnow changed the base to v2.0.0 you should be able to simpy `git merge origin v2.0.0`

SighingSnow:
Thanks for all your patience! I have created [a new PR ](https://github.com/embeddings-benchmark/mteb/pull/2815) on the branch v2.0.0. Looking forward to receive your feedback. 
And I am glad to take your comments then.

Issue #2770, opened 2025-06-04
Caltech101ZeroShot fails with missing names


wwu137:
After https://github.com/embeddings-benchmark/mteb/pull/2759, Caltech101ZeroShot will fail with the following error:

```
 results, tick, tock = self._run_eval(
  File "/usr/local/lib/python3.10/dist-packages/mteb/evaluation/MTEB.py", line 307, in _run_eval
    results = task.evaluate(
  File "/usr/local/lib/python3.10/dist-packages/mteb/abstasks/AbsTask.py", line 137, in evaluate
    scores[hf_subset] = self._evaluate_subset(
  File "/usr/local/lib/python3.10/dist-packages/mteb/abstasks/Image/AbsTaskZeroShotClassification.py", line 128, in _evaluate_subset
    candidate_labels = self.get_candidate_labels()
  File "/usr/local/lib/python3.10/dist-packages/mteb/tasks/Image/ZeroShotClassification/eng/Caltech101.py", line 56, in get_candidate_labels
    for name in self.dataset["test"].features[self.label_column_name].names
AttributeError: 'Value' object has no attribute 'names'
```

@KennethEnevoldsen  could you please help take a look?

Samoed:
I think previously in `label` column where label-names were for each label was a name, but now there only labels

KennethEnevoldsen:
Ah sorry, this is my mistake - @Samoed is the solution to reupload the dataset independently? Or is there a way where we can automatically upload it such that they can share the source?

Samoed:
There can be 3 ways
1. Re-upload separately
2. Create independent column in same dataset with text description
3. Add [`ClassLabel`](https://huggingface.co/docs/datasets/v3.6.0/en/package_reference/main_classes#datasets.ClassLabel) mapping to label column

PR #2777, opened 2025-06-05
fix: Update Caltech101 datasets to latest revision [v2]

KennethEnevoldsen:
related: #2770

Fixes the issue, but only in v2

```
# tested using:

task: mteb.AbsTask = mteb.get_task("Caltech101ZeroShot")
task.load_data()
task.get_candidate_labels()
```

### DIFF
--- a/mteb/tasks/Image/ZeroShotClassification/eng/Caltech101.py
+++ b/mteb/tasks/Image/ZeroShotClassification/eng/Caltech101.py
@@ -13,7 +13,7 @@ class Caltech101ZeroShotClassification(AbsTaskZeroShotClassification):
         reference="https://ieeexplore.ieee.org/document/1384978",
         dataset={
             "path": "mteb/Caltech101",
-            "revision": "52439cf6d4f6ebf563d8cdc7f2c5371d9efd2686",
+            "revision": "20a80c5a6993dce5c2a5c3135f96c625477b34a2",
         },
         type="ZeroShotClassification",
         category="i2t",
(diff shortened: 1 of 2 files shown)


mteb/tasks/Image/ZeroShotClassification/eng/Caltech101.py
```diff
@@ -46,6 +44,10 @@ class Caltech101ZeroShotClassification(AbsTaskZeroShotClassification):
   year = {2004},
 }
 """,
+        descriptive_stats={
+            "n_samples": {"test": 6084},
+            "avg_character_length": {"test": 431.4},
+        },
```
Samoed: `descriptive_stats` should be stored in `json` 
```suggestion
```

mteb/tasks/Image/ZeroShotClassification/eng/Caltech101.py
```diff
@@ -12,10 +12,8 @@ class Caltech101ZeroShotClassification(AbsTaskZeroShotClassification):
         description="Classifying images of 101 widely varied objects.",
         reference="https://ieeexplore.ieee.org/document/1384978",
         dataset={
-            "path": "HuggingFaceM4/Caltech-101",
-            "name": "with_background_category",
-            "revision": "851374102055782c84f89b1b4e9d128a6568847b",
-            "trust_remote_code": True,
+            "path": "mteb/Caltech101",
+            "revision": "20a80c5a6993dce5c2a5c3135f96c625477b34a2",
```
Samoed: I don't see labels as text in dataset

KennethEnevoldsen: It should be fixed now

mteb/tasks/Image/ZeroShotClassification/eng/Caltech101.py
KennethEnevoldsen: removed

Samoed (commented):
Why do we need to commit this to `v2` diirectly? Maybe it would be better to just merge `main` to `v2` with this fix?

Issue #2416, opened 2025-03-23
Update ruff rules


Samoed:
I think we can enable:
- [x] [`T20`](https://docs.astral.sh/ruff/rules/#flake8-print-t20) - flake8-print print found
- [X] [`B`](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b) - flake8-bugbear. https://github.com/embeddings-benchmark/mteb/pull/5189, https://github.com/embeddings-benchmark/mteb/pull/5205
- [X] [`A`](https://docs.astral.sh/ruff/rules/#flake8-builtins-a) - flake8-builtins. To see if shadow builtins functions https://github.com/embeddings-benchmark/mteb/pull/5328
- [x] [`LOG`](https://docs.astral.sh/ruff/rules/#flake8-logging-log) - flake8-logging. Simple logging checkers https://github.com/embeddings-benchmark/mteb/pull/3820
- [X] [`ISC`](https://docs.astral.sh/ruff/rules/#flake8-implicit-str-concat-isc) - flake8-implicit-str-concat https://github.com/embeddings-benchmark/mteb/pull/5181
- [X] [`ICN`](https://docs.astral.sh/ruff/rules/#flake8-import-conventions-icn) - flake8-import-conventions https://github.com/embeddings-benchmark/mteb/pull/5108
- [x] [`PT`](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt) - flake8-pytest-style https://github.com/embeddings-benchmark/mteb/pull/5238
- [X] [`RET`](https://docs.astral.sh/ruff/rules/#flake8-return-ret) - flake8-return https://github.com/embeddings-benchmark/mteb/pull/5335
- [X] [`SIM`](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim) - flake8-simplify https://github.com/embeddings-benchmark/mteb/pull/5336
- [x] [`PTH`](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth) - flake8-use-pathlib https://github.com/embeddings-benchmark/mteb/pull/3197
- [x] [`NPY`](https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy) - NumPy-specific rules https://github.com/embeddings-benchmark/mteb/commit/04d0cf423d178273b881fce12a500f2fde507c4a
[...]

PR #2784, opened 2025-06-07
bump ruff

Samoed:
Update ruff from `0.9.7` to `0.11.13`
### Checklist

- [ ] I did not add a dataset, or if I did, I added the [dataset checklist](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#submit-a-pr) to the PR and completed it.
- [ ] I did not add a model, or if I did, I added the [model checklist](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_model.md#submitting-your-model-as-a-pr) to the PR and completed it.

### DIFF
--- a/docs/create_tasks_table.py
+++ b/docs/create_tasks_table.py
@@ -79,7 +79,7 @@ def create_task_lang_table(tasks: list[mteb.AbsTask], sort_by_sum=False) -> str:
             if lang in PROGRAMMING_LANGS:
                 lang = "code"
             if table_dict.get(lang) is None:
-                table_dict[lang] = {k: 0 for k in sorted(get_args(TASK_TYPE))}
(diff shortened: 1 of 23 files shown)


KennethEnevoldsen:
I understant the bump, but maybe worth resolving v2 first? Or at least making a merge directly afterwards (I don't feel like we are in a rush to do it now).

Samoed:
Do you mean that we should firstly finish `v2` before bumping `ruff`?

KennethEnevoldsen:
We already have #2416, so yeah, waiting is probably reasonable

Samoed:
This is updating only the base version to make it possible to use more rules, because this is inconvenient to work locally with different ruff versions

KennethEnevoldsen:
Hmm right, but feel free to merge this, but then we should do an immediate merge to v2.0.0 as well

PR #2740, opened 2025-05-30
Add R2MED Benchmark

ll0ruc:
### Checklist

Reason for dataset addition: (R2MED dataset, https://huggingface.co/R2MED) R2MED, the first benchmark explicitly designed for reasoning-driven medical retrieval.  More details in [Paper](https://arxiv.org/abs/2505.14558), [Homepage](https://r2med.github.io/)

- [x] I did not add a dataset, or if I did, I added the [dataset checklist](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#submit-a-pr) to the PR and completed it.
- [x] I have tested that the dataset runs with the mteb package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the mteb run -m {model_name} -t {task_name} command.
    - [x] sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2
    - [x] intfloat/multilingual-e5-small
- [x]  I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] I have considered the size of the dataset and reduced it if it is too big (2048 examples is typically large enough for most tasks)
- [x] I did not add a model, or if I did, I added the [model checklist](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_model.md#submitting-your-model-as-a-pr) to the PR and completed it.

PR #2795, opened 2025-06-09
dataset: Add R2MED Benchmark

KennethEnevoldsen:
PR for #2740 (can't reopen the original)

@ll0ruc tried to resolve the PR, but couldn't, so reverted the merge (thought it was an error on our end). You are missing the `date` in your metadata. Will you add it?


ll0ruc:
> PR for #2740 (can't reopen the original)
> 
> @ll0ruc tried to resolve the PR, but couldn't, so reverted the merge (thought it was an error on our end). You are missing the `date` in your metadata. Will you add it?

@KennethEnevoldsen I noticed that the TaskMetadata class specifies date as a tuple of two dates, described as date: `The date when the data was collected. Specified as a tuple of two dates`. I wanted to clarify:

Does the first date refer to when the original data was collected or published, and the second to when the R2MED subset was created?

KennethEnevoldsen:
It is the range of when the original data was created (so if you are using tweets from 2000-2020  that is your range).

ll0ruc:
I’ve added the `date` field for all R2MED tasks as requested.

KennethEnevoldsen:
@ll0ruc we just need the lint, and then we are good (I think it is the missing newline at the end of the file)

ll0ruc:
@KennethEnevoldsen ok, It looks like the lint is already successful

KennethEnevoldsen:
@ll0ruc great! Lovely to get this added

PR #2799, opened 2025-06-09
[v2] refactor types into a module

### LINKED ISSUE #1822: [v2] Remove `LANG_MAPPING`
For now we have `LANG_MAPPING` and `LanguageScripts`. I think we can only have one

KennethEnevoldsen:
Decided not to move types in TaskMetadata that was only used for task metadata.

However was thinking of refactoring TaskMetadata into:
```
task_metadata
| - __init__.py
| - task_metadata.py
| - aggregate_task_metadata.py
```
this should prevent cases where TaskMetadata is imported from `mteb.abstasks` (already removed it from the abstasks init).

Also considered removing the `ScoringFunction` as it adds a lot of interdependence (it is also the only metadata object that it not json serializable by default).

Things that I know are missing:
- METRIC (in `mteb/task_selection.py`): Script could probably be removed (#2809)

Notes:
- `normalize_embeddings.py` was unused so removed it
- `languages.py` could probably be refactored into a module (added issue #2808) 

fixes #1822


Samoed:
> METRIC (in mteb/task_selection.py): Script could probably be removed

I can't find usage of these functions too, so this can be removed. 

> normalize_embeddings.py seem unused. Can I remove this?

I think yes

> languages.py could probably be refactored into a module (to contains the relevant json files)

Yes, I think languages needs to be refactored properly https://github.com/embeddings-benchmark/mteb/issues/1822

mteb/types/_metadata.py
```diff
@@ -1,22 +1,20 @@
...
-    "image",
-]
 
-http_url_adapter = TypeAdapter(AnyUrl)
-STR_URL = Annotated[
-    str, BeforeValidator(lambda value: str(http_url_adapter.validate_python(value)))
-]  # Allows the type to be a string, but ensures that the string is a URL
+## LANGUAGE TYPES ##
+ISOLanguageScript = str  # a 3-letter ISO 639-3 language code followed by a 4-letter ISO 15924 script code (e.g. "eng-Latn")
+ISOLanguage = str  # a 3-letter ISO 639-3 language code
```
Samoed, line 10: I think this can be updated to
```python
import annotated_types as at

ISOLanguageScript = Annotated[str, at.Len(3, 3)]  # a 3-letter ISO 639-3 language code followed by a 4-letter ISO 15924 script code (e.g. "eng-Latn")
ISOLanguage = Annotated[str, at.Len(3, 3)]  # a 3-letter ISO 639-3 language code
ISOScript = Annotated[str, at.Len(4, 4)]  # a 4-letter ISO 15924 script code
```
This library is used by pydantic, so we don't need to install something

mteb/abstasks/TaskMetadata.py
```diff
@@ -323,18 +237,18 @@ class TaskMetadata(BaseModel):
...
 
     eval_splits: list[str] = ["test"]
-    eval_langs: LANGUAGES
+    eval_langs: Languages
     main_score: str
 
-    date: tuple[STR_DATE, STR_DATE] | None = None
+    date: tuple[StrDate, StrDate] | None = None
     domains: list[TASK_DOMAIN] | None = None
     task_subtypes: list[TASK_SUBTYPE] | None = None
```
Samoed: Should we also update `TASK_TYPE`, `TASK_DOMAIN`, `TASK_SUBTYPE`?

isaac-chung (commented):
Looks great so far. 

1. I too believe normalize_embedding.py is unused. Good to remove.
2. There's a test failure: `ERROR tests/test_load_results/test_task_results.py - TypeError: 'module' object is not callable`. Maybe it's an import thing?

mteb/abstasks/TaskMetadata.py
isaac-chung: How do you mean?

Samoed: I mean rename them like `TaskType` and move out from `TaskMetadata`

mteb/types/_metadata.py
KennethEnevoldsen: This will sadly not work as some languages are programming languages (which do not have ISO codes). We do validate these in the relevant objects.

mteb/abstasks/TaskMetadata.py
KennethEnevoldsen: Will go through and rename for consistency, will also check for a ruff lint rule

KennethEnevoldsen:
Fixed import issue, removed normalize_embeddings.py and for METRIC, I added another issue to it ( #2809)

> Yes, I think languages needs to be refactored properly https://github.com/embeddings-benchmark/mteb/issues/1822

Added this to fixes, and opened a new issue on language refactors (#2808) 

Anything else on this PR?

isaac-chung:
Should be good if tests pass.

PR #2803, opened 2025-06-10
fix: Adding client arg to init method of OpenAI models wrapper

malteos:
This PR adds the OpenAI client arg to init method (e.g., for already initialized AzureOpenAI client). 

The arg is needed if you want to use with a different client, e.g. OpenAI embedding models via Azure.

### DIFF
--- a/mteb/models/openai_models.py
+++ b/mteb/models/openai_models.py
@@ -27,7 +27,7 @@ class OpenAIWrapper(Wrapper):
         max_tokens: int,
         tokenizer_name: str = "cl100k_base",
         embed_dim: int | None = None,
-        client: None | Any = None,
+        client: Any | None = None,  # OpenAI
         **kwargs,
     ) -> None:
         """Wrapper for OpenAIs embedding API.


mteb/models/openai_models.py
```diff
@@ -27,6 +27,7 @@ def __init__(
         max_tokens: int,
         tokenizer_name: str = "cl100k_base",
         embed_dim: int | None = None,
+        client: None | Any = None,
```
Samoed: I think this is simplier
```suggestion
        client: OpenAI = OpenAI(),
```

mteb/models/openai_models.py
```diff
@@ -49,7 +50,10 @@ def __init__(
         )
         import tiktoken
 
-        self._client = OpenAI()
+        if client is None:
+            self._client = OpenAI()
+        else:
+            self._client = client  # use client from args (e.g., already initialized AzureOpenAI client)
```
KennethEnevoldsen:
```suggestion
```

KennethEnevoldsen (approved):

mteb/models/openai_models.py
KennethEnevoldsen: We don't have OpenAI defined in this scope (due to optional dependency)

PR #2813, opened 2025-06-11
fix: refactor `languages.py` [v2]

### LINKED ISSUE #2808: [v2] refactor `languages.py`

languages.py could probably be refactored into a module (to contains the relevant json files)

Originally mentioned here: https://github.com/embeddings-benchmark/mteb/pull/2799

KennethEnevoldsen:
Refactored languages.py into a module. This should be fully backwards compatible.

Considered if I should make the inner files `_*.py` to encourage people to import from `mteb.languages`, but I think this is reasonably stable as is.

Fixes #2808


mteb/languages/iso_mappings.py
```diff
@@ -0,0 +1,25 @@
...
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+# Language mappings
+path_to_lang_codes = Path(__file__).parent / "iso_639_3_to_language.json"
+path_to_lang_scripts = Path(__file__).parent / "iso_15924_to_script.json"
+path_to_lang_fam = Path(__file__).parent / "language_family.json"
```
Samoed, line 13: I think you need to move json files too

mteb/languages/check_language_code.py
```diff
@@ -0,0 +1,29 @@
...
+            return  # override for code
+        else:
+            raise ValueError(
+                f"Programming language {lang} is not a valid programming language."
+            )
+    if lang not in ISO_TO_LANGUAGE:
+        raise ValueError(
+            f"Invalid language code: {lang}, you can find valid ISO 639-3 codes in {path_to_lang_codes}"
+        )
+    if script not in ISO_TO_SCRIPT:
```
Samoed, line 26: Do we need to support script part? I think if we have 1 language it will use 1 script most of the time.

mteb/languages/iso_mappings.py
isaac-chung: Ya - think this is causing test failures.

mteb/languages/check_language_code.py
KennethEnevoldsen: Some languages use multiple scripts and I could imagine that we could analyze the influence of script change on embedding quality.

mteb/languages/iso_mappings.py
KennethEnevoldsen: solved

PR #2819, opened 2025-06-15
fix: Reuploaded previously unavailable SNL datasets

### LINKED ISSUE #2477: Dataset navjordj/SNL_summarization no longer available
Seems like the dataset navjordj/SNL_summarization is no longer available. 

I have written to the authors so if why it is the case.

KennethEnevoldsen:
closes #2477


Samoed:
Can you check if this version of the dataset produces the same results as the previous one?

KennethEnevoldsen:
added it! Anything else?

Samoed (approved):
Great!

PR #2823, opened 2025-06-15
[v2] Refactor descriptive stats

Samoed:
Continue ideas from https://github.com/embeddings-benchmark/mteb/pull/2537. Make all statistics in format
```json
{
    "text_statistics": {"min_text_length": 1, ... },
    "label_statistics": {"min_labels_per_text": 1, ...},
}
```

For now, I haven't changed MIEB statistics


mteb/abstasks/statistics_calculation.py
KennethEnevoldsen, line 1: Should we make this module + methods private?

mteb/abstasks/statistics_calculation.py
```diff
@@ -0,0 +1,97 @@
...
+        img_widths.append(width)
+
+    return ImageStatistics(
+        min_image_width=min(img_widths),
+        average_image_width=sum(img_widths) / len(img_widths),
+        max_image_width=max(img_widths),
+        min_image_height=min(img_heights),
+        average_image_height=sum(img_heights) / len(img_heights),
+        max_image_height=max(img_heights),
+    )
```
KennethEnevoldsen: I am not sure if these are the statistics that people would normally be interested in?

Should we add duplicates as well?

KennethEnevoldsen (commented):
Looks good - we might discuss what are relevant image metrics, but the structue change it very decent

mteb/abstasks/statistics_calculation.py
Samoed: I've added it as it was previously in MIEB.  I think we can expand it, but I'm not sure what to add. CC @isaac-chung

KennethEnevoldsen: I would like duplicates - otherwise, I don't think I have a lot.

isaac-chung: What was previously in MIEB is fine

isaac-chung:
> Looks good - we still need to resolve the conflicts though. I think we can add image metrics in another PR if needed
> 
>  @isaac-chung any image metrics that we need to add?

Can't think of any extra from the MIEB existing ones. Think this PR just needs to resolve the conflicts and that's it.

PR #2826, opened 2025-06-16
model: add Hakim and TookaSBERTV2 models

mehran-sarmadi:
In this PR, we add five models: three from the Hakim family (hakim, hakim-small, and hakim-unsup), and two from the TookaSBERTV2 family (TookaSBERT-V2-Small and TookaSBERT-V2-Large).


mteb/models/fa_models.py
```diff
@@ -166,3 +166,43 @@
...
+    similarity_fn_name="cosine",
+    framework=["Sentence Transformers", "PyTorch"],
+    use_instructions=False,
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
+)
+
+tooka_sbert_v2_large = ModelMeta(
+    name="PartAI/Tooka-SBERT-V2-Large",
```
Samoed, line 191: Can you add loader to models?

mteb/models/mcinext_models.py
```diff
@@ -0,0 +1,578 @@
...
+    def __init__(self, message: str, status_code: int | None = None):
+        super().__init__(
+            f"API Error: {message} (Status Code: {status_code})"
+            if status_code
+            else f"API Error: {message}"
+        )
+        self.status_code = status_code
+
+
+class OurInstructModelWrapper(Wrapper):
```
Samoed:
```suggestion
class HakimModelWrapper(Wrapper):
```

mteb/models/mcinext_models.py
```diff
@@ -0,0 +1,578 @@
...
+    "hakim": "Hakim",
+    "hakim-small": "Hakim_small",
+    "hakim-unsup": "Hakim_unsuper",
+}
+
+# Task configurations
+TASK_CONFIGS = {
+    "1_1": (
+        "دسته بندی , تحلیل احساس رضایت کاربر در مکالمه با چت بات",
+        ["عالی", "خوب", "متوسط", "بد", "خیلی بد"],
```
Samoed: What mean second argument of your cofigs? You don't use it in your code

mteb/models/mcinext_models.py
```diff
@@ -0,0 +1,578 @@
...
+    "3_14": (
+        "تشخیص ارتباط , متن اول مکالمه ی کاربر با چت بات است. آیا متن دوم موضوع استخراج شده ی متن اول است ؟",
+        None,
+    ),
+}
+
+# Dataset task mappings
+DATASET_TASKS = {
+    "PersianTextEmotion": (1, 18),
+    "PersianFoodSentimentClassification": (1, 190),
```
Samoed: Can you add instructions directly to the task names?

mteb/models/fa_models.py
mehran-sarmadi: These models should be fine loading with the default Sentence Transformers loader. In that case, is it still necessary to add a specific loader for them?

mteb/models/mcinext_models.py
mehran-sarmadi: Oh, yes, I change it now.

mteb/models/fa_models.py
Samoed: It's not nesessary, but will be simplier understand what they're using

mehran-sarmadi: I will add it.

mteb/models/mcinext_models.py
mehran-sarmadi: You're right, the second argument in the configs refers to the class labels of the data. We tried several instruction setups, and in some of them those labels were necessary, while in others we left them out to compare results. In this final version (which gave us the best results), they’re actually not needed anymore. So yes, you're absolutely right, I can remove them to make the code cleaner.

mteb/models/mcinext_models.py
mehran-sarmadi: Sure, if you think that would be better, I can make that change as well

mteb/models/fa_models.py
mehran-sarmadi: Now that I’ve thought about it more, I’d prefer not to add a SentenceTransformers wrapper. I’ve done my testing without it, and other Farsi models are set up the same way. It helps maintain consistency and doesn’t seem functionally necessary. What do you think?

Samoed: Ok, we can don't add to them

mteb/models/mcinext_models.py
```diff
@@ -0,0 +1,504 @@
...
+    "BeytooteClustering": ("دسته بندی , دسته بندی موضوعی متن", 1),
+    "DigikalamagClustering": ("دسته بندی , دسته بندی موضوعی متن", 1),
+    "NLPTwitterAnalysisClustering": ("دسته بندی , دسته بندی موضوعی متن", 1),
+    "HamshahriClustring": ("دسته بندی , دسته بندی موضوعی متن", 1),
+    "SIDClustring": ("دسته بندی , دسته بندی موضوعی متن", 1),
+    "MIRACLReranking": ("تشخیص ارتباط , آیا متن دوم پاسخ متن اول است ؟", 3),
+    "WikipediaRerankingMultilingual": (
+        "تشخیص ارتباط , آیا متن دوم پاسخ متن اول است ؟",
+        3,
+    ),
```
Samoed, line 131: I think this is still a bit strange that order of id and instruction is changing

mehran-sarmadi: It looks like a display issue due to mixing Persian (RTL) and English (LTR). If you copy the code into VS Code, it should show correctly.
image

Samoed: Interesting bug

mteb/models/mcinext_models.py
```diff
@@ -0,0 +1,504 @@
...
+                logger.error(f"Failed to process batch starting at index {i}: {e}")
+                raise e
+
+        logger.info(
+            f"Encoding completed successfully for {len(all_embeddings)} sentences."
+        )
+        return np.array(all_embeddings, dtype=np.float32)
+
+
+# Model metadata
```
KennethEnevoldsen:
```suggestion
```

KennethEnevoldsen (approved):
nothing major here - looks good!

PR #2828, opened 2025-06-16
refactor task import

KennethEnevoldsen:
This refactors imports following this pattern:

```py
# tasks/__init__
from .Retrieval import *
# tasks/retrieval/__init__
from .eng import *
# tasks/retrieval/eng/__init__
from .task1 import Task1
```
proposed by @Samoed in #2825. This should reduce the number of imports required, while not exposing any of the module requires at the task definition.

If you add a model or a dataset, please add the corresponding checklist:

* [dataset checklist](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#submit-a-pr)
* [model checklist](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_model.md#submitting-your-model-as-a-pr)

Issue #2714, opened 2025-05-22
[v2] Add typechecker


Samoed:
Currently, we have a lot of `# type: ignore` comments but haven’t specified a type checker. In the `v2` branch, we've added more type annotations in most places, and I think we could introduce `mypy` (or `ty`, once it reaches at least version 0.0.1). However, I'm unsure about the configuration—if we set it to `strict`, we might end up with too many errors to resolve easily.

Also maybe soon `v2` version of mypy will be released https://github.com/python/mypy/issues/17967

KennethEnevoldsen:
It sounds like `ty` might be what we should go for. I don't want to add it before it is stable, though. We could, however, add it to the makefile (before adding it to the CI)

Samoed:
`ty` 0.0.2 (first beta) was released https://github.com/astral-sh/ty/releases/tag/0.0.2

Issue #1124, opened 2024-07-30
Evaluators ambiguity: class/module


r0mer0m:
There is ambiguity between classes and modules under the `evalutors` module. Specifically, evaluators objects under [mteb](https://github.com/embeddings-benchmark/mteb/tree/main/mteb)/[evaluation](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/evaluation)/[evaluators](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/evaluation/evaluators) are named the same as the modules which contain them. For example:

`mteb.evaluation.evaluators.RerankingEvaluator` references the class `RerankingEvaluator` in `mteb/evaluation/evaluators/RerankingEvaluator.py`. 

This is problematic because it does not allow to reference the `mteb/evaluation/evaluators/RerankingEvaluator.py` module and therefore referencing any other definition within it.

KennethEnevoldsen:
I agree that this is a general problem within mteb. Ideally, we should have the modules in lowercase to avoid confusing the module with the object. I believe this is also the [pep8 recommendation](https://peps.python.org/pep-0008/#package-and-module-names).

We also have quite a high usage of `*` imports, which also leads to a few oddities e.g.:

```
from mteb import load_dataset
```

I would love a fix for this. Will just check in with @Muennighoff and @isaac-chung as well to make sure that everyone agrees

isaac-chung:
Thanks for bringing this up! I think it's great if we can follow something like pep8. This would lead to a breaking change that might not be backwards compatible, so this would trigger a major version bump. We should consider this for perhaps for 2.0.

Muennighoff:
> This is problematic because it does not allow to reference the mteb/evaluation/evaluators/RerankingEvaluator.py module and therefore referencing any other definition within it.

But there is no other definition within `RerankingEvaluator.py` besides `RerankingEvaluator`, no?

Note that we also name all task & abstask files in this style e.g. https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/Classification/ara/TweetEmotionClassification.py - I think if we change it for evaluators we should also change it there but still unsure if the change is needed. 🤔

r0mer0m:
Thanks @KennethEnevoldsen and @isaac-chung, +1 that following pep8 standards is a good path forwards. 

[...]

PR #2829, opened 2025-06-16
fix: rename TaskMetadata.py to resolve class/module ambiguity

KennethEnevoldsen:
related to: #1124
required for: #2714

Seems like we in multiple places denote the module instead of the intended TaskMetada. This rename should fix that issue

relies on PR #2828


mteb/encoder_interface.py
```diff
@@ -4,10 +4,11 @@
 
 from torch.utils.data import DataLoader
 
+from mteb.abstasks.task_metadata import TaskMetadata
 from mteb.types import Array, BatchedInput, PromptType
 
 if TYPE_CHECKING:
-    from mteb.abstasks import TaskMetadata
+    pass
```
Samoed:
```suggestion
```

Issue #1955, opened 2025-02-04
Tests for MIEB evaluators


Muennighoff:
Should have tests for all new evaluators from MIEB and other things that are kept (https://github.com/embeddings-benchmark/mteb/tree/main/tests/test_evaluators)

fzowl:
@Samoed Can you please take a look if there is anything i missed in the PRs i raised?

Samoed:
This issue about testing evaluators that inside `Image` folder https://github.com/embeddings-benchmark/mteb/tree/main/mteb/evaluation/evaluators/Image. You've added tests only to the main mteb tasks

KennethEnevoldsen:
There is only one image evaluator left (`ImageTextPairClassificationEvaluator`). The rest has been merged with the existing text models (which has tests where relevant)

I would probably close? (@Samoed)

Samoed:
Yes, agree

PR #2838, opened 2025-06-20
Add Classification Evaluator unit test

fzowl:
First step in resolving the https://github.com/embeddings-benchmark/mteb/issues/1955 issue.


tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,569 @@
...
+            scores, _ = eval_limited(model_pytorch)
+            assert "accuracy" in scores
+            assert "f1" in scores
+            if is_binary_after_limit:
+                assert "ap" in scores
+            else:
+                assert "ap" not in scores
+
+
+class TestLogRegClassificationEvaluator:
```
KennethEnevoldsen: I don't think we need the class here. It also leads to duplication across the three different evaluators. Since we in v2 wil reduce the ClassificationEvalutors to just one class (see comment above), let us just focus on just logreg for now, feel free to delete the rest.

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,569 @@
...
+        )
+
+    @pytest.mark.parametrize(
+        "evaluator_fixture, is_binary",
+        [
+            ("eval_logreg_binary", True),
+            ("eval_logreg_multiclass", False),
+        ],
+    )
+    def test_output_structure(self, evaluator_fixture, is_binary, model, request):
```
KennethEnevoldsen: I would probably change this to:

```suggestion
    def test_binary_output_structure(self, evaluator, model, x_train: list[str], y_train: list[int], x_test, x_train, expected_score: float):
```

That way, you can parametrize both the model and the evaluator - you will initialize the evaluator a few more times than you do now, but this way it is very easy to add new test cases. You can also extract whether the task is binary or not from the labels.

Note that it might be easier to define:
```py
ClassificationTestCase:
  x_train: list[str]
  y_train: list[int]
  x_test
  x_train
  expected_score: float

# which leads to:  
def test_binary_output_structure(self, evaluator, model, testcase: ClassificationTestCase):
```

KennethEnevoldsen: We naturally can't check for the expected output here with the random `MockNumpyEncoder()`, however, changing it to:
```
class MockNumpyEncoder(mteb.Encoder):
    def __init__(self):
        self.rng_state = np.random.default_rng(42)

        return self.rng_state.random.rand(len(sentences), 10)
```

should fix the issue and should not introduce new issues.

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,569 @@
...
+            assert "ap" not in scores
+
+    @pytest.mark.parametrize(
+        "evaluator_fixture, is_binary",
+        [
+            ("eval_logreg_binary", True),
+            ("eval_logreg_multiclass", False),
+        ],
+    )
+    def test_score_ranges(self, evaluator_fixture, is_binary, model, request):
```
KennethEnevoldsen: delete, checking for expected values is better

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,569 @@
...
+        evaluator = request.getfixturevalue(evaluator_fixture)
+        scores, _ = evaluator(model)
+        metrics_to_check = ["accuracy", "f1", "f1_weighted"]
+        if is_binary:
+            metrics_to_check.extend(["ap", "ap_weighted"])
+
+        for metric in metrics_to_check:
+            assert 0 <= scores[metric] <= 1
+
+    def test_cache_usage_binary(self, model, eval_logreg_binary):
```
KennethEnevoldsen: Would be great with a docstring on this test (took me some time to decode)

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,569 @@
...
+                "test_logreg_limit_not_binary_train",
+                1,
+                1,
+                1,
+                False,
+                ValueError,
+            ),
+        ],
+    )
+    def test_limit_parameter(
```
KennethEnevoldsen: No need to add a test for the limit parameter, it will be deprecated in v2 (you can see the deprecation warning is already in v1)

KennethEnevoldsen (commented):
Thanks for the PR!

Generally I think this looks really well, but I think we can simplify it down a bit.

In v2 we combine these into a single `ClassificationEvaluator` (to prevent discrepancies across), which has a classifier (sklearn interface) attached. This classifier will by default be logistic, but in principle you could change it out with any sklearn compatible classifier.

tests/test_evaluators/test_ClassificationEvaluator.py
fzowl: Removed

tests/test_evaluators/test_ClassificationEvaluator.py
fzowl: Docstring added

tests/test_evaluators/test_ClassificationEvaluator.py
fzowl: Removed

fzowl:
@KennethEnevoldsen Thank you for the review! Can you please take another look now if the test is better this time?

tests/test_evaluators/test_ClassificationEvaluator.py
fzowl: Removed

tests/test_evaluators/test_ClassificationEvaluator.py
fzowl: @KennethEnevoldsen I refactored this a bit, can you please take a look?

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,194 @@
...
+def model():
+    return MockNumpyEncoder()
+
+
+@pytest.fixture(params=[BINARY_TEST_CASE, MULTICLASS_TEST_CASE])
+def test_case(request):
+    return request.param
+
+
+def test_output_structure(model, test_case: ClassificationTestCase):
```
KennethEnevoldsen:
```suggestion
# Fixtures
@pytest.fixture
def model():
    return MockNumpyEncoder()


@pytest.fixture(params=[BINARY_TEST_CASE, MULTICLASS_TEST_CASE])
def test_case(request):
    return request.param

@pytest.fixture("test_case", [BINARY_TEST_CASE, MULTICLASS_TEST_CASE])
def test_output_structure(model, test_case: ClassificationTestCase):
```

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,194 @@
...
+from mteb.evaluation.evaluators import logRegClassificationEvaluator
+
+
+@dataclass
+class ClassificationTestCase:
+    x_train: list[str]
+    y_train: list[int]
+    x_test: list[str]
+    y_test: list[int]
+    task_name: str
```
KennethEnevoldsen:
```suggestion
```
Not required for the test case

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,194 @@
...
+    expected_score: float | None = None  # For deterministic tests
+
+
+class MockNumpyEncoder(mteb.Encoder):
```
KennethEnevoldsen: I would update the existing MockNumpyEncoder

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,194 @@
...
+)
+
+MULTICLASS_TEST_CASE = ClassificationTestCase(
+    x_train=SENTENCES_TRAIN_MULTICLASS,
+    y_train=Y_TRAIN_MULTICLASS.tolist(),
+    x_test=SENTENCES_TEST_MULTICLASS,
+    y_test=Y_TEST_MULTICLASS.tolist(),
+    task_name="test_logreg_multiclass",
+    expected_score=0.0,  # Expected accuracy with deterministic MockNumpyEncoder
+)
```
KennethEnevoldsen: simplify to:
```suggestion
BINARY_TEST_CASE = ClassificationTestCase(
    x_train=["pos1", "pos2", "neg1", "neg2"],
    y_train=[1, 1, 0, 0],
    x_test=["new pos", "new neg"],
    y_test=[1, 0],
    model= MockNumpyEncoder(),
    expected_score=0.5,
)

MULTICLASS_TEST_CASE = ClassificationTestCase(
    x_train=["cls 1", "still cls 1", "cls 2", "also cls 2", "cls 3", "cls 3 too"],
    y_train=[0, 0, 1, 1, 2, 2],
    x_test=["new cls 1", "new cls 2", "new cls 3"],
    y_test=[0, 1, 2],
    model= MockNumpyEncoder(),
    expected_score=0.0,
)
```

KennethEnevoldsen (approved):
Looking a lot better - Added a few simplifications to make it simpler

tests/test_evaluators/test_ClassificationEvaluator.py
fzowl: @KennethEnevoldsen Done

fzowl:
@KennethEnevoldsen Can you please take a look now?

tests/test_evaluators/test_ClassificationEvaluator.py
```diff
@@ -0,0 +1,155 @@
...
+    expected_score=0.5,
+)
+
+MULTICLASS_TEST_CASE = ClassificationTestCase(
+    x_train=["cls 1", "still cls 1", "cls 2", "also cls 2", "cls 3", "cls 3 too"],
+    y_train=[0, 0, 1, 1, 2, 2],
+    x_test=["new cls 1", "new cls 2", "new cls 3"],
+    y_test=[0, 1, 2],
+    expected_score=0.0,
+)
```
Samoed: Maybe would be better to use MockTask from https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_benchmark/mock_tasks.py?

fzowl: @Samoed Can you check now? Is this what you meant? (the test looks much cleaner to me)

Samoed (approved):

PR #2846, opened 2025-06-23
fix: prompt validation for tasks with `-`

Samoed:
Fix validation for tasks with multiple `-`

>  if the task dataset contains a "-", there will be an error in parsing the task_name, such as when specifying "NQ-PL-query".

Ref https://github.com/embeddings-benchmark/results/pull/217#issuecomment-2996133931
CC @YanshekWoo


KennethEnevoldsen (approved):

KennethEnevoldsen:
Should we make a test for this?

Samoed:
I tried a bit to add test, but this is a bit tricky. I will try to add it again

PR #2855, opened 2025-06-25
fix: Update model selection for the leaderboard

### LINKED ISSUE #2834: Add size ticks to leaderboard

[Some people on Twitter](https://x.com/antoine_chaffin/status/1934990854890090670) were discussing the leaderboard, particularly the size picker. 

As an enhancement, we could add some tickets for general size buckets (although they are somewhat disputed) for small/base/large models.  Probably, < 150M, < 350M, < 1B, < 5B, < 10B ranges although they can be tweaked

KennethEnevoldsen:
fixes #2834

This removed the lower bound selection, but generally I don't think people should care about the models being too small.

Had x-tabdeveloping review in person.

Screenshot 2025-06-25 at 13 47 35

edit: have fixed the 1M -> 1B error

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -157,10 +157,10 @@ def filter_models(
     availability: bool | None,
     compatibility: list[str],
     instructions: bool | None,
-    model_size: int,
+    max_model_size: int,
     zero_shot_setting: Literal["only_zero_shot", "allow_all", "remove_unknown"],
 ):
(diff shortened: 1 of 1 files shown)


orionw:
Looks great to me! Only suggestion, can the user select multiples categories or is it limited to one? EDIT: ah, maybe the higher category includes all smaller versions. That works!

KennethEnevoldsen:
Yep higher includes all smaller as well

PR #2861, opened 2025-06-27
Adding nvidia/llama-nemoretriever-colembed models

bschifferer:

Both Models will be released in the next hours:
https://huggingface.co/nvidia/llama-nemoretriever-colembed-1b-v1
https://huggingface.co/nvidia/llama-nemoretriever-colembed-3b-v1

`mteb.get_model_meta(args.model_name_or_path)` is used to run MTEB VisualDocumentRetrieval leaderboad (python file provided in the HuggingFace Model Repository


mteb/models/nvidia_llama_nemoretriever_colemb.py
```diff
@@ -0,0 +1,139 @@
...
+    name="nvidia/llama-nemoretriever-colembed-1b-v1",
+    languages=["eng-Latn"],
+    revision="1f0fdea7f5b19532a750be109b19072d719b8177",
+    release_date="2025-06-27",
+    modalities=["image", "text"],
+    n_parameters=2_418_000_000,
+    memory_usage_mb=9224,
+    max_tokens=8192,
+    embed_dim=2048,
+    license="not specified",
```
isaac-chung: it's a customized NVIDIA license: https://huggingface.co/nvidia/llama-nemoretriever-colembed-1b-v1/blob/main/LICENSE

mteb/models/nvidia_llama_nemoretriever_colemb.py
```diff
@@ -0,0 +1,139 @@
...
+    open_weights=True,
+    public_training_code="Proprietary Code",
+    public_training_data="Proprietary",
+    framework=["PyTorch"],
+    reference="https://huggingface.co/nvidia/llama-nemoretriever-colembed-1b-v1",
+    similarity_fn_name="max_sim",
+    use_instructions=True,
+    training_datasets={},
```
isaac-chung: Let's fill in the training sets as well: https://huggingface.co/nvidia/llama-nemoretriever-colembed-1b-v1#training-dataset. [See bge as an example](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/bge_models.py).

isaac-chung (commented):
Hi @bschifferer, thanks for adding this. Just 2 small suggestions, otherwise LGTM!

mteb/models/nvidia_llama_nemoretriever_colemb.py
bschifferer: thanks - I tried adding the name before and then I got an error, that I can only select from a list of licenses. "not specified" sounded the closest. I updated it with the LICENSE URL

mteb/models/nvidia_llama_nemoretriever_colemb.py
bschifferer: Thanks - I updated the code to add all the training datasets from our model card

PR #2879, opened 2025-07-05
fix: Seperate model implementations and encoder specifications

KennethEnevoldsen:
- Move all implementations into a seperate folder called `model_implementations`
- moved `encoder_interface.py` and `model_meta.py` into `models`
- renamed `models/*` to `encoder_implementions/*` to make the distinction from between the two folder clear
- merged `models/utils.py` into the only model that used it

We seems to have a few differing names when referring to a model (ModelMeta, get_model, etc.) and encoders (Encoder, AbsEncoder). Should we try to do something about this or just leave it as is?

Note: There is a few refactors that I would like to do on top of this, but will add that to seperate PR (since it is too hard to review here)

Fixes


isaac-chung:
> We seems to have a few differing names when referring to a model (ModelMeta, get_model, etc.) and encoders (Encoder, AbsEncoder). Should we try to do something about this or just leave it as is?

ModelMeta, get_model are separate things. What do you mean here?
Looks like AbsEncoder is a wrapper base class - maybe that can benefit from a better name?

isaac-chung:
> Move all implementations into a seperate folder called model_implementations
moved encoder_interface.py and model_meta.py into models
renamed models/* to encoder_implementions/* to make the distinction from between the two folder clear
merged models/utils.py into the only model that used it

- Cannot find `encoder_implementions`. Do you mean `model_implementations`?
- Instead of `model_implementations`, since it's within the models folder, can rename to just `implementations`?

Otherwise, looks good!

KennethEnevoldsen:
Yeah did mean model_implementations (used encoder implementations earlier)

So let us just leave the naming as is and then take a look at it later

Issue #2869, opened 2025-06-30
Model Links do not work when switching leaderboards


bschifferer:
### Describe the bug

When I visit https://huggingface.co/spaces/mteb/leaderboard and switch to another leaderboard, the URLs of models cannot be open correctly. It will just show an empty white browser.

### To reproduce

Reproducible: Using Chrome (tested other browsers, etc)
1. Visit https://huggingface.co/spaces/mteb/leaderboard
2. Select any other leaderboard (e.g click on "English" below "Multilingual")
3. Click on the model name in the table (e.g. "LGAI-Embedding-Preview")

It will just keep the URL "https://huggingface.co/spaces/mteb/leaderboard" and keeps a white space.

For some reason, it works on the default view (e.g. Multlingual)
For some reason, it works when using a direkt URL http://mteb-leaderboard.hf.space/?benchmark_name=MTEB%28eng%2C+v2%29

### Are you interested to contribute a fix for this bug?

No

Samoed:
I can verify this. I can open links in another tab without problem. Maybe problem that other leaderboards are hidden in drop down. Otherwise I don't understand why this can happen

![Image](https://github.com/user-attachments/assets/98860632-a55d-445e-9b8c-a532ccba0abc)

KennethEnevoldsen:
Opening it in a new tab seems to fix the issue.

@Samoed, not sure why there would be an issue with the dropdown?

If I examine the html they seem to look equivalent:
```
# from "Multlingual"
<p><a href="https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-text-embeddings">gemini-embedding-001</a></p>
```

```
# from "English"
<p><a href="https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-text-embeddings">gemini-embedding-001</a></p>
```

Samoed:
currently I don't have ideas. I will try to debug it a bit later

KennethEnevoldsen:
@Samoed can you still reproduce this in the current version?

Samoed:
Yes, I can reproduce

KennethEnevoldsen:
Hmm can reproduce on firefox, but not on chrome or Safari - what are you on?

Samoed:
I can reproduce on all of them

Samoed:
I'm getting it again

KennethEnevoldsen:
can we change the link such that it opens in a new tab by default?

[...]

PR #2881, opened 2025-07-05
fix selector state

Samoed:
I tried to replicate issue https://github.com/embeddings-benchmark/mteb/issues/2869, but locally everything works normal. I  tried to look to html and css, but I don't see difference between locally and on spaces. 

I've solved problems when `selector_state`

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -74,6 +74,8 @@ def download_table(table: pd.DataFrame) -> str:
 
 
 def update_citation(benchmark_name: str) -> str:
+    if benchmark_name == "selector_state":
+        return ""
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen (commented):
Can you push this to the demo leaderboard? Just to  check that everything works

We should also add some documentation in a comment if it works (just to make sure it isn't removed in the future

Samoed:
I have deployed this branch here  https://huggingface.co/spaces/mteb/leaderboard_2_demo, but it has not solved the problems

KennethEnevoldsen:
Hmm, yeah, so seems like it is still an issue...

KennethEnevoldsen:
close in favor of #2939

Issue #2692, opened 2025-05-11
[v2] Refactor `Retrieval` tasks to use dataset directly


Samoed:
Currently, the corpus and queries are being transformed from the dataset to a dictionary. This process can be extremely slow for large datasets—for example, it took more than an hour to download `MiracleRetrieval`, with most of the time spent on type conversion to a dictionary

KennethEnevoldsen:
@Samoed I believe this is fixed?

PR #2885, opened 2025-07-06
Retrieval ds

Samoed:
Fixes https://github.com/embeddings-benchmark/mteb/issues/2692

### DIFF
--- a/mteb/abstasks/AbsTaskRetrieval.py
+++ b/mteb/abstasks/AbsTaskRetrieval.py
@@ -166,18 +166,31 @@ class AbsTaskRetrieval(AbsTask):
         )
 
         if self.metadata.is_multilingual:
-            # todo add conversion from dict to dataset
             for subset in self.queries:
                 for split in self.queries[subset]:
-                    self.dataset[subset][split]["queries"] = self.queries[subset][split]
-                    self.dataset[subset][split]["corpus"] = self.corpus[subset][split]
+                    queries = self.queries[subset][split]
+                    corpus = self.corpus[subset][split]
+                    self.dataset[subset][split]["queries"] = Dataset.from_list(
+                        [{"id": k, "text": v} for k, v in queries.items()]
+                    )
+                    self.dataset[subset][split]["corpus"] = Dataset.from_list(
+                        [
+                            {
+                                "id": k,
+                                "text": v["text"],
+                                "title": v.get("title", ""),
+                            }
+                            for k, v in corpus.items()
+                        ]
+                    )
                     self.dataset[subset][split]["relevant_docs"] = self.relevant_docs[
                         subset
                     ][split]
                     if hasattr(self, "instructions"):
-                        self.dataset[subset][split]["instructions"] = self.instructions[
-                            subset
-                        ][split]
+                        instructions = self.instructions[subset][split]
+                        self.dataset[subset][split]["instructions"] = Dataset.from_list(
+                            [{"id": k, "text": v} for k, v in instructions.items()]
+                        )
                     if hasattr(self, "top_ranked"):
                         self.dataset[subset][split]["top_ranked"] = self.top_ranked[
                             subset
@@ -185,15 +198,29 @@ class AbsTaskRetrieval(AbsTask):
         else:
             subset = "default"
             for split in self.queries:
-                self.dataset[subset][split]["queries"] = self.queries[split].copy()
-                self.dataset[subset][split]["corpus"] = self.corpus[split].copy()
+                queries = self.queries[split]
+                corpus = self.corpus[split]
+                self.dataset[subset][split]["queries"] = Dataset.from_list(
+                    [{"id": k, "text": v} for k, v in queries.items()]
+                )
+                self.dataset[subset][split]["corpus"] = Dataset.from_list(
+                    [
+                        {
+                            "id": k,
+                            "text": v["text"],
+                            "title": v.get("title", ""),
+                        }
+                        for k, v in corpus.items()
+                    ]
+                )
                 self.dataset[subset][split]["relevant_docs"] = self.relevant_docs[
                     split
                 ].copy()
                 if hasattr(self, "instructions"):
-                    self.dataset[subset][split]["instructions"] = self.instructions[
-                        split
-                    ].copy()
+                    instructions = self.instructions[split]
+                    self.dataset[subset][split]["instructions"] = Dataset.from_list(
+                        [{"id": k, "text": v} for k, v in instructions.items()]
+                    )
                 if hasattr(self, "top_ranked"):
                     self.dataset[subset][split]["top_ranked"] = self.top_ranked[
                         split
--- a/mteb/abstasks/AbsTaskRetrieval.py
+++ b/mteb/abstasks/AbsTaskRetrieval.py
@@ -16,7 +16,7 @@ from mteb.types.statistics import DescriptiveStatistics
 from ..evaluation.evaluators import RetrievalEvaluator
 from ..evaluation.evaluators.retrieval_metrics import make_score_dict
 from .AbsTask import AbsTask
-from .dataset_loaders import RetrievalDatasetLoader, RetrievalSplitData
+from .retrieval_dataset_loaders import RetrievalDatasetLoader, RetrievalSplitData
 
 logger = logging.getLogger(__name__)
 
@@ -135,11 +135,12 @@ class AbsTaskRetrieval(AbsTask):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        empty_dataset = Dataset.from_dict({})
         self.dataset = defaultdict(
             lambda: defaultdict(
                 lambda: RetrievalSplitData(
-                    corpus={},
-                    queries={},
+                    corpus=empty_dataset,
+                    queries=empty_dataset,
                     relevant_docs={},
                     instructions=None,
                     top_ranked=None,
@@ -150,11 +151,13 @@ class AbsTaskRetrieval(AbsTask):
     def convert_v1_dataset_format_to_v2(self):
         if not hasattr(self, "queries"):
             return
+        empty_dataset = Dataset.from_dict({})
+
         self.dataset = defaultdict(
             lambda: defaultdict(
                 lambda: RetrievalSplitData(
-                    corpus={},
-                    queries={},
+                    corpus=empty_dataset,
+                    queries=empty_dataset,
                     relevant_docs={},
                     instructions=None,
                     top_ranked=None,
@@ -163,6 +166,7 @@ class AbsTaskRetrieval(AbsTask):
         )
 
         if self.metadata.is_multilingual:
+            # todo add conversion from dict to dataset
             for subset in self.queries:
                 for split in self.queries[subset]:
                     self.dataset[subset][split]["queries"] = self.queries[subset][split]
@@ -273,7 +277,7 @@ class AbsTaskRetrieval(AbsTask):
     def _evaluate_subset(
         self,
         model: Encoder,
-        data_split: DatasetDict | Dataset,
+        data_split: RetrievalSplitData,
         encode_kwargs: dict[str, Any],
         hf_split: str,
         hf_subset: str,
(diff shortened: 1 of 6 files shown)


mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -116,11 +120,13 @@ def __init__(self, **kwargs):
     def convert_v1_dataset_format_to_v2(self):
```
KennethEnevoldsen, line 124: I am unsure how our changes influence the loading of v2 retrieval dataset (do we have any of those? It would be great to see how that would influence the load data setup.

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -116,11 +120,13 @@ def __init__(self, **kwargs):
...
         if not hasattr(self, "queries"):
```
KennethEnevoldsen, line 126: Is this a way to check if it is v2? Might be worth adding a comment on this

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -399,25 +433,42 @@ def _calculate_metrics_from_split(
...
             for hf_subset in self.metadata.eval_langs:
                 split_data = self.dataset[hf_subset][split]
-                queries.update(process_docs(split_data["queries"], hf_subset, split))
-                corpus.update(process_docs(split_data["corpus"], hf_subset, split))
+                cur_queries = split_data["queries"].map(
+                    map_indexes_for_ds, fn_kwargs=dict(hf_subset=hf_subset, split=split)
+                )
+                cur_corpus = split_data["corpus"].map(
+                    map_indexes_for_ds, fn_kwargs=dict(hf_subset=hf_subset, split=split)
+                )
```
KennethEnevoldsen: isn't there already an index, why do we need to create a new one?

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -577,17 +636,24 @@ def calculate_corpus_length(
 
 
 def process_docs(
-    collection: dict[str, dict[str, str] | str], hf_subset: str, split: str
-) -> dict[str, str]:
+    collection: dict[str, list[str]], hf_subset: str, split: str
```
KennethEnevoldsen: Do we still need this function? (seems like the following superseedes it)

mteb/create_dataloaders.py
```diff
@@ -62,7 +62,7 @@ def corpus_to_dict(
 
 
 def create_dataloader_for_retrieval_corpus(
```
KennethEnevoldsen, line 45: in this function we take a dataset, convert it to a dictionary and then back to a dataset - could we avoid the transformation?

mteb/evaluation/evaluators/model_classes.py
```diff
@@ -70,15 +71,15 @@ def __init__(
 
     def search(
         self,
-        corpus: dict[str, dict[str, str]],
-        queries: dict[str, str | list[str] | list[dict[str, str]]],
+        corpus: Dataset,
```
KennethEnevoldsen: We could introduce two new types:

```
QueryDataset = Dataset
CorpusDataset = Dataset
```
To differentiate between the two. It also allows us to add some documentation on what those entail, but I am not sure if it is worth it.

mteb/abstasks/AbsTaskRetrieval.py
KennethEnevoldsen: How does this influence pushing datasets to the hub? Does it push in a v2 compatible manner, and do we need to update now?

tests/test_benchmark/mock_tasks.py
```diff
@@ -1419,12 +1439,12 @@ class MockRerankingTask(AbsTaskRetrieval):
     expected_stats = {
         "test": {
             "num_samples": 4,
-            "number_of_characters": 112,
+            "number_of_characters": 136,
```
KennethEnevoldsen, line 1442: were these incorrect before?

tests/test_tasks/test_mteb_rerank.py
```diff
@@ -351,7 +351,9 @@ def test_mteb_rerank(tmp_path: Path):
     with (tmp_path / "SciFact_default_predictions.json").open() as f:
         results = json.load(f)
 
-    results = sorted(results["1"].keys(), key=lambda x: (results["1"][x], x))[:2]
+    results = sorted(
+        results["1"].keys(), key=lambda x: (results["1"][x], x), reverse=True
+    )[:2]
```
KennethEnevoldsen: reversed=True. Isn't that a bit worrying?

tests/test_benchmark/test_models.py
```diff
@@ -27,7 +27,7 @@ def test_colbert_model_e2e(task: AbsTask, model: str, tmp_path: Path):
     )
     result = results[0]
 
-    assert result.scores["test"][0]["ndcg_at_1"] == 1.0
+    assert result.scores["test"][0]["ndcg_at_1"] == 0.0
```
KennethEnevoldsen, line 30: Why would this change?

Samoed:
> Yes. I have had that problem before as well. I have actually considered replacing the mockencoder with a small model2vec model 

We can use approach as is in langchain fake embedding by setting seed from the text https://github.com/langchain-ai/langchain/blob/master/libs%2Fcore%2Flangchain_core%2Fembeddings%2Ffake.py#L120

KennethEnevoldsen:
> We can use approach as is in langchain fake embedding by setting seed from the text https://github.com/langchain-ai/langchain/blob/master/libs%2Fcore%2Flangchain_core%2Fembeddings%2Ffake.py#L120

Great idea! Let us do that instead.

tests/test_benchmark/test_models.py
Samoed: As I said in description, there is issue due to sorting and embedding creation 

> I also modified the test a bit because we’re creating mock embeddings independently of the input. Previously, the corpus dict was sorted by keys and, for some reason, ended up as [d2, d1], so the first embedding was used.

tests/test_benchmark/mock_tasks.py
Samoed: I've added title to the corpus

mteb/evaluation/evaluators/model_classes.py
Samoed: I think this would be good addition

mteb/abstasks/AbsTaskRetrieval.py
Samoed: Some datasets could have same indexes between different splits (as in our tests), so we need to add additional values to index to avoid this problem

tests/test_tasks/test_mteb_rerank.py
Samoed: I tried to debug this part and found that we’re handling the reranking task a bit incorrectly, because it passes a `tuple[dict]` to `corpus_to_dict`, and this ends up storing all information in the `text` field. This might be part of the issue discussed in https://github.com/embeddings-benchmark/mteb/issues/2933.

After making this change, I looked at the `results`, and they contain values like:
```python
{'18670': -10.466630935668945, '4983': -8.812150955200195, '19238': -11.240396499633789}
````

So we need to sort them in reverse.

File: mteb/abstasks/AbsTaskRetrieval.py (hunk 1 above)
Comment (Samoed) [reply mteb/abstasks/AbsTaskRetrieval.py:124]: Updated function for pushing datasets

File: mteb/abstasks/AbsTaskRetrieval.py (hunk 2 above)
Comment (Samoed) [reply mteb/abstasks/AbsTaskRetrieval.py:126]: Added comment

File: mteb/abstasks/AbsTaskRetrieval.py (hunk 3 above)
Comment (Samoed) [reply mteb/abstasks/AbsTaskRetrieval.py]: Yes. I just reimplemented logic with dict corpus/queries, but now we don't need to create new indexes to calculate statistics

File: mteb/abstasks/AbsTaskRetrieval.py (hunk 4 above)
Comment (Samoed) [reply mteb/abstasks/AbsTaskRetrieval.py]: Removed

File: mteb/create_dataloaders.py (hunk 5 above)
Comment (Samoed) [reply mteb/create_dataloaders.py:45]: Updated to use map directly

File: mteb/abstasks/AbsTaskRetrieval.py (hunk 1 above)
Comment (Samoed) [reply mteb/abstasks/AbsTaskRetrieval.py:124]: I tried to measure time with this script
```python
import mteb
from datetime import datetime

task = mteb.get_task("SCIDOCS")
start = datetime.now()
task.load_data()
print((datetime.now() - start).seconds)
```
And for `v2` I got times 10-30 seconds, and for this PR from 7-20 seconds. I don't understand why there is so huge difference between same runs on same branch (dataset was preloaded locally).

Samoed:
@KennethEnevoldsen can you rereview this?

tests/test_tasks/test_mteb_rerank.py
KennethEnevoldsen: hmm, interesting - we should maybe test for the exact values here then - to ensure that we don't change this behaviour going forward 

We of course also have to test https://github.com/embeddings-benchmark/mteb/issues/2933

tests/test_tasks/test_mteb_rerank.py
```diff
@@ -345,14 +345,17 @@ def test_mteb_rerank(tmp_path: Path):
         eval_splits=["test"],
         previous_results=tmp_file,
         save_predictions=True,
+        co2_tracker=False,
```
KennethEnevoldsen, line 348: Yea, will be good to have it as default off in v2's `evaluate`

mteb/models/encoder_interface.py
```diff
@@ -5,14 +5,18 @@
...
+    PromptType,
+)
 
 
 @runtime_checkable
 class Encoder(Protocol):
     """The interface for an encoder in MTEB.
 
-    Besides the required functions specified below, the encoder can additionally specify the the following signatures seen below.
+    Besides the required functions specified below, the encoder can additionally specify the following signatures seen below.
```
KennethEnevoldsen, line 19: Hmm, assuming we will make docs on this, then I could imagine that above/below will not work (it is also likely that you move stuff around)

mteb/abstasks/_statistics_calculation.py
```diff
@@ -147,18 +148,16 @@ def calculate_top_ranked_statistics(
...
-    ]
+    qrels_lengths = [len(relevant_docs[qid]) for qid in relevant_docs]
     unique_qrels = len({doc for qid in relevant_docs for doc in relevant_docs[qid]})
     # number of qrels that are not 0
     num_qrels_non_zero = sum(
         sum(1 for doc_id in docs if docs[doc_id] != 0)
         for docs in relevant_docs.values()
     )
-    qrels_per_doc = num_qrels_non_zero / len(relevant_docs) if len(queries_ids) else 0
+    qrels_per_doc = num_qrels_non_zero / len(relevant_docs)
```
KennethEnevoldsen, line 160: Are we sure we won't get a ZeroDivisionError here?

KennethEnevoldsen (approved):
A few minor things, generally looks very good!

mteb/models/encoder_interface.py
Samoed: Yes, but I think we can work on docs later. I just removed duplicated `the`

mteb/abstasks/_statistics_calculation.py
Samoed: This would only happens when we don't have relevant documents at all

tests/test_tasks/test_mteb_rerank.py
Samoed: Yes, I think we would create mock cross-encoders to test them more reproducible

mteb/abstasks/_statistics_calculation.py
KennethEnevoldsen: If that happens a ZeroDivisionError is fine

mteb/models/encoder_interface.py
KennethEnevoldsen: but the code could also move around (which would make the comment wrong). I would instead just refer directly by name.

Samoed: I think we can leave this for another pr

PR #2887, opened 2025-07-06
MAEB Classification Datasets Downsampling/Formatting + MTEB UPLOAD

AdnanElAssadi56:
- All classification datasets have been downsampled so that each is under 1.5 GB, with most now in the 500 MB range.
- Uploaded every downsampled classification dataset to the mteb Hugging Face repo with our agreed-upon fixed naming convention.
- I noticed that the current SIBFleurs dataset supports 100+ languages, so I selected a subset of 15 languages for memory efficiency and added it as a new sibfleurs_multilingual configuration replacing the previous English‑only entry. I also did stratified downsampling because it was large.

- I have noticed that the current evaluation pipeline does not accept a “test” split for multilingual datasets.
  My Temporary workaround for SibFleurs: renamed the split to “train” and used cross-validation for evaluation.


KennethEnevoldsen (commented):
Looks good here - the last comment worries be a bit though. Seems like we have to fix the implementation (@Samoed if you have the time it would be great to ensure that it is easy for us to merge it with v2 in the future)

isaac-chung (approved):

isaac-chung:
> Looks good here - the last comment worries be a bit though. Seems like we have to fix the implementation (@Samoed if you have the time it would be great to ensure that it is easy for us to merge it with v2 in the future)

Let's create an issue for that.

Issue #2832, opened 2025-06-16
Handle MRL


orionw:
### Description of the feature

Do we handle MRL with some config parameter? Seems like an easy enhancement if not.

KennethEnevoldsen:
Hi @orionw what would handling mean here? Is it possible to evaluate with it (I believe it is) or is it present on the leaderboard?

orionw:
Mainly for evaluation purposes.  Do we have it already and I missed it @KennethEnevoldsen?

KennethEnevoldsen:
isn't is possible to just pass `encode_kwargs`?

orionw:
I don't think it's something the models themselves handle, I think typically you just truncate them after they generate the full embedding. 

So something like [this](https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/evaluation/evaluators/model_classes.py#L373):

```python
with torch.inference_mode():
    scores = self.model.similarity(query_embeddings, sub_corpus_embeddings)
```

Would become
```python
with torch.inference_mode():
    scores = self.model.similarity(query_embeddings[:, mrl_dimension], sub_corpus_embeddings[:, mrl_dimension])
```

It seems actually quite simple to add if we add as a new argument, I can add it tomorrow or so.

Samoed:
I think we can duplicate some instances of models with different encoding sizes for MRL

orionw:
Is it preferable to duplicate the models or just let users pass in the argument? Like custom instructions, this is a case where the eval score will be quite different depending on the value used but duplicating models seems like a lot of work since there are ~4 sizes of MRL per model. 

But arguments are saved out somewhere right (the `encode_kwargs`) so perhaps that could be the reference.

Samoed:
> but duplicating models seems like a lot of work since there are ~4 sizes of MRL per model.

I don't think that this a lot of work, because only model meta would be duplicated

> But arguments are saved out somewhere right (the encode_kwargs) so perhaps that could be the reference.

Yes, but it would be hard to fit them in layout of results

KennethEnevoldsen:
So for one model that supports MRL the current interface is:

```py
import mteb

tasks = mteb.get_tasks(tasks=["LccSentimentClassification"])
eval = mteb.MTEB(tasks=tasks)


meta = mteb.get_model_meta("openai/text-embedding-3-small")


model = meta.load_model()  # Default embed_dim is 1536
results = eval.run(model, overwrite_results=True)
[...]

PR #2893, opened 2025-07-10
model: add image support for jina embeddings v4

makram93:
This PR adds support for image embeddings in Jina Embeddings v4 and refactors the `encode()` method to use `get_text_embedding()` for consistency.

   - [x] mteb.get_model(model_name, revision) and
   - [x] mteb.get_model_meta(model_name, revision)

### DIFF
--- a/mteb/models/jina_models.py
+++ b/mteb/models/jina_models.py
@@ -230,13 +230,8 @@ class JinaV4Wrapper(Wrapper):
         attn_implementation="flash_attention_2",
         trust_remote_code: bool = True,
         model_prompts: dict[str, str] | None = None,
-        vector_type: str = "multi_vector",
         **kwargs,
     ) -> None:
-        if CURRENT_SENTENCE_TRANSFORMERS_VERSION < MIN_SENTENCE_TRANSFORMERS_VERSION:
-            raise RuntimeError(
-                f"sentence_transformers version {st_version} is lower than the required version 3.1.0"
-            )
         requires_package(
             self,
             "flash_attn",
@@ -258,15 +253,7 @@ class JinaV4Wrapper(Wrapper):
             revision=revision,
         ).eval()
         self.model_prompts = model_prompts or {}
-        self.vector_type = self._validate_vector_type(vector_type)
-
-    def _validate_vector_type(self, vector_type: str) -> str:
-        """Validate and return vector type."""
-        if vector_type not in self.SUPPORTED_VECTOR_TYPES:
-            raise ValueError(
-                f"vector_type must be one of {self.SUPPORTED_VECTOR_TYPES}, got {vector_type}"
-            )
-        return vector_type
+        self.vector_type = "single_vector"  # default vector type
 
     def _resolve_task_parameters(
         self, task_name: str | None, prompt_type: PromptType | None = None
@@ -274,7 +261,7 @@ class JinaV4Wrapper(Wrapper):
         """Resolve task parameters from task_name and prompt_type.
 
         Returns:
-            tuple: (jina_prompt, base_task, prompt_name_param)
+            tuple: (base_task, prompt_name_param, task_type)
         """
         task_type = self.get_prompt_name(self.model_prompts, task_name, prompt_type)
         jina_task_name = self.model_prompts.get(task_type) if task_type else None
@@ -302,26 +289,10 @@ class JinaV4Wrapper(Wrapper):
     ) -> None:
         """Log task and prompt information."""
         if prompt_name:
-            logger.info(
-                f"Using prompt_name={prompt_name} for task={task_name} prompt_type={prompt_type}"
-            )
+            logger.info(f"Using {prompt_name=} for {task_name=} {prompt_type=}")
         else:
-            logger.info(
-                f"No model prompts found for task={task_name} prompt_type={prompt_type}"
-            )
+            logger.info(f"No model prompts found for {task_name=} {prompt_type=}")
         logger.info(f"Encoding {sentences_count} sentences.")
-
-    def _warn_vector_type_mismatch(self, task_type: str | None) -> None:
-        """Warn about vector type mismatches for DocumentUnderstanding tasks."""
-        if (
-            task_type
-            and task_type.startswith("DocumentUnderstanding")
-            and self.vector_type != "multi_vector"
-        ):
-            logger.warning(
-                "Please use 'multi_vector' mode for DocumentUnderstanding tasks to get better results. "
-                "Load the model with: meta.load_model(vector_type='multi_vector')"
-            )
 
     def encode(
         self,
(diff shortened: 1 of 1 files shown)


mteb/models/jina_models.py
```diff
@@ -213,72 +216,327 @@ def encode(
...
+        """Compute similarity between embeddings.
+
+        Args:
+            a: First embedding
+            b: Second embedding
+        """
+        a_torch = self._convert_to_torch_if_needed(a)
+        b_torch = self._convert_to_torch_if_needed(b)
+
+        if self.vector_type == "single_vector":
```
Samoed, line 432: I don't think that we would handle different similarities like this, because they would output different results

makram93: >I don't think that we would handle different similarities like this

Sorry, could you clarify what you're suggesting?

Just to explain: our v4 model supports both modes—returning a single embedding vector or a ColBERT-style multi-vector representation.

Samoed: I think you can leave only a single embedding. We have issues (https://github.com/embeddings-benchmark/mteb/issues/2832, https://github.com/embeddings-benchmark/mteb/issues/1211) with how to handle different parameters during evaluation, but we haven't decided yet.


CC @KennethEnevoldsen

makram93: In my current implementation, I handle different embedding configurations like this:

```
meta = mteb.get_model_meta("jinaai/jina-embeddings-v4")
custom_model = meta.load_model(vector_type="multi_vector") # vector_type="single_vector"
```
I’ve tested this setup on image-related tasks, and while the results do vary between `single_vector` and `multi_vector` modes, the differences are consistent with the results previously obtained using the `vidore_benchmark` package.

However, if the preference is to use only the single_vector output for now, I’m happy to make that change. Let me know your recommendation, and I’ll update accordingly.

Samoed: Yes, loading model with these args would work correctly, but we can't correctly show results with different parameters and for now we can handle correctly only one configuration for the model. For now, I think you can leave only single_vector and create issue to add support of multivector

mteb/models/jina_models.py
```diff
@@ -213,72 +216,327 @@ def encode(
...
+        attn_implementation="flash_attention_2",
+        trust_remote_code: bool = True,
         model_prompts: dict[str, str] | None = None,
+        vector_type: str = "multi_vector",
         **kwargs,
     ) -> None:
+        if CURRENT_SENTENCE_TRANSFORMERS_VERSION < MIN_SENTENCE_TRANSFORMERS_VERSION:
+            raise RuntimeError(
+                f"sentence_transformers version {st_version} is lower than the required version 3.1.0"
+            )
```
isaac-chung: If Jinav4 is loaded from transformers, maybe we don't need to check sentence transformers version?

mteb/models/jina_models.py
```diff
@@ -213,72 +216,327 @@ def encode(
...
+        return vector_type
+
+    def _resolve_task_parameters(
+        self, task_name: str | None, prompt_type: PromptType | None = None
+    ) -> tuple[str, str, str]:
+        """Resolve task parameters from task_name and prompt_type.
+
+        Returns:
+            tuple: (jina_prompt, base_task, prompt_name_param)
```
isaac-chung: Should this be updated? Couldn't find `jina_prompt` in the code.

mteb/models/jina_models.py
```diff
@@ -213,72 +216,327 @@ def encode(
...
+    def _log_task_info(
+        task_name: str,
+        prompt_type: PromptType | None,
+        prompt_name: str | None,
+        sentences_count: int,
+    ) -> None:
+        """Log task and prompt information."""
+        if prompt_name:
+            logger.info(
+                f"Using prompt_name={prompt_name} for task={task_name} prompt_type={prompt_type}"
```
isaac-chung:
```suggestion
                f"Using {prompt_name=} for {task_name=} {prompt_type=}"
```

mteb/models/jina_models.py
```diff
@@ -213,72 +216,327 @@ def encode(
...
+            )
+        else:
+            logger.info(
+                f"No model prompts found for task={task_name} prompt_type={prompt_type}"
```
isaac-chung:
```suggestion
                f"No model prompts found for {task_name=} {prompt_type=}"
```

mteb/models/jina_models.py
```diff
@@ -213,72 +216,327 @@ def encode(
...
+            iterator = DataLoader(images, batch_size=batch_size)
+
+        for batch in iterator:
+            for b in batch:
+                pil_img = (
+                    F.to_pil_image(b.to("cpu")) if not isinstance(b, Image.Image) else b
+                )
+                all_images.append(pil_img)
+
+        batch_size = 1
```
isaac-chung, line 387: Could we not use the param that's passed in, i.e. `batch_size: int = 32`?

isaac-chung (commented):
Looks good overall. Just got a few questions.

mteb/models/jina_models.py
makram93: @Samoed Thanks for clarifying that. I'll go ahead and make the necessary changes. 

In that case, can we submit the ViDoRe results for the `multi-vector` embedding to the [results](https://github.com/embeddings-benchmark/results) repo? The difference (between `single-vector` and `multi-vector`) is significant and impacts our ranking on the ViDoRe leaderboard.

isaac-chung: Our goal here is for the results to be reproducible, and to do that, we want all parameters used to produce the results logged in the ModelMeta, which is saved in the corresponding results folder. If the PR can do that, I don't see a problem with the LB (that way you will have the same model name on the LB instead of jina-v4-single-vec and jina-v4-multi-vec).

mteb/models/jina_models.py
makram93: Given the specified image resolution and model size, we can handle at most ~2 images per batch before running into an out-of-memory (OOM) error.

makram93:
@Samoed @isaac-chung Thanks for the input and suggestions. 

I’ve refactored the code to load the `ModelMeta` without any parameters. It currently runs `single_vec` for MTEB and `multi_vec` for ViDoRe. It’s a bit inelegant for now, but once support is added in MTEB for handling different configurations during evaluation, we can revisit and make changes accordingly.

Samoed (approved):
Great!

isaac-chung (approved):

isaac-chung:
@makram93 thanks for iterating. Since you're now more familiar with our goal, if you have any suggestions on how to implement the extra parameter that can be saved in ModelMeta, feel free to raise an issue/ PR in the future. I will merge this now.

PR #2632, opened 2025-05-02
Classification dataset cleaning

AlexeyVatolin:
Fixes [comment](https://github.com/embeddings-benchmark/mteb/issues/1049#issuecomment-2468424773) from #1049
### Scores comparison
As expected, scores slightly drop after my changes for most tasks.
**multilingual-e5-small**
|    | task_name                                 |   main_score_old |   main_score_new |   delta_percent |
|---:|:------------------------------------------|-----------------:|-----------------:|----------------:|
|  0 | AJGT                                      |         0.7455   |         0.727835 |           -2.37 |
|  2 | BengaliDocumentClassification             |         0.506934 |         0.496289 |           -2.1  |
|  4 | BengaliHateSpeechClassification           |         0.492376 |         0.497559 |            1.05 |
|  6 | BengaliSentimentAnalysis                  |         0.831888 |         0.805349 |           -3.19 |
|  8 | CSFDCZMovieReviewSentimentClassification  |         0.275586 |         0.277197 |            0.58 |
| 11 | CzechProductReviewSentimentClassification |         0.507227 |         0.504395 |           -0.56 |
| 12 | HotelReviewSentimentClassification        |         0.494775 |         0.492434 |           -0.47 |
| 14 | OnlineStoreReviewSentimentClassification  |         0.362256 |         0.311644 |          -13.97 |
| 16 | RestaurantReviewSentimentClassification   |         0.635986 |         0.631377 |           -0.72 |
| 18 | TweetEmotionClassification                |         0.511279 |         0.491574 |           -3.85 |
| 20 | TweetSarcasmClassification                |         0.609289 |         0.564206 |           -7.4  |

**paraphrase-multilingual-MiniLM-L12-v2**
|    | task_name                                 |   main_score_old |   main_score_new |   delta_percent |
|---:|:------------------------------------------|-----------------:|-----------------:|----------------:|
|  1 | AJGT                                      |         0.699278 |         0.693557 |           -0.82 |
|  3 | BengaliDocumentClassification             |         0.335498 |         0.300439 |          -10.45 |
|  5 | BengaliHateSpeechClassification           |         0.340113 |         0.323974 |           -4.75 |
|  7 | BengaliSentimentAnalysis                  |         0.606135 |         0.562747 |           -7.16 |
|  9 | CSFDCZMovieReviewSentimentClassification  |         0.268652 |         0.256982 |           -4.34 |
| 10 | CzechProductReviewSentimentClassification |         0.513428 |         0.493799 |           -3.82 |
[...]

PR #2900, opened 2025-07-13
Classification dataset cleaning

AlexeyVatolin:
I am continuing the work on data cleaning in MTEB. This pull request is a follow-up to the work started in https://github.com/embeddings-benchmark/mteb/pull/2632. The same cleaning rules described in the comment https://github.com/embeddings-benchmark/mteb/pull/2632#issuecomment-2849288975 are applied here.

Additionally, I have introduced a new filter, filter_one_sample_labels, which is useful in rare cases. This filter is applied to datasets that lack a test sample and contain classes with only one example. Such classes are removed because it is not possible to split the sample into train and test sets.

Here is a [comparison of scores](https://gist.github.com/AlexeyVatolin/9bcbb01c07135eb65edc05d623fc4a2b) for all modified datasets on the `intfloat/multilingual-e5-small` and `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2` models.

--- 
I am adding this as a separate file in the gist because the table is too large to include in a comment.


Samoed (approved):

AlexeyVatolin:
@Samoed, could we merge this pull request?

Samoed:
For me looks good, I requested review for second opinion

isaac-chung:
This is great! Could you point me to the script you used to clean these datasets? I wasn't able to find them in this PR or in [the linked PR](https://github.com/embeddings-benchmark/mteb/pull/2632).

AlexeyVatolin:
@isaac-chung, I didn't add this to the repository because the code is somewhat tangled and redundant in some places. Additionally, it currently supports only classification (except for multilingual tasks). I plan to improve it. [Here is](https://gist.github.com/AlexeyVatolin/ea3adc21aa7a767603ff393b22085adc) the link to the Gist with the code.

isaac-chung:
Great stuff! Let's commit this to the script folder as is, and then improve on it. At the end of the day, this code is used to make changes to the public datasets, so it should be (and should have been) recorded. The script folder is not super cleaned up anyway, so we shouldn't let perfect get in the way of good. 

Also, that way others can contribute too if they wanted to for other tasks.

AlexeyVatolin:
@isaac-chung, I agree with you, I have added my script.

isaac-chung (approved):

PR #2901, opened 2025-07-14
SpeechT5 Model

AdnanElAssadi56:
This was supposed to be done for A2T and T2A tasks in retrieval, but I've noticed that their is no unified encoder for handling both audio and text. I tried getting the text encoder from the ```microsoft/speecht5_tts``` variant and the audio encoder from the ```microsoft/speecht5_asr```variant, but the results were very low. Thus, I've pivoted to just add them as regular model for other MAEB tasks since implementation was already mostly done.

Results from ```microsoft/speecht5_vc``` ```microsoft/speecht5_asr```:

[JamAltArtistA2ARetrieval (2).json](https://github.com/user-attachments/files/21207790/JamAltArtistA2ARetrieval.2.json)
[JamAltArtistA2ARetrieval (1).json](https://github.com/user-attachments/files/21207792/JamAltArtistA2ARetrieval.1.json)
[BeijingOpera (1).json](https://github.com/user-attachments/files/21207793/BeijingOpera.1.json)
[BeijingOpera.json](https://github.com/user-attachments/files/21207794/BeijingOpera.json)


AdnanElAssadi:
Comparing with other models and Clap, I noticed it is not bad after all (the task seems hard). Following the approach mentioned above of using the two different variants for audio and text encoders respectively, here are the results:

[JamAltArtistA2ARetrieval.json](https://github.com/user-attachments/files/21246391/JamAltArtistA2ARetrieval.7.json)
[JamAltLyricT2ARetrieval.json](https://github.com/user-attachments/files/21246392/JamAltLyricT2ARetrieval.4.json)
[JamAltLyricA2TRetrieval.json](https://github.com/user-attachments/files/21246393/JamAltLyricA2TRetrieval.5.json)
[AudioCapsA2TRetrieval.json](https://github.com/user-attachments/files/21246434/AudioCapsA2TRetrieval.4.json)
[AudioCapsT2ARetrieval.json](https://github.com/user-attachments/files/21246466/AudioCapsT2ARetrieval.3.json)


P.S  Decided to remove ```microsoft/speecht5_vc``` so that implementation is cleaner.

mteb/models/speecht5_models.py
```diff
@@ -0,0 +1,287 @@
...
+                last_hidden = outputs.last_hidden_state
+                embeddings = torch.mean(last_hidden, dim=1)
+                all_embeddings.append(embeddings.cpu())
+
+        if all_embeddings:
+            return torch.cat(all_embeddings, dim=0)
+        else:
+            return torch.zeros((0, self.tts_model.config.hidden_size))
+
+    def encode(
```
isaac-chung, line 198: encode is really just for text embeddings. It has been confused and we should also fix the rest of the audio models separately.

mteb/models/speecht5_models.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+class SpeechT5Wrapper(Wrapper):
+    def __init__(
+        self,
+        model_name: str,
+        device: str = "cuda" if torch.cuda.is_available() else "cpu",
+        **kwargs: Any,
+    ):
+        self.model_name = model_name
```
isaac-chung, line 30: is `self.model_name` used anywhere else? is it meant to handle different models?
> I've pivoted to just add them as regular model for other MAEB tasks since implementation was already mostly done.

mteb/models/speecht5_models.py
isaac-chung: For example: https://github.com/embeddings-benchmark/mteb/pull/2914

Issue #2896, opened 2025-07-10
Prompts handling with Sentence Transformers


arthurbr11:
### Describe the bug

Hello ! With Sentence Transformers v5, the default prompt initialization creates empty prompts that cause MTEB to reject valid model prompts.

**Root Cause:**
In ST v5, models are initialized with default empty prompts (for better consistency with update include in the release) as you can see  [here](https://github.com/UKPLab/sentence-transformers/blob/4f8862e58fba8feafe3abc8cca9caa5adb5ba490/sentence_transformers/SentenceTransformer.py#L189-L191) :
```python
self.prompts = {"query": "", "document": ""}
if prompts:
    self.prompts.update(prompts)
```

**The Problem:**
Even when a model has valid prompts (i.e. task specific), the empty default prompts (`""`) make the prompt to have now necessary this two keys empty or not. This means `self.prompts` might contain a mix of valid prompts, empty ones and ones with wrong format for MTEB.

When MTEB validates these prompts, it calls `self.validate_task_to_prompt_name(self.model.prompts)`. This appears to trigger a `KeyError` exception, which is caught in this code block : 
https://github.com/embeddings-benchmark/mteb/blob/cfa27d7209f066a96a619c9e21a0cc6781d11fd6/mteb/models/sentence_transformer_wrapper.py#L46-L51

The `KeyError` exception causes MTEB to set `model_prompts = None` and log the warning message: `"Model prompts are not in the expected format. Ignoring them."`

So what we have right now is : 

- No prompts in config → No warning message
- Wrong format prompts in config → Warning message "ignoring prompts"

And what i think should be nice to have is : 
- If only empty default prompts are present (`{"query": "", "document": ""}`) => Ignore them silently without raising the warning message
- If there are prompts with valid formats => Use them for evaluation
- If there are generic prompts for "query" and "document" in the config but no task-specific prompts ) => Use the generic ones as defaults

This approach would allow MTEB to handle ST v5's initialization pattern while still utilizing any valid prompts that are available. I'm not really aware of the bigger picture of the repo so i might miss some complications in my solution.

Thanks ! 

* Arthur BRESNU

### To reproduce

Having Latest versions of MTEB and Sentence Transformers v5 with a model containing task-specific prompts.

### Are you interested to contribute a fix for this bug?

No

Samoed:
> If there are prompts with valid formats => Use them for evaluation

[...]

PR #2912, opened 2025-07-17
fix: change `passage` prompt to `document` 

Samoed:
Change `passage` prompt name to `document` in all places to make compatible with `SentenceTransformers` `v5`.

Closes https://github.com/embeddings-benchmark/mteb/issues/2896


isaac-chung:
Can we please get the tests passing?

isaac-chung:
There's no change to pyproject.toml yet. Would allowing ST v5 pass all the tests?

Samoed:
I don't think we need to change anything in pyproject, because tests is run with sentence transformers `v5` and everything is working

isaac-chung (approved):
You're right! My bad. Good to go.

KennethEnevoldsen:
isn't this a breaking change for anyone implementing their own models?

Samoed:
I think probably yes

KennethEnevoldsen:
Hmm, probably shouldn't have been merged as a fix then (feat at least if not included in v2)

Well don't think we should change it now, but worth keeping in mind in the future

PR #2915, opened 2025-07-20
dataset: add Clotho by creating the datasets on the fly

### LINKED ISSUE #2053: Add Clotho
https://arxiv.org/pdf/1910.09387

isaac-chung:
Fixes #2053

model: laion/clap-htsat-fused.
```
ClothoA2TRetrieval
        "cv_recall_at_1": 0.17895,
        "cv_recall_at_3": 0.31962,
        "cv_recall_at_5": 0.40574,
        "cv_recall_at_10": 0.53397,

ClothoT2ARetrieval
        "cv_recall_at_1": 0.13142,
        "cv_recall_at_3": 0.26285,
        "cv_recall_at_5": 0.33572,
        "cv_recall_at_10": 0.44655,
```

### DIFF
--- a/mteb/tasks/Audio/Any2AnyRetrieval/Clotho.py
+++ b/mteb/tasks/Audio/Any2AnyRetrieval/Clotho.py
@@ -46,7 +46,7 @@ class ClothoA2TRetrieval(AbsTaskAny2AnyRetrieval):
         if self.data_loaded:
             return
 
-        ds = load_dataset(**self.metadata.dataset, split="test")
+        ds = load_dataset(**self.metadata.dataset, split="test", keep_in_memory=False)
 
         queries_ = {"id": [], "modality": [], "audio": []}
         corpus_ = {"id": [], "modality": [], "text": []}
@@ -130,7 +130,7 @@ class ClothoT2ARetrieval(AbsTaskAny2AnyRetrieval):
         if self.data_loaded:
             return
 
-        ds = load_dataset(**self.metadata.dataset)
+        ds = load_dataset(**self.metadata.dataset, split="test", keep_in_memory=False)
 
         queries_ = {"id": [], "modality": [], "text": []}
         corpus_ = {"id": [], "modality": [], "audio": []}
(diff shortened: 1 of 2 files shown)


AdnanElAssadi:
Hey @isaac-chung 

Any reason why you are doing the preprocessing step at runtime? Why don't you just preprocess the dataset and push it to MTEB with its correct format?

isaac-chung:
> Hey @isaac-chung
> 
> Any reason why you are doing the preprocessing step at runtime? Why don't you just preprocess the dataset and push it to MTEB with its correct format?

Hey @AdnanElAssadi56 , good question. It's something I just realized recently. Right now we'd run both T2A and A2T - this way, the dataset is only downloaded once. For SoundDescs, my machine struggles to download the T2A dataset even though A2T download was fine.

AdnanElAssadi:
> > Hey @isaac-chung
> > Any reason why you are doing the preprocessing step at runtime? Why don't you just preprocess the dataset and push it to MTEB with its correct format?
> 
> Hey @AdnanElAssadi56 , good question. It's something I just realized recently. Right now we'd run both T2A and A2T - this way, the dataset is only downloaded once. For SoundDescs, my machine struggles to download the T2A dataset even though A2T download was fine.

Yeah, especially since the current script assumes that machines can handle the entire dataset without issues. On Colab, for example, I couldn’t run the script on the Fleurs dataset directly , I had to stream it and use generators just to avoid memory crashes.

Another thing worth mentioning is the current approach to handling multilingual datasets (e.g., Jamalt) could probably be improved. From what I understand, we just use the usual configs and add a language column, then group and split by language at runtime. For something like Fleurs, that would mean simultaneously having 102 languages for each of the queries, qrels, and corpus components (which is probably very huge). Someone already picked up the issue for Fleurs, but from what I’ve seen, handling it properly isn’t very straightforward.

isaac-chung:
@AdnanElAssadi56 feel free to assign Fleurs to yourself.

PR #2922, opened 2025-07-21
Use `mteb.get_model` in adding_a_dataset.md

Samoed:
Ref https://github.com/embeddings-benchmark/results/pull/240#discussion_r2218387162

### DIFF
--- a/docs/adding_a_dataset.md
+++ b/docs/adding_a_dataset.md
@@ -242,13 +242,12 @@ Before you commit, here is a checklist you should complete before submitting:
 
 An easy way to test it is using:
 ```python
-from mteb import MTEB
-from sentence_transformers import SentenceTransformer
+from mteb import MTEB, get_model
 
 # Define the sentence-transformers model name
 model_name = "sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2"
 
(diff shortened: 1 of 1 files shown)


isaac-chung (changes requested):
I'm not convinced that this is the correct fix.

> Yep so I reran the model using mteb.get_model and it seems to produce the right model meta data. I believe the sentence transformers version of the model is newer, leading to different meta data.

What is the root issue here? Is it that 
a) The revisions are the same, but the ST model meta has params that are needed and the model meta in mteb is lacking? or
b) or ST model has a newer revision?
c) something else?

Samoed:
> b) or ST model has a newer revision?

Yes, model was updated on the hub

isaac-chung:
> > b) or ST model has a newer revision?
> 
> Yes, model was updated on the hub

Then the model meta revision should be updated.

Samoed:
The problem was with e5-small model and I don't think we need update anything on our side. 

Overall, we recommend users to use `mteb.get_model`, and it's strange that our documentation variant with passing a `SentenceTransformer` model has been left out, because they would use same code with different models

isaac-chung (approved):
You're right! I had read this backwards. All good here.

PR #2932, opened 2025-07-22
Audio Models Batch Fix

AdnanElAssadi56:
Added batch implementation to discussed models.

### DIFF
--- a/mteb/models/msclap_models.py
+++ b/mteb/models/msclap_models.py
@@ -201,7 +201,7 @@ class MSClapWrapper:
         prompt_type: PromptType | None = None,
         **kwargs: Any,
     ) -> np.ndarray:
-        return self.get_text_embeddings(inputs)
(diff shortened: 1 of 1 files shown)


mteb/models/msclap_models.py
```diff
@@ -227,7 +201,7 @@ def encode(
         prompt_type: PromptType | None = None,
         **kwargs: Any,
     ) -> np.ndarray:
-        return self.get_text_embeddings(inputs, **kwargs)
+        return self.get_text_embeddings(inputs)
```
isaac-chung: Add back please?

isaac-chung (approved):
In general looks good. Have you run a small batch on these to test if they work?

AdnanElAssadi:
> In general looks good. Have you run a small batch on these to test if they work?

Yup

PR #2939, opened 2025-07-22
fix: Prevent incorrectly passing "selector_state" to `get_benchmark`

KennethEnevoldsen:
The leaderboard would have (silent) errors where `get_benchmark` lead to a KeyError due to "selector_state" being passed as a default value. Setting `DEFAULT_BENCMARK_NAME` as the value solves this issue.


Samoed (approved):
This is better fix than https://github.com/embeddings-benchmark/mteb/pull/2881

KennethEnevoldsen:
Yea so it doesn't target that specific issue, but it is something we should fix regardless

PR #2942, opened 2025-07-22
dataset: Add JinaVDR

maximilianwerk:
Hey, we would like to contribute the JinaVDR benchmark to MTEB.

Our aim with the JinaVDR (Visual Document Retrieval) benchmark is to expand upon the work of these prior benchmarks by incorporating visually rich multilingual documents with complex layouts like graphs, charts, and tables (mixed with text and images), as well as adding real-world queries and questions.

The benchmarks were each run on an H100 GPU.

I experienced some OOM errors when running with big batch sizes. Thus, the change for the colpali models in the `similarity` function.

Please find the results here: https://github.com/embeddings-benchmark/results/pull/242

- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command. It is important to reduce the `batch_size` in order to have enough memory to run these models.
  - [x] vidore/colpali-v1.2
  - [x] vidore/colpali-v1.3
  - [x] vidore/colqwen2.5-v0.2
  - [x] jinaai/jina-embeddings-v4
- [x] I have considered the size of the dataset and reduced it if it is too big (2048 examples is typically large enough for most tasks): All datasets have a maximum of 1000 documents.

### DIFF
--- a/mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
+++ b/mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
@@ -39,7 +39,6 @@ def get_langs(langs: list[str]) -> dict[str, list[str]]:
 
 
 COMMON_METADATA = {
-    "description": "Please refer to the reference repo in order to get more information about the benchmark.",
     "type": "DocumentUnderstanding",
     "category": "t2i",
     "eval_splits": ["test"],
@@ -178,6 +177,7 @@ def load_data(self, **kwargs):
 class JinaVDRMedicalPrescriptionsRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRMedicalPrescriptionsRetrieval",
+        description="Retrieve medical prescriptions based on templated queries.",
         reference="https://huggingface.co/datasets/jinaai/medical-prescriptions_beir",
         dataset={
             "path": "jinaai/medical-prescriptions_beir",
@@ -198,6 +198,7 @@ class JinaVDRMedicalPrescriptionsRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRStanfordSlideRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRStanfordSlideRetrieval",
+        description="Retrieve scientific and engineering slides based on human annotated queries.",
         reference="https://huggingface.co/datasets/jinaai/stanford_slide_beir",
         dataset={
             "path": "jinaai/stanford_slide_beir",
@@ -218,6 +219,7 @@ class JinaVDRStanfordSlideRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRDonutVQAISynHMPRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRDonutVQAISynHMPRetrieval",
+        description="Retrieve medical records based on templated queries.",
         reference="https://huggingface.co/datasets/jinaai/donut_vqa_beir",
         dataset={
             "path": "jinaai/donut_vqa_beir",
@@ -238,6 +240,7 @@ class JinaVDRDonutVQAISynHMPRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRTableVQARetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRTableVQARetrieval",
+        description="Retrieve scientific tables based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/table-vqa_beir",
         dataset={
             "path": "jinaai/table-vqa_beir",
@@ -258,6 +261,7 @@ class JinaVDRTableVQARetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRChartQARetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRChartQARetrieval",
+        description="Retrieve charts based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/ChartQA_beir",
         dataset={
             "path": "jinaai/ChartQA_beir",
@@ -278,6 +282,7 @@ class JinaVDRChartQARetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRTQARetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRTQARetrieval",
+        description="Retrieve textbook pages (images and text) based on LLM generated queries from the text.",
         reference="https://huggingface.co/datasets/jinaai/tqa_beir",
         dataset={
             "path": "jinaai/tqa_beir",
@@ -298,6 +303,7 @@ class JinaVDRTQARetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDROpenAINewsRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDROpenAINewsRetrieval",
+        description="Retrieve news articles from the OpenAI news website based on human annotated queries.",
         reference="https://huggingface.co/datasets/jinaai/openai-news_beir",
         dataset={
             "path": "jinaai/openai-news_beir",
@@ -318,6 +324,7 @@ class JinaVDROpenAINewsRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDREuropeanaDeNewsRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDREuropeanaDeNewsRetrieval",
+        description="Retrieve German news articles based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/europeana-de-news_beir",
         dataset={
             "path": "jinaai/europeana-de-news_beir",
@@ -338,6 +345,7 @@ class JinaVDREuropeanaDeNewsRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDREuropeanaEsNewsRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDREuropeanaEsNewsRetrieval",
+        description="Retrieve Spanish news articles based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/europeana-es-news_beir",
         dataset={
             "path": "jinaai/europeana-es-news_beir",
@@ -358,6 +366,7 @@ class JinaVDREuropeanaEsNewsRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDREuropeanaItScansRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDREuropeanaItScansRetrieval",
+        description="Retrieve Italian historical articles based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/europeana-it-scans_beir",
         dataset={
             "path": "jinaai/europeana-it-scans_beir",
@@ -378,6 +387,7 @@ class JinaVDREuropeanaItScansRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDREuropeanaNlLegalRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDREuropeanaNlLegalRetrieval",
+        description="Retrieve Dutch historical legal documents based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/europeana-nl-legal_beir",
         dataset={
             "path": "jinaai/europeana-nl-legal_beir",
@@ -398,6 +408,7 @@ class JinaVDREuropeanaNlLegalRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRHindiGovVQARetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRHindiGovVQARetrieval",
+        description="Retrieve Hindi government documents based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/hindi-gov-vqa_beir",
         dataset={
             "path": "jinaai/hindi-gov-vqa_beir",
@@ -418,6 +429,7 @@ class JinaVDRHindiGovVQARetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRAutomobileCatelogRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRAutomobileCatelogRetrieval",
+        description="Retrieve automobile marketing documents based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/automobile_catalogue_jp_beir",
         dataset={
             "path": "jinaai/automobile_catalogue_jp_beir",
@@ -438,6 +450,7 @@ class JinaVDRAutomobileCatelogRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRBeveragesCatalogueRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRBeveragesCatalogueRetrieval",
+        description="Retrieve beverages marketing documents based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/beverages_catalogue_ru_beir",
         dataset={
             "path": "jinaai/beverages_catalogue_ru_beir",
@@ -458,6 +471,7 @@ class JinaVDRBeveragesCatalogueRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRRamensBenchmarkRetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRRamensBenchmarkRetrieval",
+        description="Retrieve ramen restaurant marketing documents based on LLM generated queries.",
         reference="https://huggingface.co/datasets/jinaai/ramen_benchmark_jp_beir",
         dataset={
             "path": "jinaai/ramen_benchmark_jp_beir",
@@ -478,6 +492,7 @@ class JinaVDRRamensBenchmarkRetrieval(AbsTaskAny2AnyRetrieval):
 class JinaVDRJDocQARetrieval(AbsTaskAny2AnyRetrieval):
     metadata = TaskMetadata(
         name="JinaVDRJDocQARetrieval",
+        description="Retrieve Japanese documents in various formats based on human annotated queries.",
         reference="https://huggingface.co/datasets/jinaai/jdocqa_beir",
         dataset={
             "path": "jinaai/jdocqa_beir",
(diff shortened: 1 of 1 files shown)


mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1520 @@
...
+        dataset={
+            "path": "jinaai/infovqa_beir",
+            "revision": "main",
+        },
+        eval_langs=["eng-Latn"],
+        domains=["Web"],
+        license="mit",
+        annotations_creators="derived",
+        sample_creation="found",
+        descriptive_stats={
```
Samoed: Can you remove descriptive_stats everywhere? This is deprecated format

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1520 @@
...
+        splits=self.metadata_dict["eval_splits"],
+        langs=self.metadata_dict["eval_langs"],
+        cache_dir=kwargs.get("cache_dir", None),
+        revision=self.metadata_dict["dataset"]["revision"],
+    )
+
+    self.data_loaded = True
+
+
+class JinaVDRMedicalPrescriptionsRetrieval(MultilingualTask, AbsTaskAny2AnyRetrieval):
```
Samoed: Can you remove `MultilingualTask` from tasks with `eval_langs=["eng-Latn"],`

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1520 @@
...
+    load_data = load_data
+
+
+class JinaVDRStanfordSlideRetrieval(MultilingualTask, AbsTaskAny2AnyRetrieval):
+
+    metadata = TaskMetadata(
+        name="JinaVDRStanfordSlideRetrieval",
+        dataset={
+            "path": "jinaai/stanford_slide_beir",
+            "revision": "main",
```
Samoed: Can you specify revision everywhere?

mteb/models/colpali_models.py
```diff
@@ -131,7 +131,7 @@ def calculate_probs(self, text_embeddings, image_embeddings):
         return scores.softmax(dim=-1)
 
     def similarity(self, a, b):
-        return self.processor.score(a, b)
+        return self.processor.score(a, b, batch_size=8)
```
KennethEnevoldsen: This should allow the user to pass different processor kwargs along.

```suggestion
        return self.processor.score(a, b, **self.processor_kwargs)
```

mteb/models/jina_models.py
```diff
@@ -337,13 +337,13 @@ def get_text_embeddings(
         else:
             self.vector_type = "single_vector"
 
+        if not isinstance(texts, list):
+            texts = list(texts)
```
KennethEnevoldsen: What are the cases where it is not a list?

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1520 @@
...
+    "zh": ["zho-Hans"],
+}
+
+
+def get_langs(langs: list[str]) -> dict[str, list[str]]:
+    return {lang: _LANGS[lang] for lang in langs}
+
+
+COMMON_METADATA = {
+    "description": "Retrieve associated pages according to questions or related text.",
```
KennethEnevoldsen: Can you specify this per dataset? It needs to be detailed enough for the user to understand what the task is about and how it differs from other tasks.

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1520 @@
...
+
+    metadata = TaskMetadata(
+        name="JinaVDRStanfordSlideRetrieval",
+        dataset={
+            "path": "jinaai/stanford_slide_beir",
+            "revision": "main",
+        },
+        eval_langs=["eng-Latn"],
+        domains=["Academic"],
+        license="not specified",
```
KennethEnevoldsen, line 210: Ideally we want the license to be specifed?

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1520 @@
...
+}
+
+
+def get_langs(langs: list[str]) -> dict[str, list[str]]:
+    return {lang: _LANGS[lang] for lang in langs}
+
+
+COMMON_METADATA = {
+    "description": "Retrieve associated pages according to questions or related text.",
+    "reference": "https://arxiv.org/abs/2506.18902",
```
KennethEnevoldsen: While this reference makes great sense for the benchmark itself, I would probably add more detailed refs to the individual datasets (e.g. for EuropeanaItScans it seems like this [repo](https://huggingface.co/datasets/jinaai/europeana-it-scans) is a better reference if I want to understand something about the data).

mteb/benchmarks/benchmarks.py
```diff
@@ -2029,3 +2029,69 @@
...
+            "JinaVDRTatQARetrieval",
+            "JinaVDRInfovqaRetrieval",
+            "JinaVDRDocVQARetrieval",
+            "JinaVDRDocQAGovReportRetrieval",
+            "JinaVDRTabFQuadRetrieval",
+            "JinaVDRDocQAEnergyRetrieval",
+            "JinaVDRArxivQARetrieval",
+        ],
+    ),
+    description="Multilingual, domain-diverse and layout-rich document retrieval benchmark.",
```
KennethEnevoldsen, line 2205: It sounds like (from the paper) that this extends the existing Visual Document Retrieval benchmark, which is already implemented in MTEB. Any reason to use `JinaVDRTabFQuadRetrieval` instead of `VidoreTabfquadRetrieval`?

In addition, it would be great with an expansion of the description - e.g., mentioning that it is an expansion  and indicating what way it expands the benchmark

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
maximilianwerk: Since you already warn, that `MultilingualTask` will be removed with 2.0, should I remove it from all Tasks? That was something I was also wondering while creating the tasks.

Samoed: No, it's required to run correctly your tasks in `main`

KennethEnevoldsen:
What a great PR @maximilianwerk! Added a few additional comments

mteb/models/jina_models.py
maximilianwerk: When I used an older version of MTEB (like two months old) I got sometimes other iterables then lists. If this is fixed on MTEB side and there is more consistency, great. I'm very happy to remove it. :)

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
maximilianwerk: I got "inspired" by the vidore2 benchmark and would prefer to keep the original datasets on huggingface as the only source of truth. When we update information there, we would always need to sync it with here, which seems like a bad idea and will rot soon, non?

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
maximilianwerk: I understand that. For some datasets, that is hardly doable, since the original authors did not always specified the license. How would you like to proceed? Remove the task for the benchmark?

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
maximilianwerk: These repos are always referenced in the actual huggingface datasets. But if you prefer, I'll double that here.

mteb/benchmarks/benchmarks.py
maximilianwerk: I will come back with a proper answer here soon. We did some automatic filtering on the queries, since some where bad (like "what can you see in this image?"), but we may have filtered a bit too aggressively.

maximilianwerk: Finally, here are some more details from our side:
We recognized, that synthetic queries often have the above mentioned problem: In order to make sense of the query, you actually need to know the document. This is not specific to the vidore benchmark, but true for all synthetic generated queries. As a consequence, we did a rather restrictive query classification, whether a query makes sense, without knowing the document. We used the following prompt for filtering with a LLM:

```
Your task is to categorize each search query into one of the following two classes: VALID or INVALID.

Criteria for VALID queries:
1. VALID queries should not be vague or ambiguous, they must provide enough context for search outside a specific set of documents.
2. VALID queries should not depend on specific documents, charts, tables, but can mention known entities (like individuals, institutions, etc.).

Queries that do not meet the given criteria should be classified as INVALID.

Format for response:
Query: "..."
Class: VALID/INVALID
Explanation: "..."

Examples for reference:

Query: "How are concerns logged and tracked throughout the process?"
Class: INVALID
Explanation: This query does not contain enough information, it is not clear what "process" is being referenced.

Query: "For a married couple filing jointly, what is the withholding amount according to the Tax Withholding table?"
Class: INVALID
Explanation: This query depends on a specific "Tax Withholding" table.

Query: "What is the role of Günther Oberhofer at Conrad Electronic?"
Class: VALID
Explanation: This query provides enough context by asking about a specific person at a known company.

Query: "Under what circumstances might the store send emails to customers?" 
Class: INVALID
Explanation: This query is too vague because it does not specify which store is being referred to.

Query: "What is the premise of the story in Star Divide Ascension Series Book 2?"
CLASS: VALID
Explanation: This query provides enough context for a search by specifying the title of a particular book within a series.

Query: "What action will be taken regarding the trading of BROKEN HILL PROSPECTING LIMITED's securities?" 
Class: INVALID
Explanation: The query lacks context such as timeframe, specific events, or responsible entities, making it vague.

Query: "What is the purpose of Tallan's Accessible Web Portal?"
Class: VALID
Explanation: This query inquires the purpose of a well known portal.

Query: "What are some examples of how pupils at Doncaster School for the Deaf are involved in enrichment opportunities?"
Class: VALID
Explanation: This query provides enough context for a search as it specifies a particular school (Doncaster School for the Deaf).

Using the guidelines and format provided above, categorize the following query: "{{ query }}".
```

When you look for example at https://huggingface.co/datasets/vidore/docvqa_test_subsampled/, quite some queries have this property (e.g. `What is the name of the person in the CC field ?`, ` What is the personnel costs in the 4th year?`). We recognize, that for some datasets we over-filtered quite heavily. However, from our perspective, we are fine to sacrifice a good amount of "valid queries" in order to get more relevance in the remaining ones.

If you don't like our filtering approach on existing benchmarks, we understand that. Then, we would prefer to remove the vidore tasks from our benchmark instead of using the original ones. Please tell me, what you prefer. :)

mteb/models/jina_models.py
KennethEnevoldsen: If that is the case, let us keep it (v2 does a lot of this standardization)

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
KennethEnevoldsen: The dataset will be revision-locked, so it should not change unless a new version is created.

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
KennethEnevoldsen: If it is not specified we of course can't add it - I just want to make sure that it has been checked

mteb/benchmarks/benchmarks.py
KennethEnevoldsen: Ahh, if that is the case, I would add `superseeded_by` to the relevant video tasks. E.g. `superseeded_by = "JinaVDRTabFQuadRetrieval"` to `VidoreTabfquadRetrieval`.I would also describe this in the description (short, potentially just with a markdown comment link to this comment or reference to paper section). 

You also need to add the `adapted_from` to the tasks so that we can track cross-contamination across the two datasets.

Of course, it can be argued that it does not directly `superseeded_by` the previous dataset, but in this case I would probably be fine with raising a warning if people use `VidoreTabfquadRetrieval`, to let them know that a filtered version is available - they can always check the description to see what has changed.

danielcer:
Is this PR active with JinaVDR still likely to be added to MTEB?

maximilianwerk:
> Is this PR active with JinaVDR still likely to be added to MTEB?

Yes, I will fix the missing requirements today/tomorrow.

mteb/benchmarks/benchmarks.py
maximilianwerk: I would leave this decision to the vidore maintainers. They originally created or curated the datasets and as such they should have the decision, whether our filtering it too strict. Especially, since we heavily over-filtered.

maximilianwerk: Nevertheless, I'll add a link to this discussion in the huggingface dataset descriptions and update the refs to the correct version afterwards.

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
maximilianwerk: I specified it for all datasets, where I could find a license.

mteb/benchmarks/benchmarks.py
maximilianwerk: @KennethEnevoldsen I added the link to the original dataset and to this discussion to our non-beir-format datasets (e.g. https://huggingface.co/datasets/jinaai/docqa_healthcare_industry). As a consequence, I did not need to update the references.

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
maximilianwerk: I get your point, but the rotting of information is still true. Furthermore, I cannot find, where the description of a dataset is displayed on the MTEB leaderboard. When I go to "Task information", I only get languages, task type etc. And the actual link to the dataset. I'd assume, whoever is already at the state looking into the code, is already far beyond reading the description.

image

KennethEnevoldsen: While I understand your concern, this is currently a requirement for all task submissions. It should be possible to inspect it using `mteb.get_task`. I also don't expect this description ever to be outdated (since it is revision-locked).

The dataset description in the reference also doesn't necessarily match the implementation in the codebase.

maximilianwerk: I created my descriptions similar to other existing descriptions. I hope they match your needs. If not, let me know. :)

maximilianwerk:
Hey, is there anything missing from my side?

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1126 @@
...
+        sample_creation="found",
+        **COMMON_METADATA,
+    )
+
+    load_data = load_data
+
+
+class JinaVDRAutomobileCatelogRetrieval(AbsTaskAny2AnyRetrieval):
+    metadata = TaskMetadata(
+        name="JinaVDRAutomobileCatelogRetrieval",
```
Samoed, line 431:
```suggestion
        name="JinaVDRAutomobileJaCatalogRetrieval",
```

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1126 @@
...
+        sample_creation="found",
+        **COMMON_METADATA,
+    )
+
+    load_data = load_data
+
+
+class JinaVDRBeveragesCatalogueRetrieval(AbsTaskAny2AnyRetrieval):
+    metadata = TaskMetadata(
+        name="JinaVDRBeveragesCatalogueRetrieval",
```
Samoed, line 452:
```suggestion
        name="JinaVDRBeveragesRuCatalogueRetrieval",
```

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1126 @@
...
+        license="not specified",
+        annotations_creators="LM-generated",
+        sample_creation="found",
+        **COMMON_METADATA,
+    )
+
+    load_data = load_data
+
+
+class JinaVDRBeveragesCatalogueRetrieval(AbsTaskAny2AnyRetrieval):
```
Samoed, line 450:
```suggestion
class JinaVDRBeveragesRuCatalogueRetrieval(AbsTaskAny2AnyRetrieval):
```

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1126 @@
...
+        license="not specified",
+        annotations_creators="LM-generated",
+        sample_creation="found",
+        **COMMON_METADATA,
+    )
+
+    load_data = load_data
+
+
+class JinaVDRAutomobileCatelogRetrieval(AbsTaskAny2AnyRetrieval):
```
Samoed, line 429:
```suggestion
class JinaVDRAutomobileJaCatalogRetrieval(AbsTaskAny2AnyRetrieval):
```

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1126 @@
...
+        sample_creation="found",
+        **COMMON_METADATA,
+    )
+
+    load_data = load_data
+
+
+class JinaVDRRamensBenchmarkRetrieval(AbsTaskAny2AnyRetrieval):
+    metadata = TaskMetadata(
+        name="JinaVDRRamensBenchmarkRetrieval",
```
Samoed, line 473:
```suggestion
        name="JinaVDRRamensJaBenchmarkRetrieval",
```

mteb/tasks/Image/Any2AnyRetrieval/multilingual/JinaVDRBenchRetrieval.py
```diff
@@ -0,0 +1,1126 @@
...
+        license="not specified",
+        annotations_creators="LM-generated",
+        sample_creation="found",
+        **COMMON_METADATA,
+    )
+
+    load_data = load_data
+
+
+class JinaVDRRamensBenchmarkRetrieval(AbsTaskAny2AnyRetrieval):
```
Samoed, line 471:
```suggestion
class JinaVDRRamensJaBenchmarkRetrieval(AbsTaskAny2AnyRetrieval):
```

KennethEnevoldsen (approved):
Thanks for adding the descriptions

I am not sure I agree with the name changes from @Samoed

Samoed:
> I am not sure I agree with the name changes from @Samoed. I am assuming that they are minor suggestions - so I think this is good to merge (@Samoed can you confirm here?)

Yes, they're minor, but I tried to follow naming scheme. Also some of tasks are not multilingual and maybe they should be in task specific folders instead in one big

KennethEnevoldsen:
yeah. I think I would prefer them in one file or all in individual files. So I think this is preferable (if not it will not be a breaking change to fix it)

PR #2946, opened 2025-07-23
dataset: Add Dapfam patent retrieval tasks

iliass-y:
**Dataset**: DAPFAM –  Domain-Aware Patent Retrieval Dataset
Aggregated at the Family Level (Ayaou et al. 2025,  [paper link](https://arxiv.org/abs/2506.22141) ).
* **New TaskSubtype**: `Patent retrieval` (added to `TaskMetadata`).
* **Variants (18)**  
  * Domains: ALL / IN / OUT  
  * Query text: **TitleAbstract**  or  **TitleAbstractClaims**  
  * Corpus text: **TitleAbstract**, **TitleAbstractClaims** *(paper)*, **TitleAbstractClaimsDescription** *(extra)*  
  * Classes are named `Dapfam_{Domain}_{Query}_{Corpus}`.

This table lists the 18 tasks generated by `DAPFAMPatentRetrieval.py`.  
Columns:

- **Task name**: the Python class registered in MTEB  
- **Domain**: ALL / IN / OUT  
- **Query fields**: which fields make up the query text  
- **Corpus fields**: which fields make up the corpus text  
- **In paper**: ✅ were these variants in Ayaou et al. (2025)?

| Task name                                                         | Domain | Query fields           | Corpus fields               | In paper? |
|-------------------------------------------------------------------|:------:|:-----------------------:|:---------------------------:|:---------:|
| `Dapfam_ALL_TitleAbstract_TitleAbstract`                          | ALL    | Title, Abstract        | Title, Abstract             | ✨ extra   |
| `Dapfam_ALL_TitleAbstract_TitleAbstractClaims`                    | ALL    | Title, Abstract        | Title, Abstract, Claims     | ✅         |
| `Dapfam_ALL_TitleAbstract_TitleAbstractClaimsDescription`         | ALL    | Title, Abstract        | Title, Abstract, Claims, Description | ✨ extra   |
| `Dapfam_ALL_TitleAbstractClaims_TitleAbstract`                    | ALL    | Title, Abstract, Claims| Title, Abstract             | ✨ extra   |
| `Dapfam_ALL_TitleAbstractClaims_TitleAbstractClaims`              | ALL    | Title, Abstract, Claims| Title, Abstract, Claims     | ✅         |
| `Dapfam_ALL_TitleAbstractClaims_TitleAbstractClaimsDescription`   | ALL    | Title, Abstract, Claims| Title, Abstract, Claims, Description | ✨ extra   |
| `Dapfam_IN_TitleAbstract_TitleAbstract`                           | IN     | Title, Abstract        | Title, Abstract             | ✨ extra   |
| `Dapfam_IN_TitleAbstract_TitleAbstractClaims`                     | IN     | Title, Abstract        | Title, Abstract, Claims     | ✅         |
| `Dapfam_IN_TitleAbstract_TitleAbstractClaimsDescription`          | IN     | Title, Abstract        | Title, Abstract, Claims, Description | ✨ extra   |
| `Dapfam_IN_TitleAbstractClaims_TitleAbstract`                     | IN     | Title, Abstract, Claims| Title, Abstract             | ✨ extra   |
| `Dapfam_IN_TitleAbstractClaims_TitleAbstractClaims`               | IN     | Title, Abstract, Claims| Title, Abstract, Claims     | ✅         |
| `Dapfam_IN_TitleAbstractClaims_TitleAbstractClaimsDescription`    | IN     | Title, Abstract, Claims| Title, Abstract, Claims, Description | ✨ extra   |
| `Dapfam_OUT_TitleAbstract_TitleAbstract`                          | OUT    | Title, Abstract        | Title, Abstract             | ✨ extra   |
| `Dapfam_OUT_TitleAbstract_TitleAbstractClaims`                    | OUT    | Title, Abstract        | Title, Abstract, Claims     | ✅         |
| `Dapfam_OUT_TitleAbstract_TitleAbstractClaimsDescription`         | OUT    | Title, Abstract        | Title, Abstract, Claims, Description | ✨ extra   |
| `Dapfam_OUT_TitleAbstractClaims_TitleAbstract`                    | OUT    | Title, Abstract, Claims| Title, Abstract             | ✨ extra   |
| `Dapfam_OUT_TitleAbstractClaims_TitleAbstractClaims`              | OUT    | Title, Abstract, Claims| Title, Abstract, Claims     | ✅         |
| `Dapfam_OUT_TitleAbstractClaims_TitleAbstractClaimsDescription`   | OUT    | Title, Abstract, Claims| Title, Abstract, Claims, Description | ✨ extra   |

> **Note:** “✨ extra” variants were _not_ in the original paper but are provided here for extended evaluation.

### How to replicate paper document neural embedding tests :
```python
import mteb, torch

model = SentenceTransformer(
    "Snowflake/snowflake-arctic-embed-m-v2.0",
    trust_remote_code=True,
    model_kwargs={"torch_dtype":"float16"},
).cuda().eval()

tasks = [t for t in mteb.get_tasks() if t.startswith("Dapfam") and "_Description" not in t]
mteb.MTEB(tasks).run(model, output_folder="results/snowflake_arctic_fp16")
```

- [X] I have outlined why this dataset is filling an existing gap in `mteb`

Thank you for reviewing! I’m happy to address any feedback.


mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+                task_subtypes=["Patent retrieval", "Article retrieval"],
+                eval_splits=["test"],
+                eval_langs=["eng-Latn"],
+                main_score="ndcg@10",
+                date=("2025-06-30", "2025-06-30"),
+                domains=["Engineering", "Chemistry", "Legal"],
+                license="not specified",
+                annotations_creators="derived",
+                sample_creation="created",
+                judged_docs_only_flag=False,
```
KennethEnevoldsen: This seems to have been accidentally added

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+# ——— register all tasks ———
+for domain, domlbl in DOMAIN_LABELS.items():
+    for qn, qf in QUERY_VARIANTS.items():
+        for cn, cf in CORPUS_VARIANTS.items():
+            task_name = f"Dapfam_{domain}_{qn}_{cn}"
+            metadata = TaskMetadata(
+                name=task_name,
+                description=f"DAPFAM [{domain}] Q={qn} / C={cn}",
```
KennethEnevoldsen: The description has to be detailed enough for a user to know what the dataset is about (it should also be clear how the datasets are different

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+                dataset={"path": HF_REPO, "revision": "main"},
```
KennethEnevoldsen: if you want to save on reused metadata you can always:

```
# define default metadata
_default_metadata = dict(...)
# unpack during metadata definition 
TaskMetadata(task_name="name", ..., **_default_metadata)

File: mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py [hunk 4]
```diff
@@ -0,0 +1,220 @@
...
+                reference=REFERENCE,
+                type="Retrieval",
+                category="p2p",
+                task_subtypes=["Patent retrieval", "Article retrieval"],
+                eval_splits=["test"],
+                eval_langs=["eng-Latn"],
+                main_score="ndcg@10",
+                date=("2025-06-30", "2025-06-30"),
+                domains=["Engineering", "Chemistry", "Legal"],
+                license="not specified",
```
KennethEnevoldsen: license has to be specified

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+                description=f"DAPFAM [{domain}] Q={qn} / C={cn}",
+                dataset={"path": HF_REPO, "revision": "main"},
+                reference=REFERENCE,
+                type="Retrieval",
+                category="p2p",
+                task_subtypes=["Patent retrieval", "Article retrieval"],
+                eval_splits=["test"],
+                eval_langs=["eng-Latn"],
+                main_score="ndcg@10",
+                date=("2025-06-30", "2025-06-30"),
```
KennethEnevoldsen: What does this date correspond to? It should be time of data creation (patents were written from date to date)

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+        self.corpus = {"test": corpus}
+        self.queries = {"test": queries}
+        self.relevant_docs = {"test": qrels}
+        self.data_loaded = True
+        return self.corpus, self.queries, self.relevant_docs
+
+    return load_data
+
+
+def make_evaluate(domain_filter):
```
KennethEnevoldsen: Is this redefinition needed? If it is, we should make it clear in the docstring why it is needed.

One way to do that is. 

In v2 (which will soon be merged) we do allow for custom metrics

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+                category="p2p",
+                task_subtypes=["Patent retrieval", "Article retrieval"],
+                eval_splits=["test"],
+                eval_langs=["eng-Latn"],
+                main_score="ndcg@10",
+                date=("2025-06-30", "2025-06-30"),
+                domains=["Engineering", "Chemistry", "Legal"],
+                license="not specified",
+                annotations_creators="derived",
+                sample_creation="created",
```
KennethEnevoldsen: Not `"found"`?

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+                type="Retrieval",
+                category="p2p",
+                task_subtypes=["Patent retrieval", "Article retrieval"],
+                eval_splits=["test"],
+                eval_langs=["eng-Latn"],
+                main_score="ndcg@10",
+                date=("2025-06-30", "2025-06-30"),
+                domains=["Engineering", "Chemistry", "Legal"],
+                license="not specified",
+                annotations_creators="derived",
```
KennethEnevoldsen: Can you explain how it is derived? Just want to make sure that the label is correct

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+            def __init__(self):
+                super(self.__class__, self).__init__()
+                self.load_data()
+                self.calculate_metadata_metrics()
+
+            attrs = {
+                "__init__": __init__,
+                "metadata": metadata,
+                "load_data": make_load_data(qf, cf),
+                "evaluate": make_evaluate(domlbl),
```
KennethEnevoldsen: I would probably either 1) create a parent `DAPFAMRetrieval` class to create which the individual classes can inherit from or 2) create a function and then call that in the evaluate.

```
Class MyTask(AbsTaskRetrieval):
  def evaluate(...):
    return _my_custom_evaluate(...)
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+            }
+            globals()[task_name] = type(task_name, (AbsTaskRetrieval,), attrs)
```
KennethEnevoldsen: Let's not do this in a loop. It will be hard for us to maintain in the future

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,220 @@
...
+    "TitleAbstractClaimsDescription": [
+        "title_en",
+        "abstract_en",
+        "claims_text",
+        "description_en",
+    ],
+}
+
+
+def make_load_data(q_fields, c_fields):
```
KennethEnevoldsen: We generally need some type hints

KennethEnevoldsen (commented):
Thanks the PR! And congratulations on the paper - big fan of the making the evaluation accecible.

Generally there is nothing in the PR that seems to be breaking or incorrect, but we will need to do some refactors to support it long-term. There is some minor things that need to added(checked on the documentation side)

I would also strongly prefer not introducing the `.evaluate` method, as it will break in the soon-to-be-released v2 due to different model input which will allow for multiple modalities.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: First of all, thank you for the kind words.

indeed, this is fixed in the latest commit.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: I have added some extra details for each task.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Yes, license is now specified.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: At first it represented the date on which i published the paper and the first version of the data.
But i have switched it now to represent the min and max dates of available patents in the dataset.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: explanation was added as well as a "custom" evaluation function wrapped in evaluate, hopefully that makes it easier to maintain.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Same problem as above here.
Basically there was no human or expert annotation here nor was there an LLM generation.
For the relevance judgement it was a deterministic algorithmic approach to obtain them, hence my choice of "derived".

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: I actually ended up doing both in the last commit.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Loop dropped, tasks are now defined individually by inheriting from a base class and the AbsTaskRetrieval task.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Added type hints were it seemed appropriate.

iliass-y:
Hello @KennethEnevoldsen ,
Again, thank you for the kind words and the detailed comments.
I think that i have addressed as best as i could the various points you brought up.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,415 @@
...
+import numpy as np
+from datasets import load_dataset
+from sentence_transformers.quantization import quantize_embeddings
+from sklearn.metrics import average_precision_score
+
+from ....abstasks.AbsTaskRetrieval import AbsTaskRetrieval
+from ....abstasks.TaskMetadata import TaskMetadata
+
+# ───────────────────────────────────────────────────
+# GLOBAL CONSTANTS
```
KennethEnevoldsen:
```suggestion
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,415 @@
+# mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```
KennethEnevoldsen:
```suggestion
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,415 @@
...
+  title        = {DAPFAM: A Domain-Aware Patent Retrieval Dataset Aggregated at the Family Level},
+  author       = {Ayaou, Iliass and Cavallucci, Denis and Chibane, Hicham},
+  year         = {2025},
+  eprint       = {2506.22141},
+  archivePrefix= {arXiv},
+  primaryClass = {cs.CL}
+}"""
+
+# shared metadata
+_DEFAULT_META = dict(
```
KennethEnevoldsen:
```suggestion
_SHARED_METADATA = dict(
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,415 @@
...
+            qrels.setdefault(qid, {})[pid] = (
+                float(r["relevance_score"]),
+                r["domain_rel"],
+            )
+        # 4) Assign for MTEB
+        self.relevant_docs = {"test": qrels}
+        self.data_loaded = True
+        return self.corpus, self.queries, self.relevant_docs
+
+    # ------------ evaluation  (faithful to the paper) ------------
```
KennethEnevoldsen:
```suggestion
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,415 @@
...
+# ───────────────────────────────────────────────────
+# MIX-IN with shared logic + metric implementation
+class _DAPFAMMixin:
+    # class-level attributes are filled in each concrete subclass
+    domain_filter: Optional[str] = None
+    query_fields: List[str] = []
+    corpus_fields: List[str] = []
+    in_paper: bool = False
+
+    # ------------ data loading (identical for all variants) ------------
```
KennethEnevoldsen:
```suggestion
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,415 @@
...
+}
+
+# paper variants used in Table 4
+_IN_PAPER = {
+    ("TitleAbstract", "TitleAbstractClaims"),
+    ("TitleAbstractClaims", "TitleAbstractClaims"),
+}
+
+
+# ───────────────────────────────────────────────────
```
KennethEnevoldsen:
```suggestion
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,415 @@
...
+# AbsTaskRetrieval.__init__ will call self.load_data() and compute metadata automatically.
+
+
+# ---------- ALL domain ----------
+class Dapfam_ALL_TitleAbstract_TitleAbstract(_DAPFAMMixin, AbsTaskRetrieval):
+    domain_filter = None
+    query_fields = _QUERY_FIELDS["TitleAbstract"]
+    corpus_fields = _CORPUS_FIELDS["TitleAbstract"]
+    in_paper = False
+    metadata = _meta(__qualname__, "ALL • Query: TA  | Corpus: TA")
```
KennethEnevoldsen: No need for a function here

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,415 @@
...
+
+# ───────────────────────────────────────────────────
+# 18 explicit task classes (no loops)
+
+# NOTE: Each class only sets class-level attributes
+
+# AbsTaskRetrieval.__init__ will call self.load_data() and compute metadata automatically.
+
+
+# ---------- ALL domain ----------
```
KennethEnevoldsen:
```suggestion
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
KennethEnevoldsen: The current description seems to be: `"ALL • Query: TA | Corpus: TA"`. The description has to be clearer than this.

E.g., something like: 

> "A task consisting of documents from Wikipedia concerning birds and LLM-generated questions regarding the articles. The goal is to retrieve the document that contains the answer to the question."

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
KennethEnevoldsen: Hmm, thanks for adding it. Is the only difference the quantization? I couldn't find anything on why this is in the paper. Couldn't imagine that it makes a big difference, so I am unsure if the added complexity is justified.

We typically consider normalization a model setting and not something that happens during evaluation.

Are there any other variations from MTEB that I should be aware of?

I see that your implementation assumes that the model is built for cosine similarity - generally reasonable assumption - but as a want to encourage diverse submissions we are trying to built these assumptions out of the benchmark

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: expanded the description of each task.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: I was under the impression that i answered this one, must have skipped it by mistake.
The data in it's current format doesn't exist anywhere else because although i obviously didn't write the patents, the representation of the data was quite altered through various processing/data engineering stages and data aggregation strategies (the dataset is about patent families and not patent at the base level as well).

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Done in the last iteration of changes.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Okay,
you're right, it's not currently clear in the paper and i will get around to it to add it.

So besides the quantization and the cosine similarity ( that are now changeable through passing "similarity" and "quantize" arguments, there are some other differences that also warranted a custom eval function.

- For mAP weighting, w build a deterministic rank-based weight vector [K, K–1, …, 1] for the top-K positions instead of using the similarity scores, this is motivated by the fact that cosine scores obtained after quantizing have more chances of having equal values.

- If a query has no positive ground truth documents in the whole dataset, we assign metric = 1.0.
on average, out-of-domain relevants documents per query is only 5.74 ( out of domain linked patent are quite rare especially in some IPC domains), so some queries have zero out-of-domain targets, and we had to decide how to aggregate their scores.
Setting those query metrics to 1.0 ensures we don’t penalize the system simply because no ground-truth existed in that split. And it makes sens from a specific system usage perspective for out of domain patent retrieval.

Both of these differences do not affect the deterministic nature of the metrics and allow for comparing various models correctly.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,551 @@
...
+
+class Dapfam_OUT_TitleAbstractClaims_TitleAbstractClaimsDescription(
+    _DAPFAMMixin, AbsTaskRetrieval
+):
+    domain_filter = "OUT"
+    query_fields = _QUERY_FIELDS["TitleAbstractClaims"]
+    corpus_fields = _CORPUS_FIELDS["TitleAbstractClaimsDescription"]
+    in_paper = False
+    metadata = TaskMetadata(
+        name=__qualname__,
```
KennethEnevoldsen: "Dapfam_OUT_TitleAbstractClaims_TitleAbstractClaimsDescription" is too long of a title (general comment)

Generally we do not use underscores in task names

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,551 @@
...
+        description=(
+            "Out-of-domain retrieval: query and target share no IPC top-three codes. "
+            "Queries use title, abstract, and claims; corpus adds description. "
+            "Goal: evaluate complete-text cross-domain retrieval."
```
KennethEnevoldsen: From this description, I only know that it is an out-of-domain retrieval task. It is not clear to the user that this has anything to do with patents (unless you Google what IPC means, but that shouldn't be required)

Also do write out segments:

> "Goal: evaluate complete-text cross-domain retrieval."

It is better written out as:

> The goal of the task is to ...

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
KennethEnevoldsen: Can you refer to the section in the paper where this is discussed, then I will read through it

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
KennethEnevoldsen: It's already much better, but it still needs a bit more work. I gave a few pointers below

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
KennethEnevoldsen: > So besides the quantization and the cosine similarity ( that are now changeable through passing "similarity" and "quantize" arguments, there are some other differences that also warranted a custom eval function.

For `similarity`, can you change it such that it uses the model's similarity function?

> For mAP weighting, we build a deterministic rank-based weight vector [K, K–1, …, 1] for the top-K positions instead of using the similarity scores. This is motivated by the fact that cosine scores obtained after quantization have a higher likelihood of having equal values.

I am still missing a reason for the quantization. Can you help me understand why that is of interest during evaluation?

@Samoed can I ask you to look over this as well? I prefer not to have a custom evaluation, but only if it makes sense.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,551 @@
...
+    )
+
+
+class Dapfam_OUT_TitleAbstractClaims_TitleAbstractClaimsDescription(
+    _DAPFAMMixin, AbsTaskRetrieval
+):
+    domain_filter = "OUT"
+    query_fields = _QUERY_FIELDS["TitleAbstractClaims"]
+    corpus_fields = _CORPUS_FIELDS["TitleAbstractClaimsDescription"]
+    in_paper = False
```
KennethEnevoldsen:
```suggestion
):
     # In the paper
        domain_filter = "OUT"
    query_fields = _QUERY_FIELDS["TitleAbstractClaims"]
    corpus_fields = _CORPUS_FIELDS["TitleAbstractClaimsDescription"]
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,551 @@
...
+_SHARED_METADATA = dict(
+    dataset={"path": HF_REPO, "revision": "main"},
+    reference=REFERENCE,
+    type="Retrieval",
+    category="p2p",
+    task_subtypes=["Article retrieval", "Patent retrieval"],
+    eval_splits=["test"],
+    eval_langs=["eng-Latn"],
+    main_score="ndcg@10",
+    date=("1964-06-26", "2023-06-20"),  # dataset card coverage
```
KennethEnevoldsen:
```suggestion
    date=("1964-06-26", "2023-06-20"),
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,551 @@
+from __future__ import annotations
+
+import math
+from typing import Dict, List, Optional, Tuple
```
KennethEnevoldsen:
```suggestion
```
You don't need these when you import annotations from __future__. 

simply use dict, list, `str | None` (union with none) and tuple

KennethEnevoldsen (commented):
We are getting closer :)

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Yes of course, you can look at the chapter `3. Construction of DAPFAM`, there is even a quite large diagram/schema that summarizes the main steps taken to construct the dataset. I have a little free time this morning so i will get on the rest of the issues.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: I hope that the new changes are more in line with the naming conventions in mteb.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: I hope that i got it right this time.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Removed almost all abbreviations, and added more in depth explanation especially since the tasks are a little bit uncommon.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: The main reason i need the quantization is to faithfully reproduce the values shown in the paper.

In fact the whole reason i did quantize in the first place is that i was working on a quite limited hardware and i needed to do a lot of experiments (especially the passage level ones, not included in this batch of tasks), so quantization seemed like a good way to save on disk space and used computation power since i computed the embeddings, stored them using zstd compression, then computed the similarity scores later on.

i made a mistake in the commit message regarding the scores disparity that i tested.
if set to False\not set, the obtained NDCG, MAP and RECALL will be from ~0.004 to ~0.01 lower than the values shown in the paper (depending on the domain filter and the text chosen to represent the patent families).
right now it is `False` by default, so unless set explicitly by the user it will not be taken into account.

iliass-y:
Hi @KennethEnevoldsen ,
Any news ?

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
Samoed: I think it’s fine if the scores are slightly different due to quantization differences compared to the paper

KennethEnevoldsen (commented):
Sorry for the delay, I was at a conference last well (ACL).

Generally everything looks good except that the task is not imported, which means that `mteb.get_task` will not work.

iliass-y:
> Sorry for the delay, I was at a conference last well (ACL).
> 
> Generally everything looks good except that the task is not imported, which means that `mteb.get_task` will not work.

Ah, that's okay @KennethEnevoldsen .
Is there something i need to do now ?

KennethEnevoldsen:
You will need to add the import for the task and check that it can be fetched using `mteb.get_task`. Generally you can go through the checklist [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#submit-a-pr)

iliass-y:
> You will need to add the import for the task and check that it can be fetched using `mteb.get_task`. Generally you can go through the checklist [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_dataset.md#submit-a-pr)

All done @KennethEnevoldsen  !
tasks added and checks made.

KennethEnevoldsen:
All good to me - @Samoed anything to add here?

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,599 @@
...
+        qrels = self.relevant_docs[split]
+
+        corp_ids, corp_texts = zip(*corpus.items())
+        qry_ids, qry_texts = zip(*queries.items())
+
+        encode_kwargs = kwargs.get("encode_kwargs", {})
+        quantize = kwargs.get("quantize", False)
+
+        # check similarity function name :
+        print(model_wrapper.model.similarity_fn_name)
```
Samoed:
```suggestion
        logger.info(model_wrapper.model.similarity_fn_name)
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,599 @@
...
+  title        = {DAPFAM: A Domain-Aware Patent Retrieval Dataset Aggregated at the Family Level},
+  author       = {Ayaou, Iliass and Cavallucci, Denis and Chibane, Hicham},
+  year         = {2025},
+  eprint       = {2506.22141},
+  archivePrefix= {arXiv},
+  primaryClass = {cs.CL}
+}"""
+
+_SHARED_METADATA = dict(
+    dataset={"path": HF_REPO, "revision": "main"},
```
Samoed: Can you specify revision?

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,599 @@
...
+# MIX-IN with shared logic + metric implementation
+class _DAPFAMMixin:
+    # class-level attributes are filled in each concrete subclass
+    domain_filter: str | None = None
+    query_fields: list[str] = []
+    corpus_fields: list[str] = []
+    in_paper: bool = False
+
+    def load_data(self, **_) -> tuple[dict, dict, dict]:
+        ds_c = load_dataset(HF_REPO, "corpus", split="train")
```
Samoed: In `eval_splits` you've specifed only `test`. Also you should pass revision here

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,599 @@
...
+            qrels.setdefault(qid, {})[pid] = (
+                float(r["relevance_score"]),
+                r["domain_rel"],
+            )
+        # 4) Assign for MTEB
+        self.relevant_docs = {"test": qrels}
+        self.data_loaded = True
+        return self.corpus, self.queries, self.relevant_docs
+
+    def _dapfam_evaluate(
```
Samoed: I'm not sure if we want to support custom evaluator. @KennethEnevoldsen

Samoed (changes requested):
The main concern about this PR that it specifies custom evaluator. I've tried to look throught comments, but didn't found why it's reqired

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: A also went through the various comments and found the detailed explanation i gave as to why a custom evaluator ( would have been easier for me too if it wasn't the case ) :


... there are some other differences that also warranted a custom eval function.

For mAP weighting, w build a deterministic rank-based weight vector [K, K–1, …, 1] for the top-K positions instead of using the similarity scores, this is motivated by the fact that cosine scores obtained after quantizing have more chances of having equal values.

If a query has no positive ground truth documents in the whole dataset, we assign metric = 1.0.
on average, out-of-domain relevants documents per query is only 5.74 ( out of domain linked patent are quite rare especially in some IPC domains), so some queries have zero out-of-domain targets, and we had to decide how to aggregate their scores.
Setting those query metrics to 1.0 ensures we don’t penalize the system simply because no ground-truth existed in that split. And it makes sens from a specific system usage perspective for out of domain patent retrieval.

Both of these differences do not affect the deterministic nature of the metrics and allow for comparing various models correctly.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Done, revision number clearly stated.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Done, revision number clearly stated in all data loading operations.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
KennethEnevoldsen: Thanks for raising this @Samoed. @iliass-y I am not sure the quantization has been justified yet (which is required for your argument on the mAP).

> If a query has no positive ground truth documents in the whole dataset, we assign metric = 1.0.

Does it then provide any information for evaluation? Couldn't we simply delete those from the evaluation dataset

iliass-y: @KennethEnevoldsen ,

so for quantization, as i have said previously :

```
The main reason i need the quantization is to faithfully reproduce the values shown in the paper.

In fact the whole reason i did quantize in the first place is that i was working on a quite limited hardware and i needed to do a lot of experiments (especially the passage level ones, not included in this batch of tasks), so quantization seemed like a good way to save on disk space and used computation power since i computed the embeddings, stored them using zstd compression, then computed the similarity scores later on.

if set to False\not set, the obtained NDCG, MAP and RECALL will be from ~0.004 to ~0.01 lower than the values in the paper ...
```
but if it is such a roadblock i don't mind dropping it since the delta is still small enough.

As far as the ground truth assignment, if i delete them from the evaluation dataset it will have a big impact on the OUT evaluations. 
By design, it is very rare to have patent families referencing cross domain / out of domain patent families. And some domains have even less than the mean amount of out of domain links.
the idea is that since in real usage most patents don't have out of domain linked patents, it made no sens to penalize the system when no positives exist.
If i delete those, it will affect the IN and ALL evaluation somewhat, and it will tank the OUT evaluation.

There is no risk of triviality or non deterministic behavior in my implementation, a better system WILL systematically have better scores.
I also wish i didn't have to do this, implementing the custom function took me quite some time to align it with the MTEB code, and i am aware that this is the main reason why this PR is taking so long.

KennethEnevoldsen: Sorry for the late reply here @iliass-y - I was away last week and it took me a while to catch back up.

And do know that we appreciate the effort that making a PR takes, but we also have to ensure that we can keep this project going forward, so we have to make some calls to ensure that it is easy to keep it up to date.

decision on quantization: While I understand the initial reasoning for it, I don't think it is something that we wish to maintain (we simply don't have the resources to do that).

regarding the ground truth assignment I must say that I am not quite sure how it would influence the metrics. 

We have one dataset (TwitterhjerneRetrieval), which has 1 case of a query without a postive document. Removing that query does not influence the current performance metrics (which also fits my understanding of our metrics). So if that is the case it here, do help me locate where that happens?

KennethEnevoldsen: @iliass-y we are waiting on a response here

iliass-y: @KennethEnevoldsen Sorry,
I am working on my PHD defense and manuscript so i am spread thin.
I will soon get back at you, but i think the direction i am going to take is to remove custom evaluation logic completely, it's too much hassle, much simpler without.

KennethEnevoldsen: Perfect, feel free to ping us once you have an update (I get that working on the PhD takes priority)

iliass-y:
Hi @KennethEnevoldsen, @Samoed,

So i have finally gotten around to standardize the task and remove the custom evaluation (a made also a revision to the paper in the meantime to reflect the change in the metric computing logic among other things, feel free to check it out).
The tasks now do not use any custom evaluation functions.
The tasks metadata shows correctly as well :
image

I hope that we can finally finish up this PR this time.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,497 @@
...
+    eval_splits=["test"],
+    eval_langs=["eng-Latn"],
+    main_score="ndcg_at_100",
+    date=("1964-06-26", "2023-06-20"),
+    domains=["Engineering", "Chemistry", "Legal"],
+    license="cc-by-nc-sa-4.0",
+    annotations_creators="derived",
+    sample_creation="created",
+    bibtex_citation=BIBTEX,
+    judged_docs_only_flag=True,
```
Samoed: Can you remove this from metadata?
```suggestion
```

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,497 @@
...
+}
+
+
+# MIX-IN with shared logic + metric implementation
+class _DAPFAMMixin:
+    # class-level attributes are filled in each concrete subclass
+    domain_filter: str | None = None
+    query_fields: list[str] = []
+    corpus_fields: list[str] = []
+    in_paper: bool = False
```
Samoed: I can't find where this is used

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,497 @@
...
+        "claims_text",
+        "description_en",
+    ],
+}
+
+# Text representations used in the paper
+_IN_PAPER = {
+    ("TitleAbstract", "TitleAbstractClaims"),
+    ("TitleAbstractClaims", "TitleAbstractClaims"),
+}
```
Samoed: And this

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: That (the in_paper argument) was left for consistency reasons and since all the task classes inherit from this one.
the rest are present in the `load_data` method.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Done.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Done.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
Samoed: Yes, but I can't find usage of in_paper parameter in load_data

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,489 @@
...
+    query_fields: list[str] = []
+    corpus_fields: list[str] = []
+    in_paper: bool = False
+
+    def load_data(self, **_) -> tuple[dict, dict, dict]:
+        ds_c = load_dataset(
+            HF_REPO,
+            "corpus",
+            split="train",
+            revision="3ad6eab6ed9b5fb1c0609b4dbf40e391ebb5a544",
```
KennethEnevoldsen: should be taken from the metadata

KennethEnevoldsen (commented):
Thanks for taking the time on - I know that long reviews can be frustrating.

A few minor things, then I think you also need to format the code and the citations `make lint` and `make format-citations` should do the trick.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: I wanted to keep it just for consistence reasons but i have removed it now since it doesn't affect the code.

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
iliass-y: Done for all except for the "corpus" since i load 3 different datasets (corpus, queries, and relations) from my repo on the hub. If you have a better way please let me know.

iliass-y:
hi @KennethEnevoldsen and take you as well for taking the time to review this rather lengthy PR.
I hope that all the changes are aligned with your remarks.

Also, i wanted to inform you that from my experience, `make format-citations` does not format the citations to the acceptable format, but simply skips it :
image
However the test still failed, so i decided to make a custom script to see exactly were the formatting is erroneous, and i found out that the assertion expects an exact string match with the formatted version (from what i observed, even the order of the bibtex keys matter here). This is a little bit confusing since the formatted order does not match the default order obtained from arXiv.
For the task metadata, test kept pointing out that metadata was unfilled for the tasks, but no indication from my understanding as to which tasks (especially since the test passed for previous versions), maybe an indication of what metadata keys are missing during the checks would be helpful ?

again, thank you for your time you and @Samoed .

mteb/tasks/Retrieval/eng/DAPFAMPatentRetrieval.py
```diff
@@ -0,0 +1,496 @@
...
+    corpus_fields: list[str] = []
+
+    def load_data(self, **kwargs) -> tuple[dict, dict, dict]:
+        ds_c = load_dataset(
+            kwargs.get("dataset", {}).get("path", HF_REPO),
+            "corpus",
+            split=kwargs.get("eval_splits", "train"),
+            revision=kwargs.get("dataset", {}).get(
+                "revision", "780f4011d60297fc6e97a4119b0c516d13afea2d"
+            ),
```
Samoed: Can you change everywhere to this?
```suggestion
            split=self.metadata.eval_splits[0],
            revision=self.metadatadataset["revision"],
```

iliass-y: Done, i have also changed the dataset path argument to `self.metadata.dataset["path"]`.

Samoed:
>  For the task metadata, test kept pointing out that metadata was unfilled for the tasks, but no indication from my understanding as to which tasks (especially since the test passed for previous versions), maybe an indication of what metadata keys are missing during the checks would be helpful ?

I don't have errors locally too. Maybe new commits will fix it

KennethEnevoldsen (approved):
I feel like this it good to merge - @Samoed how about you?

Samoed (approved):
Yes, we can merge. @iliass-y Thank you for your work!

iliass-y:
Thanks to both of you for your work as well !

PR #2956, opened 2025-07-29
MAEB Model Evaluation Fixes

AdnanElAssadi56:
- Added 30-second truncation to all models.
- Added show_progress_bar kwarg for enabling/disabling progress display.

### DIFF
--- a/mteb/models/ast_model.py
+++ b/mteb/models/ast_model.py
@@ -100,7 +100,10 @@ class ASTWrapper(Wrapper):
         all_embeddings = []
 
         with torch.no_grad():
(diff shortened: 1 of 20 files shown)


AdnanElAssadi:
P.S. 
- I've tested some of the models after the changes, but it would probably be a good idea to run all of them. I tried to follow the original implementations as close as possible.
- If anyone notices missing models, feel free to mention, but I am pretty sure I've done them all.

KennethEnevoldsen:
@gowitheflow-1998 I think you are the best one to review this PR - do you have the time? We need to invalidate and rerun relevant models after this as well

Issue #2449, opened 2025-03-27
Move validation of metadata object to tests instead of import


KennethEnevoldsen:
Originally discussed in #1463. 

```
import mteb # building registry happens as import, but no validation
mteb.get_model_meta(name)

# during meta definition
meta = ModelMeta.model_construct(**kwargs) # This looks horrible we might be able to do it in another way

# During testing
meta.validate_model(meta)
```

A few potential solutions is to use: 
[datasclasses](https://docs.pydantic.dev/latest/concepts/dataclasses/#dataclass-config)
converting to [strict mode](https://docs.pydantic.dev/latest/concepts/strict_mode/#strict-mode-in-method-calls) (to prevent time spend on coercion). I am, however, not sure that it actually makes a difference.

KennethEnevoldsen:
As noted in #2958, this did not result in an import gain, so I will close this for now.

PR #2958, opened 2025-07-30
[v2] Remove validation on import

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/2449
I've change all `TaskMetadata` to `TaskMetadata.model construct`, but it seems it have no difference on import. I tried with this script
```python
from datetime import datetime

start = datetime.now()
import mteb
print(datetime.now() - start)
```
And on this branch and `v2` I have same 4 seconds for import


KennethEnevoldsen:
Sounds like there is no reason to do it then - let us just close this issue and this PR then

PR #2963, opened 2025-08-01
Editing init to add GovReport and BarExamQA tasks

abdurrahmanbutler:
I have edited the Retrieval init file to import the previously added GovReport and BarExamQA tasks so that they are visible to users.  This PR is to update the changes.


Samoed (approved):

KennethEnevoldsen:
Thanks @abdurrahmanbutler - it seems like the citation is incorrectly formatted. You can format it using:

```
python scripts/format_citations.py tasks
```

Since I was unfortunately unable to review the original PR, could I also request that you rename the functions to "GovReportRetrieval" and "BarExamQARetrieval"? This is to keep a consistent format for newly added tasks

abdurrahmanbutler:
Apologies for late reply. I have opened a new PR to fix these issues as well as the naming inconsistency of my other dataset BillSum.

New pr #3026

PR #2965, opened 2025-08-01
upload qzhou-embedding model meta

PennyYu123:

### DIFF
--- a/mteb/models/qzhou_embedding.py
+++ b/mteb/models/qzhou_embedding.py
@@ -1,0 +1,22 @@
+
+from mteb.model_meta import ModelMeta
+
+my_model = ModelMeta(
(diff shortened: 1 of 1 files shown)


mteb/models/qzhou_embedding.py
```diff
@@ -0,0 +1,22 @@
+
+from mteb.model_meta import ModelMeta
+
+my_model = ModelMeta(
+    name="QZhou-Embedding",
```
Samoed: Can you publish your model?

KennethEnevoldsen:
```suggestion
    name="Kingsoft-LLM/QZhou-Embedding",
```

PR #2966, opened 2025-08-01
[v2] split batched input by subtypes

Samoed:
I think we currently have too large a batched input, which doesn’t specify its input type. But I’m not sure how to properly handle this separation in the code


KennethEnevoldsen (approved):
Looks good to me

Issue #2728, opened 2025-05-26
Create interface with `search`


Samoed:
This would allow simplifying `Retrieval` and `Reranking` tasks. Start discussion https://github.com/embeddings-benchmark/mteb/pull/2710

This will likely consist of three steps:
- Refactoring CrossEncoders into its own DRESModel
- Refactor BM25 into its own DRES Models
- Define a consistent interface (should be derived from the two above)
- (Add Pylate PLAID index)

Samoed:
@orionw what do you think?

orionw:
Yeah I like this direction. We need some way of adding Pylate/Splade/other retrievers in a way that is easily extendable. 

I wrote out initial thoughts, then realized that we have most of the interface already: we seem to [have a `DenseRetrievalExactSearch` class with  `init` and `search`](https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/evaluation/evaluators/model_classes.py#L416C9-L416C29) functions. 

I think what we're missing is the indexing portion right? So if we abstract this setup a bit more (move encoding into a new `index` function that takes in the corpus and queries) before the `search` function then we can implement pretty much everything - BM25, Splade, PyLate models, etc.

Is that what you were thinking @Samoed?

Also I would really love to rename `DenseRetrievalExactSearch` since it is becoming so broad (#488) -- cross-encoders and late-interaction models are not dense retrieval exact search IMO. We could name it something generic like `SearchModel` or `Retriever` or something

Samoed:
```python
class SearchInterface:
    def __init__(
        self,
        model: Encoder | CrossEncoder,
        **kwargs: dict[str, Any],
    ):
        ...
        
    def search(
        self,
        corpus: dict[str, dict[str, str]],
        queries: dict[str, str],
        task_metadata: TaskMetadata,
        hf_split: str,
        hf_subset: str,
        instructions: dict[str, str] | None = None,
        **kwargs,
    ) -> dict[str, dict[str, float]]:
        ...
        
    # previously _rerank_documents
    def rerank(
        self,
        corpus: dict[str, dict[str, str]],
        queries: dict[str, str],
        task_metadata: TaskMetadata,
        hf_split: str,
        hf_subset: str,
        instructions: dict[str, str] | None = None,
        top_ranked: dict[str, list[str]] | None = None,
        **kwargs,
    ) -> dict[str, dict[str, float]]:
        ...  
```
[...]

PR #2970, opened 2025-08-02
[v2] Integrate search interface

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/2728

Added this interface instead of `DenseRetrievalExactSearch`
```python
class SearchInterface(Protocol):
    """Interface for searching models."""

    def index(
        self,
        corpus: CorpusDatasetType,
        *,
        task_metadata: TaskMetadata,
        hf_split: str,
        hf_subset: str,
    ) -> None:
        ...

    def search(
        self,
        queries: QueryDatasetType,
        *,
        task_metadata: TaskMetadata,
        hf_split: str,
        hf_subset: str,
        top_k: int,
        instructions: InstructionDatasetType | None = None,
        top_ranked: TopRankedDocumentsType | None = None,
    ) -> RetrievalOutputType:
     ...
```


mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -54,42 +51,31 @@ def __init__(
...
         **kwargs: Any,
-    ) -> dict[str, dict[str, float]]:
-        self.is_cross_encoder = is_cross_encoder_compatible(model)
-        if self.is_cross_encoder:
-            logger.info(
-                "The custom predict function of the model will be used if not a SentenceTransformer CrossEncoder"
+    ) -> RetrievalOutputType:
+        if not isinstance(model, SearchInterface):
+            raise TypeError(
+                f"RetrievalEvaluator expects a SearchInterface, got {type(model)}"
```
KennethEnevoldsen: We should just autowrap any encoder interfaces

mteb/models/abs_encoder.py
```diff
@@ -17,151 +18,21 @@
...
 logger = logging.getLogger(__name__)
 
 
-class AbsEncoder(ABC):
-    """Base class to indicate that this is a wrapper for a model.
-    Also contains some utility functions for wrappers for working with prompts and instructions.
-    """
-
-    model: Any
+class ModelMixin:
```
KennethEnevoldsen: I don't think a model mixin is required

mteb/models/abs_encoder.py
```diff
@@ -291,18 +164,95 @@ def get_task_instruction(
             return self.format_instruction(instruction)
         return instruction
 
-    def combine_query_and_instruction(
+
+class AbsEncoder(ABC, ModelMixin, AbsEncoderSearch):
```
KennethEnevoldsen: I actually don't think this is required - it competes 1-1 with the encoder interface so we should just use that

mteb/models/abs_search.py
```diff
@@ -0,0 +1,286 @@
...
+    RetrievalOutputType,
+    TopRankedDocumentsType,
+)
+
+from .encoder_interface import Encoder
+
+logger = logging.getLogger(__name__)
+
+
+class AbsEncoderSearch(Encoder):
```
KennethEnevoldsen: It is currently an interface/protocol, not an abstract class - I don't think it needs to inherit from the encoder interface

mteb/models/abs_search.py
```diff
@@ -0,0 +1,286 @@
...
+    corpus_chunk_size = 50_000
+    task_corpus: CorpusDatasetType | None = None
```
KennethEnevoldsen: This seems weird to me. Why is this needed? Task corpus is given in `index`. 

Is corpus_chunk_size not an encode_kwarg?

mteb/models/abs_search.py
KennethEnevoldsen: This seems more like a wrapper class around an encoder to me (essentially what DenseRetrievalExactSearch was before)

mteb/models/encoder_interface.py
```diff
@@ -118,6 +162,25 @@ def similarity_pairwise(
         """
         ...
 
+
+@runtime_checkable
+class CrossEncoderProtocol(Protocol):
```
KennethEnevoldsen, line 164: Is this only used for retrieval + reranking? Do we need it, or can we use the search interface? It might still be convenient to have even if we dont need it)

mteb/models/get_model_meta.py
```diff
@@ -76,7 +76,9 @@ def get_model_metas(
     return res
 
 
-def get_model(model_name: str, revision: str | None = None, **kwargs: Any) -> Encoder:
+def get_model(
+    model_name: str, revision: str | None = None, **kwargs: Any
+) -> Encoder | SearchInterface:
```
KennethEnevoldsen: Yeah so now not all models will be an encoder, which means that we probably need to add some error handling when these models are passed to `evaluate` or `MTEB`

mteb/models/model_implementations/rerankers_custom.py
```diff
@@ -19,7 +18,7 @@
 logger = logging.getLogger(__name__)
 
 
-class RerankerWrapper(DenseRetrievalExactSearch):
+class RerankerWrapper:
```
KennethEnevoldsen, line 21: This doesn't break anything?

mteb/models/model_meta.py
```diff
@@ -59,47 +59,53 @@ class ModelMeta(BaseModel):
...
+    loader: (
+        Callable[..., Encoder | SearchInterface]
+        | type[Encoder | SearchInterface]
+        | None
+    )
+    loader_kwargs: dict[str, Any] = field(default_factory=dict)
+    # metadata (required)
     name: str | None
     revision: str | None
+    # metadata (optional)
```
KennethEnevoldsen: I am afraid that these comments might be misleading (they are all required when making a PR.

if name and revision is required we should just remove None (though I know that we have a few cases where we use it)

mteb/models/model_meta.py
```diff
@@ -59,47 +59,53 @@ class ModelMeta(BaseModel):
...
+        is_cross_encoder: Whether the model can act as a cross-encoder or not.
+        modalities: A list of strings representing the modalities the model supports. Default is ["text"].
     """
 
     model_config = ConfigDict(extra="forbid")
 
+    # loaders
+    loader: (
+        Callable[..., Encoder | SearchInterface]
+        | type[Encoder | SearchInterface]
```
```suggestion
        Callable[..., Encoder | SearchInterface]
```
Right?

mteb/models/encoder_interface.py
```diff
@@ -8,10 +8,70 @@
...
+    InstructionDatasetType,
     PromptType,
+    QueryDatasetType,
+    RetrievalOutputType,
+    TopRankedDocumentsType,
 )
 
 
+@runtime_checkable
+class SearchInterface(Protocol):
```
KennethEnevoldsen: we probably should put this in a seperate file or rename the file to model_interfaces.py

mteb/models/search/search_encoder.py
KennethEnevoldsen, line 1: are these two files needed? Seems like it is just a restatement of the search interface

mteb/evaluation/evaluators/RetrievalEvaluator.py
```diff
@@ -99,6 +85,7 @@ def evaluate(
         k_values: list[int],
         ignore_identical_ids: bool = False,
     ) -> tuple[
+        # todo add NamedTuple
```
KennethEnevoldsen: should we create this?

KennethEnevoldsen (commented):
Great starting point! I think we can simplify it a bit - would especially like if we can avoid having multiple sources of how the interfaces should look

mteb/models/abs_encoder.py
Samoed: I think we can create `AbsEncoder` and `AbsCrossEncoder` and inherit from this class to reuse some functions for instructions

mteb/evaluation/evaluators/RetrievalEvaluator.py
Samoed: But then how models would implement their custom implementations?

mteb/models/abs_search.py
Samoed: > This seems more like a wrapper class around an encoder to me

Initially, I thought that way, but I think you meant to add a function to the model class https://github.com/embeddings-benchmark/mteb/issues/2728#issuecomment-3145539096

> It is currently an interface/protocol, not an abstract class - I don't think it needs to inherit from the encoder interface

Yes, I inhereted from `Encoder` to show that we reuse `encode` function from it

mteb/evaluation/evaluators/RetrievalEvaluator.py
Samoed: Yes, will create later

mteb/models/abs_encoder.py
Samoed: > I actually don't think this is required

What doesn't required?  Mixin or AbsEncoderSearch?

mteb/models/abs_encoder.py
KennethEnevoldsen: Why not simply re-use a function?

mteb/models/abs_encoder.py
KennethEnevoldsen: AbsEncoder as a whole (but that might be a another PR)

mteb/models/abs_search.py
KennethEnevoldsen: I am not sure what you mean here (is it the right comment you are referring to?)

mteb/models/abs_search.py
Samoed: > Is corpus_chunk_size not an encode_kwarg?

Corpus_chunk_size is used to not store all corpus emeddings at once to minimaize memory usage for huge datasetse, e.g. `Miracl`. Corpus will be process by chunks with this size and will be compared with queries. I can remove this and this will be more simplier to use.

> Why is this needed? Task corpus is given in index.

Yes, just to add correct type hint for it

mteb/models/encoder_interface.py
Samoed: > Is this only used for retrieval + reranking? 

I think reranking only.

> Do we need it, or can we use the search interface?

I thought about creating loader for crossencoders, becuase for now they're inherit from `DRES` directly, which is strange https://github.com/embeddings-benchmark/mteb/blob/e4f30e9bbc0beec516ef8a98d93076ce751124f3/mteb/models/rerankers_custom.py#L20

Also if user will pass cross encoder to mteb directly it would use same wrapper as encoder, that won't be correct

mteb/models/model_implementations/rerankers_custom.py
Samoed: Yes, at the moment this will not work. I will update it later. I have just deleted `DenseRetrievalExactSearch` and will need to update it like this to run the tests.

mteb/models/abs_search.py
KennethEnevoldsen: Yea so this is where the abstract and interfaces have issues.

What I want is:
1) An interface (the goal standard reference) 
2) instances that implement the interface

If multiple instances might justify an abstract class, but we don't have that at the moment. Even for the encoder, almost everything uses the sentence-transformer wrapper

Generic code (like similarity) can also be inherited from the interface.

mteb/models/model_meta.py
Samoed: > if name and revision is required we should just remove None (though I know that we have a few cases where we use it)

I then remove it from this PR and update in different PR

mteb/models/model_meta.py
Samoed: We have some implementations that passing class to the loader, so we need to annotate with `type` here https://github.com/embeddings-benchmark/mteb/blob/7c60af51b838be88bb2126ebdc8307dd0f313ae8/mteb/models/model_implementations/align_models.py#L107

mteb/models/search/search_encoder.py
Samoed: Yes, would remove for now

mteb/models/abs_encoder.py
Samoed: They're using prompts from class direclty, but I can move out them from class

mteb/models/abs_encoder.py
Samoed: It have functions for working with similarity and prompts

mteb/models/abs_search.py
Samoed: > Def. implement it as two different interfaces - but there is not reason why one class can't adhere to two interfaces

I thought this mean to add `search` and `index` to the models classes directly instead of creating wrapper for them 

Also you've said 
> the search interface does not need an init

And I thought you meant that any model would be used as a `SearchInterface`, but with two additional functions rather than a wrapper

mteb/models/abs_search.py
Samoed: This is not like abastract class, but more default search class for the encoders. (Btw we might want to rename it like `Embedder`)

mteb/evaluation/evaluators/RetrievalEvaluator.py
Samoed: What should be `Search`? Some additional functions or wrapper?

mteb/models/model_meta.py
KennethEnevoldsen: Shouldn't be a problem - it is also callable and returns an compatible instance:

```
>>> class TestClass():
...     def __init__(self) -> None:
...         pass
... 
>>> from typing import Callable
>>> isinstance(TestClass, Callable)
True

File: mteb/evaluation/evaluators/RetrievalEvaluator.py (hunk 1 above)
Comment (Samoed) [reply mteb/evaluation/evaluators/RetrievalEvaluator.py]: @KennethEnevoldsen what we should do wrapper or not?

File: mteb/models/abs_search.py (hunk 5 above)
Comment (KennethEnevoldsen) [reply mteb/models/abs_search.py]: Hmm, right then, what about `EncoderWrapperForSearch`?

File: mteb/evaluation/evaluators/RetrievalEvaluator.py (hunk 1 above)
Comment (KennethEnevoldsen) [reply mteb/evaluation/evaluators/RetrievalEvaluator.py]: Can't we simply do:

```
if not isinstance(model, SearchInterface) and isinstance(model, Encoder):
  logger.info("The supplied model does not implement the search interface. Attempting to wrap the model using EncoderWrapperForSearch")
  search_model = EncoderWrapperForSearch(model)
else:
  raise TypeError("The supplied model implements neither the SearchInterface nor Encoder")
```

We should probably also align naming here: If we use `Encoder` (and don't specify the interface), then we should probably rename to exclude the interface part.

**Comment by KennethEnevoldsen** [comment]
Will wait 'till we feedback from the others for the remaining comments

File: mteb/evaluation/evaluators/RetrievalEvaluator.py (hunk 1 above)
Comment (Samoed) [reply mteb/evaluation/evaluators/RetrievalEvaluator.py]: Ah, I see what you've meant. That's a good solution

Comment (Samoed) [reply mteb/evaluation/evaluators/RetrievalEvaluator.py]: Updated as you've suggested

File: mteb/models/abs_encoder.py (hunk 2 above)
Comment (Samoed) [reply mteb/models/abs_encoder.py]: Rollback these changes

File: mteb/models/encoder_interface.py (hunk 11 above)
Comment (Samoed) [reply mteb/models/encoder_interface.py]: Renamed

File: mteb/models/get_model_meta.py (hunk 7 above)
Comment (Samoed) [reply mteb/models/get_model_meta.py]: Added additional checks

File: mteb/models/model_implementations/rerankers_custom.py (hunk 8 above)
Comment (Samoed) [reply mteb/models/model_implementations/rerankers_custom.py:21]: Added wrapper for `CrossEncoderProtocols`. These models are compatible with it, and they will be used with `SearchCrossEncoderWrapper`

**Comment by Samoed** [comment]
I've implemented a search wrapper for the cross-encoder to support only reranking tasks. I think the `v2` implementation of cross-encoder search is a bit incorrect. In this part, we're selecting all corpus IDs, and I believe it always selects the same IDs every time:  
https://github.com/embeddings-benchmark/mteb/blob/64478e746f889fbd05dd226f456755be2bac9caa/mteb/evaluation/evaluators/dense_retrieval_exact_search.py#L460-L468

I think we can change to encode all pairs from qrels instead. 

For now, I think we don't want to support two-stage encoding. We can add it later.

**Comment by Samoed** [comment]
Also I'm not sure if we need `top_ranked` split if it have same logic and values as in `qrels` (relevant_document) split? It's used only to get corpus ids of query, but I think this can be done with `relevant_document` too. https://github.com/embeddings-benchmark/mteb/blob/64478e746f889fbd05dd226f456755be2bac9caa/mteb/evaluation/evaluators/dense_retrieval_exact_search.py#L279-L283

I've look to the some datasets and they have same information in `qrels` and `top_ranked`



CC @orionw

**Comment by orionw** [comment]
> Also I'm not sure if we need top_ranked split if it have same logic and values as in qrels (relevant_document) split? It's used only to get corpus ids of query, but I think this can be done with relevant_document too.

> For now, I think we don't want to support two-stage encoding. We can add it later.

I think these are related. `top_ranked` is simulating the 2nd stage pass where your first stage has found some documents but you aren't sure if they're relevant or not. If you just used the `qrels` than for some datasets then you'd be limited to reranking documents that have been judged (which in some cases is just relevant docs). 

My guess is there are some datasets in MTEB for reranking that have nearly identical qrels and top_ranked, but that they are other datasets where that also doesn't happen (and/or future datasets/flexibility).

> I've implemented a search wrapper for the cross-encoder to support only reranking tasks. I think the v2 implementation of cross-encoder search is a bit incorrect. In this part, we're selecting all corpus IDs, and I believe it always selects the same IDs every time:

We do want to only pick the ids they give for reranking, but it should differ per query. I think `top_ranked` is a mapping for this reason, to select the ids to use for each query.

**Comment by Samoed** [comment]
> My guess is there are some datasets in MTEB for reranking that have nearly identical qrels and top_ranked, but that they are other datasets where that also doesn't happen (and/or future datasets/flexibility).

For now I think they're always the same, because they were reuploaded by you.

> top_ranked is simulating the 2nd stage pass where your first stage has found some documents but you aren't sure if they're relevant or not. 

But this can be achived by using qrels too

> We do want to only pick the ids they give for reranking, but it should differ per query. I think top_ranked is a mapping for this reason, to select the ids to use for each query.

I think same information is stored in qrels

File: mteb/models/search_wrappers.py [hunk 13]
```diff
@@ -0,0 +1,344 @@
...
+        hf_split: str,
+        hf_subset: str,
+        top_k: int,
+        encode_kwargs: dict[str, Any],
+        top_ranked: TopRankedDocumentsType | None = None,
+    ) -> RetrievalOutputType:
+        if top_ranked is None:
+            raise ValueError(
+                "CrossEncoder search requires top_ranked documents for reranking."
+            )
```
Samoed, line 305: I've implemented cross-encoders to support reranking task only, but we have test:
1. that testing cross-encoders on retrieval
2. 2 stage reranking
Do we want to support these cases?

Samoed:
@KennethEnevoldsen What do you thing about `relevant_docs`/`top_ranked`?

mteb/models/search_wrappers.py
orionw: +1 to what @KennethEnevoldsen thinks, but the 2nd stage reranking is nearly identical to just normal reranking and almost comes for free. I would vote to keep it if there's not much added complexity (unless things have changed and now there's more complexity)

orionw:
> But this can be achived by using qrels too

Not always. Maybe the easiest example is that you don't need to include non-relevant documents in qrels. So if I were to upload a new reranking dataset with only positive qrels then we'd only ever get 100% (since we're only reranking the positives with no negatives).

I think the old format used to **always** specify that they were either non-relevant or relevant, and so when I uploaded them those went to the qrels even though they wouldn't usually (but for equal scores with the old format they had to be). 

But you don't *have* to specify the non-relevant documents for most retrieval/reranking datasets so forcing it to be that way would limit future datasets.

KennethEnevoldsen:
> For now, I think we don't want to support two-stage encoding. We can add it later.

We should at least have it in v2. However, we might limit the scope of this PR and just create an issue on it. 

`mteb.evaluate` currently doesn't support it, though, but it should be possible to solve this using a wrapper: 

I am thinking something like this:
```
retrieval_task = ... 
model_stage1 = ...
model_stage2 = ...

model_stage1 = SavePredictionsWrapper(model_stage1, path=...)
res = mteb.evaluate(model_stage1, retrieval_task)

model_stage2 = LoadPredictions(model_stage2, path=...)
res = mteb.evaluate(model_stage1, retrieval_task)
```

The hope is that this would maintain a fairly simple evaluation loop and move the two-stage into a different component. WDYT?

Note: This does not support saving the errors (which is also not supported in `mteb.evaluate`, but it also currently not documented anywhere.

**regarding `qrels` and `top_ranked`**

I am not entirely sure what the right move is, but I would probably trust @orionw's intuition more than my own here. In which case I would keep them. 

However, if there are no cases where they are different, then I am not sure it is needed. Is it used to compute certain metrics?

Samoed:
> However, if there are no cases where they are different, then I am not sure it is needed. Is it used to compute certain metrics?

It was only used to take pairs of query id, doc id for reranking tasks for the encoders. 

Also I'm not sure if cross-encoders should support retrieval tasks or not. If we make them to use qrels, then it can be used with both reranking and and retrieval. If only top ranked then retrieval task only (as implemented now)

> The hope is that this would maintain a fairly simple evaluation loop and move the two-stage into a different component. WDYT?

I'm not sure about wrappers for the models, because this problem is more related to task, than to the models and results of previous run should be passed to the evaluator somehow. 

Also we might want to find a solution how to save results of these run, e.g. to show them on leaderboard

KennethEnevoldsen:
> Also I'm not sure if cross-encoders should support retrieval tasks or not. If we make them to use qrels, then it can be used with both reranking and and retrieval. If only top ranked then retrieval task only (as implemented now)

I think they should. We want people to evaluate realistic use-cases. I would argue that it is more realistic than the reranking case.

(@orionw)

> I'm not sure about wrappers for the models, because this problem is more related to task, than to the models and results of previous run should be passed to the evaluator somehow.

Hmm I don't think it is a concern for the evaluator. The model can respond with the same documents as the previous model + its own edits. No reason for the evaluator to know what the previous model predicted.

Samoed:
> I think they should

So cross-encoders would use only qrels or we can make them to use qrels by default and too_ranked if they passed.

> Hmm I don't think it is a concern for the evaluator.

But in that case evaluator will pass pairs of quries and documents (as standart retrieval) to the model and will expect their scores and I don't understand how model will use top ranked then.

orionw:
> However, if there are no cases where they are different, then I am not sure it is needed. Is it used to compute certain metrics?

There are cases where they are different - at least all of the `InstructionReranking` tasks

The somewhat long background is that for MTEB v1, the reranking tasks used to use custom metrics and contain these fields:
```
        query: str
        positive: list[str]
        negative: list[str]
```

In #1359 we consolidated retrieval/reranking so they could both use the TREC evaluator and now they share much of the same logic.

So in order to maintain the same scores as MTEB v1 reranking, we had to add all the negatives to the negative qrels. But that is not how reranking is normally done, so it's a bit of a hack. More recent and future datasets won't do it this way.

Any reranking task formulation has to have both `qrels` and `top_ranked` since they are generally quite disinct. Imagine I wanted to create a new NQ reranking task (NQ has two positives per query): NQ doesn't have qrels for documents that are non-relevant since humans only validated the positives. If I didn't upload a `top_ranked`, it would only rerank from the `qrels`, which would only rerank the two relevant documents per query. This would always achieve 100% on all metrics. The `top_ranked` field instead tells the model what to rerank and the `qrels` is what has been judged to be relevant or not.

> Also I'm not sure if cross-encoders should support retrieval tasks or not. If we make them to use qrels, then it can be used with both reranking and and retrieval. If only top ranked then retrieval task only (as implemented now)

Sorry, I don't think I fully follow this @Samoed. Cross-Encoders shouldn't generally do retrieval without `top_ranked` since they would have to rerank every combination of query and doc (very expensive), although it is technically feasible. But Embedding models could do either one and should already by able to handle both. 

> Also we might want to find a solution how to save results of these run, e.g. to show them on leaderboard

I do think we will have to come up with a solution for leaderboard style evaluation, but I think the wrapper and current solution works great for practical evaluation/use. 

I think the wrapper style would be quite easy to use @KennethEnevoldsen!

Samoed:
> Cross-Encoders shouldn't generally do retrieval without top_ranked since they would have to rerank every combination of query and doc (very expensive), although it is technically feasible.

I agree with you, but in our tests, one of the failures in this branch is that cross-encoders are being tested on retrieval tasks (other cases are two-stage reranking, but still on retrieval) and in main branch we have handling of cross-encoders in retrieval https://github.com/embeddings-benchmark/mteb/blob/4adf56523ca64275150771aeb7b84b14e92b1a9d/mteb/evaluation/evaluators/RetrievalEvaluator.py#L245, and that’s why I raised this question.

> I do think we will have to come up with a solution for leaderboard style evaluation, but I think the wrapper and current solution works great for practical evaluation/use.

I think we could create a wrapper for tasks to store information about `top_ranked` so it can be reused in the evaluator instead of in the model.

orionw:
> I agree with you, but in our tests, one of the failures in this branch is that cross-encoders are being tested on retrieval tasks (other cases are two-stage reranking, but still on retrieval), and that’s why I raised this question.

Ah I see. Yeah I think we could drop that test and honestly we should print a warning to the user if they are doing a retrieval task with a cross-encoder. It's possible they mean to (and we could allow it for people who really want to or are using small datasets), but more likely they are making a mistake. 

> I think we could create a wrapper for tasks to store information about top_ranked so it can be reused in the evaluator instead of in the model.

Yeah this is a nice approach! It would be basically like creating a mini-reranking task that then is reproducible. The only issue I could see is that any leaderboard would have to use the same `top_ranked` but it would solve reproducibility issues which is big.

Samoed:
So, @orionw can you review my implementation then? I will work on multistage on another pr

orionw:
Yes, sorry, will review. Is it this one?

Samoed:
Yes, it is

mteb/evaluate.py
```diff
@@ -256,9 +263,12 @@ def evaluate(
...
     model_revision = cast(str, meta.revision)
-    if isinstance(model, (SentenceTransformer, CrossEncoder)):
-        model = SentenceTransformerWrapper(model)  # type: ignore[assignment] # TODO: SentenceTransformerWrapper should be a subclass of Encoder
-    model = cast(Encoder, model)
+    if isinstance(model, SentenceTransformer):
+        model = SentenceTransformerEncoderWrapper(model)
+        model = cast(Encoder, model)
+    elif isinstance(model, CrossEncoder):
+        model = CrossEncoderWrapper(model)
+        model = cast(CrossEncoderProtocol, model)
```
orionw, line 271: I'm not familiar with this `cast` ,  what does that do?

mteb/evaluation/MTEB.py
```diff
@@ -250,7 +253,7 @@ def _merge_split_scores(
 
     def run(
         self,
-        model: SentenceTransformer | Encoder,
+        model: MtebSupportedModelProtocols | CrossEncoder | SentenceTransformer,
```
orionw: Would it be easier to create a meta-list at this point? Or does this cover all the protocols we want to expand to someday?

mteb/models/models_protocols.py
```diff
@@ -1,17 +1,74 @@
...
+    TopRankedDocumentsType,
 )
 
 
+@runtime_checkable
+class SearchProtocol(Protocol):
+    """Interface for searching models."""
+
+    def index(
+        self,
```
orionw, line 24: Love this!

mteb/models/models_protocols.py
```diff
@@ -132,8 +192,8 @@ def predict(
         """Predicts relevance scores for pairs of inputs. Note that, unlike the encoder, the cross-encoder can compare across inputs.
 
         Args:
-            inputs1: First Dataloader of inputs to encode.
-            inputs2: Second Dataloader of inputs to encode.
+            inputs1: First Dataloader of inputs to encode. For reranking will be queries.
```
orionw: Maybe add it may contain instructions in the docstring?

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -123,18 +120,20 @@ def encode(
             embeddings = embeddings.cpu().detach().float()
         return embeddings
 
-    def _predict(
+
+class CrossEncoderWrapper:
```
orionw, line 124: I will say I am a bit confused. We have a `CrossEncoderWrapper` and a `CrossEncoderProtocol`.  The protocol is just the abstract methods being used?

mteb/evaluation/MTEB.py
Samoed: `MtebSupportedModelProtocols` is meta-list with all suported protocols. I've added here `CrossEncoder | SentenceTransformer` to allow to pass `SentenceTranformer` models wihout additional work

mteb/models/sentence_transformer_wrapper.py
Samoed: > The protocol is just the abstract methods being used?

Yes,`CrossEncoderProtocol` like specification and `CrossEncoderWrapper` is implementation

mteb/evaluate.py
Samoed: This is for typechecking that model is type of `CrossEncoderProtocol`, but I think we can remove it

Samoed:
@KennethEnevoldsen can you review PR and give your thoughts about 2 stage reranking?

mteb/models/models_protocols.py
```diff
@@ -145,3 +205,6 @@ def predict(
             The predicted relevance scores for each inputs pair.
         """
         ...
+
+
+MtebSupportedModelProtocols = Union[Encoder, CrossEncoderProtocol, SearchProtocol]
```
KennethEnevoldsen:
```suggestion
MtebSupportedModelProtocols = Union[Encoder, CrossEncoderProtocol, SearchProtocol]
```
Generally very happy with this, but I would rephrase these to:

```
Models = Union[EncoderModel, CrossEncoderModel, SearchModel]
# or MTEBModels
```
However, this is something we can take in another PR to get this one merged

mteb/models/sentence_transformer_wrapper.py
KennethEnevoldsen: Specifically, it is a wrapper that wraps the sentence transformer CrossEncoder to follow the CrosEncoderProtocol (other languages use 'interface' as the term)

mteb/models/models_protocols.py
KennethEnevoldsen: Feel free to move this into an issue

KennethEnevoldsen:
Yeah, generally think that this looks really good - very happy about where we are heading

Let us do the two-stage in another PR. 

I am more than happy with disabling the test here and merging

PR #2973, opened 2025-08-02
qzhou-embedding model_meta & implementation

PennyYu123:
I have corrected the wrong info in model_meta and added the model implementation code. 
Our QZhou-Embedding model is currently private on huggingface and will be released tomorrow. Please check it out then. Thank you. 

### DIFF
--- a/mteb/models/qzhou_embedding.py
+++ b/mteb/models/qzhou_embedding.py
@@ -1,0 +1,282 @@
+import mteb
+from mteb.model_meta import ModelMeta
+from mteb.models.wrapper import Wrapper
+from mteb.encoder_interface import PromptType
+from functools import partial
+from typing import Dict, List, Optional, Union
+from tqdm.autonotebook import trange
+
+import numpy as np
+import torch
+from torch import Tensor, nn
+import torch.nn.functional as F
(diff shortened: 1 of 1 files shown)


mteb/models/qzhou_embedding.py
```diff
@@ -0,0 +1,282 @@
...
+        'CovidRetrieval': 'Given a question on COVID-19, retrieve news articles that answer the question',
+        'DuRetrieval': 'Given a Chinese search query, retrieve web passages that answer the question',
+        'MMarcoReranking': 'Given a Chinese search query, retrieve web passages that answer the question',
+        'MMarcoRetrieval': 'Given a web search query, retrieve relevant passages that answer the query',
+        'T2Reranking': 'Given a Chinese search query, retrieve web passages that answer the question',
+        'ThuNewsClusteringP2P': 'Identify the topic or theme of the given news articles based on the titles and contents',
+        'VideoRetrieval': 'Given a video search query, retrieve the titles of relevant videos'
+        }
+
+class QZhouModel(nn.Module):
```
Samoed, line 104: I think it would be better if you would integrate this to your model implementation directly instead of only to mteb

Issue #1337, opened 2024-10-27
MIEB: Figure out folder structure


KennethEnevoldsen:
Figure out what to do with folder structure. Should we create a folder like evaluation/evaluators/{modality}, or do we want to do it in another way? I also have some concerns re. evaluators (as I would like them to be the same across text and images, but I see that it might not be possible)

isaac-chung:
Some can certainly be shared without modifications, like the `Any2AnyRetrievalEvaluator`. The rest seem to need some modifications, i.e. unify the modalities from which embeddings are generated. e.g. STSEvaluator and VisualSTSEvaluator share pretty much everything below the line `logger.info("Evaluating...")`.

isaac-chung:
We still have the efforts like creating the MIEB leaderboard (whether separate or aligning the results + adding benchmarks such that it can used in leaderboard v2+). For simplicity, I propose that we:

1. Merge MIEB into main with the current structure, i.e. `evaluation/evaluators/{modality}`
2. Lay out what modifications are needed to reuse evaluators across modalities or create modality-agnostic evaluators / abstasks.
3. Consider how much effort we'd like to spend in extensibility in anticipation of MOEB modalities, e.g. audio.
4. In new issues/PRs, add the new modality-agnostic evaluators / abstasks.
5. When applicable, switch tasks over to use the new modality-agnostic evaluators / abstasks.
6. Mark the older modality-specific evaluators / abstasks as deprecated.

This way, we move this issue to non-blocking, and we can directly start the final merge of mieb into main.

KennethEnevoldsen:
I agree with all of these points. Backward compatibility should be a thing we worry about a lot more after v2 and this seems to align with what you are thinking as well.

wouldn't we want to move text tasks into a "text" folder though?

isaac-chung:
> wouldn't we want to move text tasks into a "text" folder though?

I was hoping the status quo could remain as is, seeing that MTEB was originally set up for `text` only. If we really wanted to do that, maybe v2 would be a good place to do that? I'd say we could leave it out of the MIEB merge to simplify things.

x-tabdeveloping:
I don't think the number of tasks would be a problem, since we can separate benchmarks by modality.

Samoed:
I think we can do:
```
- tasks_classes
  - any_2_any
  - image
  - text
  - AbsTaskClassification
  - AbsTaskSTS
  - ...
 ```
And similar to evaluators and tasks.

[...]

PR #2882, opened 2025-07-06
Implemented Audio Any2AnyRetrieval + 3 Datasets for A2A, A2T, T2A

kkaitlyn111:
Resolves #2068

- use the Any2AnyRetrieval abstask and evaluator from image (updated) cc @Samoed , @KennethEnevoldsen
- converted audiocaps to the normal retrieval format (included a script)
- added mock AbsTaskAny2AnyRetrieval classes (T2A, A2A, A2T) to mock tasks

----
Old description:

Implemented the Any2Any Retrieval pipeline for audio (similar to Any2Any for image and text). I also added 3 audio2audio (JamAltArtist), audio2text (AudioCaps), and text2audio (JamAltLyrics) datasets to test compatibility with the new formats

Note: Closed #2684 and also fixed up suggestions from reviewers from that thread!
(Also, we can probably remove the AudioCapsReversed toy task, just used it here to test that we got similar results both ways)

PR #2980, opened 2025-08-03
fix: restructure evaluators

KennethEnevoldsen:
- [x] restructure evaluators into folders to match supported modalities
- [x] rename evaluators to reflect supported modalities. I would probably drop `Any` and instead rename old ones to `Text*`
- [x] make evaluators a private module

I will restructure the tasks themselves in another PR.

related to #1337


mteb/evaluation/evaluators/any_sts_evaluator.py
```diff
@@ -23,7 +23,7 @@
 logger = logging.getLogger(__name__)
 
 
-class AnySTSEvaluator(Evaluator):
+class AnySTSEvaluator(Evaluator):  # TODO: Should we rename this to just STSEvaluator?
```
Samoed: No, it's used in visual sts https://github.com/embeddings-benchmark/mteb/blob/64478e746f889fbd05dd226f456755be2bac9caa/mteb/tasks/Image/VisualSTS/eng/STS12VisualSTS.py#L7

mteb/evaluation/evaluators/image/multi_choice_evaluator.py
```diff
@@ -35,7 +35,7 @@
 
 
 # Adapted from https://github.com/beir-cellar/beir/blob/f062f038c4bfd19a8ca942a9910b1e0d218759d4/beir/retrieval/search/dense/exact_search.py#L12
-class Any2AnyMultiChoiceSearch:
+class Any2AnyMultiChoiceSearch:  # TODO: Is this actually any2any? -> Seems like it should be ImageMultiChoiceSearch
```
Samoed: I think in maeb this evaluator used too, but I think we would refactor this after search interface

mteb/evaluation/evaluators/image/retrievalevaluator.py
```diff
@@ -173,7 +173,9 @@ def load_results_file(self):
 
 
 # Adapted from https://github.com/beir-cellar/beir/blob/f062f038c4bfd19a8ca942a9910b1e0d218759d4/beir/retrieval/evaluation.py#L9
-class Any2AnyRetrievalEvaluator(Evaluator):
+class Any2AnyRetrievalEvaluator(
+    Evaluator
```
Samoed: Same as MultiChoice

mteb/evaluation/evaluators/summarization_evaluator.py
```diff
@@ -25,7 +25,9 @@
 logger = logging.getLogger(__name__)
 
 
-class SummarizationEvaluator(Evaluator):
+class SummarizationEvaluator(
+    Evaluator
+):  # TODO: Should we rename this to SummarizationTextEvaluator?
```
Samoed, line 30: Maybe we should name them in format `{modality}{task type}{task/evaluator}`

mteb/evaluation/evaluators/any_sts_evaluator.py
KennethEnevoldsen: Right, but is it also used for text - what I am proposing is removing "Any" (in general only add modalities if the task does not allow all modalities)

mteb/evaluation/evaluators/image/multi_choice_evaluator.py
KennethEnevoldsen: Seems like it only work for image-text though?

mteb/evaluation/evaluators/summarization_evaluator.py
KennethEnevoldsen: That works - will rename

mteb/evaluation/evaluators/image/multi_choice_evaluator.py
Samoed: In this branch yes, but in maeb there is small additional for creating embeddings, but rest is the same https://github.com/embeddings-benchmark/mteb/pull/2882

KennethEnevoldsen: renamed to `Image2TextMultiChoiceSearch`

mteb/evaluation/evaluators/image/retrievalevaluator.py
KennethEnevoldsen: changed this to `ImageTextRetrievalEvaluator`

mteb/evaluation/evaluators/any_sts_evaluator.py
KennethEnevoldsen: Ahh but we already have `STSEvaluator`, but these are not the same? Can both be used for images and text?

Samoed: I think it was added during merge and shouldn't be used in `v2`

Samoed: Hm. I can't find `STSEvaluator` in `v2`

KennethEnevoldsen: Hmm there are two is this branch - but I see that the code is the same so I suspect is must be a merge thing

KennethEnevoldsen:
Will close and open another PR for it

PR #2981, opened 2025-08-03
Refactor CLI to enable changes

KennethEnevoldsen:
- [x] move cli, create_meta into module
- [x] rename create_meta > generate_readme
- [x] rename cli.py > ~~build_cli~~ main
  - refactored build_cli() out of main()
  - ~~move the main function into `__main__`~~
- [x] moved docstring into documentation
- [x] made function with add parsers private


mteb/cli/build_cli.py
```diff
@@ -92,7 +8,7 @@
 import torch
 
 import mteb
```
Samoed, line 10: I think we can move these imports inside functions to make `mteb --help` generated instantly

mteb/cli/run.py
Samoed, line 1: In `build_cli` run function exists too

mteb/cli/run.py
```diff
@@ -0,0 +1,164 @@
+"""This is the command line interface for `mteb`.
```
Samoed: Do we need this docstring since you're updated `cli.md`?

mteb/cli/run.py
```diff
@@ -0,0 +1,164 @@
...
+        tasks = mteb.get_benchmarks(names=args.benchmarks)
+    else:
+        tasks = mteb.get_tasks(
+            categories=args.categories,
+            task_types=args.task_types,
+            languages=args.languages,
+            tasks=args.tasks,
+        )
+
+    eval = mteb.MTEB(tasks=tasks)
```
Samoed: Also, should we change this to new run format, or you will change this in another PR?

mteb/cli/build_cli.py
KennethEnevoldsen: That would have been great, but it seems like it imports `__init__.py` :/

mteb/cli/run.py
KennethEnevoldsen: yes - i have deleted this

mteb/cli/run.py
KennethEnevoldsen: another PR (evaluate is not fully ready to replace MTEB - we will need the search interface for that to work)

mteb/cli/main.py
```diff
@@ -362,20 +278,27 @@ def add_create_meta_parser(subparsers) -> None:
...
+    _add_available_tasks_parser(subparsers)
+    _add_available_benchmarks_parser(subparsers)
+    _add_create_meta_parser(subparsers)
 
+    return parser
 
-if __name__ == "__main__":
-    main()
+
+def main() -> None:
```
Samoed: Maybe it would be better to rename function or file, because imports would be confusing `from main import main`

Issue #2969, opened 2025-08-02
Merge `instruction` with `queries` for retrieval


Samoed:
> Can we add instructions to queries as an additional column? I think would simplify implementation of reranking tasks 

 _Originally posted by @Samoed in [#2728](https://github.com/embeddings-benchmark/mteb/issues/2728#issuecomment-3145596105)_

PR #2984, opened 2025-08-04
combine instructions with queries

Samoed:
Closes https://github.com/embeddings-benchmark/mteb/issues/2969

I've merged `instruction` and `queries` datasets. Now queries dataset would look like
```json
{"id": "123", "text": "some text", "insruction": "instruction"}
```

### DIFF
--- a/mteb/create_dataloaders.py
+++ b/mteb/create_dataloaders.py
@@ -15,7 +15,6 @@ from mteb.types import (
     PromptType,
     QueryDatasetType,
 )
-from mteb.types import Conversation, ConversationTurn
 from mteb.types._encoder_io import CorpusInput, ImageInput, QueryInput, TextInput
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


mteb/create_dataloaders.py
```diff
@@ -64,6 +64,16 @@ def create_dataloader_for_retrieval_corpus(
     return torch.utils.data.DataLoader(dataset, **dataloader_kwargs)
 
 
+def combine_queries_with_instruction_text(row: dict[str, str]) -> dict[str, str]:
```
orionw, line 67: This can still be over-ridden right? That way custom models can add their custom SEP tokens?

Samoed: To the `encode` function will be passed 
```
{
   "text": "query" + "instruction",
   "query": "query",
   "instruction": "instruction",
}
```
and in model implementation can be combined query with instruction as needed

orionw: Beautiful, thanks!

Samoed:
For `InstructIR` queries are already duplicated to match with instructions https://huggingface.co/datasets/mteb/InstructIR-mteb/viewer/queries

Also in docstring we have that we're alredy need to duplicate queries to match with instruction https://github.com/embeddings-benchmark/mteb/blob/64478e746f889fbd05dd226f456755be2bac9caa/mteb/abstasks/AbsTaskRetrieval.py#L95

orionw:
I see, thanks for pointing out. I think the FollowIR and mFollowIR tasks don't require duplication and the only other Instruction* task I see is IFIR, which seems to only have one instruction per query. So I think we're good then?

Samoed:
I think yes. Can you also review https://github.com/embeddings-benchmark/mteb/pull/2970. I have questions here about top_ranked and how to integrate cross-encoders

KennethEnevoldsen:
Looks good to me as well

PR #3005, opened 2025-08-07
Add Voyage 3.5 model configuration

fzowl:
- [ ]  I have ensured that my model can be loaded using
  - [ ]  mteb.get_model(model_name, revision) and
  - [ ]  mteb.get_model_meta(model_name, revision)
- [ ] The model is public, i.e. is available either as an API or the wieght are publicly avaiable to download


isaac-chung (approved):
Thanks!

Issue #2992, opened 2025-08-06
[maeb] clap&encodec bugs from model running


gowitheflow-1998:
some new bugs from results running.
Might be due to the truncation PR, otherwise worth fixing.

clap
```
Error while evaluating AmbientAcousticContextClustering: the wav size should be less than or equal to the swin input size
Traceback (most recent call last):
  File "/mnt/workspace/workgroup/chx/mteb-maeb/run_models5.py", line 84, in <module>
    results = evaluation.run(model, output_folder="/mnt/workspace/workgroup/chx/mteb-maeb/results", batch_size=8)
  File "/mnt/workspace/workgroup/chx/mteb-maeb/mteb/evaluation/MTEB.py", line 672, in run
    raise e
  File "/mnt/workspace/workgroup/chx/mteb-maeb/mteb/evaluation/MTEB.py", line 625, in run
    results, tick, tock = self._run_eval(
  File "/mnt/workspace/workgroup/chx/mteb-maeb/mteb/evaluation/MTEB.py", line 307, in _run_eval
    results = task.evaluate(
  File "/mnt/workspace/workgroup/chx/mteb-maeb/mteb/abstasks/AbsTask.py", line 137, in evaluate
    scores[hf_subset] = self._evaluate_subset(
  File "/mnt/workspace/workgroup/chx/mteb-maeb/mteb/abstasks/Audio/AbsTaskAudioClustering.py", line 88, in _evaluate_subset
    embeddings = model.get_audio_embeddings(
  File "/mnt/workspace/workgroup/chx/mteb-maeb/mteb/models/clap_models.py", line 137, in get_audio_embeddings
    audio_features = self.model.get_audio_features(**inputs)
  File "/mnt/workspace/workgroup/chx/env/maeb/lib/python3.10/site-packages/transformers/models/clap/modeling_clap.py", line 1920, in get_audio_features
    audio_outputs = self.audio_model(
  File "/mnt/workspace/workgroup/chx/env/maeb/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/mnt/workspace/workgroup/chx/env/maeb/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1762, in _call_impl
    return forward_call(*args, **kwargs)
[...]

PR #3010, opened 2025-08-08
MAEB Models Eval Fixes 2

AdnanElAssadi56:
Solves #2992 

- Removed 30 seconds max length for clap model.
- Did 30 seconds truncation manually for encodec.

### DIFF
--- a/mteb/models/encodec_model.py
+++ b/mteb/models/encodec_model.py
@@ -124,7 +124,7 @@ class EncodecWrapper(Wrapper):
                 # Process audio through EnCodec's processor
                 max_samples = int(self.max_audio_length_seconds * self.sampling_rate)
                 batch_np = [audio[:max_samples].cpu().numpy() for audio in batch]
-          
+
                 inputs = self.processor(
                     raw_audio=batch_np,
                     sampling_rate=self.sampling_rate,
--- a/mteb/models/mctct_model.py
+++ b/mteb/models/mctct_model.py
@@ -191,7 +191,6 @@ class MCTCTWrapper(Wrapper):
                     truncation=True,
                     max_length=int(self.max_audio_length_seconds * self.sampling_rate),
                 ).to(self.device)
-    
 
                 # Get embeddings from the model
                 outputs = self.model(


AdnanElAssadi:
Hi @gowitheflow-1998 

I've tried to fix the problems you encountered in the issue. Please tell me if you encounter issues with the other models as well.

I think it is best to do a quick test over all models before we merge so that we don't encounter the same problem.

isaac-chung:
> I think it is best to do a quick test over all models before we merge 

@AdnanElAssadi56 can you run that please?

isaac-chung:
@KennethEnevoldsen @gowitheflow-1998 if we're truncating to 30sec, can we still say we process long audio?

KennethEnevoldsen:
> @KennethEnevoldsen @gowitheflow-1998 if we're truncating to 30sec, can we still say we process long audio?

So it is only in the model implementation that we limit to 30 seconds, not in the evaluation. 

We can note in the limitations sections that we do this and that this is mainly a computational constraint, but that the leaderboard allows evaluation of model without this constraint

isaac-chung:
> So it is only in the model implementation that we limit to 30 seconds, not in the evaluation.

What's the difference? It seems like we are _adding_ this constraint here

Samoed:
> What's the difference?

Models that can processing longer input audio sequences can use different lengths. Some 30 seconds, some several minutes etc

isaac-chung:
My question is more related to why we're adding the truncation when the models can handle more _and_ when we want to evaluate long audio processing.

KennethEnevoldsen:
@AdnanElAssadi56 after having looked at the errors it doesn't seem like they require us to limit the sequence length to 30 - seems like we can make a more lenient fix?

AdnanElAssadi:
> @AdnanElAssadi56 after having looked at the errors it doesn't seem like they require us to limit the sequence length to 30 - seems like we can make a more lenient fix?

Yes, for the clap model, I removed the 30s limit. 

For other models, we can probably get away with just turning truncation on and not have a max length. However, for models in which truncation is done manually that would be not possible, and we will have to enforce a specific limit.

KennethEnevoldsen:
@gowitheflow-1998 any good ideas on how to best solve this issue?

I would probably merge for now, and if specific models are very fast to run, we can then increase max length?

gowitheflow-1998:
@AdnanElAssadi56  thanks! Have tested them. For CLAP models, I needed to remove `truncation=True,` to get it to run otherwise would get a truncation not implemented error. encodec is runnable now. Others all look good! @KennethEnevoldsen

isaac-chung:
Maybe we want to keep both truncated and not truncated results to compare? Thinking about how we can claim that we process "long audios"

gowitheflow-1998:
@isaac-chung sure. agree!

AdnanElAssadi:
Just to summarize:

- Are we confirming the 30s limit?
- Should we check which models handle truncation natively and only enforce the 30s limit for those that don’t, or should we just stick with the uniform approach I used earlier?
- For the extra run, depending on compute, should we try running models without truncation and compare results (separate branch or something)?

KennethEnevoldsen:
@AdnanElAssadi56, it seems like there are still some bugs left (@gowitheflow-1998 will create an issue on them). It also seems like there is a bit of an efficiency issue (the GPUs are not fully utilized). So it might be that we could increase the duration simply by utilizing a dataloader similar to how we do with images.

AdnanElAssadi:
> @AdnanElAssadi56, it seems like there are still some bugs left (@gowitheflow-1998 will create an issue on them). It also seems like there is a bit of an efficiency issue (the GPUs are not fully utilized). So it might be that we could increase the duration simply by utilizing a dataloader similar to how we do with images.

Yes, I’ve also noticed some padding-related bugs, which I’ll work on fixing (they only show up when evaluating on certain datasets).

Regarding the dataloader, I would assume that’s handled on the task side, not the model side. Is it implemented in a specific image task so I can look at the reference?

KennethEnevoldsen:
> Yes, I’ve also noticed some padding-related bugs, which I’ll work on fixing (they only show up when evaluating on certain datasets).

Let us just start with these 

> Regarding the dataloader, I would assume that’s handled on the task side, not the model side. Is it implemented in a specific image task so I can look at the reference?

You can take a look at the `ImagekNNClassificationEvaluator`

AdnanElAssadi:
>> Yes, I’ve also noticed some padding-related bugs, which I’ll work on fixing (they only show up when evaluating on certain datasets).

> Let us just start with these
 
@KennethEnevoldsen They should be good now.

@gowitheflow-1998 Feel free to share other issues you've encountered from the models side.

I think two tasks remain before we finalize this PR. 

1.  Implement Dataloader for GPU utilization

 2. Decide whether to:
     - Let each model handle truncation natively (flexible, respects training assumptions).  
     - Enforce a fixed max length (e.g., 30s) across all models or just models that don't support truncation natively (may mismatch some models’ training).
     
 For the models I've worked with above, I've added an optional max length argument which would allow us to decide max length when doing evaluation (defaults to 30s if missing).
 
 EDIT: For 1, I think it is better that we do it in a separate https://github.com/embeddings-benchmark/mteb/pull/3131.

isaac-chung:
Might want to log "max length" in the results file somehow so that they are reproducible.

KennethEnevoldsen:
We def. want:

> Let each model handle truncation natively (flexible, respects training assumptions).

commented on the GPU implementation in the PR

AdnanElAssadi:
> We def. want:

I'll go ahead then and give models truncation flexibility. For ones that don't have, I'll enforce a max limit that can be passed (30s for now).

AdnanElAssadi:
> I'll go ahead then and give models truncation flexibility. For ones that don't have, I'll enforce a max limit that can be passed (30s for now).

Done. 

@KennethEnevoldsen @isaac-chung Please give it a final confirmation, so we can merge.

@gowitheflow-1998 will be communicating to us any error that arises, and I'll fix asap.

PR #3023, opened 2025-08-10
Supplement missing training sets

PennyYu123:
I added the training data that was not included before.  


mteb/models/qzhou_models.py
```diff
@@ -26,6 +26,35 @@ def instruction_template(
     return f"Instruct: {instruction}\nQuery:"
 
 
+qzhou_training_data = {
```
Samoed, line 29: Maybe to use then
```python
qzhou_training_data = {
       **bge_m3_training_data,
        **bge_chinese_training_data,
        **bge_full_data,
        **E5_MISTRAL_TRAINING_DATA,
        "AFQMC": ["train"],
        ...
}
```

PennyYu123: OK

PennyYu123:
https://github.com/embeddings-benchmark/results/pull/250

KennethEnevoldsen (approved):

PR #3024, opened 2025-08-11
model: Add M2-BERT model

redbrain:
- [x] The model is public, i.e. is available either as an API or the weights are publicly avaiable to download

Note that `trust_remote_code=True` is required to load the model.

### DIFF
--- a/mteb/models/m2bert_models.py
+++ b/mteb/models/m2bert_models.py
@@ -17,5 +17,5 @@ m2_32k = ModelMeta(
     use_instructions=False,
     public_training_code="https://github.com/HazyResearch/m2",
     public_training_data="https://huggingface.co/datasets/allenai/c4",
-    training_datasets=None, # unknown
+    training_datasets={}, # C4
 )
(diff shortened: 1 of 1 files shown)


mteb/models/m2bert_models.py
```diff
@@ -0,0 +1,21 @@
...
+    embed_dim=768,
+    license="apache-2.0",
+    max_tokens=32768,
+    reference="https://huggingface.co/togethercomputer/m2-bert-80M-32k-retrieval",
+    similarity_fn_name="cosine",
+    framework=["Sentence Transformers", "PyTorch"],
+    use_instructions=False,
+    public_training_code="https://github.com/HazyResearch/m2",
+    public_training_data="https://huggingface.co/datasets/allenai/c4",
+    training_datasets=None, # unknown
```
Samoed:
```suggestion
    training_datasets={}, # C4
```

mteb/models/m2bert_models.py
```diff
@@ -0,0 +1,21 @@
+from mteb.model_meta import ModelMeta
+
+m2_32k = ModelMeta(
+    name="togethercomputer/m2-bert-80M-32k-retrieval",
```
Samoed, line 4: It seems that model uses `AutoModelForSequenceClassification` and won't work with `SentenceTransformers` correctly. Also it reqires `trust_remote_code`

PR #3035, opened 2025-08-17
ci: reduce parallel runs for when checking if a dataset exists

KennethEnevoldsen:
The hope is that this will prevent many of the current [errors](https://github.com/embeddings-benchmark/mteb/actions/runs/17019125199/job/48245690831)

I am welcoming of alternative solutions.


Samoed (approved):
I think this can solve some issues

KennethEnevoldsen:
Let us see - if it doesn't we can easily roll it back

PR #3040, opened 2025-08-17
[v2] Two stage rerank

Samoed:
```python
import mteb

model = mteb.get_model("minishlab/potion-base-2M")
task = mteb.get_task("NanoArguAnaRetrieval")

res = mteb.evaluate(
    model,
    task,
    save_retrieval_results=True,
    # overwrite_strategy="always",
)
task.convert_to_reranking(model)
model = mteb.get_model("minishlab/potion-base-4M")
mteb.evaluate(
    model,
    task,
)
```

### DIFF
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -6,7 +6,7 @@ import random
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from copy import copy
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import datasets
 import numpy as np
@@ -18,7 +18,6 @@ from sklearn.preprocessing import MultiLabelBinarizer
 
 from mteb.abstasks._stratification import _iterative_train_test_split
 from mteb.abstasks.task_metadata import TaskMetadata
-from mteb.cache import ResultCache
 from mteb.languages import LanguageScripts
 from mteb.models.models_protocols import (
     CrossEncoderProtocol,
@@ -28,6 +27,9 @@ from mteb.models.models_protocols import (
 )
 from mteb.types import HFSubset, ScoresDict
 from mteb.types.statistics import DescriptiveStatistics
+
+if TYPE_CHECKING:
+    from mteb.cache import ResultCache
 
 logger = logging.getLogger(__name__)
 
@@ -132,7 +134,7 @@ class AbsTask(ABC):
         subsets_to_run: list[HFSubset] | None = None,
         *,
         encode_kwargs: dict[str, Any],
-        cache: ResultCache | None,
+        cache: ResultCache | None = None,
         **kwargs: Any,
     ) -> dict[HFSubset, ScoresDict]:
         """Evaluates a Sentence Embedding Model on the task.
--- a/mteb/abstasks/AbsTask.py
+++ b/mteb/abstasks/AbsTask.py
@@ -18,6 +18,7 @@ from sklearn.preprocessing import MultiLabelBinarizer
 
 from mteb.abstasks._stratification import _iterative_train_test_split
 from mteb.abstasks.task_metadata import TaskMetadata
+from mteb.cache import ResultCache
 from mteb.languages import LanguageScripts
 from mteb.models.models_protocols import (
     CrossEncoderProtocol,
@@ -131,6 +132,7 @@ class AbsTask(ABC):
         subsets_to_run: list[HFSubset] | None = None,
         *,
         encode_kwargs: dict[str, Any],
+        cache: ResultCache | None,
         **kwargs: Any,
     ) -> dict[HFSubset, ScoresDict]:
         """Evaluates a Sentence Embedding Model on the task.
@@ -141,6 +143,7 @@ class AbsTask(ABC):
             split: Which datasplit to be used.
             subsets_to_run: List of HFSubsets to evaluate. If None, all subsets are evaluated.
             encode_kwargs: Additional keyword arguments that are passed to the model's `encode` method.
+            cache: Cache of the tasks.
             kwargs: Additional keyword arguments that are passed to the _evaluate_subset method.
         """
         if isinstance(model, CrossEncoderProtocol) and not self.support_cross_encoder:
--- a/mteb/abstasks/AbsTaskRetrieval.py
+++ b/mteb/abstasks/AbsTaskRetrieval.py
@@ -9,8 +9,8 @@ from typing import Any, Callable
 
 from datasets import Dataset, DatasetDict, concatenate_datasets
 
-from mteb import get_model_meta
 from mteb.cache import ResultCache
+from mteb.models.get_model_meta import get_model_meta
 from mteb.models.model_meta import ModelMeta
 from mteb.models.models_protocols import Encoder, MTEBModels
 from mteb.types import (
@@ -261,7 +261,7 @@ class AbsTaskRetrieval(AbsTask):
         subsets_to_run: list[HFSubset] | None = None,
         *,
         encode_kwargs: dict[str, Any],
-        cache: ResultCache | None,
+        cache: ResultCache | None = None,
         save_retrieval_results: bool = False,
         **kwargs,
     ) -> dict[HFSubset, ScoresDict]:
@@ -344,19 +344,25 @@ class AbsTaskRetrieval(AbsTask):
         if self.save_retrieval_results:
             if hasattr(model, "mteb_model_meta"):
                 model_meta = model.mteb_model_meta
-                previous_results_path = self.previouse_results_path(
+                previous_results_path = self.previous_results_path(
                     model_meta, self.cache
                 )
                 previous_results = {}
                 if previous_results_path.exists():
-                    with open(previous_results_path) as f:
+                    with previous_results_path.open() as f:
                         previous_results = json.load(f)
 
                 if hf_subset not in previous_results:
                     previous_results[hf_subset] = {}
-                previous_results[hf_subset][hf_split] = results
 
-                with open(previous_results_path, "w") as f:
+                top_k_sorted = {}
+                for query_id, values in results.items():
+                    sorted_keys = sorted(values, key=values.get, reverse=True)
+                    top_k_sorted[query_id] = sorted_keys[: self.top_k]
+
+                previous_results[hf_subset][hf_split] = top_k_sorted
+
+                with previous_results_path.open("w") as f:
                     json.dump(previous_results, f)
             else:
                 logger.warning(
@@ -385,6 +391,13 @@ class AbsTaskRetrieval(AbsTask):
             hf_split=hf_split,
             hf_subset=hf_subset,
         )
+        previous_results_data = {}
+        if self.previous_results_model_meta:
+            previous_results_data = {
+                "top_ranked_creation": self.previous_results_model_meta.to_dict(),
+                "top_k": self.top_k,
+            }
+
         return make_score_dict(
             ndcg,
             _map,
@@ -394,7 +407,7 @@ class AbsTaskRetrieval(AbsTask):
             naucs,
             naucs_mrr,
             task_specific_scores,
-            self.previous_results_model_meta,
+            previous_results_data,
         )
 
     def task_specific_scores(
@@ -570,7 +583,7 @@ class AbsTaskRetrieval(AbsTask):
     def previous_results_name(self) -> str:
         return f"{self.metadata.name}_previous.json"
 
-    def previouse_results_path(
+    def previous_results_path(
         self,
         model_meta: ModelMeta,
         cache: ResultCache | None,
@@ -586,6 +599,7 @@ class AbsTaskRetrieval(AbsTask):
         model: MTEBModels | ModelMeta | str,
         cache: ResultCache | None = ResultCache(),
         remote: bool = False,
+        top_k: int = 10,
     ) -> None:
         """Converts a reranking task to re-ranking by loading results from model run
 
@@ -593,6 +607,7 @@ class AbsTaskRetrieval(AbsTask):
             model: Model which have previous results
             cache: Cache to use
             remote: If true, re-ranking task is re-ranked
+            top_k: top k
         """
         if isinstance(model, MTEBModels):
             if not hasattr(model, "mteb_model_meta"):
@@ -610,7 +625,7 @@ class AbsTaskRetrieval(AbsTask):
                 "Cannot convert re-ranking task to re-ranking, because cache is None."
             )
 
-        previous_results_path = self.previouse_results_path(
+        previous_results_path = self.previous_results_path(
             model_meta, cache, remote=remote
         )
         if not previous_results_path.exists():
@@ -629,7 +644,10 @@ class AbsTaskRetrieval(AbsTask):
                 top_ranked = previous_results[subset][split]
                 if not isinstance(top_ranked, dict):
                     raise ValueError("Previous top ranked results is not a dictionary.")
-                self.dataset[subset][split]["top_ranked"] = top_ranked
+                self.dataset[subset][split]["top_ranked"] = {
+                    q_id: c_ids[:top_k] for q_id, c_ids in top_ranked.items()
+                }
+        self.top_k = top_k
         self.previous_results_model_meta = model_meta
 
 
--- a/mteb/abstasks/AbsTaskRetrieval.py
+++ b/mteb/abstasks/AbsTaskRetrieval.py
@@ -9,7 +9,10 @@ from typing import Any, Callable
 
 from datasets import Dataset, DatasetDict, concatenate_datasets
 
-from mteb.models.models_protocols import Encoder
+from mteb import get_model_meta
+from mteb.cache import ResultCache
+from mteb.models.model_meta import ModelMeta
+from mteb.models.models_protocols import Encoder, MTEBModels
 from mteb.types import (
     HFSubset,
     RelevantDocumentsType,
@@ -108,6 +111,9 @@ class AbsTaskRetrieval(AbsTask):
     cross_encoder_top_k: int = 100
     support_cross_encoder: bool = True
     support_search: bool = True
+    save_retrieval_results: bool = False
+    cache: ResultCache | None = None
+    previous_results_model_meta: ModelMeta | None = None
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -255,6 +261,8 @@ class AbsTaskRetrieval(AbsTask):
         subsets_to_run: list[HFSubset] | None = None,
         *,
         encode_kwargs: dict[str, Any],
+        cache: ResultCache | None,
+        save_retrieval_results: bool = False,
         **kwargs,
     ) -> dict[HFSubset, ScoresDict]:
         """Evaluate the model on the retrieval task.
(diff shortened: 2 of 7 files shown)


tests/test_cli.py
```diff
@@ -187,12 +187,3 @@ def test_create_meta_from_existing(
     command = f"{sys.executable} -m mteb create_meta --results_folder {results.as_posix()} --output_path {output_path.as_posix()} --from_existing {existing_readme.as_posix()} --overwrite"
     result = subprocess.run(command, shell=True, capture_output=True, text=True)
     assert result.returncode == 0, "Command failed"
-
-
-def test_save_predictions():
```
KennethEnevoldsen, line 192: isn't this still used by MTEB()?

mteb/evaluation/evaluators/retrieval_evaluator.py
```diff
@@ -49,21 +48,10 @@ def __init__(
 
     def __call__(
         self,
-        model: SearchProtocol | Encoder | CrossEncoderProtocol,
+        search_model: SearchProtocol,
         encode_kwargs: dict[str, Any],
         **kwargs: Any,
     ) -> RetrievalOutputType:
-        if isinstance(model, Encoder) and not isinstance(model, SearchProtocol):
```
KennethEnevoldsen, line 56: Why delete?

mteb/evaluate.py
```diff
@@ -313,45 +325,39 @@ def evaluate(
...
-        model_name=model_name,
-        model_revision=model_revision,
-        task_results=[result],
-    )
+        return ModelResult(
+            model_name=model_name,
+            model_revision=model_revision,
+            task_results=[result],
+        )
+    except Exception as e:
```
KennethEnevoldsen: I very explicitly don't want a try except here. It makes debugging much harder as the error is raised after the function scopes are deleted (so no chance to trace it back through the stack).

mteb/cache.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, cache_path: Path | str | None = None) -> None:
             self.cache_path = Path(cache_path)
         else:
             self.cache_path = self.default_cache_path
+        self.cache_path.mkdir(parents=True, exist_ok=True)
```
KennethEnevoldsen: I would do it when we try to write, not before

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -591,6 +600,86 @@ def _push_section(
...
+        )
+        return full_task_path.parent / self.previous_results_name
+
+    def convert_to_reranking(
+        self,
+        model: MTEBModels | ModelMeta | str,
+        cache: ResultCache | None = ResultCache(),
+        remote: bool = False,
+        top_k: int = 10,
+    ) -> None:
```
KennethEnevoldsen: I would expect this to return a reranking task (and not convert the existing)

Also, ResultCache isn't really intended for predictions (I wouldn't overload it)

KennethEnevoldsen (commented):
Hmm thanks for making the PR @Samoed. 

I really don't like the `save_retrieval_results=True` argument as it clearly a task specific argument.

I do however like that it is possible to create a reranking from a retrieval task. How about 

```py
import mteb

model = mteb.get_model("minishlab/potion-base-2M")
retrieval_task = mteb.get_task("NanoArguAnaRetrieval")

model = SaveRetrievaPredictionsWrapper(model)

res = mteb.evaluate(model, retrieval_task)
reranking_task = retrieval_task.convert_to_reranking(model.predictions, top_k=100)

model = mteb.get_model("jinaai/jina-reranker-v2-base-multilingual")
mteb.evaluate(model, reranking_task)
```

Alternatively, we could add a more general `save_predictions` flag, that can be used for all tasks and just add a NotImplementedError() for all tasks but retrieval. I could see that as being an option as well. Besides the many not-implemented errors this would cause, it would also lead to a requirement to handle new arguments, such as where to save predictions, etc.

The Wrapper isolate the concern and leave the evalautor only focused on evaluation

mteb/evaluation/evaluators/retrieval_evaluator.py
Samoed: I moved to `RetrievalTask`

mteb/abstasks/AbsTaskRetrieval.py
Samoed: > Also, ResultCache isn't really intended for predictions (I wouldn't overload it)

Yes, but I'm taking only path to save predictions from it, because I don't want to pass additional task specific parameters. Do you have other suggestions?

Samoed:
> I really don't like the save_retrieval_results=True argument as it clearly a task specific argument.

I defenetly agree, but I don't see another solution to the problem. And this is because I didn't want to add it implement firstly.

> model = SaveRetrievaPredictionsWrapper(model)

I don't think that this would solve the problems, because models can save only embedings, but not predictions for the retrieval task. I don't know how resuse results from the wrapper then.

> Alternatively, we could add a more general save_predictions flag, that can be used for all tasks and just add a NotImplementedError() for all tasks but retrieval

I can make implementation for it, because this would solve problems for the tasks

mteb/evaluate.py
Samoed: The current implementation raise error as
```
  File "/.../mteb/t.py", line 9, in <module>
    res = mteb.evaluate(
  File "/.../mteb/mteb/evaluate.py", line 357, in evaluate
    raise e
  File "/.../mteb/mteb/evaluate.py", line 330, in evaluate
    result = _evaluate(
  File "/.../mteb/mteb/evaluate.py", line 116, in _evaluate
    result = _evaluate(
  File "/.../mteb/mteb/evaluate.py", line 140, in _evaluate
    task_results[split] = task.evaluate(
  File "/.../mteb/mteb/abstasks/AbsTaskRetrieval.py", line 296, in evaluate
    return super().evaluate(
  File "/.../mteb/mteb/abstasks/AbsTask.py", line 190, in evaluate
    scores[hf_subset] = self._evaluate_subset(
  File "/.../mteb/mteb/abstasks/AbsTaskRetrieval.py", line 331, in _evaluate_subset
    0/0
ZeroDivisionError: division by zero

```
I don't see a problem currenly. I can cahange to similar as previous solution. I've cahnged it because previously there was a difference if run model with `raise_error` or not.

KennethEnevoldsen: I would change it back and fix the issue (this is not for this PR)

mteb/abstasks/AbsTaskRetrieval.py
KennethEnevoldsen: Hmm I would probably don't want to use the `.cache` folder here as both the files are larger and the content would probably rarely be re-used.

KennethEnevoldsen:
> I don't think that this would solve the problems, because models can save only embedings, but not predictions for the retrieval task. I don't know how resuse results from the wrapper then.

Why can't they save predicitons? doesn't the search interface return RetrievalOutputType?

> I can make implementation for it, because this would solve problems for the tasks

Let us decide on a good solutions before spending the implementation time

Samoed:
> Why can't they save predicitons? doesn't the search interface return RetrievalOutputType?

Oh, I forgot that we have search interface. Yes, wrapper should work then

tests/test_cli.py
Samoed: Yes, this still used by `MTEB`, but I don't want to support it, because otherwise it would required to support a lot of extra parameters in `AbsRetrieval` which we don't want I think

mteb/evaluate.py
Samoed: Revert this change

mteb/abstasks/AbsTaskRetrieval.py
Samoed: Removed cache usage

orionw (approved):
Sorry for the late review. In general this looks fine to me.

As a meta-comment though, I will say I'm nervous we're adding too many wrappers. Do they stack? Can we save predictions and cache embeddings and ...? 

Maybe a personal preference but I think arguments to functions are a much nicer UI from a users perspective than doing a bunch of wrappers. However, we have so many it's probably too late to change all of these at this point. Just worth noting for future extensions.

Samoed:
> Can we save predictions and cache embeddings and ...?

I thought about this, too. This will be like `RetrievalSaveResultsWrapper(CacheEmbeddingWrapper(model))` and should work

> Maybe a personal preference but I think arguments to functions are a much nicer UI from a users perspective than doing a bunch of wrappers. However, we have so many it's probably too late to change all of these at this point. Just worth noting for future extensions.

In new `mteb.evaluate` we have just 5 parameters https://github.com/embeddings-benchmark/mteb/blob/cf01bc89ad053ecc4df3c576a430e1a363708d79/mteb/evaluate.py#L162-L171

tests/test_cli.py
KennethEnevoldsen: So this will break MTEB()?

Samoed: I've chedked a bit futher. It will break only cli for some part, becuase we don't support this flag for now, but we need to refactor CLI a bit to use `mteb.evaluate` instead of `MTEB`. We can add support of this flag to save backward compatibility. But this flag is not supported by `MTEB` directly, it's just passed thought kwargs and not documented https://github.com/embeddings-benchmark/mteb/blob/cf01bc89ad053ecc4df3c576a430e1a363708d79/mteb/evaluation/MTEB.py#L254-L266

mteb/abstasks/AbsTaskRetrieval.py
```diff
@@ -108,6 +115,7 @@ class AbsTaskRetrieval(AbsTask):
     cross_encoder_top_k: int = 100
     support_cross_encoder: bool = True
     support_search: bool = True
+    previous_results_model_meta: dict[str, Any] | None = None
```
KennethEnevoldsen, line 118: Not entirely sure why this is needed. Can you clarify?

mteb/abstasks/AbsTaskRetrieval.py
KennethEnevoldsen, line 1: generally like the rewrite here

mteb/models/search_wrappers.py
```diff
@@ -344,3 +346,80 @@ def search(
             results[query_id][corpus_id] = float(score)
 
         return results
+
+
+class SaveRetrievaPredictionsWrapper:
+    def __init__(
+        self,
+        model: SearchProtocol,
+        results_path: str | Path,
```
KennethEnevoldsen: So based on @orionw and @Samoed comment, I might be better to add the following arguments:

save_retrieval_predictions: bool
prediction_save_path: Path

And then in the future we can convert raise a depreciationwarning on `save_retrieval_predictions ` and instead refer to `save_predictions`.

let me know what you think?

(sorry @Samoed i understand that this is a digression from before)

KennethEnevoldsen: We might still want to keep the wrapper internally (without exposing it to the user)

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -45,6 +45,11 @@ def __init__(
         """
         if isinstance(model, str):
             self.model = SentenceTransformer(model, revision=revision, **kwargs)
+            from mteb.models.get_model_meta import (
```
KennethEnevoldsen, line 48: Is this due to circular imports?

Samoed: Yes

mteb/abstasks/AbsTaskRetrieval.py
Samoed: This needed to add some reproducibility of results. `previous_results_model_meta` will be passed to results file for 2 stage rerank

mteb/models/search_wrappers.py
Samoed: > And then in the future we can convert raise a depreciationwarning on save_retrieval_predictions  and instead refer to save_predictions.

Where do you want to add it? To `mteb.evaluate`?

mteb/models/sentence_transformer_wrapper.py
KennethEnevoldsen: A bit annoying. Ideally, we shouldn't have problems with circular imports within the same module (but not for this PR)

mteb/models/search_wrappers.py
KennethEnevoldsen: Yea that would be the idea

mteb/abstasks/AbsTaskRetrieval.py
KennethEnevoldsen: So how will this look?

mteb/models/search_wrappers.py
Samoed: Should we add method `save_predictions` to all tasks then? And we can use only `prediction_save_path: Path` then

KennethEnevoldsen: Hmm but that required a lot of implementation across the different tasks?

mteb/abstasks/AbsTaskRetrieval.py
Samoed: Might be we should add only name + revision, but there can be other changes, so I'm dumping full model meta 

```json
{
  "dataset_revision": "NA",
  "task_name": "MockRetrievalTask",
  "mteb_version": "1.38.34",
  "scores": {
    "val": [
      {
        "ndcg_at_1": 0.5,
        "ndcg_at_3": 0.5,
        "ndcg_at_5": 0.5,
        "ndcg_at_10": 0.5,
        "ndcg_at_20": 0.5,
        "ndcg_at_100": 0.5,
        ...
        "previous_results_model_meta": {
          "loader_kwargs": {},
          "name": "sentence-transformers/average_word_embeddings_komninos",
          "revision": "21eec43590414cb8e3a6f654857abed0483ae36e",
          "release_date": null,
          "languages": null,
          "n_parameters": null,
          "memory_usage_mb": null,
          "max_tokens": null,
          "embed_dim": null,
          "license": "apache-2.0",
          "open_weights": true,
          "public_training_code": null,
          "public_training_data": null,
          "framework": [
            "PyTorch",
            "Sentence Transformers"
          ],
          "reference": null,
          "similarity_fn_name": null,
          "use_instructions": null,
          "training_datasets": null,
          "adapted_from": null,
          "superseded_by": null,
          "modalities": [
            "text"
          ],
          "is_cross_encoder": null,
          "citation": null,
          "loader": "sentence_transformers_loader"
        },
        "main_score": 0.5,
        "hf_subset": "default",
        "languages": [
          "eng-Latn"
        ]
      }
    ],
}
```

mteb/models/search_wrappers.py
Samoed: Yes, but I don't like that we're passing task-specific args. Also, for some users this can be helpful

mteb/abstasks/AbsTaskRetrieval.py
KennethEnevoldsen: Hmmm, I see. It might be cleaner to just use the name and revision, since we have the mteb version there).

mteb/models/search_wrappers.py
KennethEnevoldsen: Sure, it was simply that we don't have it implemented yet so I am afraid it would give a poor user experience

mteb/abstasks/AbsTaskRetrieval.py
Samoed: Changed and now it looks like:
```json
{
  "dataset_revision": "NA",
  "task_name": "MockRetrievalTask",
  "mteb_version": "1.38.34",
  "scores": {
    "val": [
      {
        "ndcg_at_1": 0.5,
        "ndcg_at_3": 0.5,
        ...
        "previous_results_model_meta": {
          "model_name": "sentence-transformers/average_word_embeddings_komninos",
          "revision": "21eec43590414cb8e3a6f654857abed0483ae36e"
        },
        "main_score": 0.5,
        "hf_subset": "default",
        "languages": [
          "eng-Latn"
        ]
      }
    ]
  },
  "evaluation_time": 0.12439084053039551,
  "kg_co2_emissions": null
}
```

mteb/models/search_wrappers.py
Samoed: I will implement it for retrieval in this pr and will add to other tasks later

tests/test_cli.py
KennethEnevoldsen: Yeah indeed!

KennethEnevoldsen: Let is keep it like this, but remove it from the CLI args/docs

mteb/models/search_wrappers.py
KennethEnevoldsen: Alright let us do it this way (sorry for backtracking)

tests/test_cli.py
Samoed: If we implement an additional method for save results, I think we can leave it as is

mteb/models/sentence_transformer_wrapper.py
Samoed: This is hard to overcome in this file, because it importing it for the default loader

mteb/evaluate.py
```diff
@@ -168,8 +172,9 @@ def evaluate(
     encode_kwargs: dict[str, Any] | None = None,
     cache: ResultCache | None = ResultCache(),
     overwrite_strategy: str | OverwriteStrategy = "only-missing",
+    results_folder: Path | str | None = None,
```
KennethEnevoldsen: result_folder is a poor name as it confuses model predictions with evaluation results. I would rename it to `prediction_folder`

mteb/evaluate.py
```diff
@@ -188,6 +193,7 @@ def evaluate(
                 changed.
             - "only-cache": Only load the results from the cache folder and do not run the task. Useful if you just want to load the results from the
                 cache.
+        results_folder: Folder to save predictions for the task results
```
```suggestion
        prediction_folder: Optional folder in which to save model predictions for the task.
```

KennethEnevoldsen: We should also specify how the predictions are saved.

mteb/abstasks/AbsTask.py
```diff
@@ -201,12 +205,51 @@ def _evaluate_subset(
...
+        results_folder: Path | None = None,
         **kwargs: Any,
     ) -> ScoresDict:
         raise NotImplementedError(
             "If you are using the default evaluate method, you must implement _evaluate_subset method."
         )
 
+    def save_task_predictions(
+        self,
+        results: dict[str, Any],
```
KennethEnevoldsen:
```suggestion
        predictions: dict[str, Any],
```

mteb/evaluate.py
Samoed: This is hard to specify, because this will be task-type specific

Samoed:
Updated example in description

mteb/evaluate.py
KennethEnevoldsen: Yeah, so just say that it will be saved under the `{taskname}_predictions.json`.

KennethEnevoldsen:
Looks great @Samoed - I think this is good to merge - we need to update some docs away from using MTEB() for two-stage retrieval, but I can do that in a separate PR

mteb/evaluate.py
Samoed: added

PR #3046, opened 2025-08-18
fix: Updated revision for jina-embeddings-v4

jupyterjazz:
We should remove the hardcoded revision from the jina-embeddings-v4 loader, as the specified version is incompatible with the latest Transformers release and causes evaluation to fail

### DIFF
--- a/mteb/models/jina_models.py
+++ b/mteb/models/jina_models.py
@@ -545,6 +545,7 @@ jina_embeddings_v4 = ModelMeta(
     loader=partial(  # type: ignore
         JinaV4Wrapper,
         model="jinaai/jina-embeddings-v4",
+        revision="4a58ca57710c49f51896e4bc820e202fbf64904b",
         trust_remote_code=True,
         model_prompts={
             "Retrieval-query": "retrieval.query",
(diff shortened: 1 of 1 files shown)


Samoed:
Can you open an issue in transformers to fix this?

jupyterjazz:
The issue was in the v4 model code, which has been fixed, but because the loader specifies an older revision, it still fails with MTEB

Samoed:
But you can bump revision instead of removing it

mteb/models/jina_models.py
```diff
@@ -545,7 +545,6 @@ def get_programming_task_override(
     loader=partial(  # type: ignore
         JinaV4Wrapper,
         model="jinaai/jina-embeddings-v4",
```
KennethEnevoldsen, line 547:
```suggestion
        model="jinaai/jina-embeddings-v4",
        revision="4a58ca57710c49f51896e4bc820e202fbf64904b",
```

KennethEnevoldsen: Also do run it on a task to confirm that this solves the problem

jupyterjazz:
> you can bump revision instead of removing it

Ok

> Also do run it on a task to confirm that this solves the problem

Verified

PR #3050, opened 2025-08-18
model: add granite-embedding-english R2 models

aashka-trivedi:
Adds Granite Embedding English R2 models

    - [x] mteb.get_model(model_name, revision) and
    - [x] mteb.get_model_meta(model_name, revision)


aashka-trivedi:
Thank you @KennethEnevoldsen for the lightning-fast response!

KennethEnevoldsen:
You arrived at just the right time - Are there plans for a multilingual version of this? I am really liking the look of the model so far

aashka-trivedi:
@KennethEnevoldsen we have multilingual models in the plan :)

Issue #3064, opened 2025-08-21
Automatically label stale PRs


isaac-chung:
https://github.com/actions/stale

With something like:

```
name: Close stale issues and PRs
on:
  schedule:
    - cron: '30 1 * * *' # runs daily at 1:30 AM
jobs:
  stale:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/stale@v9
        with:
          stale-pr-label: 'stale'
          pull-request-days-before-stale: 30
          pull-request-days-before-close: 10
          exempt-pr-labels: 'work-in-progress,awaiting-review'
          stale-pr-message: 'This pull request has been automatically marked as stale due to inactivity.'
```

Stale PRs will also be closed after certain days of inactivity.

PR #3066, opened 2025-08-22
ci: Add stale workflow

isaac-chung:
Fixes https://github.com/embeddings-benchmark/mteb/issues/3064
- Labels a "needs-more-infomation" issue stale after 60 days
- Closes a stale "needs-more-infomation" issue after 7 days
- Labels a PR stale after 14 days
- Closes a stale PR after 7 days

Any update (update/comment) can reset the stale idle time on the issues/PRs.

### DIFF
--- a/.github/workflows/stale_pr.yml
+++ b/.github/workflows/stale_pr.yml
@@ -15,6 +15,6 @@ jobs:
           stale-pr-label: 'stale'
           days-before-pr-stale: 14
           days-before-pr-close: 7
-          any-of-issue-labels: 'Bug'
+          any-of-issue-labels: 'more information needed'
           stale-pr-message: 'This pull request has been automatically marked as stale due to inactivity.'
           close-pr-message: 'This pull request has been automatically closed due to inactivity.'
--- a/.github/workflows/stale_pr.yml
+++ b/.github/workflows/stale_pr.yml
@@ -1,4 +1,4 @@
-name: Close stale PRs
+name: Label and Close stale PRs
 on:
   schedule:
     - cron: '30 1 * * *' # runs daily at 1:30 AM
--- a/.github/workflows/stale_pr.yml
+++ b/.github/workflows/stale_pr.yml
@@ -1,0 +1,16 @@
+name: Close stale PRs
+on:
+  schedule:
+    - cron: '30 1 * * *' # runs daily at 1:30 AM
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9.1.0
+        with:
+          stale-pr-label: 'stale'
+          days-before-pr-stale: 14
+          days-before-pr-close: 7
+          any-of-issue-labels: 'Bug'
+          stale-pr-message: 'This pull request has been automatically marked as stale due to inactivity.'
+          close-pr-message: 'This pull request has been automatically closed due to inactivity.'
--- a/.github/workflows/stale_pr.yml
+++ b/.github/workflows/stale_pr.yml
@@ -13,6 +13,7 @@ jobs:
       - uses: actions/stale@v9.1.0
         with:
           stale-pr-label: 'stale'
+          stale-issue-label: 'stale'
           days-before-pr-stale: 14
           days-before-pr-close: 7
           any-of-issue-labels: 'more information needed'
--- a/.github/workflows/stale_pr.yml
+++ b/.github/workflows/stale_pr.yml
@@ -5,6 +5,10 @@ on:
 jobs:
   stale:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
Hmm, I understand the desire to add this, but I have generally found in other projects that it adds more work as opposed to removing it.

Have you tried in elsewhere @isaac-chung?

Samoed:
I'm not sure what it give to us

isaac-chung:
Closes stale PRs and issues automatically. How does it _add work_?

isaac-chung:
> I'm not sure what it give to us

Helps us declutter. Would you prefer to manually check every PR and issue for time since last activity instead?

Samoed:
> Closes stale PRs and issues automatically. How does it add work?

We have a lot of old prs with requests for adding dataset and models and I'm not sure if we want to close them.

isaac-chung:
> * Labels a Bug issue stale after 60 days
> * Closes a stale Bug issue after 7 days

Good point. I have anticipated this by ONLY processing Bug issues -> see PR description:
> Labels a Bug issue stale after 60 days
Closes a stale Bug issue after 7 days

KennethEnevoldsen:
> Closes stale PRs and issues automatically. How does it add work?

e.g. It closes a PR as stale even though it was solved or it marks stuff as stale even though it is ongoing, but might take some time.

isaac-chung:
> > Closes stale PRs and issues automatically. How does it add work?
> 
> e.g. It closes a PR as stale even though it was solved 

Where do you see that?

> or it marks stuff as stale even though it is ongoing, but might take some time.

Any update (update/comment) can reset the stale idle time on the issues/PRs.

.github/workflows/stale_pr.yml
```diff
@@ -0,0 +1,20 @@
...
+      actions: write
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v9.1.0
+        with:
+          stale-pr-label: 'stale'
+          days-before-pr-stale: 14
+          days-before-pr-close: 7
+          any-of-issue-labels: 'Bug'
```
Samoed: Currently we don't use label `bug`, but type `bug`

isaac-chung: Good eye. There are some options:
1. Most specific: Add 'bug' label for existing bug type tasks. Add 'bug' label in issue template
2. Less specific: exclude issues with label 'new dataset' and 'new model'

Samoed: > Less specific: exclude issues with label 'new dataset' and 'new model'

They're not bugs and won't be affected by this I think

isaac-chung: Do you prefer option 1? It's not clear to me from your comment.

Samoed: We can use option 1. I meant that `new datsaet` and `new models` don't have type bug.

isaac-chung: Added label to issue template.

isaac-chung: > any-of-issue-labels: 'Bug'

So actually this will not process any issues right now, which is a nice safe default.

isaac-chung: I'll add the bug labels to bug type issues after this is merged.

isaac-chung:
Thanks both for your feedback. I'd suggest us to try a few cycles and see how we like it, then update this workflow as we see fit. 

I'll proceed to merge later today, and label all bug type issues with the bug label. Let me know if you have more concerns on what we've discussed so far.

KennethEnevoldsen:
I really don't think we should merge this before we agree on the setup.

> bug type issues with the bug label.

Why do we want to have both label and type?

Here are a few pointers:

1) Confirmed bugs should never be closed unless they have been resolved (this is important documentation)
2) There is plenty of stuff that we can actively close (e.g. needs-more-infomation can be closed if info is not provided). So maybe these are good places to start?
3) If there are cases we are unsure we can start by only marking and not actively closing
4) A stale issue/PR costs us almost nothing. It only takes up time if we have to consider it

Having used this workflow before, I must say I wasn't pleased with it (I suspect 4 is a good reason why). I am more than happy to try it out again, but perhaps we can scope it down and then scale it up if it works well?

isaac-chung:
`needs-more-infomation` is a good call! Let's start with that. Let's start small yeah 👍

Edit: Updated. Other than needs-more-infomation, anything else is an obvious choice? repro?

KennethEnevoldsen:
sorry for the late response here, repro might also be a good one (unlabelled might also be a good one)

isaac-chung:
I can add repro, and see how to add unlabeled ones as well.

PR #3079, opened 2025-08-26
fix: Improving validate_task_to_prompt_name logs and error messages

RyanMullins:
## Problem

As released, the `Wrapper.validate_task_to_prompt_name()` function raises a `KeyError` if it encounters any key in the prompts dictionary that doesn't match a specific formatting requirement, even if the provided prompts dictionary contained _some_ valid keys. 

In most cases, `Wrapper` subclasses would pass this `KeyError` through and cause the MTEB wrapper initialization or evaluation run to fail.

The `SentenceTransformerWrapper` behaved differently in that it would catch the error raised by `Wrapper.validate_task_to_prompt_name()` and set the model prompts to `None`, allowing the evaluation to run but without any of the prompts that would make the model perform as expected. 

The `ColBERTWrapper` subclass was trying to catch a `ValueError` that did not seem like it would ever come up.

The problems documented above were observed when testing a `SentenceTransformer` model that included some custom task names in its `prompts` configuration. Sentence Transformers does not place restrictions on the values of keys in the `prompts` dictionary, so: 

* Having task names that are unknown to MTEB should be acceptable for the model given the Sentence Transformers best practices; and
* Ideally MTEB would use the prompts it does recognize instead of refusing the use any prompts if it encounters one unrecognized prompt. 

## Proposed Solution

I propose refactoring `Wrapper.validate_task_to_prompt_name()` to support options for:

1.  Raising a `KeyError`, for backward compatibility; or 
1.  Filtering to the valid prompts for MTEB and reporting the invalid prompts without raising.

In either case of the above, `Wrapper.validate_task_to_prompt_name()` should record the entire set of problems with the keys being validated, so that users can see and address all issues at once instead of having to iteratively run the configured model prompts through `Wrapper.validate_task_to_prompt_name()` to fix them all.

This PR: 

* Refactors `Wrapper.validate_task_to_prompt_name()` to address the above
* Refactors most `Wrapper` subclasses to 1) move a `None`-ness check inside the initializer function to the `Wrapper.validate_task_to_prompt_name()` function and 2) use the "raises" variant of the refactored `Wrapper.validate_task_to_prompt_name()` function.
* Refactors the `SentenceTransformerWrapper` class to use the "filters and reports" variant of the refactored `Wrapper.validate_task_to_prompt_name()` function.
* Updates existing unit tests as appropriate
* Adds unit tests for the "filters and reports" variant of the `Wrapper.validate_task_to_prompt_name()` function.

### DIFF
--- a/mteb/leaderboard/benchmark_selector.py
+++ b/mteb/leaderboard/benchmark_selector.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from dataclasses import dataclass
 
 import gradio as gr
-from build.lib.mteb.benchmarks.benchmarks import MTEB_multilingual
 
 import mteb
 from mteb import Benchmark
--- a/mteb/leaderboard/benchmark_selector.py
+++ b/mteb/leaderboard/benchmark_selector.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 
 import gradio as gr
+from build.lib.mteb.benchmarks.benchmarks import MTEB_multilingual
 
 import mteb
 from mteb import Benchmark
--- a/mteb/models/sentence_transformer_wrapper.py
+++ b/mteb/models/sentence_transformer_wrapper.py
@@ -38,29 +38,33 @@ class SentenceTransformerWrapper(Wrapper):
         else:
             self.model = model
 
-        if (
-            model_prompts is None
-            and hasattr(self.model, "prompts")
-            and len(self.model.prompts) > 0
-        ):
-            try:
-                model_prompts = self.validate_task_to_prompt_name(self.model.prompts)
-
-                if (
-                    len(self.model.prompts) == 2
-                    and self.model.prompts.get("query", "") == ""
-                    and self.model.prompts.get("document", "") == ""
-                ):
-                    model_prompts = None
-            except KeyError:
-                model_prompts = None
-                logger.warning(
-                    "Model prompts are not in the expected format. Ignoring them."
-                )
-        elif model_prompts is not None and hasattr(self.model, "prompts"):
-            logger.info(f"Model prompts will be overwritten with {model_prompts}")
+        if (built_in_prompts := getattr(self.model, "prompts", None)) and not model_prompts:
+            model_prompts = built_in_prompts
+        elif model_prompts and built_in_prompts:
+            logger.warning(f"Model prompts will be overwritten with {model_prompts}")
             self.model.prompts = model_prompts
-        self.model_prompts = self.validate_task_to_prompt_name(model_prompts)
+
+        self.model_prompts, invalid_prompts = self.validate_task_to_prompt_name(
+            model_prompts, raise_for_invalid_keys=False
+        )
+
+        if invalid_prompts:
+            logger.warning(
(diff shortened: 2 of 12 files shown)


mteb/models/sentence_transformer_wrapper.py
```diff
@@ -39,28 +39,35 @@ def __init__(
...
+
+        self.model_prompts, invalid_prompts = self.validate_task_to_prompt_name(
+            model_prompts, raise_for_invalid_keys=False
+        )
+
+        if invalid_prompts:
+            logger.warning(
+                "Some prompts are not in the expected format and will be ignored. Problems:\n\n"
+                f"{'\n'.join(invalid_prompts)}"
+            )
```
Samoed: I think this can be still in one line. I think it's a bit unexpected, when logs are multiline

mteb/models/wrapper.py
```diff
@@ -65,29 +66,95 @@ def get_prompt_name(
...
 
     @staticmethod
+    @overload
     def validate_task_to_prompt_name(
-        task_to_prompt_name: dict[str, str] | None,
-    ) -> dict[str, str] | None:
-        if task_to_prompt_name is None:
-            return task_to_prompt_name
+        task_to_prompt: dict[str, str] | None,
+        raise_for_invalid_keys: Literal[True] = True,
```
Samoed, line 72: Nice hack, but I'm not sure about it, because it's a bit tricky

RyanMullins: Yeah... makes me really wish Python had a better typing system. Open to suggestions.

mteb/models/sentence_transformer_wrapper.py
RyanMullins: Fixed.

tests/test_reproducible_workflow.py
```diff
@@ -67,11 +74,43 @@ def test_validate_task_to_prompt_name(task_name: str | mteb.AbsTask):
...
+    ],
+)
+def test_validate_task_to_prompt_name_fails_and_raises(
+    task_prompt_dict: dict[str, str],
+):
     with pytest.raises(KeyError):
-        Wrapper.validate_task_to_prompt_name(
-            {"task_name": "prompt_name", "task_name-query": "prompt_name"}
-        )
+        Wrapper.validate_task_to_prompt_name(task_prompt_dict)
```
Samoed, line 97: Can you add test for `None` input prompts?

Samoed (commented):
Overall I think this is good changes. Thanks!

tests/test_reproducible_workflow.py
RyanMullins: Done

mteb/leaderboard/benchmark_selector.py
```diff
@@ -3,6 +3,7 @@
 from dataclasses import dataclass
 
 import gradio as gr
+from build.lib.mteb.benchmarks.benchmarks import MTEB_multilingual
```
Samoed:
```suggestion
```

KennethEnevoldsen (approved):
I agree with @Samoed here I think this is a good change.

PR #3081, opened 2025-08-27
mdbr-leaf models

robin-vjc:
This PR submits the new mdbr-leaf models.

- [x] I have ensured that my model can be loaded using `mteb.get_model(model_name, revision)` and `mteb.get_model_meta(model_name, revision)`


mteb/models/mdbr_models.py
```diff
@@ -0,0 +1,65 @@
...
+    max_tokens=512,
+    embed_dim=768,
+    license="apache-2.0",
+    reference="https://huggingface.co/MongoDB/mdbr-leaf-ir",
+    similarity_fn_name="cosine",
+    use_instructions=True,
+    adapted_from="nreimers/MiniLM-L6-H384-uncased",
+    superseded_by=None,
+    public_training_code=None,
+    public_training_data=None,
```
Samoed, line 44: Can you add training datasets?

robin-vjc: I have updated the PR.

We had already included all the datasets used to train the models we distil from (`arctic-embed-m-v1.5` and `mxbai-embed-large-v1` respectively).

mteb/models/mdbr_models.py
```diff
@@ -0,0 +1,74 @@
...
+from mteb.models.arctic_models import arctic_v1_training_datasets
+from mteb.models.mxbai_models import mixedbread_training_data
+
+model_prompts = {"query": "Represent this sentence for searching relevant passages: "}
+
+LEAF_TRAINING_DATASETS = {
+    "AmazonQA": ["train"],
+    # FineWeb
+    # CC-News
+    # LoTTE
```
Samoed: We have `LoTTE` as a task, but only in `v2` branch https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/tasks/Retrieval/eng/LoTTERetrieval.py
```suggestion
    "LoTTE": ["train"],
```

robin-vjc: Yes happy to add it. 

We have used both `dev` and `test` of `pooled` data (see https://ir-datasets.com/lotte.html). 
Looking at the code you have linked, it seems like it should be
```
"LoTTE": ["dev", "test"],
```
instead of "train". Is that right?

Samoed: Yes

robin-vjc: submitted a fix

PR #3082, opened 2025-08-27
model: add Taichu mRerank

minmummax:
- [x ] I have filled out the ModelMeta object to the extent possible
- [ x] I have ensured that my model can be loaded using
  - [x ] `mteb.get_model(model_name, revision)` and
  - [ x] `mteb.get_model_meta(model_name, revision)`
- [ x] I have tested the implementation works on a representative set of tasks.
- [ x] The model is public, i.e. is available either as an API or the wieght are publicly avaiable to download

### DIFF
--- a/mteb/models/taichu_mReranker_models.py
+++ b/mteb/models/taichu_mReranker_models.py
@@ -39,7 +39,7 @@ def taichu_multimodal_rerank(query=None, candidate=None):
         "query": query,
         "candidate": candidate,
     }
-    
+
     try:
         response = requests.post(url=api_url, headers=headers, json=payload, timeout=10)
 
@@ -58,15 +58,13 @@ def taichu_multimodal_rerank(query=None, candidate=None):
     return None
 
 
-class TaiChuReranker():
-    name: str = "TaiChu"
-
+class TaiChuReranker:
     def __init__(
         self,
         model_name_or_path="ZiDongTaiChu/Taichu-mReranker-v1.0",
         max_tokens: int = 32000,
         **kwargs,
-    ):       
+    ):
         self._model_name = model_name_or_path
 
     def predict(self, text: str, image: List[Image.Image], **kwargs):
@@ -80,19 +78,9 @@ class TaiChuReranker():
         """
         image_base64_list = [pil_to_base64(image) for image in image]
 
-        query = [
-            {
-                "type": "text",
-                "text": text
-            }
-        ]
+        query = [{"type": "text", "text": text}]
         candidate = [
-            [
-                {
-                    "type": "image",
-                    "image": image_base64
-                }
-            ]
+            [{"type": "image", "image": image_base64}]
             for image_base64 in image_base64_list
         ]
         result_response = taichu_multimodal_rerank(query, candidate)
@@ -121,7 +109,7 @@ TRAINING_DATA = {
 Taichu_mReranker_v1_0 = ModelMeta(
     name="ZiDongTaiChu/Taichu-mReranker-v1.0",
     revision="1",
-    release_date="2025-08-18",
+    release_date="2025-08-26",
     languages=[
         "eng-Latn",
         "zho-Hans",
@@ -143,5 +131,5 @@ Taichu_mReranker_v1_0 = ModelMeta(
     public_training_code=None,
     public_training_data=None,
     is_cross_encoder=True,
-
+    modalities=["text", "image"],
 )
--- a/mteb/models/taichu_mReranker_models.py
+++ b/mteb/models/taichu_mReranker_models.py
@@ -1,0 +1,147 @@
+from __future__ import annotations
+
+import base64
+import logging
+import os
+
+from io import BytesIO
+from typing import Any, List
+from functools import partial
+
+import numpy as np
+import requests
+import torch
+from PIL import Image
+
+from mteb.model_meta import ModelMeta
+
+logger = logging.getLogger(__name__)
+
+
+def pil_to_base64(image, format="jpeg"):
+    buffer = BytesIO()
+    image.save(buffer, format=format)
+    img_bytes = buffer.getvalue()
+    encoded_bytes = base64.b64encode(img_bytes)
+    return encoded_bytes.decode("utf-8")
+
+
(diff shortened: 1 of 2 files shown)


mteb/models/taichu_mReranker_models.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __init__(
+        self,
+        model_name_or_path="ZiDongTaiChu/Taichu-mReranker-v1.0",
+        max_tokens: int = 32000,
+        **kwargs,
+    ):       
+        self._model_name = model_name_or_path
+
+    def predict(self, text: str, image: List[Image.Image], **kwargs):
```
Samoed, line 70: I think that Any2Any reranking tasks don't support cross encoders, but we're working on `v2` version where they will be supporter. Cc @isaac-chung

mteb/models/taichu_mReranker_models.py
```diff
@@ -0,0 +1,135 @@
...
+    loader=partial(
+        TaiChuReranker,
+        model_name_or_path="ZiDongTaiChu/Taichu-mReranker-v1.0",
+        max_tokens=32768,
+    ),
+    max_tokens=32768,
+    open_weights=False,
+    n_parameters=None,
+    memory_usage_mb=None,
+    embed_dim=None,
```
Samoed, line 128: Can you add embeding dimensionals?

minmummax: Our model is a cross-encoder model. I think I should just remove this key

isaac-chung (dismissed):
Looks good. @Samoed do you have any additional comments?

Samoed:
@isaac-chung I don't think it's working. I tried to run tests with this mock model and it failed with error `AttributeError: 'MockMultimodalReranker' object has no attribute 'get_image_embeddings'`. `MIEB` haven't integrated `predict` function and I don't understand how this model implementation is working

```python
class MockMultimodalReranker:
    mteb_model_meta = ModelMeta(
        name="mock/MockMocoModel",
        languages=["eng-Latn"],
        revision="7d091cd70772c5c0ecf7f00b5f12ca609a99d69d",
        release_date="2024-01-01",
        modalities=["image", "text"],
        n_parameters=86_600_000,
        memory_usage_mb=330,
        max_tokens=None,
        embed_dim=768,
        license=None,
        open_weights=True,
        public_training_code=None,
        public_training_data=None,
        framework=["PyTorch"],
        reference="https://github.com/facebookresearch/moco-v3",
        similarity_fn_name=None,
        use_instructions=False,
        training_datasets=None,
    )

    def __init__(
        self,
        **kwargs,
    ):
        pass

    def predict(self, text: str, image: list[Image], **kwargs):
        raise Exception
        return 1
```

Samoed (changes requested):
@isaac-chung I don't think it's working. I tried to run tasks with this mock model and it failed with error `AttributeError: 'MockMultimodalReranker' object has no attribute 'get_image_embeddings'`. `MIEB` haven't integrated `predict` function and I don't understand how this model implementation is working

```python
from PIL.Image import Image

import mteb
from mteb import ModelMeta


class MockMultimodalReranker:
    mteb_model_meta = ModelMeta(
        name="mock/MockMocoModel",
        languages=["eng-Latn"],
        revision="7d091cd70772c5c0ecf7f00b5f12ca609a99d69d",
        release_date="2024-01-01",
        modalities=["image"],
        n_parameters=86_600_000,
        memory_usage_mb=330,
        max_tokens=None,
        embed_dim=768,
        license=None,
        open_weights=True,
        public_training_code=None,
        public_training_data=None,
        framework=["PyTorch"],
        reference="https://github.com/facebookresearch/moco-v3",
        similarity_fn_name=None,
        use_instructions=False,
        training_datasets=None,
    )

    def __init__(
        self,
        **kwargs,
    ):
        pass

    def predict(self, text: str, image: list[Image], **kwargs):
        raise Exception
        return 1


evaluator = mteb.MTEB(
    tasks=mteb.get_tasks(tasks=["BLINKIT2IMultiChoice"])
)
evaluator.run(MockMultimodalReranker())
```

isaac-chung:
What's the confusion? I see submitted results on Vidore tasks for this model.

isaac-chung:
@Samoed it seems like `is_cross_encoder=True` is missing from your example code's `ModelMeta` to make it a cross encoder model.

Samoed:
`is_cross_encoder=True` don't change anything. I get same error. We don't check it when running models 

> What's the confusion? I see submitted results on Vidore tasks for this model.

This is an API-only model with signature that does not match either `Encoder` or `ImageEncoder`. I cannot find that `Any2AnyRetrievalTask` uses the `predict` method, so I am unsure how the results were produced. I tried to run multiple times model with the same signature, but I just didn't get any results

isaac-chung:
I see. This PR is also missing updates to overview.py (I just pushed changes). And running this mteb command: `mteb run -m ZiDongTaiChu/Taichu-mReranker-v1.0 -t VidoreArxivQARetrieval` yields this error:
```
AttributeError: 'TaiChuReranker' object has no attribute 'get_text_embeddings'
```

@minmummax since we are unable to reproduce the results, we cannot accept the PR as is. Please make sure that the results are reproducible by running the aforementioned mteb command. Please also share how the current results were obtained in detail.

minmummax:
> I see. This PR is also missing updates to overview.py (I just pushed changes). And running this mteb command: `mteb run -m ZiDongTaiChu/Taichu-mReranker-v1.0 -t VidoreArxivQARetrieval` yields this error:
> 
> ```
> AttributeError: 'TaiChuReranker' object has no attribute 'get_text_embeddings'
> ```
> 
> @minmummax since we are unable to reproduce the results, we cannot accept the PR as is. Please make sure that the results are reproducible by running the aforementioned mteb command. Please also share how the current results were obtained in detail.

OK, thanks. I will make this reproducible by running this command. This model is a multimodal reranking model (Cross-encoder Architecture )like[ jina-rerank-m0](https://huggingface.co/jinaai/jina-reranker-m0). Our results are reranked based on an Embedding model, which recalls top 25 candidates, then we use our model to rerank those candidates. I will make it detailed on the pr information。

Issue #2602, opened 2025-04-30
Move developer dependencies to groups PEP 735


Samoed:
`pip 25.1` supports dependency groups. I think we can move test and lint dependencies to their own groups. 

`uv` support https://github.com/astral-sh/uv/issues/8090
PEP https://peps.python.org/pep-0735/

PR #3088, opened 2025-08-28
Move dev to dependency groups

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/2602

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,16 +53,6 @@ mteb = "mteb.cli:main"
 
 [project.optional-dependencies]
 image = ["torchvision>0.2.1"]
-dev = [
-"ruff==0.11.13", # locked so we don't get PRs which fail only due to a lint update
-"pytest>=8.3.4",
-"pytest-xdist>=3.6.1",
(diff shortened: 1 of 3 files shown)


KennethEnevoldsen:
I believe this will break results:

https://github.com/embeddings-benchmark/results/blob/2369024b6fbe154d83ab89ef0de6efe6829ceec0/pyproject.toml#L7

but we should just add these to the results repo.

pyproject.toml
```diff
@@ -98,8 +88,26 @@ xet = ["huggingface_hub>=0.32.0"]
...
+    "pytest>=8.3.4",
+    "pytest-xdist>=3.6.1",
+    "pytest-coverage>=0.0",
+    "pytest-rerunfailures>=15.0",
+    "iso639>=0.1.4", # used for tests/scripts/test_generate_model_meta.py
+]
+dev = [
+    {include-group = "lint"},
+    {include-group = "test"},
+]
```
KennethEnevoldsen, line 107: Ahh, that is clean!

PR #3089, opened 2025-08-28
fix: Add dedicated display for RTEB benchmark results

q275343119:
## 📌 Live Demo
A working preview of these changes can be viewed here:
[**https://huggingface.co/spaces/SmileXing/leaderboard**](https://huggingface.co/spaces/SmileXing/leaderboard)

---

## Changes Proposed

This Draft PR introduces a dedicated display for the RTEB benchmark results. The changes include:

1.  **Sidebar Addition:** Added a dedicated "RTEB" option to the benchmarks sidebar selector.
2.  **Data Filtering:** Modified the `summary` and `performance per task` modules to exclusively use and display data from the RTEB benchmark.
3.  **Field Removal:**
    *   Removed the `zero-shot` field from the summary table.
    *   Consequently, removed the `zero-shot` filter option from the advanced model filters.
4.  **Ranking Logic:** Updated the ranking logic in the summary table to be based on `mean(Task)` in descending order.
5.  **Data Source:** The results data is now pulled from the [`embedding-benchmark/results`](https://github.com/embedding-benchmark/results) repository, specifically the [`feat-use-rteb-results`](https://github.com/embedding-benchmark/results/tree/feat-use-rteb-results) branch.

---

**Note:** This is a work in progress (Draft PR) to showcase the initial implementation for discussion and feedback.


mteb/benchmarks/rteb_benchmark.py
```diff
@@ -0,0 +1,192 @@
+# RTEB Benchmarks - Retrieval Embedding Benchmark
+from __future__ import annotations
+
+from mteb.benchmarks.benchmark import RTEBBenchmark
+from mteb.overview import get_tasks
+
+RTEB_MODELS = ['voyageai/voyage-3.5', 'sentence-transformers/all-mpnet-base-v2', 'BAAI/bge-m3',
```
Samoed: I don't think that we want always select same models

KennethEnevoldsen:
Hi @q275343119, there are several changes here that could require some discussion. I think it would be easier to discuss them separately (so that a single one doesn't block changes).

I think this is true for 2, 3, 4 (5 is a given)

for 1) it might be a possibility to set it as a flag (that way we can inspect it without adding it to main before it is ready. WDYT?

mteb/benchmarks/rteb_benchmark.py
q275343119: You're right, this isn't the ideal approach. Currently, I don't have a method to determine whether a model belongs to RTEB, so I used a hardcoded solution for now. Once the criteria are more defined, I will refactor this into a more elegant solution.

q275343119: You're right, this isn't the ideal approach. Currently, I don't have a method to determine whether a model belongs to RTEB, so I used a hardcoded solution for now. Once the criteria are more defined, I will refactor this into a more elegant solution.

q275343119:
> Hi @q275343119, there are several changes here that could require some discussion. I think it would be easier to discuss them separately (so that a single one doesn't block changes).
> 
> I think this is true for 2, 3, 4 (5 is a given)
> 
> for 1) it might be a possibility to set it as a flag (that way we can inspect it without adding it to main before it is ready. WDYT?

Thanks a lot for the suggestion! I’ll add a flag so that the RTEB-related features can be toggled and won’t affect the main display by default.
I’ll also make some adjustments on top of the latest branch, including reverting the ranking logic (as this will be handled later by @fzoll).
In addition, I’ll open a separate issue to discuss the changes (filtering, zero-shot column, ranking, etc.) so that we can agree on the details before creating new draft PRs.

q275343119:
> > Hi @q275343119, there are several changes here that could require some discussion. I think it would be easier to discuss them separately (so that a single one doesn't block changes).
> > I think this is true for 2, 3, 4 (5 is a given)
> > for 1) it might be a possibility to set it as a flag (that way we can inspect it without adding it to main before it is ready. WDYT?
> 
> Thanks a lot for the suggestion! I’ll add a flag so that the RTEB-related features can be toggled and won’t affect the main display by default. I’ll also make some adjustments on top of the latest branch, including reverting the ranking logic (as this will be handled later by @fzoll). In addition, I’ll open a separate issue to discuss the changes (filtering, zero-shot column, ranking, etc.) so that we can agree on the details before creating new draft PRs.

I have opened an issue to discuss the RTEB-specific changes:
https://github.com/embeddings-benchmark/mteb/issues/3095

.github/workflows/leaderboard_build.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
       - name: Install dependencies (incl. leaderboard extra)
         run: |
-          pip install ".[dev,leaderboard]"
+          pip install ".[leaderboard]" --group dev
```
Samoed: I think you've merged commit incorrectly and all commits displayed here somehow

q275343119: Yes, I noticed there are many duplicate commits. I’ll try to fix this and let you know once it’s adjusted

q275343119:
I add a new startup argument `--show_rteb` to control the visibility of branch-specific changes.
we can start the project using the following command:
```bash
python -m mteb.leaderboard.app --show_rteb
```

mteb/benchmarks/benchmark.py
```diff
@@ -72,3 +73,24 @@ def load_results(
         results = base_results.select_tasks(self.tasks)
         self.results_cache[base_results] = results
         return results
+
+
+@dataclass
+class RTEBBenchmark(Benchmark):
```
KennethEnevoldsen: Hmm, any reason for having this as a separate class?

mteb/leaderboard/table.py
```diff
@@ -344,3 +344,192 @@ def create_tables(
         joint_table, per_task, score_columns, column_types
     )
     return summary_table, per_task_table
+
+def create_rteb_tables(
```
KennethEnevoldsen: Hmm I understand that you did it this way, but I think I would do it differently to make it consistent across benchmarks. This also makes the approach more expandable in the future:

So my thoughts now are to move it to the benchmark object to have it as:

```
summary_table = Benchmark.create_summary_table() # default function but can be overwritten
```

We should be able to do this in a separate PR without RTEB benchmark (to see that everything works as intended), then adding customization for RTEB is simply a matter of replacing the `create_summary_table` function. I want the per_task_table to be the same across benchmarks.

The long-context retrieval or BRIGHT is a good test case for custom summary tables. 

Let me know what you think

mteb/benchmarks/benchmark.py
q275343119: As discussed with @fzliu, I've extended the Benchmark class and modified the `load_results` method to show specific models when RTEB benchmarks are selected. This change implements the requirement for model-specific result filtering.

This can be removed if we decide not to differentiate models in the final implementation.

mteb/leaderboard/table.py
q275343119: > So my thoughts now are to move it to the benchmark object to have it as:
> 
> ```
> summary_table = Benchmark.create_summary_table() # default function but can be overwritten
> ```
That’s a great idea！！！！！

> Let me know what you think

I have a couple of follow-up questions:

* Previously, `create_tables()` returned both `summary_table` and `per_task_table`. With the new approach, should I generate `summary_table` via `benchmark.create_summary_table() `and `per_task_table` via `create_task_table()` separately?

* In [#3096](https://github.com/embeddings-benchmark/mteb/discussions/3096)
 the discussion around `rank` and `zero-shot` still seems undecided. If later we need RTEB benchmarks to handle `rank` or `zero-shot` differently, would using inheritance to override create_summary_table() be an acceptable way to implement that customization(something like  RTEBBenchmark)?

KennethEnevoldsen: I would create them separately

Regarding how to implement changes, I think all of it can be implemented in `benchmark.create_summary_table()`, which I would probably just do as:

```
rteb_bench = Benchmark()
rteb_bench.create_summary_table = rteb_create_summary_table
```
That way, there is no confusion when people suddenly get an RTEBBenchmark instead of a Benchmark.

mteb/benchmarks/benchmark.py
KennethEnevoldsen: Hmm, do you want to curate the models that appear on RTEB, or do you want to filter them differently? (Is there anything wrong with the current filtering?)

q275343119:
Odd, there are still too many commits, please wait while I process them

mteb/leaderboard/table.py
q275343119: Got it. I'll create a separate PR for this great idea

Samoed:
I'm not sure about adding `RTEB (Retrieval)` as separate. Maybe we should convert it to MenuItem too? 

image

q275343119:
> I'm not sure about adding `RTEB (Retrieval)` as separate. Maybe we should convert it to MenuItem too?
> 
Sure, no problem. The main reasons are that RTEB is quite popular and has some key differences from MTEB, RTEB includes closed datasets. If you have any feedback or suggestions, we can discuss in https://github.com/embeddings-benchmark/mteb/discussions/3096. 
Also, would it be possible to reflect these changes in the UI as well? I mean remove the flag(`show_rteb`),or set default value to True
@KennethEnevoldsen

mteb/benchmarks/get_rteb_benchmark.py
KennethEnevoldsen, line 1: delete?

mteb/leaderboard/benchmark_selector.py
```diff
@@ -25,74 +27,164 @@ class MenuEntry:
...
+                       ),
+                       MenuEntry(
+                           "Miscellaneous",  # All of these are retrieval benchmarks
+                           mteb.get_benchmarks(
+                               [
+                                   "BEIR",
+                                   "BEIR-NL",
+                                   "NanoBEIR",
+                                   "BRIGHT",
+                                   "BRIGHT (long)",
```
KennethEnevoldsen: Some of these are retrieval - it seems a bit weird that they are not in the retrieval section

For now, it is not a problem, as it is just an optional flag for development, but it will be something we need to solve

mteb/leaderboard/benchmark_selector.py
```diff
@@ -25,74 +27,164 @@ class MenuEntry:
...
+                                   "RAR-b",
+                               ]
+                           ),
+                       ),
+                   ],
+    ),
+]
+
+RTEB_BENCHMARK_ENTRIES = [
+    MenuEntry(
```
KennethEnevoldsen, line 106: I would avoid the duplication on this item - just define and reuse across both (for now)

mteb/benchmarks/get_rteb_benchmark.py
q275343119: Yes, since RTEB no longer requires special filtering for models, benchmarks that belong to RTEB will now from Benchmark instead of RTEBBenchmark. As a result, the `get_rteb_benchmark.py` script is no longer necessary.

Originally, this script generated the BENCHMARK_REGISTRY using the following code:
```python
BENCHMARK_REGISTRY = {
    inst.name: inst
    for name, inst in benchmark_module.__dict__.items()
    if isinstance(inst, (Benchmark, RTEBBenchmark))
}
```
Now, this approach is no longer needed, so I have removed it.

mteb/leaderboard/benchmark_selector.py
q275343119: I'll modify this part. I will move the benchmarks that are exclusive to RTEB into `RTEB_BENCHMARK_ENTRIES`.

q275343119:
Hi @KennethEnevoldsen ,I've made adjustments to the RTEB_BENCHMARK_ENTRIES to remove duplicate Benchmarks. Could you please review it again to see if there are any other issues? If everything looks good, could we merge it into the main branch?

mteb/leaderboard/table.py
q275343119: Hi @KennethEnevoldsen ,here is the PR(https://github.com/embeddings-benchmark/mteb/pull/3126)

Samoed:
@q275343119 Can you run linter?

q275343119:
> @q275343119 Can you run linter?

Thanks for the reminder about running lint. I just realized we have a convenient lint command set up in the project. I'll make sure to use it from now on.

mteb/load_results/benchmark_results.py
```diff
@@ -730,7 +730,7 @@ def to_disk(self, path: Path | str) -> None:
     @classmethod
     def from_validated(cls, **data) -> BenchmarkResults:
         model_results = []
-        for model_res in data["model_results"]:
+        for model_res in data["model_results"][:1]:
```
Samoed:
```suggestion
        for model_res in data["model_results"]:
```

q275343119:
Hi @Samoed ,could you please review it again to see if there any other issues?

PR #3092, opened 2025-08-28
fix: add voyage quantization models

fzoll:
Adding quantization support


Samoed:
Why do you want to add quanyization and add it only for voyage?

fzoll:
@Samoed This way a model's performance becomes comparable with different quantization.
I added Voyage quantization as a first step, since I know this API best.

mteb/model_meta.py
```diff
@@ -120,6 +122,7 @@ class ModelMeta(BaseModel):
     superseded_by: str | None = None
     is_cross_encoder: bool | None = None
     modalities: list[MODALITIES] = ["text"]
+    quantization: QUANTIZATION_TYPES = "float"
```
KennethEnevoldsen: seems reasonable - did a check through current models

KennethEnevoldsen (commented):
This looks fine to me. 

It primarily adds the annotations (which we probably want to start annotating) and then implements it in a few models. I have made a few comment on the implementation though.

mteb/models/voyage_models.py
```diff
@@ -125,15 +125,55 @@ def _batched_encode(
                 batch.append(sentences[index])
                 index += 1
 
+            # Get quantization type from model metadata
+            output_dtype = getattr(self, "mteb_model_meta", None)
+            if output_dtype is not None:
+                output_dtype = output_dtype.quantization
```
KennethEnevoldsen: Instead of fetching this from the meta I would probably pass it in as a kwarg during initialization. That also make the difference clear in the modelmeta specification (otherwise they seem exactly the same)

mteb/models/voyage_models.py
```diff
@@ -166,6 +206,87 @@ def _batched_encode(
...
+    use_instructions=True,
+    training_datasets=VOYAGE_TRAINING_DATA,
+    public_training_code=None,
+    public_training_data=None,
+    quantization="binary",
+    adapted_from="voyageai/voyage-3.5",
+)
+
+voyage_3_5_ubinary = ModelMeta(
+    name="voyageai/voyage-3.5-ubinary",
```
KennethEnevoldsen: For other models we have done it like so: `"openai/text-embedding-3-small (embed_dim=512)"`

so would probably:
```suggestion
    name="voyageai/voyage-3.5 (quantization=ubinary)",
```

mteb/model_meta.py
```diff
@@ -42,6 +42,7 @@
     "ColPali",
 ]
 DISTANCE_METRICS = Literal["cosine", "max_sim", "dot"]
+QUANTIZATION_TYPES = Literal["float", "int8", "uint8", "binary", "ubinary"]
```
Samoed:
```suggestion
QUANTIZATION_TYPES = Literal["float32", "bf16", "int8", "uint8", "binary", "ubinary"]
```

mteb/model_meta.py
Samoed:
```suggestion
    quantization: QUANTIZATION_TYPES = "float32"
```

mteb/model_meta.py
Samoed: Can you add comment to ubinary, because this is not standart type

mteb/model_meta.py
Samoed: I think jasper/stella only is BF16

Samoed: Also this should be `None` for `API` based models. And I don't think that we want to add different annotations to voyage models, because I'm not sure if it different models or the same
```suggestion
    quantization: QUANTIZATION_TYPES | None = "float"
```

KennethEnevoldsen: Hmm I was thinking that this was the quantization of the embedding (so will always be known)

Samoed: Then we can rename it to `output_dtype` and also should add pytorch/numpy. Maybe sparse models have their own additional type

KennethEnevoldsen: I don't think numpy/pytorch is needed, but output_dtype might be a better name.

fzoll:
@KennethEnevoldsen @Samoed I just simplified this and moved the `output_dtype` to the loader from `ModelMeta`. This solution is very similar to how we dealt with the `Matryoshka`/different dimensions.

@KennethEnevoldsen, Probably this is what you mentioned in your comment to move to kwargs, am i correct?

mteb/model_meta.py
```diff
@@ -42,6 +42,7 @@
     "ColPali",
 ]
 DISTANCE_METRICS = Literal["cosine", "max_sim", "dot"]
+OUTPUT_DATA_TYPES = Literal["float32", "bf16", "int8", "uint8", "binary"]
```
Samoed: I think you can add this to `ModelMeta` sicnce you've specified literal

KennethEnevoldsen: I would just remove
```suggestion
```

mteb/models/voyage_models.py
```diff
@@ -201,6 +234,60 @@ def _batched_encode(
...
+    framework=["API"],
+    use_instructions=True,
+    training_datasets=VOYAGE_TRAINING_DATA,
+    public_training_code=None,
+    public_training_data=None,
+    adapted_from="voyageai/voyage-3.5",
+)
+
+voyage_3_5_binary = ModelMeta(
+    name="voyageai/voyage-3.5 (output_dtype=binary)",
```
KennethEnevoldsen, line 265: Yes this works - we might add it as a column as well in the future (and then we would need it in the metadata), but this works well for now

mteb/model_meta.py
fzoll: I just removed this as unused.

Issue #3097, opened 2025-08-29
CI not working


Samoed:
### Describe the bug

```
.........................................INTERNALERROR> def worker_internal_error(
INTERNALERROR>         self, node: WorkerController, formatted_error: str
INTERNALERROR>     ) -> None:
INTERNALERROR>         """
INTERNALERROR>         pytest_internalerror() was called on the worker.
INTERNALERROR>     
INTERNALERROR>         pytest_internalerror() arguments are an excinfo and an excrepr, which can't
INTERNALERROR>         be serialized, so we go with a poor man's solution of raising an exception
INTERNALERROR>         here ourselves using the formatted message.
INTERNALERROR>         """
INTERNALERROR>         self._active_nodes.remove(node)
INTERNALERROR>         try:
INTERNALERROR> >           assert False, formatted_error
INTERNALERROR> E           AssertionError: Traceback (most recent call last):
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.10.18/x64/lib/python3.10/site-packages/_pytest/main.py", line 289, in wrap_session
INTERNALERROR> E                 session.exitstatus = doit(config, session) or 0
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.10.18/x64/lib/python3.10/site-packages/_pytest/main.py", line 343, in _main
INTERNALERROR> E                 config.hook.pytest_runtestloop(session=session)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.10.18/x64/lib/python3.10/site-packages/pluggy/_hooks.py", line 512, in __call__
INTERNALERROR> E                 return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.10.18/x64/lib/python3.10/site-packages/pluggy/_manager.py", line 120, in _hookexec
INTERNALERROR> E                 return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.10.18/x64/lib/python3.10/site-packages/pluggy/_callers.py", line 167, in _multicall
INTERNALERROR> E                 raise exception
INTERNALERROR> E               File "/opt/hostedtoolcache/Python/3.10.18/x64/lib/python3.10/site-packages/pluggy/_callers.py", line 139, in _multicall
INTERNALERROR> E                 teardown.throw(exception)
[...]

PR #3098, opened 2025-08-29
CI: Set upper limit for xdist version 

Samoed:
Try to fix https://github.com/embeddings-benchmark/mteb/issues/3097

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,12 +21,12 @@ lint-check:
 
 test:
 	@echo "--- 🧪 Running tests ---"
-	pytest -n auto -m "not (test_datasets or leaderboard_stability)"
+	pytest -m "not (test_datasets or leaderboard_stability)"
 
 
 test-with-coverage:
 	@echo "--- 🧪 Running tests with coverage ---"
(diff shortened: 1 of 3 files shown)


KennethEnevoldsen:
(will just convert to drat for now)

Samoed:
@KennethEnevoldsen test are working now. I've removed `-n auto`, but I don't have ideas what borken, because pytest libs haven't been updated for month

isaac-chung:
Since we used to require deps like `pytest-xdist>=3.6.1`, there was no upperbound. It seems like newer versions are not compatible - the error is likely due to a test that fails in a way that cannot be properly serialized by xdist.

isaac-chung:
Will likely need a merge to v2 as well for PRs that merge into v2.

Samoed:
I think it was caused by `pytest-rerunfailures`, but I don't see anything super breaking in their release. https://github.com/pytest-dev/pytest-rerunfailures/blob/master/CHANGES.rst#160-2025-08-29, but issues related to our problem https://github.com/pytest-dev/pytest-rerunfailures/issues/303 https://github.com/pytest-dev/pytest-rerunfailures/issues/302

PR #3100, opened 2025-08-30
fix: Add @classmethod for @field_validators in TaskMetadata 

Samoed:
To use `field_validator` we need to add `classmethod` after it https://docs.pydantic.dev/latest/concepts/validators/#__tabbed_1_2

```python
class Model(BaseModel):
    number: int

    @field_validator('number', mode='after')  
    @classmethod
    def is_even(cls, value: int) -> int:
        if value % 2 == 1:
            raise ValueError(f'{value} is not an even number')
        return value  
```


isaac-chung (approved):
Nice. Does this mean that these were not validated, until now?

Samoed:
I'm not sure, because we don't have errors, but field validators should be with `classmethod` to work

PR #3101, opened 2025-08-30
Align task prompt dict with `PromptType`

Samoed:
Currenly `PromptType` contains `query` and `document`, but in `TaskMetadata` `PromptDict` contains `query` and `passage`. This might lead to incorrect checking. Now `PromptDict` would have this fields:

```python
from mteb.abstasks.TaskMetadata import PromptDict
print(PromptDict.__annotations__)
# {'query': <class 'str'>, 'document': <class 'str'>}
```

### DIFF
--- a/mteb/abstasks/TaskMetadata.py
+++ b/mteb/abstasks/TaskMetadata.py
@@ -164,7 +164,7 @@ METRIC_VALUE = Union[int, float, dict[str, Any]]
 
 
 PromptDict = TypedDict(
-    "PromptDict", dict.fromkeys(PromptType, str), total=False
+    "PromptDict", {prompt_type.value: str for prompt_type in PromptType}, total=False
 )
 """A dictionary containing the prompt used for the task.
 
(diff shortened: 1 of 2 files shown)


mteb/abstasks/TaskMetadata.py
```diff
@@ -181,9 +180,6 @@ class DescriptiveStatistics(TypedDict):
     pass
 
 
-METRIC_VALUE = Union[int, float, dict[str, Any]]
```
Samoed, line 184: Duplicated above

Samoed:
> This would be breaking if people try to access "passage"?

We're already changed from `document` to `passage`. I just fixed validation of `TaskMetadata.prompt` to use `document` instead of `pasage`, but we don't have any tasks

KennethEnevoldsen:
Right, let us just merge this in then, but generally, I would really prefer not to have breaking changes in fixes

Samoed:
> Right, let us just merge this in then, but generally, I would really prefer not to have breaking changes in fixes

Yes, I agree

PR #3103, opened 2025-08-30
ci: Dataset check on new PR

### LINKED ISSUE #3065: Dataset tests on new PR
Discover new datasets in the new PR, and run the HF dataset existence test on those datasets only.

isaac-chung:
Fixes #3065 

1. Checks for new datasets from changes files in the PR only (within mteb/tasks). 
2. Then run the dataset revision check on only the detected datasets. (see [here](https://github.com/embeddings-benchmark/mteb/actions/runs/17348084796/job/49250690297?pr=3103), only 1 dataset was run) 

------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,8 +54,7 @@ dataset-load-test:
 
 dataset-load-test-pr:
 	@echo "--- 🚀 Running dataset load test for PR ---"
-	python -m scripts.extract_datasets $(BASE_BRANCH)
-	pytest -m test_datasets
+	eval "$$(python -m scripts.extract_datasets $(BASE_BRANCH))" && pytest -m test_datasets
 
 leaderboard-build-test:
 	@echo "--- 🚀 Running leaderboard build test ---"
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ dataset-load-test:
 
 dataset-load-test-pr:
 	@echo "--- 🚀 Running dataset load test for PR ---"
-	python scripts/extract_datasets.py $(BASE_BRANCH)
+	python -m scripts.extract_datasets $(BASE_BRANCH)
 	pytest -m test_datasets
 
 leaderboard-build-test:
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,11 @@ dataset-load-test:
 	@echo "--- 🚀 Running dataset load test ---"
 	pytest -m test_datasets
 
+dataset-load-test-pr:
+	@echo "--- 🚀 Running dataset load test for PR ---"
+	python scripts/extract_datasets.py $(BASE_BRANCH)
+	pytest -m test_datasets
+
 leaderboard-build-test:
 	@echo "--- 🚀 Running leaderboard build test ---"
 	pytest -n auto -m leaderboard_stability
--- a/scripts/extract_datasets.py
+++ b/scripts/extract_datasets.py
@@ -114,6 +114,8 @@ def parse_args():
 
 if __name__ == "__main__":
     """
+    Extract datasets from changed task files compared to a base branch.
+
     Can pass in base branch as an argument. Defaults to 'main'.
     e.g. python -m scripts.extract_datasets mieb
     """
--- a/scripts/extract_datasets.py
+++ b/scripts/extract_datasets.py
@@ -62,8 +62,9 @@ def extract_datasets(files: list[str]) -> list[tuple[str, str]]:
             f"{path}:{revision}" for path, revision in unique_datasets
         )
         os.environ["CUSTOM_DATASET_REVISIONS"] = custom_revisions
-        logging.info(f"Set CUSTOM_DATASET_REVISIONS={custom_revisions}")
+        logging.debug(f"Set CUSTOM_DATASET_REVISIONS={custom_revisions}")
 
+        print(f'export CUSTOM_DATASET_REVISIONS="{custom_revisions}"')
     return unique_datasets
 
 
@@ -123,4 +124,4 @@ if __name__ == "__main__":
     changed_files = get_changed_files(base_branch, startswith="mteb/tasks/")
     dataset_tuples = extract_datasets(changed_files)
 
-    logging.info(f"Found {len(dataset_tuples)} unique datasets:")
+    logging.debug(f"Found {len(dataset_tuples)} unique datasets:")
--- a/scripts/extract_datasets.py
+++ b/scripts/extract_datasets.py
@@ -120,10 +120,7 @@ if __name__ == "__main__":
     args = parse_args()
 
     base_branch = args.base_branch
-    changed_files = get_changed_files(base_branch)
-    print(f"{changed_files=}")
+    changed_files = get_changed_files(base_branch, startswith="mteb/tasks/")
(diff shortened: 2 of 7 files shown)


scripts/extract_datasets.py
```diff
@@ -0,0 +1,127 @@
...
+    return None
+
+
+def extract_dataset_from_dict(dict_node: ast.Dict) -> tuple[str, str] | None:
+    """Extract path and revision from a dataset dictionary."""
+    path = None
+    revision = None
+
+    for key, value in zip(dict_node.keys, dict_node.values):
+        if isinstance(key, ast.Constant) and key.value == "path":
```
Samoed, line 85: Will this work?
```suggestion
        if isinstance(key, (ast.Constant, ast.Str)) and key.value == "path":
```

scripts/extract_datasets.py
```diff
@@ -0,0 +1,127 @@
...
+            if isinstance(value, ast.Constant):
+                path = value.value
```
Samoed, line 87: Maybe be move to parent if. Do we have cases when value is not a constant/string?

Samoed (approved):
Great!

scripts/extract_datasets.py
isaac-chung: I can take a look :)

tests/test_tasks/test_all_abstasks.py
```diff
@@ -47,6 +48,15 @@
     }
 )
 
+custom_revisions = os.getenv("CUSTOM_DATASET_REVISIONS")
```
KennethEnevoldsen, line 51: I would imagine this would be hard to debug in the future - Why is this needed?

scripts/extract_datasets.py
```diff
@@ -0,0 +1,127 @@
...
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    """
+    Can pass in base branch as an argument. Defaults to 'main'.
+    e.g. python -m scripts.extract_datasets mieb
+    """
+
+    args = parse_args()
```
KennethEnevoldsen: Let us add a bit more docs to this so that simply running `--help` should allow you to figure out what it does

scripts/extract_datasets.py
```diff
@@ -0,0 +1,127 @@
...
+
+    base_branch = args.base_branch
+    changed_files = get_changed_files(base_branch, startswith="mteb/tasks/")
+    dataset_tuples = extract_datasets(changed_files)
+
+    logging.debug(f"Found {len(dataset_tuples)} unique datasets:")
```
KennethEnevoldsen: ":" seems to suggest a list coming up?

scripts/extract_datasets.py
```diff
@@ -0,0 +1,127 @@
+from __future__ import annotations
+
+import argparse
+import ast
+import logging
+import os
+
+from .extract_model_names import get_changed_files
+
+logging.basicConfig(level=logging.INFO)
```
KennethEnevoldsen, line 10: move to `if __name__`?

scripts/extract_datasets.py
```diff
@@ -0,0 +1,127 @@
+from __future__ import annotations
+
+import argparse
+import ast
+import logging
+import os
+
+from .extract_model_names import get_changed_files
```
KennethEnevoldsen, line 8: I would probably just copy it over (no need to make the two dependent on eachother for a single function)

Makefile
```diff
@@ -52,6 +52,10 @@ dataset-load-test:
 	@echo "--- 🚀 Running dataset load test ---"
 	pytest -m test_datasets
 
+dataset-load-test-pr:
+	@echo "--- 🚀 Running dataset load test for PR ---"
+	eval "$$(python -m scripts.extract_datasets $(BASE_BRANCH))" && pytest -m test_datasets
```
KennethEnevoldsen, line 57: I would probably refactor to not use tests (`mteb.get_task(...).load_data()` in script)

KennethEnevoldsen (commented):
Ahh great!!

If we already load the data would it be too costly to just run a small model on it as well (it could be just a "random" encoder, but ideally a model2vec model).

tests/test_tasks/test_all_abstasks.py
isaac-chung: The models check relies on writing to a text file, which I thought was less reliable than this method of setting an environment variable.

scripts/extract_datasets.py
isaac-chung: Nit. Would prefer not to duplicate code.

tests/test_tasks/test_all_abstasks.py
KennethEnevoldsen: Could we move the loading directly to the script - I think that would be clearer when it fails

scripts/extract_datasets.py
isaac-chung: Done.

scripts/extract_datasets.py
isaac-chung: Done.

Makefile
isaac-chung: Where is that in this PR?

tests/test_tasks/test_all_abstasks.py
isaac-chung: > If we already load the data would it be too costly to just run a small model on it as well (it could be just a "random" encoder, but ideally a model2vec model).


So I think there's some confusion over what the test is. This is the test in question:
```
@pytest.mark.parametrize("dataset_revision", dataset_revisions)
def test_dataset_on_hf(dataset_revision: tuple[str, str]):
    repo_id, revision = dataset_revision
    try:
        huggingface_hub.dataset_info(repo_id, revision=revision)
    except (
        huggingface_hub.errors.RepositoryNotFoundError,
        huggingface_hub.errors.RevisionNotFoundError,
    ):
        assert False, f"Dataset {repo_id} - {revision} not available"
    except Exception as e:
        assert False, f"Dataset {repo_id} - {revision} failed with {e}"
```

Which AFAIK does not load any actual data from the dataset.

Makefile
isaac-chung: Maybe see [this comment](https://github.com/embeddings-benchmark/mteb/pull/3103#discussion_r2314445569)? I think there's some misunderstanding here over what the test is.

tests/test_tasks/test_all_abstasks.py
KennethEnevoldsen: Ahh, right, so it only does an existence check

KennethEnevoldsen (approved):

isaac-chung:
> Ahh great!!
> 
> If we already load the data would it be too costly to just run a small model on it as well (it could be just a "random" encoder, but ideally a model2vec model).

@KennethEnevoldsen @Samoed  just read this again. We could totally try running a random encoder + any small reference model to see if GH actions can handle it, if we haven't tried this yet.

Samoed:
We can, but some running audio and video tasks can take a lot of time. We also have suggestion to check model per batch_size, but I think this will be overkill https://github.com/embeddings-benchmark/mteb/pull/5217#discussion_r3809925326

PR #3108, opened 2025-09-01
Fix VocalSound split naming

### LINKED ISSUE #3106: [maeb][dev] 7 AudioClassification dataset issues
from my understanding `AudioClassification` is a linear probe task and therefore requires both a train and a test set? Currently the following 7 tasks fail to run or doesn't make sense because only one split presents.
```
"VocalSound",
"VoxCelebSA",
"VoxLingua107_Top10",
"VoxPopuliAccentID",
"MInDS14",
"VoxPopuliGenderID",
"VoxPopuliLanguageID"
```

if there's only `test` split, you would get an error that train set not found (e.g., `VocalSound`).
if there's only `train` split and you specify `train` as `eval_split` in the task metadata, you would get 100% accuracy because of training and testing both on the `train` set which doesn't make sense. (e.g., all tasks that use `mteb/voxpopuli-accent-mini`, including `VoxPopuliLanguageID`, `VoxPopuliGenderID`,`VoxPopuliAccentID` etc).

isaac-chung:
Fixes #3106 


isaac-chung:
Ah. Now we need to merge main to maeb.

cc @Samoed @KennethEnevoldsen

gowitheflow-1998 (approved):

PR #3120, opened 2025-09-02
[v2] Lots of docs updates

KennethEnevoldsen:
Lots of updates the the docs!

I will start slowly chopping away at these.

Currently going through the API section.


KennethEnevoldsen:
Great - there is quite a lot to start on, but I feel like this is already quite a solid boost for package transparency

isaac-chung:
Nice work!

PR #3122, opened 2025-09-03
Deprecated `load_results` and added an update page

KennethEnevoldsen:
- Deprecated `load_results`
- Added New in v2.0 page, which include "upgrading from v1" section
- Added usage.md to docs (could probably use some work, but that is another pr)
- added empty landing page (and renamed the old index.md)

Pushed current website to:
https://embeddings-benchmark.github.io/mteb/


mteb/load_results/load_results.py
```diff
@@ -36,6 +37,10 @@ def _model_name_and_revision(
     return model_name, revision
 
 
+@deprecated(
```
Samoed, line 45: Decorator was added only in 3.13 https://peps.python.org/pep-0702/

PR #3126, opened 2025-09-04
Refactor split create_tables into static Benchmark methods

q275343119:
I've decomposed the monolithic `create_tables` method into tow focused methods `create_summary_table` and `create_per_task_table`.
And both new methods are implemented as static methods on the Benchmark class.

### DIFF
--- a/mteb/benchmarks/benchmark.py
+++ b/mteb/benchmarks/benchmark.py
@@ -72,3 +72,17 @@ class Benchmark:
         results = base_results.select_tasks(self.tasks)
         self.results_cache[base_results] = results
         return results
+
+    @staticmethod
+    def create_summary_table(scores_long: list[dict], search_query: str | None = None):
+        """create_summary_table"""
+        # Avoid circular references
+        from mteb.leaderboard.table import create_summary_table
+        return create_summary_table(scores_long, search_query)
+
+    @staticmethod
+    def create_per_task_table(scores_long: list[dict], search_query: str | None = None):
+        """create_per_task_table"""
+        # Avoid circular references
+        from mteb.leaderboard.table import create_per_task_table
+        return create_per_task_table(scores_long, search_query)
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -23,7 +23,6 @@ from mteb.leaderboard.benchmark_selector import (
     make_selector,
 )
 from mteb.leaderboard.figures import performance_size_plot, radar_chart
-from mteb.leaderboard.table import create_tables
 from mteb.leaderboard.text_segments import ACKNOWLEDGEMENT, FAQ
 
 logger = logging.getLogger(__name__)
@@ -218,10 +217,10 @@ def get_leaderboard_app() -> gr.Blocks:
         max_model_size=MAX_MODEL_SIZE,
         zero_shot_setting="allow_all",
     )
+    default_filtered_scores = [entry for entry in default_scores if entry["model_name"] in filtered_models]
+    summary_table = default_benchmark.create_summary_table(default_filtered_scores)
+    per_task_table = default_benchmark.create_per_task_table(default_filtered_scores)
 
-    summary_table, per_task_table = create_tables(
-        [entry for entry in default_scores if entry["model_name"] in filtered_models]
-    )
     lang_select = gr.Dropdown(
         LANGUAGE,
         value=sorted(default_results.languages),
@@ -763,7 +762,8 @@ def get_leaderboard_app() -> gr.Blocks:
                     filtered_scores.append(entry)
             else:
                 filtered_scores = scores
-            summary, per_task = create_tables(filtered_scores)
+            summary = default_benchmark.create_summary_table(filtered_scores)
+            per_task = default_benchmark.create_per_task_table(filtered_scores)
             elapsed = time.time() - start_time
             logger.debug(f"update_tables callback: {elapsed}s")
             return summary, per_task
--- a/mteb/leaderboard/table.py
+++ b/mteb/leaderboard/table.py
@@ -332,7 +332,7 @@ def apply_styling(
 
 
 def create_tables(
-    scores_long: list[dict], search_query: str | None = None
+        scores_long: list[dict], search_query: str | None = None
 ) -> tuple[gr.DataFrame, gr.DataFrame]:
     result = scores_to_tables(scores_long, search_query)
     # dataframe with No Results is returned, so no need to apply styling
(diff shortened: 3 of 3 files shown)


KennethEnevoldsen:
Hi @q275343119 is this still a draft or is it ready for review?

q275343119:
> Hi @q275343119 is this still a draft or is it ready for review?

Yes, this PR is ready for review (not a draft).
The changes are reflected in [the leaderboard](https://huggingface.co/spaces/q275343119/leaderboard)
, though they don’t involve frontend adjustments, so the differences may not be immediately visible.

I’d appreciate your feedback on whether this implementation makes sense, as I’m not fully sure if it’s the best approach.

mteb/benchmarks/benchmark.py
```diff
@@ -72,3 +72,19 @@ def load_results(
...
+
+        return create_summary_table(scores_long, search_query)
+
+    @staticmethod
+    def create_per_task_table(scores_long: list[dict], search_query: str | None = None):
+        """create_per_task_table"""
+        # Avoid circular references
+        from mteb.leaderboard.table import create_per_task_table
+
+        return create_per_task_table(scores_long, search_query)
```
KennethEnevoldsen: Hmm, the circular import here is a bit odd. I would just move the code here (note that gradio isn't a required dependency in mteb)

KennethEnevoldsen: I would also make both of these private

mteb/leaderboard/table.py
```diff
@@ -344,3 +344,239 @@ def create_tables(
         joint_table, per_task, score_columns, column_types
     )
     return summary_table, per_task_table
+
+
+def _prepare_data(
+    scores_long: list[dict], search_query: str | None = None
+) -> tuple[pd.DataFrame, list[str]] | None:
+    """Prepare raw dataframe and filter invalid models.
```
KennethEnevoldsen:
```suggestion
    """Prepare raw dataframe and filter models without any scores.
```

mteb/leaderboard/table.py
```diff
@@ -344,3 +344,239 @@ def create_tables(
...
+    # setting model name column to markdown
+    column_types[1] = "markdown"
+    score_columns = ["Mean (Task)", "Mean (TaskType)", *mean_per_type.columns]
+
+    return apply_summary_styling(joint_table, score_columns, column_types)
+
+
+def create_per_task_table(
+    scores_long: list[dict], search_query: str | None = None
+) -> gr.DataFrame:
```
KennethEnevoldsen: I don't believe the search query was used, which I believe simplified things quite a bit

KennethEnevoldsen: you can probably remove _prepare_data

mteb/leaderboard/table.py
```diff
@@ -344,3 +344,239 @@ def create_tables(
...
+        per_task_style,
+        interactive=False,
+        pinned_columns=1,
+        show_fullscreen_button=True,
+        show_copy_button=True,
+        show_search="filter",
+    )
+
+
+def create_summary_table(
```
KennethEnevoldsen: I would probably just take BenchmarkResults as the input here.

I think you can even use the `BenchmarkResults.to_dataframe` attribute here.
related docs: https://github.com/embeddings-benchmark/mteb/blob/main/docs/usage/results.md

mteb/benchmarks/benchmark.py
KennethEnevoldsen: It would also be great if this just returns a dataframe, and styling is applied in the leaderboard. That way, certain columns are always consistently styled, and user can get a daframe in a working format from this.

q275343119:
> I think there is a few simplications to do:
> 
> 1. remove search query
> 2. take benchmark result as input
> 3. utilize existing data wrangling for BenchmarkResults (see below)
> 
> Should we also create a custom leaderboard e.g. for BRIGHT (for the different subsets) just to see how a custom imp. would look

Thanks a lot for the detailed feedback, @KennethEnevoldsen  🙏.
I’ll follow your suggestions — removing the unused search_query, restructuring the methods to take BenchmarkResults (and return clean DataFrames), addressing the circular import, and moving the styling logic to the leaderboard side.
I’ll update the PR in the coming days.

q275343119:
Hi @KennethEnevoldsen,I've updated the PR. Could you please take another look when you have time

mteb/benchmarks/benchmark.py
```diff
@@ -18,6 +25,252 @@
...
+    Returns a DataFrame with one row per model and one column per task.
+
+    Args:
+        benchmark_results: BenchmarkResults object containing model results
+
+    Returns:
+        DataFrame with per-task scores, ready for styling in the leaderboard
+    """
+    # Get scores in long format
+    scores_long = benchmark_results.get_scores(format="long")
```
KennethEnevoldsen: why not use the `.to_dataframe` method?

mteb/benchmarks/benchmark.py
```diff
@@ -72,3 +325,33 @@ def load_results(
...
+        """Create per-task table from BenchmarkResults.
+
+        This method can be overridden by subclasses to provide custom table generation logic.
+
+        Args:
+            benchmark_results: BenchmarkResults object containing model results (may be pre-filtered)
+
+        Returns:
+            DataFrame with per-task scores, ready for styling in the leaderboard
+        """
```
KennethEnevoldsen:
```suggestion
        """Create per-task table. Called by the leaderboard app."""
```

mteb/benchmarks/benchmark.py
```diff
@@ -72,3 +325,33 @@ def load_results(
...
+        """Create summary table from BenchmarkResults.
+
+        This method can be overridden by subclasses to provide custom table generation logic.
+
+        Args:
+            benchmark_results: BenchmarkResults object containing model results (may be pre-filtered)
+
+        Returns:
+            DataFrame with model summaries, ready for styling in the leaderboard
+        """
```
KennethEnevoldsen:
```suggestion
        """Create summary table. Called by the leaderboard app."""

```

mteb/benchmarks/benchmark.py
KennethEnevoldsen, line 1: Let us move this to `benchmark._create_table.py`

mteb/leaderboard/table.py
KennethEnevoldsen, line 1: I think you are missing the clean-up of this script - plenty of functions are no longer used.

mteb/benchmarks/benchmark.py
KennethEnevoldsen: (would prefer it)

mteb/benchmarks/benchmark.py
q275343119: I'm not sure about what should I do, so I just want to confirm — should I move the `create_table` logic into `benchmark/_create_table.py`, or is there something else you had in mind?

mteb/benchmarks/_create_table.py
```diff
@@ -0,0 +1,256 @@
...
+    return str(int(max_tokens))
+
+
+def _failsafe_get_model_meta(model_name):
+    try:
+        from mteb.models.overview import get_model_meta
+
+        return get_model_meta(model_name)
+    except Exception:
+        return None
```
q275343119, line 60: I noticed a strange issue after moving the code into `_create_table.py`.
Placing this line at the top:
```python
from mteb.models.overview import get_model_meta
```
causes a circular import, even though it works fine in `benchmark.py`.
For now, I applied lazy loading inside the function to avoid the problem.
Do you have some better different solution.

q275343119: Hi @KennethEnevoldsen ,could you please take a look,when you have time.

Samoed:
I'm not sure how custom tables would work with filters, because they could have missing columns and etc. Maybe we should also provide possible custom filters, but this requires to customize whole view then

q275343119:
> I'm not sure how custom tables would work with filters, because they could have missing columns and etc. Maybe we should also provide possible custom filters, but this requires to customize whole view then

From what I’ve seen in the current frontend implementation, the filters (e.g. taskType, domain, etc.) operate on the BenchmarkResult data first, and only then the table is rendered based on the filtered results. So as long as the filtered BenchmarkResult objects carry the standard fields, the generated tables shouldn’t be missing any required columns.

In the near future, we do plan to customize some table columns in the RTEB part — for example, replacing mean(Task) / mean(TaskType) with mean(public) / mean(private). That’s where the custom table logic would start to matter, but even then the core fields used by filters will remain available.

By the way, would it be possible to merge the PR today if everything looks good?

KennethEnevoldsen:
I will just double-check that the logic works, but otherwise, I think it looks good

KennethEnevoldsen:
Hmm processing time seems higher locally (typically faster than online):

Screenshot 2025-09-29 at 17 45 24

It might just have been a start-up thing 

edit: After a bit more testing I couldn't find any notable loss in performance, but we might consider rolling it back if we see notable effects in the leaderboard

mteb/leaderboard/app.py
```diff
@@ -751,19 +765,43 @@ def update_tables(
...
+                if entry["task_name"] not in tasks:
+                    continue
+                if (models_to_keep is not None) and (
+                    entry["model_name"] not in models_to_keep
+                ):
+                    continue
+                filtered_model_names.add(entry["model_name"])
+                filtered_task_names.add(entry["task_name"])
+
+            # Create filtered BenchmarkResults as required by Kenneth
```
KennethEnevoldsen, line 783:
```suggestion
```

KennethEnevoldsen: not sure I understand?

q275343119: > not sure I understand?

Maybe one possible reason is that the original create_table function returned both summary_table and per_task in a single pass.
In the updated version, these two DataFrames are computed separately, which introduces some duplicated processing and may cause the slowdown.

PR #3129, opened 2025-09-04
model: EmbeddingGemma 300M

RyanMullins:
Adds EmbeddingGemma 300M to MTEB.

* [x] I have filled out the `ModelMeta` object to the extent possible
* [x] I have ensured that my model can be loaded using `mteb.get_model(model_name, revision)` and `mteb.get_model_meta(model_name, revision)`
* [x] I have tested the implementation works on a representative set of tasks.
* [x] The model is public, i.e., is available either as an API or the weights are publicly available to download

cc @KennethEnevoldsen @Samoed


mteb/models/google_models.py
```diff
@@ -237,3 +237,23 @@ def encode(
...
+embedding_gemma_300m = ModelMeta(
+    name="google/embeddinggemma-300m",
+    languages=MULTILINGUAL_EVALUATED_LANGUAGES,
+    open_weights=True,
+    revision=None,
+    release_date="2025-09-04",
+    n_parameters=307_581_696,
+    embed_dim=768,
+    max_tokens=2048,
+    license=None,
```
Samoed: I think you can add your license `gemma`, but you also need to update available licences https://github.com/embeddings-benchmark/mteb/blob/647c8c3d54296d9b4a4384c0809bd958caeceea8/mteb/custom_validators.py#L20

mteb/models/google_models.py
```diff
@@ -237,3 +237,23 @@ def encode(
     public_training_data=None,
     training_datasets=GECKO_TRAINING_DATA,
 )
+
+embedding_gemma_300m = ModelMeta(
+    name="google/embeddinggemma-300m",
+    languages=MULTILINGUAL_EVALUATED_LANGUAGES,
+    open_weights=True,
+    revision=None,
```
Samoed:
```suggestion
    revision="64614b0b8b64f0c6c1e52b07e4e9a4e8fe4d2da2",
```

mteb/models/google_models.py
RyanMullins: Done. Thanks.

mteb/models/google_models.py
RyanMullins: Done. Thanks.

Samoed:
Does your model support images? I see `image_soft_token` in your tokenizer config

RyanMullins:
@Samoed no image support for this model, but the image token remains in the tokenizer because all Gemma 3 models use the same one.

Samoed (approved):

PR #3124, opened 2025-09-03
[v2] Split up usage.md for better readability 

KennethEnevoldsen:
- [x] Split up the usage.md for better overall readability
- [x] Added documentation for the CLI.
- [x] Added outline for starting page
- [x] Added installation guide

changes can be seen here:
https://embeddings-benchmark.github.io/mteb/

PR #3136, opened 2025-09-06
[v2]: Make the `evaluator` module private

KennethEnevoldsen:
- [x] rename `evaluators` to `_evaluators`
- [x] delete `evaluations` folder and move `MTEB.py` and `_evaluators` out


Samoed:
What is the source branch of this PR? I think this is too many changes

Samoed (approved):
I think this is good, but we need to change target branch to https://github.com/embeddings-benchmark/mteb/pull/3124

PR #3125, opened 2025-09-03
[v2] Refactor mieb retrieval

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/2435

I make retrieval tasks to work with `MIEB` `Retrieval` tasks

I've run some tasks with `openai/clip-vit-large-patch14`. For now, they're very different. I'll try to look what's the issue

| Task name | `PR` | `v2` |
| - | - | - |
| BLINKIT2IMultiChoice | 0.71891 | 0.71891 |
| BLINKIT2IRetrieval | 0.35821 | 0.35821 |
| BLINKIT2TMultiChoice | 0.40227 | 0.40227 |
| BLINKIT2TRetrieval | 0.09332 | 0.09332 |
| CVBenchCount | 0.03173 | 0.03173 |

PR #3137, opened 2025-09-06
[v2] Refactor multilabel classification

Samoed:
Unfiy Image and Text multilabel classification


mteb/abstasks/AbsTaskMultilabelClassification.py
```diff
@@ -26,14 +28,21 @@ def evaluate_classifier(
...
     y_test: np.ndarray,
-    classifier: ClassifierMixin,
+    classifier: SklearnClassifierProtocol,
 ) -> dict[str, float]:
-    classifier = clone(classifier)
+    classifier: SklearnClassifierProtocol = clone(classifier)
     classifier.fit(embeddings_train, y_train)
     y_pred = classifier.predict(embeddings_test)
     accuracy = classifier.score(embeddings_test, y_test)
+    if isinstance(classifier, MultiOutputClassifier):
```
Samoed, line 37: We don't have an evaluator for multlabel classification, but I can move logic from task to it

mteb/create_dataloaders.py
```diff
@@ -319,7 +319,7 @@ def create_dataloader(
 
     if "image" in task_metadata.modalities:
         return create_image_dataloader(
-            (dataset.select_columns(input_column).rename_column(input_column, "image")),
+            dataset,
```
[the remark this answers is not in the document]
isaac-chung: This may be relevant to memory usage?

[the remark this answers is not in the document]
Samoed: Yes, but I have moved this part to the task code. I don't think this is the best place for `select_columns` as we might have multimodal tasks, and with this part, only image features would be selected

mteb/abstasks/AbsTaskMultilabelClassification.py
KennethEnevoldsen: Yeah, this is something that both x-tabdeveloping and I have done (also the case for ClusteringFast). I would generally say that it is not needed to have an evaluator. Gives another layer, which can often lead to some odd implementations.

KennethEnevoldsen: but maybe I am wrong?

KennethEnevoldsen (approved):
loving it!

mteb/abstasks/AbsTaskMultilabelClassification.py
Samoed: This is only task without evaluator, and we can add it to have more consistency. Also, we have cases like `STS` where `_evaluate_subset` just call evaluator https://github.com/embeddings-benchmark/mteb/blob/7075fbb14c1d3e67f132c70268f47781bc5aabc9/mteb/abstasks/AbsTaskAnySTS.py#L80-L92

KennethEnevoldsen: `AbsTaskClusteringFast` also does not have an evaluator

Samoed: Hm, right. Do we want to add evaluator to classification then or we will leave it as is?

isaac-chung: I think it's fine without. No need to add it just for the sake of adding code.

Samoed:
I will merge this PR after https://github.com/embeddings-benchmark/mteb/pull/3125, because it depends on it and I don't want to mess retrival PR

Issue #3134, opened 2025-09-06
[v2] Merge scores work in evaluation working incorectly


Samoed:
### Describe the bug

I've found problems from https://github.com/embeddings-benchmark/mteb/pull/3125#issuecomment-3261813856. Now when results are merging https://github.com/embeddings-benchmark/mteb/blob/632cfdfffd7b14a85d87cefc365cf2316dea89ce/mteb/evaluate.py#L364 they're overriding with existing results, because in merge split scores passed existing_scores, new_scores instead of `new scores, old_scoes`

cc @KennethEnevoldsen 

### To reproduce

```python
import mteb
from mteb.load_results.task_results import TaskResult

old_task_result = TaskResult(
    dataset_revision="1",
    task_name="Task 1",
    mteb_version=mteb.__version__,
    scores={
        "test": [
            {
                "main_score": 0,
                "hf_subset": "default",
                "languages": ["eng-Latn"],
            }
        ]
    },
    evaluation_time=1,
)

new_task_result = TaskResult(
    dataset_revision="1",
    task_name="Task 1",
    mteb_version=mteb.__version__,
    scores={
        "test": [
            {
                "main_score": 1,
                "hf_subset": "default",
                "languages": ["eng-Latn"],
            }
        ]
    },
    evaluation_time=1,
)
final_task_result = new_task_result.merge(old_task_result)
print(final_task_result.scores)
# {'test': [{'main_score': 0, 'hf_subset': 'default', 'languages': ['eng-Latn']}]}
```

### Are you interested to contribute a fix for this bug?

No

KennethEnevoldsen:
Hmm, unsure if this is a bug. This is probably what I would expect, similarly to dict update:
```python
d = {"a": 1}
d.update({"a": 0})
print(d)
# {'a': 0}
```

But this situation shouldn't have happened in the first place, right? We shouldn't have two "test" splits. I assume this was fixed in #3130.

Samoed:
Problem is that result from current run will have less priority than result from the cache. I've run with `overwrite_strategy=always` and because of that cache wasn't hit. 

https://github.com/embeddings-benchmark/mteb/blob/632cfdfffd7b14a85d87cefc365cf2316dea89ce/mteb/evaluate.py#L363-L364

<details>
<summary>Script</summary>

```python
import logging
from pathlib import Path

branch = "refactor"
logging.basicConfig(level=logging.INFO, filename=f"{branch}.log", filemode="w")

import mteb
[...]

PR #3140, opened 2025-09-07
fix save result override

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3134


KennethEnevoldsen:
Let us finish the discussion in the issue first - I might be missing something, but I don't think the issue is currently there

PR #3147, opened 2025-09-07
dataset: Add PosIR retrieval datasets

DunZhang:
checklist:

### DIFF
--- a/mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
+++ b/mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
@@ -121,9 +121,9 @@ class PosIRAccommodationCateringHotelRetrieval(MultilingualTask, AbsTaskRetrieva
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRAccommodationCateringHotelRetrieval",
-        domains=["Non-fiction"],
+        domains=["Written"],
         description=(
-            "This test set is based on the accommodation_catering_hotel subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the accommodation_catering_hotel subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -136,9 +136,9 @@ class PosIRAerospaceRetrieval(MultilingualTask, AbsTaskRetrieval):
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRAerospaceRetrieval",
-        domains=["Engineering"],
+        domains=["Engineering", "Written"],
         description=(
-            "This test set is based on the aerospace subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the aerospace subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -153,9 +153,9 @@ class PosIRAgricultureForestryAnimalHusbandryFisheryRetrieval(
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRAgricultureForestryAnimalHusbandryFisheryRetrieval",
-        domains=["Non-fiction"],
+        domains=["Written"],
         description=(
-            "This test set is based on the agriculture_forestry_animal_husbandry_fishery subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the agriculture_forestry_animal_husbandry_fishery subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -170,9 +170,9 @@ class PosIRArtificialIntelligenceMachineLearningRetrieval(
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRArtificialIntelligenceMachineLearningRetrieval",
-        domains=["Engineering"],
+        domains=["Academic", "Written"],
         description=(
-            "This test set is based on the artificial_intelligence_machine_learning subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the artificial_intelligence_machine_learning subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -185,9 +185,9 @@ class PosIRAutomobileRetrieval(MultilingualTask, AbsTaskRetrieval):
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRAutomobileRetrieval",
-        domains=["Engineering"],
+        domains=["Engineering", "Written"],
         description=(
-            "This test set is based on the automobile subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the automobile subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -200,9 +200,9 @@ class PosIRBiomedicineRetrieval(MultilingualTask, AbsTaskRetrieval):
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRBiomedicineRetrieval",
-        domains=["Medical"],
+        domains=["Medical", "Written"],
         description=(
-            "This test set is based on the biomedicine subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the biomedicine subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -215,9 +215,9 @@ class PosIRComputerCommunicationRetrieval(MultilingualTask, AbsTaskRetrieval):
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRComputerCommunicationRetrieval",
-        domains=["Engineering"],
+        domains=["Web", "Written"],
         description=(
-            "This test set is based on the computer_communication subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the computer_communication subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -230,9 +230,9 @@ class PosIRComputerProgrammingCodeRetrieval(MultilingualTask, AbsTaskRetrieval):
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRComputerProgrammingCodeRetrieval",
-        domains=["Engineering"],
+        domains=["Programming"],
         description=(
-            "This test set is based on the computer_programming_code subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the computer_programming_code subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -247,9 +247,9 @@ class PosIRCurrentAffairsGovernmentAdministrationRetrieval(
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRCurrentAffairsGovernmentAdministrationRetrieval",
-        domains=["News"],
+        domains=["News", "Government"],
         description=(
-            "This test set is based on the current_affairs_government_administration subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the current_affairs_government_administration subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -262,9 +262,9 @@ class PosIRElectricPowerEnergyRetrieval(MultilingualTask, AbsTaskRetrieval):
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRElectricPowerEnergyRetrieval",
-        domains=["Engineering"],
+        domains=["Engineering", "Written"],
         description=(
-            "This test set is based on the electric_power_energy subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the electric_power_energy subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -277,9 +277,9 @@ class PosIRFilmEntertainmentRetrieval(MultilingualTask, AbsTaskRetrieval):
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRFilmEntertainmentRetrieval",
-        domains=["Entertainment"],
+        domains=["Entertainment", "Written"],
         description=(
-            "This test set is based on the film_entertainment subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the film_entertainment subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
@@ -292,9 +292,9 @@ class PosIRFinanceEconomicsRetrieval(MultilingualTask, AbsTaskRetrieval):
     metadata = TaskMetadata(
         **_POSIR_METADATA,
         name="PosIRFinanceEconomicsRetrieval",
-        domains=["Financial"],
+        domains=["Financial", "Written"],
         description=(
-            "This test set is based on the finance_economics subset of the BAAI/IndustryCorpus2 dataset. "
+            "This test set is based on the finance_economics subset of the BAAI/IndustryCorpus2 sampled from NLP pre-trained corpora (such as WuDaoCorpora, redpajama, and SkyPile-150B, among others). "
             + _COMMON_DESCRIPTION_SUFFIX
         ),
     )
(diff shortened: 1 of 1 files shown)


DunZhang:
Dear MTEB team,

Because this PR adds many 124 files and modifies 1 file, the change set is relatively large. I am therefore providing a detailed description to facilitate your review.

## 1. Necessity of this test suite
Recently, our paper [An Empirical Study of Position Bias in Modern Information Retrieval](https://arxiv.org/abs/2505.13950) (has been accepted by EMNLP 2025) investigates position bias in information retrieval, where models tend to overemphasize content at the beginning of passages while neglecting semantically relevant information that appears later.

Our experiments show that when relevant information appears later in the passage, dense embedding models and ColBERT-style models suffer significant performance degradation (an average drop of 15.6%).

This benchmark is an official extension of the paper. We believe the paper and this test suite provide a new perspective for evaluating retrieval models, and we would like to contribute it to MTEB.

## 2. The test results of the datasets
I have tested that the dataset runs with the mteb package.  
Three models are tested: `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`, `intfloat/multilingual-e5-small` and `BAAI/bge-m3`.

We carefully examined the evaluation results for the three models and found no anomalies; the results are fully consistent with the paper’s conclusions. The average scores are as follows:
- `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`: 0.2078
- `intfloat/multilingual-e5-small`: 0.4918
- `BAAI/bge-m3`: 0.5893

We can observe that `BAAI/bge-m3` outperforms the other two models. The primary reason is that our corpus passages have a maximum length of 2048, while `bge-m3` supports inputs up to 8192, and the other two models are limited to 512.

**The performance ranking of these three models is consistent with the MMTEB leaderboard.**

**Please see the attachment for detailed results.**
[posir_results.zip](https://github.com/user-attachments/files/22197647/posir_results.zip)


## 3. Basic information about the datasets
This benchmark covers 31 domains in total. Of these, 30 domains’ corpora are drawn from [BAAI/IndustryCorpus2](https://huggingface.co/datasets/BAAI/IndustryCorpus2). The 30 specific domains are:
```json
{
    "交通运输": "transportation",
    "医学_健康_心理_中医": "medicine_health_psychology_traditional_chinese_medicine",
    "数学_统计学": "mathematics_statistics",
    "时政_政务_行政": "current_affairs_government_administration",
    "消防安全_食品安全": "fire_safety_food_safety",
    "石油化工": "petrochemical",
    "计算机_通信": "computer_communication",
    "人工智能_机器学习": "artificial_intelligence_machine_learning",
    "其他信息服务_信息安全": "other_information_services_information_security",
    "学科教育_教育": "subject_education_education",
    "文学_情感": "literature_emotion",
    "水利_海洋": "water_resources_ocean",
    "游戏": "game",
    "科技_科学研究": "technology_scientific_research",
    "采矿": "mining",
    "住宿_餐饮_酒店": "accommodation_catering_hotel",
    "其他制造": "other_manufacturing",
    "影视_娱乐": "film_entertainment",
    "新闻传媒": "news_media",
    "汽车": "automobile",
    "生物医药": "biomedicine",
    "航空航天": "aerospace",
    "金融_经济": "finance_economics",
    "体育": "sports",
    "农林牧渔": "agriculture_forestry_animal_husbandry_fishery",
    "房地产_建筑": "real_estate_construction",
    "旅游_地理": "tourism_geography",
    "法律_司法": "law_judiciary",
    "电力能源": "electric_power_energy",
    "计算机编程_代码": "computer_programming_code"
}

```
The remaining domain is selected from FineWeb and can be regarded as a general-purpose test set.

Each dataset has both a Chinese version and an English version (not translations). By using strong commercial LLMs and retrieval–reranking models, we designed a reasonable question-generation pipeline and a stringent filtering procedure to ensure data quality (see the paper for details).

For each domain, there are two test sets: Head and Tail (e.g., PosIRAutomobileHeadRetrieval and PosIRAutomobileTailRetrieval). They share the same corpus; the difference is that the reference text for the test queries is located in the first half versus the second half of the passage, respectively.

Therefore, we have 31 × 2 = 62 test sets.

Each test set’s corpus contains roughly 60,000 passages. Passage lengths range between 100 and 2048. During sampling, we applied length-based uniform sampling, i.e., the number of passages is approximately balanced across equal-length intervals. Each test set contains about 500–700 queries.

## 4. Two perspectives for interpreting results
1) As in standard retrieval benchmarks, we use nDCG@10 as the evaluation metric; higher scores indicate better performance.

2) Within the same domain, compare the Head and Tail test sets. If a model scores higher on Head and lower on Tail, it indicates a stronger position bias.

(PS: For point 2, we have a dedicated formula to quantify position bias, but since it is not central to this PR, we omit the details here.)

## 5. About the code review
Although this PR includes 62 Python files, they share the same code structure. Only a few fields differ:
```
_INDUSTRY_DOMAIN variable
_POSITION variable
In TaskMetadata: the values of the name and domains parameters
```

## 6. Who We Are
We are the NovaSearch Team, the creators of the Jasper and Stella distillation models. We hope to contribute to retrieval.

mteb/tasks/Retrieval/multilingual/PosIRAccommodationCateringHotelHeadRetrieval.py
```diff
@@ -0,0 +1,116 @@
...
+from mteb.abstasks.AbsTaskRetrieval import AbsTaskRetrieval
+from mteb.abstasks.MultilingualTask import MultilingualTask
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+_INDUSTRY_DOMAIN = "accommodation_catering_hotel"
+# head or tail
+_POSITION = "Head"
+
+
+def _load_data(
```
Samoed: Maybe it would be easier to move all your tasks in singe file, because you're coping same `_load_data` function in all places? WDYT @KennethEnevoldsen? Also you can reupoad your task from `v2` branch, but this would be harder

Samoed (commented):
Congratulations with release! You also can add your benchmark to benchmarks file https://github.com/embeddings-benchmark/mteb/blob/main/mteb/benchmarks/benchmarks/benchmarks.py

mteb/tasks/Retrieval/multilingual/PosIRAccommodationCateringHotelHeadRetrieval.py
KennethEnevoldsen: I would move it into one script as well

Agree that at some point, we have to decide when to transition PRs to v2. If we add/update the documentation on contributing (retrieval) tasks, then we can start doing that.

mteb/tasks/Retrieval/multilingual/PosIRAccommodationCateringHotelHeadRetrieval.py
```diff
@@ -0,0 +1,116 @@
...
+
+from datasets import load_dataset
+
+from mteb.abstasks.AbsTaskRetrieval import AbsTaskRetrieval
+from mteb.abstasks.MultilingualTask import MultilingualTask
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+_INDUSTRY_DOMAIN = "accommodation_catering_hotel"
+# head or tail
+_POSITION = "Head"
```
KennethEnevoldsen: I think that all the Head/Tail datasets can be combined in one:

So that the subsets would be:
```py
eval_langs={"cmn-Hans-head": ["cmn-Hans"], "cmn-Hans-tail": ["cmn-Hans"], ...},
```

@samoed I suppose we don't have metrics across subsets? It would be great if we could add the comparison score as well. (seems like `task_specific_scores` are constrained in the subset)

I suppose it would be possible to compute it using something like:

```python
    def evaluate(
        self,
        model: Encoder,
        split: str = "test",
        subsets_to_run: list[HFSubset] | None = None,
        *,
        **kwargs,
    ) -> dict[HFSubset, ScoresDict]:
        """Custom evaluate function that only adds computation of custom metric across subsets"""

        scores =  super().evaluate(
            model,
            split,
            subsets_to_run,
            encode_kwargs=encode_kwargs,
            prediction_folder=prediction_folder,
            **kwargs,
        )
       psi = self.compute_psi()
       # add to each subset in scores
       return scores
```

then 
```py
main_score = task_results.get_score() # by default computes mean across each hf_subset of the main score
```
would be the "desired" PSI.

This also wouldn't be too much of a hassle to maintain, as the score interface is unlikely to change.

mteb/tasks/Retrieval/multilingual/PosIRAccommodationCateringHotelHeadRetrieval.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class PosIRAccommodationCateringHotelHeadRetrieval(MultilingualTask, AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        dataset={
+            "path": "infgrad/PosIR-Benchmark-v1",
+            "revision": "8faa66a19e4b2cf5764bc2e2adb19e9eca503d63",
+        },
+        name="PosIRAccommodationCateringHotelHeadRetrieval",
+        description="The Position-aware Information Retrieval (PosIR) Benchmark spans 31 domains and 2 languages, featuring query-related content that appears at varying positions within passages.",
```
KennethEnevoldsen: We will need a slightly more detailed description. It should be clear to the reader what the queries and corpus contain.

mteb/tasks/Retrieval/multilingual/PosIRAccommodationCateringHotelHeadRetrieval.py
Samoed: In `v2` we can add custom scores for the tasks. I think this should be easier

DunZhang:
@KennethEnevoldsen @Samoed 

Thank you both for your support and discussion. I have carefully read the comments and have the following thoughts:

### Regarding the Addition of a New Benchmark

Yes, once the PR is done, we will proceed to open another PR to add this benchmark, following the instructions at `https://github.com/embeddings-benchmark/mteb/blob/main/docs/adding_a_benchmark.md`.

### Regarding `custom scores`

We have not yet reached a definitive conclusion on the exact evaluation metric for this dataset. However, after some discussion, we have decided not to use the Position Sensitivity Index (PSI) mentioned in our paper. The PSI was designed for the base dataset in the paper, where the reference positions were discrete (i.e., head, middle, and tail). In contrast, this enhanced version of the dataset features continuous reference positions with specific start and end points, and we believe a more suitable metric can be developed.

In any case, I suggest we put the custom score issue on hold for now. I will revisit this and make the necessary modifications once we have a complete proposal for the new metric.

### Regarding `merge to one file`

I've noticed that in the MTEB codebase, each dataset typically has its own Python script. Could you please provide some guidance on how to merge them into a single file? The `_load_data` function is common across all the dataset files, but I'm unsure of the best place to locate it.

### Regarding `Head/Tail datasets can be combined in one`

Based on my debugging, if the Head/Tail datasets are combined into one (e.g., `{"cmn-Hans-head": ["cmn-Hans"],"cmn-Hans-tail": ["cmn-Hans"], "eng-Latn-head": ["eng-Latn"], "eng-Latn-tail": ["eng-Latn"]}`), a single domain would have four subsets. The final score displayed on the leaderboard would be the average of these four subsets, which would prevent us from observing the positional bias phenomenon.

In other words, **a `custom score` would be necessary to reflect the positional bias**.

Is my understanding correct?

Based on my current understanding of MTEB, I see two possible approaches:
1. Maintain the current structure.
2. Merge `PosIR***HeadRetrieval` and `PosIR***TailRetrieval` into a single dataset and treat it as a standard retrieval test set.

**Now that I know custom scores are an option, I am inclined towards the second approach.** We will definitely implement a custom score in the future, and once that is in place, the current simple binary Head/Tail division will become meaningless. I estimate we will release the custom score in three to six months.

However, regardless of the custom score, merging into a single dataset (the second approach) still holds value. For a model to achieve high performance on this combined dataset, it must not exhibit significant positional bias.

### Regarding `a slightly more detailed description`

I have already made the revisions and updated the code accordingly.

Finally, if there is anything else in the comments that requires further clarification, please let me know.

Samoed:
> I've noticed that in the MTEB codebase, each dataset typically has its own Python script. Could you please provide some guidance on how to merge them into a single file?

Not all tasks have 1 task - 1 file. You can put all your tasks just in 1 file. You can locate this file in same dir where your current tasks

KennethEnevoldsen:
> Based on my debugging, if the Head/Tail datasets are combined into one (e.g., {"cmn-Hans-head": ["cmn-Hans"],"cmn-Hans-tail": ["cmn-Hans"], "eng-Latn-head": ["eng-Latn"], "eng-Latn-tail": ["eng-Latn"]}), a single domain would have four subsets. The final score displayed on the leaderboard would be the average of these four subsets, which would prevent us from observing the positional bias phenomenon.

Indeed, but it would be impossible to compute the metric across two different tasks, so adding them as separate would prevent you from adding something like PSI as it would require information from another task.

DunZhang:
Based on the previous discussion, I have updated the code:  
1) Combined all PosIR tasks into one py
2) Removed the head/tail of the dataset (a custom score will be added in the v2 version later)  

The new evaluation results have been uploaded.
[posir_results_20250909.zip](https://github.com/user-attachments/files/22227825/posir_results_20250909.zip)

mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
```diff
@@ -0,0 +1,591 @@
...
+    "main_score": "ndcg_at_10",
+    "date": ("2025-03-04", "2025-08-20"),
+    "task_subtypes": ["Article retrieval"],
+    "license": "mit",
+    "annotations_creators": "LM-generated and reviewed",
+    "dialect": [],
+    "sample_creation": "LM-generated and verified",
+    "prompt": {
+        "query": "Given a web search query, retrieve relevant passages that answer the query",
+        "passage": "",
```
Samoed:
```suggestion
        "document": "",
```

Samoed (approved):
I think PR looks good

mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
```diff
@@ -0,0 +1,591 @@
...
+
+# 31 test sets
+class PosIRAccommodationCateringHotelRetrieval(MultilingualTask, AbsTaskRetrieval):
+    _industry_domain = "accommodation_catering_hotel"
+    metadata = TaskMetadata(
+        **_POSIR_METADATA,
+        name="PosIRAccommodationCateringHotelRetrieval",
+        domains=["Non-fiction"],
+        description=(
+            "This test set is based on the accommodation_catering_hotel subset of the BAAI/IndustryCorpus2 dataset. "
```
KennethEnevoldsen: I need a description of what the corpus and queries are. This is better, but it requires the user to know what the accommodation_catering_hotel contains

mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
```diff
@@ -0,0 +1,591 @@
...
+
+    load_data = _common_load_data
+
+
+class PosIRFireSafetyFoodSafetyRetrieval(MultilingualTask, AbsTaskRetrieval):
+    _industry_domain = "fire_safety_food_safety"
+    metadata = TaskMetadata(
+        **_POSIR_METADATA,
+        name="PosIRFireSafetyFoodSafetyRetrieval",
+        domains=["Non-fiction"],
```
KennethEnevoldsen: can you add non-fiction to the cases where it applied. I also seem like you can add "Written" for all datasets

mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
```diff
@@ -0,0 +1,591 @@
...
+
+    load_data = _common_load_data
+
+
+class PosIRLiteratureEmotionRetrieval(MultilingualTask, AbsTaskRetrieval):
+    _industry_domain = "literature_emotion"
+    metadata = TaskMetadata(
+        **_POSIR_METADATA,
+        name="PosIRLiteratureEmotionRetrieval",
+        domains=["Written"],
```
KennethEnevoldsen: fiction?

KennethEnevoldsen (commented):
so head-tail is completely removed? Seems like it is impossible to extract the targeted metrics?

DunZhang:
> so head-tail is completely removed? Seems like it is impossible to extract the targeted metrics?

I previously thought that it wasn't possible to customize the scoring, so I used the head/tail subset method, which is not very elegant, to demonstrate Position Bias.

However, after discussion, it was discovered that custom scoring is indeed possible. Yet, my teammates and I haven’t come up with specific metric ideas yet, so we decided to postpone this and add the custom scoring feature in the V2 version.

In any case, the current conventional metric calculation method can still indirectly reflect the model’s Position Bias.

mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
DunZhang: I double-checked the domains and made reasonable revisions.

mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
DunZhang: Yes, fiction

mteb/tasks/Retrieval/multilingual/PosIRRetrieval.py
DunZhang: Hi, the corpus comes from well-known NLP pre-training datasets. I have already modified the corresponding part in the description, so users should not be confused about the corpus.

Regarding the queries, they are all uniformly generated by a large model. The description includes an explanation of the problem types, which is specified in the variable `_COMMON_DESCRIPTION_SUFFIX`.

KennethEnevoldsen: Thanks for the clarification. I think that is reasonable

mteb/descriptive_stats/Retrieval/PosIRLawJudiciaryRetrieval.json
```diff
@@ -0,0 +1,58 @@
...
+        "num_samples": 126910,
+        "num_queries": 2647,
+        "num_documents": 124263,
+        "min_document_length": 6,
+        "average_document_length": 0.9512244191754585,
+        "max_document_length": 130,
+        "unique_documents": 124263,
+        "min_query_length": 3,
+        "average_query_length": 140.83452965621458,
+        "max_query_length": 3,
```
KennethEnevoldsen, line 13: @Samoed this seems like a bug?

KennethEnevoldsen:
> However, after discussion, it was discovered that custom scoring is indeed possible. Yet, my teammates and I haven’t come up with specific metric ideas yet, so we decided to postpone this and add the custom scoring feature in the V2 version.

> In any case, the current conventional metric calculation method can still indirectly reflect the model’s Position Bias.


Yeah so it is simply reflected in the NDCG@10 (given the way that the synthetic dataset is constructed)  

Hmm, but currently the datasets would have to be restructured to allow for the metric computation, no? Once this is merged, we will have to ensure compatibility, so I would prefer not to merge datasets that we know will be changed.

It is already possible to add it to v2. We can definitely help you out if needed.

mteb/descriptive_stats/Retrieval/PosIRLawJudiciaryRetrieval.json
Samoed: Yes, I had a bug. I've swapped queries and documents from one function return, but I think I've fixed it a long ago in `v2`. Also, we have `docuemnts` as dict and I just take `len` from it and there it contains `id`, `text`, `title`. I think can just recompute statistics in `v2`

DunZhang:
@Samoed @KennethEnevoldsen 

After discussing with our team members, we believe that a rigorous and comprehensive metric (as a supplement to the current Position Sensitivity Index, PSI) is crucial for clearly reflecting position bias. At the same time, it is also important to evaluate across as many models as possible to validate the reasonableness of the metric, which can provide the community with a solid baseline perspective. We would like to emphasize that, in addition to the position bias metric, the original NDCG metric should also be included, as it measures the absolute effectiveness of retrieval models.

Therefore, we have decided to close this PR, spend 1–2 months improving the current metric design and experiments, and then resubmit directly to v2.

KennethEnevoldsen:
Thanks for the message @DunZhang. That is a reasonable decision. I would also rather add it at a stage where we and you are both happy with it, rather than rushing it in.

Feel free to start a discussion if you need feedback on metrics development or integration. There is a great community in MTEB, and I suspect that multiple people could have valuable feedback. 

Regardless, I will be looking forward to the new PR - I think this dataset would be a great addition

PR #3154, opened 2025-09-07
Started adopting PEP8 naming convention

KennethEnevoldsen:
for #1124

@Samoed you are currently working with a few abstask, can I ask you to adopt snakecase for these (so that I don't cause conflicts for you)

We also have to decide if we want to snake_case task (should be easy enough to do, but will take a while)


mteb/_evaluators/regression_evaluator.py
```diff
@@ -21,8 +21,8 @@
 
 
 class SklearnRegressorModel(Protocol):
-    def fit(self, X: np.ndarray, y: np.ndarray, sample_weight=None) -> None: ...
-    def predict(self, X: np.ndarray) -> np.ndarray: ...
+    def fit(self, X: np.ndarray, y: np.ndarray, sample_weight=None) -> None: ...  # noqa: N803
+    def predict(self, X: np.ndarray) -> np.ndarray: ...  # noqa: N803
```
Samoed, line 25: I think this can be just renamed

Samoed:
> you are currently working with a few abstask, can I ask you to adopt snakecase for these (so that I don't cause conflicts for you)


I will adopt. You can just merge this.

> We also have to decide if we want to snake_case task (should be easy enough to do, but will take a while)

I think we should do, but this is too much work for now (I think we have like 600+ files) and I don't have ideas how to automate it.

Samoed:
I wanted to update ruff rules too, but I've found that Any2Any task would reqire too much work to adapt

mteb/_evaluators/regression_evaluator.py
KennethEnevoldsen: but the X is standard on the sklearn API, I would follow that

KennethEnevoldsen:
> I think we should do, but this is too much work for now (I think we have like 600+ files) and I don't have ideas how to automate it.

I can give it a go in another PR

KennethEnevoldsen:
> I wanted to update ruff rules too, but I've found that Any2Any task would reqire too much work to adapt

too many things to change because of the rules? Or do you mean that it takes too long to convert to image+text?

If it is the first one, it is perfectly fine to ignore errors, this helps us ensure new code is compliant.

Issue #3128, opened 2025-09-04
Can't run dialogs with different lenghts


Samoed:
### Describe the bug

Dataloader can't pass dialogs with different lengths

```python
>               raise RuntimeError("each element in list of batch should be of equal size")
E               RuntimeError: each element in list of batch should be of equal size
```

 Maybe we can convert each conversation in a list additionally  

### To reproduce

````python
from mteb.abstasks import AbsTaskRetrieval
from mteb import TaskMetadata
from datasets import Dataset
import mteb

from mteb.abstasks.retrieval_dataset_loaders import RetrievalSplitData

def base_retrieval_datasplit() -> RetrievalSplitData:
    return RetrievalSplitData(
        queries=Dataset.from_list(
            [
                {
                    "id": "q1",
                    "text": "This is a test sentence",
                },
                {
                    "id": "q2",
                    "text": "This is another test sentence",
                },
            ]
        ),
        corpus=Dataset.from_list(
            [
                {
                    "id": "d2",
                    "text": "This is a positive sentence",
                    "title": "Title of d1",
                },
                {
                    "id": "d1",
                    "text": "This is another positive sentence",
                    "title": "Title of d2",
                },
            ]
        ),
        relevant_docs={
            "q1": {"d1": 1, "d2": 0},
            "q2": {"d1": 0, "d2": 1},
        },
        top_ranked={
            "q1": ["d1", "d2"],
            "q2": ["d2", "d1"],
        },
    )

class MockRetrievalDialogTask(AbsTaskRetrieval):
    top_k = 1


    metadata = TaskMetadata(
        type="Retrieval",
        name="MockRetrievalDialogTask",
        main_score="ndcg_at_10",
        eval_splits=["val", "test"],
        dataset={
            "path": "test",
            "revision": "1"
        },
        description="",
        eval_langs=["eng-Latn"]
    )

    def load_data(self, **kwargs):
        base_datasplit = base_retrieval_datasplit()

        base_datasplit["top_ranked"] = None
        base_datasplit["queries"] = Dataset.from_dict(
            {
                "id": ["q1", "q2"],
                # dialogs with different lengths
                "text": [
                    [
                        {"role": "user", "content": "What is the weather like today?"},
                    ],
                    [
                        {"role": "user", "content": "What is the capital of France?"},
                        {
                            "role": "assistant",
                            "content": "The capital of France is Paris.",
                        },
                    ],
                ],
            }
        )

[...]

PR #3156, opened 2025-09-07
[v2] Support dialogs with different lengths

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3128


tests/test_benchmark/test_benchmark.py
```diff
@@ -454,3 +455,22 @@ def test_task_modality_filtering_model_modalities_more_than_task_modalities(task
         output_folder="tests/results",
         overwrite_results=True,
     )
+
+
+def test_dialogs_with_different_length_dataloader(tmp_path: Path):
+    class MockNumpyTextEncoder(AbsMockEncoder):
```
KennethEnevoldsen: Why not simply use the MockNumpyEncoder?

Samoed: I need to iterate through dataloader to trigger error. I can change `MockNumpyEncoder`
from 
```py
return np.random.rand(len(inputs.dataset), 10)
```
to 
```py
iter_key = list(batch[0].keys())[0]
n = [item for batch in inputs for item in batch[iter_key]]
return np.random.rand(n, 10)
```

KennethEnevoldsen: wouldn't that work?

Samoed: I've updated `MockNumpyEncoder.encode` and removed this test, because it will be triggered when evaluating this dialog task

Samoed:
I found that some datasets contain `None` values in certain columns, which causes the dataloader to fail when attempting to process them. To try to solve this issue, we can select only the columns that are required for the task in `_evaluate_subset`. However, I am not sure if this approach is the best solution

KennethEnevoldsen:
> I found that some datasets contain None values in certain columns

Hmm that seems odd, do you have an example?

Shouldn't we just clean up the dataset (feel like a dataset bug)

Samoed:
Yes. In test we have `IndicSentiment` and some rows of it in `ar` split in `BRAND` [column contains `None`](https://huggingface.co/datasets/mteb/IndicSentiment/viewer/as?row=19), but we don't use this column as text input.

We can just add in `_eval_subset` `dataset.select_columns([self.input_column, self.label_column])` to solve this issue

KennethEnevoldsen:
Ahh, I thought it was in the used column, yeah, so simply selecting relevant columns (dropping non-relevant) seems like a good idea.

Samoed:
@KennethEnevoldsen Can I merge this PR and make fix for task types later?

KennethEnevoldsen:
Feel free to

PR #3164, opened 2025-09-09
[v2]: Added task, benchmark, model overview to docs

### LINKED ISSUE #3166: Can't run documentation
### Describe the bug

When I'm trying to run `make serve-docs` I get error 
```
ERROR   -  Config value 'plugins': The "exclude" plugin is not installed
```

### To reproduce

`make serve-docs` on clean installation.

CC @KennethEnevoldsen

KennethEnevoldsen:
We could consider removing the task and benchmark tables. However, it is currently not a big cost to keep it.

Updates are available at:
https://embeddings-benchmark.github.io/mteb/overview/

This mostly makes tasks searchable:

Screenshot 2025-09-09 at 21 28 17

fixed #3166


Samoed:
1. Can we move List of tasks from TOC (from right sidebar) to overview (left sidebar) and in TOC will be name of tasks? (I this can be done by splitting each category to its own file)  Additionally, can we categorize the tasks by modality, but I think that this might be a bit difficult to implement.
2. Maybe citation can be added?
3. We can add a link to our source file path with 
```python
import mteb
import inspect

task = mteb.get_task("SCIDOCS")
inspect.getfile(task.__class__)  # '.../mteb/mteb/tasks/Retrieval/eng/SCIDOCSRetrieval.py'
```
but maybe this is irrelevant, because we can add link to the task API page and from it to github
4. And for dataset size estimation `n_samples` can be added https://github.com/embeddings-benchmark/mteb/blob/73d90207fa346850aeeef1e84570ab93231cbbcc/mteb/abstasks/task_metadata.py#L420
5. Link to descriptive stats. I think if we put them directly page will be too big to load

KennethEnevoldsen:
We could create separate files for it:

Screenshot 2025-09-10 at 10 39 41

```yaml
      - Tasks: 
        - Available Tasks: # extra layer here is required for drop-down
          - overview/test/a.md # can be detected using a plugin
          - overview/test/b.md
```

The plan is to also add models and datasets (so we can't have it expanded)

KennethEnevoldsen:
That way we can also add other types of overviews to tasks Wouldn't mind that.

KennethEnevoldsen:
Decided against adding tasks to TOC:

Screenshot 2025-09-10 at 13 19 54

As it will overload the UI

I also added benchmarks and model overview

Makefile
```diff
@@ -34,12 +34,18 @@ pr:
 	make test
 
 
-build-docs:
-	@echo "--- 📚 Building documentation ---"
-	# since we do not have a documentation site, this just build tables for the .md files
+build-tables:
```
Samoed, line 37: Maybe better to use hooks? https://www.mkdocs.org/user-guide/configuration/#hooks

KennethEnevoldsen: As I see it hooks are intended as plugins?

docs/overview/create_available_models.py
```diff
@@ -0,0 +1,155 @@
...
+####  `{model_name}`
+
+**Revision:** `{revision}` • **License:** {license} • [Learn more →]({reference})
+
+
+Max Tokens: {max_tokens}
+Embedding dimension: {embed_dim}
+Parameters: {n_parameters}
+Release date: {release_date}
+Languages: {languages}
```
Samoed: This duplicate table

docs/overview/create_available_models.py
```diff
@@ -0,0 +1,155 @@
...
+
+| Max Tokens | Embedding dimension | Parameters | Release date | Languages |
```
Samoed: Can we add required memory for model?

Samoed:
For models, I see only `Non-instruction Model` in TOC
image

docs/overview/create_available_models.py
```diff
@@ -0,0 +1,155 @@
...
+
+import mteb
+
+START_INSERT = "<!-- START TASK DESCRIPTION -->"
+END_INSERT = "<!-- END TASK DESCRIPTION -->"
+
+model_entry = """
+####  `{model_name}`
+
+**Revision:** `{revision}` • **License:** {license} • [Learn more →]({reference})
```
Samoed: I don't think that revision would be informative
```suggestion
**License:** {license} • [Learn more →]({reference})
```

docs/overview/create_available_models.py
```diff
@@ -0,0 +1,155 @@
...
+
+from pathlib import Path
+
+import mteb
+
+START_INSERT = "<!-- START TASK DESCRIPTION -->"
+END_INSERT = "<!-- END TASK DESCRIPTION -->"
+
+model_entry = """
+####  `{model_name}`
```
Samoed: I think learn more can be changed to model name
```suggestion
####  [`{model_name}`]({reference})
```

Makefile
Samoed: Hm, yes. Maybe scripts can be converted in one to not wait mteb import each time?

KennethEnevoldsen: Hmm would rather keep them seperate. It is not the CI that will take the longest.

docs/overview/create_available_benchmarks.py
```diff
@@ -0,0 +1,75 @@
...
+
+START_INSERT = "<!-- START TASK DESCRIPTION -->"
+END_INSERT = "<!-- END TASK DESCRIPTION -->"
+
+benchmark_entry = """
+####  {benchmark_name}
+
+{description}
+
+??? info Tasks
```
isaac-chung: Think it needs quotation marks to [change the title](https://mrkeo.github.io/reference/admonitions/#changing-the-title).
Screenshot 2025-09-10 at 19 43 40



```suggestion
??? info "Tasks"
```

docs/overview/create_available_models.py
```diff
@@ -0,0 +1,161 @@
...
+
+import mteb
+
+START_INSERT = "<!-- START TASK DESCRIPTION -->"
+END_INSERT = "<!-- END TASK DESCRIPTION -->"
+
+model_entry = """
+####  [`{model_name}`]({reference})
+
+ **License:** {license}
```
isaac-chung, line 16: For datasets, the name and license are on the same line. I randomly picked one convention to stick with.
```suggestion
####  [`{model_name}`]({reference}) • **License:** {license} 
```

docs/overview/create_available_benchmarks.py
```diff
@@ -0,0 +1,75 @@
...
+
+def format_benchmark_entry(benchmark: mteb.Benchmark) -> str:
+    description = benchmark.description or "No description available."
+    if benchmark.reference:
+        # add learn more link to description
+        description += (
+            "\n\n" + learn_more.format(reference=benchmark.reference) + "\n\n"
+        )
+
+    tasks_md = benchmark.tasks.to_markdown(["type", "modalities"])
```
isaac-chung: In tasks and models, languages are lists that formatted so that they don't have `[]`. Can that be done for modalities here as well?

docs/overview/create_available_benchmarks.py
```diff
@@ -0,0 +1,75 @@
...
+
+    task_md_indented = "\n".join([f"    {line}" for line in tasks_md.split("\n")])
```
isaac-chung, line 75: How do we feel about: clickable task names. Once we expand a benchmark to see the task list, each of the task names are clickable, and connects to the task page as if one navigated via the "available task" section.

isaac-chung (commented):
I'm really loving how this is looking so far. One general question (across models, tasks, benchmarks): How do we want to handle citations? Could they be an expandable text box that has a button to be copied?

docs/overview/create_available_models.py
KennethEnevoldsen: Hmm, but this will put it in the title, which is a bit odd. For dataset it is also not the name it is the "huggingface repository" which can be used for multiple datasets.

docs/overview/create_available_benchmarks.py
KennethEnevoldsen: I will see if it is doable using:
```
https://embeddings-benchmark.github.io/mteb/overview/available_tasks/{task_type}/?h={task_name}#{task_name}
```
But I think it might be

docs/overview/create_available_benchmarks.py
KennethEnevoldsen: If it is possible, I will give it a go

KennethEnevoldsen:
> How do we want to handle citations? 

Sure, we could add an expandable text box. I actually had it at some point, but decided to remove it, maybe I should add it back

docs/overview/create_available_models.py
isaac-chung: Oh right, that'd be odd. Let's ignore then.

docs/overview/create_available_benchmarks.py
KennethEnevoldsen: Added

docs/overview/create_available_benchmarks.py
KennethEnevoldsen: added

KennethEnevoldsen:
I think I fixes all of the issues. I will merge this in but if there is anything else, then feel free to let me know and I will fix it

Samoed:
In `v2` models have citation field too. Can we add citation to them? https://github.com/embeddings-benchmark/mteb/blob/f4c30aef6c6e23c09f0df3c753baa0d533da9d2e/mteb/models/model_meta.py#L121

mteb/overview.py
```diff
@@ -7,6 +7,7 @@
 from collections import Counter, defaultdict
 
 import pandas as pd
+from traitlets import Any
```
Samoed: @KennethEnevoldsen tests are failing because of this import

KennethEnevoldsen: a damn autofix! fixed

KennethEnevoldsen:
> In v2 models have citation field too. Can we add citation to them?

It is already added (though not many models have it annotated):

https://embeddings-benchmark.github.io/mteb/overview/available_models/text/#baaibge-reranker-v2-m3

KennethEnevoldsen:
> For models, I see only Non-instruction Model in TOC

Good catch - fixed:
https://embeddings-benchmark.github.io/mteb/overview/available_models/text/#instruction-model

PR #3167, opened 2025-09-10
fix: make sentence_transformers an optional import

KennethEnevoldsen:
It is still a required dependency, but this makes the import time consistenly 0.5-1s faster.

Also did a minor refactor of `evaluate.py`

related to #1463

### DIFF
--- a/mteb/models/get_model_meta.py
+++ b/mteb/models/get_model_meta.py
@@ -201,7 +201,6 @@ def model_meta_from_hf_hub(model_name: str) -> ModelMeta:
 
 
 def _model_meta_from_cross_encoder(model: CrossEncoder) -> ModelMeta:
-    
     return ModelMeta(
         loader=CrossEncoderWrapper,
         name=model.model.name_or_path,


mteb/models/sentence_transformer_wrapper.py
```diff
@@ -4,28 +4,26 @@
 from typing import TYPE_CHECKING, Any
 
 import numpy as np
-import sentence_transformers
+
+# import sentence_transformers
```
Samoed:
```suggestion
```

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -54,19 +54,20 @@ def __init__(
...
         else:
             self.model = model
+        from mteb.models.get_model_meta import (
+            _model_meta_from_sentence_transformers,
+        )
 
-        if (
-            built_in_prompts := getattr(self.model, "prompts", None)
-        ) and not model_prompts:
+        self.mteb_model_meta = _model_meta_from_sentence_transformers(self.model)
```
Samoed, line 65: I don't think this is reqired. This needed only if user passes SentenceTransformer or CrossEncoder wihout `mteb.get_model`

Samoed (approved):

mteb/models/sentence_transformer_wrapper.py
KennethEnevoldsen: Sure, but it will just be overwritten if you use `mteb.get_model`.

Before you could pass a SentenceTransformer without having it assigned. So if you just passed a SentenceTransformer to evaluate you wouldn't get the meta.

Samoed: Model meta will be added in evaluator https://github.com/embeddings-benchmark/mteb/blob/f4c30aef6c6e23c09f0df3c753baa0d533da9d2e/mteb/evaluate.py#L271

KennethEnevoldsen: But that requires:
https://github.com/embeddings-benchmark/mteb/blob/f4c30aef6c6e23c09f0df3c753baa0d533da9d2e/mteb/evaluate.py#L65-L82

Samoed: Yes, but wouldn't be in that case sentence transformer will be passed and metadata created? This is minor, you can ignore this

Samoed:
I think we can move torch imports inside functions to speedup imports

mteb/models/sentence_transformer_wrapper.py
Samoed: Ah, I see that you've removed creation of model meta in this function. Yes, then this is better solution then. But this should be added to cross encoders to fix the test

KennethEnevoldsen:
atm. I am just taking the main things first. Currently, torch is not at the top of the list. When it is, I will let you know.
Transformers is a big one

PR #3168, opened 2025-09-10
dataset: Adding JapaneseCode1Retrieval as the first non-public dataset

fzoll:
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the mteb run -m {model_name} -t {task_name} command.
  - [ ] sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2
  - [x] intfloat/multilingual-e5-small

### DIFF
--- a/.github/workflows/dataset_loading.yml
+++ b/.github/workflows/dataset_loading.yml
@@ -23,7 +23,5 @@ jobs:
           make install-for-tests
 
     - name: Run dataset loading tests
-      env:
-        HF_TOKEN: ${{ secrets.MTEB_PRIVATE }}
       run: |
         make dataset-load-test
--- a/.github/workflows/dataset_loading_pr.yml
+++ b/.github/workflows/dataset_loading_pr.yml
@@ -24,7 +24,5 @@ jobs:
           make install-for-tests
 
     - name: Run dataset loading tests
-      env:
-        HF_TOKEN: ${{ secrets.MTEB_PRIVATE }}
       run: |
         make dataset-load-test-pr BASE_BRANCH=${{ github.event.pull_request.base.ref }}


mteb/tasks/Retrieval/code/JapaneseCode1Retrieval.py
```diff
@@ -0,0 +1,95 @@
...
+logger = logging.getLogger(__name__)
+
+
+class JapaneseCode1Retrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="JapaneseCode1Retrieval",
+        description="Japanese code retrieval dataset. Japanese natural language queries paired with Python code snippets for cross-lingual code retrieval evaluation.",
+        reference="https://huggingface.co/datasets/mteb-private/JapaneseCode1Retrieval",
+        dataset={
+            "path": "local",
```
Samoed: Can you download from hf repo directly? Also can you convert to HF format to load this task without `load_data`

mteb/tasks/Retrieval/code/JapaneseCode1Retrieval.py
```diff
@@ -0,0 +1,31 @@
...
+
+from mteb.abstasks.AbsTaskRetrieval import AbsTaskRetrieval
+from mteb.abstasks.TaskMetadata import TaskMetadata
+
+
+class JapaneseCode1Retrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="JapaneseCode1Retrieval",
+        description="Japanese code retrieval dataset. Japanese natural language queries paired with Python code snippets for cross-lingual code retrieval evaluation.",
+        reference="https://huggingface.co/datasets/mteb-private/JapaneseCode1Retrieval",
```
KennethEnevoldsen: We state in the blog (will forward it to you @Samoed) that we want to share samples from this. So should this lint be to a `mteb-private/JapaneseCode1Retrieval-sample`?

Samoed: I think yes. Great idea!

KennethEnevoldsen:
Seems like CI is failing because the ds is not on HF. You can fix this by simply "ignoring" closed datasets in a list "accepted" closed datasets (specified in tests).

mteb/tasks/Retrieval/code/JapaneseCode1Retrieval.py
fzoll: @KennethEnevoldsen @Samoed  Ok, sure! I will tweak the sample a bit as well. Also, i guess, the sample dataset should be public.

fzoll:
@KennethEnevoldsen Don't you want to allow the CI/tests to reach the private datasets instead? (I guess we just need to add a token as a secret and modify the GH action to use the secret as an env.variable).

mteb/tasks/Retrieval/code/JapaneseCode1Retrieval.py
fzoll: @KennethEnevoldsen @Samoed https://huggingface.co/datasets/mteb-private/JapaneseCode1Retrieval-sample

Samoed: I see only qrels split on HF preview. Can you upload it to see qrels and corpus splits too? Also do we want to add samples as tasks too? I think this can help to debug inputs for users if they need this

fzoll: @Samoed Please check the dataset now. I don't know if a task is required for the sample, @KennethEnevoldsen ?

Samoed: Yes, it's better now

KennethEnevoldsen: Usage sample is wrong. I would do:

```py
import mteb

# Load the sample dataset
task = mteb.get_task("JapaneseCode1Retrieval")
evaluator = mteb.MTEB(tasks=[task])

# Run evaluation with your model
model = mteb.get_model("your-model-name")
results = evaluator.run(model) # requires hf_token to run as it is a closed dataset
```

KennethEnevoldsen:
@fzoll that would be much better. I have added a token `MTEB_PRIVATE` that should have access.

fzoll:
@KennethEnevoldsen Ok, let me check the required changes in the CI.

Samoed:
Action needed to be updated to use `MTEB_PRIVATE`

Samoed:
I will fix token

Samoed:
I don't have ideas why action is not triggered. I think it might by due branch protection rule @KennethEnevoldsen

fzoll:
@Samoed Thanks for the CI change! It seems to me that the tests run and succeeded at the end.

Samoed:
It was successes, but `Datasets available on HuggingFace` wasn't triggered

.github/workflows/dataset_loading_pr.yml
```diff
@@ -1,17 +1,24 @@
 name: Datasets available on HuggingFace - PR
 
 on:
-  pull_request:
-    paths:
-      - "mteb/tasks/**.py"
```
isaac-chung: Should add this back, otherwise unrelated changes will also trigger this run, which we were trying to avoid in the first place.

Samoed: Added back

KennethEnevoldsen:
Looks good! I think this is ready to merge then

.github/workflows/dataset_loading_pr.yml
```diff
@@ -1,17 +1,26 @@
 name: Datasets available on HuggingFace - PR
 
 on:
-  pull_request:
+  pull_request_target:
+    types: [opened, synchronize, edited]
```
isaac-chung: Still not seeing this workflow being run. I think these 3 lines should be reverted.

Samoed: I reverted them, but wihtout these lines token from fork will be used, but it not exist there

isaac-chung: Oh I see.

Samoed: I don't know how to make this action working with pr's from forks. Maybe we can ignore private dataset in action that test them in PR

isaac-chung: Yeah, let's ignore for now and start an issue to ask for help.

Samoed: I've rolled backed changes in this action. We can igonre CI and merge it

KennethEnevoldsen: Can we just do a simple test to see that the secret is working? (e.g. create a specific CI which loads this private dataset

```
python -c "import datasets; ds = load_dataset(...); print("dataset loaded")"
```

Samoed: I tried and only passing `token` directly to `load_dataset` worked. `hf auth login` or python `login` didn't work

isaac-chung: So probably better off skipping private ones in the existing test, and write a separate one for private datasets then eh

pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers = [
 ]
 requires-python = ">=3.9,<3.13"
 dependencies = [
-    "datasets>=2.19.0, <4.0.0",
+    "datasets==4.0.0",
```
Samoed: We don't support datasets `v4`. I've tested on this version
```suggestion
    "datasets<4.0.0",
```

Samoed (changes requested):

pyproject.toml
fzoll: @Samoed Ok, got it. I will revert this change (probably i will revert the last few), i just wanted to see why the token is not picked up.

Samoed: I think with `pull_request` token from your repo is taken, but you don't have it. To fix this `pull_request_target` should be used probably

fzoll: @Samoed Ok, i changed to `pull_request_target`. I added my token to my repo as well.

fzoll: @Samoed ...but as i see, with `pull_request_target`, `Datasets available on HuggingFace - PR` is not triggered

fzoll: I reverted my changes

.github/workflows/dataset_loading_pr.yml
fzoll: @isaac-chung @Samoed I can exclude the private datasets from this test...but how to test the private datasets?

isaac-chung: > only passing token directly to load_dataset worked

> write a separate one for private datasets

.github/workflows/dataset_loading.yml
```diff
@@ -23,5 +23,7 @@ jobs:
           make install-for-tests
 
     - name: Run dataset loading tests
+      env:
+        HF_TOKEN: ${{ secrets.MTEB_PRIVATE }}
```
KennethEnevoldsen: After some discussion with @fzoll, I think it is better to ignore the private datasets in the test.

There are a couple of reasons:
- This does not seem to be working, and @fzoll suggested that this might be due to GitHub-based token protection to prevent leaking a token. Not entirely sure if it is the case though, couldn't find a source on it.
- If there is issue, it will only be us that have to deal with it so by not testing this we will not expose any issues other to the users
- It makes it possible to run all tests locally without having to have a token, thus making it easier for contributors.

fzoll:
@KennethEnevoldsen What do you think?

mteb/tasks/Retrieval/code/JapaneseCode1Retrieval.py
KennethEnevoldsen: @fzoll this is still a issue

.github/workflows/dataset_loading.yml
KennethEnevoldsen: delete?

.github/workflows/dataset_loading_pr.yml
```diff
@@ -24,5 +24,7 @@ jobs:
           make install-for-tests
 
     - name: Run dataset loading tests
+      env:
+        HF_TOKEN: ${{ secrets.MTEB_PRIVATE }}
```
KennethEnevoldsen: delete

KennethEnevoldsen (approved):

.github/workflows/dataset_loading.yml
KennethEnevoldsen:
```suggestion
```

.github/workflows/dataset_loading_pr.yml
KennethEnevoldsen:
```suggestion
```

PR #3170, opened 2025-09-11
[v2] don't import transformers on import

KennethEnevoldsen:
set_seed is called during the initialization of the aggregate tasks. I have refactored it to take the main components from transformers set_seed. We do not initialize the seed for specialized hardware, but maybe that is also not needed?

With this change we no longer need make transformers an optional install.

This give a notably increase in load time
6.511, 6.798, 6.423
to
5.213, 5.230, 4.929

related to #1463

### DIFF
--- a/mteb/set_seed.py
+++ b/mteb/set_seed.py
@@ -1,0 +1,46 @@
+"""Utilities for setting seeds for reproducibility.
+
+Derived from `transformers.trainer_utils.set_seed`. It assumes torch is installed.
+"""
+
+from __future__ import annotations
+
+import logging
+import random
+
+import numpy as np
+import torch
+
+logger = logging.getLogger(__name__)
+
+
+def set_seed(seed: int) -> tuple[random.Random, np.random.Generator]:
+    """Helper function for reproducible behavior to set the seed in `random`, `numpy`, `torch` and/or `tf` (if installed).
+
+    If does not set the seed for npu, musa, mlu, hpu, xpu devices. To set the seed for those devices, we recommend using:
+
+    ```python
+    from transformers import set_seed
+    set_seed(seed)
+    ```
+
+    Args:
+        seed: The seed to set.
+
+    Returns:
+        A tuple of (random.Random, np.random.Generator) initialized with the given seed.
+    """
+    random.seed(seed)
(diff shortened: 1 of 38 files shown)


mteb/set_seed.py
```diff
@@ -0,0 +1,46 @@
...
+    Args:
+        seed: The seed to set.
+
+    Returns:
+        A tuple of (random.Random, np.random.Generator) initialized with the given seed.
+    """
+    random.seed(seed)
+    np.random.seed(seed)  # noqa: NPY002
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
```
Samoed, line 36: Can we add `transformers.set_seed(seed)` too? just in case?

KennethEnevoldsen: No, because this is called in the aggregated tasks, so it will be run on import

Which is not ideal - I tried to refactor them, but I am not sure what the best solution is, especially the visualSTS tasks are a bit problematic:
`STSBenchmarkMultilingualVisualSTSEng`

(would love to not have these be aggregated tasks)

Samoed: > (would love to not have these be aggregated tasks)

I agree. They're a bit strange to me too. Maybe we can remove set seed from the aggregated task and use it only in other tasks? 


Also, maybe we should make the seed hard coded in order to ensure reproducibility?

KennethEnevoldsen: No it is the task itself that is initialised. So it is either refactoring those or make the seed initialisation not depend on transformers.

> Also, maybe we should make the seed harder to change in order to ensure reproducibility?

~~Not sure what you mean by "harder"?~~

Ahh, I see. Hmm, I think the way it is not is pretty reasonable. There are plenty of good reasons to change the seed, e.g., to ensure that the results we see are not just luck.

Samoed: I think we should make it harder to change the seed. At the very least, we could remove it from the task initialization arguments. https://github.com/embeddings-benchmark/mteb/blob/6d69d7fe3c9f654b4db31e5158f13769aa521001/mteb/abstasks/AbsTask.py#L102

KennethEnevoldsen: Hmm. I do think there is good reasons to change it

PR #3176, opened 2025-09-13
[v2] Start type checking

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/issues/2714

- Aligned `task._evaluate_subset` with each other
- Aligned `Evaluator.__call__`, but classification and regression task return 2 items instead of 1


mteb/_evaluators/classification_evaluator.py
```diff
@@ -111,13 +111,13 @@ def calculate_scores(
...
 
-    def __call__(
+    def __call__(  # type: ignore[override]
         self,
         model: Encoder,
         *,
         encode_kwargs: dict[str, Any],
         test_cache: np.ndarray | None = None,
-    ) -> tuple[dict[str, float], Any]:
+    ) -> tuple[dict[str, float], np.ndarray | None]:
```
KennethEnevoldsen: a docstring would be good here

pyproject.toml
```diff
@@ -107,12 +107,24 @@ docs = [
...
     "mkdocs-exclude>=1.0.2",
     "mkdocs-include-dir-to-nav>=1.2.0",
-    "tabulate>=0.9.0"
+    "tabulate>=0.9.0",
 ]
 dev = [
     {include-group = "lint"},
     {include-group = "test"},
 ]
+typing = [
```
KennethEnevoldsen, line 112: Let us add a makefile command as well

pyproject.toml
```diff
@@ -262,3 +274,34 @@ conflicts = [
     [{ extra = "colpali-engine" }, { extra = "pylate" }],
     [{ extra = "colpali-engine" }, { extra = "llm2vec" }]
 ]
+
+[tool.mypy]
+plugins = ['pydantic.mypy']
+
+[[tool.mypy.overrides]]
```
KennethEnevoldsen, line 284: Should we add some comments to these overrides - not quite sure what they do

pyproject.toml
```diff
@@ -107,12 +107,24 @@ docs = [
     "mkdocs-bibtex>=2.16.2",
     "mkdocs-exclude>=1.0.2",
     "mkdocs-include-dir-to-nav>=1.2.0",
-    "tabulate>=0.9.0"
+    "tabulate>=0.9.0",
 ]
 dev = [
```
KennethEnevoldsen: include typing?

pyproject.toml
```diff
@@ -107,12 +107,24 @@ docs = [
...
+typing = [
+    "mypy>=1.18.1",
+    "pandas-stubs>=2.3.2.250827",
+    "scipy-stubs>=1.15.3.0",
+    "types-cachetools>=6.2.0.20250827",
+    "types-pysocks>=1.7.1.20250828",
+    "types-pyyaml>=6.0.12.20250822",
+    "types-requests>=2.32.4.20250913",
+    "types-simplejson>=3.20.0.20250822",
+    "types-tensorflow>=2.18.0.20250809",
```
KennethEnevoldsen: do we want to check for TensorFlow typing? It is a very small subset of the codebase (would probably just ignore them) - assuming that it install tensorflow/keras?

mteb/create_dataloaders.py
```diff
@@ -291,14 +308,22 @@ def create_document_dataloader(
     dataset: Dataset,
     task_metadata: TaskMetadata,
     input_column: str | None = None,
-    **dataloader_kwargs: dict[str, Any],
+    batch_size: int = 32,
```
KennethEnevoldsen, line 311: not dataloader_kwargs?

pyproject.toml
Samoed: I've made `mypy --install-types` and it was added. I can remove it for now

mteb/create_dataloaders.py
Samoed: I don't think we need them, because only `batch_size` is helpful for us, I think https://docs.pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader

pyproject.toml
Samoed: Added

pyproject.toml
Samoed: Was added because of https://github.com/embeddings-benchmark/mteb/blob/45114a584c72545a08c0eb35d0b5d847c71353aa/mteb/set_seed.py#L40

mteb/_evaluators/classification_evaluator.py
Samoed: Added

mteb/create_dataloaders.py
KennethEnevoldsen: What about num_workers?

pyproject.toml
KennethEnevoldsen: great - yeah let us remove it

mteb/create_dataloaders.py
Samoed: I don't think that we require this, because we don't do something tricky, but I will test this

Issue #1593, opened 2024-12-14
Add Support for Index-Based Retrieval for ColBERT


sam-hey:
The current implementation of ColBERT only utilizes the MaxSim operator for retrieval and does not leverage an index. While this works for smaller datasets, it is not scalable for large datasets, as the resource consumption becomes impractical.

To enable efficient retrieval for large-scale datasets, it is essential to integrate an indexing mechanism. This will ensure that resource consumption remains manageable while maintaining retrieval performance.

PyLate plans to introduce the PLAID Index to address this limitation. This index is designed to optimize retrieval tasks for larger datasets. For details on the current progress, refer to the [PyLate Issue](https://github.com/lightonai/pylate/issues/72).

Needs: https://github.com/embeddings-benchmark/mteb/issues/1826

NohTow:
Hey @sam-hey,
We already discussed on the PyLate repository about this (thank you for being active there btw).
I plan to move forward with the merge of the PLAID branch soon-ish.
I can also work on a minimal boilerplate with it.
It should be fairly similar to our existing BEIR boilerplates, I did one with the PLAID index [here](https://github.com/lightonai/pylate/compare/main...PLAID#diff-f2509658d1da5c0d8f573d5598c6d720eadbb60709b5484bef55097294bf4f57), the main difference being that right now we are directly calling the index with `scores = index(queries_embeddings)` instead of using retrieve, but it'll change in the final release and be usable as any of our index and so the boilerplate will be totally the same (up to `index = indexes.PLAID(override=True)`).

I have something working with MTEB dataloader as well, but I am not sure what you would need exactly?

sam-hey:
Hey @NohTow,
That’s awesome to hear—thanks for reaching out!
I don’t think you need to worry about the MTEB implementation. You can just focus on building PyLate, as MTEB will simply use the provided programmatic endpoints to retrieve the documents along with their attached metadata.

NohTow:
Alright, I started [drafting the PR](https://github.com/lightonai/pylate/pull/116), should be merged soon!
I'll ping you once it's done to let you know

NohTow:
Hey @sam-hey,
We merged the branch, PLAID indexes are now available (from source now, but we'll be doing a release today normally)!
I shared some of my boilerplates [here](https://huggingface.co/lightonai/GTE-ModernColBERT-v1/discussions/1), notably the LongEmbed one should be very helpful as it directly uses MTEB data :)

[...]

PR #3183, opened 2025-09-16
[v2] add multi-vector indexes so we can scale multi-vector search

orionw:
I need a multi-index vector database with MTEB for a different project I'm involved in and thanks to @Samoed's recent efforts I can do it without too much effort!

I don't have time to finish this PR for the next month probably (unless it's just minor changes) but this is a working example of PyLate using an index. Perhaps if someone else has the time to finish this, it could be a very useful addition to v2.0.0.

I tried to use ColBERT style models without this and without the index it's impossible to run larger datasets since the max sim operation would require hundreds of GBs of GPU VRAM. 

What do you think @KennethEnevoldsen @Samoed?  I tried to keep it somewhat open since I imagine we will need to incorporate sparse models (I hear that's coming to SentenceTransformers soon ish).

cc @NohTow who gave me some pointers.

Close https://github.com/embeddings-benchmark/mteb/issues/1593


mteb/models/search_wrappers.py
```diff
@@ -204,6 +306,137 @@ def _full_corpus_search(
                         heapq.heappushpop(result_heaps[query_id], (score, corpus_id))
         return result_heaps
 
+    def _index_multivector_pylate(
```
Samoed: Great! But I think it would be better to add `search` and `index` functions to the ColBERT wrapper to make it compatible with the search protocol instead of changing wrapper

Samoed:
I think this PR will close https://github.com/embeddings-benchmark/mteb/issues/1593. I wanted to add it too. I can finish this PR if you don't have time for it

mteb/models/search_wrappers.py
NohTow: The corresponding PyLate functions should be quite easy to write!

NohTow:
Thanks for the draft!
I can give an hand/answer any question if needed.
I think it is pretty cool to have the index/modeling decoupled in MTEB, we built our indexes without modeling so it can be used with any kind of models. So IMHO the best way of integrating PyLate into MTEB is to use have a PLAID index and the PyLate model separated, so people can also plug their own models with an encode function leveraging those indexes, for example with ColPali or the model from Orion!

Samoed:
I think we can create plaid search class and inherit pylate models from it

Samoed:
I've created new class for `pylate` index. I've got these results for `colbert-ir/colbertv2.0`

| | New (this PR) | old (v2) |
| - | - | - |
| NanoArguAnaRetrieval | `0.51162` or `0.50584` | 0.50672 |

@NohTow I remember you mentioned that the index is not deterministic somewhere. Is there a way we can improve this? I think there I have large difference between runs, because I'm using `Nano` dataset and on a bigger this shouldn't be a problem


<details>
<summary>Script</summary>

```python
import logging
from pathlib import Path

import mteb

model = mteb.get_model("colbert-ir/colbertv2.0", device="cuda")
tasks = mteb.get_tasks(
    tasks=[
        "NanoArguAnaRetrieval",
    ]
)

evaluator = mteb.MTEB(tasks)
evaluator.run(
    model,
    output_folder="pylate",
    overwrite_results=True,
)
```

</details>

mteb/models/search_wrappers.py
orionw: Ah great idea @Samoed!

orionw:
Thanks so much @Samoed for picking this up 🙏  

@NohTow would know best, but I think it's expected to get different scores when switching from full inference with exact scores for every doc to the new version which uses plaid (not exact inference for speed). I'm not sure why there are two numbers you list though. I would expect the model to be slightly worse, but I suppose the non-exact scoring could make it better?

mteb/models/model_implementations/colbert_models.py
Samoed, line 1: Was renamed

Samoed:
@orionw I have 2 numbers in results, because I've run multiple times and got these 2 scores as final main metrics. I've updated comment to refer this

NohTow:
Hello,

About the difference in scores of full ranking vs PLAID, as pointed out by Orion, PLAID is an approximate search and also apply some quantization, so it is definitely expected to not be able to have 100% matching. Playing around with PLAID parameters (number of candidates, quantization bits, ...) could make it more or less exact search in certain case, but it was obviously not the goal. About having "better scores", you would indeed think that an approximate search should have worst scores, but somehow it can also acts as a regularization and dismiss some candidates that would have scored too high, so it is not really impossible. It is especially true for a small dataset that exhibits high variance, as pointed out by your other results.

So about the variance, this is something that I also do not like but unfortunately was not able to really totally fix. PLAID starts of by doing k-nn, which affect a lot of things downstream (how cells are defined, what base are the residual computed on...).
I believe maybe adding some steps to the k-nn might make the centroïds more deterministic and should reduce variance.
Another possibility would be to try to seed everything to get the same everytime
cc @raphaelsty (because we need to do it for both PLAID and fast-plaid)

mteb/models/model_implementations/pylate_models.py
```diff
@@ -0,0 +1,430 @@
...
+    PromptType,
+    QueryDatasetType,
+    RetrievalOutputType,
+    TopRankedDocumentsType,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class PylateSearchEncoder:
```
KennethEnevoldsen, line 33: Would inherit from the desired protocols here, so we know what it is implementing

KennethEnevoldsen (commented):
I think @Samoed has this one covered, but do feel free to @ me if there is anything :)

mteb/models/model_implementations/pylate_models.py
Samoed: Yes, we can, but in that case it will always pass check on `isinstance` and that this is a good solution. I can add a comment that this implement this protocol

Samoed:
@KennethEnevoldsen So, what should we do? Can we go ahead and merge, or should we wait for the seed fix (or something else)?

mteb/models/model_implementations/pylate_models.py
KennethEnevoldsen: Ohh thought it would fail in that case, but you are right:

```py
class MyProtocol:
    def my_method(self):
        pass


class NotTheProtocol(MyProtocol):
    def another_method(self):
        pass

isinstance(NotTheProtocol(), MyProtocol)  # True

class NotTheProtocol():
    def another_method(self):
        pass

isinstance(NotTheProtocol(), MyProtocol)  # False
```

Then comment or docstring is better

KennethEnevoldsen:
> @KennethEnevoldsen So, what should we do? Can we go ahead and merge, or should we wait for the seed fix (or something else)?

Unsure how big the difference is in practice? If it is less than eps=0.001 (so score of 0.773 and 0.774 are close enough), then we can merge. That is also the tolerance we use in tests.

I would even consider 0.001-0.01 to be reasonable

Samoed:
> Unsure how big the difference is in practice?

In my test on `NanoArguAnaRetrieval` with `colbert-ir/colbertv2.0`, I got `0.51162` and `0.50584` from the table in the comment: https://github.com/embeddings-benchmark/mteb/pull/3183#issuecomment-3304437637. For larger datasets, the difference will probably be smaller, since it can only swap a few top corpus entries for queries, which causes this change.

NohTow:
When I run `GTE-ModernColBERT-v1` on nfcorpus I have (using Fast-Plaid):
0.38033, 0.38033, 0.3801
so definitely within range, but let me check if we can seed things/reduce it even more

Edit: I can confirm that, with the OG PLAID and setting k_means iter to 100 I can get consistent results (like really the exact same)!
I still have different results when using fast-plaid, will investigate.
Unfortunately, I was not able to make seeding work for now. I guess if reproductibility is the key objective there, we can increase the number of kmeans iter, it's _a bit_ slower but should be very fine w.r.t encoding and stuff.
Will continue exploring and report!

Edit2: added some results so you can see the variance for yourself. Raphaël is not available right now, I'll check with him about fast-plaid when he is available

<details>
<summary>10 runs result fast-plaid (default params)</summary>
================================================================================
RESULTS SUMMARY (10 successful runs)
================================================================================
Metric                  Mean          Std          Min          Max
---------------------------------------------------------------
map                   0.1859       0.0003       0.1851       0.1864
ndcg@10               0.3795       0.0009       0.3783       0.3815
ndcg@100              0.3423       0.0006       0.3415       0.3435
recall@10             0.1802       0.0007       0.1788       0.1814
recall@100            0.3232       0.0015       0.3201       0.3261

================================================================================
DETAILED PER-RUN SCORES
================================================================================
Run           map       n@10      n@100       r@10      r@100
-------------------------------------------------------------
1          0.1864     0.3790     0.3427     0.1802     0.3241
2          0.1856     0.3792     0.3420     0.1802     0.3217
3          0.1859     0.3806     0.3435     0.1807     0.3233
4          0.1863     0.3796     0.3421     0.1808     0.3231
5          0.1861     0.3785     0.3428     0.1796     0.3232
6          0.1857     0.3797     0.3415     0.1806     0.3234
7          0.1851     0.3790     0.3424     0.1803     0.3261
8          0.1858     0.3783     0.3424     0.1792     0.3235
9          0.1859     0.3815     0.3421     0.1814     0.3236
10         0.1857     0.3793     0.3415     0.1788     0.3201

================================================================================
95% CONFIDENCE INTERVALS
================================================================================
map            : 0.1859 ± 0.0002 [0.1856, 0.1861]
ndcg@10        : 0.3795 ± 0.0007 [0.3788, 0.3801]
ndcg@100       : 0.3423 ± 0.0004 [0.3419, 0.3427]
recall@10      : 0.1802 ± 0.0005 [0.1796, 0.1807]
recall@100     : 0.3232 ± 0.0011 [0.3222, 0.3243]
</details>

<details>
<summary>10 runs fast-plaid kmeans iter = 100</summary>

================================================================================
RESULTS SUMMARY (10 successful runs)
================================================================================
Metric                  Mean          Std          Min          Max
---------------------------------------------------------------
map                   0.1861       0.0004       0.1854       0.1867
ndcg@10               0.3797       0.0012       0.3773       0.3814
ndcg@100              0.3430       0.0009       0.3412       0.3446
recall@10             0.1805       0.0006       0.1791       0.1813
recall@100            0.3259       0.0026       0.3207       0.3302

================================================================================
DETAILED PER-RUN SCORES
================================================================================
Run           map       n@10      n@100       r@10      r@100
-------------------------------------------------------------
1          0.1863     0.3802     0.3433     0.1806     0.3264
2          0.1863     0.3814     0.3432     0.1812     0.3250
3          0.1867     0.3807     0.3446     0.1808     0.3293
4          0.1854     0.3790     0.3422     0.1804     0.3238
5          0.1864     0.3795     0.3436     0.1805     0.3272
6          0.1865     0.3806     0.3435     0.1808     0.3266
7          0.1856     0.3802     0.3427     0.1791     0.3252
8          0.1858     0.3773     0.3430     0.1801     0.3302
9          0.1863     0.3800     0.3428     0.1813     0.3250
10         0.1859     0.3778     0.3412     0.1802     0.3207

================================================================================
95% CONFIDENCE INTERVALS
================================================================================
map            : 0.1861 ± 0.0003 [0.1858, 0.1864]
ndcg@10        : 0.3797 ± 0.0009 [0.3788, 0.3805]
ndcg@100       : 0.3430 ± 0.0006 [0.3424, 0.3436]
recall@10      : 0.1805 ± 0.0004 [0.1801, 0.1809]
recall@100     : 0.3259 ± 0.0018 [0.3241, 0.3278]

</details>


<details>
<summary>OG PLAID kmeans_iter = 100</summary>

================================================================================
RESULTS SUMMARY (10 successful runs)
================================================================================
Metric                  Mean          Std          Min          Max
---------------------------------------------------------------
map                   0.1866       0.0000       0.1866       0.1866
ndcg@10               0.3792       0.0000       0.3792       0.3792
ndcg@100              0.3432       0.0000       0.3432       0.3432
recall@10             0.1800       0.0000       0.1800       0.1800
recall@100            0.3269       0.0000       0.3269       0.3269

================================================================================
DETAILED PER-RUN SCORES
================================================================================
Run           map       n@10      n@100       r@10      r@100
-------------------------------------------------------------
1          0.1866     0.3792     0.3432     0.1800     0.3269
2          0.1866     0.3792     0.3432     0.1800     0.3269
3          0.1866     0.3792     0.3432     0.1800     0.3269
4          0.1866     0.3792     0.3432     0.1800     0.3269
5          0.1866     0.3792     0.3432     0.1800     0.3269
6          0.1866     0.3792     0.3432     0.1800     0.3269
7          0.1866     0.3792     0.3432     0.1800     0.3269
8          0.1866     0.3792     0.3432     0.1800     0.3269
9          0.1866     0.3792     0.3432     0.1800     0.3269
10         0.1866     0.3792     0.3432     0.1800     0.3269

================================================================================
95% CONFIDENCE INTERVALS
================================================================================
map            : 0.1866 ± 0.0000 [0.1866, 0.1866]
ndcg@10        : 0.3792 ± 0.0000 [0.3792, 0.3792]
ndcg@100       : 0.3432 ± 0.0000 [0.3432, 0.3432]
recall@10      : 0.1800 ± 0.0000 [0.1800, 0.1800]
recall@100     : 0.3269 ± 0.0000 [0.3269, 0.3269]
</details>


Edit3: re seed: actually, I noticed our kmeans should already be seeded, so it's very bizarre we are having different results, and even more so that, for OG PLAID, adding more iterations resolve the issue, because it emphasizes that the difference comes from the kmeans while it is seeded... continuing the exploration

KennethEnevoldsen:
Alright, if a seed is easy to do then we should add it, but this looks very acceptable to me.

NohTow:
Sorry maybe a dumb question, but why are we getting the embeddings of the documents to rerank using the index?
Neither Fast-Plaid and OG PLAID actually implement do not implement the `get_documents_embeddings`, it was a function I created for the now deprecated Voyager index, which is quite slow. We could see if we could add these functions to PLAID indexes, but in any case we are storing a residual quantized version of the embeddings, couldn't we use the actual embeddings?
How is it done usually in MTEB when you do not have indexes (or is it coming from the fact that we actually use an index and so when reranking with an index you always get the embeddings from the index?)

Samoed:
That's a good question. For sentence transformer models, we just compute similarity between query and document embeddings https://github.com/embeddings-benchmark/mteb/blob/59f6c01c3f86ef2e1a94923d20e320a9a2f7ce2e/mteb/models/search_wrappers.py#L237-L253 For plaid index I don't have ideas how to make retrieve with specified `ids` without taking embeddings from index. We can use embedding directly, as was originally done in this PR https://github.com/embeddings-benchmark/mteb/blob/0c232d95da93c3ff5bd8984d1c0c2f3b97369245/mteb/models/search_wrappers.py#L413-L428. I just wanted to simplify working with index a bit

I think you can add reranking functions to PLAID and this would be a better solution

NohTow:
> I think you can add reranking functions to PLAID and this would be a better solution

So this is a possibility, but as I said, the embeddings are unfortunately not stored "as is" within the index (to allow smaller footprint). I can see with Raphaël to add those features to the indexes (actually we would need them to rebuild the indexes), but I am not sure this will be trivial (as in added tomorrow), so maybe we can start with using the raw embeddings and do another PR to use the index when we add this feature in PyLate if that works for you?

Samoed:
Do you want to use raw only for reranking or for the retrieval too?

NohTow:
> Do you want to use raw only for reranking or for the retrieval too?

Only for reranking, it is impossible to do retrieval on large datasets without indexes! Reranking should be fine no?

Samoed:
I think so

mteb/models/model_implementations/pylate_models.py
```diff
@@ -0,0 +1,446 @@
...
+
+        if self.task_corpus is None:
+            raise ValueError("Corpus must be set before reranking.")
+
+        result_heaps = {qid: [] for qid in query_idx_to_id.values()}
+        doc_id_to_idx = {doc["id"]: idx for idx, doc in enumerate(self.task_corpus)}
+
+        all_doc_embeddings = self.encode(
+            create_dataloader(
+                self.task_corpus,
```
Samoed, line 221: @KennethEnevoldsen Maybe we can to pass `corpus` to search function, because we're using in search encoder and cross encoder too

NohTow:
Re variance:
I did some tests and when using OG PLAID with faiss to compute the kmeans, I can get consistent results (the seed defined in the code is enough)
However, when using fast kmeans, the centroïds are inconsistent, which leads to some variance in the results.
As fast-plaid is using fast kmeans, it means we cannot get 100% reproduction until this is fixed. I reached out to Ben to ask him to fix this in fast kmeans. I'll try to see whether there is also other source of variance in fast plaid than just the kmeans!

Edit: so I had to learn a bit of Rust, but I found out that fast-plaid is missing one seeding, Raphaël will be updating it ASAP.
When this seed is fixed, if we use Faiss kmeans, there is 0 variance!
However, even with this fix, since fast-kmeans is not deterministic, there is some variance, so we'll have to wait for Ben to fix it to have 100% reproductibility!

Edit2: I tracked the origin of the non determinism in fast kmeans! Everything is actually correctly seeded, the differences arise from using triton kernels, which race conditions can affect results!
So I'll discuss with Ben to see whether we want to make this kernel deterministic, else what we will have to do is expose a use_triton parameters into PLAID indexes from Pylate, also expose it in fast plaid and forward it to fast kmeans, and set it to False when being using in MTEB to make variance equals to 0 (at the cost of a _tad_ slower processing)!

<details>
<summary>10 runs result fast-plaid without tritons kernels and fix to rust seeding(default params, not any additional seeding than defaults)</summary>
================================================================================
RESULTS SUMMARY (10 successful runs)
================================================================================
Metric                  Mean          Std          Min          Max
---------------------------------------------------------------
map                   0.1860       0.0000       0.1860       0.1860
ndcg@10               0.3796       0.0000       0.3796       0.3796
ndcg@100              0.3418       0.0000       0.3418       0.3418
recall@10             0.1802       0.0000       0.1802       0.1802
recall@100            0.3223       0.0000       0.3223       0.3223

================================================================================
DETAILED PER-RUN SCORES
================================================================================
Run           map       n@10      n@100       r@10      r@100
-------------------------------------------------------------
1          0.1860     0.3796     0.3418     0.1802     0.3223
2          0.1860     0.3796     0.3418     0.1802     0.3223
3          0.1860     0.3796     0.3418     0.1802     0.3223
4          0.1860     0.3796     0.3418     0.1802     0.3223
5          0.1860     0.3796     0.3418     0.1802     0.3223
6          0.1860     0.3796     0.3418     0.1802     0.3223
7          0.1860     0.3796     0.3418     0.1802     0.3223
8          0.1860     0.3796     0.3418     0.1802     0.3223
9          0.1860     0.3796     0.3418     0.1802     0.3223
10         0.1860     0.3796     0.3418     0.1802     0.3223

================================================================================
95% CONFIDENCE INTERVALS
================================================================================
map            : 0.1860 ± 0.0000 [0.1860, 0.1860]
ndcg@10        : 0.3796 ± 0.0000 [0.3796, 0.3796]
ndcg@100       : 0.3418 ± 0.0000 [0.3418, 0.3418]
recall@10      : 0.1802 ± 0.0000 [0.1802, 0.1802]
recall@100     : 0.3223 ± 0.0000 [0.3223, 0.3223]
</details>

Edit3: we'll be making the changes to FastPlaid and PyLate today, so you'll be able to get stable results by setting use_triton to false when creating the PyLate index soon!

mteb/models/model_implementations/pylate_models.py
KennethEnevoldsen: can that just be passed to self during the indexing?

KennethEnevoldsen:
Great to see these changes!

mteb/models/model_implementations/pylate_models.py
Samoed: Yes, it made like this now, but this is a bit not clear

KennethEnevoldsen: Hmm, that would probably be what I suspected would happen.

The object itself is free to build a vector store or something

NohTow:
Hello,
We released new versions for Fast-Plaid and PyLate
You should be able to get 100% reproductibility right now by just updating and feeding use_triton=False when creating the PLAID index!

raphaelsty:
Hi everyone, following @NohTow comment, the `1.3.2` version of PyLate which we just released should do the job :)

Samoed:
I've updated implementation to disable triton kernels

NohTow:
Please tell me if you still have any issue with reproducibility!
Also do not hesitate to ping if you need anything, getting embeddings from the index is not planned in the next days, but else happy to help!

raphaelsty:
Sorry forgot to push the `1.3.2` version of PyLate on pypi, it's done now :)

Samoed:
I run multiple times and scores are matching now! Thanks for the fix!

NohTow:
Is there anything still blocking on PyLate side?
i'll be off next week so won't be able to make updates, but if there is something I can change today I can give it a shot

Samoed:
I don't think so. I'm waiting for @KennethEnevoldsen. Thank you for helping!

NohTow:
Hello,

Back from vacation I figured I should give it a try to make sure everything was alright.
So the boilerplates on the readmes seems out to date (?), because I cannot use get_model and get_tasks, but the CLI seems to run:
`mteb run -m lightonai/GTE-ModernColBERT-v1 -t NFCorpus` works but surprisingly gives non deterministic results: "ndcg_at_10": 0.37692," "ndcg_at_10": 0.37717," while PyLate consistently gives the same value (which is also a bit higher): 38.02.

For Scifact, PyLate gives 76.46 while mteb gives "ndcg_at_10": 0.7589, seems to be consistent though


Any idea why the results seem to be *a bit* variable?
Also, any idea why the results between PyLate directly and MTEB are different?

I can explore the query length and document length, but I would need to be able to run the evaluation programmatically, could you point me towards the latest working boilerplates?
Or where the CLI code is actually (I tried the one I found but it is not running somehow)

Samoed:
Hm, I didn't get any errors when run in main or `v2` branches
```
import mteb

m = mteb.get_model_meta("lightonai/GTE-ModernColBERT-v1")
```

>  Also, any idea why the results between PyLate directly and MTEB are different?

This is hard to tell without your implementation

NohTow:
I managed to hack the CLI so I think I can get code to work/change things if needed
I do not think the results are non deterministic, it's prolly just rounding erros that happens sometimes (most of the time the results are the same and the diff is very small)

About the implementation, I am using our evaluation [boilerplate](https://github.com/lightonai/pylate/blob/main/examples/evaluation/beir_dataset.py) for BEIR.
The query length is different for scifact, but should be the default one for nfcorpus and the document length is correct.
I am using the beir package directly, I can give a shot to [my custom script using MTEB data](https://github.com/lightonai/pylate/blob/main/examples/evaluation/longembed_dataset.py) to see if it comes from this

Also, is RTEB supposed to be up to date and working on the v2.0.0 branch? I have errors because document_type is "c" for some splits (code I suppose), and also the results on the running splits are somewhat bad (this could be the expected results, but since I am not sure).

Samoed:
I tried to run in colab https://colab.research.google.com/drive/1O5sNuooFiXAYwQNRLgG6SYfqPG4dOnCL?usp=sharing and imports fine, but I got `"ndcg_at_10": 0.25668,` for `NFCorpus` and `"ndcg_at_10": 0.73017,` for `SciFact`. I run 2 times and scores are the same

> About the implementation, I am using our evaluation [boilerplate](https://github.com/lightonai/pylate/blob/main/examples/evaluation/beir_dataset.py) for BEIR.

I don't think that there is much difference, but we have more code, but most of the logic still in model implementation

> Also, is RTEB supposed to be up to date and working on the v2.0.0 branch? I have errors because document_type is "c" for some splits (code I suppose), and also the results on the running splits are somewhat bad (this could be the expected results, but since I am not sure).

Yes, it should be up-to-date. Can you open an issue with full error trace and I'll investigate this

NohTow:
Hello,

I tried various things:
- Code that use MTEB data but use our inference/index code
The resulting NDCG on SciFact is 76.34 (consistent) which seems closer to what I have with BEIR.
<details>
<summary>Corresponding code</summary>

```python

import mteb

from pylate import evaluation, indexes, models, retrieve

if __name__ == "__main__":
    tasks = mteb.get_tasks(
        tasks=[
            "SciFact",
        ]
    )
    for task in tasks:
        task.load_data()
        model_name = "lightonai/GTE-ModernColBERT-v1"
        model = models.ColBERT(
            model_name_or_path=model_name,
        )
     
        index = indexes.PLAID(
            override=True,
            use_triton=False,
            use_fast=True,
        )

        retriever = retrieve.ColBERT(index=index)

        documents = [elem['title'] + " " + elem['text'] if "title" in elem else elem['text'] for elem in task.dataset["default"]["test"]["corpus"]]
       
        documents_embeddings = model.encode(
            sentences=documents,
            batch_size=10,
            is_query=False,
            show_progress_bar=True,
        )
        documents_ids = [elem["id"] for elem in task.dataset["default"]["test"]["corpus"]]
        index.add_documents(
            documents_ids=documents_ids,
            documents_embeddings=documents_embeddings,
        )
        queries = [elem['text'] for elem in task.dataset["default"]["test"]["queries"]]
        queries_embeddings = model.encode(
            sentences=queries,
            is_query=True,
            show_progress_bar=True,
            batch_size=32,
        )

        scores = retriever.retrieve(queries_embeddings=queries_embeddings)

        evaluation_scores = evaluation.evaluate(
            scores=scores,
            qrels=task.dataset["default"]["test"]["relevant_docs"],
            queries=[elem['id'] for elem in task.dataset["default"]["test"]["queries"]],
            metrics=[
                "map",
                "ndcg@1",
                "ndcg@10",
                "ndcg@100",
                "recall@10",
                "recall@100",
            ],
        )

        print(evaluation_scores)
```
</details>

- Code using MTEB fully (without the CLI) based off your code
The resulting NDCG is 75.89 (consistent), which seems to be the same value than with the CLI (thankfully).

<details>
<summary>Corresponding code</summary>

```python
import mteb
tasks = mteb.get_tasks(["SciFact"])
model = mteb.get_model("lightonai/GTE-ModernColBERT-v1")
evaluator = mteb.MTEB(tasks)
evaluator.run(model, co2_tracker=False)
```
</details>


So there seems to be an "issue" in the MTEB code that is not related to data


About your results on collab, I'll try creating a new env from scratch and see how it goes!

> Can you open an issue with full error trace and I'll investigate this

I can open an issue, but essentially some of the datasets have document_type=="c" (code), and so it raises a ValueError because it can't handle queries type "c" (note that the Value Error message is wrong because it mentions queries while it is about documents). You can see that handling "c" is not handled [there](https://github.com/embeddings-benchmark/mteb/blob/9325dfaef80fce6291681f46645243f976cd58df/mteb/create_dataloaders.py#L307)

Samoed:
> I can open an issue, but essentially some of the datasets have document_type=="c" (code), and so it raises a ValueError because it can't handle queries type "c" (note that the Value Error message is wrong because it mentions queries while it is about documents). You can see that handling "c" is not handled [there](https://github.com/embeddings-benchmark/mteb/blob/9325dfaef80fce6291681f46645243f976cd58df/mteb/create_dataloaders.py#L307)

I see, maybe I've merged incorrectly. I will fix it
UPD created PR https://github.com/embeddings-benchmark/mteb/pull/3277

Samoed:
> Code using MTEB fully (without the CLI) based off your code

This is strange that we have 3% difference in results

NohTow:
About your issue in collab, I found the root.
You are not installing flash attention, which makes query expansion bugged for this model. I fixed this in a latest release (originally to fix cpu inference) and the fix is in the release you are using BUT somehow, when using MTEB to load the model, it uses an older version of the _model_ where the config parameter has not been added.
More surprisingly, when I try to add the revision manually, it says that the revision is not found and refer to a previous commit
```ValueError: Model revision c1647d10d6edc6f70837f42f0a978f2df53f51dd not found for model lightonai/GTE-ModernColBERT-v1. Expected 78d50a162b04dfdc45c3af6b4294ba77c24888a3.```
image


When loading the model in PyLate, the updated readme is downloaded and so the param is correctly set and it works, even without FA.

Edit: it seems to be because ```self.model = ColBERT(self.model_name, revision=revision, **kwargs)``` receive the commit 78d

Edit2: ok this is actually because of this:

```lightonai__gte_moderncolbert_v1 = ModelMeta(
    loader=MultiVectorModel,
    name="lightonai/GTE-ModernColBERT-v1",
    languages=[
        "eng-Latn",
    ],
    open_weights=True,
    revision="78d50a162b04dfdc45c3af6b4294ba77c24888a3",```



Should be updated with the latest commit, will do a PR

NohTow:
> This is strange that we have 3% difference in results

Yeah I am not really sure why but it seems like there is a noticeable difference that I would like to fix
I'll try to see if I can find anything, but my boilerplate using the MTEB data seems pretty staightforward so I am not sure why the mteb code would be different

NohTow:
Found this issue (at least partly)
In the encode function, the inputs are always only the [text field of the data](https://github.com/embeddings-benchmark/mteb/blob/32975796b1b3fc08aa16b2a9a341829097e0c119/mteb/models/model_implementations/pylate_models.py#L324), missing out the potential titles
Adding titles seems to partly close the gap

I still have a bit of a gap though: 75.98 & 37.37 vs 76.34 & 37.96

Samoed:
This shouldn't be the case, because title and text are combine and passed as text https://github.com/embeddings-benchmark/mteb/blob/32975796b1b3fc08aa16b2a9a341829097e0c119/mteb/create_dataloaders.py#L46-L55

NohTow:
Yeah actually the results were still at the same relative gap, sorry!
But then I really have no clue what could be the issue :(

NohTow:
Ok I think I might have a good lead
When I [do not convert to tensor to then cast to cpu/numpy](https://github.com/embeddings-benchmark/mteb/blob/55de95eff3503108f7f2a16cb5ef9976766d6063/mteb/models/model_implementations/pylate_models.py#L330) (which I do not do in my usual scripts), the results are way closer:
76,35 37.92 vs 76,34 38,03
Is there any reason why you did this casting or we can remove it (as it is handled by PyLate indexes and just add rounding noise)?
It is still not 100% equals, which is frustrating, but at this point I am not sure what could be different tbh

It should also be noted that the batch size has a slight impact, when I remove the batch_size=10 from the kwargs, I get 76,39 and 37,84. It's fairly minor and in any case the evaluation batch size should always be the same

Samoed:
Very unexpected! I think we return as tensor because we're using later `torch.nn.utils.rnn.pad_sequence`, but maybe this is irrelevant now original comment to this in main https://github.com/embeddings-benchmark/mteb/blob/30de6197c5acd4fbfe3e6f452aba796bf6d08b9e/mteb/models/colbert_models.py#L100-L106

NohTow:
The padding thing is handled by our indexes, by removing the cast to tensor -> pad -> .cpu().numpy() (just not returning tensor and directly returning pred), it works great

So essentially remove the convert_to_tensor param, and directly returning the output is fine!
There is is still a delta, but I am not finding why... even with the same bs and the same k for retrieve, it seems to be a bit different

Samoed:
Let's continue this discussion in https://github.com/embeddings-benchmark/mteb/issues/3283

Issue #3163, opened 2025-09-09
Datasets contain `None` values in certain columns, which causes the dataloader to fail


Samoed:
I found that some datasets contain `None` values in certain columns, which causes the dataloader to fail when attempting to process them. To try to solve this issue, we can select only the columns that are required for the task in `_evaluate_subset`. However, I am not sure if this approach is the best solution

_Originally posted by @Samoed in https://github.com/embeddings-benchmark/mteb/issues/3156#issuecomment-3267108546_

KennethEnevoldsen:
I think it is a good solution - what are your worries?

Samoed:
I don't have worries. Just created issue to not forget this change

PR #3190, opened 2025-09-17
[v2] Select required columns to run

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3163

### DIFF
--- a/mteb/abstasks/AbsTaskAnyClustering.py
+++ b/mteb/abstasks/AbsTaskAnyClustering.py
@@ -76,8 +76,8 @@ class AbsTaskAnyClustering(AbsTask):
         ):
             v_measures = []
             for cluster_set in tqdm.tqdm(dataset, desc="Clustering"):
-                clustering_dataset = Dataset.from_dict(cluster_set).rename_column(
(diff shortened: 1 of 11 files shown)


mteb/abstasks/AbsTaskAnyClustering.py
```diff
@@ -76,8 +76,8 @@ def _evaluate_subset(
         ):
             v_measures = []
             for cluster_set in tqdm.tqdm(dataset, desc="Clustering"):
-                clustering_dataset = Dataset.from_dict(cluster_set).rename_column(
-                    original_column_name="sentences", new_column_name="text"
+                clustering_dataset = Dataset.from_dict(cluster_set).select_columns(
```
KennethEnevoldsen, line 79: do we still want to do the rename as well?

Samoed: No, we don't need to add a rename here, because the `evaluator` will select the required column in `create_dataloader`. Pushed fix for this

KennethEnevoldsen: check, works for me

Issue #3117, opened 2025-09-02
New contributors CI don't work


Samoed:
### Describe the bug

Ref https://github.com/embeddings-benchmark/mteb/pull/3115

### To reproduce

https://github.com/embeddings-benchmark/mteb/actions/runs/17403514660/job/49402152389
CC @isaac-chung 

### Are you interested to contribute a fix for this bug?

No

isaac-chung:
It's odd that the first time contributor tag is missing.

isaac-chung:
Let's monitor it for 2 more weeks. Happy to revert if it still doesn't work.

KennethEnevoldsen:
Hmm have we seen cases where it worked?

isaac-chung:
Doesn't seem so. Might be a bug related to PR/PR request?

Samoed:
No. We have new contributors, but CI wasn't triggered. For example, in https://github.com/embeddings-benchmark/mteb/pull/3205

PR #3193, opened 2025-09-20
[v2] Merge main 20 09

Samoed:


.github/workflows/welcome_new_comers.yml
```diff
@@ -0,0 +1,40 @@
+name: Welcome New Contributors
```
KennethEnevoldsen, line 1: Ahh didn't see this one - looking forward to seeing it in action

docs/benchmarks.md
KennethEnevoldsen, line 1: unsure if we should delete these in v2

.github/workflows/welcome_new_comers.yml
Samoed: Sadly it's not working https://github.com/embeddings-benchmark/mteb/issues/3117

KennethEnevoldsen (approved):
Looks good no particular worries

docs/benchmarks.md
Samoed: I think we can delete this, because we have site with all documentation

Issue #2273, opened 2025-03-06
Remove `create_metadata`


Samoed:
Since we've changed results loading to use the `results` repository instead of `README`, should we remove `create_metadata` to prevent user confusion?

KennethEnevoldsen:
I think it is good to keep, but we should update it to be more relevant, e.g., only use its benchmark scores (mean on MTEB(eng, v2)). 

What about something like:
```
mteb create_metadata --benchmarks "MTEB(eng, v2)" "MTEB(eng, v1)" \
 --tasks MiraclRetrieval # default to None, but you can optionally tasks that you are interested in
```
 
Could use implementation from #2038

Samoed:
I think we can create a table in readme with scores on mteb and add main scores of tasks and benchmarks to the `frontmatter`

PR #3199, opened 2025-09-20
[v2] Refactor `create metadata`

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/2273

- Now model card data would have only `main_scores`
- Refactored the CLI command from `create-meta` to `create-model-results`, added `--model-name`, `--tasks`, `--benchmarks`. The CLI now supports merging results with existing model cards from either a file or a Hugging Face model ID.
- Refactored the dataset card metadata generation in `mteb/abstasks/task_metadata.py` by modularizing language, license, task type, and subtype handling into separate helper methods (`_hf_task_type`, `_hf_task_category`, `_hf_subtypes`, `_hf_languages`, `_hf_license`). 

Example repo with new model card https://huggingface.co/Samoed/test_model

### DIFF
--- a/tests/create_meta/model_card_gold_existing.md
+++ b/tests/create_meta/model_card_gold_existing.md
@@ -499,12 +499,3 @@ We sampled each dataset given a weighted probability which configuration is deta
 | [SQuAD2.0](https://rajpurkar.github.io/SQuAD-explorer/) | [paper](https://aclanthology.org/P18-2124.pdf) | 87,599 |
 | [TriviaQA](https://huggingface.co/datasets/trivia_qa) | - | 73,346 |
 | **Total** | | **1,170,060,424** |
-
-
-
-# MTEB results
-| task_name            |   all-MiniLM-L6-v2 |
-|:---------------------|-------------------:|
-| BSARDRetrieval       |           0        |
-| BornholmBitextMining |           0.296813 |
-| STS22                |           0.408369 |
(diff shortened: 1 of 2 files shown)


mteb/cli/build_cli.py
```diff
@@ -226,32 +226,70 @@ def _add_run_parser(subparsers: argparse._SubParsersAction) -> None:
...
-
-    with output_path.open("w") as f:
-        f.write(frontmatter)
+    tasks = []
+    if tasks_names is not None:
+        tasks = mteb.get_tasks(tasks_names)
+    if benchmarks is not None:
+        benchmarks = mteb.get_benchmarks(benchmarks)
+        for benchmark in benchmarks:
+            tasks.extend(benchmark.tasks)
```
KennethEnevoldsen, line 252: Hmm do we want to merge benchmarks scores when we add it to the card so that you would get `mteb(eng, v2) (mean): 0.61`

Samoed:
> users would be more interested in benchmark scores that task specific scores.

I agree, but I think we should update `BenchmarkResults` to output results of benchmark itself. 

Something like
```python
results: BenchmarkResult = load_results(models, tasks)

models_scores = results.get_benchmark_result()
# {"intfloat/e5-small": 0.5}
results.benchmark_name
# MTEB(eng, v2)
```
And with these changes will be easy to add benchmark scores

KennethEnevoldsen:
agree

Issue #2515, opened 2025-04-07
[v2] make a clear seperation of public and non-public functions


KennethEnevoldsen:
We currently do not have a clear separation of public functions and non-public functions. This makes it hard to make changes that are non-breaking.

@Samoed would like your thoughts on this

Samoed:
We can make it, but this is not clear what is public or private.

Obiously:
1. AbsTask.evaluate
2. Benchmark __init__, run, but that would be changed in https://github.com/embeddings-benchmark/mteb/issues/1272

But for other parts, I'm not sure which parts are `public` for us. Are we able to call evaluators or tasks directly?

KennethEnevoldsen:
Yeah, it will probably require some discussion:

- Evaluators are probably a big one. I would like them to be private, but I can see that some would like to call them.
- Tasks: I would keep them public. Basically the same as `get_task`, but all task utility function can def. be private

Samoed:
Do we want to add new functions here?

KennethEnevoldsen:
Do you have something we should add? Any sections we want to refactor?

Samoed:
No, I don't think so

KennethEnevoldsen:
I will let you merge yours in and then I will also look over and then we can close this

PR #3200, opened 2025-09-20
make more functions private

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/2515

- `cli` -> `_cli`
- `languages` -> `_languages`
- Made some helper functions in `abstasks` and `overview` private


mteb/abstasks/AbsTask.py
```diff
@@ -14,8 +14,8 @@
 from datasets import Dataset, DatasetDict
 from sklearn.preprocessing import MultiLabelBinarizer
 
+from mteb._languages import LanguageScripts
```
KennethEnevoldsen: It should be valid for the user to import this - I would not make it private

mteb/cli/__init__.py
KennethEnevoldsen, line 1: I am unsure about the CLI. I think there are a few cases where it is legitimate to call the CLI function. 

e.g. `run` so I wouldn't make the entire module private

KennethEnevoldsen (commented):
would revert cli + languages, otherwise all good

mteb/cli/__init__.py
Samoed: I don't think that `run` will be used by someone, because it's required to pass `Namesapce` and simpler to use evaluate then

mteb/abstasks/AbsTask.py
Samoed: Rolled back

KennethEnevoldsen:
@Samoed feel free to merge this

PR #3218, opened 2025-09-27
fix: prevent EOS token truncation

### LINKED ISSUE #3217: `EOS` token is lost on truncation for last-token pooling
The recently added **`BMRetriever`** models rely on **last-token pooling**, which uses the embedding of the final `EOS` token to generate sentence representations. This is critical for their performance.

However, the current `InstructSentenceTransformerWrapper` implementation handles truncation in a way that can inadvertently remove this `EOS` token. The wrapper appends the `eos_token` string to the input *before* tokenization. If the input sentence's length exceeds `max_seq_length`, the tokenizer truncates the end of the string, cutting off the `EOS` token along with it.

This causes the `BMRetriever` model to perform pooling on the last word token of the truncated sequence instead of the `EOS` token. As a result, the generated embeddings are incorrect, which invalidates the model's performance evaluation in the MTEB benchmark.

The following code snippet demonstrates how the `EOS` token is dropped when an input is too long for the `BMRetriever` model wrapper:

```python
import mteb
import torch

# 1. Load the BMRetriever model using the MTEB wrapper
model_name = "BMRetriever/BMRetriever-7B"
# Set a short max_seq_length to easily reproduce the issue
model_wrapper = mteb.get_model(
    model_name,
    max_seq_length=10,
    model_args={"torch_dtype": torch.bfloat16}
)

# 2. Prepare a long sentence and simulate adding the EOS token
long_sentence = "This is a very long sentence designed to exceed the max sequence length."
# The internal logic in InstructSentenceTransformerWrapper is:
# sentences = [example + self.model.tokenizer.eos_token for example in sentences]
sentence_with_eos = [long_sentence + model_wrapper.model.tokenizer.eos_token]
print(f"Before : {sentence_with_eos[0]}")

# 3. Tokenize the sentence (this happens inside model_wrapper.encode)
# We call tokenize directly to inspect the result.
tokens_after_truncation = model_wrapper.model.tokenize(sentence_with_eos)
decoded_tokens = model_wrapper.model.tokenizer.decode(
    tokens_after_truncation["input_ids"][0],
    skip_special_tokens=True
)
print(f"After  : {decoded_tokens}")
```

Execution output:
```
Before : This is a very long sentence designed to exceed the max sequence length.</s>
After  : This is a very long sentence designed to exceed
```

As shown, the critical </s> token required by BMRetriever is missing from the final tokenized input.

For `BMRetriever `to function correctly, the `EOS` token must always be the final token in the sequence, even after truncation. The tokenization logic should prioritize keeping the `EOS` token at the end.

An ideal tokenized output (conceptually) would look something like this:

`['This', 'is', 'a', 'very', 'long', 'sentence', 'designed', 'to', 'exceed', '</s>']`

whybe-choi:
Close #3217

This pull request enhances the `BMRetrieverWrapper` model in `bmretriever_models.py` by adding a new `encode` method that supports eos token handling, and updates the default model revision to a newer version.

### DIFF
--- a/mteb/models/bmretriever_models.py
+++ b/mteb/models/bmretriever_models.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+from collections.abc import Sequence
 from functools import partial
 from typing import Any, Callable
 
+import numpy as np
 import torch
 from sentence_transformers import SentenceTransformer
 from sentence_transformers.models import Pooling, Transformer
@@ -55,6 +57,48 @@ class BMRetrieverWrapper(InstructSentenceTransformerWrapper):
 
         if padding_side is not None:
             self.model.tokenizer.padding_side = padding_side
+
+    def encode(
+        self,
+        sentences: Sequence[str],
+        *,
+        task_name: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> np.ndarray:
+        instruction = self.get_task_instruction(
+            task_name, prompt_type, self.prompts_dict
+        )
+        # to passage prompts won't be applied to passages
+        if (
+            not self.apply_instruction_to_passages
+            and prompt_type == PromptType.document
+        ):
+            instruction = None
+
+        if instruction:
+            sentences = [instruction + sentence for sentence in sentences]
+
+        if self.add_eos_token:
+            batch_tokens = self.model.tokenizer(
+                sentences, truncation=True, max_length=self.model.max_seq_length - 1
+            )
+            decoded_sentences = self.model.tokenizer.batch_decode(
+                batch_tokens["input_ids"], skip_special_tokens=True
+            )
+            sentences = [
+                sentence + self.model.tokenizer.eos_token
+                for sentence in decoded_sentences
+            ]
+
+        embeddings = self.model.encode(
+            sentences,
+            **kwargs,
+        )
+
+        if isinstance(embeddings, torch.Tensor):
+            embeddings = embeddings.cpu().detach().float().numpy()
+        return embeddings
 
 
 # https://huggingface.co/datasets/BMRetriever/biomed_retrieval_dataset
@@ -158,7 +202,7 @@ BMRetriever_7B = ModelMeta(
     loader=partial(
         BMRetrieverWrapper,
         model_name="BMRetriever/BMRetriever-7B",
-        config_args={"revision": "e3569bfbcfe3a1bc48c142e11a7b0f38e86065a3"},
+        config_args={"revision": "13e6adb9273c5f254e037987d6b44e9e4b005b9a"},
         model_args={"torch_dtype": torch.float32},
         instruction_template=instruction_template,
         padding_side="left",
@@ -168,7 +212,7 @@ BMRetriever_7B = ModelMeta(
     name="BMRetriever/BMRetriever-7B",
     languages=["eng-Latn"],
     open_weights=True,
-    revision="e3569bfbcfe3a1bc48c142e11a7b0f38e86065a3",
+    revision="13e6adb9273c5f254e037987d6b44e9e4b005b9a",
     release_date="2024-04-29",
     embed_dim=4096,
     n_parameters=7_110_660_096,
--- a/mteb/models/bmretriever_models.py
+++ b/mteb/models/bmretriever_models.py
@@ -1,10 +1,8 @@
 from __future__ import annotations
 
-from collections.abc import Sequence
 from functools import partial
 from typing import Any, Callable
 
-import numpy as np
 import torch
 from sentence_transformers import SentenceTransformer
 from sentence_transformers.models import Pooling, Transformer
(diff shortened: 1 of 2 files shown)


Samoed:
In their tokenizer config stated `"add_eos_token": false,`, so I'm not really sure if this model is need `EOS`

https://huggingface.co/BMRetriever/BMRetriever-7B/blob/main/tokenizer_config.json#L3

UPD see in their example
```
# Important! Adding EOS token at the end
batch_dict['input_ids'] = [input_ids + [tokenizer.eos_token_id] for input_ids in batch_dict['input_ids']]
```

mteb/models/bmretriever_models.py
```diff
@@ -56,6 +58,48 @@ def __init__(
...
+        if (
+            not self.apply_instruction_to_passages
+            and prompt_type == PromptType.document
+        ):
+            instruction = None
+
+        if instruction:
+            sentences = [instruction + sentence for sentence in sentences]
+
+        if self.add_eos_token:
```
Samoed: You can just can update tokenizer like
```python
tokenizer = AutoTokenizer.from_pretrained("BMRetriever/BMRetriever-7B", add_eos_token=True)
```
Also, I think it would be better to also update `InstructSentenceTransformer` class. To pass to `SentenceTransformers`
```python
model = SentenceTransformer(model_name, tokenizer_kwargs=dict(add_eos_token=True))
```

whybe-choi:
> In their tokenizer config stated `"add_eos_token": false,`, so I'm not really sure if this model is need `EOS`
> 
> https://huggingface.co/BMRetriever/BMRetriever-7B/blob/main/tokenizer_config.json#L3

But, the usage section in the model card adds the EOS token like this:
```python
# Tokenize the input texts
batch_dict = tokenizer(input_texts, max_length=max_length-1, padding=True, truncation=True, return_tensors='pt')

# Important! Adding EOS token at the end
batch_dict['input_ids'] = [input_ids + [tokenizer.eos_token_id] for input_ids in batch_dict['input_ids']]
batch_dict = tokenizer.pad(batch_dict, padding=True, return_attention_mask=True, return_tensors='pt').to("cuda")
```

Samoed:
Yes, I've found this too. (updated comment)

mteb/models/bmretriever_models.py
Samoed: To pass to `Transformers` use `tokenizer_args`
https://github.com/UKPLab/sentence-transformers/blob/7341bf155b4349b88690b78c84beb5aa658c439f/sentence_transformers/models/Transformer.py#L70

whybe-choi: How about this code?

```diff
--- a/mteb/models/instruct_wrapper.py
+++ b/mteb/models/instruct_wrapper.py
@@ -48,12 +48,13 @@
         self.model = SentenceTransformer(model_name, revision=revision, **kwargs)
         self.instruction_template = instruction_template
         self.apply_instruction_to_passages = apply_instruction_to_passages
-        self.add_eos_token = add_eos_token
         self.prompts_dict = prompts_dict
         if max_seq_length is not None:
             self.model.max_seq_length = max_seq_length
         if padding_side is not None:
             self.model.tokenizer.padding_side = padding_side
+        if add_eos_token:
+            self.model.tokenizer.add_eos_token = True
 
     def encode(
         self,
@@ -64,12 +65,6 @@
         prompt_type: PromptType | None = None,
         **kwargs: Any,
     ) -> np.ndarray:
-        if self.add_eos_token:
-            sentences = [
-                example + self.model.tokenizer.eos_token for example in sentences
-            ]
-
         instruction = self.get_task_instruction(
             task_name, prompt_type, self.prompts_dict
         )
```

Samoed: Yes! But I think this can be updated a bit.
```python
tokenizer_params = {}
if add_eos_token:
    tokenizer_params["add_eos_token"] = add_eos_token
if max_seq_length is not None:
    # https://github.com/UKPLab/sentence-transformers/blob/7341bf155b4349b88690b78c84beb5aa658c439f/sentence_transformers/models/Transformer.py#L115
    tokenizer_params["model_max_length"] = max_seq_length
if padding_side is not None:
    tokenizer_params["padding_side"] = padding_side

kwargs.setdefault("tokenizer_kwargs", {}).update(tokenizer)

self.model = SentenceTransformer(model_name, revision=revision, **kwargs)
```
Also, other tokenizer arguments can be converted this way

whybe-choi: I'm wondering if it would be better to work on these changes in a new PR. And I think this is what you mean:

```python
if add_eos_token:
    if "tokenizer_kwargs" not in kwargs:
        kwargs["tokenizer_kwargs"] = {}
    kwargs["tokenizer_kwargs"].update(add_eos_token=add_eos_token)

if padding_side is not None:
    if "tokenizer_kwargs" not in kwargs:
        kwargs["tokenizer_kwargs"] = {}
    kwargs["tokenizer_kwargs"].update(padding_side=padding_side)

self.model_name = model_name
self.model = SentenceTransformer(model_name, revision=revision, **kwargs)
self.instruction_template = instruction_template
self.apply_instruction_to_passages = apply_instruction_to_passages
self.prompts_dict = prompts_dict

if max_seq_length is not None:
    self.model.max_seq_length = max_seq_length
```

Samoed: > I'm wondering if it would be better to work on these changes in a new PR

I don't think so, because we're fixing `eos` token in tokenize still. You can just update title of the PR. I updated code in my comment to make it more straightforward

whybe-choi: Thanks! I've committed the changes reflecting the feedback.

Samoed:
Thank you for fix!

PR #3222, opened 2025-09-29
Feat rteb sider UI

KennethEnevoldsen:
fixes to #3212

reran it locally as it seems like the current image it outdated

 - removed RTEB
 - renamed to general purpose
 - renamed task to "{prev_name} Retrieval" to prevent future bugs in people removing the "_". If we want to use the space we probably need to add a test to ensure distinct display names

### DIFF
--- a/mteb/benchmarks/benchmarks/rteb_benchmarks.py
+++ b/mteb/benchmarks/benchmarks/rteb_benchmarks.py
@@ -12,7 +12,7 @@ RTEB_CITATION = r"""@article{rteb2024,
 
 RTEB_MAIN = Benchmark(
     name="RTEB(beta)",
-    display_name="RTEB Retrieval Embedding Benchmark",
+    display_name="Retrieval Embedding Benchmark",
     icon="https://github.com/DennisSuitters/LibreICONS/raw/2d2172d15e3c6ca03c018629d60050e4b99e5c55/svg-color/libre-gui-search.svg",
     tasks=get_tasks(
         tasks=[
@@ -54,7 +54,7 @@ RTEB_MAIN = Benchmark(
 
 RTEB_ENGLISH = Benchmark(
     name="RTEB(eng, beta)",
-    display_name="English ",
+    display_name="English Retrieval",
     icon="https://github.com/lipis/flag-icons/raw/refs/heads/main/flags/4x3/us.svg",
     tasks=get_tasks(
         tasks=[
@@ -89,7 +89,7 @@ RTEB_ENGLISH = Benchmark(
 
 RTEB_FRENCH = Benchmark(
     name="RTEB(fr, beta)",
-    display_name="French ",
+    display_name="French Retrieval",
     icon="https://github.com/lipis/flag-icons/raw/260c91531be024944c6514130c5defb2ebb02b7d/flags/4x3/fr.svg",
     tasks=get_tasks(
         tasks=[
@@ -107,7 +107,7 @@ RTEB_FRENCH = Benchmark(
 
 RTEB_GERMAN = Benchmark(
     name="RTEB(deu, beta)",
-    display_name="German ",
+    display_name="German Retrieval",
     icon="https://github.com/lipis/flag-icons/raw/260c91531be024944c6514130c5defb2ebb02b7d/flags/4x3/de.svg",
     tasks=get_tasks(
         tasks=[
@@ -141,7 +141,7 @@ RTEB_JAPANESE = Benchmark(
 
 RTEB_FINANCE = Benchmark(
     name="RTEB(fin, beta)",
-    display_name="Finance ",
+    display_name="Finance Retrieval",
     icon="https://github.com/DennisSuitters/LibreICONS/raw/2d2172d15e3c6ca03c018629d60050e4b99e5c55/svg-color/libre-gui-price-tag.svg",
     tasks=get_tasks(
         tasks=[
@@ -162,7 +162,7 @@ RTEB_FINANCE = Benchmark(
 
 RTEB_LEGAL = Benchmark(
     name="RTEB(Law, beta)",
-    display_name="Legal ",
+    display_name="Legal  Retrieval",
     icon="https://github.com/DennisSuitters/LibreICONS/raw/2d2172d15e3c6ca03c018629d60050e4b99e5c55/svg-color/libre-map-library.svg",
     tasks=get_tasks(
         tasks=[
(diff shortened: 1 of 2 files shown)


mteb/leaderboard/benchmark_selector.py
```diff
@@ -50,7 +50,7 @@ class MenuEntry:
                 ),
             ),
             MenuEntry(
-                "Domain-Specific",
+                "Domain -Specific",
```
Samoed: This looks very hacky, maybe we should change it somehow
```suggestion
                "Domain-Specific ",
```

KennethEnevoldsen:
@isaac-chung can I get a second opinion on this?

Samoed:
I think tab retrieval might be a bit confusing. People might think that we just have all tasks with additional filter on retrieval (and language and domain etc). Maybe we should change name somehow

isaac-chung:
Right now it seems like we're putting RTEB as a spotlight. Do we want to do that? If we have a retrieval dedicated tab, would it makes sense to group similar benchmarks, like BEIR?

Samoed:
> If we have a retrieval dedicated tab, would it makes sense to group similar benchmarks, like BEIR?

They're in Retrieval/`Miscellaneous`

isaac-chung:
Nice but, why misc?

mteb/leaderboard/benchmark_selector.py
```diff
@@ -88,17 +88,9 @@ class MenuEntry:
                 "Miscellaneous",  # All of these are retrieval benchmarks
                 mteb.get_benchmarks(
                     [
-                        "BEIR",
                         "BEIR-NL",
```
Samoed: Should be `"BEIR-NL"` moved too?

fzliu:
> Nice but, why misc?

Simply because they were in Misc originally. If there is community feedback on a better structure for the sidebar, I think it definitely makes sense to make changes (but we need to get the feedback first).

KennethEnevoldsen:
Hmm, my two cents (and I do disagree with those) are that retrieval and general-purpose embeddings are the two main use case areas. So, making the UI reflect that. 

Of course, the Retrieval Section is not only for RTEB tasks - I have adapted that by moving around tasks

## Example
Screenshot 2025-09-30 at 10 22 00

Samoed:
Should we move image tasks to separate view then too?

KennethEnevoldsen:
Yeah with the IBM pr I realised that we should restructure a bit more:

Screenshot 2025-09-30 at 10 43 18

mteb/benchmarks/benchmarks/rteb_benchmarks.py
```diff
@@ -10,9 +10,9 @@
   year = {2024},
 }"""
 
-RTEB_MAIN = Benchmark(
+RTEB_MAIN = RtebBenchmark(
     name="RTEB(beta)",
-    display_name="RTEB Retrieval Embedding Benchmark",
+    display_name="Multilingual Retrieval",
```
isaac-chung: If the blogpost is advertising "RTEB", I'd not be able to automatically connect that RTEB and "Multilingual Retrieval" are the same thing.

KennethEnevoldsen: So revert to "RTEB Retrieval Embedding Benchmark"?

KennethEnevoldsen: or just "RTEB retrieval"?

isaac-chung: If it's already under "retrieval", what do you think of just "RTEB"?

KennethEnevoldsen: Will go for that

isaac-chung: We could do RTEB(multilingual) and RTEB(English) ?

KennethEnevoldsen: Hmm, right, just for all other benchmarks we haven't used the shorthand (we have used a general purpose name which does not require people to know the content)

but RTEB English is probably reasonable so will change it throughout

KennethEnevoldsen:
Fixed @isaac-chung's comment on names, but it seems like there are some issues in the aggregation after I merged in #3226 (@q275343119)

Screenshot 2025-09-30 at 20 39 58

Will take a closer look

KennethEnevoldsen:
It seems like it is only using some of the tasks, and this is what causes the issue, but I am not sure where the other tasks get filtered out.

mteb/load_results/benchmark_results.py
```diff
@@ -107,26 +107,29 @@ def filter_tasks(
...
-            if (task_types is not None) and (task_result.task_type not in task_types):
+            if (task_types is not None) and (task_meta.type not in task_types):
                 continue
             if modalities is not None:
-                task_modalities = getattr(task_result, "modalities", [])
-                if not any(modality in task_modalities for modality in modalities):
+                if not any(modality in task_meta.modalities for modality in modalities):
                     continue
+            if (is_public is not None) and (task_meta.is_public is is_public):
+                continue
```
fzowl, line 132: @KennethEnevoldsen @fzliu @Samoed I think this is wrong, it should be:
```
            if (is_public is not None) and (task_meta.is_public is not is_public):
                continue

```

Screenshot 2025-10-01 at 0 26 28

Samoed: I think yes, you're right

q275343119: Yes, @fzowl you are right, I made a mistake here, I will fix this bug as soon as possible

q275343119: > Yes, @fzowl you are right, I made a mistake here, I will fix this bug as soon as possible

Hi @Samoed and @KennethEnevoldsen ,I fix it in new pr:https://github.com/embeddings-benchmark/mteb/pull/3230

KennethEnevoldsen:
Thanks for the fix @q275343119 - did you guys run the leaderboard?

I at least still only have a few of the datasets:

Screenshot 2025-10-01 at 09 01 20

If I check the cache it shouldn't be an issue
 ```py
import mteb

rteb = mteb.get_benchmark("RTEB(beta)")

benchmark_res = mteb.load_results(
    tasks=rteb.tasks, only_main_score=True, validate_and_filter=True
)

len(benchmark_res.filter_tasks(is_public=False).task_names)  # 16
len(benchmark_res.filter_tasks(is_public=True).task_names)  # 13

len([t for t in rteb.tasks if not t.metadata.is_public])  # 16
len([t for t in rteb.tasks if t.metadata.is_public])  # 13
```

KennethEnevoldsen:
This bug also seems to influence other benchmarks:

Screenshot 2025-10-01 at 09 06 56

Which works just fine in the live leaderboard

KennethEnevoldsen:
@Samoed and @isaac-chung can I ask any of you to reproduce this? I am afraid it might be a local issue

Samoed:
I build on HF space and leaderboard look correct (I see that you build in space too)

KennethEnevoldsen:
Ran it here seems like it is a local bug: https://huggingface.co/spaces/mteb/leaderboard_dev

KennethEnevoldsen:
This was also confirmed by @q275343119. So I think this should be ready to merge!

PR #3226, opened 2025-09-30
feat:Update RTEB summary columns

q275343119:
1.  Enhanced BenchmarkResult.filter_tasks method to support filtering by is_public status.

2. Updated RTEB summary: added mean(public) & mean(private) columns, removed mean(task) & mean(taskType).

Preview: https://huggingface.co/spaces/SmileXing/leaderboard

### DIFF
--- a/mteb/benchmarks/benchmark.py
+++ b/mteb/benchmarks/benchmark.py
@@ -10,6 +10,7 @@ from pydantic import AnyUrl, BeforeValidator, TypeAdapter
 from mteb.benchmarks._create_table import (
     _create_per_task_table_from_benchmark_results,
     _create_summary_table_from_benchmark_results,
+    _create_summary_table_mean_public_private,
 )
 from mteb.load_results.load_results import load_results
 
@@ -89,3 +90,11 @@ class Benchmark:
     ) -> pd.DataFrame:
         """Create per-task table. Called by the leaderboard app."""
         return _create_per_task_table_from_benchmark_results(benchmark_results)
+
+
+class RtebBenchmark(Benchmark):
+    def _create_summary_table(
+        self, benchmark_results: BenchmarkResults
+    ) -> pd.DataFrame:
+        """Create summary table. Called by the leaderboard app."""
+        return _create_summary_table_mean_public_private(benchmark_results)
--- a/mteb/load_results/benchmark_results.py
+++ b/mteb/load_results/benchmark_results.py
@@ -107,6 +107,7 @@ class ModelResult(BaseModel):
         domains: list[TASK_DOMAIN] | None = None,
         task_types: list[TASK_TYPE] | None = None,
         modalities: list[MODALITIES] | None = None,
+        privacy: Literal["public", "private"] | None = None,
     ) -> ModelResult:
         # TODO: v2 see filter_tasks in BenchmarkResults - but can be moved to a private function or removed
         new_task_results = []
@@ -127,6 +128,10 @@ class ModelResult(BaseModel):
                 task_modalities = getattr(task_result, "modalities", [])
                 if not any(modality in task_modalities for modality in modalities):
                     continue
+            if (privacy is not None) and (
+                task_result.is_public != (privacy == "public")
+            ):
+                continue
             new_task_results.append(task_result)
         return type(self).model_construct(
             model_name=self.model_name,
@@ -395,6 +400,7 @@ class BenchmarkResults(BaseModel):
         domains: list[TASK_DOMAIN] | None = None,
         task_types: list[TASK_TYPE] | None = None,  # type: ignore
         modalities: list[MODALITIES] | None = None,
+        privacy: Literal["public", "private"] | None = None,
     ) -> BenchmarkResults:
         # TODO: Same as filter_models
         model_results = [
@@ -404,6 +410,7 @@ class BenchmarkResults(BaseModel):
                 domains=domains,
                 task_types=task_types,
                 modalities=modalities,
+                privacy=privacy,
             )
             for res in self.model_results
         ]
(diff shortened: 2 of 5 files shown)


q275343119:
Hi @KennethEnevoldsen ,could you please take a look when you have time ?

mteb/load_results/task_results.py
```diff
@@ -167,6 +167,8 @@ def from_task_results(
                     "languages": eval_langs,
                 }
                 flat_scores[split].append(_scores)
+        # AttributeError: 'Namespace' object has no attribute 'is_public'
```
Samoed: When do you have this exception?

mteb/load_results/task_results.py
```diff
@@ -175,6 +177,7 @@ def from_task_results(
             scores=flat_scores,
             evaluation_time=evaluation_time,
             kg_co2_emissions=kg_co2_emissions,
+            is_pubic=is_public,
```
Samoed: Can you take it later from `task.metadata`?

mteb/load_results/task_results.py
SmileLikeYe: 327264985889e02dcb9373f908d7a7f9
Hi, here is my GitHub alt account。
When I  want to use task.metadata.is_public , it will raises error like the screenshot. So I have to use the get attar

mteb/benchmarks/benchmark.py
```diff
@@ -89,3 +90,11 @@ def _create_per_task_table(
     ) -> pd.DataFrame:
         """Create per-task table. Called by the leaderboard app."""
         return _create_per_task_table_from_benchmark_results(benchmark_results)
+
+
+class RtebBenchmark(Benchmark):
```
KennethEnevoldsen, line 95: Unsure if we want to do this or 

```
RTEB_MAIN._create_summary_table = _create_summary_table_mean_public_private
```
@Samoed do you have a preference? I think I am leaning toward using a class

mteb/load_results/task_results.py
SmileLikeYe: @Samoed As I menthoned above, When I want to use task.metadata.is_public , it will raises error like the screenshot. 
327264985889e02dcb9373f908d7a7f9

mteb/load_results/task_results.py
```diff
@@ -228,6 +231,10 @@ def domains(self) -> list[str]:
     def task_type(self) -> str:
         return self.task.metadata.type
 
+    @property
+    def task_is_public(self) -> bool:
+        return getattr(self.task.metadata, "is_public", True)
```
KennethEnevoldsen, line 233:
```suggestion
        return self.task.metadata.is_public
```

mteb/load_results/benchmark_results.py
```diff
@@ -395,6 +400,7 @@ def filter_tasks(
         domains: list[TASK_DOMAIN] | None = None,
         task_types: list[TASK_TYPE] | None = None,  # type: ignore
         modalities: list[MODALITIES] | None = None,
+        privacy: Literal["public", "private"] | None = None,
```
KennethEnevoldsen, line 403: would convert to `is_public: bool`

KennethEnevoldsen:
Can we also add the "is_public" column to the task_information table (otherwise the mean column are hard to interpret)

mteb/benchmarks/benchmark.py
Samoed: I agree, class will be better

KennethEnevoldsen:
We also have to figure out a way to handle the plots - both currently does not work:

Screenshot 2025-09-30 at 14 27 49

KennethEnevoldsen:
A simple fix to replace Retrieval with Mean(Task). I would also add a mean(No Code), Mean(eng), Mean(deu), etc., including one for code.

mteb/load_results/task_results.py
Samoed: I think this is because old tasks have namespaces instead of taskmetadata. Should we change that @KennethEnevoldsen? https://github.com/embeddings-benchmark/mteb/blob/fb5902ad6bfdb9b9cac9c21458f88a6b31e71e17/mteb/load_results/task_results.py#L27-L84

Samoed: But overall, I don't think you should return `is_public` in results. I think it would be better to return full metadata

KennethEnevoldsen:
Also, currently, mean(public) and mean(private) are quite hard to compare; I'm not sure if there is a good solution to fix it.

SmileLikeYe:
> Also, currently, mean(public) and mean(private) are quite hard to compare; I'm not sure if there is a good solution to fix it.

Do you have any suggestion?

KennethEnevoldsen:
> Do you have any suggestion?

We could do a normalised diff - i.e. 
```
normalised_diff = z_score(mean_public - mean_private)
```
But I am unsure how it would be displayed meaningfully 

Normalised Diff, seems unclear on what direction is better if not stated

q275343119:
@KennethEnevoldsen I've update the PR accrding to your comments. here are the results screenshot:
90ea398109422df66ee8e1620b2bb2ed
image
image

Can you review it?

mteb/load_results/benchmark_results.py
q275343119: But if we treat it as a filter field, using an enum seems more appropriate.
It could allow selecting private / public / all etc ？

KennethEnevoldsen:
Will merge this into #3222 as most things here are fixed

Samoed:
@KennethEnevoldsen Nither of my comments weren't resolved. Should we fix them in another PR or what do you think about them? I don't think we need `is_public` in tasks results

q275343119:
> @KennethEnevoldsen Nither of my comments weren't resolved. Should we fix them in another PR or what do you think about them? I don't think we need `is_public` in tasks results

@Samoed You can check the last commit, I have dropped the is_public from the tasks resutls.

Samoed:
Oh, great!

PR #3228, opened 2025-09-30
fix: Further specified macro-language code for Norwegian

KennethEnevoldsen:
"nor" is a macro-language code that covers bokmål and nynorsk (both norwegian), but this means that these datasets will be missed if using "nob" or "nno". Specifying it like this should allow this.

### DIFF
--- a/mteb/models/cohere_models.py
+++ b/mteb/models/cohere_models.py
@@ -82,6 +82,8 @@ supported_languages = [
     "nep-Deva",
     "nld-Latn",
     "nor-Latn",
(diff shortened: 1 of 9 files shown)


Samoed:
I also found `not-Latn⁣` in:
- XM3600T2IRetrieval
- WebFAQBitextMiningQuestions
- WikipediaRerankingMultilingual
- MKQARetrieval
- WikipediaRetrievalMultilingual

KennethEnevoldsen:
fixed

PR #3231, opened 2025-10-01
Feat: Move zero-shot percentage calculation to the end of summary

q275343119:
Move zero-shot percentage calculation to the end of summary table creation which only apply to RTEB table.

You can see the changes in: https://huggingface.co/spaces/SmileXing/leaderboard.

### DIFF
--- a/mteb/benchmarks/_create_table.py
+++ b/mteb/benchmarks/_create_table.py
@@ -372,11 +372,6 @@ def _create_summary_table_mean_public_private(
     # Move borda rank to front
     joint_table.insert(0, "Rank (Borda)", joint_table.pop("borda_rank"))
 
-    # Add zero-shot percentage at the end
-    tasks = get_tasks(tasks=list(data["task_name"].unique()))
-    joint_table["Zero-shot"] = model_metas.map(lambda m: m.zero_shot_percentage(tasks))
-    joint_table["Zero-shot"] = joint_table["Zero-shot"].fillna(-1)
-
     return joint_table
 
 
--- a/mteb/benchmarks/_create_table.py
+++ b/mteb/benchmarks/_create_table.py
@@ -344,13 +344,6 @@ def _create_summary_table_mean_public_private(
         ),
     )
 
-    # Add zero-shot percentage
-    tasks = get_tasks(tasks=list(data["task_name"].unique()))
-    joint_table.insert(
-        1, "Zero-shot", model_metas.map(lambda m: m.zero_shot_percentage(tasks))
-    )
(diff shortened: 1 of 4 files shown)


Samoed:
For context 
image

Samoed:
By the way should we change `RTEB(beta)` before release?

q275343119:
> By the way should we change `RTEB(beta)` before release?

I've changed "RTEB(beta)" to "RTEB"

Samoed:
I don't think that this should be just `RTEB`. I've talked to Kenneth, and he want to leave it as beta for now

KennethEnevoldsen:
Yep, should still be beta - I would just remove the zero-shot column, but maybe a good idea to replace with some sort of diff from private/public?

q275343119:
Hi @KennethEnevoldsen and @Samoed ,I have changed RTEB to RTEB(beta), and removed the zero-shot column, and when there is no zero-shot in the `summary_table`, the "Advanced Model Filters" also no longer has the zero-shot option

You can see the changes in: https://huggingface.co/spaces/SmileXing/leaderboard.

Samoed:
Can you look why 3.9 tests are failing?

q275343119:
> Can you look why 3.9 tests are failing?

It seems there is an issue here:
```
/opt/hostedtoolcache/Python/3.9.23/x64/lib/python3.9/site-packages/transformers/trainer.py:5593: in Trainer
    def _get_num_items_in_batch(self, batch_samples: list, device: torch.device) -> int | None:
E   TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

q275343119:
> Can you look why 3.9 tests are failing?

The project requires transformers version>=4.52.0, the latest version is 4.57.0 (updated on Oct 4, 2025), the default uses the latest version, and there should be no issues after downgrading, I'll give it a try

q275343119:
> Can you look why 3.9 tests are failing?

@Samoed passed

q275343119:
@Samoed Datasets available on HuggingFace are failing:

```
We had to rate limit your IP . To continue using our service, create a HF account or login to your existing account, and make sure you pass a HF_TOKEN if you're using the API.
```
Is the HF_TOKEN used too frequently and has been restricted?

Samoed:
I think commits are added incorrectly, a bit

Samoed:
> Is the HF_TOKEN used too frequently and has been restricted?

Maybe. On main branch in ci dataset check never passed

q275343119:
> I think commits are added incorrectly, a bit

The reason may be that I used rebase. I tried to fix it manually. If it doesn't work, I will submit a new PR.

q275343119:
Hi @Samoed ,it looks better

PR #3255, opened 2025-10-05
Aggregate by subset for HUMEv1

isaac-chung:
Goal of this PR is to align the LB aggregation with the paper so that the Humans rank 4th. 

Notably:
1. created a dedicated HUMEBenchmark object, similar to how RTEB's method.
2. `[HUMEMultilingualSentimentClassification](https://huggingface.co/datasets/mteb/HUMEMultilingualSentimentClassification)` had used `nor` instead of `nob`. Changing the actual dataset and results should be done separately.
3. "Human Benchmark" added under "General Purpose" on the left side bar.

Screenshot 2025-10-05 at 16 10 32

### DIFF
--- a/mteb/benchmarks/_create_table.py
+++ b/mteb/benchmarks/_create_table.py
@@ -270,7 +270,6 @@ def _create_summary_table_mean_public_private(
     Returns:
         DataFrame with model summaries, ready for styling in the leaderboard
     """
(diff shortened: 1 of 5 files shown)


Samoed:
Updated leaderboard from this branch https://huggingface.co/spaces/mteb/leaderboard_2_demo

Samoed (approved):
Great!

Issue #3085, opened 2025-08-27
[v2] Add support for `save_predictions` to the tasks


Samoed:
- [x] STS https://github.com/embeddings-benchmark/mteb/pull/3262
- [x] Classification https://github.com/embeddings-benchmark/mteb/pull/3262
- [x] MultilabelClassification https://github.com/embeddings-benchmark/mteb/pull/3262
- [x] ZeroShotClassification https://github.com/embeddings-benchmark/mteb/pull/3292
- [x] BitextMining https://github.com/embeddings-benchmark/mteb/pull/3292
- [x] PairClassifiation https://github.com/embeddings-benchmark/mteb/pull/3298
- [x] Summarization https://github.com/embeddings-benchmark/mteb/pull/3322
- [x] Clustering https://github.com/embeddings-benchmark/mteb/pull/3322
- [x] ClusteringFaset https://github.com/embeddings-benchmark/mteb/pull/3322
- [x] Retrieval
- [x] AbsTaskImageTextPairClassification https://github.com/embeddings-benchmark/mteb/pull/3322
- [x] Regression https://github.com/embeddings-benchmark/mteb/pull/3322
- [x] add test https://github.com/embeddings-benchmark/mteb/pull/3262#discussion_r2405991639

Samoed:
I've noticed that `Retrieval` task returns predictions for the inputs, while other evaluators returns metrics. Maybe we should standardize this a bit, but I'm not sure if evaluator should return only predictions

KennethEnevoldsen:
We should def. save predictions - I am not sure if we need the standardisation for the evaluator though (not all tasks use them) - However it might be easier that way

Samoed:
I mean, most of the tasks compute metrics in `evaluators`. I think we can make all evaluators to output predictions instead of scores and compute scores in tasks

KennethEnevoldsen:
perfectly fine with me

PR #3262, opened 2025-10-06
Add save predictions to `STS`, `Classification`, `MultilabelClassification`

Samoed:
Add save predictions to `STS`, `Classification`, `MultilabelClassification`. Ref https://github.com/embeddings-benchmark/mteb/issues/3085. I moved calculation of scores from evaluator to tasks and added typed dict to metrics

### DIFF
--- a/mteb/_evaluators/any_sts_evaluator.py
+++ b/mteb/_evaluators/any_sts_evaluator.py
@@ -4,21 +4,37 @@ import logging
 from typing import Any
 
 from datasets import Dataset
-from scipy.stats import pearsonr, spearmanr
 from sklearn.metrics.pairwise import (
     paired_cosine_distances,
     paired_euclidean_distances,
     paired_manhattan_distances,
 )
+from typing_extensions import TypedDict
 
 from mteb.abstasks.task_metadata import TaskMetadata
 from mteb.create_dataloaders import create_dataloader
 from mteb.models import Encoder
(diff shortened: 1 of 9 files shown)


mteb/_evaluators/any_sts_evaluator.py
```diff
@@ -4,29 +4,44 @@
 from typing import Any
 
 from datasets import Dataset
-from scipy.stats import pearsonr, spearmanr
 from sklearn.metrics.pairwise import (
     paired_cosine_distances,
     paired_euclidean_distances,
     paired_manhattan_distances,
 )
+from typing_extensions import TypedDict
```
KennethEnevoldsen:
```suggestion
from typing import TypedDict
```

mteb/_evaluators/any_sts_evaluator.py
KennethEnevoldsen, line 1: Can you add a test for STS, multilabel CLF and CLF? Just to see that the results is as expected.

Samoed: I run implementation in main and from this branch and results are the same. What test do you want to add? That scores are matching?

KennethEnevoldsen: Mainly that the saved output is as expected - however it is minor feel free to merge

Samoed: I will add this test after I finish all predictions

PR #3266, opened 2025-10-06
fix: add prompt for MIRACLRetrievalHardNegatives

Drozhzhinastya:
added prompt query for MIRACLRetrievalHardNegatives run task to mteb/mteb/tasks/Retrieval/multilingual/MIRACLRetrieval.py

### DIFF
--- a/mteb/tasks/Retrieval/multilingual/MIRACLRetrieval.py
+++ b/mteb/tasks/Retrieval/multilingual/MIRACLRetrieval.py
@@ -300,9 +300,77 @@ def _load_miracl_data_hard_negatives(
 
 
 class MIRACLRetrievalHardNegatives(MultilingualTask, AbsTaskRetrieval):
+    # in current version prompt for instruction models different from original task
+    superseded_by = "MIRACLRetrievalHardNegatives.v2"
+
     metadata = TaskMetadata(
         name="MIRACLRetrievalHardNegatives",
         description="MIRACL (Multilingual Information Retrieval Across a Continuum of Languages) is a multilingual retrieval dataset that focuses on search across 18 different languages. The hard negative version has been created by pooling the 250 top documents per query from BM25, e5-multilingual-large and e5-mistral-instruct.",
+        reference="http://miracl.ai",
+        dataset={
+            "path": "mteb/miracl-hard-negatives",
+            "revision": "95c8db7d4a6e9c1d8a60601afd63d553ae20a2eb",
+            "trust_remote_code": True,
+        },
+        type="Retrieval",
+        category="s2p",
+        modalities=["text"],
+        eval_splits=[_EVAL_SPLIT],
+        eval_langs=_LANGUAGES,
+        main_score="ndcg_at_10",
+        date=("2022-06-01", "2023-01-30"),
+        domains=["Encyclopaedic", "Written"],
+        task_subtypes=[],
+        license="cc-by-sa-4.0",
+        annotations_creators="expert-annotated",
+        dialect=[],
+        sample_creation="created",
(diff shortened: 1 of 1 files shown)


Samoed:
@KennethEnevoldsen It seems, that we didn't have prompt for MiraclHardNegatives, but it exists in `MiraclRetrieval`. Maybe we need to reevaluate some `RTEB` models

Drozhzhinastya:
I tested the models from RuMTEB leaderboard on MIRACLRetrievalHardNegatives task, and it turned out that the instruction models show lower performance compared to the full version of MIRACLRetrieval.
[df_compare_miracles.csv](https://github.com/user-attachments/files/22726405/df_compare_miracles.csv)

KennethEnevoldsen:
Hmm, ahh, this is a problem. It is very hard to change a task prompt once it is used in a benchmark. It would invalidate previous results and might be hard to track down...

Not sure what the best solution is here

Samoed:
Maybe we can add `v2` version of miraclhardnegatives and update `RTEB` tasks (or just add v2 version of it)?

KennethEnevoldsen:
But it is used in more than just RTEB - Making a v2 sounds reasonable

Samoed:
Addede `v2` task

mteb/tasks/Retrieval/multilingual/MIRACLRetrieval.py
```diff
@@ -337,6 +340,71 @@ class MIRACLRetrievalHardNegatives(MultilingualTask, AbsTaskRetrieval):
...
+class MIRACLRetrievalHardNegativesV2(MultilingualTask, AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="MIRACLRetrievalHardNegatives.v2",
+        description=(
+            "MIRACL (Multilingual Information Retrieval Across a Continuum of Languages) is a multilingual retrieval "
+            "dataset that focuses on search across 18 different languages. The hard negative version has been "
+            "created by pooling the 250 top documents per query from BM25, e5-multilingual-large and e5-mistral-instruct."
+            "V2 version differs only by prompt."
+        ),
+        reference="http://miracl.ai",
```
KennethEnevoldsen: should we define this in a dictionary outside and unpack? (to avoid changes in one that doesn't happen in the other)

mteb/tasks/Retrieval/multilingual/MIRACLRetrieval.py
```diff
@@ -337,6 +340,71 @@ class MIRACLRetrievalHardNegatives(MultilingualTask, AbsTaskRetrieval):
...
+
+
+class MIRACLRetrievalHardNegativesV2(MultilingualTask, AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="MIRACLRetrievalHardNegatives.v2",
+        description=(
+            "MIRACL (Multilingual Information Retrieval Across a Continuum of Languages) is a multilingual retrieval "
+            "dataset that focuses on search across 18 different languages. The hard negative version has been "
+            "created by pooling the 250 top documents per query from BM25, e5-multilingual-large and e5-mistral-instruct."
+            "V2 version differs only by prompt."
```
KennethEnevoldsen:
```suggestion
            "V2 uses a more appropriate prompt rather than the default prompt for retrieval."
```

KennethEnevoldsen (approved):
A few minor things

mteb/tasks/Retrieval/multilingual/MIRACLRetrieval.py
Samoed: Moved

PR #3267, opened 2025-10-06
[v2] fix ci

Samoed:
Change `[docs]` to `--group docs`. Fixed name of `BRIGHT(long)`


Samoed:
It seems that leaderboard building, but test still fail

KennethEnevoldsen:
Leaderboard seems to fail (I can't reproduce it locally though...)

PR #3276, opened 2025-10-07
[v2] update gh readme to point to the new documentation

KennethEnevoldsen:
Also made a few minor fixes to the datasets

### DIFF
--- a/citation.cff
+++ b/citation.cff
@@ -1,0 +1,205 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite both of the below papers. In addition please cite any paper associated with the specific datasets and benchmarks you use."
+title: "MTEB: Your toolkit for evaluating embeddings"
+type: software
+url: "https://arxiv.org/abs/2502.13595"
+repository-code: "https://github.com/embeddings-benchmark/mteb"
+license: "Apache-2.0"
+keywords:
+  - text embeddings
+  - multilingual
+  - benchmark
+preferred-citation:
+  type: article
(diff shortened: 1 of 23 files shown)


citation.cff
Samoed, line 1: Great! I wanted to add this too, but I thought we want to add multiple papers, but citation.cff support only 1

mteb/tasks/Clustering/nob/snl_clustering.py
```diff
@@ -27,11 +27,11 @@ class SNLClustering(AbsTaskAnyClustering):
     metadata = TaskMetadata(
         name="SNLClustering",
         dataset={
-            "path": "navjordj/SNL_summarization",
-            "revision": "3d3d27aa7af8941408cefc3991ada5d12a4273d1",
+            "path": "adrlau/navjordj-SNL_summarization_copy",  # TODO: replace with mteb/SNLRetrieval after #2820 is resolved.
```
Samoed, line 30: Why revision was changed?

KennethEnevoldsen: It is the same as on main (it is not a direct copy of the git so revs are different)

Samoed: This task was reuploaded https://huggingface.co/datasets/mteb/SNLHierarchicalClusteringP2P

KennethEnevoldsen: Hmm yea so it just the old one that doesn't work

PR #3278, opened 2025-10-07
Update PyLate model revision to fix inference without FA

NohTow:
Hello,

As detailed [here](https://github.com/embeddings-benchmark/mteb/pull/3183#issuecomment-3376052542), there was an issue with ColBERT models using Flash Attention during training but not inference. This has been fixed in a new version of PyLate that has been released (to fix CPU inference and GPU without FA), but it also requires to use the new config from the model, which was not done because the revision of the model was set to an older one.

This fixes the issue by updating the revision to get the correct config


Samoed:
I run from this branch and got `0.75799` on SciFact too. So this was FA problem. I will try to look later why we have lower performance

Issue #2790, opened 2025-06-08
[v2] rework printing when running tasks


KennethEnevoldsen:
Currently, we have a lot of printing when running tasks. I would like to rework printing such that it is both pleasant and considerate when running many tasks, but also such that the `mteb` interfaces nicely with the tqdm from `SentenceTransformers`.

PR #3286, opened 2025-10-07
Better prints classification [draft]

KennethEnevoldsen:
First shot at improving prints. I am starting with

- fixed warning for np.mean
- redid experiments to use tqdm
  - I am unsure if people would like to be able to turn if off and if so how?
- I added two logs*.txt files to show how the logging changed (will be removed later)
- Added a logOnce function to reduce unec. repititions
- Added nesting for tqdm (unsure if the complexity here is too much)

run using:
```py
import logging
import os

import mteb

os.environ["TOKENIZERS_PARALLELISM"] = "false"
# happens to to sklearn .fit parallelism (unsure if we should handle this)

model_name = "sentence-transformers/static-similarity-mrl-multilingual-v1"
model = mteb.get_model(model_name)
results = mteb.evaluate(model, tasks=tasks, cache=None, overwrite_strategy="always")
```

I purposefully didn't use rich as I want it work cleanly without it.


logs-new.txt
```diff
@@ -0,0 +1,45 @@
...
+INFO:mteb.models.sentence_transformer_wrapper:No model prompts found for task=Banking77Classification prompt_type=None
+Running Experiment (10/10) - Evaluating Classifier...: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [00:27<00:00,  2.72s/it]
+  └─ Batches: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 20/20 [00:00<00:00, 345.41it/s]
+INFO:mteb.evaluate:✓ Finished evaluation for Banking77Classification
```
KennethEnevoldsen: Example of new logs

Samoed: That looks great!

KennethEnevoldsen:
@Samoed is the progress bar something that should be possible to disable? (we could read encode_kwargs["show_progress_bar"])

If not using this approach, it would require some work

Samoed:
Yes, I tried to use it in some model implementations, e.g. in clip https://github.com/embeddings-benchmark/mteb/blob/b3e013fec8f678c549e73107ee6bde7f0842cb1a/mteb/models/model_implementations/clip_models.py#L66 to follow sentence transformers approach

KennethEnevoldsen:
Added show progress bar now (wasn't that bad actually)

mteb/_evaluators/classification_evaluator.py
```diff
@@ -104,24 +109,37 @@ def __call__(  # type: ignore[override]
...
-        if test_cache is None:
-            test_cache = model.encode(
-                dataloader_test,
+        pbar_desc = ""
+        if pbar is not None:
+            pbar_desc = pbar.desc.removesuffix(": ")
+
+        ntqdm = nested_tqdm(pbar) if pbar is not None else nullcontext()
+
+        with ntqdm:
```
Samoed: Do we need tqdm here?

mteb/abstasks/AbsTask.py
```diff
@@ -125,6 +125,7 @@ def evaluate(
         *,
         encode_kwargs: dict[str, Any],
         prediction_folder: Path | None = None,
+        show_progress_bar: bool = True,
```
Samoed, line 128: Should we add support in cli?

mteb/_evaluators/classification_evaluator.py
KennethEnevoldsen: removed

mteb/abstasks/AbsTask.py
KennethEnevoldsen: Yes - Added it to the todo list here: #2790

PR #3290, opened 2025-10-07
[v2] Use `tqdm.auto` everywhere

Samoed:
Change tqdm for consistency


KennethEnevoldsen:
@Samoed we have to take a look at the leaderboard thing - can you reproduce it locally?

PR #3291, opened 2025-10-07
[v2] Add logging for clustering

KennethEnevoldsen:
- Added logging for AbstaskClusteringFast
- Added logging for AbsTaskAnyClustering
- Minor tweaks to Classification clustering to ensure consistent style

### DIFF
--- a/mteb/_evaluators/clustering_evaluator.py
+++ b/mteb/_evaluators/clustering_evaluator.py
@@ -6,6 +6,7 @@ from typing import Any
 from datasets import Dataset
 from scipy.optimize import linear_sum_assignment
 from sklearn import cluster, metrics
+from tqdm import tqdm
 
 from mteb.abstasks.task_metadata import TaskMetadata
 from mteb.create_dataloaders import create_dataloader
@@ -43,6 +44,7 @@ class ClusteringEvaluator(Evaluator):
         *,
         encode_kwargs: dict[str, Any],
         v_measure_only: bool = False,
+        pbar: tqdm | None = None,
     ) -> dict[str, float]:
         data_loader = create_dataloader(
             self.dataset,
@@ -50,6 +52,13 @@ class ClusteringEvaluator(Evaluator):
             input_column=self.input_column_name,
             batch_size=encode_kwargs["batch_size"],
         )
+
+        pbar_desc = ""
+        if pbar is not None:
+            pbar_desc = pbar.desc.removesuffix(": ")
+
+        if pbar is not None:
+            pbar.set_description(pbar_desc + " - Encoding samples...")
 
         embeddings = model.encode(
             data_loader,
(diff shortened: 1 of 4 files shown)


mteb/abstasks/AbsTaskClusteringFast.py
```diff
@@ -152,6 +154,15 @@ def _evaluate_subset(
             raise Exception(
                 "Both max_document_to_embed and max_fraction_of_documents_to_embed are set. Please only set one."
             )
+    
+        pbar = tqdm(
+            desc="Running Clustering",
+            total=4,
+            disable=not show_progress_bar,
+        )
```
KennethEnevoldsen: Unsure if it is actually needed here - the majority of the task is in the encoding step (could easily be removed), but doing something like this would make all task logging somewhat consistent.

We could also restructure it slightly so that a good chunk of the percentage is in the encoding stage. WDYT?

mteb/_evaluators/clustering_evaluator.py
```diff
@@ -69,10 +80,12 @@ def __call__(
         clustering_model.fit(embeddings)
         cluster_assignment = clustering_model.labels_
 
-        logger.info("Evaluating...")
+        if pbar is not None:
+            pbar.set_description(pbar_desc + " - Evaluating clustering...")
```
Samoed: I think it would be better to keep logging separate, for when users disable the progress bar but not logging. Because of that, I’m not sure we want to add logging messages to the progress bar here, since it would make the code a bit harder to read.

mteb/abstasks/AbsTaskClusteringFast.py
Samoed: I think some users would want to disable progress bars but keep logging. Also, users can customize logging, and for long tasks it can be more helpful than a progress bar, since they can add timestamps or other information with `basicConfig`. I’m also not sure that all terminals would support progress bars and render everything correctly.

Samoed:
I also made pr for updating tqdm to auto everywhere https://github.com/embeddings-benchmark/mteb/pull/3290

mteb/_evaluators/clustering_evaluator.py
KennethEnevoldsen: I can keep them as debug messages

Samoed: Can we leave as info? Datasets and other libraries have a lot of messages in debug and because of them it would be hard to read

mteb/abstasks/AbsTaskClusteringFast.py
KennethEnevoldsen: Hmm but these take quite a short time and I was thinking of setting info as we the default for the CLI (with the progress bars), however, these will break those

KennethEnevoldsen: an alternative might be to log only is progress bar is turned off but that also seems weird

mteb/_evaluators/clustering_evaluator.py
KennethEnevoldsen: (see above)

KennethEnevoldsen:
@Samoed after having worked a bit more on this. I think it might be better to:

1) Remove all progress bars from tasks, only classification and AbsTaskAnyClustering need it. In general, it is only the encode step that takes a while on a task, and that has a progress bar
2) Add tqdm to the evaluate for running a model on multiple tasks - this is where you really need a progress bar

~~WDYT?~~

Changed it to that

mteb/abstasks/AbsTaskClusteringFast.py
Samoed: Let's leave as is then

Samoed:
Yes, I agree

PR #3292, opened 2025-10-07
Add predictions for `ZeroShot` and `BitextMining`

Samoed:
I found out that during reuploads I changed `load_dataset` wrong and these datasets didn't work.

Refactored bitextmining to make it more clear

Added tests for predictions

### DIFF
--- a/mteb/_evaluators/_download.py
+++ b/mteb/_evaluators/_download.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from pathlib import Path
 
 import requests
-import tqdm
+from tqdm.auto import tqdm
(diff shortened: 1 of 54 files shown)


KennethEnevoldsen:
@Samoed generally looks good though there seems to be a failing test

Samoed:
It seems that cache of reranking task is failing, but I didn't change this task. Maybe if we rerun it will be good, but windows ci is too long, but seems from ubuntu work fine then this is fine

KennethEnevoldsen:
Feel free to just merge (we can speed up tests in another PR

PR #3296, opened 2025-10-08
fix: Change language for task SlovakMovieReviewSentimentClassification

andrejridzik:
Fixing the language for SlovakMovieReviewSentimentClassification task

* The original task was incorrectly categorized under language code [svk](http://github.com/embeddings-benchmark/mteb/blob/main/mteb/iso_639_3_to_language.json#L6143), which corresponds to *Slovakian Sign Language*
* This PR just changes the language to use the correct code [slk](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/iso_639_3_to_language.json#L5928)

## Checklist
- [x] I have tested that the dataset runs with the mteb package.


Samoed (approved):

Samoed:
Good catch!

Issue #3294, opened 2025-10-07
add support for python 3.14


Samoed:
https://docs.python.org/3.14/whatsnew/3.14.html

KennethEnevoldsen:
I think we are limited by PyTorch (to 3.13)

Samoed:
Pytrec PR for supporting 3.14 https://github.com/terrierteam/pytrec_eval/pull/12

KennethEnevoldsen:
Any chance we can ping someone from there?

Samoed:
Pinged in PR. By the way, pytorch supports py3.14 already

cmacdonald:
pytrec_eval released with support for 3.14. https://pypi.org/project/pytrec-eval-terrier/0.5.10/#files

Raise a new issue if you find any problems.

Samoed:
I've started action for python3.14 support [#3450](https://github.com/embeddings-benchmark/mteb/pull/3450)

Samoed:
But it seems that pyarrow don't support 3.14 yet https://github.com/apache/arrow/issues/47438

Samoed:
Arrow was released, but datasets have issues https://github.com/huggingface/datasets/issues/7813

Samoed:
Datasets with support 3.14 was released and ci is passing in https://github.com/embeddings-benchmark/mteb/pull/3450

Issue #2538, opened 2025-04-12
Drop Python 3.9 support


isaac-chung:
Python 3.9 will stop receiving security updates in [Oct 2025](https://peps.python.org/pep-0596). Propose to remove 3.9 tests and any references in the repo.

KennethEnevoldsen:
I would probably keep the 3.9 support until Oct. 2025. Though I am looking forward to it.

Is there anything specifically that this bump would fix?

KennethEnevoldsen:
@isaac-chung does this mean we should close this for now?

isaac-chung:
I don't have a preference. If we close it now, let's mark it as won't do

KennethEnevoldsen:
Let us keep it open to track relevant changes.

Just for completion:

- this would fix the issue with Gradio, which requires 3.10

isaac-chung:
Et voila! We're in October.

Samoed:
I think we need to wait until November, until it totally EOL

KennethEnevoldsen:
We can implement the drop in v2?

Samoed:
Yes, we can do this there

PR #3304, opened 2025-10-10
[v2] remove support for 3.9 and add support for 3.13

KennethEnevoldsen:
related to 

#2538 (only 3.13)
and
#3294 

set upper bound for pylate as pylate (due to voyager) requires python 3.12

### DIFF
--- a/mteb/types/_encoder_io.py
+++ b/mteb/types/_encoder_io.py
@@ -11,7 +11,6 @@ from PIL import Image
 from typing_extensions import NotRequired
 
 # --- Output types ---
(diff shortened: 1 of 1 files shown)


mteb/types/_encoder_io.py
```diff
@@ -105,12 +105,12 @@ class MultimodalInput(TextInput, CorpusInput, QueryInput, ImageInput, AudioInput
 
 
 # should be as Union, because `|` won't work for python3.9
```
Samoed: Comments about python3.9 in this file can be removed

PR #3102, opened 2025-08-30
[v2] Merge main 30 08

Samoed:

PR #3306, opened 2025-10-10
[v2]: Starting to clean up tests

KennethEnevoldsen:
Plan:
- Started to remove tests that tests already tested functionality (minimal only when I am 100%)
- if a test uses MTEB I generally refactor it to use evaluate
- generally move stuff around to reflect the structure on the package
- If test file contains semantically different tests split it up into multiple
- for all deprecated functions I moved the tests into `test_deprecated/*`
- speed:
  - removed `intfloat/multilingual-e5-small` from the integration test so reduce the number of model downloads required
  - replaced "IntructIR" with "IFIRNFCorpus" in the `task_grid.py` as it is notably smaller
  - Reduce grid for testing, `get_tasks` from 2592 to 16+18+24+54=112 by splitting it up into multiple instances

I would love to get some feedback on if people agree with these changes, then I will continue

One of the major refactors was `test_benchmark`, which contained both integration tests, tests of MTEB, and tests of `get_benchmark` and the benchmarks themselves (e.g. names must be unique)

TODO:
- [ ] fix remaining issues in getting tests to run
- [ ] ~figure out where the file saves come from~ Will fix it in another PR

### DIFF
--- a/tests/mock_tasks.py
+++ b/tests/mock_tasks.py
@@ -110,7 +110,8 @@ def instruction_retrieval_datasplit() -> RetrievalSplitData:
 
 
 class MockClassificationTask(AbsTaskAnyClassification):
-    classifier = LogisticRegression(n_jobs=1, max_iter=2)  # type: ignore
+    classifier = LogisticRegression(n_jobs=1, max_iter=10)  # type: ignore
+
     expected_stats = {
         "test": {
             "num_samples": 2,
@@ -185,7 +186,7 @@ class MockClassificationTask(AbsTaskAnyClassification):
 
 
 class MockMultilingualClassificationTask(AbsTaskAnyClassification):
-    classifier = LogisticRegression(n_jobs=1, max_iter=2)
+    classifier = LogisticRegression(n_jobs=1, max_iter=10)
 
     expected_stats = {
         "test": {


Samoed:
I've started looking when our time on tests was increased.
1. After updating retrieval format +10 min https://github.com/embeddings-benchmark/mteb/pull/2885 (52 min https://github.com/embeddings-benchmark/mteb/actions/runs/16714678318/usage, previous pr 42 min https://github.com/embeddings-benchmark/mteb/actions/runs/16706198422/usage) Maybe we need to change `TwitterHjerneRetrieval` as test for retrieval dataloader and use `NanoBeir` tasks
2. And after this merge of main another +10 min https://github.com/embeddings-benchmark/mteb/pull/3102 (this 1h tests https://github.com/embeddings-benchmark/mteb/actions/runs/17370564430/usage, previous pr has 50 min tests https://github.com/embeddings-benchmark/mteb/actions/runs/17233888466/usage?pr=3040) maybe evaluators run takes too long


I will try to review tests that we want to remove

KennethEnevoldsen:
@Samoed 

This test seem to be especially problematic:`test_benchmark_integration_with_datasets.py`

task 11 in `test_benchmark_datasets` seems to be `Core17InstructionRetrieval` or `InstructIR`

```
1216.05s call     tests/test_benchmark/test_benchmark_integration_with_datasets.py::test_benchmark_datasets[model0-task11]
47.03s call     tests/test_cli.py::test_create_meta
38.62s call     tests/test_overview.py::test_get_tasks_size_differences
20.43s call     tests/test_benchmark/test_benchmark.py::test_multiple_mteb_tasks[model0-tasks0]
19.23s call     tests/test_cli.py::test_run_task[intfloat/multilingual-e5-small-BornholmBitextMining-fd1525a9fd15316a2d503bf26ab031a61d056e98]

**Comment by Samoed** [comment]
This should be `IntructIR`, it has 9906 queries, maybe this is the reason https://github.com/embeddings-benchmark/mteb/blob/a95c931fc4574635182574a4497700411dd44ce6/mteb/descriptive_stats/Reranking/InstructIR.json#L10, because Core17InstructionRetrieval have similar corpus, but run fast, but only 40 queries https://github.com/embeddings-benchmark/mteb/blob/a95c931fc4574635182574a4497700411dd44ce6/mteb/descriptive_stats/InstructionReranking/Core17InstructionRetrieval.json#L10

**Comment by Samoed** [comment]
We also can split tests run. E.g. we can move out tests related to metadata to separate to increase feedback speed. Now we're having test that format citation and this is a bit strange, because we need to check what's wrong with them.

Maybe it would be better to create github action that test this automatically (because github don't support multiple templates for PR and users don't fill them), but this can be done later and of course in separate issue

**Comment by KennethEnevoldsen** [comment]
> because github don't support multiple templates for PR and users don't fill them

Should we just delete these? (seperate issue)

**Comment by KennethEnevoldsen** [comment]
>This should be IntructIR, it has 9906 queries, maybe this is the reason

replaced it with "IFIRNFCorpus" which is ~3000 samples and ~200 queries

**Comment by Samoed** [comment]
> > because github don't support multiple templates for PR and users don't fill them
> 
> Should we just delete these? (seperate issue)

I was talking about our checklists, not sure if we want to delete them, but we need to automate them

File: tests/test_tasks/test_predictions.py
Comment (KennethEnevoldsen) [inline tests/test_tasks/test_predictions.py:1]: @Samoed, these tasks now fail. I assume it is because the MockNumpyEncoder is random.

should I fix this to make it more consistent?

File: tests/test_integrations/test_prompts.py [hunk 1]
```diff
@@ -0,0 +1,209 @@
...
+    _task_name = (
+        task_name.metadata.name if isinstance(task_name, AbsTask) else task_name
+    )
+
+    class MockEncoderWithInstructions(MockSentenceTransformer):
+        def encode(
+            self, sentences: DataLoader, prompt_name: str | None = None, **kwargs
+        ):
+            assert prompt_name == _task_name
+            return np.zeros((len(sentences.dataset), 10))
```
KennethEnevoldsen: @Samoed, I am unsure why these are MockSentenceTransformer. Shouldn't this work regardless of whether they are wrapped or not?

Similarly, I also don't see the need to wrap them in `MockSentenceTransformerWrapper`. 

Are we testing something specific to the SentenceTransformerWrapper? If so, it is not entirely clear to me

KennethEnevoldsen: Currently it fails since the `mteb_model_meta` is set to `None`. This does not work as either it should be set to a `ModelMeta` or it shouldn't be set (if it can be `None` we need to fix the `evaluate`)

.gitignore
```diff
@@ -157,3 +157,7 @@ powermetrics_log.txt
 
 # vscode
 .vscode/launch.json
+
+# produced during testing
+results/mock__MockSentenceTransformer/1.0.0/MockMultilingualSTSTask.json
+results/mock__MockSentenceTransformer/1.0.0/model_meta.json
```
Samoed: Can we avoid creation?

tests/mock_models.py
```diff
@@ -106,6 +133,7 @@ def encode(
 
 
 class MockCLIPEncoder(AbsMockEncoder):
+    mteb_model_meta = ModelMeta(name="mock/MockCLIPEncoder", **empty_metadata_kwargs)
```
Samoed: `mteb_model_meta` duplicated

tests/test_deprecated/test_MTEB.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@pytest.mark.parametrize("model", [MockNumpyEncoder()])
+def test_run_using_benchmark(model: mteb.Encoder, tmp_path: Path):
+    """Test that a benchmark object can be run using the MTEB class."""
+    bench = mteb.Benchmark(
+        name="test_bench", tasks=mteb.get_tasks(tasks=["STS12", "SummEval"])
+    )
+
+    eval = mteb.MTEB(tasks=[bench])
```
Samoed, line 24: Should we use `mteb.evaluate` here?

tests/test_deprecated/test_MTEB_create_model_meta.py
```diff
@@ -0,0 +1,54 @@
+# TODO: KCE would probably consider removing this
```
Samoed: We can change this to only create metadata object from Mock `SentenceTransformer` and `CrossEncoder`

tests/test_integrations/test_dtype_encoders.py
```diff
@@ -0,0 +1,42 @@
...
+    ],
+)
+def test_benchmark_encoders_on_task(
+    task: str | AbsTask, model: mteb.Encoder, tmp_path: Path
+):
+    if isinstance(task, str):
+        tasks = mteb.get_tasks(tasks=[task])
+    else:
+        tasks = [task]
+
```
Samoed:
```suggestion
```

tests/test_tasks/test_predictions.py
Samoed: I don't see that this test is failing in logs

tests/test_integrations/test_prompts.py
Samoed: This test is testing that we're correctly passing `prompt_name` to the `encode`, but because we're testing on mock tasks I've added `MockSentenceTransformerWrapper`. I think we can mock tasks registry a bit to add mock tasks during testing

tests/test_integrations/test_modality.py
```diff
@@ -0,0 +1,61 @@
...
+        f"'{modality}'" for modality in sorted(task.metadata.modalities)
+    )
+    mock_logger.assert_called_with(
+        f"mock/MockMocoModel only supports ['image'], but the task modalities are [{task_modalities}]."
+    )
+
+
+# TODO: KCE: Unsure if we need this test:
+@pytest.mark.parametrize("task", [MockImageClusteringTask()])
+def test_task_modality_filtering_model_modalities_more_than_task_modalities(task):
```
Samoed, line 51: I think we need this to that we don't filter out tasks. When model supports both image and text, but we're passing either image or text tasks

tests/test_validate_metadata/test_citation_formatting.py
Samoed, line 1: Should we move out tests for task metadata to separate workflow?

tests/test_integrations/test_dtype_encoders.py
```diff
@@ -0,0 +1,42 @@
...
+    mteb.evaluate(model, tasks)
```
Samoed: I think we need to test that output scores are matching with expected score

tests/test_deprecated/test_MTEB.py
isaac-chung: If it's testing a deprecated part as the name suggested, it should remain the same.

.gitignore
KennethEnevoldsen: let me look into it

tests/test_deprecated/test_MTEB_create_model_meta.py
KennethEnevoldsen: So this is only called from MTEB, which is deprecated, which is why I considered removing it

tests/test_integrations/test_prompts.py
KennethEnevoldsen: But isn't there tests below that test for exactly this? Any reason to wrap it?

tests/test_tasks/test_predictions.py
KennethEnevoldsen: Hmm ahh, it fails locally though...

tests/test_validate_metadata/test_citation_formatting.py
KennethEnevoldsen: I think this is fine (no need to add the complexity)

tests/test_integrations/test_dtype_encoders.py
KennethEnevoldsen: Would be happy to do that in another PR

tests/test_integrations/test_prompts.py
Samoed: Yes, I think they're the same

tests/test_validate_metadata/test_citation_formatting.py
Samoed: Yes, but this can be a bit frustrating for users that tests are failing because wrong formated citation

tests/test_tasks/test_predictions.py
Samoed: Maybe something can be different between numpy versions

tests/test_validate_metadata/test_citation_formatting.py
isaac-chung: You mean like how linting is separate in CI?

tests/test_tasks/test_predictions.py
KennethEnevoldsen: I will try to make the encoder a bit more consistent

.gitignore
KennethEnevoldsen: I believe it should be fixed now

tests/test_validate_metadata/test_citation_formatting.py
KennethEnevoldsen: I don't think this is needed. We could, but ideally, we should just make the tests faster to run.

This PR shows that there is plenty of space for that

tests/test_tasks/test_load_data.py
KennethEnevoldsen, line 1: I would love to remove these tests. I am not entirely sure what it tests (that every task has a `load_data`?) as everything that we would want to test is mocked.

tests/test_tasks/test_convert_to_reranking.py
```diff
@@ -66,6 +67,8 @@ def test_mteb_rerank(tmp_path: Path):
 
 
 def test_reranker_same_ndcg1(tmp_path: Path):
```
KennethEnevoldsen, line 69: @Samoed can you help me figure out this test?

Samoed:
We can limit `max_iter` for classification task to speedup tests https://github.com/embeddings-benchmark/mteb/blob/f1c6cf0a81bf24801179ddaf09af9a0bf88472b7/mteb/abstasks/AbsTaskAnyClassification.py#L113-L116

tests/test_tasks/test_load_data.py
Samoed: I agree

tests/test_tasks/test_convert_to_reranking.py
Samoed: This test checks that scores from the encoder can be used for the cross-encoder. But the part that compares it to the same NDCG as the one-stage setup is a bit strange, and I also wanted to remove it (or change it to use mock models). For that, we first need to create a mock cross-encoder. We also need to add a few more tests for cross-encoders.

KennethEnevoldsen:
@isaac-chung and @Samoed one of the slowest test I found was the test for get_task, which had a grid of 2592 and had to loop to a lot tasks (sometimes all, depending on the filter). While not individually very slow these were a major part of the overall runtime.

I split them up into multiple segments and this reduced it down to 16+18+24+54=112 by splitting it up into multiple instances, which is still a quite extensive grid.

tests/test_validate_metadata/test_citation_formatting.py
KennethEnevoldsen: Had a closer look at this, seems like we have a grid that basically tests across all tasks, some aspect of the metadata. This naturally gives us quite a large set of tests. I would (in a separate PR) combine these

isaac-chung:
It would be great to have this merged sooner rather than later (suggest to leave the file saving stuff separately) and fully leverage the sped up tests. Having tests run at 1hr+ each PR is pretty not ideal 🙏

tests/test_tasks/test_predictions.py
Samoed: I've rounded predictions in https://github.com/embeddings-benchmark/mteb/pull/3322

KennethEnevoldsen:
Alright, figured out the mistake here. Took some time as the error only happens when running multiple tests at once.

Since `mteb.evaluate` unloads the dataset after it has been run (added by @Samoed and a very reasonable default for memory) this causes the `task.dataset` to be set to `None`, which leads future test to fail. This only happens in `Mock*Task`s though as they set `self.dataset["train"]` using a function to generate the data. The fix is simple simply recreate `self.dataset` in the `load_data` function.

KennethEnevoldsen:
I think this is the remaining tests that needs to be solved, but it is down to 3 min:

Screenshot 2025-10-12 at 17 56 52

In comparison, we started at >30min locally (all of it is with a hot cache of models and datasets)

I will take a shower and make some dinner, though, so I will leave it here.

It seems to only be a local error, though (I suspect it's something to do with the numpy version, seed, or something like that). Anyway, that shouldn't happen, so I will fix that. However, assuming this runs fine - do feel free to merge (and then I will do it in another PR)

KennethEnevoldsen:
These are the slowest test with a hot cache, without it is almost always the integration tests with datasets:

Screenshot 2025-10-12 at 18 02 18

There are some obvious fixes:

3) Change out Colbert with a smaller model e.g. `answerdotai/answerai-colbert-small-v1`
1, 5) Reduce the number of iterations of the linear classifier used in MockClassificationTask (I actually just added this)   
2, 4) are both CLI, which I imagine takes some time to start up (mostly from loading mteb, I imagine)

KennethEnevoldsen:
I will just enable auto-merge here - feel free to do a post-review. I will do a few follow-up PRs, but having a test run 10x faster should speed up development a bit

tests/test_integrations/test_modality.py
```diff
@@ -0,0 +1,59 @@
...
+from tests.mock_tasks import (
+    MockImageClusteringTask,
+    MockImageTextPairClassificationTask,
+    MockRetrievalTask,
+)
+
+logging.basicConfig(level=logging.INFO)
+
+
+# NOTE: Covers image and image-text tasks. Can be extended to cover new mixed-modality task types.
```
Samoed, line 26: Should we add issues for this?

tests/mock_tasks.py
```diff
@@ -109,6 +110,7 @@ def instruction_retrieval_datasplit() -> RetrievalSplitData:
 
 
 class MockClassificationTask(AbsTaskAnyClassification):
+    classifier = LogisticRegression(n_jobs=1, max_iter=2)  # type: ignore
```
Samoed: Interesting that classificator now output `[[1, 1], [0, 0], [1, 1], [0, 0], [1, 1], [0, 0], [1, 1], [0, 0], [1, 1], [0, 0]]` and because of that tests are failing. I increased max iter a bit and this solved an error

tests/mock_models.py
```diff
@@ -16,6 +16,27 @@
 from mteb.models.sentence_transformer_wrapper import SentenceTransformerEncoderWrapper
 from mteb.types import Array, BatchedInput, PromptType
 
+empty_metadata_kwargs = dict(
+    loader=None,
+    languages=["eng-Latn"],
+    revision="1",
+    release_date=None,
+    modalities=["text"],
```
Samoed, line 24: Most of the time, mock should support both modalities
```suggestion
    modalities=["text", "image"],
```

tests/test_integrations/test_image_model_task.py
```diff
@@ -0,0 +1,17 @@
...
+
+import mteb
+from mteb.abstasks import AbsTask
+from tests.mock_models import MockCLIPEncoder
+from tests.task_grid import MOCK_MIEB_TASK_GRID
+
+logging.basicConfig(level=logging.INFO)
+
+
+@pytest.mark.parametrize("task", MOCK_MIEB_TASK_GRID)
```
Samoed, line 13: Should we combine with mteb tasks test?

tests/test_integrations/test_integration_with_sentencetransformers.py
```diff
@@ -0,0 +1,38 @@
...
+    # this behavior and keeps the prompts as configured by the model, so this
+    # test now sets the prompts to an empty dict explicitly to preserve the legacy
+    # behavior and focus the test on the tasks instead of the prompts.
+    # Using empty dict instead of None to avoid TypeError in SentenceTransformers 5.0.0+
+    model.prompts = {}
+    mteb.evaluate(model, task, cache=None)
+
+
+@pytest.mark.parametrize("task", MOCK_MIEB_TASK_GRID)
+@pytest.mark.parametrize("model", [MockCLIPEncoder()])
```
Samoed, line 34: I think there shouldn't be mock model or mock tasks, because this should be integration test

tests/test_integrations/test_integration_with_datasets.py
```diff
@@ -0,0 +1,22 @@
...
+
+import mteb
+from mteb.abstasks import AbsTask
+from tests.mock_models import MockNumpyEncoder
+from tests.task_grid import TASK_TEST_GRID
+
+logging.basicConfig(level=logging.INFO)
+
+
+@pytest.mark.parametrize("task", TASK_TEST_GRID)
```
Samoed, line 18: We need to add tests for mieb tasks too, but they're huge most of time

isaac-chung:
> I will just enable auto-merge here - feel free to do a post-review. I will do a few follow-up PRs, but having a test run 10x faster should speed up development a bit

Thanks so much!

tests/test_integrations/test_integration_with_datasets.py
KennethEnevoldsen: Do we need to do that if we have the mock tasks?

Samoed:
Great work!

tests/test_integrations/test_integration_with_sentencetransformers.py
KennethEnevoldsen: Oh, this is an accidental duplicate. Removed in #3327

tests/test_integrations/test_image_model_task.py
KennethEnevoldsen: We can't given the model right? But we might be able to move it to a better spot

tests/mock_models.py
KennethEnevoldsen: but the data is text right?

tests/test_integrations/test_integration_with_datasets.py
Samoed: We need to have integration tests with real datasets format to verify that we can download them correctly

tests/mock_models.py
Samoed: Yes, I just think it would be easier if mock models would support all modalities by default, because if they won't support something then tasks should be skipped

PR #3307, opened 2025-10-10
WIP Financial pairs

hladek:
Slovak financial question answering - multiple choice.

All models return zero evaluation. Probably fine-tuning of model  is needed.


hladek:
Sorry, this is a mistake.

Issue #3181, opened 2025-09-15
v2 leaderboard does not reflect v1


KennethEnevoldsen:
### Describe the bug

Seemingly, v1 has more models than v2 - e.g., looking at the default benchmark, we see 299 models for v1, but only 247 for v2.

v1:
<img width="358" height="388" alt="Image" src="https://github.com/user-attachments/assets/c15d5192-8c06-46ff-a5a9-5cd871fcd0fb" />

v2:

<img width="482" height="526" alt="Image" src="https://github.com/user-attachments/assets/2dc23c43-1f88-41be-9aeb-ff3505938a0a" />

### To reproduce

```bash
# on branch: v2.0.0
make run-leaderboard
```
Requires KeyError fix in: #3180 

(Removing the fix with result loading from #3180 does not influence the number of models)

### Are you interested to contribute a fix for this bug?

Yes

KennethEnevoldsen:
@Samoed, any ideas on this?

Samoed:
Hm, I don't have ideas

PR #3308, opened 2025-10-10
[v2] Fix leaderboard

Samoed:
Now leaderboard will use `ResultCache`. I've pushed demo leaderboard here https://huggingface.co/spaces/mteb/leaderboard_2_demo

### DIFF
--- a/mteb/leaderboard/table.py
+++ b/mteb/leaderboard/table.py
@@ -1,5 +1,3 @@
-from typing import TYPE_CHECKING
-
 import gradio as gr
 import matplotlib.pyplot as plt
 import numpy as np
@@ -7,9 +5,8 @@ import pandas as pd
 from matplotlib.colors import LinearSegmentedColormap
 from pandas.api.types import is_numeric_dtype
 
-if TYPE_CHECKING:
-    from mteb.benchmarks.benchmark import Benchmark
-    from mteb.load_results.benchmark_results import BenchmarkResults
+from mteb.benchmarks.benchmark import Benchmark
+from mteb.results.benchmark_results import BenchmarkResults
 
 
 def borda_count(scores: pd.Series) -> pd.Series:


mteb/cache.py
```diff
@@ -361,32 +361,24 @@ def _filter_paths_by_model_and_revision(
         paths: list[Path],
         models: Sequence[str] | Sequence[ModelMeta] | None = None,
     ) -> list[Path]:
-        model_revisions: list[str] | None = None
-        if models is not None:
-            if isinstance(models[0], ModelMeta):
-                models = cast(list[ModelMeta], models)
-                model_names = {m.model_name_as_path() for m in models}
-                model_revisions = [
```
Samoed, line 367: Previously, `model_names` didn’t match with `model_revision` because the set sorted the names, while the revisions remained in their original order

mteb/cache.py
```diff
@@ -361,32 +361,24 @@ def _filter_paths_by_model_and_revision(
...
-                    for m in models
-                ]
+        """Filter a list of paths by model name and optional revision."""
+        if not models:
+            return paths
+
+        if isinstance(models[0], ModelMeta):
+            models = cast(list[ModelMeta], models)
+            name_and_revision = {
+                (m.model_name_as_path(), m.revision or "no_revision_available")
```
Samoed, line 369: Currently we're taking results for models with same revision as metadata, but previous leaderboard was taking results from all revisions
https://github.com/embeddings-benchmark/mteb/blob/67f7ad92d2d8db8e9ee6332e542f27a038ccb232/mteb/leaderboard/app.py#L44

https://github.com/embeddings-benchmark/mteb/blob/67f7ad92d2d8db8e9ee6332e542f27a038ccb232/mteb/leaderboard/app.py#L44

I think this is the reason of issue https://github.com/embeddings-benchmark/mteb/issues/3181

Samoed:
image

mteb/cache.py
KennethEnevoldsen: Ohh.. great fix!

KennethEnevoldsen:
v3.13 is failing. With:

> FAILED tests/test_benchmark/test_benchmark.py::test_benchmark_encoders_on_task[model3-task11] - AttributeError: 'MockSentenceTransformersbf16Encoder' object has no attribute 'prompts'

Probably not an issue introduced here, though

mteb/cache.py
KennethEnevoldsen: Hmm, right, but we do want to use all.

E.g., if mE5 uses rev1, but for some tasks were evaluated using rev2, we would be fine with using those. Rev1 is, however, the revision we prioritise.

mteb/leaderboard/app.py
```diff
@@ -33,16 +34,20 @@
...
-            only_main_score=True, require_model_meta=False, models=ALL_MODELS
-        )._filter_models()
-        all_results.to_disk(results_cache_path)
+        results_cache = ResultCache()
+        results_cache.download_from_remote()
+
+        all_results = results_cache.load_results(
+            models=mteb.get_model_metas(),
+            only_main_score=True,
+            require_model_meta=True,
```
KennethEnevoldsen:
```suggestion
            require_model_meta=False,
```
Some do not have a model meta

mteb/leaderboard/app.py
```diff
@@ -33,16 +34,20 @@
...
     if not results_cache_path.exists():
-        all_results = mteb.load_results(
-            only_main_score=True, require_model_meta=False, models=ALL_MODELS
-        )._filter_models()
-        all_results.to_disk(results_cache_path)
+        results_cache = ResultCache()
+        results_cache.download_from_remote()
+
+        all_results = results_cache.load_results(
+            models=mteb.get_model_metas(),
```
KennethEnevoldsen: I think we need to convert this to just strings to get all the models' revisions (which we can then filter later)

mteb/cache.py
KennethEnevoldsen: Will you make a test for this?

Samoed: Add test in https://github.com/embeddings-benchmark/mteb/pull/3308/commits/7969a15a04b59519ba0e5b359ad320bdf2d9e9ad

mteb/cache.py
Samoed: Updated in https://github.com/embeddings-benchmark/mteb/pull/3308/commits/9f57e1433a71bb5c1e4da2e0c80ef55b3acb7c63

Samoed:
> Probably not an issue introduced here, though

Yes, it didn't pass in https://github.com/embeddings-benchmark/mteb/pull/3304 already. I will try to test this

KennethEnevoldsen:
@Samoed can we do a comparison on the number of models as well? (just to get a status update on https://github.com/embeddings-benchmark/mteb/issues/3181), but regardless I think this is good to merge

Samoed:
We can, but for now we need to fix this issue https://github.com/embeddings-benchmark/mteb/issues/3317

Samoed:
`v2`
image
image

`Main`
image
image

Samoed:
I don't see difference even in MIRACLReranking task

Samoed:
I'll merge and we will test scores later

PR #3311, opened 2025-10-10
remove `from __future__ import annotations`

Samoed:
Since we're dropped `python3.9` we can remove `from future` in most places Close https://github.com/embeddings-benchmark/mteb/issues/3309

### DIFF
--- a/mteb/MTEB.py
+++ b/mteb/MTEB.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 import logging
 import os
@@ -25,6 +27,7 @@ import datasets
 
 import mteb
 from mteb.abstasks import AbsTask
+from mteb.benchmarks import Benchmark
 from mteb.models import (
     CrossEncoderWrapper,
     Encoder,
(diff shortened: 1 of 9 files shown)


mteb/abstasks/AbsTask.py
```diff
@@ -438,7 +437,7 @@ def filter_eval_splits(self, eval_splits: list[str] | None) -> AbsTask:
 
     def filter_modalities(
         self, modalities: list[str] | None, exclusive_modality_filter: bool = False
-    ) -> AbsTask:
+    ) -> Self:
```
Samoed, line 440: In some parts I've added return `Self`, because we can't use class name directly here

mteb/MTEB.py
```diff
@@ -35,13 +36,11 @@
     SentenceTransformerEncoderWrapper,
 )
 from mteb.results import TaskResult
+from mteb.types import ScoresDict
 
 if TYPE_CHECKING:
     from sentence_transformers import CrossEncoder, SentenceTransformer
```
Samoed, line 42: I left `from __future__` in some places where we import some library in `TYPE_CHECKING`

KennethEnevoldsen:
Generally looks good here - tests of course have to work

Samoed:
They're working, but the datasets have just been rate-limited due to my changes. Other tests are expected to fail.

Issue #3314, opened 2025-10-11
Fix python 3.13 tests


Samoed:
v3.13 is failing. With:

> FAILED tests/test_benchmark/test_benchmark.py::test_benchmark_encoders_on_task[model3-task11] - AttributeError: 'MockSentenceTransformersbf16Encoder' object has no attribute 'prompts'

Probably not an issue introduced here, though

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/issues/3308#issuecomment-3393020058_

PR #3315, opened 2025-10-11
[v2] fix 3.13 tests

Samoed:
When I was integrating `encode_query` and `encode_documents` I used nonexistent function, but this is strange, that this appears only in `3.13` 

Close https://github.com/embeddings-benchmark/mteb/issues/3314


KennethEnevoldsen:
@Samoed seems like tests are still failing

PR #3322, opened 2025-10-12
[v2] Finish predictions

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3085

- I made `DeprecatedSummarizationEvaluator` inherit from `SummarizationEvaluator`, since they now share the same `__call__` method but have different `_calculate_metrics`.
- I unified the `Regression` and `Classification` evaluators into a single one called `SklearnModelEvaluator`, which supports all `sklearn-like` models.
- `Classification` and `Regression` tasks now use the same `evaluate_subset` and `_evaluate_subset` methods.
- Added saving of predictions for `Clustering`, `ClusteringFast`, and `ImageTextPairClassification`.

### DIFF
--- a/tests/mock_tasks.py
+++ b/tests/mock_tasks.py
@@ -3952,7 +3952,20 @@ class MockRegressionTask(AbsTaskRegression):
                 "unique_texts": 2,
             },
             "values_statistics": {"min_score": 0.0, "avg_score": 0.5, "max_score": 1.0},
-        }
+        },
+        "train": {
+            "num_samples": 2,
+            "number_of_characters": 53,
+            "num_texts_in_train": None,
+            "text_statistics": {
+                "total_text_length": 53,
+                "min_text_length": 23,
+                "average_text_length": 26.5,
+                "max_text_length": 30,
+                "unique_texts": 2,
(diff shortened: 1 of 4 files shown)


mteb/abstasks/any_classification.py
```diff
@@ -187,8 +188,6 @@ def _evaluate_subset(
     ) -> FullClassificationMetrics:
         train_split = data_split[self.train_split]
         eval_split = data_split[hf_split]
-        params = {"k": self.k}
-        params.update(kwargs)
```
Samoed, line 191: I think I added this when we had different evaluator classes and `k` was used by knn

tests/test_tasks/test_predictions.py
KennethEnevoldsen, line 1: I think these changes are superseded by #3327

Samoed:
Yes, I saw that you used approx. I should wait until you merge your pr?

KennethEnevoldsen:
If it is good to merge then I can just merge it now

Samoed:
Yes, I've finished this PR and you don't have comments we can merge it then

Issue #2957, opened 2025-07-30
Create input specific mock model


Samoed:
> Yes. I have had that problem before as well. I have actually considered replacing the mockencoder with a small model2vec model 

We can use approach as is in langchain fake embedding by setting seed from the text https://github.com/langchain-ai/langchain/blob/master/libs%2Fcore%2Flangchain_core%2Fembeddings%2Ffake.py#L120

_Originally posted by @Samoed in https://github.com/embeddings-benchmark/mteb/issues/2885#issuecomment-3118785857_

PR #3327, opened 2025-10-12
[v2] Additional fixes for tests

### LINKED ISSUE #3324: [v2] fix tests that fail only when run locally
I think this is the remaining tests that needs to be solved, but it is down to 3 min:

Screenshot 2025-10-12 at 17 56 52

[...]

It seems to only be a local error, though (I suspect it's something to do with the numpy version, seed, or something like that). Anyway, that shouldn't happen, so I will fix that. However, assuming this runs fine - do feel free to merge (and then I will do it in another PR)

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/issues/3306#issuecomment-3394764102_

KennethEnevoldsen:
- [x] removed duplicate test testing for MIEB models + encoders the current one is in `test_integrations/test_image_model_task.py`
- [x] figure out where additional  files come from and send them to tmp_path
- [x] fix test which fails locally only by introducing a new model `mteb/random-baseline` that produced a random vector which is consistent on the input
- [x] minor fixes to cases which didn't handle numpy as output
- [x] removed test which tests that private datasets are in fact listed as private (we have the inverse)
- [x] reduced grid size of `test_encoder_dtype_on_task`
- [x] combined tests for metadata so that we have ~1k test instead of ~3k for metadata (runs a bit faster)
- [x] reduced task_grid to only include one classification task and switched out the classification task for a smaller one
- [x] replaced test_cli classification task with a smaller one

All of this combined put us under 3 mins locally (hot cache)

Screenshot 2025-10-12 at 21 09 54

fixes https://github.com/embeddings-benchmark/mteb/issues/2957
fixes #3324 

Added the `mteb/random-baseline` as a fully fledged model, as I could imagine it being relevant as a baseline as well.

### DIFF
--- a/mteb/models/model_implementations/random_baseline.py
+++ b/mteb/models/model_implementations/random_baseline.py
@@ -1,6 +1,7 @@
 from typing import Any
 
 import numpy as np
+from PIL import Image
 from torch.utils.data import DataLoader
 
 from mteb.abstasks.task_metadata import TaskMetadata
@@ -33,9 +34,22 @@ class RandomBaseline:
     ) -> Array:
         all_embeddings = []
         for batch in inputs:
-            for text in batch["text"]:
-                vector = self._string_to_vector(text, self.embedding_dim)
-                all_embeddings.append(vector)
+            if "text" in batch and "image" in batch:
+                for text, image in zip(batch["text"], batch["image"]):
+                    text_vector = self._string_to_vector(text, self.embedding_dim)
+                    image_vector = self._image_to_vector(image, self.embedding_dim)
+                    combined_vector = (text_vector + image_vector) / 2
+                    all_embeddings.append(combined_vector)
+            elif "image" in batch:
+                for image in batch["image"]:
+                    vector = self._image_to_vector(image, self.embedding_dim)
+                    all_embeddings.append(vector)
+            elif "text" in batch:
+                for text in batch["text"]:
+                    vector = self._string_to_vector(text, self.embedding_dim)
+                    all_embeddings.append(vector)
+            else:
+                raise KeyError("Input batch must contain 'text' and/or 'image' keys.")
         return np.vstack(all_embeddings)
 
     @staticmethod
@@ -47,6 +61,15 @@ class RandomBaseline:
         seed = int.from_bytes(s.encode("utf-8"), byteorder="big") % (
             2**32
         )  # numpy rng seed must be between 0 and 2**32
+        rng = np.random.default_rng(seed)
+        return rng.random(size)
+
+    @staticmethod
+    def _image_to_vector(image: Image.Image, size: int):
+        """Generate a deterministic random vector based on image content."""
+        # Convert image to bytes and then to a numeric seed
+        image_bytes = image.tobytes()
+        seed = int.from_bytes(image_bytes, byteorder="big") % (2**32)
         rng = np.random.default_rng(seed)
         return rng.random(size)
 
@@ -78,6 +101,7 @@ class RandomBaseline:
 random_baseline = ModelMeta(
     loader=RandomBaseline,  # type: ignore
     name="mteb/random-baseline",
+    modalities=["text", "image"],
     languages=None,
     open_weights=True,
     revision="1",
--- a/tests/test_tasks/test_predictions.py
+++ b/tests/test_tasks/test_predictions.py
@@ -1,6 +1,5 @@
 import json
 from pathlib import Path
-from typing import Any
 
 import pytest
 
@@ -15,7 +14,6 @@ from tests.mock_tasks import (
     MockRegressionTask,
     MockRetrievalTask,
     MockSTSTask,
-    MockSummarizationTask,
     MockTextZeroShotClassificationTask,
 )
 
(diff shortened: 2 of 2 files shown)


mteb/models/model_implementations/random_baseline.py
```diff
@@ -0,0 +1,93 @@
...
+        if s is None:
+            s = ""
+        # Convert string to a numeric seed
+        seed = int.from_bytes(s.encode("utf-8"), byteorder="big") % (
+            2**32
+        )  # numpy rng seed must be between 0 and 2**32
+        rng = np.random.default_rng(seed)
+        return rng.random(size)
+
+    def similarity(
```
Samoed, line 76: I think this would be automatically cosine based on `model_meta`

tests/test_tasks/test_predictions.py
```diff
@@ -22,62 +21,62 @@
...
+                [0, 0],
+                [0, 0],
+                [0, 0],
+                [0, 0],
+                [0, 0],
+                [0, 0],
+                [0, 0],
+                [0, 0],
+                [0, 0],
+                [0, 0],
```
Samoed, line 45: This is strange that output of classification task only 0

mteb/models/model_implementations/random_baseline.py
```diff
@@ -0,0 +1,93 @@
...
+import numpy as np
+from torch.utils.data import DataLoader
+
+from mteb.abstasks.task_metadata import TaskMetadata
+from mteb.models import Encoder
+from mteb.models.model_meta import ModelMeta
+from mteb.types._encoder_io import Array, BatchedInput, PromptType
+
+
+class RandomBaseline(Encoder):
```
Samoed: Great addition! We can reuse it in mock models. Also close https://github.com/embeddings-benchmark/mteb/issues/2957

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: Hmm it fails if it is not there. Do we add the similarity function somewhere if it is not?

KennethEnevoldsen: (might be because I am inheriting from Encoder)

tests/test_tasks/test_predictions.py
KennethEnevoldsen: Hmm yeah can't tell you why though

mteb/models/model_implementations/random_baseline.py
Samoed: Yes, I've added in absencoder https://github.com/embeddings-benchmark/mteb/blob/60adf65d542d2de075f759ed32fd1d4f38618b96/mteb/models/abs_encoder.py#L237

tests/test_integrations/test_dtype_encoders.py
```diff
@@ -7,20 +7,54 @@
...
+        MockRerankingTask(),
+        MockRetrievalTask(),
+        MockSTSTask(),
+        MockMultilabelClassification(),
+        MockSummarizationTask(),
+        MockInstructionRetrieval(),
+        MockInstructionReranking(),
+        MockRetrievalDialogTask(),
+        MockTextZeroShotClassificationTask(),
+    ],
```
Samoed: Why not to use mock task test grid here?

mteb/models/model_implementations/random_baseline.py
```diff
@@ -0,0 +1,93 @@
...
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> Array:
+        all_embeddings = []
+        for batch in inputs:
+            for text in batch["text"]:
```
Samoed: Maybe create for each input embedding to make random model multimodal?

tests/test_integrations/test_dtype_encoders.py
KennethEnevoldsen: This is mostly to test the flow of dtypes throughout the codebase. Multilingual subsets generally use the same underlying flow (evaluate_subset) so we don't need to test it twice. This reduced the size of the grid quite a lot, given that we are testing multiple models

Samoed: Maybe we can create grid with these tasks and then combine them with multilingual to full grid?

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: Right, so it only works when you inherit from AbsEncoder? I would prefer keeping this function independent. It has already revealed some issues with similarity being assumed to be a torch tensor

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: I am all for doing this in another PR when relevant

mteb/models/model_implementations/random_baseline.py
Samoed: > Right, so it only works when you inherit from AbsEncoder?

Yes

> Right, so it only works when you inherit from AbsEncoder?

Leave only in `AbsEncoder`? Don’t we support both numpy and torch?

tests/test_integrations/test_dtype_encoders.py
KennethEnevoldsen: agree - fixed!

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: In this PR, I fixed a few cases that didn't work if the output of similarity was a numpy vector. 

Maybe the MockEncoders shouldn't inherit from AbsMockEncoder?

KennethEnevoldsen: I also see that we have `MockSentenceTransformersbf16Encoder` which I would probably convert to `MockSentenceTransformersbf16Encoder` > `MockTorchbf16Encoder`

Samoed: Maybe, I want to refactor test models a bit

KennethEnevoldsen: alright I will leave it for another PR

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: Added this here as it was needed for there merge

PR #3333, opened 2025-10-13
Remove CHAIN19 from leaderboard

Samoed:
I was looking on leaderboard and found that in `v1` leaderboard it doesn't exist, but we have it on `v2`


Samoed:
They've deleted their repo, so I will remove it

PR #3343, opened 2025-10-13
activate `TID`

Samoed:
Make imports cleaner


KennethEnevoldsen:
The fact that we can now do this without circular import issues tells us a lot about how nice it is to work with now

Issue #3325, opened 2025-10-12
Add tests with crossencoders


Samoed:
`v2` have better support for `cross-encoders`, but we don't have good tests for them

KennethEnevoldsen:
I believe this was added in #3347

Samoed:
Yes, I thought this was closed, but GitHub only automatically closes issues for PRs that target the `main` branch.

PR #3347, opened 2025-10-13
Refactor models tests

Samoed:
- Renamed `mteb/random-baseline` to `mock/random-encoder-baseline`
- Changed `int.from_bytes(s.encode("utf-8"), byteorder="big")` to `int(hashlib.sha256(text.encode("utf-8")).hexdigest(), 16)` because it not correctly produced seeds
```python
texts = ['This is a test sentence', 'This is another test sentence']
seed_func = lambda text: int.from_bytes(text.encode("utf-8"), byteorder="big") % (2 ** 32)
print([seed_func(text) for text in texts])
# [1701733221, 1701733221]  # same seeds for different texts
```
- Added `mteb/random-crossencoder-baseline` and run retrieval tasks on it
- Added integration test with `cross-encoder/ms-marco-TinyBERT-L2-v2`
- Make only 1 test for 2 stage reranking with mock models and task `test_two_stage_reranking`
- Changed usage `MockNumpyEncoder` to `mock/random-encoder-baseline`

Close https://github.com/embeddings-benchmark/mteb/issues/3325


mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen, line 1: Looks great. I am unsure why the bytes didn't work though?

tests/test_abstasks/test_convert_to_reranking.py
KennethEnevoldsen, line 1: Thanks, love the change!

tests/test_integrations/test_mock_cross_encoder.py
```diff
@@ -0,0 +1,25 @@
...
+        MockRerankingTask(),
+        MockMultilingualRerankingTask(),
+        MockInstructionReranking(),
+        MockMultilingualInstructionReranking(),
+    ],
+)
+def test_mock_cross_encoder(task: AbsTask):
+    """Test that a task can be fetched and run"""
+    model = mteb.get_model("mteb/random-crossencoder-baseline")
+    mteb.evaluate(model, task, cache=None)
```
KennethEnevoldsen, line 25: great addition, but there seems to have been some copy paste



```suggestion
@pytest.mark.parametrize(
    "task",
    [
        MockRerankingTask(),
        MockMultilingualRerankingTask(),
        MockInstructionReranking(),
        MockMultilingualInstructionReranking(),
    ],
)
def test_cross_encoder_on_task(task: AbsTask):
    """Ensures that cross-encoders can be run on retrieval and reranking tasks"""
    model = mteb.get_model("mteb/random-crossencoder-baseline")
    mteb.evaluate(model, task, cache=None)
```

We should also add the expected output here

tests/test_integrations/test_integration_with_sentencetransformers.py
```diff
@@ -26,3 +32,19 @@ def test_sentence_transformer_integration(task: AbsTask, model_name: str):
...
+        MockMultilingualRerankingTask(),
+        MockInstructionReranking(),
+        MockMultilingualInstructionReranking(),
+    ],
+)
+@pytest.mark.parametrize("model_name", ["cross-encoder/ms-marco-TinyBERT-L2-v2"])
+def test_sentence_transformer_integration_cross_encoder(task: AbsTask, model_name: str):
+    """Test that a task can be fetched and run"""
+    model = CrossEncoder(model_name)
+    mteb.evaluate(model, task, cache=None)
```
KennethEnevoldsen, line 46: Do we already download this model? I don't want to increase the download time. I would also reduce to only non-multilingual tasks (the more extensive grid is tested with the random baseline)

Should we move this together with the other cross-encoder test?

(I am unsure if these tests are integrations or model tests)

mteb/models/model_implementations/random_baseline.py
```diff
@@ -20,7 +67,7 @@ class RandomBaseline:
 
     def __init__(self, model_name: str, revision: str | None, **kwargs: Any) -> None:
```
KennethEnevoldsen: We could give this "dtype" as input and essentially remove many of our mock cases - WDYT?

mteb/models/model_implementations/random_baseline.py
```diff
@@ -20,7 +67,7 @@ class RandomBaseline:
...
         self.rng_state = np.random.RandomState(42)
```
KennethEnevoldsen: I know I wrote this but should we use the recommended default instead? (numpy.random.default_rng)

mteb/models/model_implementations/random_baseline.py
Samoed: I think this because you've selected `2**32` as mod value. If you select `1e7+7` numbers are different. This happens because `from_bytes` produced very big numbers
```python
2 ** 32
# 4294967296
int.from_bytes("This is a test sentence".encode("utf-8"), byteorder="big")
# 8084666724066882667111151064961595047191437202623980389
int.from_bytes("sentence".encode("utf-8"), byteorder="big")
# 8315173733206483813
int.from_bytes("ence".encode("utf-8"), byteorder="big")
# 1701733221
seed_func("sentence")
# 1701733221
seed_func("ence")
# 1701733221
seed_func("This is a test sentence")
# 1701733221
```

UPD Tried to find what happening
> When you’re taking modulo 2**32, only the last 4 bytes of the entire UTF-8 representation affect the remainder. If the last few bytes are the same (like "ence" in both sentences), you’ll get identical seeds.

mteb/models/model_implementations/random_baseline.py
Samoed: Yes, I thought about how to modify this, and I think this would work

tests/test_integrations/test_integration_with_sentencetransformers.py
Samoed: > Do we already download this model?

Previously it was in 2 stage rerank test and it weights only 17mb. (In this test was loading `SciFact` and `average_word_embeddings_kominos_dependency`) `average_word_embeddings_levy_dependency` [weights 200mb](https://huggingface.co/sentence-transformers/average_word_embeddings_levy_dependency/tree/main/0_WordEmbeddings) * 2 so I don't see a problem to load it here

> Should we move this together with the other cross-encoder test?

This file about integration test with sentence transformers, so I think this is appropriate place

> (I am unsure if these tests are integrations or model tests)

This test tests that we correctly work with `CrossEncoder` passed directly as previous test with `SentenceTransformers`

mteb/models/model_implementations/random_baseline.py
Samoed: Removed most of the mock models. Only `MockSetnenceTrsnformer` models are left

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: Ahh that make sense. Yeah a hash is better

mteb/models/model_implementations/random_baseline.py
```diff
@@ -91,31 +138,96 @@ def similarity_pairwise(
...
-    max_tokens=np.inf,
-    reference=None,
-    similarity_fn_name="cosine",  # type: ignore
-    framework=[],
-    use_instructions=False,
-    public_training_code=None,  # No training code, as this is a random baseline
-    public_training_data=None,  # No training data, as this is a random baseline
-    training_datasets=set(),
+    **_common_mock_metadata,
+)
```
KennethEnevoldsen, line 159: Hmm, I wouldn't have a `mock` model in our suite of models, but a random baseline seems reasonable (would keep as mteb)

mteb/models/model_implementations/random_baseline.py
```diff
@@ -91,31 +138,96 @@ def similarity_pairwise(
...
+random_torch_encoder_baseline = ModelMeta(
+    loader=RandomEncoderBaseline,  # type: ignore
+    loader_kwargs=dict(
+        is_torch=True,
+        torch_dtype=torch.float32,
+    ),
+    name="mock/random-torch-encoder-baseline",
+    modalities=["text", "image"],
+    **_common_mock_metadata,
+)
```
KennethEnevoldsen: we can replace this in the test with:

`mteb.get_model("mteb/random-baseline", use_torch=True, torch_dtype=torch.float32)`

I don't think we need to specify a new model

mteb/models/model_implementations/random_baseline.py
```diff
@@ -91,31 +138,96 @@ def similarity_pairwise(
...
+
+random_image_only_encoder_baseline = ModelMeta(
+    loader=RandomEncoderBaseline,  # type: ignore
+    name="mock/random-image-only-encoder-baseline",
+    modalities=["image"],
+    **_common_mock_metadata,
+)
```
KennethEnevoldsen: Here we would have to do slightly more:

```
mdl = mteb.get_model("mteb/random-baseline")
mdl.mteb_model_meta.modalities = ["image"]
```

mteb/models/model_implementations/random_baseline.py
```diff
@@ -18,9 +88,18 @@ class RandomBaseline:
...
     mteb_model_meta: ModelMeta | None = None
 
-    def __init__(self, model_name: str, revision: str | None, **kwargs: Any) -> None:
-        self.rng_state = np.random.RandomState(42)
-        self.embedding_dim = 32  # not sure it matters what dimension we use here
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None,
+        is_torch: bool = False,
```
KennethEnevoldsen: use_torch? Maybe it is better with `array_framework`?

tests/mock_models.py
KennethEnevoldsen, line 1: Should we also remove MockSentenceTransformersbf16Encoder?

KennethEnevoldsen:
@Samoed ran these new tests locally and it seems like the clustering task test fails, when I run it in parallel (using `make test`:

Screenshot 2025-10-14 at 09 41 58

However, if I just run the test individually, this does not happen. Ran it twice to ensure that it was not just accidental.

The test also seems to have gotten 20% slower? Not that it is a big deal.

mteb/models/model_implementations/random_baseline.py
Samoed: Numpy is an array framework too. We have some torch specific parts that I wanted to test

mteb/models/model_implementations/random_baseline.py
Samoed: I thought putting `mteb` as prefix can confuse people that we have some models in our organization

mteb/models/model_implementations/random_baseline.py
Samoed: Yes, I thought about it too, but thought creating different models would be cleaner, but agree this is better

Samoed:
> The test also seems to have gotten 20% slower? Not that it is a big deal.

I don't see difference in CI comparing to other PRs

tests/mock_models.py
Samoed: We can, but numpy doesn't support bf16, and it tests that our sentence transformer wrapper can support bf16 models. I think we can change them later, because for now I'm not sure what to do with them

Samoed:
> it seems like the clustering task test fails, when I run it in parallel

I had a problem with clustering where it passed locally but failed on CI. Setting `random_seed` solved the issue. Maybe sklearn handles random seeds a bit differently in parallel mode (but on CI uses two workers), or perhaps on macOS sklearn manages seeds differently. I don't know what to do.

mteb/models/model_implementations/random_baseline.py
Samoed: Removed other models in https://github.com/embeddings-benchmark/mteb/pull/3347/commits/538c5e3d46e18c05fce6c442f3bd2eb64e8bdb00

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: Hmm yeah. I would say that is fine. Alternatively, `baseline/random-baseline`?

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: Sure I was thinking:

```
array_framework: list[Literal["numpy", "torch"]]
```

tests/mock_models.py
KennethEnevoldsen: Ahh, so there are two things to test:
1) does the wrapper support bf16
2) does the tasks support bf16 

I was mostly thinking about the second one. I think it is fine to leave as is then

KennethEnevoldsen:
> I had a problem with clustering where it passed locally but failed on CI. Setting random_seed solved the issue. Maybe sklearn handles random seeds a bit differently in parallel mode (but on CI uses two workers), or perhaps on macOS sklearn manages seeds differently. I don't know what to do.

I will give it a go

mteb/models/model_implementations/random_baseline.py
Samoed: Changed to `array_framework: Literal["numpy", "torch"]` and added `dtype` argument to suport numpy too

Samoed:
Interesting. Now python3.10 fails

KennethEnevoldsen:
Hmm, it might be that it is an edge case because there isn't really enough info in the task so it essentially just becomes random, what if we instead use:

```
            "This is a test sentence",
            "This is a test sentence",
            "This is a test sentence",
            "very different, very unique",
            "very different, very unique",
            "very different, very unique",
        ]
        labels = [0, 0, 0, 1, 1, 1]
```

Samoed:
@KennethEnevoldsen It's passed now somehow... Should we try more, or we can leave it as is then?

KennethEnevoldsen:
Oh let us just leave it then

Issue #3350, opened 2025-10-13
Filter out unsupported modalities by `mteb.evaluate`


Samoed:
https://github.com/embeddings-benchmark/mteb/blob/ce42692115d4941adde4f7ed75dc986965982ecb/tests/test_integrations/test_modality.py#L26

PR #3351, opened 2025-10-13
[v2] Filter out tasks by models modality

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3350

### DIFF
--- a/tests/test_integrations/test_modality.py
+++ b/tests/test_integrations/test_modality.py
@@ -8,7 +8,6 @@ from copy import deepcopy
 import pytest
 
 import mteb
-import mteb.overview
 from mteb.types import PromptType
 from tests.mock_tasks import (
     MockImageClusteringTask,


mteb/evaluate.py
```diff
@@ -141,6 +141,21 @@ def _evaluate_task(
 
     task.check_if_dataset_is_superseded()
 
+    if model.mteb_model_meta.modalities is not None and not (
```
KennethEnevoldsen: Hmm, but could I run a text-only model on an image-text retrieval task (just on the text  part?)

I would probably just give a warning

mteb/evaluate.py
```diff
@@ -141,6 +141,21 @@ def _evaluate_task(
...
+        logger.warning(
+            f"Model {model.mteb_model_meta.name} support modalities {model.mteb_model_meta.modalities} but the task "
+            f"{task.metadata.name} only supports {task.modalities}. Skipping task."
+        )
+        return TaskResult.from_task_results(
+            task,
+            task_results,
+            evaluation_time=0,
+            kg_co2_emissions=None,
+        )
```
KennethEnevoldsen: I would not return an empty task result if the task does not run - maybe we should move the check up to the evaluate function (where an empty list is a viable solution)

mteb/evaluate.py
KennethEnevoldsen: However if it does not overlap (e.g. image task and text model) then I would just raise an error

Samoed: > Hmm, but could I run a text-only model on an image-text retrieval task (just on the text part?)

I don't think that you can do this, because you will have an error during running an `encode` when you will try to encode images that model doesn't support, and it probably output just empty array or error during looping thorough batches

mteb/evaluate.py
Samoed: Moved into `evaluate`

mteb/evaluate.py
Samoed: Hm, maybe you mean tasks like `it2t`. We can add support for this kind of cases but I unsure if we want to

KennethEnevoldsen: Hmm, but for a text retrieval from a corpus of markdown with images. A solid baseline to consider is retrieval with just a text encoder.

Samoed: I will add support then for this case

mteb/evaluate.py
```diff
@@ -344,6 +344,20 @@ def evaluate(
...
         logger.info("✓ Model loaded")
 
+    if model.mteb_model_meta.modalities is not None and not (
+        # all task modalities are contained within model modalities
+        set(task.metadata.modalities).issubset(model.mteb_model_meta.modalities)
+    ):
+        logger.warning(
+            f"Model {model.mteb_model_meta.name} support modalities {model.mteb_model_meta.modalities} but the task "
+            f"{task.metadata.name} only supports {task.modalities}. Skipping task."
+        )
```
```suggestion
            f"Model {model.mteb_model_meta.name} support modalities {model.mteb_model_meta.modalities} but the task "
            f"{task.metadata.name} contains the following {task.modalities}. Skipping task."
        )
```

mteb/evaluate.py
```diff
@@ -344,6 +344,20 @@ def evaluate(
         model = model.load_model()
         logger.info("✓ Model loaded")
 
+    if model.mteb_model_meta.modalities is not None and not (
```
KennethEnevoldsen: Should we still raise a warning if the model does not support ALL modalities

mteb/evaluate.py
```diff
@@ -344,6 +344,20 @@ def evaluate(
...
+    ):
+        logger.warning(
+            f"Model {model.mteb_model_meta.name} support modalities {model.mteb_model_meta.modalities} but the task "
+            f"{task.metadata.name} only supports {task.modalities}. Skipping task."
+        )
+        return ModelResult(
+            model_name=model_name,
+            model_revision=model_revision,
+            task_results=[],
+        )
```
KennethEnevoldsen: Wouldn't we rather raise an error here? Or maybe this is friendlier (we, however, have the raise_errors=False)

mteb/evaluate.py
Samoed: Yes, but this can be unexpected if during evaluation model will fail. I will move up this check then to raise error before task running

mteb/evaluate.py
Samoed: Updated to support partly supported modalities runs

Samoed:
Seems that clustering task failed again (I rerun python3.11). I will try to fix it

mteb/evaluate.py
```diff
@@ -176,6 +178,70 @@ def _evaluate_task(
     return result
 
 
+def _check_model_modalities(
```
KennethEnevoldsen, line 181: Oh wow this is nice!

PR #3353, opened 2025-10-14
[v2] Rename task folders to align pep8 and move image tasks

### LINKED ISSUE #1337: MIEB: Figure out folder structure

isaac-chung:
Fixes #1337 . Groups tasks by task type [as per this comment](https://github.com/embeddings-benchmark/mteb/issues/1337#issuecomment-3395874976).

- Rename folder like `STS` -> `sts`
- Break `Image` folder and distribute tasks by task type, e.g. `VisualSTS` -> `sts`


KennethEnevoldsen:
Looking good! 

So we keep task scripts and non-pep8 compliant (e.g. [`KlueNLI.py`](https://github.com/embeddings-benchmark/mteb/blob/rename-task-folders-to-align-pep8/mteb/tasks/pair_classification/kor/KlueNLI.py)). Anyway this a great change.

Developers on macOS might have import issues, as the system does not detect the change in casing (the file system is case-insensitive). Just to make people aware.

(I am fine with merging this as it is)

isaac-chung:
Merging!

PR #3358, opened 2025-10-14
[v2] refactor filter_tasks.py out of overview.py and rename overview.py to get_tasks.py

KennethEnevoldsen:
Refactored filter_tasks out of overview.py, this both makes it easier to test segments of get_tasks and it also makes it accecible for use throughout the repo.

I also:
- refactored tests to reflect the new structure (gives a few minor speedups)
- added documentation for filter_tasks
- renamed overview.py to get_tasks.py

Q: `abstask.filter_modalities`, does it actually filter anything (doesn't it just change the metadata which isn't used anywhere?)

### DIFF
--- a/tests/test_deprecated/test_MTEB.py
+++ b/tests/test_deprecated/test_MTEB.py
@@ -8,7 +8,6 @@ from pathlib import Path
 import pytest
 
 import mteb
-import mteb.get_tasks
 
 logging.basicConfig(level=logging.INFO)
 


Samoed:
> Q: abstask.filter_modalities, does it actually filter anything (doesn't it just change the metadata which isn't used anywhere?)

Yes, it's weird. Maybe this was done like this to be used in leaderboard, but I'm not sure (source PR https://github.com/embeddings-benchmark/mteb/pull/2262)

KennethEnevoldsen:
> Yes, it's weird. Maybe this was done like this to be used in leaderboard, but I'm not sure (source PR https://github.com/embeddings-benchmark/mteb/pull/2262)

Hmm, I will try to remove it

KennethEnevoldsen:
Couldn't find any place where it was used

PR #3371, opened 2025-10-15
[v2]: Denote non-public functions

### LINKED ISSUE #2515: [v2] make a clear seperation of public and non-public functions

KennethEnevoldsen:
Should create_dataloaders also be private? (@Samoed)

- moved the generate model card to the CLI (should it be private?)
- made most leaderboard functions private
- made set_seed private
- made a few similarity function utilities private as well

fixes #2515 

(^maybe not fully)

### DIFF
--- a/mteb/cli/build_cli.py
+++ b/mteb/cli/build_cli.py
@@ -8,8 +8,8 @@ from rich.logging import RichHandler
 
 import mteb
 from mteb.cache import ResultCache
-from mteb.evaluate import OverwriteStrategy
 from mteb.cli.generate_model_card import generate_model_card
+from mteb.evaluate import OverwriteStrategy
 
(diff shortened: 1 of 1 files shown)


Samoed:
> Should create_dataloaders also be private

Yes, I think it's better

Samoed (approved):

Samoed (approved):

Samoed:
I think we would find more functions if we activate `DOC` rule for ruff, but it would produce 150+ errors of missing docstrings or missing parameters (tested couple days ago)

KennethEnevoldsen:
Would probably be great to do that

PR #3375, opened 2025-10-15
[V2] Activate ruff `D` and `W` rules. Partly run `DOC`

Samoed:
Mostly it requires adding `Returns` if there is any docstring in function. Also activated `W`, but it didn't produce any changes


mteb/models/model_implementations/random_baseline.py
```diff
@@ -56,7 +72,19 @@ def _image_to_vector(image: Image.Image, size: int):
...
+    Args:
+        inputs: A DataLoader yielding batches of inputs, where each batch is a dictionary
+                that may contain 'text' and/or 'image' keys.
+        embedding_dim: The dimensionality of the output embeddings.
+
+    Returns:
+        A 2D numpy array of shape (num_samples, embedding_dim) containing the embeddings
+
+    Raises:
+        KeyError: If neither 'text' nor 'image' keys are present in the input batch.
```
KennethEnevoldsen: Do we require this? I often don't find it that useful, as it doesn't tell you anything about what subfunctions might raise

mteb/leaderboard/table.py
```diff
@@ -207,7 +211,11 @@ def _apply_summary_table_styling(joint_table: pd.DataFrame) -> gr.DataFrame:
 
 
 def _apply_per_task_table_styling(per_task: pd.DataFrame) -> gr.DataFrame:
-    """Apply styling to a raw per-task DataFrame"""
+    """Apply styling to a raw per-task DataFrame
+
+    Returns:
+        Styled gr.DataFrame ready for display in the leaderboard
```
KennethEnevoldsen, line 217: Is this required? Seems odd to me when it is private

mteb/models/model_implementations/vista_models.py
```diff
@@ -76,9 +76,15 @@ def __init__(
...
-        def encode_text(self, texts):
+        def encode_text(self, texts: dict[str, torch.Tensor]) -> Array:
             """Currently override Visualized_BGE's the original implementation
             to fix attention_mask & embedding_output dtype misalignment
+
+            Args:
+                texts (dict): {"input_ids": ..., "attention_mask": ...}
+
+            Returns:
+                Array: text embeddings
```
KennethEnevoldsen:
```suggestion
            Args:
                texts: {"input_ids": ..., "attention_mask": ...}

            Returns:
                Array of text embeddings
```
If there is a rule to prevent types in docstrings that would be great

Samoed: > If there is a rule to prevent types in docstrings that would be great

I don't think so. Ruff doesn't have good support for docstring, but there are no tools left

mteb/leaderboard/table.py
Samoed: Yes, this is required by `DOC` rule, but we can disable it since I've added most of the doc and leave only `D` to validate missing args and etc

mteb/models/model_implementations/random_baseline.py
Samoed: This required by DOC, but we can disable it

mteb/models/model_implementations/vista_models.py
KennethEnevoldsen: Yeah, I actually just saw that I have an open issue on it over on ruff - they are working on getting better support for docstrings though

mteb/leaderboard/table.py
KennethEnevoldsen: What about just disabling the specific subrule for this?

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: I would just disable the subrule, but maybe it is fine to disable it completely also. Some tests are fine just to run once without maintaining them

Samoed: We can disable then
1. DOC201 [docstring-missing-returns](https://docs.astral.sh/ruff/rules/docstring-missing-returns/)
2. DOC501 [docstring-missing-exception](https://docs.astral.sh/ruff/rules/docstring-missing-exception/)

Samoed: Hm, but without them this rule useless I think. We can disable it then

Samoed: I disable it. Should I remove raises or returns?

mteb/leaderboard/table.py
Samoed: Found issue on private/public functions https://github.com/astral-sh/ruff/issues/20280

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: Yeah let us remove raises (returns are generally good to have (I was hoping we could just keep it for public)

KennethEnevoldsen:
In think once we have removed raises this is good to merge

mteb/models/model_implementations/random_baseline.py
Samoed: Fond that we have disabled most of `D` rules. I will try to run them again and select useful

Samoed:
Activated `D101-103` and `D205` rules

```
    "D101",     # Missing docstring in public class
    "D102",     # Missing docstring in public method
    "D103",     # Missing docstring in public function
    "D205",     # 1 blank line required between summary line and description
```

KennethEnevoldsen:
Merge conflicts have to be resolved, but otherwise, I think this is good to merge

Issue #3145, opened 2025-09-07
[v2] refactor or benchmark.load_results to use


KennethEnevoldsen:
### Description of the feature

Currently, it uses the deprecated load_results, which seems unideal. I would probably deprecate this and instead use the `cache.load_results(tasks=benhcmark.tasks)`

PR #3376, opened 2025-10-15
benchmark use cache directly

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3145

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -217,9 +217,7 @@ def get_leaderboard_app(cache: ResultCache = ResultCache()) -> gr.Blocks:
         mteb.get_benchmarks(display_on_leaderboard=True), key=lambda x: x.name
     )
     all_benchmark_results = {
-        benchmark.name: all_results._filter_tasks(
-            [t.metadata.name for t in benchmark.tasks]
-        ).join_revisions()
+        benchmark.name: all_results.select_tasks(benchmark.tasks).join_revisions()
         for benchmark in benchmarks
     }
     default_benchmark = mteb.get_benchmark(DEFAULT_BENCHMARK_NAME)


mteb/benchmarks/benchmark.py
```diff
@@ -51,6 +51,7 @@ class Benchmark:
     display_on_leaderboard: bool = True
     icon: str | None = None
     display_name: str | None = None
+    _results_cache: dict = field(default_factory=dict, init=False, repr=False)
```
KennethEnevoldsen: I am unsure why this is needed. Is it used in the leaderboard?

Samoed: This used in `load_results`. I just moved declaration of this field directly. Yes, this used in leaderboard
https://github.com/embeddings-benchmark/mteb/blob/2ac453d54e556c8bdaddcbd27efc3aba72302792/mteb/leaderboard/app.py#L220

KennethEnevoldsen: hmm right. This function currently seems more suited for the leaderboard than for user. I would probably create a function in the leaderboard instead:

```
def load_benchmark_results(
        benchmark,
        base_results: BenchmarkResults | None = None,
        cache: ResultCache = ResultCache(),
    )

Samoed: I changed logic a bit, but I think that should work

Samoed:
Results are a bit different for `mteb(eng, v2)`. E.g. misses top 1 model QZhou-Embedding, but I don't understand why. Maybe something different in `_filter_tasks`. Can you why this can happen?

Samoed:
After merging `v2` results are the same. You can check them in https://huggingface.co/spaces/mteb/leaderboard_2_demo

KennethEnevoldsen:
Hmm it does seems the same to me:

Screenshot 2025-10-16 at 13 37 14
Screenshot 2025-10-16 at 13 37 25

Samoed:
Yes, maybe I saw version from `v2`. Will try to debug a bit then

Samoed:
Fixed problem. Previously I used `_filter_tasks` which recives just task names and load all data for the task. Changed to `select_task` which load tasks from `AbsTask`
image

PR #3372, opened 2025-10-15
(v2.0.0) (bugfix) Correcting the task filtering issue

fzoll:
Correcting the task filtering issue and adding tests

PR #3387, opened 2025-10-16
Move `superseded_by` to `TaskMetadata`

Samoed:
Move `superseded_by` to `TaskMetadata`. Added `AbsTask.superseded_by` property for compatablity. 

Added `supressed_by` to similar tasks computing


isaac-chung:
Might have missed the discussion. Wondering why?

Samoed:
There is no discussion, just this https://github.com/embeddings-benchmark/mteb/pull/3372#discussion_r2433540544. I just think that `suppresed_by` is used in all tasks, and it would be better to put it in `TaskMetadata`

isaac-chung:
_How_ is it better?

Samoed:
1. TasksMetadata validates all fields, and we can avoid mistakes of typos.
2. This is just metadata field and not used during evaluation. This is better separation between task type specific variables (attributes of classes) and metadata fields that can be used in all task types
3. We can simplify computation of zero shot score possibly by updating `collect_similar_tasks` (update, added to zero shot computation)

Samoed:
Previously it was just a field in abstasks https://github.com/embeddings-benchmark/mteb/blob/721e8a34743c050179ee5e9093ed24ee7ab37fba/mteb/abstasks/AbsTask.py#L63 and I moved it to task metadata
https://github.com/embeddings-benchmark/mteb/blob/b6496477873856e48d0a78a8e48646b4304e995c/mteb/abstasks/task_metadata.py#L270

KennethEnevoldsen:
Ahh so `supressed_by` (spelling is different) wasn't introduced. Then I think this PR is all good

Samoed:
Oh, sorry :sweat_smile:

PR #3390, opened 2025-10-16
fully activate `N`

Samoed:
Finish adopting `N` pep8 names rule. Put all tasks folder in ignore rule

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,12 +9,17 @@ description = "Massive Text Embedding Benchmark"
 readme = "README.md"
 authors = [
     { name = "MTEB Contributors", email = "niklas@huggingface.co" },
-    { email = "kenneth.enevoldsen@cas.au.dk" },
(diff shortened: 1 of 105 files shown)


pyproject.toml
```diff
@@ -243,12 +226,21 @@ ignore = [
 ]
 
 [tool.ruff.lint.per-file-ignores]
-"scripts/*" = ["PTH"]
+"scripts/*" = ["PTH", "N"]
 "tests/*" = ["RUF012"]
+"mteb/tasks/*" = ["N999"]
+"mteb/MTEB.py" = ["N999"]
```
Samoed: We can rename this file to `mteb.py` or `legacy_evaluator` and users still can use `mteb.MTEB`, but this is very minor

KennethEnevoldsen: what about `deprecated_evaluator`

Samoed: Renamed

Issue #3192, opened 2025-09-18
Add faiss for single vector models and deprecate CacheWrapper


orionw:
### Description of the feature

Since we now can have indexing as a first class function, we should deprecate the `CacheWrapper` and add `faiss` as an optional indexing step. 

This opens up a lot more options for storing vectors and should be a lot faster as well. 

Happy to take a first stab at this sometime if helpful @KennethEnevoldsen @Samoed.  We can start with a simple flat index ([very simple](https://github.com/facebookresearch/faiss/wiki/Getting-started#in-python-1)) but could also let the user pass `index_kwargs` to make it blazingly fast at huge scale.

Samoed:
Yes, we can update `CacheWrapper` to use `faiss`, but I don’t think that would be a good solution since it’s more targeted at ANN, while the cache wrapper just gets cached embedding for a query.

I think it would be better to use `faiss` as a new search index for retrieval instead.

orionw:
Faiss does do approximate, but it also has exact (ie "flat"). You can then extract by id from Faiss also, but I agree the CacheWrapper is easier to use.  

If we want to do both we can keep CacheWrapper and add optional Faiss indexing

Samoed:
Then I think we can change it then

Samoed:
I tried to play with faiss a bit and I got x3 speed up on basic search over niave looping over sentence transformers embeddings. Maybe we can add FAISS index to the `SearchWrapper` instead. Datasets have integration with it and this will simplify a lot https://huggingface.co/docs/datasets/v4.1.1/en/package_reference/main_classes#datasets.Dataset.add_faiss_index. @KennethEnevoldsen What do you think? At least we can add this as an alternative method to the current

orionw:
> I tried to play with faiss a bit and I got x3 speed up

🚀🚀🚀 Amazing, thanks for looking into this!

KennethEnevoldsen:
Sounds great!!

PR #3402, opened 2025-10-17
feat: Add faiss cache backend

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3192

Added `FAISS` to cache backend. Now `CacheEmbeddingsWrapper` can support multiple backends that can be passed directly.

```python
from mteb.models.cache_wrappers.cache_backends.faiss_cache import FaissCache

model = mteb.get_model(...)
cachedmodel = CachedEmbeddingWrapper(model, cache_dir, cache_backend=FaissCache)
```

Probably something similar can be done for `Retrieval` tasks

### DIFF
--- a/mteb/models/cache_wrappers/cache_backends/__init__.py
+++ b/mteb/models/cache_wrappers/cache_backends/__init__.py
@@ -1,0 +1,7 @@
+from .faiss_cache import FaissCache
+from .numpy_cache import NumpyCache
+
+__all__ = [
+    "FaissCache",
+    "NumpyCache",
+]


tests/test_models/test_cached_embedding_wrapper.py
KennethEnevoldsen, line 1: Should we test the different cache backends

mteb/models/cache_wrappers/cache_backends/cache_map.py
```diff
@@ -0,0 +1,189 @@
...
+import numpy as np
+
+from mteb.types import BatchedInput
+
+from ._hash_utils import _hash_item
+
+logger = logging.getLogger(__name__)
+
+
+class VectorCacheMap:
```
KennethEnevoldsen: Should we rename this to NumpyCache?

mteb/models/cache_wrappers/cache_backend_protocol.py
```diff
@@ -0,0 +1,57 @@
...
+from pathlib import Path
+from typing import Any, Protocol, runtime_checkable
+
+import numpy as np
+
+from mteb.types import BatchedInput
+
+
+@runtime_checkable
+class CacheBackendProtocol(Protocol):
```
KennethEnevoldsen, line 12: Loving it - this makes it very easy to change + test

mteb/models/cache_wrappers/cache_backend_protocol.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __init__(self, directory: Path | None = None, **kwargs: Any) -> None:
+        """Initialize the cache backend.
+
+        Args:
+            directory: Directory path to store cache files.
+            **kwargs: Additional backend-specific arguments.
+        """
+
+    def add(self, item: BatchedInput, vector: np.ndarray) -> None:
```
KennethEnevoldsen: Isn't this non-batched? (typing might be slightly confusing)

mteb/models/cache_wrappers/cache_wrapper.py
```diff
@@ -268,9 +89,7 @@ def encode(
         task_name = task_metadata.name
         try:
             if task_name not in self.cache_dict:
-                self.cache_dict[task_name] = _VectorCacheMap(
-                    self.cache_path / task_name
-                )
+                self.cache_dict[task_name] = self.backend(self.cache_path / task_name)
```
KennethEnevoldsen: So, a whole new cache PR task - I am fine with this, but it prevents re-use across

tests/test_models/test_cached_embedding_wrapper.py
Samoed: Yes, but I'm not sure if we want to add new dependency in tests

mteb/models/cache_wrappers/cache_wrapper.py
Samoed: Yes, agree. Can remove it

mteb/models/cache_wrappers/cache_backend_protocol.py
Samoed: For now, it is not batched, but I agree it's better to change

tests/test_models/test_cached_embedding_wrapper.py
KennethEnevoldsen: Is that a big problem?

mteb/models/cache_wrappers/cache_wrapper.py
KennethEnevoldsen: does the check include prompt (if not we probably have to handle that)

tests/test_models/test_cached_embedding_wrapper.py
Samoed: No, just I wasn't sure

mteb/models/cache_wrappers/cache_wrapper.py
```diff
@@ -220,19 +34,26 @@ class CachedEmbeddingWrapper:
         >>> mteb.evaluate(cached_model, task)
     """
 
-    def __init__(self, model: EncoderProtocol, cache_path: str | Path):
+    def __init__(
+        self,
+        model: EncoderProtocol,
+        cache_path: str | Path,
+        backend: type[CacheBackendProtocol] = VectorCacheMap,
```
KennethEnevoldsen: would rename to 
```suggestion
        cache_backend: type[CacheBackendProtocol] = VectorCacheMap,
```

mteb/models/cache_wrappers/cache_wrapper.py
Samoed: Yes, maybe this is the reason behind different indexes for different tasks. I think the simplest solution would be to keep cache per task to ensure that prompts will apply correctly, because they can change in `model.encode`

tests/test_models/test_cached_embedding_wrapper.py
Samoed: Add to the tests

pyproject.toml
```diff
@@ -91,6 +91,8 @@ ark = ["volcengine-python-sdk[ark]==3.0.2", "tiktoken>=0.8.0"]
 colpali_engine = ["colpali_engine>=0.3.12"]
 xet = ["huggingface_hub>=0.32.0"]
 youtu = ["tencentcloud-sdk-python-common>=3.0.1454", "tencentcloud-sdk-python-lkeap>=3.0.1451"]
+faiss-cpu = ["faiss-cpu>=1.12.0"]
+faiss-gpu = ["faiss-gpu>=1.12.0"]
```
orionw: Did you test with cpu and gpu? It's been a minute since I used them, but IIRC the gpu one has some quirks.

mteb/models/cache_wrappers/cache_backend_protocol.py
orionw: +1, this is very nice!

orionw (approved):
Looks great! Excited to have this in 🙌

mteb/models/cache_wrappers/cache_backend_protocol.py
Samoed: Changed

pyproject.toml
Samoed: I faiss-gpu is deprecated on pypi, but exists only on conda. I'll leave only cpu version then

mteb/models/cache_wrappers/cache_wrapper.py
KennethEnevoldsen: yeah propably the best choice

Samoed:
@KennethEnevoldsen Is it good to merge?

docs/advanced_usage/cache_embeddings.md
```diff
@@ -29,3 +29,16 @@ vectors = np.asarray(vector_map.vectors)
...
 vectors = vectors[~zero_mask]
 ```
+
+### Different Cache backends
+
+By default, the `CachedEmbeddingWrapper` uses a NumPy memmap backend (`NumpyCache`) to store embeddings. However, you can also use other backends. Currently, only `FAISS` is implemented, but you can provide your own custom backend that implements the `CacheBackendProtocol` by passing it as the `search_backend` parameter when initializing the `CachedEmbeddingWrapper`. For example:
+
+```python
+import mteb
+from mteb.models.cache_wrappers.cache_backends.faiss_cache import FaissCache
```
KennethEnevoldsen: Shouldn't the import ideally look like this:
```suggestion
from mteb.models.cache_wrappers.cache_backends import FaissCache
```

PR #3404, opened 2025-10-17
make search encoder compatible with encoder

Samoed:
Previously, if execute this

```python
import mteb

task = model.get_task("SomeSTS")
```
would produce error. I think it's better to add this option if user wrap encoder model with search encoder to be able to execute all tasks

### DIFF
--- a/mteb/models/search_wrappers.py
+++ b/mteb/models/search_wrappers.py
@@ -4,6 +4,7 @@ from typing import Any
 
 import torch
 from datasets import Dataset
+from torch.utils.data import DataLoader
 
 from mteb._create_dataloaders import (
     create_dataloader,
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
I am not sure I get the error here, did you mean to write random-cross-encoder?

Samoed:
I forgot to add `SearchEncoderWrapper` to the example like in test
```python
import mteb
from mteb.models.sentence_transformer_wrapper import SentenceTransformerEncoderWrapper

model = SentenceTransformerEncoderWrapper(mteb.get_model("baseline/random-encoder-baseline"))
task = model.get_task("SomeSTS")
```

KennethEnevoldsen:
makes sense - all good

PR #3412, opened 2025-10-18
Make pair classification multimodal

Samoed:
I checked `MAEB` tasks https://github.com/embeddings-benchmark/mteb/issues/3410 and found that they're using `PairClassification` too. I've made it to support multimodal tasks now


KennethEnevoldsen (approved):

KennethEnevoldsen:
once the conflict is resolved it is all good

Issue #3405, opened 2025-10-17
Invalid links for models with huggingface references


KennethEnevoldsen:
Currently we by default add a link even if invalid:
 
```
[`baseline/random-cross-encoder-baseline`](https://huggingface.co/baseline/random-cross-encoder-baseline)

whybe-choi:
Hello, @KennethEnevoldsen !

This is a case of an invalid link that I found while adding citations.﻿

https://github.com/embeddings-benchmark/mteb/blob/634a1bbf33865b9f14261753c6015060e66ab2f2/mteb/models/model_implementations/codi_models.py#L128-L130

I think this is the same model as this one: 
https://github.com/embeddings-benchmark/mteb/blob/v2.0.0/mteb/models/model_implementations/youtu_models.py

KennethEnevoldsen:
Thanks @whybe-choi added a PR to remove it

ZBWpro:
Hello everyone, CoDi Embedding V1 has been renamed Youtu Embedding. So you can remove the old version.

KennethEnevoldsen:
Great to know @ZBWpro - thanks for the clarification!

ayush1298:
@KennethEnevoldsen In #3695 , I will be adding this functionality, so at that time, we can make a request on that link, and check whether we are getting status=200 in return or not?
Will it work by doing this?

Samoed:
I think this issue was fixed in https://github.com/embeddings-benchmark/mteb/pull/3426 and wasn't automatically closed because it was target non default branch (was targeted to v2) and was related to documentation

PR #3425, opened 2025-10-19
remove non-public model without API

KennethEnevoldsen:
The models seems to no longer be available. So I will take it down.

Mentioned https://github.com/embeddings-benchmark/mteb/issues/3405#issuecomment-3418074994

@ZBWpro seems like you submitted the original model


Samoed:
Maybe I've merged it iv v2 incorretly, because in main this is a different org https://github.com/embeddings-benchmark/mteb/blob/721e8a34743c050179ee5e9093ed24ee7ab37fba/mteb/models/youtu_models.py#L117

KennethEnevoldsen:
Yeah @ZBWpro just let us know that it was replaced - so we are free to remove

Samoed:
Maybe in main file was renamed, but in `v2` git lost that they're same files because we have tencent model https://github.com/embeddings-benchmark/mteb/blob/b0bf193843d9d1397ce4b37a0de4061f0ba783a4/mteb/models/model_implementations/youtu_models.py#L117

PR #3435, opened 2025-10-20
add citations to models

whybe-choi:
Ref #3388 

This pull request adds citation metadata to a wide range of model implementations in the repository, ensuring that each model now includes a proper reference to its source paper or resource. This improves the transparency and traceability of the models for users and researchers.

The most important changes are:

**Arctic Models:**
* Introduced `ARCTIC_V1_CITATION` and `ARCTIC_V2_CITATION` constants, and added the appropriate citation to all `arctic_embed` model variants in `arctic_models.py`. [[1]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R7-R26) [[2]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R161) [[3]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R186) [[4]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R211) [[5]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R236) [[6]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R260) [[7]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R289) [[8]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R314) [[9]](diffhunk://#diff-4a06b28579b4bba28b5430a0b8b2b35c1058671d6a8a7327d26cea7f59819517R338)

**Other Model Families:**
* Added citation strings to the following model files and attached them to their respective `ModelMeta` definitions:
  - B1ade models (`b1ade_models.py`)
  - BGE models (`bge_models.py`)
  - E5-V models (`e5_v.py`) [[1]](diffhunk://#diff-9fbf693d11bf8b5dd6d8a5d0c618f0cf5052eb7f71e3333950b358a5ab37232eR17-R26) [[2]](diffhunk://#diff-9fbf693d11bf8b5dd6d8a5d0c618f0cf5052eb7f71e3333950b358a5ab37232eR182)
[...]

PR #3439, opened 2025-10-20
fix: add citations to models (#3435)

Samoed:
Reapply #3435 to main. I thought `v2` will be automatically removed and target will be changed to `main`, but because of protection rules (I think) target wasn't changed


KennethEnevoldsen (approved):

KennethEnevoldsen:
Waiting to merge this to see if a release workflow is triggered (if not we need to resolve that)

Issue #3441, opened 2025-10-20
Resolve flaky test for Clustering in `test_predictions`


KennethEnevoldsen:
Example of it being flaky: https://github.com/embeddings-benchmark/mteb/actions/runs/18648452821/job/53160655450

Previous attempt at fixing it: #3400

KennethEnevoldsen:
@Samoed did we end up fixing this? (I don't recall us fixing it)

Samoed:
No, we haven't

KennethEnevoldsen:
Alright, I will close this for now, but if re-appears we can open it

PR #3442, opened 2025-10-20
test: disable flaky test

KennethEnevoldsen:
Added issue to readd #3441


Samoed:
Maybe we can mark it xfail? or add retries on it?

Issue #3446, opened 2025-10-20
Find better ways for package publishing


Samoed:
### Description of the feature

With `v2` release, we have issues with semantic release action. We should change our release ci.

Examples: 
1. I have project where I trigger release ci by creating release (create prerelease -> ci trigger with publishing -> new version on github / pypi) https://github.com/deeppavlov/AutoIntent/blob/dev/.github/workflows/release.yaml (or we can push on tag creation as done in python package guide https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)

2. Fastapi have similar ci that triggered on release creation https://github.com/fastapi/fastapi/blob/master/.github/workflows/publish.yml

KennethEnevoldsen:
Manual release workflow would work fine with me: `uv` also have a decent approach here: https://docs.astral.sh/uv/guides/integration/github/#publishing-to-pypi

Essentially I want something that is easier to debug. Semantic release works well until something breaks

Samoed:
I will update ci then

KennethEnevoldsen:
We could also easily do a workflow which bump based on tag:

```
# extract git commit tag
# if tag is in ...
uv version --bump major/minor/patch
```
But it might be simpler to just manually bump the version.

PR #3448, opened 2025-10-20
ci: New release workflow

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3446

I've added 2 workflow
1. Reimplementation of semantic release that create releases based on prefix of commit.
2. And release of project with uv that will be triggered by new releases. 

But we need to add project to the trusted publisher as said in https://docs.astral.sh/uv/guides/integration/github/#publishing-to-pypi

I'm testing this workflow in my fork https://github.com/Samoed/mteb

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,7 +57,7 @@ jobs:
           bump="none"
           if [[ $low == breaking:* ]]; then
             bump="major"
-          elif [[ $low == feat:* ]]; then
+          elif [[ $low == feat:* ]] || [[ $low == benchmark:* ]]; then
             bump="minor"
           elif [[ $low == fix:* ]] || [[ $low == dataset:* ]] || [[ $low == model:* ]]; then
             bump="patch"
@@ -98,7 +98,6 @@ jobs:
 
           # push changes and tag safely
           git push origin HEAD:main
-          git push origin "$tag"
 
       - name: Create GitHub Release
         env:


.github/workflows/release.yml
```diff
@@ -9,42 +11,122 @@
...
-name: Release
+name: Version & Release on Commit Prefix
 on:
-  workflow_run:
-    workflows: ["test"]
-    types:
-      - completed
+  push:
+    branches:
+      - main
```
KennethEnevoldsen, line 21: Shouldn't we wait for tests to avoid pushing a package without tests passing

.github/workflows/release.yml
```diff
@@ -9,42 +11,122 @@
...
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+          git add pyproject.toml
+          git commit -m "chore: bump version to ${{ steps.version_bump.outputs.new_version }}" || echo "No changes to commit."
+
+          tag="v${{ steps.version_bump.outputs.new_version }}"
```
KennethEnevoldsen: we don't use v* currently

.github/workflows/release.yml
```diff
@@ -9,42 +11,122 @@
...
+          git push --force-with-lease origin HEAD:main
+          git push origin "$tag"
+
+      - name: Create GitHub Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          tag="v${{ steps.version_bump.outputs.new_version }}"
+          echo "Creating release for tag $tag"
+          gh release create "$tag" --generate-notes --repo "${{ github.repository }}"
```
KennethEnevoldsen, line 108: doesn't this also create the tag?

.github/workflows/release.yml
```diff
@@ -9,42 +11,122 @@
...
+
+      - name: Build
+        run: uv build
+
+      # Check that basic features work and we didn't miss to include crucial files
+      - name: Smoke test (wheel)
+        run: uv run --isolated --no-project --with dist/*.whl python -c "import mteb"
+      - name: Smoke test (source distribution)
+        run: uv run --isolated --no-project --with dist/*.tar.gz python -c "import mteb"
```
KennethEnevoldsen, line 117: great to have these

.github/workflows/release.yml
Samoed: Yes, this will create tag and will release a new version based on previous tag

KennethEnevoldsen: but I don't think you need the git tag above

KennethEnevoldsen:
Do I need to set up trusted publishing for this?

Samoed:
Yes

.github/workflows/release.yml
Samoed: Added

KennethEnevoldsen:
Alright, I will do this tomorrow

KennethEnevoldsen:
Actually just did it now (or didn't as it seems like it is already set up - we used it before as well)

Should we get this linted. I would also like to add "benchmark" to major bump.

Samoed:
The issue with lint caused by leaderboard merged before v2. I can push empty commit to trigger

Samoed:
> "benchmark" to major bump

I think it's better to use minor, because major will be for compatability

KennethEnevoldsen:
Ahh yeah minor of course!

Issue #3453, opened 2025-10-20
Don't shorten embedding size in models docs


KennethEnevoldsen:
### Description of the feature

e.g. currently it is shortened as 4.1k

PR #3455, opened 2025-10-20
docs: Don't shorten embedding size

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3453


docs/overview/available_models/image.md
Samoed, line 1: Maybe we should add them to gitignore, because they're generating many changes?

KennethEnevoldsen: yep let us just delete them and add them to the gitignore

PR #3367, opened 2025-10-14
feat: Add mteb nl v2.0.0

nikolay-banar:
- [x] I have outlined why this dataset is filling an existing gap in `mteb`
- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [x] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [x] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] I have considered the size of the dataset and reduced it if it is too big (2048 examples is typically large enough for most tasks)

PR #3464, opened 2025-10-21
feat: Add mteb nl

nikolay-banar:

Ref https://github.com/embeddings-benchmark/mteb/pull/3367

Also, I was checking some default metrics for some datasets. Does it make sense to use accuracy as a main metric for MultiEURLEXMultilabelClassification?

### DIFF
--- a/mteb/tasks/pair_classification/nld/sick_nl_pair_classification.py
+++ b/mteb/tasks/pair_classification/nld/sick_nl_pair_classification.py
@@ -1,5 +1,5 @@
+from mteb.abstasks.pair_classification import AbsTaskPairClassification
 from mteb.abstasks.task_metadata import TaskMetadata
-from mteb.abstasks.text.pair_classification import AbsTaskPairClassification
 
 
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
> Also, I was checking some default metrics for some datasets. Does it make sense to use accuracy as a main metric for MultiEURLEXMultilabelClassification?

I see that the original paper seems to use mean R-precisio (MRP)

$$
\text{R-Precision}_i = \frac{|\text{relevant labels} \cap \text{top-}R_i \text{ predictions}|}{R_i}
$$

Where $R_i$ is the number of relevant labels for instance $i$.

$$
\text{MRP} = \frac{1}{N} \sum_{i=1}^{N} \text{R-Precision}_i
$$

Where N is the total number of instances.

Would this be better? ([paper](https://aclanthology.org/2021.emnlp-main.559v2.pdf))

KennethEnevoldsen:
I will merge this in - but I think we can continue the discussion on metric - feel free to open an issue on it

PR #3472, opened 2025-10-22
fix: verify languages during filtering

Samoed:
We also can check if languages or codes in `ISO_TO_LANGUAGE` and raise an error

### DIFF
--- a/mteb/benchmarks/benchmarks/benchmarks.py
+++ b/mteb/benchmarks/benchmarks/benchmarks.py
@@ -975,8 +975,6 @@ MTEB_INDIC = Benchmark(
                 # Bitext
                 "IN22ConvBitextMining",
                 "IN22GenBitextMining",
-                "IndicGenBenchFloresBitextMining",
-                "LinceMTBitextMining",
                 # clustering
                 "SIB200ClusteringS2S",
                 # classification
@@ -985,7 +983,6 @@ MTEB_INDIC = Benchmark(
                 "HindiDiscourseClassification",
                 "SentimentAnalysisHindi",
                 "MalayalamNewsClassification",
-                "IndicLangClassification",
                 "MTOPIntentClassification",
                 "MultiHateClassification",
                 "TweetSentimentClassification",
@@ -1084,7 +1081,6 @@ MTEB_EU = Benchmark(
             "AmazonCounterfactualClassification",
             "MassiveScenarioClassification",
             "MultiHateClassification",
-            "NordicLangClassification",
             "ScalaClassification",
             "SwissJudgementClassification",
             "TweetSentimentClassification",


mteb/get_tasks.py
```diff
@@ -338,6 +338,16 @@ def get_task(
     Examples:
         >>> get_task("BornholmBitextMining")
     """
+    if languages is not None:
+        for lang in languages:
+            if not (
+                len(lang) == 3  # ISO 639-3 language code
+                or len(lang) == 4  # script code
```
KennethEnevoldsen: I would probably do this in the `LanguageScripts`, which I believe is being called further down.

Also, `lang` can't be a script, only a language-script. 
Language can also be a programming language.

mteb/abstasks/abstask.py
```diff
@@ -491,6 +491,11 @@ def filter_languages(
                 if lang_scripts.contains_languages(langs):
                     subsets_to_keep.append(hf_subset)
 
+        if len(subsets_to_keep) == 0:
+            logger.warning(
+                "No matching languages found for the specified filters. The task will be empty."
+            )
```
KennethEnevoldsen, line 497: does it make any sense to have an "empty" task?

Samoed: I don't think so. I found this because it vllm test they selected `en` for `NFCorpus` which produced empty splits https://github.com/vllm-project/vllm/pull/27329

```python
import mteb 

task = mteb.get_task("NFCorpus", languages=["en"])
task.hf_subsets
# []
```

Samoed:
Moved verification to langcodes, but maybe you have better solution

Samoed:
I found that in `MTEB(indic, v1)` we filter out all subsets for `IndicGenBenchFloresBitextMining`.
```python
import mteb

indic = mteb.get_benchmark("MTEB(Indic, v1)")
indic.tasks[2].hf_subsets
# []
```

This happens, because we're using `exclusive_language_filter=True`, but it has `eng` in all subsets, and they are filtered out.

To fix this, we can just remove this task from mteb indic, because all splits for this task are removed and it not affect leaderboard.

Samoed:
I removed from benchmarks:
- IndicGenBenchFloresBitextMining, LinceMTBitextMining, IndicLangClassification from `MTEB(Indic, v1)`
- NordicLangClassification from `MTEB(Europe, v1)`

mteb/abstasks/abstask.py
KennethEnevoldsen: Hmm. Should we just throw an error?

KennethEnevoldsen: I see you already do

Issue #3478, opened 2025-10-23
Evaluation are broken for datasets that have negatives qrels (e.g trec-covid)


NohTow:
### Describe the bug

Hello,
When trying to run BEIR, I noticed that trec-covid was crashing because `pyarrow.lib.ArrowInvalid: Float value -1 was truncated converting to uint16`.
Tracked back the root issue, which is the conversion of [scores to uint16](https://github.com/embeddings-benchmark/mteb/blob/fcd3b71a09bef1d141686ecf74e46fc285ae6509/mteb/abstasks/retrieval_dataset_loaders.py#L179), which range is 0 - 65535, so does not support the negatives values from trec-covid.

I do not know how you handle this usually, the solution could be to either use int16 (because qrels won't ever be this big) or to clip negative values to 0
cc @tomaarsen who confirmed the issue

### To reproduce

`mteb run -m sentence-transformers/static-retrieval-mrl-en-v1 -t TRECCOVID`

### Are you interested to contribute a fix for this bug?

Yes if needed, the fixes are simple

Samoed:
I didn't think that we have negative qrels

NohTow:
<img width="983" height="337" alt="Image" src="https://github.com/user-attachments/assets/6e06355d-94f2-4198-bd0b-33233255c5f3" />
TIL as well

PR #3479, opened 2025-10-23
fix: qrels selection negative scores

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3478


tomaarsen:
int16 should also cover it right? It doesn't matter much either way, as long as -1 is supported.
P.s. I tested `int16` myself with the CLI using https://huggingface.co/sentence-transformers/static-retrieval-mrl-en-v1 and TRECCOVID, worked well.

- Tom Aarsen

Samoed:
Probably yes. Overall, I think this should be float, but `pytrec_eval` changed scores from float to int in recent release. I left as int32 for compatibility, but since qrels are transformed from datasets to python dict I don't think that this would produce difference in memory

Issue #3406, opened 2025-10-17
Add similarity search backend to Retrieval tasks


Samoed:
### Description of the feature

As https://github.com/embeddings-benchmark/mteb/issues/3192 added possible different backends to the cache wrapper, we can add similar backends to retrieval tasks.

It can have interface like
```python
class SearchBackend:
    def add_document(
        self,
        embeddings: Array,
        idxs: list[str],
    ) -> None:
        """Add documents to the search backend.
        
        Args:
            embeddings: Embeddings of the documents to add.
            idxs: IDs of the documents to add.
        """
        
    def search(
        self,
        embeddings: Array,
    ) -> list[tuple[str, float]]:
        """Search the backend for the given embeddings.

        Args:
            embeddings: Embeddings to search for.

        Returns:
            List of tuples containing document IDs and their relevance scores.
        """

    def save(self, **kwargs):
        ...

    def load(self, **kwargs):
        ...
```

And then this can be used with `SearchEncoderWrapper` as:
```python
model = SearchEncoderWrapper(
    mteb.get_model(...),
    backend=SearchBackend(),
)
```

This method wouldn't require changes for other search models (pylate), but propose change for the encoders

CC @orionw

orionw:
+1, this would be quite nice and mirror what is going on in `pylate` and in `transformers` with their attention variants.

At some point it could wrap the many different ways people use backends, e.g. variants of `faiss` etc.

Issue #3401, opened 2025-10-17
Add a changelog


KennethEnevoldsen:
After v2.0.0, it would be great to maintain a changelog for more notable changes (it is possible to auto-generate it with current CI it is just currently disabled)

Samoed:
For ref changelog for semantic release https://python-semantic-release.readthedocs.io/en/stable/misc/psr_changelog.html

KennethEnevoldsen:
I would probably just manually write it up if we are not using PSR anyway

PR #3492, opened 2025-10-25
feat: add search encoder backend

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3406 

I've implemented SearchEncoder protocol and now can be selected faiss or direct search (as previous). 

https://github.com/embeddings-benchmark/mteb/blob/c8b2bd3a5d1af6a8639da567d6dfd30cbbd25bdb/mteb/models/search_encoder_index/search_backend_protocol.py#L7-L29

I've saved "batched" approach for retrieval to store less memory during evaluation. Backend can be changed by 

```python
import mteb
from mteb.models.search_encoder_index import (
    DefaultEncoderSearchBackend,
    FaissEncoderSearchBackend,
)
from mteb.models import SearchEncoderWrapper


python_backend = SearchEncoderWrapper(
    model, index_backend=DefaultEncoderSearchBackend()
)
faiss_backend = SearchEncoderWrapper(
    model, index_backend=FaissEncoderSearchBackend(model)
)
```

I've tested on Scifact using `potion-2M` and got 2s evaluation for default search and 3s for `FAISS`.

<details>
<summary>Script to test</summary>

```python
import mteb
from mteb.cache import ResultCache
from mteb.models import SearchEncoderWrapper
from mteb.models.search_encoder_index import (
    DefaultEncoderSearchBackend,
    FaissEncoderSearchBackend,
)

model = mteb.get_model("minishlab/potion-base-2M")

python_backend = SearchEncoderWrapper(
    model, index_backend=DefaultEncoderSearchBackend()
)
faiss_backend = SearchEncoderWrapper(
    model, index_backend=FaissEncoderSearchBackend(model)
)

task = mteb.get_task("SciFact")

python_cache = ResultCache("python_backend_cache")
faiss_cache = ResultCache("faiss_backend_cache")

# warmup
mteb.evaluate(
    model,
    task,
    cache=None,
)

mteb.evaluate(
    python_backend,
    task,
    cache=python_cache,
)

mteb.evaluate(
    faiss_backend,
    task,
    cache=faiss_cache,
)
```

</details>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,9 +35,8 @@ pr:
 	make lint
 	make test
 
-build-docs:
+build-docs: build-docs-overview
 	@echo "--- 📚 Building documentation ---"
-	make build-docs-overview
 	python -m mkdocs build
 
 
--- a/docs/api/model.md
+++ b/docs/api/model.md
@@ -33,3 +33,7 @@ length, valid frameworks, license, and degree of openness.
 :::mteb.models.CrossEncoderProtocol
 
 :::mteb.models.MTEBModels
+
+:::mteb.models.IndexEncoderSearchProtocol
+
+:::mteb.models.CacheBackendProtocol
--- a/mteb/__init__.py
+++ b/mteb/__init__.py
@@ -9,8 +9,10 @@ from mteb.filter_tasks import filter_tasks
 from mteb.get_tasks import get_task, get_tasks
 from mteb.load_results import load_results
 from mteb.models import (
+    CacheBackendProtocol,
     CrossEncoderProtocol,
     EncoderProtocol,
+    IndexEncoderSearchProtocol,
     SearchProtocol,
     SentenceTransformerEncoderWrapper,
 )
@@ -27,8 +29,10 @@ __all__ = [
     "AbsTask",
     "Benchmark",
     "BenchmarkResults",
+    "CacheBackendProtocol",
     "CrossEncoderProtocol",
     "EncoderProtocol",
+    "IndexEncoderSearchProtocol",
     "SearchProtocol",
     "SentenceTransformerEncoderWrapper",
     "TaskMetadata",
--- a/mteb/models/__init__.py
+++ b/mteb/models/__init__.py
@@ -1,4 +1,4 @@
-from .cache_wrappers import CachedEmbeddingWrapper
+from .cache_wrappers import CacheBackendProtocol, CachedEmbeddingWrapper
 from .model_meta import ModelMeta
 from .models_protocols import (
     CrossEncoderProtocol,
@@ -6,6 +6,7 @@ from .models_protocols import (
     MTEBModels,
     SearchProtocol,
 )
+from .search_encoder_index.search_backend_protocol import IndexEncoderSearchProtocol
 from .search_wrappers import SearchCrossEncoderWrapper, SearchEncoderWrapper
 from .sentence_transformer_wrapper import (
     CrossEncoderWrapper,
@@ -14,10 +15,12 @@ from .sentence_transformer_wrapper import (
 )
 
 __all__ = [
+    "CacheBackendProtocol",
     "CachedEmbeddingWrapper",
     "CrossEncoderProtocol",
     "CrossEncoderWrapper",
     "EncoderProtocol",
+    "IndexEncoderSearchProtocol",
     "MTEBModels",
     "ModelMeta",
     "SearchCrossEncoderWrapper",
--- a/mteb/models/cache_wrappers/__init__.py
+++ b/mteb/models/cache_wrappers/__init__.py
@@ -1,3 +1,4 @@
+from .cache_backend_protocol import CacheBackendProtocol
 from .cache_wrapper import CachedEmbeddingWrapper
 
-__all__ = ["CachedEmbeddingWrapper"]
+__all__ = ["CacheBackendProtocol", "CachedEmbeddingWrapper"]


orionw (approved):
Looks great overall!

I think it doesn't take advantage of `faiss`'s built in scoring functionality but by not doing that we can have more control (so that can be ignored). If we wanted to use faiss we could do something like

```
# Batch reconstruct candidate embeddings
candidate_embs = np.vstack([
    self.index.reconstruct(idx) for idx in candidate_indices
])

# Create temporary index to let FAISS handle scoring
temp_index = self.index_type(d)
temp_index.add(candidate_embs)

# Search returns scores and indices in one call
scores, local_indices = temp_index.search(
    query_emb.reshape(1, -1).astype(np.float32),
    min(top_k, len(candidate_indices))
)
```

But I think it just does dot product. So it looks great as is, but just mentioning this in case that's helpful.

Samoed:
Yes, I think that's better. I've added support of cosine and dot product similarity support and scores are nearly the same (same for `1e-6`).

mteb/models/search_encoder_index/search_backend_protocol.py
```diff
@@ -0,0 +1,50 @@
+from collections.abc import Callable
+from typing import Protocol
+
+from mteb.types import Array, TopRankedDocumentsType
+
+
+class IndexEncoderSearchProtocol(Protocol):
```
KennethEnevoldsen, line 7:
```suggestion
class EncoderSearchProtocol(Protocol):
```

mteb/models/search_encoder_index/default_backend_search.py
```diff
@@ -0,0 +1,99 @@
...
+from collections.abc import Callable
+
+import torch
+
+from mteb.types import Array, TopRankedDocumentsType
+
+logger = logging.getLogger(__name__)
+
+
+class DefaultEncoderSearchBackend:
```
KennethEnevoldsen:
```suggestion
class StreamingSearchBackend:
```
Should we call it backend or "index"

mteb/models/search_encoder_index/faiss_search_backend.py
```diff
@@ -0,0 +1,157 @@
...
+
+from mteb._requires_package import requires_package
+from mteb.models.model_meta import ScoringFunction
+from mteb.models.models_protocols import EncoderProtocol
+from mteb.types import Array, TopRankedDocumentsType
+
+logger = logging.getLogger(__name__)
+
+
+class FaissEncoderSearchBackend:
```
KennethEnevoldsen:
```suggestion
class FaissSearchBackend:
```

tests/test_abstasks/test_predictions.py
```diff
@@ -125,3 +134,68 @@ def test_predictions(tmp_path: Path, task, expected):
 
     predictions = full_predictions["default"]["test"]
     assert predictions == expected
+
+
+@pytest.mark.parametrize(
```
KennethEnevoldsen: should this be placed here? I would say that it neither under abstasks nor test_predictions

Samoed:
> We also need to add documentation

Yes, wanted to add after your check on pr

mteb/models/search_encoder_index/search_backend_protocol.py
Samoed: I think we can specify that this is only for index and only for encoder, because this can be confused that `SentenceTransformerEncoderWrapper` will implement it (probably)

mteb/models/search_encoder_index/default_backend_search.py
Samoed: I think index would be better to use

mteb/models/search_encoder_index/faiss_search_backend.py
Samoed: Renamed to `FaissSearchIndex`

mteb/models/search_encoder_index/default_backend_search.py
Samoed: Renamed to `StreamingSearchIndex`

tests/test_abstasks/test_predictions.py
Samoed: Created new folder test_search_index

KennethEnevoldsen: wouldn't you put it test_models (to match the structure of the repo?)

Samoed: Yes, will move it here then

Samoed:
I've run this script and both evaluation method took same time, so I'm unsure a bit what to add in advantages of FAISS, except of dumping index, but we're clearing it after evaluation.

| task | Stream | FAISS |
|-|-|-|
| SWEbenchVerifiedRR | 536 | 541 |
| ClimateFEVERHardNegatives | 9 | 12 | 

```python
import logging

import mteb
from mteb.cache import ResultCache
from mteb.models import SearchEncoderWrapper
from mteb.models.search_encoder_index import StreamingSearchIndex, FaissSearchIndex

model = SearchEncoderWrapper(mteb.get_model("minishlab/potion-base-2M"))
tasks = mteb.get_tasks(
    tasks=[
        "ClimateFEVERHardNegatives",
        "SWEbenchVerifiedRR",
    ],
)

cache = ResultCache("stream")

mteb.evaluate(
    model,
    tasks,
    cache=cache,
)

### FAISS
index_backend = FaissSearchIndex(model)
model = SearchEncoderWrapper(
    mteb.get_model("minishlab/potion-base-2M"),
    index_backend=index_backend
)
cache = ResultCache("FAISS")

mteb.evaluate(
    model,
    tasks,
    cache=cache,
)
```

orionw:
I think faiss is not ideal for smaller reranking cases (~100-1000 docs to search for). We should see dramatic gains for retrieval though, with a large enough corpus. For `ClimateFEVERHardNegatives` it could just be initialization differences. Maybe if you try MS MARCO for retrieval?

I asked Claude what it thinks we should do for reranking and it suggested we retrieve the vectors from `faiss` for reranking but just use standard numpy afterwards. We could do this, but if it's roughly the same to use faiss then we might as well keep what we have for that.

If large scale retrieval is much faster I think that's the main benefit

Samoed:
I tried running it on `MSMARCO`, and both backends showed similar times on sub-batches. If we remove the search over each sub-corpus batch, FAISS would probably show a speedup, but I’m not sure how to do that while still supporting the "streaming" backend.

mteb/models/search_encoder_index/search_backend_protocol.py
```diff
@@ -0,0 +1,50 @@
...
+    """Protocol for search backends used in encoder-based retrieval."""
+
+    def add_document(
```
KennethEnevoldsen:
```suggestion
    def add_documents(
```

mteb/models/search_wrappers.py
```diff
@@ -173,23 +181,13 @@ def _full_corpus_search(
                 prompt_type=PromptType.document,
                 **encode_kwargs,
             )
-
+            self.index_backend.add_document(sub_corpus_embeddings, sub_corpus_ids)
```
KennethEnevoldsen: shouldn't things we added in the index step? (I suspect it it here because of the streaming backend)

Samoed: Yes, but I don't know how to make it compatible with streaming without creating new functions. If we want to use the index step, we need duplicate functions with similar logic. Maybe we can create different wrapper to use without hacks to use index directly

KennethEnevoldsen: Hmm can you outline the current approach in the streaming just to clarify why it is the way it is now (Just to make sure that we agree on the problem).

Samoed: Current approach with "streaming" don't store corpus embeddings fully for retrieval. It processes corpus in "chunks" of 50_000 documents and then finds similarities between documents and queries. Because of that we're adding `sub_corpus_embeddings` here (chunk embeddings). If we move encoding of documents in index, then we need to store all embeddings of documents, but this will cause problems with large datasets like miracl or msmarco

KennethEnevoldsen: Hmm, so I see two possible issues. 

1) Either the abstraction is faulty, and we should try to come up with a different protocol. E.g., one that takes in the encoder.
2) or the streaming backend doesn't fit into the current abstraction, and we should remove the streaming from it. E.g. by making `SearchEncoderWrapper` (takes a search engine) and `StreamingSearchEncoderWrapper` (we probably have to word with the naming to avoid breaking backward compat).

Samoed: > Either the abstraction is faulty, and we should try to come up with a different protocol. E.g., one that takes in the encoder.

I think if we do that then it will be almost like the search protocol 

> the streaming backend doesn't fit into the current abstraction, and we should remove the streaming from it

I think yes, but I don't know how to do this without code duplication. We can make if no index backend is passed, then a search approach will be used, otherwise index

KennethEnevoldsen: > I think if we do that then it will be almost like the search protocol

Yep, that was also my worry - good to see that we agree

> I think yes, but I don't know how to do this without code duplication. We can make if no index backend is passed, then a search approach will be used, otherwise index

Yeah, that might actually be a pretty decent approach

Samoed: Made backend optional

mteb/models/search_encoder_index/search_indexes/streaming_search_index.py
KennethEnevoldsen, line 1: doesn't that mean that we should delete this?  (since it is the default)

Samoed: Removed

docs/advanced_usage/retrieval_backend.md
```diff
@@ -0,0 +1,23 @@
+# Retrieval Search backend
+
+!!! note "Available since 2.3.0"
+    This feature was introduced in version **2.3.0**.
+
+For some large dataset search can take a lot of time. To reduce this you can use `FaissSearchIndex`. To work with it install `pip install mteb[faiss]`.
```
KennethEnevoldsen: Wasn't the main concern memory?

docs/advanced_usage/retrieval_backend.md
KennethEnevoldsen, line 1: shouldn't we also add documentation on the search backend?

Maybe we should also add something here so people can discover what has been added:

Screenshot 2025-11-25 at 16 23 28

A kind of user-friendly changelog

Samoed: It will be shown in `advanced usage`

KennethEnevoldsen: Yeah, but people will not know what has happened since 2.0.0

I would probably change `New in v2.0` to 

```
- What is new
  - v2.3
  - v2.2
  - v2.1
  - v2.0
```

Samoed: I think this is more about changelog https://github.com/embeddings-benchmark/mteb/issues/3401

docs/advanced_usage/retrieval_backend.md
Samoed: Also added memory

docs/advanced_usage/retrieval_backend.md
KennethEnevoldsen: Fair we still need the API docs though

Makefile
```diff
@@ -35,9 +35,8 @@ pr:
 	make lint
 	make test
 
-build-docs:
+build-docs: build-docs-overview
```
KennethEnevoldsen, line 38: oO does this work?

KennethEnevoldsen:
I think this is good to merge

Makefile
Samoed: Yes, everything after `:` will be triggered before running a function
```makefile
targets: prerequisites
	command
```

Issue #3423, opened 2025-10-18
Add model: facebook/SONAR


Samoed:
### Model link on Hugging Face

https://huggingface.co/facebook/SONAR

yaswanth169:
Hi @Samoed @AdnanElAssadi56,

I'd like to claim this one for MOEB : targeting the original SONAR (text_sonar_basic_encoder + speech encoders), not the newer OmniSONAR variant. Fits the "audio embedding models" priority.

Quick specs: 200 languages text / 37 languages speech, fixed 1024-dim output, clean Python API (TextToEmbeddingModelPipeline / SpeechToEmbeddingModelPipeline). Code is MIT, model weights are CC-BY-NC-4.0 — text encoder ~750M params, speech encoder ~1B params (MMS-based), so no GPU strictly required.

Plan: custom AbsEncoder wrapper handling text/speech modality routing to the right pipeline. Will validate against a known audio retrieval task and try to reproduce a reported number where possible before submitting.

Yashwanth

PR #3497, opened 2025-10-26
Refactor tasks and models to new interface

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3413

Now all models and tasks using `ecode`.

Audio input looks like
```python
class AudioInputItem(TypedDict):
    array: np.ndarray
    sampling_rate: int | None

class AudioInput(TypedDict):
    audio: list[AudioInputItem]
```

Refactored tasks to use `model.encode`:
- [X] reranking
- [X] classification
- [X] clustering
- [X] pair classification
- [X] multilabel classification
- [X] zero shot classification
- [X] retrieval

Updated models `model.encode`:
- [x] ast
- [x] clap
- [x] cnn14
- [x] data2vec
- [x] encodec
- [x] hubert
- [x] mctct
- [x] mms
- [x] msclap
- [x] muq
- [x] qwen2
- [x] seamlessm4t
- [x] sewd
- [x] speecht5
- [x] unispeech
- [x] wav2clip
- [x] wav2vec2
- [x] wavlm
- [x] whisper
- [x] yamnet
- [x] vggish

### DIFF
--- a/mteb/models/model_implementations/vggish_models.py
+++ b/mteb/models/model_implementations/vggish_models.py
@@ -92,9 +92,6 @@ def vggish_loader(*args, **kwargs):
                 self.device
             )
 
-            if input_tensor.numel() == 0:
-                return None
-
             # Handle different tensor dimensions
             if input_tensor.dim() == 4 and input_tensor.shape[1] == 3:
                 # [batch, 3, height, width] -> [batch, 1, height, width]
@@ -145,18 +142,6 @@ def vggish_loader(*args, **kwargs):
 
                     with torch.no_grad():
                         input_tensor = self._prepare_input_tensor(audio)
-
-                        if input_tensor is None:
-                            # Create zero embedding for empty tensor
-                            logger.debug("Creating zero embedding for empty tensor")
-                            zero_embedding = torch.zeros(
-                                self.embed_dim, device=self.device
-                            )
-                            batch_embeddings.append(
-                                zero_embedding.cpu().detach().numpy()
-                            )
-                            continue
-
                         embedding = self.model(input_tensor)
 
                         # Use mean pooling if needed
--- a/mteb/models/model_implementations/yamnet_models.py
+++ b/mteb/models/model_implementations/yamnet_models.py
@@ -147,18 +147,6 @@ def yamnet_loader(*args, **kwargs):
 
                     with torch.no_grad():
                         input_tensor = self._prepare_input_tensor(audio)
-
-                        if input_tensor is None:
-                            # Create zero embedding for empty tensor
-                            logger.debug("Creating zero embedding for empty tensor")
-                            zero_embedding = torch.zeros(
-                                self.embed_dim, device=self.device
-                            )
-                            batch_embeddings.append(
-                                zero_embedding.cpu().detach().numpy()
-                            )
-                            continue
-
                         # discard logits
                         embedding, _ = self.model(input_tensor)
 


mteb/evaluation/evaluators/audio/clustering_evaluator.py
Samoed, line 1: Wasn't used

mteb/models/model_implementations/ast_model.py
Samoed, line 1: @AdnanElAssadi56 Can you review refactored model implementation?

AdnanElAssadi: Overall looks good, but removing those zero/empty audio checks has caused some errors when evaluating on some datasets before. Not sure if current data-loader will fix the problem.

Samoed: I think this is dataset problem rather than model problem and this should be fixed in `task.load_data`, because taks shouldn't have empty inputs

AdnanElAssadi:
I’ve refactored 15 out of 21 models and tested all of them on "BeijingOpera" as a sanity check. (Except for Qwen2, which runs out of memory - likely due to its large size and CNN14 - still haven't done it)

We’ll probably need to handle the problematic datasets separately. Some models fail on empty inputs (as mentioned earlier), and others require a minimum input window length for audio, which can cause crashes. I was able to reproduce these issues with "MridinghamStroke" and "MridinghamTonic".

Samoed:
> others require a minimum input window length for audio, which can cause crashes. I was able to reproduce these issues with "MridinghamStroke" and "MridinghamTonic".

Which models reqired this?

mteb/models/model_implementations/encodec_model.py
```diff
@@ -31,166 +35,81 @@ def __init__(
...
+                if sr is None:
+                    warnings.warn(
+                        f"No sampling_rate provided for an audio sample. "
+                        f"Assuming {self.sampling_rate} Hz (model default)."
+                    )
+                    sr = self.sampling_rate
+
+                # Convert to mono if needed
+                if array.dim() > 1 and array.shape[0] > 1:
+                    array = torch.mean(array, dim=0, keepdim=True)
```
Samoed, line 70: Do we support mono/stereo separation of datasets? Should we add same conversion for other models, or we process only mono?

mteb/models/model_implementations/mms_models.py
```diff
@@ -1,16 +1,18 @@
...
+from mteb import TaskMetadata
+from mteb._requires_package import requires_audio_dependencies
 from mteb.models import ModelMeta
 from mteb.models.abs_encoder import AbsEncoder
-from mteb.types import Array, PromptType
+from mteb.types import Array, BatchedInput, PromptType
+from mteb.types._encoder_io import AudioInput
 
 
 class MMSWrapper(AbsEncoder):
```
Samoed, line 17: We're always load only english adapter for this model which is probably incorrect

mteb/models/model_implementations/mms_models.py
```diff
@@ -48,114 +51,52 @@ def __init__(
...
+                        f"Assuming {self.sampling_rate} Hz (model default)."
+                    )
+                    sr = self.sampling_rate
+
+                if sr != self.sampling_rate:
+                    resampler = torchaudio.transforms.Resample(
+                        orig_freq=sr, new_freq=self.sampling_rate
+                    )
+                    array = resampler(array)
+                audio_arrays.append(array.numpy())
```
Samoed, line 83: I think we should move this to utils functions, because we're repeating this for all models

mteb/models/model_implementations/audio_qwen2_models.py
Samoed, line 1: Should we add support for text for qwen audio?

AdnanElAssadi:
@Samoed Have you tested on a dataset?

Samoed:
Not yet, will try a bit later

mteb/models/model_implementations/speecht5_models.py
Samoed, line 1: @AdnanElAssadi56 Should we split `SpeechT5Wrapper` into tts and stt? This is strange that in init we have asr and tts models and they're downloading different models

mteb/models/model_implementations/speecht5_models.py
```diff
@@ -301,10 +249,7 @@ def encode(
 
 # Voice Conversion model - Optimized for Speech-to-Speech conversion tasks
 speecht5_multimodal = ModelMeta(
-    loader=partial(
-        SpeechT5Wrapper,
-        model_name="microsoft/speecht5_multimodal",
-    ),
+    loader=SpeechT5Wrapper,
     name="microsoft/speecht5_multimodal",
```
Samoed, line 330: BTW I can't find it. Does it exist?

mteb/models/model_implementations/qwen2_models.py
Samoed, line 1: Renamed to audio qwen2

mteb/models/model_implementations/speecht5_models.py
AdnanElAssadi: If you go back to corresponding PR, I think this was discussed. SpeechT5 has these two models separate, and I did this workaround to be able have it do A2T and T2A tasks. We didn't have many Audio & Text models, so I decided to use it this way.

Samoed: Yes, this was only one comment, but I don't think that this is good solution https://github.com/embeddings-benchmark/mteb/pull/2901

Samoed:
I've run models
 
Working models:
1. "laion/clap-htsat-fused",
2. "facebook/mms-1b-all",
3. "facebook/hubert-base-ls960",
4. "asapp/sew-d-tiny-100k-ft-ls100h",
5. "facebook/encodec_24khz",
6. data2vec-audio-base-960h",
7. unispeech-sat-base-100h-libri-ft",
8. wav2vec2-xls-r-300m",
9. wavlm-base",
10. whisper-tiny",
11. vggish",
12. yamnet",
13. cnn14-esc50",
14. MuQ-MuLan-large",

Not working:
1. "speechbrain/m-ctc-t-large. I have `NAN` as output sometimes and warning during eval
```
/usr/local/lib/python3.12/dist-packages/transformers/models/deprecated/mctct/feature_extraction_mctct.py:145: RuntimeWarning: invalid value encountered in divide
  x = np.divide(x, std)
```
It seems that this model deprecated and I'm not sure if we want to support it
2. lyrebird/wav2clip
Problem with library
```
OSError: lyrebird/wav2clip is not a local folder and is not a valid model identifier listed on 'https://huggingface.co/models'
If this is a private repository, make sure to pass a token having permission to this repo either by logging in with `hf auth login` or by passing `token=<your_token>`
```
3. microsoft/msclap-2023
Problem with librosa 
```
TypeError: list indices must be integers or slices, not tuple inside librosa
```

I didn't run:
1. Qwen (too big)
2. Speecht5, need to figure out what to do with `tts/stt`

mteb/models/model_implementations/encodec_model.py
Samoed: @AdnanElAssadi56 Do we need to convert in mono for all models?

mteb/models/model_implementations/speecht5_models.py
AdnanElAssadi: How would you suggest doing it?

Samoed: I think we should have separate stt and tts. We can have separete wrapper for their combination, as in file `microsoft/speecht5_multimodal` which don't exists on hf

Samoed: Refactored in https://github.com/embeddings-benchmark/mteb/pull/3497/commits/3a1ef1c0726df5a3acf5620df3c77eb7a15295ef

AdnanElAssadi:
@Samoed I say we let @gowitheflow-1998 resume evaluation and confirm for us before we merge, at least for a small subset of the datasets.

AdnanElAssadi:
> 2. lyrebird/wav2clip Problem with library
> 
> ```
> OSError: lyrebird/wav2clip is not a local folder and is not a valid model identifier listed on 'https://huggingface.co/models'
> If this is a private repository, make sure to pass a token having permission to this repo either by logging in with `hf auth
> ```

I didn't face this problem when integrating the model a few months ago. Could the refactoring be the problem? I'll try to test it from my end.

mteb/models/model_implementations/audio_qwen2_models.py
AdnanElAssadi: Are you sure there is an option for it? Last time I checked it was hard to integrate. Seamless could also have text option.

Samoed:
> > 2. lyrebird/wav2clip Problem with library
> > 
> > ```
> > OSError: lyrebird/wav2clip is not a local folder and is not a valid model identifier listed on 'https://huggingface.co/models'
> > If this is a private repository, make sure to pass a token having permission to this repo either by logging in with `hf auth
> > ```
> 
> I didn't face this problem when integrating the model a few months ago. Could the refactoring be the problem? I'll try to test it from my end.


No, this is inside from library during init. I'll try to fix it but I don't think that I can

mteb/models/model_implementations/wav2clip_model.py
```diff
@@ -31,194 +35,124 @@ def __init__(
         self.max_audio_length_s = max_audio_length_s
 
         # text side (CLIP)
-        self.clip = CLIPModel.from_pretrained("openai/clip-vit-base-patch32").to(device)
+        self.clip = CLIPModel.from_pretrained(model_name, revision=revision).to(device)
+        self.clip.eval()
         self.clip_processor = CLIPProcessor.from_pretrained(
-            "openai/clip-vit-base-patch32"
+            model_name, revision=revision
         )
```
Samoed, line 42: Why do we use clip here? This is method to make wav2clip multimodal with completely different model?

mteb/models/model_implementations/vggish_models.py
```diff
@@ -153,45 +109,52 @@ def _prepare_input_tensor(self, audio_data):
...
                         if input_tensor is None:
                             # Create zero embedding for empty tensor
                             logger.debug("Creating zero embedding for empty tensor")
                             zero_embedding = torch.zeros(
                                 self.embed_dim, device=self.device
                             )
-                            batch_embeddings.append(zero_embedding.cpu().numpy())
+                            batch_embeddings.append(
+                                zero_embedding.cpu().detach().numpy()
+                            )
```
KennethEnevoldsen: I don't think we need the numpy conversion here

mteb/models/model_implementations/vggish_models.py
```diff
@@ -153,45 +109,52 @@ def _prepare_input_tensor(self, audio_data):
...
+                    sr = a.get("sampling_rate", None)
+                    if sr is None:
+                        warnings.warn(
+                            f"No sampling_rate provided for an audio sample. "
+                            f"Assuming {self.sampling_rate} Hz (model default)."
+                        )
+                        sr = self.sampling_rate
+
+                    # Resample if needed
+                    audio = self._resample_audio(array.float(), sr)
```
KennethEnevoldsen, line 139: should this be done in the collate_fn to avoid stalling the GPU

Samoed:
@KennethEnevoldsen What do you think about wav2clip and speecht5? I don't think we want to merge multiple models or clip in wav2clip

mteb/models/model_implementations/wav2clip_model.py
KennethEnevoldsen: But isn't this what they do in the paper?

https://arxiv.org/pdf/2110.11499

Samoed: Yes, but clip here used only in `get_text_embeddings`. Clip for wav2clip added automatically in their library

KennethEnevoldsen:
> speecht5

I could see it both ways. Splitting them, however, simplified inference, so I would go for that. 

It is a bit similar to the SONAR models, which can either be seen as one multilingual model or one for all modalities #3423

Samoed:
I've separated them in different wrappers and added another wrapper that uses both models in https://github.com/embeddings-benchmark/mteb/pull/3497/commits/3a1ef1c0726df5a3acf5620df3c77eb7a15295ef, but still combining two models seems a bit strange for me

mteb/models/model_implementations/vggish_models.py
Samoed: Great idea! But we need to pass to collator model's `sampling_rate` somehow then. We can standardize this to move this code in collator, but I don't have ideas currently

mteb/models/model_implementations/vggish_models.py
Samoed: Overall I don't think that we should handle empty input audios here. I will remove it then

KennethEnevoldsen:
Splitting them is more than fine with me

Samoed:
But do we want to keep "multimodal" approach

KennethEnevoldsen:
Not sure what you mean?

Samoed:
For speech t5 we have multimodal version combines asr and tts variants together to support audio and speech modalities together https://github.com/embeddings-benchmark/mteb/blob/3a1ef1c0726df5a3acf5620df3c77eb7a15295ef/mteb/models/model_implementations/speecht5_models.py#L328

Same for wav2clip that's adding clip just to support text modality https://github.com/embeddings-benchmark/mteb/blob/9c52fe41b0d8fa503e6074e85f535684e69d2bfc/mteb/models/model_implementations/wav2clip_model.py#L38

Overall what left for this pr?

AdnanElAssadi:
@Samoed Maybe we can go with this approach for the initial release. We only have around five audio+text models, if I remember correctly. Once we integrate more later on, we can refactor or remove anything that doesn’t fit well.

@gowitheflow-1998 Were you able to run an evaluation pass to confirm that there aren’t any major issues?

mteb/models/model_implementations/vggish_models.py
KennethEnevoldsen: Let us work on it later and leave it for now

mteb/models/model_implementations/wav2clip_model.py
KennethEnevoldsen: Hmm but don't they align the audio model into the clip space? (so you could do text to i audio retrieval)

KennethEnevoldsen:
@Samoed sorry for the late response here - I think we can merge this

mteb/models/model_implementations/wav2clip_model.py
Samoed: Yes, but this is not clip that they're using in their library. They're downloading it from https://github.com/descriptinc/lyrebird-wav2clip/blob/1864b3924be5a785e2d49d975b8a26ff93f62951/wav2clip/__init__.py#L7
and it seems that they don't have a method to extract text embedding https://github.com/descriptinc/lyrebird-wav2clip/blob/1864b3924be5a785e2d49d975b8a26ff93f62951/wav2clip/model/encoder.py#L30

KennethEnevoldsen: Maybe move this to an issue an merge? (@AdnanElAssadi56 seems like you made the first version of this?)

Samoed: Added an issue https://github.com/embeddings-benchmark/mteb/issues/3545

Issue #3480, opened 2025-10-23
Release action not working


Samoed:
### Describe the bug

It tries to update pyproject, but it can't commit because `main` is protected branch. @KennethEnevoldsen 

### To reproduce

https://github.com/embeddings-benchmark/mteb/actions/runs/18748369318/job/53481167054

```
[main c4ddc7c0] chore: bump version to 2.1.1
 1 file changed, 1 insertion(+), 1 deletion(-)
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: 
remote: - Changes must be made through a pull request.        
remote: 
remote: - 3 of 3 required status checks are expected.        
To https://github.com/embeddings-benchmark/mteb
 ! [remote rejected]   HEAD -> main (protected branch hook declined)
error: failed to push some refs to 'https://github.com/embeddings-benchmark/mteb'
```

### Are you interested to contribute a fix for this bug?

Yes

KennethEnevoldsen:
I believe it should now have permissions (it is my token, and I have set up the permission so that I can bypass the PR before merge now). Ideally, I would want to just allow the bot, but couldn't figure out a way to do that without creating a separate account.

Samoed:
There is discussion, but I can't find how to do this https://github.com/orgs/community/discussions/13836

KennethEnevoldsen:
Yep same

Samoed:
Still can't pass https://github.com/embeddings-benchmark/mteb/actions/runs/18837943041/job/53743199941#step:8:16. I think we can roll back to semantic release for now

PR #3502, opened 2025-10-27
ci: Rollback to semantic release

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/issues/3480


Samoed:
We can try triggering the new CI to see if it pushes the changes, but I have no idea why the previous workflow passed with the same token.

The last run of the new release action failed because of a multiline comment https://github.com/embeddings-benchmark/mteb/actions/runs/18838824823. I can fix that, but I’m not sure which CI release action we want to stick with.

KennethEnevoldsen:
I prefer the current one, but for it doesn't seem to work so then I would rather roll back and test it on an external github

PR #3516, opened 2025-10-31
fix: MTEB-NL prompts

nikolay-banar:
Following the discussion https://github.com/embeddings-benchmark/mteb/discussions/3339#discussioncomment-14814029

### DIFF
--- a/mteb/tasks/retrieval/nld/argu_ana_nl_retrieval.py
+++ b/mteb/tasks/retrieval/nld/argu_ana_nl_retrieval.py
@@ -1,9 +1,6 @@
 from mteb.abstasks.retrieval import AbsTaskRetrieval
 from mteb.abstasks.task_metadata import TaskMetadata
 
-#
-# This
-# version
 _argu_ana_nl_metadata = dict(
     reference="https://huggingface.co/datasets/clips/beir-nl-arguana",
     dataset={


mteb/tasks/clustering/nld/open_tender_clustering_p2p.py
```diff
@@ -38,6 +38,9 @@ class OpenTenderClusteringP2P(AbsTaskClustering):
   year = {2025},
 }
 """,
+        prompt={
```
KennethEnevoldsen, line 41: Are we sure we want an English prompt for a Dutch dataset?

Samoed: I think all our prompts in taskmetada are in English

KennethEnevoldsen: But I don't think there is any need to force that. For SEB v2 I will probably rework them to their respective languages)

nikolay-banar: Good question. However, it is usually model-dependent. As I remember, e5-models are trained on English instructions. Also, I guess it should not be an issue for multilingual models because they are usually trained on a large portion of English instructions.

KennethEnevoldsen: @nikolay-banar yeah the model providers can always replace the prompt if e.g. they only support English prompts. Ultimately I think it is decision for you to make as the developer of the benchmark. We can def. keep them English if you prefer

(last time I check e5 actually performed slightly better on SEB if you used a Danish prompt)

nikolay-banar: @KennethEnevoldsen I will translate the prompts then.

nikolay-banar: @KennethEnevoldsen What about multilingual datasets? Would it make sense to create Dutch versions with the corresponding prompts?

KennethEnevoldsen: > @KennethEnevoldsen What about multilingual datasets? Would it make sense to create Dutch versions with the corresponding prompts?

For multilingual dataset I would probably keep the prompts English to avoid having two versions on the task (we could probably have different prompt on different subsets - feel free to create an issue on this)

mteb/tasks/retrieval/nld/argu_ana_nl_retrieval.py
```diff
@@ -38,5 +34,27 @@ class ArguAnaNL(AbsTaskRetrieval):
...
+        name="ArguAna-NL",
         adapted_from=["ArguAna"],
+        **_argu_ana_nl_metadata,
+    )
+
+
+class ArguAnaNLv2(AbsTaskRetrieval):
+    ignore_identical_ids = True
+
+    metadata = TaskMetadata(
```
KennethEnevoldsen, line 52: we have to add to the description what has changed between the two versions

mteb/tasks/retrieval/nld/argu_ana_nl_retrieval.py
```diff
@@ -1,33 +1,29 @@
...
-        domains=["Written", "Non-fiction"],
-        task_subtypes=[],
-        license="cc-by-sa-4.0",
-        annotations_creators="derived",
-        dialect=[],
-        sample_creation="machine-translated and verified",  # manually checked a small subset
-        bibtex_citation=r"""
+#
+# This
+# version
```
Samoed:
```suggestion
```

PR #3528, opened 2025-11-05
[MAEB] Merge with `AbsRetrieval`

Samoed:
I've merged `Any2AnyRetrieval` tasks with `AbsTaskRetrieval`.

I've run evaluation on `laion/clap-htsat-fused`
| | MAEB | PR |
|-|-|-|
| AudioCapsT2ARetrieval | 0.64634 | 0.64679 |
| AudioCapsA2TRetrieval | 0.76104 | 0.75991 |

Difference probably due to additional conversion to `Audio` column type.

Ref https://github.com/embeddings-benchmark/mteb/issues/3410 forgot to create issue for retrieval

### DIFF
--- a/mteb/models/model_implementations/vggish_models.py
+++ b/mteb/models/model_implementations/vggish_models.py
@@ -92,9 +92,6 @@ def vggish_loader(*args, **kwargs):
                 self.device
             )
 
-            if input_tensor.numel() == 0:
-                return None
-
             # Handle different tensor dimensions
             if input_tensor.dim() == 4 and input_tensor.shape[1] == 3:
                 # [batch, 3, height, width] -> [batch, 1, height, width]
@@ -145,18 +142,6 @@ def vggish_loader(*args, **kwargs):
 
                     with torch.no_grad():
                         input_tensor = self._prepare_input_tensor(audio)
-
-                        if input_tensor is None:
-                            # Create zero embedding for empty tensor
-                            logger.debug("Creating zero embedding for empty tensor")
-                            zero_embedding = torch.zeros(
-                                self.embed_dim, device=self.device
-                            )
-                            batch_embeddings.append(
-                                zero_embedding.cpu().detach().numpy()
-                            )
-                            continue
-
                         embedding = self.model(input_tensor)
 
                         # Use mean pooling if needed
--- a/mteb/models/model_implementations/yamnet_models.py
+++ b/mteb/models/model_implementations/yamnet_models.py
@@ -147,18 +147,6 @@ def yamnet_loader(*args, **kwargs):
 
                     with torch.no_grad():
                         input_tensor = self._prepare_input_tensor(audio)
-
-                        if input_tensor is None:
-                            # Create zero embedding for empty tensor
-                            logger.debug("Creating zero embedding for empty tensor")
-                            zero_embedding = torch.zeros(
-                                self.embed_dim, device=self.device
-                            )
-                            batch_embeddings.append(
-                                zero_embedding.cpu().detach().numpy()
-                            )
-                            continue
-
                         # discard logits
                         embedding, _ = self.model(input_tensor)
 


mteb/_create_dataloaders.py
```diff
@@ -342,7 +343,8 @@ def _create_queries_dataloader(
     task_metadata: TaskMetadata,
     input_column: str | None = None,
     batch_size: int = 32,
-) -> DataLoader[QueryInput | ImageInput]:
+    num_workers: int = 1,
```
KennethEnevoldsen, line 362: data_loader_kwargs?

mteb/tasks/audio/any_2_any_retrieval/audio_caps.py
```diff
@@ -69,3 +84,16 @@ class AudioCapsT2ARetrieval(AbsTaskAny2AnyRetrieval):
...
+            for split in self.dataset[subset]:
+                sampling_rate = self.dataset[subset][split]["corpus"][0]["audio"][
+                    "sampling_rate"
+                ]
+                self.dataset[subset][split]["corpus"] = self.dataset[subset][split][
+                    "corpus"
+                ].cast_column(
+                    "audio",
+                    Audio(sampling_rate=sampling_rate),
+                )
```
KennethEnevoldsen, line 99: shouldn't resampling be a part of the model?

KennethEnevoldsen (commented):
Why is it we need the data_loader_kwargs set now but not for images?

mteb/tasks/audio/any_2_any_retrieval/audio_caps.py
Samoed: This is mostly not about resampling, but about converting this column to the `datasets.Audio` format, which is required for proper processing. Previously, it was a list containing the array, path, and sampling rate (at least for datasets v3, because it seems that output format was changed for v4). They are parts of the `Audio` column type, but this format doesn’t get processed correctly.

mteb/_create_dataloaders.py
Samoed: We can move batch_size and num workers to kwargs, but I think it would be better to pass explicitly

Samoed:
Added using `num_workers` in every modality

Samoed:
That's strange that in tests it tries to download with datasets `v4` and failing, but in model interfaces tests are passed. I have this error locally, but I don't know how to fix it and why this happens in CI

AdnanElAssadi:
> That's strange that in tests it tries to download with datasets `v4` and failing, but in model interfaces tests are passed. I have this error locally, but I don't know how to fix it and why this happens in CI

I didn't quite get what you mean.

Samoed:
CI is failing because in this PR datasets with version 3.6.0, but it tries to download datsets, that were uploaded with 4 version and fails because of that. I tried to add support datasets 4 https://github.com/embeddings-benchmark/mteb/pull/3536 for maeb, but it have problems, because they switched audio backend

AdnanElAssadi:
> CI is failing because in this PR datasets with version 3.6.0, but it tries to download datsets, that were uploaded with 4 version and fails because of that. I tried to add support datasets 4 #3536 for maeb, but it have problems, because they switched audio backend

Do you know which datasets are causing the v3/v4 incompatibility? If they are not too many, we could fix the audio backend issues - this seems like the better long-term solution since we would have to go to v4 eventually. Or we could skip tests on the problematic ones and open an issue to fix this later.

Samoed:
> Do you know which datasets are causing the v3/v4 incompatibility

I don't know. This seems a bit random. I tried to run locally and this worked fine with datasets 3

> If they are not too many, we could fix the audio backend issues - this seems like the better long-term solution since we would have to go to v4 eventually

Yes, but main blocker that transformers is not compatible with torchcodec, but we can to create hacky ways to resolve this

KennethEnevoldsen:
Sounds like it might be easier to wait for support and then just limit audio support to datasets v3?

Samoed:
Yes. This is how it's done in #3497

Samoed:
I don't understand why datasets are not downloading. For `BrazilianToxicTweetsClassification` (task 5) have `sequence` https://huggingface.co/datasets/mteb/BrazilianToxicTweetsClassification/blob/d7af1f3c9ff00fbb4e1e40021788a32d78a936bd/README.md?code=true#L23, but when downloading it tries to convert to list, which was added only in datasets v4. I tried to reupload with newer revision https://huggingface.co/datasets/mteb/BrazilianToxicTweetsClassification/commit/12616961c05ed37e1d5d34905c7fa273344bb54a, but it changed sequence to list and it downloaded correctly. Locally, after cache deletion even old revision downloaded correctly

AdnanElAssadi:
Yeah, this is probably from the serialization change between datasets v3 and v4.
Older revisions stored Sequence in a format that's incompatible with v4, so when datasets v4 tries to load them, it fails because it expects list instead.
By reuploading with a newer version, the schema got rewritten in the v4-compatible format (converting Sequence to list), which is why it probably works now.
As for why the old revision worked locally after clearing the cache: that's a bit odd. It could be that there was a corrupted cache entry causing the failure, or possibly some server-side metadata got updated between your attempts. Datasets v4 wouldn't probably rewrite the old revision's schema in cache, so something else must have changed to make it work.

Anyway, I think we can hold this off for a bit, it's not that urgent. I told Chenghao to resume evaluation with what we merged.

Samoed:
>  By reuploading with a newer version, the schema got rewritten in the v4-compatible format (converting Sequence to list), which is why it probably works now.

I was reupploading with datasets `3.6.0`. Maybe problem in CI cache. I will try to clean it

Samoed:
It seems problem of datset loading was in chache. Probably due to https://github.com/embeddings-benchmark/mteb/pull/3536. @KennethEnevoldsen can you rereview this, please? This PR is a bit blocking work on other audio tasks

KennethEnevoldsen (approved):
This looks good - feel free to merge

PR #3536, opened 2025-11-06
[MAEB] Support datasets 4

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3535

### DIFF
--- a/mteb/evaluation/evaluators/audio/audio_pair_classification_evaluator.py
+++ b/mteb/evaluation/evaluators/audio/audio_pair_classification_evaluator.py
@@ -83,10 +83,8 @@ class AudioPairClassificationEvaluator(Evaluator):
             encode_kwargs["batch_size"] = 32
 
         audios = set()
-        for audio in self.audio1:
-            audios.add(tuple(audio["array"].tolist()))
-        for audio in self.audio2:
-            audios.add(tuple(audio["array"].tolist()))
+        for audio in self.audio1 + self.audio2:
+            audios.add(tuple(audio))
         audios = list(audios)
 
         total_audios = len(self.audio1) + len(self.audio2)
@@ -110,12 +108,8 @@ class AudioPairClassificationEvaluator(Evaluator):
             tuple(audio["array"].tolist()): embedding
             for audio, embedding in zip(audios, embeddings)
         }
-        embeddings1 = np.array(
-            [emb_dict[tuple(audio["array"])] for audio in self.audio1]
-        )
-        embeddings2 = np.array(
-            [emb_dict[tuple(audio["array"])] for audio in self.audio2]
-        )
+        embeddings1 = np.array([emb_dict[tuple(audio)] for audio in self.audio1])
+        embeddings2 = np.array([emb_dict[tuple(audio)] for audio in self.audio2])
 
         logger.info("Computing similarity distances.")
         cosine_scores = 1 - paired_cosine_distances(embeddings1, embeddings2)
--- a/mteb/models/model_implementations/ast_model.py
+++ b/mteb/models/model_implementations/ast_model.py
@@ -5,6 +5,7 @@ from typing import Any
 import torch
 from torch.utils.data import DataLoader
 from tqdm.auto import tqdm
+from transformers import ASTFeatureExtractor, ASTModel
 
 from mteb import TaskMetadata
 from mteb._requires_package import requires_audio_dependencies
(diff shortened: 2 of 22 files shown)


Samoed:
For context. This is like almost impossible locally to run for me, because when I'm trying to use `uv sync --extra audio` it took me 50 mins and 15gb RAM (I have 16) and it didn't solve, and I can't solve this issue
```
E       RuntimeError: Could not load libtorchcodec. Likely causes:
E                 1. FFmpeg is not properly installed in your environment. We support
E                    versions 4, 5, 6, and 7 on all platforms, and 8 on Mac and Linux.
E                 2. The PyTorch version (2.9.0+cu130) is not compatible with
E                    this version of TorchCodec. Refer to the version compatibility
E                    table:
E                    https://github.com/pytorch/torchcodec?tab=readme-ov-file#installing-torchcodec.
E                 3. Another runtime dependency; see exceptions below.
E               The following exceptions were raised as we tried to load libtorchcodec:
E
E       [start of libtorchcodec loading traceback]
E       FFmpeg version 8: Could not load this library: /.../mteb/.venv/lib/python3.13/site-packages/torchcodec/libtorchcodec_core8.so
E       FFmpeg version 7: Could not load this library: /.../mteb/.venv/lib/python3.13/site-packages/torchcodec/libtorchcodec_core7.so
E       FFmpeg version 6: Could not load this library: /.../mteb/.venv/lib/python3.13/site-packages/torchcodec/libtorchcodec_core6.so
E       FFmpeg version 5: Could not load this library: /.../mteb/.venv/lib/python3.13/site-packages/torchcodec/libtorchcodec_core5.so
E       FFmpeg version 4: Could not load this library: /.../mteb/.venv/lib/python3.13/site-packages/torchcodec/libtorchcodec_core4.so
E       [end of libtorchcodec loading traceback].
```

Samoed:
Hm, it seems that on CI this is working, but I'm not sure do we want to support this approach. I will update faiss install with windows later

Samoed:
This is very strange. Locally when I do 
```python
from torchcodec._core.ops import load_torchcodec_shared_libraries
load_torchcodec_shared_libraries()

import mteb
```
I don't have errors, but when I move imports to
```python
import mteb

from torchcodec._core.ops import load_torchcodec_shared_libraries
load_torchcodec_shared_libraries()
```
I receive error 
```
OSError: /usr/lib/libtiff.so.6: undefined symbol: jpeg12_write_raw_data, version LIBJPEG_8.0
```
Probably this is problem with PIL (ref https://github.com/python-pillow/Pillow/issues/8148, but I don't use conda)

Samoed:
After step by step import moving I found that locally I have some problem in model implementations folder because if I do 
```python
MODEL_REGISTRY = {meta.name: meta for meta in get_all_model_meta_objects()}
from torchcodec._core.ops import load_torchcodec_shared_libraries

load_torchcodec_shared_libraries()
```
I would receive error. If I move import before model registry, then I won't receive error
```python
from torchcodec._core.ops import load_torchcodec_shared_libraries

# build a registry of all model meta objects
MODEL_REGISTRY = {meta.name: meta for meta in get_all_model_meta_objects()}

load_torchcodec_shared_libraries()
```

Samoed:
I've moved some imports and now importing torchcodec after mteb would work
```python
import mteb

from torchcodec._core.ops import load_torchcodec_shared_libraries
load_torchcodec_shared_libraries()
```
but running `pytest tests/test_tasks/test_maeb_datasets.py` would still fail

Samoed:
It seems that this brings a lot of problems. I think this is better to wait until `transformers` v5 with proper support will be released. Even in CI it looks fine, when I try to run 
```bash
mteb run -m MIT/ast-finetuned-audioset-10-10-0.4593 -t ESC50PairClassification
```
I get torchcodec error, but this is very strange. I think this error somewhere from transformers. Because when I do 
```python
from torchcodec._core.ops import load_torchcodec_shared_libraries
load_torchcodec_shared_libraries()

from transformers import ASTFeatureExtractor, ASTModel
```
this is fine, but when I change to 
```python
from transformers import ASTFeatureExtractor, ASTModel
from torchcodec._core.ops import load_torchcodec_shared_libraries
load_torchcodec_shared_libraries()
```
I'm getting this error again. We can try to load on startup, but this will be hacky. Ref https://github.com/huggingface/transformers/issues/42103 https://github.com/meta-pytorch/torchcodec/issues/1030

KennethEnevoldsen:
Looks like it might be better to wait with this?

Samoed:
Probably yes. We can create a workaround for this, but I don't think that good idea. But I think we still need to do this before merge into main, because otherwise ci would use old dataset version

KennethEnevoldsen:
Hmm yea I see the problem (we could have audio as a seperate step in the test that downgrade the version)

Samoed:
Yes, agree

KennethEnevoldsen:
Alright - should we then close this?

Issue #3547, opened 2025-11-11
Batch size not correctly set for Classification, PairClassification, STS, and Summarization tasks


Geralt-Targaryen:
### Describe the bug

Hi, I implemented a custom model following the specification here: https://embeddings-benchmark.github.io/mteb/api/model/#mteb.models.EncoderProtocol.encode 

And I found that in the current version, batch size is not correctly set for certain tasks:

Classification (defaults to 1 from torch):
https://github.com/embeddings-benchmark/mteb/blob/1a02edbb7d50ec9f02c258267de299a731ea10ed/mteb/_evaluators/sklearn_evaluator.py#L74-L75

PairClassification (defaults to 32, defined in mteb/_create_dataloaders.py/_create_dataloader_from_texts):
https://github.com/embeddings-benchmark/mteb/blob/1a02edbb7d50ec9f02c258267de299a731ea10ed/mteb/_evaluators/pair_classification_evaluator.py#L170-L176

STS (defaults to 32, defined in mteb/_create_dataloaders.py/create_dataloader):
https://github.com/embeddings-benchmark/mteb/blob/1a02edbb7d50ec9f02c258267de299a731ea10ed/mteb/_evaluators/any_sts_evaluator.py#L48-L57

Summarization (defaults to 32, defined in mteb/_create_dataloaders.py/_create_dataloader_from_texts):
https://github.com/embeddings-benchmark/mteb/blob/1a02edbb7d50ec9f02c258267de299a731ea10ed/mteb/_evaluators/text/summarization_evaluator.py#L106-L133

### To reproduce

Source code has been identified in the description above

### Are you interested to contribute a fix for this bug?

Yes

Samoed:
Yes, you're right. Thank you for catch! I'll fix it and try to check other places

PR #3548, opened 2025-11-11
fix: Pass encode kwargs in all dataloaders

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3547

### DIFF
--- a/mteb/_evaluators/any_sts_evaluator.py
+++ b/mteb/_evaluators/any_sts_evaluator.py
@@ -8,7 +8,7 @@ from sklearn.metrics.pairwise import (
     paired_manhattan_distances,
 )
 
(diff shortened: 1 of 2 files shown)


mteb/_evaluators/sklearn_evaluator.py
```diff
@@ -50,33 +50,20 @@ def __init__(
         self.evaluator_model = evaluator_model
 
     def create_dataloaders(
```
KennethEnevoldsen, line 52: We are breaking backward compat on a public function

Samoed: Overall evaluators are private, and create dataloaders is function that used only here, but I can change this to be compatible

PR #3549, opened 2025-11-11
Model : Tarka Embedding 350M V1

jaswanth-0821:

### DIFF
--- a/mteb/models/model_implementations/tarka_models.py
+++ b/mteb/models/model_implementations/tarka_models.py
@@ -341,7 +341,6 @@ tarka_embedding_150m_v1 = ModelMeta(
 )
 
 tark_embedding_350_v1_kwargs = dict(
-    device="cuda",  # use a gpu
     model_kwargs={
         "attn_implementation": "flash_attention_2",
         "torch_dtype": "bfloat16",


mteb/models/model_implementations/tarka_models.py
```diff
@@ -35,6 +320,71 @@
     "CodeSearchNet",
 }
 
+
+class TarkaSentenceTransformer(InstructSentenceTransformerModel):
```
Samoed: I don't think that you need to create this wrapper

jaswanth-0821: Initially, I tested the setup using the Qwen3 Embedding evaluation code. Before submitting here, I re-tested with `mteb`  with the existing wrappers and noticed a slight drop in performance. To address this, I implemented the above wrapper based on the approach used in the Qwen3 Embedding repository.
```
dataset : AskUbuntuDupQuestions
Using q3e_instruct_loader : "main_score": 0.61272,
Using TarkaSentenceTransformer : "main_score": 0.63091,
```

Samoed: But it's not correct. You don't using most of your instructions here. You shouldn't have any difference in this task using `InstructSentenceTransformerModel` directly. Most of the `get_instruction` can be added to prompt dict directly

mteb/models/model_implementations/tarka_models.py
```diff
@@ -9,6 +10,287 @@
...
+    "WikipediaRetrievalMultilingual": "Retrieval the relevant passage for the given query",
+    "Core17InstructionRetrieval": "Retrieval the relevant passage for the given query",
+    "News21InstructionRetrieval": "Retrieval the relevant passage for the given query",
+    "Robust04InstructionRetrieval": "Retrieval the relevant passage for the given query",
+    "WebLINXCandidatesReranking": "Retrieval the relevant passage for the given query",
+    "WikipediaRerankingMultilingual": "Retrieval the relevant passage for the given query",
+    "STS15": "Retrieve semantically similar text",
+    "MIRACLRetrievalHardNegatives": "Retrieval relevant passage for the given query",
+    "BIOSSES": "Retrieve semantically similar text",
+    "CQADupstackRetrieval": "Given a question, retrieve detailed question descriptions from Stackexchange that are duplicates to the given question",
```
Samoed, line 280: This prompt won't be applied, because `CQADupstackRetrieval` is aggregated task

jaswanth-0821: yes , included the individual prompts as well.

Samoed: But you added only 2 tasks from 12 https://github.com/embeddings-benchmark/mteb/blob/fe83e27968adff648de9520e187164ea5f906935/mteb/tasks/aggregated_tasks/eng/cqadupstack_retrieval.py#L19-L30

jaswanth-0821:
Thanks for pointing out the wrapper. I’ve retested using `InstructSentenceTransformerModel` on multiple tasks, and the results are now identical. I’ve also updated the changes here. -- please let me know if any further updates are needed from my side.

mteb/models/model_implementations/tarka_models.py
jaswanth-0821: Sorry for asking, but are these tasks evaluated under **MTEB (English v2)**? From what I can see, only the following two appear in the benchmark:
```
CQADupstackGamingRetrieval  
CQADupstackUnixRetrieval
```

I’m currently focusing only on **MTEB (English v2)** and will update the other benchmarks later.

mteb/models/model_implementations/tarka_models.py
```diff
@@ -56,3 +339,39 @@
...
     citation=Tarka_Embedding_150M_V1_CITATION,
 )
+
+tark_embedding_350_v1_kwargs = dict(
+    device="cuda",  # use a gpu
+    model_kwargs={
+        "attn_implementation": "flash_attention_2",
+        "device_map": "cuda",
+        "torch_dtype": "bfloat16",
+    },  # use low-precision
```
Samoed: Can you remove `device` and `device_map`? I think users can select device by themselves. 
```suggestion
    model_kwargs={
        "attn_implementation": "flash_attention_2",
        "torch_dtype": "bfloat16",
    },  # use low-precision
```

mteb/models/model_implementations/tarka_models.py
```diff
@@ -56,3 +339,39 @@
...
+    instruction_template="Instruct: {instruction}\nQuery:",
+)
+
+tarka_embedding_350m_v1 = ModelMeta(
+    loader=InstructSentenceTransformerModel,
+    loader_kwargs=tark_embedding_350_v1_kwargs,
+    name="Tarka-AIR/Tarka-Embedding-350M-V1",
+    languages=MULTILINGUAL_EVALUATED_LANGUAGES,
+    open_weights=True,
+    revision="f4b5de82060cf3a833e52580e7ce59adeacb6fb5",  # Commit of @tomaarsen
```
Samoed:
```suggestion
    revision="f4b5de82060cf3a833e52580e7ce59adeacb6fb5",
```

Samoed:
Othewise looks good

mteb/models/model_implementations/tarka_models.py
```diff
@@ -56,3 +339,38 @@
     memory_usage_mb=576,
     citation=Tarka_Embedding_150M_V1_CITATION,
 )
+
+tark_embedding_350_v1_kwargs = dict(
+    device="cuda",  # use a gpu
```
Samoed:
```suggestion
```

jaswanth-0821: My bad — Thanks for that , I will be careful next time

PR #3550, opened 2025-11-11
fix: add typecheck

Samoed:
Add typechecking

Close https://github.com/embeddings-benchmark/mteb/issues/2714

### DIFF
--- a/mteb/abstasks/aggregated_task.py
+++ b/mteb/abstasks/aggregated_task.py
@@ -6,7 +6,6 @@ from typing import Any
 
 import numpy as np
 from datasets import Dataset, DatasetDict
-from typing_extensions import Self
 
 from mteb.models.models_protocols import MTEBModels
 from mteb.results.task_result import TaskResult
@@ -121,19 +120,6 @@ class AbsTaskAggregate(AbsTask):
             task_res.mteb_version = None
         task_res.mteb_version = task_results[0].mteb_version
         return task_res
-
-    def filter_eval_splits(self, eval_splits: list[str] | None) -> Self:
-        """Filter the evaluation splits of the task.
-
-        Args:
-            eval_splits: List of splits to evaluate on. If None, all splits in metadata
-                are used.
-
-        Returns:
-            The task with filtered evaluation splits.
-        """
-        self._eval_splits = eval_splits
-        return self
 
     def evaluate(
         self,


mteb/results/benchmark_results.py
```diff
@@ -162,7 +165,7 @@ def _filter_models(
 
         return type(self).model_construct(model_results=new_model_results)
 
-    def join_revisions(self) -> Self:
+    def join_revisions(self) -> "BenchmarkResults":
```
KennethEnevoldsen: I don't think it needs the `"`

mteb/results/benchmark_results.py
```diff
@@ -361,7 +364,8 @@ def to_dataframe(
             format=format,
         )
 
-    def __iter__(self) -> Iterator[ModelResult]:
+    @override
+    def __iter__(self) -> Iterator[ModelResult]:  # type: ignore[override]
```
KennethEnevoldsen, line 413: I am not sure I get the type error here?

mteb/results/benchmark_results.py
```diff
@@ -117,9 +117,12 @@ def select_models(
 
         for name, revision in zip(names, _revisions):
             if isinstance(name, ModelMeta):
+                if name.name is None:
+                    raise ValueError("ModelMeta must have a name.")
```
KennethEnevoldsen:
```suggestion
                    raise ValueError("name in ModelMeta is None. It must be a string.")
```

mteb/models/instruct_wrapper.py
```diff
@@ -189,7 +192,7 @@ def encode(
             The encoded input in a numpy array or torch tensor of the shape (Number of sentences) x (Embedding dimension).
         """
         sentences = [text for batch in inputs for text in batch["text"]]
-        instruction = self.get_task_instruction(task_metadata, prompt_type)
+        instruction: str | None = self.get_task_instruction(task_metadata, prompt_type)
```
KennethEnevoldsen: shouldn't this be annotated by the function?

mteb/filter_tasks.py
```diff
@@ -32,54 +32,23 @@ def _check_is_valid_language(lang: str) -> None:
...
-    domains: list[TaskDomain] | None = None,
-    task_types: list[TaskType] | None = None,  # type: ignore
-    categories: list[TaskCategory] | None = None,
-    modalities: list[Modalities] | None = None,
-    exclusive_modality_filter: bool = False,
-    exclude_superseded: bool = False,
-    exclude_aggregate: bool = False,
-    exclude_private: bool = False,
-) -> list[type[AbsTask]]: ...
+T = TypeVar("T", AbsTask, type[AbsTask])
```
KennethEnevoldsen: Does this guarantee that if I provide AbsTask, then I won't get an `type[AbsTask]` out?

KennethEnevoldsen: Also how does this influence the docs?

mteb/evaluate.py
```diff
@@ -258,7 +269,7 @@ def _check_model_modalities(
 
 def evaluate(
     model: ModelMeta | MTEBModels | SentenceTransformer | CrossEncoder,
-    tasks: AbsTask | Iterable[AbsTask],
+    tasks: AbsTask | Benchmark | Iterable[AbsTask | Benchmark],
```
KennethEnevoldsen: Can it be an iterable of benchmarks? Benchmark should be an iterable of AbsTask

mteb/evaluate.py
```diff
@@ -86,27 +87,28 @@ def _sanitize_model(
 ) -> tuple[MTEBModels | ModelMeta, ModelMeta, ModelName, Revision]:
     from sentence_transformers import CrossEncoder, SentenceTransformer
 
+    wrapped: MTEBModels | ModelMeta
```
```suggestion
    wrapped_model: MTEBModels | ModelMeta
```

mteb/deprecated_evaluator.py
```diff
@@ -70,13 +71,20 @@ def __init__(
         """
         from mteb.benchmarks import Benchmark
 
-        self.tasks = list(tasks)
-        if len(self.tasks) > 0 and isinstance(self.tasks[0], Benchmark):
+        if isinstance(tasks, list) and all(
+            isinstance(task, Benchmark) for task in tasks
+        ):
             self.benchmarks = tasks
-            self.tasks = list(chain.from_iterable(self.tasks))
```
KennethEnevoldsen, line 72: unsure why this was changed?

mteb/cli/generate_model_card.py
```diff
@@ -12,7 +13,7 @@
 
 def generate_model_card(
     model_name: str,
-    tasks: list[AbsTask] | None = None,
+    tasks: Sequence[AbsTask] | None = None,
```
KennethEnevoldsen, line 17: iterable or sequence?

mteb/abstasks/abstask.py
```diff
@@ -60,6 +60,12 @@ def _multilabel_subsampling(
     return dataset_dict
 
 
+class AbsMetrics(TypedDict):
+    """The abstract class for the metrics returned by the tasks"""
+
+    ...
+
```
KennethEnevoldsen: Unsure why this is added?

mteb/results/benchmark_results.py
Samoed: BaseModel have it's own iterator

mteb/abstasks/abstask.py
Samoed: I tried to standartize, because dict is not compatible with mappting, but I will remove this I think and change evaluate subset to Mapping, which can handle both

mteb/filter_tasks.py
Samoed: > Does this guarantee that if I provide AbsTask, then I won't get an type[AbsTask] out?

Yes

> Also how does this influence the docs?

I'll check

mteb/models/instruct_wrapper.py
Samoed: Get task instruction will always return string, but later we have `instructions = None` and this cause conflict. Probably later can be changed to empty string

mteb/results/benchmark_results.py
Samoed: I can change to this, but would require `__future__`

mteb/evaluate.py
Samoed: Hm, it seems that we don't support this, because I mistakenly looked to deprecated evaluator https://github.com/embeddings-benchmark/mteb/blob/fe83e27968adff648de9520e187164ea5f906935/tests/test_deprecated/test_MTEB.py#L16

Probably we need to add a test for new evaluator too

mteb/results/benchmark_results.py
KennethEnevoldsen: Would prefer `__future__`

mteb/results/benchmark_results.py
KennethEnevoldsen: Ahh, but it still gives you a type error even with the decorator (seems to me like we are specifying it twice)

mteb/models/instruct_wrapper.py
KennethEnevoldsen: Then I would probably do like this:
```suggestion
        instruction: str | None # further up
        instruction = self.get_task_instruction(task_metadata, prompt_type)
```
otherwise we are suggesting that the function could also return `None` even though it can't

mteb/results/benchmark_results.py
Samoed: Yes, I'll remove it

mteb/evaluate.py
KennethEnevoldsen: We can add support for it, but I am leaning toward not doing it. Keeps it cleaner and it is not like a for loop across benchmarks is hard to do.

Samoed: Agree

KennethEnevoldsen:
@Samoed what the is the status on this PR?

Samoed:
I want to continue to work, but currently I don't have time now

KennethEnevoldsen:
Perfectly fine - I will leave it as is for now then

mteb/filter_tasks.py
Samoed: PR
image
main
image

mteb/deprecated_evaluator.py
Samoed: I have problems to correctly extract tasks from benchmarks objects

mteb/cli/generate_model_card.py
Samoed: We're using `Sequence` in most places, so I think we can use it here too

mteb/filter_tasks.py
Samoed: Rolled back

Samoed:
With current configuration of `mypy` was 340 (currently in main) error and now 0.
- if activate strict other 180 errors
- if add tasks and models folders then 700 errors
- if add tasks and models folders and strict then other 2_000 errors 
- `ty check mteb` show 2_900 errors

mteb/_evaluators/text/summarization_evaluator.py
```diff
@@ -216,17 +218,23 @@ def _calculate_metrics(
             strict=True,
         ):
             cosine_spearman_scores.append(
-                spearmanr(human_scores, cosine_pred_scores).statistic
+                cast(float, spearmanr(human_scores, cosine_pred_scores).statistic)
```
KennethEnevoldsen: Why not just to the float conversion 
```suggestion
                float(spearmanr(human_scores, cosine_pred_scores).statistic)
```

mteb/_evaluators/text/summarization_evaluator.py
```diff
@@ -273,10 +281,10 @@ def _calculate_metrics(
             pearson_scores.append(pearsonr(human_scores, sim_scores))
 
         return SummarizationMetrics(
-            pearson=float(np.mean(pearson_scores)),
-            spearman=float(np.mean(spearman_scores)),
-            cosine_spearman=float(np.mean(cosine_spearman_scores)),
-            cosine_pearson=float(np.mean(cosine_pearson_scores)),
-            dot_pearson=float(np.mean(dot_pearson_scores)),
-            dot_spearman=float(np.mean(dot_spearman_scores)),
+            pearson=float(np.mean(pearson_scores)),  # type: ignore[arg-type]
```
KennethEnevoldsen, line 280: Not sure what the problem is here?

mteb/_evaluators/text/summarization_evaluator.py
```diff
@@ -201,12 +203,12 @@ def _calculate_metrics(
         self,
         distances: SummarizationDistances,
     ) -> SummarizationMetrics:
-        cosine_spearman_scores = []
-        cosine_pearson_scores = []
-        dot_spearman_scores = []
-        dot_pearson_scores = []
-        pearson_scores = []
-        spearman_scores = []
+        cosine_spearman_scores: list[float] = []
```
KennethEnevoldsen: is it important that these are list float (they could be list of anything that is valid for np.mean

mteb/abstasks/clustering_legacy.py
```diff
@@ -89,6 +89,9 @@ def _evaluate_subset(
         prediction_folder: Path | None = None,
         **kwargs: Any,
     ) -> ScoresDict:
+        if not isinstance(model, EncoderProtocol):
+            raise TypeError("Expected model to be an instance of EncoderProtocol")
```
KennethEnevoldsen, line 93: This seems weird - if we suspect an EncoderProtocol, only then is there no reason to state otherwise? (same above)

mteb/abstasks/multilabel_classification.py
```diff
@@ -232,10 +234,9 @@ def _undersample_data_indices(
         """
         sample_indices = []
         if idxs is None:
-            idxs = np.arange(len(y))
+            idxs = list(range(len(y)))
```
KennethEnevoldsen: are we sure that this does not influence rng-state?

mteb/abstasks/pair_classification.py
```diff
@@ -237,6 +240,8 @@ def _compute_image_hash(inputs: list) -> list[str]:
 
     def _push_dataset_to_hub(self, repo_name: str) -> None:
         # previously pair classification datasets were stored in a single row
+        if self.dataset is None:
+            raise RuntimeError("Dataset not loaded")
```
KennethEnevoldsen: Could suggest how to load

mteb/abstasks/sts.py
```diff
@@ -103,14 +103,17 @@ class AbsTaskSTS(AbsTask):
...
-        model: EncoderProtocol,
+        model: MTEBModels,
         data_split: Dataset,
         encode_kwargs: dict[str, Any],
         hf_split: str,
         hf_subset: str,
         prediction_folder: Path | None = None,
         **kwargs: Any,
     ) -> STSMetrics:
+        if not isinstance(model, EncoderProtocol):
```
KennethEnevoldsen, line 114: just tagging one more of these but there are more

mteb/abstasks/text/bitext_mining.py
```diff
@@ -78,6 +79,23 @@ def evaluate(
         **kwargs: Any,
     ) -> dict[HFSubset, ScoresDict]:
         """Added load for "parallel" datasets"""
+        if isinstance(model, CrossEncoderProtocol) and not self._support_cross_encoder:
```
KennethEnevoldsen: Generally it seems like we have to settle on how we input models for tasks and evaluators

mteb/cli/generate_model_card.py
KennethEnevoldsen: In general, Iterable is less strict than Sequence. Sequence allows for list[1] and len(list). So it is more naive to type iterable.

You typically want to type your inputs as open as possible (maximize user flexibility) and your outputs as closed as possible (minimize user uncertainty). That being said Sequence is def. better than list.

mteb/deprecated_evaluator.py
KennethEnevoldsen: Not sure I get what the problem is?

mteb/deprecated_evaluator.py
```diff
@@ -306,13 +315,16 @@ def run(
         elif verbosity == 3:
             datasets.logging.set_verbosity(logging.DEBUG)
 
-        meta = self.create_model_meta(model)
-        output_path = self._create_output_folder(meta, output_folder)
-
+        mteb_model: MTEBModels
         if isinstance(model, SentenceTransformer):
-            model = SentenceTransformerEncoderWrapper(model)
+            mteb_model = cast(EncoderProtocol, SentenceTransformerEncoderWrapper(model))
```
KennethEnevoldsen: Minor nit, but generally prefer keeping casting operations separate from code changes (they are mostly there to appease the type checker and act as documentations so it feels weird to have them both intertwined)

mteb/deprecated_evaluator.py
```diff
@@ -333,18 +345,20 @@ def run(
...
         # To evaluate missing splits, we keep track of the task name and the corresponding splits.
-        self.last_evaluated_splits = {}
+        self._last_evaluated_splits = {}
 
         while len(self.tasks) > 0:
             task = self.tasks[0]
             logger.info(
                 f"\n\n********************** Evaluating {task.metadata.name} **********************"
             )
+            save_path: Path | None = None
```
KennethEnevoldsen:
```suggestion
            save_path: Path
```
isn't this enough?

mteb/_evaluators/text/summarization_evaluator.py
Samoed: Same as in `SummarizationEvaluator` I just didn't want to change much deprecated evaluator, but can do this similarly to not deprecated evaluator

mteb/deprecated_evaluator.py
Samoed: No, because later it would be assigned to `None`. Probably I just can rename this for `aggregated_task` and leave rest as is

mteb/_evaluators/text/summarization_evaluator.py
Samoed: No, but here we're using them as lists

mteb/cli/generate_model_card.py
Samoed: I made resolving `Benchmark` class as `Iterable[AbsTask]` 
```diff
-     def __iter__(self) -> Iterable[AbsTask]:
+     def __iter__(self) -> Iterator[AbsTask]:
```

 and I think iterable would make more reason for `AbsTasks` now

mteb/_evaluators/text/summarization_evaluator.py
KennethEnevoldsen: yeah - I was just unsure why they were typed (pyright would just guess the type)

mteb/_evaluators/text/summarization_evaluator.py
Samoed: Also cast to float here won't work, because if I do this I would get
```
Argument 1 to "float" has incompatible type "PearsonRResult[float64, float64]"; expected "str | Buffer | SupportsFloat | SupportsIndex"
```

mteb/deprecated_evaluator.py
Samoed: Don't know why I've added cast here, because it's already passed check on protocols

mteb/abstasks/multilabel_classification.py
Samoed: Changed to `list(np.arange(len(y)))`

mteb/abstasks/clustering_legacy.py
Samoed: Not sure what you want to do

mteb/abstasks/pair_classification.py
Samoed: Overall this shouldn't happen, but I updated error message

mteb/_evaluators/text/summarization_evaluator.py
Samoed: Maybe I've added this before casting to float to resolve issues, because now it's working fine

mteb/deprecated_evaluator.py
Samoed: If I change init as it on main, I would get
```
mteb/deprecated_evaluator.py:68: error: Argument 1 to "list" has incompatible type "Iterable[AbsTask | Benchmark]"; expected "Iterable[AbsTask]"  [arg-type]
mteb/deprecated_evaluator.py:71: error: Argument 1 to "from_iterable" of "chain" has incompatible type "list[AbsTask]"; expected "Iterable[Iterable[Never]]"  [arg-type]
mteb/deprecated_evaluator.py:71: error: Argument 1 to "from_iterable" of "chain" has incompatible type "list[AbsTask]"; expected "Iterable[Iterable[AbsTask]]"  [arg-type]
```

Samoed: Changed to be more similar with existing

mteb/abstasks/clustering_legacy.py
KennethEnevoldsen: I would just specify that the function only takes an encoder protocol instead of stating that it takes a MTEBModels and afterwards raising type error

mteb/abstasks/multilabel_classification.py
KennethEnevoldsen: we have the shuffle afterwards

mteb/abstasks/text/bitext_mining.py
Samoed: I don't think we can change it much now. Do you have ideas how to do this?

mteb/abstasks/pair_classification.py
KennethEnevoldsen: let us add a comment on that

mteb/deprecated_evaluator.py
KennethEnevoldsen: the cast is on the wrapper - you are stating that the wrapper is an EncoderProtocol (which it should be?)

mteb/abstasks/text/bitext_mining.py
KennethEnevoldsen: It is more about types - generally right now we specify MTEBModels even when it doesn't take all MTEBModels I would instead specify the ones that they can take

Samoed: Previously it was like this, but we `_evaluate_subset` have in annotation `MTEBModels` and I don't know how to change this

mteb/abstasks/clustering_legacy.py
Samoed: Changing to `_evaluate_subset` to `model: EncoderProtocol` would cause conflicts with `AbsTask._evaluate_subset` because it's should receive `MTEBModels`

```
mteb/abstasks/classification.py:185: error: Argument 1 of "_evaluate_subset" is incompatible with supertype "mteb.abstasks.abstask.AbsTask"; supertype defines the argument type as "EncoderProtocol | CrossEncoderProtocol | SearchProtocol"  [override]
mteb/abstasks/classification.py:185: note: This violates the Liskov substitution principle
mteb/abstasks/classification.py:185: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```

mteb/abstasks/text/bitext_mining.py
KennethEnevoldsen: But is that a problem? EncoderProtocol should still be valid to pass to MTEBModels?

mteb/abstasks/multilabel_classification.py
Samoed: This won't make difference
```python
import numpy as np

rng = np.random.default_rng(42)
idxs = np.arange(100)
rng.shuffle(idxs)
print(idxs)
# [59 21 56 18 33 42 50 27 92 71  2 25 24 52 81 99 51 61  4 97 94 37 28 80
#  7 57 79 26 76 78 38  0 86 46 10 29 70 45 17 39 91 40 89  5 68  1 95 84
# 63  3 83 93 20 87 43 73 32 75 23 96 69 15 48 55 60  9 62 31 72 16 44 82
# 67 34 30 58 74 88 98 54 49  6 90 11 85 41 22 19 35 64 47 12 53 14 13 66
# 36 65 77  8]

rng = np.random.default_rng(42)
idxs = list(np.arange(100))
rng.shuffle(idxs)
print(idxs)
# [np.int64(59), np.int64(21), np.int64(56), np.int64(18), np.int64(33), np.int64(42), np.int64(50), np.int64(27), np.int64(92), np.int64(71), np.int64(2), np.int64(25), np.int64(24), np.int64(52), np.int64(81), np.int64(99), np.int64(51), np.int64(61), np.int64(4), np.int64(97), np.int64(94), np.int64(37), np.int64(28), np.int64(80), np.int64(7), np.int64(57), np.int64(79), np.int64(26), np.int64(76), np.int64(78), np.int64(38), np.int64(0), np.int64(86), np.int64(46), np.int64(10), np.int64(29), np.int64(70), np.int64(45), np.int64(17), np.int64(39), np.int64(91), np.int64(40), np.int64(89), np.int64(5), np.int64(68), np.int64(1), np.int64(95), np.int64(84), np.int64(63), np.int64(3), np.int64(83), np.int64(93), np.int64(20), np.int64(87), np.int64(43), np.int64(73), np.int64(32), np.int64(75), np.int64(23), np.int64(96), np.int64(69), np.int64(15), np.int64(48), np.int64(55), np.int64(60), np.int64(9), np.int64(62), np.int64(31), np.int64(72), np.int64(16), np.int64(44), np.int64(82), np.int64(67), np.int64(34), np.int64(30), np.int64(58), np.int64(74), np.int64(88), np.int64(98), np.int64(54), np.int64(49), np.int64(6), np.int64(90), np.int64(11), np.int64(85), np.int64(41), np.int64(22), np.int64(19), np.int64(35), np.int64(64), np.int64(47), np.int64(12), np.int64(53), np.int64(14), np.int64(13), np.int64(66), np.int64(36), np.int64(65), np.int64(77), np.int64(8)]
```

mteb/deprecated_evaluator.py
Samoed: Yes, I removed casting here

mteb/abstasks/text/bitext_mining.py
Samoed: > EncoderProtocol should still be valid to pass to MTEBModels?

Yes, models should implement one of our protocols to work with mteb. Most of the tasks support only Encoder, so check for this should be implemented in each task

mteb/abstasks/aggregated_task.py
```diff
@@ -120,12 +122,18 @@ def combine_task_results(self, task_results: list[TaskResult]) -> TaskResult:
         task_res.mteb_version = task_results[0].mteb_version
         return task_res
 
-    def check_if_dataset_is_superseded(self) -> None:
-        """Check if the dataset is superseded by a newer version"""
-        if self.superseded_by:
-            logger.warning(
-                f"Dataset '{self.metadata.name}' is superseded by '{self.superseded_by}', you might consider using the newer version of the dataset."
-            )
```
KennethEnevoldsen: is this intentionally deleted?

mteb/_evaluators/text/summarization_evaluator.py
KennethEnevoldsen: alright - let us keep it like this for now

mteb/abstasks/clustering_legacy.py
KennethEnevoldsen: Hmm, it is because our tasks interface is not consistent. I would probably prefer stating that what it actually takes and then ignoring type errors rather than stating that we take MTEBModels

mteb/abstasks/aggregated_task.py
Samoed: This was done in different PR https://github.com/embeddings-benchmark/mteb/pull/3753. This was deleted, because this function is the same as in `AbsTask`

mteb/abstasks/clustering_legacy.py
Samoed: I've made it consistent now, and I think adding a check is better than ignoring override problems

KennethEnevoldsen: My preference is slightly on the other direction, but let us keep it as it is now

mteb/abstasks/text/bitext_mining.py
KennethEnevoldsen: let us keep it as is

Samoed:
Yes, I've added `py.typed` and checked that it would install correctly https://github.com/embeddings-benchmark/mteb/blob/a09734fbeb32a6486446a69df11f9b10d248799b/mteb/py.typed

KennethEnevoldsen:
thanks! and thanks for taking the time on this one

PR #3555, opened 2025-11-13
fix: MTEB-NL switches to v2 datasets with new prompts

nikolay-banar:
This PR changes some datasets from MTEB-NL to v2 with the new prompts.


nikolay-banar:
> typically we would bump the version for this as well but since it hasn't been up for that long we can leave it at v1
> 
> Note that this will outdate existing models (so it might leave to an unfinished table)

I have already added some results for the v2 datasets.

KennethEnevoldsen:
Great then we should be just fine

PR #3557, opened 2025-11-14
add jasper token compression model

DunZhang:

Dear MTEB maintainers, the following information may be helpful for your review of this PR:
1) This model uses the same prompt as the qzhou model
2) The training dataset for this model is consistent with qzhou's


DunZhang:
I am very excited about this model: this is the first vector model in the Jasper and Stella series to use **dynamic text token compression**. Through the combination of vector distillation and contrastive learning, our model can compress text by 3x while still achieving excellent performance!

Samoed (approved):

KennethEnevoldsen:
Great to have it! Thanks for the PR!

PR #3558, opened 2025-11-14
fix: issues on cache hits

KennethEnevoldsen:
Found a couple of issues where the cache is not hit

- the condition `overwrite_strategy == "only-missing"
        and overwrite_strategy == OverwriteStrategy.ONLY_MISSING` would never be met as it is never both (so we always rerun all splits)
- Currently, the remote cache path is not specified correctly (`remote` vs `remote/results`), which means that it is never hit.
- Added a check if a merge is required. E.g., it is often the case that you don't need to merge the new results (because the results include all the splits), however, we still rerun them if the version does not match
-  Added better error messages to merge messages
- Added test that cache re-used as intended

### DIFF
--- a/tests/test_evaluate.py
+++ b/tests/test_evaluate.py
@@ -134,7 +134,7 @@ def test_cache_hit(task: AbsTask):
     model = mteb.get_model("baseline/random-encoder-baseline")
     with pytest.raises(
         ValueError,
-        match="ValueError: overwrite_strategy is set to 'only-cache' and the results file exists. However there are the following missing splits (and subsets): {'dev': ['default']}. To rerun these set overwrite_strategy to 'only-missing'.",
+        match="overwrite_strategy is set to 'only-cache' and the results file exists",
     ):
         mteb.evaluate(model, task, overwrite_strategy="only-cache")
 


KennethEnevoldsen:
Yep probably

KennethEnevoldsen:
Added test for cache re-use. Did not add tests for remote cache.

tests/test_evaluate.py
```diff
@@ -77,6 +77,18 @@ def test_evaluate_with_cache(
...
+    cached_results = mteb.evaluate(
+        model, task, cache=cache, overwrite_strategy="only-cache"
+    )
+    cached_result = cached_results[0]
+    assert cached_result.task_name == task.metadata.name, (
+        "results should match the task"
+    )
+    assert cached_result.get_score() == expected_score, (
+        "main score should match the expected value"
+    )
```
Samoed, line 90: Can we add check that result that task result that not in cache don't load?

Samoed: Also, we can probably reuse `tests/historic_results` to test cache

KennethEnevoldsen: I am unsure how this would look?

Samoed: Oh, I forgot that we also have `tests/mock_mteb_cache`. And with it, cache can be tested.

KennethEnevoldsen: Right, but clearly it should never load results that are not in the cache, but I am unsure how I would test that (we kinda do it already by checking the scores)

Samoed: For check that datasets using `only-cache`

```python
import mteb
import pytest

def test_cache_hit():
    task = mteb.get_task("MIRACLRetrieval")
    with pytest.raises(ValueError, match="ValueError: overwrite_strategy is set to 'only-cache' and the results file exists. However there are the following missing splits (and subsets): {'dev': ['default']}. To rerun these set overwrite_strategy to 'only-missing'."):
        scores = mteb.evaluate(model, task, overwrite_strategy="only-cache")
```

Probably test for `only-missing` can be done by mocking and checking call times

KennethEnevoldsen: | will add this - we have a test for `only-missing`, but not one that tests for missing, but we have for the positive case.

Samoed:
I tried to create test for checking that `only-missing` working correctly, but not sure if we want to add it
```python
def test_only_missing(tmp_path):
    task = MockMultilingualPairClassificationTask()
    # select only 1 splits from 2
    task.hf_subsets = ["eng"]
    cache = ResultCache(tmp_path)
    with patch.object(
        AbsTaskPairClassification,
        "_evaluate_subset",
        wraps=AbsTaskPairClassification._evaluate_subset,
        autospec=True,
    ) as mock_pair_classification_evaluate:
        results = mteb.evaluate(
            model, task, cache=cache, overwrite_strategy="only-missing"
        )
        assert len(results.task_results[0].scores["test"]) == 1
        mock_pair_classification_evaluate.assert_called_once()

    # select both splits (eng, fra)
    task = MockMultilingualPairClassificationTask()
    with patch.object(
        AbsTaskPairClassification,
        "_evaluate_subset",
        wraps=AbsTaskPairClassification._evaluate_subset,
        autospec=True,
    ) as mock_pair_classification_evaluate2:
        results = mteb.evaluate(
            model, task, cache=cache, overwrite_strategy="only-missing"
        )
        assert len(results.task_results[0].scores["test"]) == 2
        mock_pair_classification_evaluate2.assert_called_once()
```

KennethEnevoldsen:
Hmm I probably wouldn't add it

Issue #3562, opened 2025-11-15
vidore-v3 private set evaluation issue


StupidBuluchacha:
### Describe the bug

Hi mteb team,

I was evaluating the vidore-v3 benchmark with 8 public and 2 private sets. The evaluation was stopped because it cannot access the private sets. I am not sure what caused this issue.

Below is my error:
`Traceback (most recent call last):
  File "/mnt/workspace/miniconda3/envs/mieb/bin/mteb", line 7, in <module>
    sys.exit(main())
  File "/mnt/workspace/mteb/mteb/cli/build_cli.py", line 392, in main
    args.func(args)
  File "/mnt/workspace/mteb/mteb/cli/build_cli.py", line 84, in run
    mteb.evaluate(
  File "/mnt/workspace/mteb/mteb/evaluate.py", line 363, in evaluate
    _res = evaluate(
  File "/mnt/workspace/mteb/mteb/evaluate.py", line 453, in evaluate
    result = _evaluate_task(
  File "/mnt/workspace/mteb/mteb/evaluate.py", line 162, in _evaluate_task
    task.load_data()
  File "/mnt/workspace/mteb/mteb/abstasks/retrieval.py", line 273, in load_data
    _process_data(split, lang)
  File "/mnt/workspace/mteb/mteb/abstasks/retrieval.py", line 262, in _process_data
    self.dataset[hf_subset][split] = RetrievalDatasetLoader(
  File "/mnt/workspace/mteb/mteb/abstasks/retrieval_dataset_loaders.py", line 74, in __init__
    self.dataset_configs = get_dataset_config_names(self.hf_repo, self.revision)
  File "/mnt/workspace/miniconda3/envs/mieb/lib/python3.10/site-packages/datasets/inspect.py", line 161, in get_dataset_config_names
    dataset_module = dataset_module_factory(
  File "/mnt/workspace/miniconda3/envs/mieb/lib/python3.10/site-packages/datasets/load.py", line 1030, in dataset_module_factory
    raise e1 from None
  File "/mnt/workspace/miniconda3/envs/mieb/lib/python3.10/site-packages/datasets/load.py", line 985, in dataset_module_factory
    raise DatasetNotFoundError(f"Dataset '{path}' doesn't exist on the Hub or cannot be accessed.") from e
datasets.exceptions.DatasetNotFoundError: Dataset 'mteb-private/Vidore3TelecomRetrieval' doesn't exist on the Hub or cannot be accessed.`

[...]

PR #3563, opened 2025-11-15
fix: add flag to run public only tasks

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3562

Added `allow-private` flag to cli and `public-only` argument to `evaluate`

### DIFF
--- a/tests/test_evaluate.py
+++ b/tests/test_evaluate.py
@@ -3,6 +3,7 @@ from copy import copy
 from pathlib import Path
 
 import pytest
+from datasets.exceptions import DatasetNotFoundError
 
 import mteb
 from mteb.abstasks.abstask import AbsTask
@@ -194,12 +195,17 @@ def test_evaluate_aggregated_task():
 def test_run_private_task_warning(caplog):
     """Test that a warning is correctly logged in an attempt run a private dataset is made"""
     task = mteb.get_task("Code1Retrieval")
+
+    def load_data_dataset_not_found():
+        raise DatasetNotFoundError
+
+    task.load_data = load_data_dataset_not_found
     model = mteb.get_model("baseline/random-encoder-baseline")
 
     with caplog.at_level(logging.WARNING):
         result = mteb.evaluate(model, task, cache=None)
         assert len(result.task_results) == 0
-        assert "Dataset for private task Code1Retrieval not found" in caplog.text
+        assert "Dataset for private task 'Code1Retrieval' not found" in caplog.text
 
 
 def test_run_private_task():


mteb/evaluate.py
```diff
@@ -267,6 +267,7 @@ def evaluate(
     overwrite_strategy: str | OverwriteStrategy = "only-missing",
     prediction_folder: Path | str | None = None,
     show_progress_bar: bool = True,
+    public_only: bool = True,
```
KennethEnevoldsen: Should we make it such that it runs private it we have permissions so like:

```
run_private: bool | None = None
```
If None, it tries to evaluate private, fails, and throws a warning - that way the user is also informed that there is something they can't do. It also prevents us from having to always set it. WDYT?

Samoed: Agree, but I'm not sure where to add. We're loading dataset in `_evaluate_task` https://github.com/embeddings-benchmark/mteb/blob/b22078ec818a8e2b7a89b028c97f60bb41e6d1db/mteb/evaluate.py#L162, but then it should return empty `TaskResult`, which is strange. If I move it to `evaluate`, then there will be huge or duplicated `try except`.

KennethEnevoldsen: Wouldn't a try except in `_evaluate_task` be doable:

```
try:
  task.load_data()
except datasets.exceptions.DatasetNotFoundError as e:
  if task.metadata.is_private and run_private is None:
    logger.warning(f"Attempting to run task '{task.metadata.name}', but the dataset ({task.metadata.dataset['path']}) is private. If you have access to the dataset make sure that you have specified your HF_TOKEN in your environmental variables. ")
  # ...

Comment (Samoed) [reply mteb/evaluate.py]: But what should be returned in that case? We had discussion about return in `_evaluate_task` https://github.com/embeddings-benchmark/mteb/pull/3351#discussion_r2428266540, and I'm not sure what we want to do there now

Comment (KennethEnevoldsen) [reply mteb/evaluate.py]: Yeah so didn't want to return an empty result object but we could return an or error object.

```
res = mteb.evaluate(model, task)
res.errors # shows the errors of tasks that failed to run

res.task_results # will remain unchanged
```

Where the object could look like:

```
class TaskError:
  task_name=...
  error=...
```

Then `_evaluate_task` will then return either a task_result or task_error. WDYT?

Comment (Samoed) [reply mteb/evaluate.py]: I've added it like this, but not sure if it is good idea

File: mteb/evaluate.py [hunk 2]
```diff
@@ -457,9 +477,18 @@ def evaluate(
...
         )
     logger.info(f"✓ Finished evaluation for {task.metadata.name}")
 
+    if isinstance(result, TaskError):
+        return ModelResult(
+            model_name=model_name,
+            model_revision=model_revision,
+            task_results=[],
+            errors=[result.error_message],
+        )
```
KennethEnevoldsen, line 492: is this correctly merged when running a list of tasks

mteb/evaluate.py
KennethEnevoldsen, line 1: We probably also need to add a task error in all the other errors

mteb/results/task_result.py
```diff
@@ -836,3 +836,15 @@ def get_hf_eval_results(self) -> list[EvalResult]:
...
+class TaskError(BaseModel):
+    """A class to represent an error that occurred during the evaluation of a task.
+
+    Attributes:
+        task_name: The name of the MTEB task.
+        error_message: The error message that occurred during the evaluation.
+    """
+
+    task_name: str
+    error_message: str
```
KennethEnevoldsen: shouldn't it just be the error?

mteb/evaluate.py
```diff
@@ -290,6 +306,7 @@ def evaluate(
         prediction_folder: Optional folder in which to save model predictions for the task. Predictions of the tasks will be sabed in `prediction_folder/{task_name}_predictions.json`
         show_progress_bar: Whether to show a progress bar when running the evaluation. Default is True. Setting this to False will also set the
             `encode_kwargs['show_progress_bar']` to False if encode_kwargs is unspecified.
+        public_only: Run only public tasks.
```
```suggestion
        public_only: Run only public tasks. If None, it will attempt to run the private task.
```

mteb/cli/build_cli.py
```diff
@@ -263,6 +264,12 @@ def _add_run_parser(subparsers: argparse._SubParsersAction) -> None:
         default=None,
         help="Folder to save the model predictions in. If None, predictions will not be saved.",
     )
+    parser.add_argument(
+        "--allow-private",
```
KennethEnevoldsen: Allow is kinda the default now - we could just remove it (since it is a matter of permissions)

but we could also add a explicit true false flag but then I think we need to rename it

mteb/evaluate.py
KennethEnevoldsen: I think I like it better, but let me what what parts you don't like

mteb/cli/build_cli.py
Samoed: I think it would be easier to remove

KennethEnevoldsen: let us remove it

mteb/results/task_result.py
Samoed: `Unable to generate pydantic-core schema for <class 'Exception'>` I think easier to leave as str

KennethEnevoldsen: but isn't it simply:

```
class Model(BaseModel):
    model_config = ConfigDict(arbitrary_types_allowed=True)
```

KennethEnevoldsen: Let us keep string

Samoed: Yes, I know, but users then can't dump model results directly, and we need to customize `model_dump_json` somehow then

KennethEnevoldsen: Let us keep string

tests/test_evaluate.py
```diff
@@ -187,3 +189,67 @@ def test_evaluate_aggregated_task():
...
+    task = MockRetrievalTask()
+    task_metadata = copy(task.metadata)
+    task_metadata.is_public = False
+    task.metadata = task_metadata
+    model = mteb.get_model("baseline/random-encoder-baseline")
+    results = mteb.evaluate(model, task, cache=None, public_only=False)
+    assert len(results.task_results) == 1
+
+
+def test_run_task_raise_error():
```
KennethEnevoldsen, line 222:
```suggestion
def test_run_task_raise_error():
    """Test that the error is not caught unintentionally"""
```

tests/test_evaluate.py
```diff
@@ -187,3 +189,67 @@ def test_evaluate_aggregated_task():
...
+    task = MockRetrievalTask()
+
+    def load_error():
+        raise RuntimeError("Test error")
+
+    task.load_data = load_error
+    model = mteb.get_model("baseline/random-encoder-baseline")
+    with pytest.raises(RuntimeError, match="Test error"):
+        mteb.evaluate(model, task, cache=None, public_only=False, raise_error=True)
```
KennethEnevoldsen:
```suggestion
        mteb.evaluate(model, task, cache=None)
```

tests/test_evaluate.py
```diff
@@ -187,3 +189,67 @@ def test_evaluate_aggregated_task():
...
+    task = mteb.get_task("Code1Retrieval")
+    model = mteb.get_model("baseline/random-encoder-baseline")
+
+    with caplog.at_level(logging.WARNING):
+        result = mteb.evaluate(model, task, cache=None)
+        assert len(result.task_results) == 0
+        assert "Dataset for private task Code1Retrieval not found" in caplog.text
+
+
+def test_run_private_task():
```
KennethEnevoldsen, line 211:
```suggestion
def test_run_private_task():
    """Tests that private task is run if it is possible to load the data"""
```

tests/test_evaluate.py
```diff
@@ -187,3 +189,67 @@ def test_evaluate_aggregated_task():
     model = mteb.get_model("baseline/random-encoder-baseline")
     task = MockAggregatedTask()
     mteb.evaluate(model, task, cache=None)
+
+
+def test_run_private_task_warning(caplog):
```
KennethEnevoldsen, line 195:
```suggestion
def test_run_private_task_warning(caplog):
    """Test that a warning is correctly logged in an attempt run a private dataset is made"""
```

tests/test_evaluate.py
```diff
@@ -187,3 +189,67 @@ def test_evaluate_aggregated_task():
...
+def test_run_task_supress_error():
+    task = MockRetrievalTask()
+
+    def load_error():
+        raise RuntimeError("Test error")
+
+    task.load_data = load_error
+    model = mteb.get_model("baseline/random-encoder-baseline")
+    results = mteb.evaluate(
+        model, task, cache=None, public_only=False, raise_error=False
```
KennethEnevoldsen:
```suggestion
        model, task, cache=None, raise_error=False
```

tests/test_evaluate.py
```diff
@@ -187,3 +189,67 @@ def test_evaluate_aggregated_task():
...
+    def load_error():
+        raise RuntimeError("Test error")
+
+    task.load_data = load_error
+    model = mteb.get_model("baseline/random-encoder-baseline")
+    with pytest.raises(RuntimeError, match="Test error"):
+        mteb.evaluate(model, task, cache=None, public_only=False, raise_error=True)
+
+
+def test_run_task_supress_error():
```
KennethEnevoldsen:
```suggestion
def test_run_task_supress_error():
    """Test that errors are correctly suppressed, when specified"""
```

tests/test_evaluate.py
```diff
@@ -187,3 +189,67 @@ def test_evaluate_aggregated_task():
...
+    task.load_data = load_error
+    model = mteb.get_model("baseline/random-encoder-baseline")
+    results = mteb.evaluate(
+        model, task, cache=None, public_only=False, raise_error=False
+    )
+    assert len(results.task_results) == 0
+    assert len(results.exceptions) == 1
+
+
+def test_run_list_with_error():
```
KennethEnevoldsen, line 235:
```suggestion
def test_run_list_with_error():
    """Test that errors are correctly suppressed, when specified"""
```
I would probably delete the previous task in favor of this one.

tests/test_evaluate.py
```diff
@@ -187,3 +189,67 @@ def test_evaluate_aggregated_task():
...
+
+    def load_error():
+        raise RuntimeError("Test error")
+
+    error_task.load_data = load_error
+    task = MockRetrievalTask()
+
+    model = mteb.get_model("baseline/random-encoder-baseline")
+    results = mteb.evaluate(
+        model, [error_task, task], cache=None, public_only=False, raise_error=False
```
KennethEnevoldsen:
```suggestion
        model, [error_task, task], cache=None, raise_error=False
```

mteb/evaluate.py
```diff
@@ -159,7 +163,20 @@ def _evaluate_task(
 
     data_loaded = task.data_loaded
     if not data_loaded:
-        task.load_data()
+        try:
+            task.load_data()
+        except DatasetNotFoundError as e:
+            if not task.metadata.is_public and public_only is None:
+                logger.warning(
+                    f"Dataset for private task {task.metadata.name} not found. "
```
```suggestion
                    f"Dataset for private task '{task.metadata.name}' not found. "
```

mteb/evaluate.py
```diff
@@ -159,7 +163,20 @@ def _evaluate_task(
...
+                    "Make sure you have access to the dataset and that you have set up the authentication correctly."
```
```suggestion
                    "Make sure you have access to the dataset and that you have set up the authentication correctly. To disable this warning set `public_only=False`"
```

More on a meta perspective, should we in general  `warnings.warn` our warnings as well?

Samoed: Agree

PR #3565, opened 2025-11-15
fix: Fix adapted from points to the models itself

KennethEnevoldsen:
This leads to the leaderboard current failing to build due to infinite recursion

Also added a test to prevent it happening going forward.


KennethEnevoldsen:
Enabling auto merge here as it causes a leaderboard crash

Samoed:
> Also added a test to prevent it happening going forward.

There is no test in pr

PR #3566, opened 2025-11-15
tests: Added test for ensuring training datasets can be computed 

KennethEnevoldsen:
missing in #3565

### DIFF
--- a/tests/test_models/test_model_meta.py
+++ b/tests/test_models/test_model_meta.py
@@ -118,8 +118,10 @@ def test_check_model_name_and_revision(model_meta: ModelMeta):
 
 @pytest.mark.parametrize("model_meta", mteb.get_model_metas())
 def test_check_training_datasets_can_be_derived(model_meta: ModelMeta):
-    # E.g. if a model if adapted_from is set to the model itself, this would cause infinite recursion. This ensures that this attribute can be called 
+    # E.g. if a model if adapted_from is set to the model itself, this would cause infinite recursion. This ensures that this attribute can be called
     # without issues.
     # https://github.com/embeddings-benchmark/mteb/pull/3565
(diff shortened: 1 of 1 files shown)


tests/test_models/test_model_meta.py
```diff
@@ -114,3 +114,11 @@ def test_model_memory_usage_api_model():
 def test_check_model_name_and_revision(model_meta: ModelMeta):
     assert model_meta.name is not None
     assert model_meta.revision is not None
+
+
+@pytest.mark.parametrize("model_meta", mteb.get_model_metas())
+def test_check_training_datasets_can_be_derived(model_meta: ModelMeta):
```
Samoed, line 120: Should we add pre-check like `assert model_meta.name != model_meta.adapted_from`

KennethEnevoldsen: Yea, but I thought this was more general (it will also catch this case, but also author cases)

Samoed: Yes, I just thought about like
```python
assert model_meta.name != model_meta.adapted_from, f"Model name and adapter model should be different. Got {model_meta.name} and {model_meta.adapted_from}"
model_meta.get_training_datasets()
```
to speed up a bit

tests/test_models/test_model_meta.py
```diff
@@ -114,3 +114,11 @@ def test_model_memory_usage_api_model():
...
+    # E.g. if a model if adapted_from is set to the model itself, this would cause infinite recursion. This ensures that this attribute can be called 
+    # without issues.
+    # https://github.com/embeddings-benchmark/mteb/pull/3565
+    model_meta.get_training_datasets()
```
```suggestion
    assert model_meta.name != model_meta.adapted_from, f"Model name and adapter model should be different. Got {model_meta.name} and {model_meta.adapted_from}"
model_meta.get_training_datasets()
```

PR #3567, opened 2025-11-15
fix: Set default input_type for VoyageMultiModalModelWrapper

KennethEnevoldsen:
if Input_type is None it leads to an error

### DIFF
--- a/mteb/models/model_implementations/voyage_v.py
+++ b/mteb/models/model_implementations/voyage_v.py
@@ -149,7 +149,7 @@ def voyage_v_loader(model_name, **kwargs):
             show_progress_bar: bool = True,
             **kwargs: Any,
         ) -> Array:
-            input_type = "document" # default
+            input_type = "document"  # default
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen:
@fzoll, @fzowl just double checking that "document" this is the desired default?

Samoed (approved):
I think this is correct, because in text wrapper document is default https://github.com/embeddings-benchmark/mteb/blob/78249e51fb6162802f2300776dd26d4d02763156/mteb/models/model_implementations/voyage_models.py#L135

mteb/models/model_implementations/voyage_v.py
```diff
@@ -149,6 +149,7 @@ def encode(
             show_progress_bar: bool = True,
             **kwargs: Any,
         ) -> Array:
+            input_type = "document" # default
```
Samoed:
```suggestion
            input_type = "document"  # default
```

Issue #3551, opened 2025-11-12
Prompt asymmetry in PairClf and STS


artemsnegirev:
### Description of the feature

Add support to prompt PairClassification and STS tasks in both sides (query/document). We have some NLI tasks (TERRa) which are asymmetric (premise, hypothesis). We can treat PairClassification as a pairwise binary task and STS as a general pairwise relevance assessment task, but add support for asymmetry. For example, we need to evaluate retrieval query/document relevance pairwise.

KennethEnevoldsen:
Makes sense. I see it as a relevant addition - Just to be upfront about it, I suspect that we will not prioritize this, but I would be more than happy to review a feature on it. It can likely be modelled based on how it is done for retrieval. The change would have to be backward compatible, which is probably the biggest challenge.

Samoed:
I think this can be easily added in backward compatible manner, as currently implemented for retrieval (when prompt for task/model is string, just return it). Do we want to expand `PromptType`? Should we name this like `input1`, `input2`?

Also, currently pair classification combines both inputs into one list for text, but I think we remove this combination 
https://github.com/embeddings-benchmark/mteb/blob/fe83e27968adff648de9520e187164ea5f906935/mteb/_evaluators/pair_classification_evaluator.py#L88-L100

artemsnegirev:
> Should we name this like `input1`, `input2`?

It's common to use query/document notation for any asymmetric task in IR. So I don't think we need this.

Samoed:
Yes, but in Pair classification and STS tasks "query" can be in both columns and this would be hard to pass correctly everywhere

Samoed:
Or alternatively we can specify which of columns is query and which is doc for these datasets, but this would require more work for annotation

KennethEnevoldsen:
We can also just default it to first is query, but it seems wierd to use query/document as it suggests that the task is assymetric (we could add annotation on task symmetri)

Samoed:
We can add to these tasks like
```python
class AbsTaskSTS:
    input1_prompt_type: PromptType | None = None
    input2_prompt_type: PromptType | None = None
    ...

class STSEvaluator:
    def __init__(..., input1_prompt_type, input2_prompt_type):
       ...
       self.input1_prompt_type = input1_prompt_type
       self.input2_prompt_type = input2_prompt_type

    def __call__(...):
        embeddings1 = model.encode(loader, prompt_type=self.input1_prompt_type, **encode_kwargs)
[...]

PR #3568, opened 2025-11-15
feat: make STS and PairClassification asymmetric

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3551

Make `STS` and `PairClassification` support `prompt_type` for different columns.

But this is not backward compatible in scores for models with `query`/`document` separation. Example for `intfloat/multilingual-e5-small`

| Task | main | PR |
| - | - | - |
| TERRa | 0.575105 | 0.589083 |

### DIFF
--- a/mteb/descriptive_stats/PairClassification/TERRa.V2.json
+++ b/mteb/descriptive_stats/PairClassification/TERRa.V2.json
@@ -1,0 +1,35 @@
+{
+    "dev": {
+        "num_samples": 307,
+        "number_of_characters": 84848,
+        "unique_pairs": 307,
+        "text1_statistics": {
+            "total_text_length": 70844,
+            "min_text_length": 39,
+            "average_text_length": 230.76221498371336,
+            "max_text_length": 717,
+            "unique_texts": 282
+        },
+        "text2_statistics": {
+            "total_text_length": 14004,
+            "min_text_length": 12,
+            "average_text_length": 45.615635179153095,
+            "max_text_length": 129,
+            "unique_texts": 307
+        },
+        "labels_statistics": {
+            "min_labels_per_text": 1,
+            "average_label_per_text": 1.0,
+            "max_labels_per_text": 1,
+            "unique_labels": 2,
+            "labels": {
+                "1": {
+                    "count": 153
+                },
+                "0": {
+                    "count": 154
+                }
+            }
+        }
+    }
+}
--- a/tests/test_evaluators/test_PairClassificationEvaluator.py
+++ b/tests/test_evaluators/test_PairClassificationEvaluator.py
@@ -21,6 +21,8 @@ class TestPairClassificationEvaluator:
             MockClassificationTask.metadata,
             "test",
             "test",
+            input1_prompt_type=None,
+            input2_prompt_type=None,
         )
         distances = evaluator(
             mteb.get_model("baseline/random-encoder-baseline"),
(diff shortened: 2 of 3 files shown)


mteb/abstasks/task_metadata.py
```diff
@@ -286,6 +294,7 @@ class TaskMetadata(BaseModel):
     adapted_from: Sequence[str] | None = None
     is_public: bool = True
     superseded_by: str | None = None
+    asymmetric_type: AsymmetricTypes | None = None
```
Samoed: Probably we can just change to `bool | None` and use `input1_prompt_type` to use choose `prompt_type`

mteb/abstasks/task_metadata.py
```diff
@@ -255,6 +262,7 @@ class TaskMetadata(BaseModel):
         is_public: Whether the dataset is publicly available. If False (closed/private), a HuggingFace token is required to run the datasets.
         superseded_by: Denotes the task that this task is superseded by. Used to issue warning to users of outdated datasets, while maintaining
             reproducibility of existing benchmarks.
+        asymmetric_type: Type of asymmetric. If `None`, then it symmetric. For PairClassification and STS tasks.
```
KennethEnevoldsen: Unsure if we want this to be just "symmetric" / "assymetric" - regardless I wouldn't want it to be None.

I would probably not include it in this PR for now

mteb/tasks/pair_classification/rus/terra.py
```diff
@@ -43,6 +47,7 @@ class TERRa(AbsTaskPairClassification):
 }
 """,
         prompt="Given a premise, retrieve a hypothesis that is entailed by the premise",
```
KennethEnevoldsen, line 49: Does this prompt still make sense if you have two prompt types?

mteb/tasks/pair_classification/rus/terra.py
```diff
@@ -1,8 +1,12 @@
 from mteb.abstasks.pair_classification import AbsTaskPairClassification
 from mteb.abstasks.task_metadata import TaskMetadata
+from mteb.types import PromptType
 
 
 class TERRa(AbsTaskPairClassification):
+    input1_prompt_type = PromptType.document
```
KennethEnevoldsen: This will break backward compat - any way around creating a v2?

Samoed: Agree. I will create other version for compatability

mteb/abstasks/task_metadata.py
Samoed: I think we need to add some to metadata still. I can convert this to `symetric: bool | None` and None for tasks that don't have this annotation for now

KennethEnevoldsen: Makes sense, but do we want to add empty metadata?

Samoed: Generally not, but we can fill it in all tasks

Samoed:
I'm not sure what to do with retrieval. Do we want to add `query_input_type`/`corpus_input_type` or model author's can explicitly add prompts (or we can do this based on metadata)?

KennethEnevoldsen:
> I'm not sure what to do with retrieval. Do we want to add query_input_type/corpus_input_type or model author's can explicitly add prompts (or we can do this based on metadata)?

I am not sure either - should we get an opinion from e.g. Orion? (in an issue)

Samoed:
I think we can create make similar for retrieval in other PR

PR #3570, opened 2025-11-16
[MAEB] Merge `AudioReranking` with `Retrieval`

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3418

Results for `MIT/ast-finetuned-audioset-10-10-0.4593`

| | MAEB | PR |
| - | - | - |
| AudioCapsMiniReranking | 0.74125 | 0.99875 |

I don't know why there is such difference in scores, but I think we can change them because it seems that previously scores were computed incorrectly (`num_neg` and `num_pos` are always 1 and `doc_idx` is changed by 2 only) https://github.com/embeddings-benchmark/mteb/blob/603f3e8e09a6e1961e089c315a492a30880e8807/mteb/evaluation/evaluators/audio/audio_reranking_evaluator.py#L139-L158


Samoed:
`numba` with `3.14` support will be released soon https://github.com/numba/numba/issues/9957. Currently, they have `3.14` support in `0.63.0b1` beta version

KennethEnevoldsen (commented):
A score of 99 seems very high to me - won't this make the task useless if even current model max out?

Samoed:
Yes, but I don't think this is related to the implementation of retrieval/reranking task

Vidore v2 also have a lot of tasks with 0.9+

KennethEnevoldsen:
Right - we can def. add it, but I don't think it makes for a good task in the leaderboard

Samoed:
Can this PR be merged?

KennethEnevoldsen:
yes

Issue #3507, opened 2025-10-28
Make `Qwen2-Audio-7B` support text


Samoed:
### Description of the feature

Now it's support only images

PR #3581, opened 2025-11-17
[MAEB] Make `Qwen2-Audio` support text

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3507

### DIFF
--- a/mteb/models/model_implementations/audio_qwen2_models.py
+++ b/mteb/models/model_implementations/audio_qwen2_models.py
@@ -116,7 +116,7 @@ class Qwen2AudioWrapper(AbsEncoder):
                 batch_indices = torch.arange(hidden.size(0), device=self.device)
                 embeddings = hidden[batch_indices, last_idx]
 
-                all_embeddings.append(embeddings.cpu())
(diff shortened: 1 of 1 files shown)


AdnanElAssadi:
Hi @Samoed

Nice work, looks good overall! However, I'd like to double-check one thing before merging:
Can you test this with both audio-only and audio+text tasks? I want to verify that the way we're fusing audio and text (by concatenating text after the audio tokens and processing everything through the audio encoder) actually works as intended for:

Audio-to-text (a2t) retrieval tasks
Text-to-audio (t2a) retrieval tasks

My concern is that since we're only using self.audio_encoder for all inputs, text-only queries in t2a tasks might not encode properly, I am not sure if the model is supposed to work this way (audio encoder handles everything).

Samoed:
Yes, I see. I'll change to extract embeddings from full model instead, rather then only from audio part

AdnanElAssadi:
Looks good to me, maybe a sanity check evaluation and then we merge.

Samoed:
I've run it and it's working

PR #3587, opened 2025-11-19
add prompt dict for Jasper_Token_Compression_600M

DunZhang:
@Samoed 
Hello, as you suggested, I have added a prompt_dict. Actual testing shows that the Chinese and English MTEB results are consistent with those already submitted.

There are changes in the MTEB (Law, v1) results, and I will update the PR(https://github.com/embeddings-benchmark/results/pull/325) for the results.

Here is the reproduction code:
```
import mteb

if __name__ == "__main__":
    benchmark_name = "MTEB(cmn, v1)"  # MTEB(eng, v2), MTEB(cmn, v1), MTEB(Law, v1),
    model = mteb.get_model("infgrad/Jasper-Token-Compression-600M")
    mteb.MTEB(tasks=list(mteb.get_benchmark(benchmark_name).tasks)).run(model)

```

### DIFF
--- a/mteb/models/model_implementations/jasper_models.py
+++ b/mteb/models/model_implementations/jasper_models.py
@@ -328,15 +328,8 @@ jasper_en_v1 = ModelMeta(
 )
 
 
-class JasperTokenCompressionLoader(InstructSentenceTransformerModel):
-    def __init__(self, model_name: str, revision: str, **kwargs: Any):
-        super().__init__(model_name, revision, **kwargs)
-        if isinstance(kwargs.get("max_seq_length"), int):
-            self.model.max_seq_length = kwargs["max_seq_length"]
-
-
 Jasper_Token_Compression_600M = ModelMeta(
-    loader=JasperTokenCompressionLoader,
+    loader=InstructSentenceTransformerModel,
     loader_kwargs=jasper_token_compression_600m_loader_kwargs,
     name="infgrad/Jasper-Token-Compression-600M",
     languages=["eng-Latn", "zho-Hans"],


mteb/models/model_implementations/jasper_models.py
```diff
@@ -135,13 +327,17 @@ def encode(
 """,
 )
 
+
+class JasperTokenCompressionLoader(InstructSentenceTransformerModel):
+    def __init__(self, model_name: str, revision: str, **kwargs: Any):
+        super().__init__(model_name, revision, **kwargs)
+        if isinstance(kwargs.get("max_seq_length"), int):
+            self.model.max_seq_length = kwargs["max_seq_length"]
```
Samoed:
```suggestion
```
`max_seq_length` will be passed to tokenizer params. https://github.com/embeddings-benchmark/mteb/blob/76be9594308a4427fef461d30b499268fb14532a/mteb/models/instruct_wrapper.py#L146-L148

Or you can add this to your tokenizer_config if this is real max len of your model

mteb/models/model_implementations/jasper_models.py
```diff
@@ -21,6 +21,198 @@
...
+    "TwitterURLCorpus": "Retrieve tweets that are semantically similar to the given tweet",
+    "VideoRetrieval": {
+        "query": "Given a video search query, retrieve the titles of relevant videos",
+        "document": "",
+    },
+    "Waimai": "Classify the customer review from a food takeaway platform into positive or negative",
+}
+jasper_token_compression_600m_loader_kwargs = dict(
+    model_kwargs={
+        "attn_implementation": "flash_attention_2",
```
Samoed: Do you require `flash_attention_2`? I tried to run multiple times with it and sometimes this is hard to install

mteb/models/model_implementations/jasper_models.py
```diff
@@ -21,6 +21,198 @@
...
+        "torch_dtype": "bfloat16",
+        "trust_remote_code": True,
+    },
+    tokenizer_kwargs={"padding_side": "left"},
```
Samoed, line 208: You can set this directly in `tokenizer_config` like https://huggingface.co/microsoft/Phi-3-mini-4k-instruct/blob/main/tokenizer_config.json#L125

mteb/models/model_implementations/jasper_models.py
```diff
@@ -21,6 +21,198 @@
...
+    "VideoRetrieval": {
+        "query": "Given a video search query, retrieve the titles of relevant videos",
+        "document": "",
+    },
+    "Waimai": "Classify the customer review from a food takeaway platform into positive or negative",
+}
+jasper_token_compression_600m_loader_kwargs = dict(
+    model_kwargs={
+        "attn_implementation": "flash_attention_2",
+        "torch_dtype": "bfloat16",
```
Samoed, line 205: Also, I think this is probably isn't required, because your model is already in bfloat16

DunZhang: That makes sense, perhaps `SDPA` has better compatibility.

mteb/models/model_implementations/jasper_models.py
DunZhang: /(ㄒoㄒ)/~~
This is a frustrating issue. The model itself supports 32k length, but was only trained on 1024 token data, so I believe the model supports 32k, just with suboptimal performance.

When reproducing results, we load it as 1024 for efficiency. Modifying `tokenizer_config.json` feels inappropriate.

Additionally, I noticed in the source code: `tokenizer_params["model_max_length"] = max_seq_length`, but unfortunately, after repeated debugging, I found it doesn't take effect. I believe this is a bug in SentenceTransformer. Here's the minimal reproduction code:

```python
    m = SentenceTransformer("intfloat/e5-small",tokenizer_kwargs={"model_max_length":3})
    print(m.tokenize(["hi hi hi hi hi hi hi hi hi hi hi hi hi"]))
    # print result
    """
    {'input_ids': tensor([[ 101, 7632, 7632, 7632, 7632, 7632, 7632, 7632, 7632, 7632, 7632, 7632,
         7632, 7632,  102]]), 'token_type_ids': tensor([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]]), 'attention_mask': tensor([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]])}
    """
```

When debugging, you'll notice that `features` isn't truncated at all.

Therefore, I had to write the following assignment statement: `self.model.max_seq_length = kwargs["max_seq_length"]`

Samoed: Hm, interesting
```python
m = SentenceTransformer("intfloat/e5-small",tokenizer_kwargs={"model_max_length":3})
m[0].tokenizer.model_max_length
# 3
print(m.tokenize(["hi hi hi hi hi hi hi hi hi hi hi hi hi"]))
# {'input_ids': tensor([[ 101, 7632, 7632, 7632, 7632, 7632, 7632, 7632, 7632, 7632, 7632, 7632,
```
I tried to debug this a bit and for `e5` it will pass `max_seq_length` from their `sentence_bert_config.json` https://huggingface.co/intfloat/e5-small/blob/main/sentence_bert_config.json#L2

In your case it will inherit 32768 https://huggingface.co/infgrad/Jasper-Token-Compression-600M/blob/main/sentence_bert_config.json#L2 (padding side too, but generally I think it should be in tokenizer config)

Samoed: Created an issue https://github.com/huggingface/sentence-transformers/issues/3575

Samoed: Now `max_seq_len` will be used correctly https://github.com/embeddings-benchmark/mteb/pull/3588

mteb/models/model_implementations/jasper_models.py
DunZhang: Cool, I already remove `JasperTokenCompressionLoader`, now this model can be loaded by `InstructSentenceTransformerModel`

mteb/models/model_implementations/jasper_models.py
DunZhang: ok. code is updated

Samoed (approved):

PR #3580, opened 2025-11-17
[MAEB] merge zeroshot classification

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3417

Results for `laion/larger_clap_music_and_speech`
| | MAEB | PR |
|-|-|-|
| ESC50_Zeroshot | 0.8485 | 0.8485 |

PR #3590, opened 2025-11-19
[MAEB] Merge classification

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3415 (reopened https://github.com/embeddings-benchmark/mteb/pull/3571)

Combined `audio classification` with `AbsTaskClassification`. Now we need to compute embeddings only once for `cross-validation` instead of recomputing on each fold.

Restuls of evaluation for `MIT/ast-finetuned-audioset-10-10-0.4593`

| | MAEB | PR |
|-|-|-|
| BeijingOpera | 0.970301 | 0.95754 |

The results are different, because previous implementation was ignoring task's `samples_per_label` parameter (because it was set in `__init__`). If I change to 16 as in `maeb` I'm getting same results (0.97)

### DIFF
--- a/mteb/tasks/audio/audio_classification/eng/ambient_acoustic_context.py
+++ b/mteb/tasks/audio/audio_classification/eng/ambient_acoustic_context.py
@@ -46,5 +46,5 @@ class AmbientAcousticContextClassification(AbsTaskClassification):
 
     input_column_name: str = "audio"
     label_column_name: str = "label"
-    samples_per_label: int = None  # Not needed as dataset is already balanced
+
     is_cross_validation: bool = False
--- a/mteb/tasks/audio/audio_classification/eng/beijing_opera.py
+++ b/mteb/tasks/audio/audio_classification/eng/beijing_opera.py
@@ -41,5 +41,5 @@ class BeijingOpera(AbsTaskClassification):
 
     input_column_name: str = "audio"
     label_column_name: str = "label"
-    samples_per_label: int = 10
+
     is_cross_validation: bool = True
--- a/mteb/tasks/audio/audio_classification/eng/bird_clef.py
+++ b/mteb/tasks/audio/audio_classification/eng/bird_clef.py
@@ -39,5 +39,5 @@ class BirdCLEFClassification(AbsTaskClassification):
 
     input_column_name: str = "recording"
     label_column_name: str = "primary_label"
-    samples_per_label: int = 20
+
     is_cross_validation: bool = True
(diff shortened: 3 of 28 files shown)


mteb/abstasks/classification.py
```diff
@@ -200,30 +221,99 @@ def _evaluate_subset(
...
+        )
+        logger.info("Running cross-validation - Encoding samples...")
+        # precompute all embeddings for cross-validation to not recomupute them in different k-folds
+        dataset_embeddings = model.encode(
+            dataloader_train,
+            task_metadata=self.metadata,
+            hf_split=hf_split,
+            hf_subset=hf_subset,
+            **encode_kwargs,
+        )
```
KennethEnevoldsen, line 283: don't we "over-embed" here? If we only need 16 samples per (I suppose we always have to embed the test so probably not

mteb/abstasks/classification.py
```diff
@@ -121,6 +135,9 @@ class AbsTaskClassification(AbsTask):
     label_column_name: str = "label"
     input_column_name: str = "text"
     abstask_prompt = "Classify user passages."
+    is_cross_validation: bool = False
+    n_splits = 5
+    cross_validation_splitter: type[KFoldSplitterProtocol] = KFold
```
KennethEnevoldsen: do we want to expose this?

mteb/abstasks/classification.py
```diff
@@ -92,6 +96,13 @@ class FullClassificationMetrics(ClassificationMetrics):
     scores_per_experiment: list[ClassificationMetrics]
 
 
+class KFoldSplitterProtocol(Protocol):  # noqa: D101
```
KennethEnevoldsen: is this a defined interface in sklearn?

Samoed:
> tests seems to fail

Cache problem https://github.com/embeddings-benchmark/mteb/pull/3580#issuecomment-3554465609

mteb/abstasks/classification.py
Samoed: Sklearn don't have interface directly, but I tried to copy same signatures as in Kfold https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html

mteb/abstasks/classification.py
Samoed: I think we can remove. I just wanted to make it as `evaluator_model`, but we can change this later

mteb/abstasks/classification.py
Samoed: > I suppose we always have to embed the test so probably not

Yes

mteb/abstasks/classification.py
KennethEnevoldsen: maybe we don't need a protocol then?

Samoed: Yes, I will remove splitter for now

KennethEnevoldsen (approved):

PR #3603, opened 2025-11-22
tests: Add tests for dataset quality

### LINKED ISSUE #375: Tracking down dataset duplication
In #354 it was discussed that some datasets might have duplicate segments. This issue is to start the discussion on how to best find and identify these.

### LINKED ISSUE #3370: Make test for duplicated texts
There seems to be duplicates in the datasets as well. Let us remove these before we merge (should we make a test for this on new tasks?)

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/pull/3367#pullrequestreview-3339018453_

KennethEnevoldsen:
This test ensures a minimum quality of task for future submissions

Currently, it tests for:
- text duplicates
- train test leakage
- too short documents

I suspect we can likely add many more tests to this in future PRs

I would argue that this fixes #375 (by preventing duplicates it in future datasets)

fixes #375
fixes #3370

### DIFF
--- a/tests/test_tasks/test_task_quality.py
+++ b/tests/test_tasks/test_task_quality.py
@@ -332,7 +332,7 @@ def _task_quality(task: AbsTask) -> list[str]:
 
 
 def test_dataset_quality():
-    tasks = get_tasks(exclude_superseded=False, exclude_aggregate=False)
+    tasks = get_tasks(exclude_superseded=False, exclude_aggregate=True)
 
     errors: list[str] = []
     for task in tasks:


tests/test_tasks/test_task_quality.py
```diff
@@ -0,0 +1,344 @@
...
+        number_texts_intersect_with_train is None
+        or number_texts_intersect_with_train == 0
+    ):
+        errors.append(
+            f"{name} ({split}) has an overlap between train and test ({number_texts_intersect_with_train=})"
+        )
+    return errors
+
+
+def _task_quality(task: AbsTask) -> list[str]:
```
KennethEnevoldsen, line 322: We could alternatively implement this as a method on the dataset itself, so that users could check dataset quality (though I am not sure how it would look).

tests/test_tasks/test_task_quality.py
```diff
@@ -0,0 +1,344 @@
...
+    desc_stats = task.metadata.descriptive_stats
+
+    errors = []
+    if desc_stats is None:
+        return []
```
Samoed, line 327: Should this be a failed test?

tests/test_tasks/test_task_quality.py
```diff
@@ -0,0 +1,344 @@
...
+    for split_name, split_stats in desc_stats.items():
+        errors += _split_quality(task.metadata.name, split_name, split_stats)
+
+    return errors
+
+
+def test_dataset_quality():
+    tasks = get_tasks(exclude_superseded=False, exclude_aggregate=False)
```
Samoed: Should we skip aggregated?
```suggestion
    tasks = get_tasks(exclude_superseded=False, exclude_aggregate=True)
```

tests/test_tasks/test_task_quality.py
Samoed: I think this is good for now as test

tests/test_tasks/test_task_quality.py
KennethEnevoldsen: there is a different test that test if desc stats is filled

PR #3613, opened 2025-11-25
model: Add IEITYuan/Yuan-embedding-2.0-zh model

wangxj12:

### DIFF
--- a/mteb/models/model_implementations/yuan_models.py
+++ b/mteb/models/model_implementations/yuan_models.py
@@ -31,4 +31,3 @@ yuan_embedding_2_zh = ModelMeta(
     public_training_data=None,
     training_datasets=yuan_emb_zh_datasets,
 )
-


mteb/models/model_implementations/yuan_models.py
```diff
@@ -0,0 +1,33 @@
...
+
+yuan_embedding_2_zh = ModelMeta(
+    name="IEITYuan/Yuan-embedding-2.0-zh",
+    loader=sentence_transformers_loader,
+    languages=["zho-Hans"],
+    open_weights=True,
+    revision="b5ebcace6f4fc6e5a4d1852557eb2dc2d1040cee",
+    release_date="2025-11-24",
+    n_parameters=326000000,
+    memory_usage_mb=None,
```
Samoed:
```suggestion
    memory_usage_mb=1242,
```

wangxj12: Thanks for pointing that out. I’ve updated memory_usage_mb accordingly. Please let me know if any further adjustments are needed.

KennethEnevoldsen (approved):

Issue #3611, opened 2025-11-24
Language filtering


antoineedy:
### Description of the feature

Hi MTEB team!

The language filtering seems to be working the following way: when selecting a language, if a task contains it, it is displayed.

This behaviour is not really compatible with the way we defined our tasks in ViDoRe V3, and I wonder if we did something wrong, or if this behaviour should be changed. For context:
- ViDoRe V3 contains 10 datasets (=10 different tasks)
- Each of them have ~300 english queries, which are translated in 5 other languages (so a total of ~1500 query per dataset). 

Thus, when we select only "english" in the language filtering area for example, nothing changes as all tasks have at least some english queries. Therefore, we have no way of displaying results for english-only queries, if I'm not mistaken. Is it the intended way? I see two ideas to clarify this filtering:
- We change this behaviour to only include the results of the chosen languages,
- We keep this behaviour of filtering out the tasks that don't contain the specified languages, but we add a tab to display per-language results (after the per-task for example).

What do you think?

PS: I am working on a demo-space with this new tab, should be available in an hour or so.

Samoed:
Results should be filtered out and only english subset should be left. But generally I think it would be better to show per-language resutls

Samoed:
This seems a bug in filtering somewhere (probably in `_get_scores`). 
```python
import mteb
from mteb.cache import ResultCache

cache = ResultCache()
cache.download_from_remote()

b = mteb.get_benchmark("ViDoRe(v3)")
vidore3_full_res = cache.load_results(tasks=b.tasks)
print(vidore3_full_res._get_scores())
new_tasks = []
for t in b.tasks:
    new_t = t.filter_languages(["eng"])
    new_tasks.append(new_t)
vidore3_eng_res = cache.load_results(tasks=new_tasks)
print(vidore3_eng_res._get_scores())
```
Results are the same


UPD: If I pass `cache.load_results(tasks=new_tasks, validate_and_filter=True)` then results would be filtered out

Samoed:
It seems that during filtering tasks, we're just returning task names instead of `AbsTask`/`TaskMetadta` and because of that nothing is changing. We need to refactor leaderboard code a lot
https://github.com/embeddings-benchmark/mteb/blob/8212389b734ddb01a86921479df624c4d6dad003/mteb/leaderboard/app.py#L565-L606

KennethEnevoldsen:
[...]

PR #3617, opened 2025-11-25
fix: Add optional "per language table"

### LINKED ISSUE #3554: Add language columns to MTEB(Scandinavian, v1)

Would be great with columns like: Mean(dan), Mean(nno), Mean(swe)...

antoineedy:
Hi! Following #3611, I propose a new **optional** tab to display results per language.

My changes:
- New `Benchmark` attribute: `supports_language_view` (`bool`, defaults to `False`),
- New table on the leaderboard (tab: Performance per language),
- Quite a big change: this tab only appears when the task displayed has `supports_language_view` to True. I think it makes sense to customize a bit the UI depending on the specificities of the benchmarks. It can make sense for other tabs, like "Performance per task type", to not be displayed under certain conditions instead of plotting "_Couldn't produce plot. Reason: Couldn't produce radar chart, the benchmark only contains one task category._".

Here is a demo leaderboard for you to have a look. I only add `supports_language_view=True` to ViDoRe V3, so the tab will only appear in this one: https://huggingface.co/spaces/antoineedy/demo-leaderboard.

Tell me what you think!

---
edit by @KennethEnevoldsen:

fixes #3554

### DIFF
--- a/mteb/leaderboard/table.py
+++ b/mteb/leaderboard/table.py
@@ -262,10 +262,16 @@ def _apply_per_task_table_styling(per_task: pd.DataFrame) -> gr.DataFrame:
         "{:.2f}", subset=task_score_columns, na_rep=""
     ).highlight_max(subset=task_score_columns, props="font-weight: bold")
 
+    # setting task name column width to 250px
+    column_widths = _get_column_widths(per_task_style.data)
+    if len(column_widths) > 0:
+        column_widths[0] = "250px"
+
     return gr.DataFrame(
         per_task_style,
         interactive=False,
         pinned_columns=1,
+        column_widths=column_widths,
         buttons=["copy", "fullscreen"],
         show_search="filter",
     )
@@ -287,10 +293,16 @@ def _apply_per_language_table_styling(per_language: pd.DataFrame) -> gr.DataFram
             "{:.2f}", subset=language_score_columns, na_rep=""
         ).highlight_max(subset=language_score_columns, props="font-weight: bold")
 
+    # setting task name column width to 250px
+    column_widths = _get_column_widths(per_language_style.data)
+    if len(column_widths) > 0:
+        column_widths[0] = "250px"
+
     return gr.DataFrame(
         per_language_style,
         interactive=False,
         pinned_columns=1,
+        column_widths=column_widths,
         buttons=["copy", "fullscreen"],
         show_search="filter",
     )


Samoed:
But don't change, wait for second opinion, just in case

antoineedy:
Alright 😁 I understand it is quite a "big" change, if this is a direction you want to take I don't mind spending some extra time on doing it the way you want

mteb/benchmarks/_create_table.py
```diff
@@ -250,6 +250,57 @@ def _create_per_task_table_from_benchmark_results(
...
+
+    if data.empty:
+        no_results_frame = pd.DataFrame(
+            {"No results": ["You can try relaxing your criteria"]}
+        )
+        return no_results_frame
+
+    per_language = data.pivot_table(
+        index="model_name", columns="subset", values="score", aggfunc="mean"
+    )
```
KennethEnevoldsen, line 275: Right now, it seems to be really displaying subsets and not language (which is not the same)

E.g., a subset could be `"en-fr"`, but contain English and French (`["eng-Latn", "fra-Latn"]`).

To make this I would probably implement it as:
```
df = benchmark_results.to_performance_by_language_dataframe(languages=...)
# just a regular df, which we can then style
```

Which we can then test and validate without running the leaderboard.

Samoed: Maybe then it would be easier to add only to vidore?

antoineedy: I am working on a new version. I don't really understand what it means to have multiple languages in one result. Should I consider pairs "eng-Latn, fra-Latn" as one language? Or should I add the result to both english and french in that case?

antoineedy:
I created a new aggregation_level to the `to_dataframe `method: `benchmark_results.to_dataframe(aggregation_level="language")` and refactored my code around this. 

Also, I updated the Benchmark attribute to be `language_view`, a list of languages to show. By default, I set it to `[]`, and in this case the tab does not show.

wdyt?

mteb/benchmarks/_create_table.py
Samoed: I think you should use subsets only, but this would apply only to your case and probably won't work for other benchmarks

mteb/results/benchmark_results.py
```diff
@@ -346,6 +347,9 @@ def to_dataframe(
         # Create DataFrame
         df = pd.DataFrame(scores_data)
 
+        if aggregation_level == "language":
+            df["language"] = df["language"].apply(lambda langs: ", ".join(langs))
```
Samoed: I think this can be a strange. For example, if one task has `[eng, fra]` and other task in this split just `[eng]`, then there will 2 different columns

mteb/benchmarks/_create_table.py
KennethEnevoldsen: > Maybe then it would be easier to add only to vidore?

> I think you should use subsets only, but this would apply only to your case and probably won't work for other benchmarks

We can do it this way, but then it should just be an extension on the summary table not a seperate tab.

I would however rather has it as a seperate tab (since other benchmarks would love benefit as well)

> I am working on a new version. I don't really understand what it means to have multiple languages in one result. Should I consider pairs "eng-Latn, fra-Latn" as one language? Or should I add the result to both english and french in that case?

I would say the score for English includes all subsets that contain `eng-Latn`. I.e. add to both

mteb/results/model_result.py
```diff
@@ -43,6 +43,9 @@ def _aggregate_and_pivot(
     elif aggregation_level == "task":
         index_columns = ["task_name"]
 
+    elif aggregation_level == "language":
+        index_columns = ["language"]
```
KennethEnevoldsen, line 47: I am unsure how this would look on other benchmarks, I suspect there might be a few cases where it doesn't work very well. Can you enable it on 

- MTEB(Scandinavian, v1): easy case to generalize to fairly small and well-contained, but contains multilingual datasets
- MTEB(Multilingual,v2): Harder case with many datasets and notably bitext mining, which might be a challenge

antoineedy:
I updated my demo leaderboard with you suggestions (+ added MTEB(Scandinavian, v1) and MTEB(Multilingual,v2)): https://huggingface.co/spaces/antoineedy/demo-leaderboard

A couple of concerns:
- The `language_view` attribute of the `Benchmark` class now takes either a list of languages to display (default to `[]`, in this case the tab is not even displayed) OR the string "all", so that all the languages present in the tasks are displayed. I think this is pretty accommodating, andit leads to this definition: `language_view: list[str] | Literal["all"] = field(default_factory=list)`. I did not find a better way of implementing it, tell me if it looks fine to you! (The other way would be to get rid of "all", so people would have to specify each language individually in a list, and this would solve the next issue)
- For benchmarks with many many languages (such as MTEB(Multilingual,v2), 1064*324 cells), the table is therefore huge and causes UI performances issues in the space. Not only performance issues, but straight gradio errors like so when applying dataframe styling:

```
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/mteb/leaderboard/app.py", line 950, in <module>
app = get_leaderboard_app()
^^^^^^^^^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/mteb/leaderboard/app.py", line 263, in get_leaderboard_app
per_language_table = apply_per_language_styling_from_benchmark(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/mteb/leaderboard/table.py", line 145, in apply_per_language_styling_from_benchmark
return _apply_per_language_table_styling(per_language_df)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/mteb/leaderboard/table.py", line 274, in _apply_per_language_table_styling
return gr.DataFrame(
^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/.venv/lib/python3.12/site-packages/gradio/component_meta.py", line 185, in wrapper
return fn(self, **kwargs)
^^^^^^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/.venv/lib/python3.12/site-packages/gradio/components/dataframe.py", line 201, in init
super().init(
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/.venv/lib/python3.12/site-packages/gradio/component_meta.py", line 185, in wrapper
return fn(self, **kwargs)
^^^^^^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/.venv/lib/python3.12/site-packages/gradio/components/base.py", line 224, in init
initial_value = self.postprocess(initial_value)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/.venv/lib/python3.12/site-packages/gradio/components/dataframe.py", line 457, in postprocess
metadata = self.get_metadata(value)
^^^^^^^^^^^^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/.venv/lib/python3.12/site-packages/gradio/components/dataframe.py", line 407, in get_metadata
return Dataframe.__extract_metadata(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/Users/antoineedy/Documents/recherche/nvidore/mteb-pipeline/.venv/lib/python3.12/site-packages/gradio/components/dataframe.py", line 492, in __extract_metadata
row_styling.append(style_dict.get(cell["id"], ""))
~~~~^^^^^^
KeyError: 'id'
make: *** [run-leaderboard] Error 1
```

In the current version of this PR, I thus disabled styling for large tables, but there are still performance issues.

In my opinion, the best way to tackle this is to accept that gradio has limits to dataframe display, and to keep this in mind when choosing for which benchmark to display this new tab.

Lastly, I did not change yet the language filtering in _Customize this benchmark_. For now, the language chosen only impact the tasks, as mentioned in the first message of this PR. Do we agree that it should also impact the columns shown in this new tab?

antoineedy:
(Note that my demo space if running on basic CPU, yours on CPU upgrade, so loading time and so might not be comparable)

mteb/benchmarks/_create_table.py
antoineedy: done, and I discovered the `explode` method of dataframes this way 😁

mteb/results/model_result.py
```diff
@@ -43,6 +43,12 @@ def _aggregate_and_pivot(
...
+        df = df.explode("language").reset_index(
```
KennethEnevoldsen, line 48: Hmm, but this treats the score of the task as the score for the language, which is not correct.

A simple example is Miracl:
https://github.com/embeddings-benchmark/results/blob/main/results/GritLM__GritLM-7B/13f00a0e36500c80ce12870ea513846a066004af/MIRACLRetrievalHardNegatives.json

Which has the main score `0.58448` for `ara-Arab` but `0.60572` for `ben-Beng`. Right now I believe you essentially average all of these (the main score for the task) and split it out.

antoineedy: In my leaderboard, the table for MTEB Multiligual, for the task MIRACLRetrievalHardNegatives only, displays this:

SCR-20251208-lvpe

Which corresponds to what you can find in the result sheet for the specified languages. Shouldn't it work that way?

Also, `df = df.explode("language")` is my way to tackle your precedent comment in case mutliple languages are present for a single score:

> I would say the score for English includes all subsets that contain eng-Latn. I.e. add to both

KennethEnevoldsen: Ahh I see that I misunderstood the function - should probably have checked first. Thanks for the clarification

mteb/benchmarks/benchmarks/benchmarks.py
```diff
@@ -892,6 +893,7 @@
 MTEB_multilingual_v2 = Benchmark(
     name="MTEB(Multilingual, v2)",
     display_name="Multilingual",
+    language_view="all",
```
KennethEnevoldsen:
```suggestion
    language_view=[
    "eng-Latn",  # English
    "zho-Hans",  # Chinese (Simplified)
    "hin-Deva",  # Hindi
    "spa-Latn",  # Spanish
    "fra-Latn",  # French
    "ara-Arab",  # Arabic
    "ben-Beng",  # Bengali
    "rus-Cyrl",  # Russian
    "por-Latn",  # Portuguese
    "urd-Arab",  # Urdu
    "ind-Latn",  # Indonesian
    "deu-Latn",  # German
    "jpn-Jpan",  # Japanese
    "swa-Latn",  # Swahili
    "mar-Deva",  # Marathi
    "tel-Telu",  # Telugu
    "tur-Latn",  # Turkish
    "tam-Taml",  # Tamil
    "vie-Latn",  # Vietnamese
    "kor-Hang",  # Korean
]
```

mteb/benchmarks/benchmarks/benchmarks.py
```diff
@@ -410,6 +410,7 @@
     name="MTEB(Scandinavian, v1)",
     display_name="Scandinavian",
     icon="https://github.com/lipis/flag-icons/raw/260c91531be024944c6514130c5defb2ebb02b7d/flags/4x3/dk.svg",
+    language_view=["dan-Latn", "swe-Latn", "nno-Latn", "nob-Latn"],
```
Samoed, line 474: Do we want both nno and nob?

KennethEnevoldsen: Knowing the norwegians I am sure that they would like to have it

antoineedy:
I think the tables look pretty good! → https://huggingface.co/spaces/antoineedy/demo-leaderboard

KennethEnevoldsen:
hmm this is a bit wierd:

Screenshot 2025-12-09 at 20 14 18

Although it has nothing to do with this PR (was added in #3661)

KennethEnevoldsen:
@antoineedy if we add the max column length then I think this is good to merge (happens in summary table for inspiration)

Samoed:
> if we add the max column length

Shouldn't we do this in other pr, because this is not related to the current?

KennethEnevoldsen:
We can do that as well, but I don't want to merge this into main before it is fixed.

antoineedy:
Should I apply this to both per-task table and per-language table? If so, will do it in a couple of minutes, then restart the space, and you'll be able to merge safely

Samoed:
This should be applied to all tables. Ref https://github.com/embeddings-benchmark/mteb/issues/3705

antoineedy:
I added the same logic as the summary table for both the per-task and per-language tables:
```python
# setting task name column width to 250px
column_widths = _get_column_widths(per_task_style.data)
if len(column_widths) > 0:
    column_widths[0] = "250px"
```
Looks great locally, [demo space](https://huggingface.co/spaces/antoineedy/demo-leaderboard) is restarting now.

antoineedy:
Just checked, everything seems good on the demo space!

KennethEnevoldsen:
Thanks @antoineedy - very happy to have this merged in!

PR #3621, opened 2025-11-25
Add token to ds check

Samoed:
ref https://github.com/embeddings-benchmark/mteb/discussions/3620

### DIFF
--- a/.github/workflows/dataset_loading.yml
+++ b/.github/workflows/dataset_loading.yml
@@ -23,5 +23,7 @@ jobs:
           make install-for-tests
 
     - name: Run dataset loading tests
+      env:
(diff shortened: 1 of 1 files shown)


.github/workflows/dataset_loading.yml
```diff
@@ -23,5 +23,7 @@ jobs:
           make install-for-tests
 
     - name: Run dataset loading tests
+      env:
+        HF_TOKEN: ${{ secrets.HF_TOKEN }}
```
Samoed, line 27: We're already have it for leaderboard refresh

KennethEnevoldsen:
I merged both workflow on #3622 so I will close this one

PR #3622, opened 2025-11-25
ci: Add HF_TOKEN to dataset loading and merge CI

### LINKED ISSUE #2950: CI: for datasets available on HF fails too often
It seems like the [ci for checking if datasets are available on Huggingface](https://github.com/embeddings-benchmark/mteb/actions/workflows/dataset_loading.yml) fails 3/4 times

It would be great to make it slightly more stable.

KennethEnevoldsen:
fixes #2950
related to #3620


Samoed (approved):

.github/workflows/dataset_loading.yml
```diff
@@ -1,27 +1,34 @@
...
     runs-on: ubuntu-latest
-
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
-
+      uses: actions/checkout@v4
+    
     - name: Set up Python
       uses: actions/setup-python@v4
```
Samoed:
```suggestion
      uses: actions/setup-python@v6
```

PR #3602, opened 2025-11-22
model: Add eager-embed embedding model

jpbalarini:
Add inference code for eager-embed embedding model.
[eager-embed-v1](https://huggingface.co/eagerworks/eager-embed-v1) is a multimodal dense embedding model with a 2560 embed dimension based on Qwen3-VL and finetuned on multiple public datasets.

More info here:
https://huggingface.co/eagerworks/eager-embed-v1
https://github.com/eagerworks/eager-embed

**Checklist**:
- [x] I have filled out the ModelMeta object to the extent possible
- [x] I have ensured that my model can be loaded using
  - [x] `mteb.get_model(model_name, revision)` and
  - [x] `mteb.get_model_meta(model_name, revision)`
- [x] I have tested the implementation works on a representative set of tasks.
- [x] The model is public, i.e. is available either as an API or the weight are publicly available to download

PR #3627, opened 2025-11-26
model: Add tomoro-colqwen3-embed embedding models

01234568:
Add inference code and requirements for tomoro-colqwen3-embed 8B and 4B models.

They are based on merged Qwen3-VL-Instruct and Qwen3-Embedding checkpoints finetuned using the ColQwen method. They produce 320-dimension embeddings per text/image token. Finetuning data is a subset of nvidia/llama-nemoretriever-colembed-3b-v1 training data.

Model checkpoints available at:
https://huggingface.co/TomoroAI/tomoro-colqwen3-embed-4b
https://huggingface.co/TomoroAI/tomoro-colqwen3-embed-8b

**Checklist**:


mteb/models/model_implementations/colqwen_models.py
```diff
@@ -73,6 +81,148 @@ def __init__(
...
+
+        # Pad to support variable sequence lengths while keeping token-level outputs for MaxSim.
+        padded = torch.nn.utils.rnn.pad_sequence(
+            all_embeds, batch_first=True, padding_value=0
+        )
+        return padded
+
+    def calculate_probs(self, text_embeddings, image_embeddings):
+        scores = self.similarity(text_embeddings, image_embeddings).T
+        return scores.softmax(dim=-1)
```
Samoed: This won't be used
```suggestion
```

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -73,6 +81,148 @@ def __init__(
...
+                inputs = self.processor.process_texts(batch["text"])
+                inputs = {k: v.to(self.device) for k, v in inputs.items()}
+                outs = self._encode_inputs(inputs)
+                all_embeds.extend(outs.cpu().to(torch.float32))
+
+        # Pad to support variable sequence lengths while keeping token-level outputs for MaxSim.
+        padded = torch.nn.utils.rnn.pad_sequence(
+            all_embeds, batch_first=True, padding_value=0
+        )
+        return padded
```
Samoed: Can you encode images and texts together?

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -73,6 +81,148 @@ def __init__(
...
+        scores = self.similarity(text_embeddings, image_embeddings).T
+        return scores.softmax(dim=-1)
+
+    def similarity(self, a, b):
+        if hasattr(self.processor, "score_multi_vector"):
+            return self.processor.score_multi_vector(a, b, device=self.device)
+
+        raise AttributeError(
+            "Processor does not implement score_multi_vector or score for similarity computation."
+        )
```
Samoed:
```suggestion
        return self.processor.score_multi_vector(a, b, device=self.device)
```

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -125,6 +275,75 @@ def __init__(
...
+  author={Xin Huang and Kye Min Tan and Albert Phelps},
+  year={2025},
+  url={https://huggingface.co/TomoroAI/tomoro-colqwen3-embed-8b}
+}
+"""
+
+colqwen3_8b = ModelMeta(
+    loader=ColQwen3Wrapper,
+    loader_kwargs=dict(
+        dtype=torch.bfloat16,
```
Samoed: Your model already in bf16

mteb/models/model_implementations/colqwen_models.py
01234568: This was copied from ColPaliEngineWrapper, I can remove it if it's unused.

mteb/models/model_implementations/colqwen_models.py
01234568: Do you mean having images and texts together in 1 item? Like get_fused_embeddings. The model supports that, but we have not tested its performance in that aspect.

mteb/models/model_implementations/colqwen_models.py
01234568: Ok, will remove this. Confirmed that the default is to load in bf16.

mteb/models/model_implementations/colqwen_models.py
Samoed: I see. I will remove from them too

mteb/models/model_implementations/colqwen_models.py
Samoed: Yes. You can test this and see when results are higher

01234568: Can I clarify under which situation get_fused_embeddings is used? For our model we embed the text+image item as one set of embeddings, some example use cases would be a document embedding of a diagram+caption, or a query where an image is attached to the question.

Our model does not differentiate between queries and docs so get_fused_embeddings can be used for either one.

Here is the implementation we're planning to go with:
```python
    def get_fused_embeddings(
        self,
        image_texts_pairs: DataLoader[BatchedInput] | None = None,
        batch_size: int = 32,
        fusion_mode="concat",
        **kwargs: Any,
    ):
        import torchvision.transforms.functional as F

        all_embeds: list[torch.Tensor] = []
        with torch.no_grad():
            for batch in tqdm(
                image_texts_pairs,
                disable=not show_progress_bar,
                desc="Encoding images+texts",
            ):
                imgs = [
                    F.to_pil_image(b.to(self.device))
                    if not isinstance(b, Image.Image)
                    else b
                    for b in batch["image"]
                ]
                texts = batch["text"]
                assert len(imgs) == len(texts), (
                    f"The number of texts and images must have the same length, got {len(imgs)} and {len(texts)}"
                )

                inputs = self.processor(images=imgs, text=texts)
                inputs = {k: v.to(self.device) for k, v in inputs.items()}
                outs = self._encode_inputs(inputs)
                all_embeds.extend(outs.cpu().to(torch.float32))

        padded = torch.nn.utils.rnn.pad_sequence(
            all_embeds, batch_first=True, padding_value=0
        )
        return padded
```

Samoed: `get_fused_embeddings` won't be used directly. Only `encode` will be used from your implementation

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -130,24 +130,12 @@ def encode(
...
-        if text_embeddings is not None:
-            return text_embeddings
-        if image_embeddings is not None:
-            return image_embeddings
+        if "text" in inputs.dataset.features and "image" in inputs.dataset.features:
+            return self.get_fused_embeddings(inputs, **kwargs)
+        elif "text" in inputs.dataset.features:
+            return self.get_text_embeddings(inputs, **kwargs)
+        elif "image" in inputs.dataset.features:
+            return self.get_image_embeddings(inputs, **kwargs)
```
Samoed: I think easier will be to have just `get_fused_embeddings`

01234568: There are still some things we need to get sorted out on our end for the huggingface processor implementation which this relies on, so we'll disable `get_fused_embeddings` for the time being. Will update later on if that's ok?

Samoed: Do you mean after PR merge?

01234568: Yeah

Samoed: I'm afraid that you will need to rerun all results then. What problem do you have? Probably you can do something similar in processor like https://github.com/embeddings-benchmark/mteb/pull/3602 if you want to add EOS token or something else

01234568: For the results we're currently submitting (vidore v1-v3) it should not be affected since it's either text only or image only. When we implemented the huggingface processor we decided not to allow text-only items in the fused embeddings, since it would result in excessive pad tokens. May I know which test sets require image+text embeddings?

Edit: And for test sets using image+text embeddings, are there text-only samples mixed with image+text samples in one batch?

Samoed: > And for test sets using image+text embeddings

For example `BLINKIT2IRetrieval` https://github.com/embeddings-benchmark/mteb/blob/7e2fa988df697a86919cc96268ec1f2fcfb261bd/mteb/tasks/retrieval/eng/blink_it2i_retrieval.py#L7 https://huggingface.co/datasets/JamieSJS/blink-it2i/

> mixed with image+text samples in one batch?

No

01234568: Updated to have everything go through get_fused_embeddings.

Samoed (approved):
Great work!

01234568:
Updated the huggingface revision with updated processor supporting fused embeddings. It was tested locally.

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -125,6 +259,69 @@ def __init__(
     citation=COLPALI_CITATION,
 )
 
+TOMORO_TRAINING_DATA = {
+    # from https://huggingface.co/datasets/vidore/colpali_train_set
+    "vdr-multilingual-train",
+    "colpali_train_set",  # as it contains PDFs
+    "VisRAG-Ret-Train-Synthetic-data",
+    "VisRAG-Ret-Train-In-domain-data",
```
KennethEnevoldsen:
```suggestion
    # not in MTEB:
    # "vdr-multilingual-train",
    # "colpali_train_set",  # as it contains PDFs
    # "VisRAG-Ret-Train-Synthetic-data",
    # "VisRAG-Ret-Train-In-domain-data",
```

This has to be a set of tasks in MTEB (though we typically keep this as well, but as comments)

KennethEnevoldsen (commented):
Great addition!

A few minor things on the metadata, otherwise do see @Samoed's comment on `get_fused_embeddings`

mteb/models/model_implementations/colqwen_models.py
KennethEnevoldsen: Do any of these datasets share content with tasks in MTEB?

Samoed: Colpali and vdr overlaps, but we need to annotate them on other models too, so I don't think this is a problem for this pr. This model is using only `fused_embeddings` and I think it's good now

KennethEnevoldsen: Do you mean vidore when you say vdr?

Samoed: Yes

Samoed: Yes

KennethEnevoldsen: Hmm - I generally don't want to merge a model that doesn't have correct annotations (as reviewing the results becomes much harder)

01234568: We followed the way training data is listed in https://github.com/embeddings-benchmark/mteb/blob/072e6ef63e961128ea60a697cc077feb2fd638d0/mteb/models/model_implementations/nvidia_llama_nemoretriever_colemb.py#L118-L136 since our data is a subset of it. vdr-multilingual-train refers to this: https://huggingface.co/datasets/llamaindex/vdr-multilingual-train

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -125,6 +259,69 @@ def __init__(
     citation=COLPALI_CITATION,
 )
 
+TOMORO_TRAINING_DATA = {
+    # from https://huggingface.co/datasets/vidore/colpali_train_set
+    "vdr-multilingual-train",
+    "colpali_train_set",  # as it contains PDFs
```
Samoed: Update annotation in https://github.com/embeddings-benchmark/mteb/pull/3636

```suggestion
    "VDRMultilingualRetrieval",
    # from https://huggingface.co/datasets/vidore/colpali_train_set
    "VidoreDocVQARetrieval",
    "VidoreInfoVQARetrieval",
    "VidoreTatdqaRetrieval",
    "VidoreArxivQARetrieval",
```

01234568:
Any remaining items from my side?

Samoed:
I don't think so

Issue #3531, opened 2025-11-06
`Vidore` zero shot annotation for colpali


Samoed:
After https://github.com/embeddings-benchmark/mteb/pull/3526 I've checked colpali training set, https://github.com/alibaba/mteb/blob/1e29d8b607918d8b17e9bb5cb5dcacda2d9b0b48/mteb/models/model_implementations/colpali_models.py#L198-L202 and some tasks in it are in `ViDoRe` (VidoreArxivQARetrieval, VidoreDocVQARetrieval, VidoreInfoVQARetrieval) benchmark, but on leaderboard these models are `zero shot` 100%, which is not correct.

<img width="1199" height="630" alt="Image" src="https://github.com/user-attachments/assets/7bee5ee0-716a-4307-8652-45a97739079d" />

KennethEnevoldsen:
We can just add them now (the leaderboard will update)

PR #3636, opened 2025-11-29
fix: `colpali_training_set` & updated `JinaVDR` and `ViDoRe` tasks annotation

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3531

@maximilianwerk @antoineedy @QuentinJGMace Can you check if annotation is correct?

- For `JINAVDR` I updated description and added first line from dataset's readme and added `adapted_from` to `ViDoRe` datasets.
- For`ViDoRe` added to some tasks `adapted_from` fields.
- Fixed annotation for models trained on `colpali_train_set`


KennethEnevoldsen (approved):
Looks good!

KennethEnevoldsen:
I will merge this, but if we get any comments we can always update the annotations

Issue #3251, opened 2025-10-04
Add model: jinaai/jina-reranker-v3


Samoed:
### Model link on Hugging Face

https://huggingface.co/jinaai/jina-reranker-v3

### Arxiv link

https://arxiv.org/abs/2509.25085

### Description of the model

Congratulations with release @hanxiao

PR #3645, opened 2025-12-02
model: jina-reranker-v3

qherreros:

Close https://github.com/embeddings-benchmark/mteb/issues/3251

### DIFF
--- a/mteb/models/model_implementations/jina_models.py
+++ b/mteb/models/model_implementations/jina_models.py
@@ -218,7 +218,7 @@ JINARerankerV3_TRAINING_DATA = {
     "DuRetrieval",
     "MMarcoReranking",
     "CornStack",
-    "MLDR",
+    "MultiLongDocRetrieval",
     "StackOverflowQA",
 }
 


mteb/models/model_implementations/jina_models.py
```diff
@@ -147,6 +227,79 @@
...
+        trust_remote_code: bool = True,
+        **kwargs: Any,
+    ) -> None:
+        if isinstance(model, str):
+            model = JinaRerankerV3(model, trust_remote_code=trust_remote_code, **kwargs)
+        self.model = model
+
+        super().__init__(
+            model, revision=revision, trust_remote_code=trust_remote_code, **kwargs
+        )
```
Samoed: It will be wrapped automatically for reranking tasks, you don't need to inherit from it
```suggestion
```

mteb/models/model_implementations/jina_models.py
```diff
@@ -553,6 +706,42 @@ def get_programming_task_override(
...
+    ),
+    name="jinaai/jina-reranker-v3",
+    languages=multilingual_langs,
+    open_weights=True,
+    revision="050e171c4f75dfec5b648ed8470a2475e5a30f30",
+    release_date="2025-09-18",  # official release date
+    modalities=["text"],
+    n_parameters=int(0.6 * 1e9),
+    memory_usage_mb=1138,
+    max_tokens=32768,
```
Samoed:
```suggestion
    max_tokens=131072,
```

mteb/models/model_implementations/jina_models.py
```diff
@@ -147,6 +227,79 @@
...
+
+    def predict(self, sentences: list[tuple[str, str]], **kwargs: Any) -> list[float]:
+        """
+        Args:
+            sentences: List of (query, document) pairs.
+            batch_size: MTEB passes this, though we group by query mostly.
+        """
+        sentences_count = len(sentences)
+        if sentences_count == 0:
+            return []
```
Samoed:
```suggestion
```

mteb/models/model_implementations/jina_models.py
```diff
@@ -147,6 +227,79 @@
...
+
+        query_groups: dict[str, list[tuple[int, str]]] = defaultdict(list)
+        for idx, (query, doc) in enumerate(sentences):
+            query_groups[query].append((idx, doc))
+
+        results = [0.0] * sentences_count
+        for query, doc_infos in query_groups.items():
+            original_indices, docs = zip(*doc_infos)
+
+            scores = self.model.rerank(query, list(docs), max_query_length=3072)
```
Samoed: Overall they're supporting up to 132k context, so I'm not sure about limit in 3072.

Samoed (changes requested):

mteb/models/model_implementations/jina_models.py
```diff
@@ -147,6 +227,79 @@
...
+            for scr in scores:
+                original_idx = original_indices[scr["index"]]
+                results[original_idx] = float(scr["relevance_score"])
+
+        return results
+
+
+class JinaRerankerV3Wrapper(CrossEncoderWrapper):
```
Samoed: Can you inherit `JinaRerankerV3` from `CrossEncoderWrapper` directly without a second wrapper?

mteb/models/model_implementations/jina_models.py
qherreros: The model does support 132K tokens, but current Jina's `rerank` implementation does truncate queries that have more than 512 tokens. I could leave it to default, but it seems to penalize evaluation on some datasets like WebLinx. What is the policy on that? Can I have a variable query truncation per task? Or should I default to only one value?

Samoed: You can make it variable between tasks from descriptive stats, but probably it would be better to tokenize all queries and passages and compute their lengths, because now it will truncate documents to 2048 tokens. https://huggingface.co/jinaai/jina-reranker-v3/blob/main/modeling.py#L203

mteb/models/model_implementations/jina_models.py
qherreros: I tried something in this spirit, please let me know if this is what you had in mind. (Sorry for the back and forth, I didn't find any example I could get inspired from).

mteb/models/model_implementations/jina_models.py
```diff
@@ -147,6 +228,69 @@
...
+            if device is None:
+                device = get_device_name()
+                logger.info(f"Use pytorch device: {device}")
+
+            self.model.to(device)
+            self.model.eval()
+        if isinstance(model, CrossEncoder):
+            self.model = model
+
+        self.mteb_model_meta = _model_meta_from_cross_encoder(self.model)
```
Samoed:
```suggestion
        from sentence_transformers.util import get_device_name
        from transformers import AutoModel

        self.model = AutoModel.from_pretrained(
            model, trust_remote_code=trust_remote_code, dtype="auto"
        )

        device = kwargs.get("device", None)
        if device is None:
            device = get_device_name()
            logger.info(f"Use pytorch device: {device}")

        self.model.to(device)
        self.model.eval()
```

mteb/models/model_implementations/jina_models.py
qherreros: Agreed, the current strategy is a bit rigid. I'll discuss that with the team. Would it block the PR or could we improve that later on?

Samoed: This could affect results, so I think this is blocking

numb3r3: @Samoed I want to clarify that our `jina-reranker-v3` model is a listwise reranker indeed. The 131K tokens is the total tokens of all of the documents together. Thus, you can see we would truncate each document to 2048 by default. Of course, we provide parameters to allow users to choose the truncation length in our `rerank` API.

Samoed: If you think that we can leave 2048 max doc lenght with this query length, than this can be merged

numb3r3: Sounds great! Given that the $2048$ max document length is fine with the current query length, we are fine with the current settings. Thanks for the clarification and your support. We can proceed with the merge.

mteb/models/model_implementations/jina_models.py
```diff
@@ -147,6 +228,62 @@
...
+        sentences_count = len(all_corpus)
+        query_groups: dict[str, list[tuple[int, str]]] = defaultdict(list)
+        for idx, (query, doc) in enumerate(zip(all_queries, all_corpus)):
+            query_groups[query].append((idx, doc))
+
+        results = np.zeros(sentences_count, dtype=np.float32)
+        for query, doc_infos in query_groups.items():
+            original_indices, docs = zip(*doc_infos)
+
+            scores = self.model.rerank(query, list(docs), max_query_length=3072)
```
Samoed: For visibility
```suggestion
            scores = self.model.rerank(query, list(docs), max_query_length=3072, max_doc_length=2048,)
```

mteb/models/model_implementations/jina_models.py
Samoed: This seems great!

mteb/models/model_implementations/jina_models.py
```diff
@@ -553,6 +692,42 @@ def get_programming_task_override(
...
+    max_tokens=131072,
+    embed_dim=None,
+    license="cc-by-nc-4.0",
+    similarity_fn_name=None,
+    framework=["PyTorch"],
+    use_instructions=None,
+    reference="https://huggingface.co/jinaai/jina-reranker-v3",
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
```
Samoed: Can you annotate your training datasets?

mteb/models/model_implementations/jina_models.py
```diff
@@ -553,6 +692,42 @@ def get_programming_task_override(
...
+    modalities=["text"],
+    n_parameters=int(0.6 * 1e9),
+    memory_usage_mb=1138,
+    max_tokens=131072,
+    embed_dim=None,
+    license="cc-by-nc-4.0",
+    similarity_fn_name=None,
+    framework=["PyTorch"],
+    use_instructions=None,
+    reference="https://huggingface.co/jinaai/jina-reranker-v3",
```
Samoed, line 736:
```suggestion
    reference="https://huggingface.co/jinaai/jina-reranker-v3",
    is_cross_encoder=True,
```

mteb/models/model_implementations/jina_models.py
```diff
@@ -200,6 +200,28 @@
...
+    "MSMARCOHardNegatives",
+    "NQ",
+    "NQHardNegatives",
+    "HotpotQA",
+    "HotpotQAHardNegatives",
+    "T2Retrieval",
+    "DuRetrieval",
+    "MMarcoReranking",
+    "CornStack",
+    "MLDR",
```
Samoed:
```suggestion
    "MultiLongDocRetrieval",
```

mteb/models/model_implementations/jina_models.py
qherreros: Done !

PR #3646, opened 2025-12-03
Model: add e5-nl models

nikolay-banar:

### DIFF
--- a/mteb/models/model_implementations/e5_models.py
+++ b/mteb/models/model_implementations/e5_models.py
@@ -126,18 +126,6 @@ E5_CITATION = """
 }
 """
 
-E5_NL_CITATION = """
-@misc{banar2025mtebnle5nlembeddingbenchmark,
-  archiveprefix = {arXiv},
-  author = {Nikolay Banar and Ehsan Lotfi and Jens Van Nooten and Cristina Arhiliuc and Marija Kliocaite and Walter Daelemans},
-  eprint = {2509.12340},
-  primaryclass = {cs.CL},
-  title = {MTEB-NL and E5-NL: Embedding Benchmark and Models for Dutch},
-  url = {https://arxiv.org/abs/2509.12340},
-  year = {2025},
-}
-"""
-
 model_prompts = {
     PromptType.query.value: "query: ",
     PromptType.document.value: "passage: ",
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/e5_models.py
```diff
@@ -399,3 +411,81 @@
...
+    embed_dim=384,
+    license="mit",
+    max_tokens=512,
+    reference="https://huggingface.co/clips/e5-small-trm-nl",
+    similarity_fn_name=ScoringFunction.COSINE,
+    framework=["Sentence Transformers", "PyTorch"],
+    use_instructions=True,
+    public_training_code="https://github.com/ELotfi/e5-nl",  # couldn't find
+    public_training_data=None,
+    training_datasets=ME5_TRAINING_DATA,  # mMARCO-NL, HotpotQA-NL, FEVER-NL, and LLM generated data
```
KennethEnevoldsen: I am unsure why the comments are here - should the set be extended?

mteb/models/model_implementations/e5_models.py
```diff
@@ -399,3 +411,81 @@
...
+    n_parameters=40_800_000,
+    memory_usage_mb=78,
+    embed_dim=384,
+    license="mit",
+    max_tokens=512,
+    reference="https://huggingface.co/clips/e5-small-trm-nl",
+    similarity_fn_name=ScoringFunction.COSINE,
+    framework=["Sentence Transformers", "PyTorch"],
+    use_instructions=True,
+    public_training_code="https://github.com/ELotfi/e5-nl",  # couldn't find
```
KennethEnevoldsen: Is this comment an accidental copy paste?

mteb/models/model_implementations/e5_models.py
```diff
@@ -399,3 +411,81 @@
     training_datasets=E5_TRAINING_DATA,
     citation=E5_CITATION,
 )
+
+e5_nl_small = ModelMeta(
```
KennethEnevoldsen: I would move the models into:

`model_implementations/clips_models.py`

KennethEnevoldsen (approved):
A few minor comments - generally it looks good. Any reason why the dataset is not shared?

mteb/models/model_implementations/e5_models.py
nikolay-banar: ME5_TRAINING_DATA already contains these datasets, but mE5s are not actually trained on them. So, I was not sure how to emphasise that.

mteb/models/model_implementations/e5_models.py
nikolay-banar: yes!

mteb/models/model_implementations/e5_models.py
KennethEnevoldsen: Ahh got it - I think this works and if not we have this PR to trace it back to

nikolay-banar:
> A few minor comments - generally it looks good. Any reason why the dataset is not shared?

We mostly used public data (some datasets from BEIR-NL). I included a link to this collection. The LLM generated dataset is not available yet, but we plan to release it soon.

KennethEnevoldsen:
Fair - you can always add the link in the future - otherwise this is all good. I have set it to auto-merge

nikolay-banar:
> Fair - you can always add the link in the future - otherwise this is all good. I have set it to auto-merge

Is auto-merge set? Do I need to do anything else?

Issue #2418, opened 2025-03-23
Number of Parameters is in alphabetical order and not numerical order.


Hobstarr:
The leaderboard summary allows you to order the models by number of parameters.

If we look at increasing size we see that the order is:

102M
102M
......
162M
17M
17M
19M
1B
2B
2M
.....

This is not a helpful method for ordering models based on their number of parameters and if possible would be better to order by size numerically.

ayush1298:
@Hobstarr This was happening because `Number of parameters` column is string and hence it is doing alphabetical sorting.  
I have tried some ways, but there is nothing to have string column but apply sorting based on numerical values. Any idea how to do this @Samoed ?

Samoed:
No, I don't have good ideas

KennethEnevoldsen:
I have made a feature requests on gradio here: https://github.com/gradio-app/gradio/issues/12407

KennethEnevoldsen:
We could convert this column to number of parameters (millions)

so 7B > 7000. This would require the column header to be larger though.

PR #3653, opened 2025-12-04
fix: Remove "Unknown" for int on leaderboard causing them to be unsortable

### LINKED ISSUE #3579: [BUG] Sorting by Memory U... sorts numbers as text instead of numeric

Sorting by Memory U... sorts numbers as text instead of numeric

KennethEnevoldsen:
Fixes #3579
Fixes https://github.com/embeddings-benchmark/mteb/issues/2418

Also fixed "Embedding Dimensions"

New LB looks like:

Screenshot 2025-12-05 at 01 06 12

### DIFF
--- a/mteb/models/model_implementations/google_models.py
+++ b/mteb/models/model_implementations/google_models.py
@@ -147,7 +147,6 @@ class GoogleTextEmbeddingModel(AbsEncoder):
 google_text_emb_004 = ModelMeta(
     loader=GoogleTextEmbeddingModel,  # type: ignore[call-arg]
     loader_kwargs=dict(
-        model_name="text-embedding-004",
         model_prompts=MODEL_PROMPTS,
     ),
     name="google/text-embedding-004",
@@ -172,7 +171,6 @@ google_text_emb_004 = ModelMeta(
 google_text_emb_005 = ModelMeta(
     loader=GoogleTextEmbeddingModel,  # type: ignore[call-arg]
     loader_kwargs=dict(
-        model_name="text-embedding-005",
         model_prompts=MODEL_PROMPTS,
     ),
     name="google/text-embedding-005",
@@ -197,7 +195,6 @@ google_text_emb_005 = ModelMeta(
 google_text_multilingual_emb_002 = ModelMeta(
     loader=GoogleTextEmbeddingModel,  # type: ignore[call-arg]
     loader_kwargs=dict(
-        model_name="text-embedding-002",
         model_prompts=MODEL_PROMPTS,
     ),
     name="google/text-multilingual-embedding-002",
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/linq_models.py
```diff
@@ -1,7 +1,7 @@
 import torch
```
isaac-chung, line 1: Are these changes related?

isaac-chung (commented):
The screenshot shows one "unknown" in  number of Params. Should we fix that column too? Otherwise I only had a question. The rest looks good.

mteb/models/model_implementations/linq_models.py
KennethEnevoldsen: No it should already be merged in - not sure why the diff is showing #3654

KennethEnevoldsen:
> The screenshot shows one "unknown" in number of Params. Should we fix that column too? Otherwise I only had a question. The rest looks good.

~~Good catch - let me track that one down~~

I only changed the embedding dimensions and Memory usage. Number of parameters is already a string (but I can still remove the Unknown for consistency).

Same thing for Max Tokens, which both have unknown and infinite (which probably could be blank as well).

If I do that is would probably change Number of parameters to one of:

- "Number of parameters (B)", would lead to values like 0.007
- "Numbers of parameters (M)", would lead to values like 56.000

Leaning B

mteb/models/model_implementations/linq_models.py
Samoed: You need to merge main into your branch

Samoed:
Test if failing because of `ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device`. So fix for results was unrelated to repository size


> Leaning B

Agree

KennethEnevoldsen:
Removed the "Unknown" in general from the leaderboard

Screenshot 2025-12-05 at 12 20 26

Screenshot 2025-12-05 at 12 16 21

I can still experience cases where models that have a NaN Mean (Task) still appear on the plot:

Screenshot 2025-12-05 at 12 22 17

KennethEnevoldsen:
Seems like the last bug is an existing bug so will not fix it here (issue #3666)

Samoed:
Will `Model Parameters` filter work with this implementation?
image

KennethEnevoldsen:
@q275343119 can I ask you to review this PR as well?

KennethEnevoldsen:
> Will Model Parameters filter work with this implementation?

~~No it does not will add a fix to it'~~ It does

KennethEnevoldsen:
Ohh it actually did work:

Screenshot 2025-12-05 at 13 11 32

I was just mixing it up with the bug https://github.com/embeddings-benchmark/mteb/issues/3666

isaac-chung:
The half circles look kinda odd. Are we ok with that?

Samoed:
This bug currently on leaderboard and not related to this PR

isaac-chung:
Oh okay, thanks

KennethEnevoldsen:
Alright - I will merge this @q275343119 I would still love your feedback when you have the time

q275343119:
> Alright - I will merge this @q275343119 I would still love your feedback when you have the time

Removing "Unknown" is a good idea to ensure the uniformity of field types and prevent string types from appearing in character columns. However, it will also result in None or NaN, which may require special handling.

Samoed:
It looks like as empty cells, which is fine I think

image

PR #3658, opened 2025-12-05
fix: GoogleTextEmbeddingModel were given multiple `model_name`

KennethEnevoldsen:


mteb/models/model_implementations/google_models.py
Samoed, line 1: I think you need to remove also for `text-embedding-002`, `text-embedding-005` etc

KennethEnevoldsen: fixed!

Samoed (approved):

Issue #3669, opened 2025-12-05
ci: Release workflow seems to fail


KennethEnevoldsen:
Don't have the time to look at it now - this is just to track it

KennethEnevoldsen:
example: https://github.com/embeddings-benchmark/mteb/actions/runs/19967709870/job/57264157207

PR #3671, opened 2025-12-05
fix: remove docker system prune command

Samoed:
Removed unnecessary docker system prune command from the workflow.

Close https://github.com/embeddings-benchmark/mteb/issues/3669


KennethEnevoldsen:
Thanks for the fix @Samoed!

PR #3673, opened 2025-12-05
fix: `get_model` now correctly assumed `SentenceTransformer`

### LINKED ISSUE #3670: `get_model` does not default to `SentenceTransformer` as noted
### Describe the bug

`get_model` should default to SentenceTransformers for models not in MTEB, but it does not.

### To reproduce

```
import mteb

mdl = mteb.get_model("NbAiLab/nb-bert-base")
```
where we get:
```
NotImplementedError: No model implementation is available for this model.
```

### Additional information

_No response_

Yes

KennethEnevoldsen:
fixes #3670

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,6 @@ jobs:
             /usr/local/share/powershell \
             /usr/share/dotnet \
             /usr/share/swift
-          docker system prune -af
 
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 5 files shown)


mteb/models/get_model_meta.py
```diff
@@ -180,6 +181,14 @@ def _model_meta_from_hf_hub(model_name: str) -> ModelMeta:
     if card_data.get("library_name", None) == "sentence-transformers":
         frameworks.append("Sentence Transformers")
         loader = sentence_transformers_loader
+    else:
+        msg = (
+            "Model library not recognized, defaulting to Sentence Transformers loader."
+        )
+        logger.warning(msg)
+        warnings.warn(msg)
```
Samoed, line 189: Should we use only one of logger or warning?

Samoed (approved):

mteb/models/get_model_meta.py
KennethEnevoldsen: I think they kinda solves two different problems

KennethEnevoldsen: https://stackoverflow.com/questions/9595009/warnings-warn-vs-logging-warning

Samoed: Yes, but here we're using both of them and probably can be used only one of them

PR #3683, opened 2025-12-06
model: add two Japanese models sarashina-embedding-{v1, v2}-1b

lsz05:
This PR adds two models in the `sarashina-embedding` series: [v1-1b](https://huggingface.co/sbintuitions/sarashina-embedding-v1-1b) and [v2-1b](https://huggingface.co/sbintuitions/sarashina-embedding-v2-1b).

### DIFF
--- a/mteb/models/model_implementations/sarashina_embedding_models.py
+++ b/mteb/models/model_implementations/sarashina_embedding_models.py
@@ -105,7 +105,7 @@ def sarashina_instruction_template(
     if not instruction:
         return ""
     if prompt_type == PromptType.document:
-        return instruction
+        return "text: "
     return f"task: {instruction}\nquery: "
 
 


mteb/models/model_implementations/sarashina_embedding_models.py
```diff
@@ -0,0 +1,107 @@
...
+    "JaCWIRReranking-document": "text: ",
+    "MIRACLReranking-query": "task: 質問を与えるので、その質問に答えるのに役立つ関連文書を検索してください。\nquery: ",
+    "MIRACLReranking-document": "text: ",
+    "MultiLongDocReranking-query": "task: 質問を与えるので、その質問に答えるのに役立つWikipediaの文章を検索してください。\nquery: ",
+    "MultiLongDocReranking-document": "text: ",
+}
+
+
+sbintuitions_sarashina_embedding_v2_1b = ModelMeta(
+    loader=sentence_transformers_loader,
```
Samoed: Are these models are instruction? I think it's better to use InstructionWrapper https://github.com/embeddings-benchmark/mteb/blob/057f0e1de22554e261447cc0ac8b1b90ad70d826/mteb/models/instruct_wrapper.py#L96 to always apply instructions for tasks rather than like prefixes

lsz05: Thank you for the advice. Turned to `InstructSentenceTransformerModel` now.
It seems good now.
Btw, only v2 is instructional. v1 doesn't use any instruction nor prefix.

mteb/models/model_implementations/sarashina_embedding_models.py
```diff
@@ -0,0 +1,166 @@
...
+) -> str:
+    """Instruction template for Sarashina v2 model.
+
+    Returns the instruction as-is since the prompts already contain the full format.
+    For document prompts, returns the instruction directly (e.g., "text: ").
+    """
+    if not instruction:
+        return ""
+    if prompt_type == PromptType.document:
+        return instruction
```
Samoed: I think it's better to return `text` to apply correctly to other tasks
```suggestion
        return "text: "
```

Samoed (approved):

Issue #3672, opened 2025-12-05
Add functionality to remove low-quality sample from a dataset


KennethEnevoldsen:
### Description of the feature

E.g., remove duplicates and text less than n characters.

Something like:
```
task = mteb.get_task("MassiveIntentClassification")
task.remove_duplicates()
task.filter_short_documents(min_length=5)
```

Samoed:
Probably we can also add this functionality similarly to #2900 https://gist.github.com/AlexeyVatolin/ea3adc21aa7a767603ff393b22085adc

KennethEnevoldsen:
Yep, could definitely be something like that

ayush1298:
Mentioning the list of filters here, which we can do. #5107 implements only the deduplication filter to start with. 

Filters
- [x] deduplication  - #5107 
- [ ] filter_short_documents
    - [ ] filter_empty  — empty / whitespace-only text
    - [ ] filter_short  - under N characters or words
- [ ] filter_train_leakage  - eval rows also present in train. Currently from old pipeline code, this function is classification specific, but we also can implement a function that would check data across splits.
- [ ] filter_unclear_label  - text appearing with contradictory labels, or unlabel documents

Not filters, but part of cleaning pipeline
- [ ] split_train_test  - creates a test split where none exists (restructuring)
- [ ] clean_dataset / process_classification — the old pipeline; likely obsolete
      once the filters above are individually exposed

Samoed:
>  filter_empty — empty / whitespace-only text
filter_short - under N characters or words

I think we should also add this with any modalities

ayush1298:
> > filter_empty — empty / whitespace-only text
> > filter_short - under N characters or words
> 
> I think we should also add this with any modalities

yes, when I will do them in separate PR, will handle those as well. Here, i have just mentioned broad catagories

PR #3607, opened 2025-11-23
dataset: Add hebrew v3

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3578

PR #3685, opened 2025-12-06
Add dataset task filter

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/issues/3672

For now, this is MVP version of cleaning. I've added it for now because I needed to filter https://github.com/embeddings-benchmark/mteb/pull/3607. I'm not sure what is best way to implement it. 

1. We can create some processing pipeline `task.filter_data(["filter1", "filter2", ...])`, but this would probably hard to use.
2. We can keep this as separate file with a bunch of util functions and maybe with some presets

### DIFF
--- a/mteb/abstasks/_data_filter/filters.py
+++ b/mteb/abstasks/_data_filter/filters.py
@@ -7,7 +7,6 @@ import pandas as pd
 from datasets import Dataset, DatasetDict
 
 from mteb import TaskMetadata
-from mteb.abstasks import AbsTaskClassification
 
 logger = logging.getLogger(__name__)
 
@@ -23,7 +22,7 @@ def deduplicate(dataset: Dataset, input_column: str) -> Dataset:
             indices_to_keep.append(i)
 
     logger.info(
-        f"[deduplicate] kept={len(indices_to_keep)}, removed={len(dataset) - len(indices_to_keep)}"
+        f"[deduplicate] removed={len(dataset) - len(indices_to_keep)}/{len(dataset)}"
     )
     return dataset.select(indices_to_keep)
 
@@ -32,7 +31,7 @@ def filter_empty(dataset: Dataset, input_column: str) -> Dataset:
     """Filter out empty or whitespace-only examples."""
     before = len(dataset)
     ds = dataset.filter(lambda x: len(x[input_column].strip()) > 0)
-    logger.info(f"[filter_empty] removed={before - len(ds)}")
+    logger.info(f"[filter_empty] removed={before - len(ds)}/{before}")
     return ds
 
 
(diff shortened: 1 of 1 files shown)


mteb/abstasks/_data_filter/classificationefilter.py
```diff
@@ -0,0 +1,203 @@
...
+) -> DatasetDict:
+    """Apply the full cleaning pipeline with logging."""
+    logger.info("[clean_dataset] Starting dataset cleaning pipeline...")
+
+    transforms = [
+        ("filter_empty", filter_empty),
+        ("deduplicate", deduplicate),
+    ]
+
+    skip_cjk_codes = {"zho", "jpn", "tha", "mya", "cmn"}
```
KennethEnevoldsen: not entirely sure why these are skipped?

mteb/abstasks/_data_filter/classificationefilter.py
```diff
@@ -0,0 +1,203 @@
...
+    unique_texts = set()
+    indices_to_keep = []
+    for i, text in enumerate(dataset[input_column]):
+        text = text.strip()
+        if text not in unique_texts:
+            unique_texts.add(text)
+            indices_to_keep.append(i)
+
+    logger.info(
+        f"[deduplicate] kept={len(indices_to_keep)}, removed={len(dataset) - len(indices_to_keep)}"
```
KennethEnevoldsen: wouldn't you rather want 10/1000 removed (10 duplicates removed out of all the documents 1000)

KennethEnevoldsen: (would probably do this generally across

mteb/abstasks/_data_filter/classificationefilter.py
```diff
@@ -0,0 +1,203 @@
...
+    indices = [
+        i
+        for i, text in enumerate(test_dataset[input_column])
+        if text not in train_texts
+    ]
+    logger.info(f"[filter_train_leakage] removed={before - len(indices)}")
+    return test_dataset.select(indices)
+
+
+def filter_controversial(
```
KennethEnevoldsen: name to filter_unclear_label

mteb/abstasks/_data_filter/classificationefilter.py
KennethEnevoldsen: I see that e.g. for number of words this is reasonably skipped

KennethEnevoldsen (commented):
Great with a first iteration

Should this a method on the AbsTaskClassification? I could also see the argument for not doing it (keeping it as a hidden thing until we have a decent implementation across tasks)

KennethEnevoldsen:
Not sure if this closes https://github.com/embeddings-benchmark/mteb/issues/3672

(I could see us merging in this partial solution and expanding upon it - it us currently "private" so it shouldn't cause in issues)

PR #3660, opened 2025-12-05
dataset: Add new benchmark JMTEB(v2)

lsz05:
This PR integrates the latest version of [JMTEB](https://github.com/sbintuitions/JMTEB) into MTEB.

PR #3699, opened 2025-12-08
dataset: Replace JMTEB(v2) with MTEB(jpn,v1) for general-purpose Japanese evaluation

lsz05:
As we have updated the general-purpose Japanese benchmark from MTEB(jpn,v1) to JMTEB(v2), I open this PR to reflect it to the leaderboard. 
- https://github.com/embeddings-benchmark/mteb/pull/3660

I have also added some models that have the full results on JMTEB(v2).
- https://github.com/embeddings-benchmark/results/pull/353


KennethEnevoldsen:
@lsz05 I have merged the results can I ask you to post a screenshot from the leaderboard (you can run it using `mteb run-leaderboard`)

mteb/leaderboard/benchmark_selector.py
```diff
@@ -71,7 +71,7 @@ class MenuEntry:
                         "MTEB(cmn, v1)",
                         "MTEB(deu, v1)",
                         "MTEB(fra, v1)",
-                        "MTEB(jpn, v1)",
```
Samoed, line 74: You can add previous version to `Other`, but you need to update its name https://github.com/embeddings-benchmark/mteb/blob/97834cbd0e150f7348edd7f96ba86ec0c76a2c33/mteb/benchmarks/benchmarks/benchmarks.py#L966
e.g. `Japanese Legacy`

lsz05:
> @lsz05 I have merged the results can I ask you to post a screenshot from the leaderboard (you can run it using `mteb run-leaderboard`)


スクリーンショット 2025-12-09 0 59 15 3

mteb/leaderboard/benchmark_selector.py
lsz05: As the old one `MTEB(jpn, v1)` is a subset of the new `JMTEB(v2)`, I believe it's okay to deprecate the old one.

lsz05:
↑ It seems I'm having trouble trouble loading the metadata of the benchmark.
スクリーンショット 2025-12-09 1 09 27

lsz05:
スクリーンショット 2025-12-09 1 35 29（3）
Everything looks good now (but the number of languages should be 1?).
About the legacy (`MTEB(jpn, v1)`), my opinion is that we don't need it anymore and we can remove it to reduce the burden of the server.
Seems English, Russian and Farsi have legacy remained, I wonder whether it is a policy to leave archives.

lsz05:
I got an error when I switch from `Chinese` tab to `Japanese` tab:
```
gradio.exceptions.Error: "Value: 'cmn' (type: <class 'str'>) is not in the list of choices: ['jpn']"
.../dev/mteb-251206/.venv/lib/python3.12/site-packages/pandas/core/arraylike.py:399: RuntimeWarning:

divide by zero encountered in log10
```

Samoed:
> About the legacy (`MTEB(jpn, v1)`), my opinion is that we don't need it anymore and we can remove it to reduce the burden of the server. Seems English, Russian and Farsi have legacy remained, I wonder whether it is a policy to leave archives. 

We don't have policy on it, but some users would want to see previous version of benchmark

> I got an error when I switch from Chinese tab to Japanese tab: 

this happens randomly. I've seen this one time and can't reproduce it

KennethEnevoldsen:
> We don't have policy on it, but some users would want to see previous version of benchmark

Yeah would add it

Samoed:
> ↑ It seems I'm having trouble trouble loading the metadata of the benchmark.

I think metadata wasn't updated completely and was shown for Mteb multilingual

mteb/benchmarks/benchmarks/benchmarks.py
```diff
@@ -961,42 +961,6 @@
     contacts=["KennethEnevoldsen", "isaac-chung"],
 )
 
-MTEB_JPN = Benchmark(
-    name="MTEB(jpn, v1)",
```
Samoed, line 965: You shouldn't delete it for backward compatibility

mteb/benchmarks/benchmarks/benchmarks.py
```diff
@@ -961,9 +961,9 @@
     contacts=["KennethEnevoldsen", "isaac-chung"],
 )
 
-MTEB_JPN = Benchmark(
+MTEB_JPN_CLASSIC = Benchmark(
```
Samoed: Please don't rename. This could break compatability too

mteb/benchmarks/get_benchmark.py
```diff
@@ -58,7 +59,8 @@ def _get_previous_benchmark_names() -> dict[str, str]:
         "MTEB(pol)": MTEB_POL.name,
         "MTEB(code)": MTEB_code.name,
         "MTEB(Multilingual)": MTEB_multilingual_v2.name,
-        "MTEB(jpn)": MTEB_JPN.name,
+        "JMTEB(v2)": JMTEB_V2.name,
```
Samoed: You don't need to add this
```suggestion
```

lsz05:
> this happens randomly. I've seen this one time and can't reproduce it

OK, I assume it's a problem of my hardware or env, as in my side it also happens on the main branch.

> I think metadata wasn't updated completely and was shown for Mteb multilingual

It happens with the error ↑.
The same thing happens on the main branch and I don't know why.

> but some users would want to see previous version of benchmark

I reverted it, and now it's placed under the `Other` tab.

KennethEnevoldsen:
Thanks for taking the time on this! I don't believe the error is related to this PR so we are fine to proceed.

PR #3700, opened 2025-12-08
fix: Add hamming score to multilabel classification

### LINKED ISSUE #1835: [Main & MIEB] potential issues for multi-label classification
@isaac-chung and I were debugging the weird scores for `VOC2007` in `mieb` (#1792) and found the following potential issues for both `mieb` and `main`.

1. `lrap` computation. Think `lrap` is supposed to operate on continuous scores instead of discrete predicted labels which I fixed for MIEB here #1834. This change largely smooths the large performance range across models when the `samples_per_label` is small.
Was trying to fix the same for `main` but got a chain of failed tests. Note that a difference is `main` is using `KNeighborsClassifier()` without a `MultiOutputClassifier` which I am not sure works in the expected way either; `mieb` instead is using `MultiOutputClassifier(estimator=LogisticRegression())` which treats each label as a separate binary cls problem. Will be great if someone can confirm the same for main! Also, why KNeighbors?

2. Logic for under sampling for `samples_per_label`. Both `main` and `mieb` are undersampling with a per single-label logic instead of a per set-label logic which I am not sure is what we want either. i.e.,
```python
if any((label_counter[label] < samples_per_label) for label in y[i]):
    sample_indices.append(i)
```
With a `samples_per_label=8` we can get something like the following, where some labels have much more number of examples than the other. e.g., for an example of label (18, 14), if the number of 18th < 8, even though the 14th is already 40+, this example will still be sampled and add to the training set.
```
defaultdict(int,
            {14: 49,
             13: 8,
             4: 12,
             10: 8,
             5: 9,
             2: 8,
             11: 8,
             6: 11,
             7: 9,
             18: 8,
             12: 8,
             19: 9,
             17: 9,
             8: 8,
             0: 8,
             3: 8,
             15: 8,
             9: 8,
             16: 8,
             1: 8})
```
Think this is less of a problem if using `MultiOutputClassifier()` because it only means that certain binary classifiers are trained with more examples while others are not affected.

3. The number of unique predicted labels are much fewer than unique groundtruth labels, which can make `accuracy` much lower because it's currently assessing in a strict set-label perfect matching logic. e.g., `[0, 0, 0, 1, 0]` will score `0` for `[0, 0, 0, 1, 1]` which is probably not optimal.

Image

Image

isaac-chung:
Fixes #1835

1. Added hamming_score function
2. Added unit test for hamming_score

### DIFF
--- a/mteb/_evaluators/classification_metrics.py
+++ b/mteb/_evaluators/classification_metrics.py
@@ -19,12 +19,8 @@ def hamming_score(y_true: np.ndarray, y_pred: np.ndarray) -> float:
         ValueError: If inputs are invalid or have incompatible shapes
         TypeError: If inputs cannot be converted to numpy arrays
     """
-    # Input validation
-    try:
-        y_true = np.asarray(y_true)
-        y_pred = np.asarray(y_pred)
-    except (ValueError, TypeError) as e:
-        raise TypeError(f"Inputs must be convertible to numpy arrays: {e}")
+    y_true = np.asarray(y_true)
+    y_pred = np.asarray(y_pred)
 
     # Check shapes
     if y_true.shape != y_pred.shape:


tests/test_evaluators/test_classification_metrics.py
Samoed, line 1: Maybe move to `tests/test_tasks`

mteb/abstasks/multilabel_classification.py
```diff
@@ -22,6 +22,63 @@
 logger = logging.getLogger(__name__)
 
 
+def hamming_score(y_true: np.ndarray, y_pred: np.ndarray) -> float:
```
Samoed: Maybe move to file for metrics?

mteb/abstasks/multilabel_classification.py
```diff
@@ -22,6 +22,63 @@
...
+    y_pred_bool = y_pred.astype(bool)
+
+    # Calculate intersection and union for each sample
+    intersection = (y_true_bool & y_pred_bool).sum(axis=1)
+    union = (y_true_bool | y_pred_bool).sum(axis=1)
+
+    # Handle division by zero: when union is 0, both are all zeros, so score is 1.0
+    scores = np.where(union == 0, 1.0, intersection / union)
+
+    return float(scores.mean())
```
Samoed: Maybe also store accuracy per label?

tests/test_evaluators/test_classification_metrics.py
isaac-chung: I see some retrieval metrics in the evaluator folder. Is that a better place?

Samoed: These metrics are also in evaluator folder, but generally I think that this would be a better place

mteb/abstasks/multilabel_classification.py
isaac-chung: That's not part of the agreement, please see the looked issue. But feel free to add an issue for this, and see who picks it up.

tests/test_evaluators/test_classification_metrics.py
isaac-chung: Do you mean Evaluator is a better place? I don't think tasks is suitable

Samoed: Yes, I think you can put it in evaluator

KennethEnevoldsen:
@isaac-chung how did you manage to get claude to co-author the commit?

isaac-chung:
@KennethEnevoldsen I asked Claude Code (CC) to create a commit msg, commit, then push the change

mteb/_evaluators/classification_metrics.py
```diff
@@ -0,0 +1,58 @@
...
+    The Hamming score is the fraction of labels that are correctly predicted for each sample,
+    averaged over all samples. For samples where both y_true and y_pred have no labels,
+    the score is 1.0 (perfect agreement).
+
+    Args:
+        y_true: Binary matrix of true labels with shape (n_samples, n_labels)
+        y_pred: Binary matrix of predicted labels with shape (n_samples, n_labels)
+
+    Returns:
+        float: Hamming score between 0.0 and 1.0
```
KennethEnevoldsen, line 16:
```suggestion
        Hamming score between 0.0 and 1.0
```

mteb/_evaluators/classification_metrics.py
```diff
@@ -0,0 +1,58 @@
...
+    Raises:
+        ValueError: If inputs are invalid or have incompatible shapes
+        TypeError: If inputs cannot be converted to numpy arrays
+    """
+    # Input validation
+    try:
+        y_true = np.asarray(y_true)
+        y_pred = np.asarray(y_pred)
+    except (ValueError, TypeError) as e:
+        raise TypeError(f"Inputs must be convertible to numpy arrays: {e}")
```
KennethEnevoldsen: is there a reason to have the try-except?

```suggestion
    y_true = np.asarray(y_true)
    y_pred = np.asarray(y_pred)
```

mteb/_evaluators/classification_metrics.py
isaac-chung: Ah, generated stuff.

mteb/_evaluators/classification_metrics.py
isaac-chung: Removing.

Issue #3438, opened 2025-10-20
Switch CI to UV


KennethEnevoldsen:
### Description of the feature

Currently our CI does not reveal that optional dependencies are incompatible. Using ´uv´ for the CI would resolve this (e.g. incompatibilities was introduced in #3407 and fixes in #3437)

isaac-chung:
I can take a first jab. Seems like it's already showing some conflicts:

```
mteb % uv sync --all-extras
warning: The `extra-build-dependencies` option is experimental and may change without warning. Pass `--preview-features extra-build-dependencies` to disable this warning.
Resolved 363 packages in 31.43s
warning: The package `google-cloud-aiplatform==1.71.1` does not have an extra named `all`
error: Extras `blip2` and `timm` are incompatible with the declared conflicts: {`mteb[blip2]`, `mteb[timm]`}
```

KennethEnevoldsen:
Would be great!

PR #3702, opened 2025-12-08
ci: Switch CI to use `uv`

isaac-chung:
First try at #3438

Use `uv` in all relevant make commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,15 @@
 
 ## Installation
 
-You can install mteb simply using pip. For more on installation please see the [documentation](https://embeddings-benchmark.github.io/mteb/installation/).
+You can install mteb simply using pip or uv. For more on installation please see the [documentation](https://embeddings-benchmark.github.io/mteb/installation/).
 
 ```bash
 pip install mteb
+```
+
+For faster installation, you can also use [uv](https://docs.astral.sh/uv/):
+```bash
+uv add mteb
 ```
 
 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -3,18 +3,26 @@ We welcome contributions. Please see the current open issues or open an issue yo
 
 
 ### Development Installation
-If you want to submit a dataset or in other ways contribute to MTEB, you can install the package in development mode:
+
+MTEB now uses [uv](https://docs.astral.sh/uv/) for fast dependency management. If you want to contribute to MTEB:
 
 ```bash
-# download the git repository
+# Install uv if you don't have it
+curl -LsSf https://astral.sh/uv/install.sh | sh
+
+# Clone and setup
 git clone https://github.com/embeddings-benchmark/mteb
 cd mteb
-
-# create your virtual environment and activate it
-make install
+make install  # Uses uv sync with dev dependencies
 ```
 
-This uses [make](https://www.gnu.org/software/make/) to define the install command. You can see what each command does in the [makefile](https://github.com/embeddings-benchmark/mteb/blob/main/Makefile).
+The `make install` command now uses uv under the hood for faster dependency resolution. You can also install manually with uv:
+
+```bash
+uv sync --extra image --group dev
+```
+
+This uses [make](https://www.gnu.org/software/make/) to define the install command. You can see what each command does in the [makefile](https://github.com/embeddings-benchmark/mteb/blob/main/Makefile). All commands now use `uv run` and `uv sync` for better performance and reliability.
 
 ### Running Tests
 To run the tests, you can use the following command:
--- a/docs/advanced_usage/retrieval_backend.md
+++ b/docs/advanced_usage/retrieval_backend.md
@@ -3,7 +3,15 @@
 !!! note "Available since 2.3.0"
     This feature was introduced in version **2.3.0**.
 
-For some large dataset search can take a lot of time and memory. To reduce this you can use `FaissSearchIndex`. To work with it install `pip install mteb[faiss]`.
+For some large dataset search can take a lot of time and memory. To reduce this you can use `FaissSearchIndex`. To work with it install:
+
+```bash
+# pip
+pip install mteb[faiss-cpu]
+
+# or uv
+uv add "mteb[faiss-cpu]"
+```
 
 Usage example:
 ```python
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -16,8 +16,43 @@ Installation is as simple as:
 
 If you want to run certain models implemented within mteb you will often need some additional dependencies. These can be installed using:
 
-```bash
-pip install mteb[cohere]
-```
+=== "pip"
+    ```bash
+    pip install mteb[cohere]
+    ```
+
+=== "uv"
+    ```bash
+    uv add "mteb[cohere]"
+    ```
 
 If a specific model requires a dependency it will raise an error with the recommended installation. To see full list of available models you can look at the [models overview](./overview/available_models/text.md).
+
+## Migrating to uv (for Contributors)
+
+If you're a contributor currently using pip, here's how to migrate to uv for faster dependency management:
+
+### Why uv?
+- **Faster**: 10-100x faster dependency resolution
+- **Reliable**: Deterministic builds with uv.lock
+- **Simpler**: One tool for virtual environments and packages
+
+### Migration Steps
+1. **Install uv**:
+   ```bash
+   curl -LsSf https://astral.sh/uv/install.sh | sh
+   ```
+
+2. **Replace your workflow**:
+   - `pip install -e .` → `uv sync`
+   - `pip install mteb[extra]` → `uv sync --extra extra`
+   - `python -m pytest` → `uv run pytest`
+
+### Development Groups
+For contributors, uv provides organized dependency groups:
+
+- `uv sync --group test` - Install test dependencies
+- `uv sync --group docs` - Install documentation dependencies
+- `uv sync --group typing` - Install type checking dependencies
+- `uv sync --group lint` - Install linting dependencies
+- `uv sync --group dev` - Install all development dependencies (recommended)
--- a/docs/mieb/readme.md
+++ b/docs/mieb/readme.md
@@ -17,7 +17,11 @@ If you’re already familiar with how MTEB works, then run any benchmark, task, 
 ### Run MIEB in 2 lines via CLI
 First, install the `mieb` dependencies:
 ```sh
+# pip
 pip install mteb[image]
+
+# or uv
+uv add "mteb[image]"
 ```
 
 Then, run the multilingual benchmark with a selected model, e.g. CLIP:
--- a/docs/usage/defining_the_model.md
+++ b/docs/usage/defining_the_model.md
@@ -23,7 +23,7 @@ openai_models = [meta for meta in model_metas if "openai" in meta.name]
 
 !!! tip
     Some models require additional dependencies to run on MTEB. An example of such a model is the OpenAI APIs.
-    These dependencies can be installed using `pip install mteb[openai]`
+    These dependencies can be installed using `pip install mteb[openai]` or `uv add "mteb[openai]"`
 
 ## Using a Sentence Transformer Model
 
--- a/docs/usage/get_started.md
+++ b/docs/usage/get_started.md
@@ -48,7 +48,7 @@ results = mteb.evaluate(model, tasks=tasks)
 MTEB is not only text evaluating, but also allow you to evaluate image and image-text embeddings.
 
 !!! Note
-    Running MTEB on images requires you to install the optional dependencies using `pip install mteb[image]`
+    Running MTEB on images requires you to install the optional dependencies using `pip install mteb[image]` or `uv add "mteb[image]"`
 
 To evaluate image embeddings you can follow the same approach for any other task in `mteb`. Simply ensuring that the task contains the modality "image":
 
--- a/docs/usage/leaderboard.md
+++ b/docs/usage/leaderboard.md
@@ -40,7 +40,7 @@ Alternatively, you can use the Makefile:
 make run-leaderboard
 ```
 
-The leaderboard requires gradio install, which can be installed using `pip install mteb[leaderboard]` and requires python >3.10.
+The leaderboard requires gradio install, which can be installed using `pip install mteb[leaderboard]` or `uv add "mteb[leaderboard]"` and requires python >3.10.
 
 ### Annotate Contamination
 
--- a/docs/usage/running_the_evaluation.md
+++ b/docs/usage/running_the_evaluation.md
@@ -160,7 +160,11 @@ A last option is to make a [custom implementation](./defining_the_model.md#using
 The simplest way to speed up downloads is by using Huggingface's [`xet`](https://huggingface.co/blog/xet-on-the-hub). You can use this simply using:
 
 ```bash
+# pip
 pip install mteb[xet]
+
+# or uv
+uv add "mteb[xet]"
 ```
 
 For one of the larger datasets, `MrTidyRetrieval` (~15 GB), we have seen speed-ups from ~40 minutes to ~30 minutes while using `xet`.


.github/workflows/test.yml
```diff
@@ -56,11 +59,10 @@ jobs:
           path: ~/.cache/huggingface
           key: ${{ runner.os }}-hf
 
-      - name: Setup Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6
+      - name: Install uv and set the Python version
+        uses: astral-sh/setup-uv@v7
         with:
           python-version: ${{ matrix.python-version }}
```
Samoed, line 62: Should we change to setup python from uv in all other actions?

Makefile
```diff
@@ -1,12 +1,14 @@
 install:
 	@echo "--- 🚀 Installing project dependencies ---"
-	pip install -e ".[image]" --group dev
+	pip install uv
```
Samoed: I don't think that `pip install uv` is needed there. I think this can produce some conflicts

isaac-chung: It's a draft. Unless you'd like to share something you have ready, I'd prefer if we hold off on commenting until later.

isaac-chung:
When [introducing Python 3.14 to tests](https://github.com/embeddings-benchmark/mteb/actions/runs/19139930620/job/54701873969), torch==2.9.0 was installed (which made tests pass in 3.14). Here, uv installed torch==2.8.0 because of a conflict with pylate:
```
 - Your project includes pylate>=1.3.1 (for certain embedding models)
  - pylate depends on fast-plaid>=1.2.3.260
  - fast-plaid constrains torch to <=2.8.0
  - You're trying to upgrade to torch>=2.9.0

  ❌ Conflict Chain:

  mteb[pylate] → pylate>=1.3.1 → fast-plaid → torch<=2.8.0
                                       ❌ conflicts with torch>=2.9.0
```

Options are:
1. Remove python 3.14, wait till pylate supports torch==2.9.0
2. Peg torch==2.9.0, but pylate will break

Samoed:
Probably we can request pylate team to release version with fast plaid support 2.9. They have this this fix in main, but didn't released yet.

Overall I think we can test pylate independently of other tests to have newer torch version in tests.

I've looked to dependencies in ci and for 3.13 pylate even is not installed for some reason (in our current ci too)

isaac-chung:
@KennethEnevoldsen what do you think?

KennethEnevoldsen:
Can't we do `package; python_version < '3.14'` and then update the error message for pylate to include that it requires python <3.14

Samoed:
It's already won't install for python > 3.13. 

https://github.com/embeddings-benchmark/mteb/blob/ce9371c50b6947cd23afb8b579d975d57ace91e8/pyproject.toml#L77

Problem that if we try to install it, then `uv` would try to sync everythng even if this is not possible. Probably this is bug on their side

KennethEnevoldsen:
Hmm I kinda agree with `uv` that it is a problem though. It is kinda odd that not all optional dependencies support 3.14

It seems like they do allow for torch==2.9.0?

https://github.com/lightonai/fast-plaid/blob/c139ec4cff698604760fc4b2ed17cfbe30c3e2af/pyproject.toml#L6

It might be due to this upper bound in pylate:
```
"fast-plaid>=1.2.4.260,<=1.3.0.290",
```
https://github.com/lightonai/pylate/blob/1def5bf6c65ae6a4f7a4a9fd4653305755b65507/pyproject.toml#L27

edit: shouldn't be a problem problem since plaid is v1.2.5

Samoed:
Yes, I found same https://github.com/embeddings-benchmark/mteb/pull/3702#issuecomment-3630647340

isaac-chung:
Ah, let me try removing the pylate / fast-plaid upper bound?

Samoed:
This won't help, because released pylate don't support 3.14

isaac-chung:
Could we just rollback 3.14 for now? Or is it a hard blocker?

isaac-chung:
@KennethEnevoldsen @Samoed I propose to temporarily remove testing for 3.14 due to this issue. The original 3.14 PR would not have passed because of the version incompatibility, and would not have been merged. So we should reopen the 3.14 issue if it exists.

Samoed:
Do we want to switch CI to `uv`? I don't see speed ups in installing dependencies

isaac-chung:
The speed up is about 2x compared to [this action](https://github.com/embeddings-benchmark/mteb/actions/runs/20428847772/job/58694884695) for instance. 

Another big benefit is the compatibility checks.

Samoed:
We can change `python_version` to `python_full_version` and this should fix problem with 3.14. Tested locally https://docs.astral.sh/uv/reference/internals/resolver/#marker-and-wheel-tag-filtering

This would install correctly
```toml
dependencies = [
    "pylate ; python_full_version < '3.13'",
    "torch==2.9.1",
]
```

isaac-chung:
> We can change `python_version` to `python_full_version` and this should fix problem with 3.14. Tested locally https://docs.astral.sh/uv/reference/internals/resolver/#marker-and-wheel-tag-filtering
> 
> This would install correctly
> 
> ```toml
> dependencies = [
>     "pylate ; python_full_version < '3.13'",
>     "torch==2.9.1",
> ]
> ```

Does this apply only to python 3.14?

Samoed:
Now problem with colpali engine

```
uv add torch==2.9.1                                                                                                                                                                                                                                                             
warning: The `extra-build-dependencies` option is experimental and may change without warning. Pass `--preview-features extra-build-dependencies` to disable this warning.                                                                                                        
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.13.*' and sys_platform == 'linux'; included: mteb[colpali-engine], mteb[colqwen3], mteb[jina-v4], mteb[model2vec], mteb[timm]; excluded: mteb[blip2],
  │ mteb[llama-embed-nemotron], mteb[llm2vec], mteb[pylate]):
  ╰─▶ Because colpali-engine==0.3.12 depends on torch>=2.5.0,<2.8.0 and only the following versions of colpali-engine{python_full_version < '3.14'} are available:
          colpali-engine{python_full_version < '3.14'}<=0.3.12
          colpali-engine{python_full_version < '3.14'}==0.3.13
      we can conclude that colpali-engine{python_full_version < '3.14'}>=0.3.12,<0.3.13 depends on torch>=2.5.0,<2.8.0.
      And because colpali-engine==0.3.13 depends on torch>=2.2.0,<2.9.0 and mteb[colpali-engine] depends on colpali-engine{python_full_version < '3.14'}>=0.3.12, we can conclude that mteb[colpali-engine] depends on torch>=2.2.0,<2.9.0.
      And because your project depends on torch==2.9.1 and your project requires mteb[colpali-engine], we can conclude that your project's requirements are unsatisfiable.

      hint: While the active Python version is 3.14, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
  help: If you want to add the package regardless of the failed resolution, provide the `--frozen` flag to skip locking and syncing.

```

isaac-chung:
Feel free to push to the branch, as I cannot reproduce your findings in the quoted message above.

Samoed:
Strange fix, but it works

isaac-chung:
Great stuff! Happy to merge?

Makefile
```diff
@@ -1,12 +1,12 @@
 install:
 	@echo "--- 🚀 Installing project dependencies ---"
-	pip install -e ".[image]" --group dev
+	uv sync --extra image --group dev
 	pre-commit install
 
 install-for-tests:
```
KennethEnevoldsen, line 6: should we just move this to default install (it is so fast now that we can simplify)

Makefile
```diff
@@ -23,12 +23,12 @@ lint-check:
 
 test:
 	@echo "--- 🧪 Running tests ---"
-	pytest -n auto -m "not (test_datasets or leaderboard_stability)"
+	uv run pytest -n auto -m "not (test_datasets or leaderboard_stability)"
```
KennethEnevoldsen: we need to add the extra and group flags to all `uv run` where needed (it will install the required dependencies)

Makefile
```diff
@@ -37,48 +37,48 @@ pr:
...
-	python -m mkdocs build
+	uv run python -m mkdocs build
 
 
 build-docs-overview:
 	@echo "--- 📚 Building documentation overview ---"
-	python docs/overview/create_available_tasks.py
-	python docs/overview/create_available_models.py
-	python docs/overview/create_available_benchmarks.py
+	uv run python docs/overview/create_available_tasks.py
```
KennethEnevoldsen: we can just add the docs flag and then remove the install block in the CI

pyproject.toml
```diff
@@ -74,7 +76,7 @@ jina-v4 = ["peft>=0.15.2", "transformers>=4.52.0", "torchvision>=0.22.1"]
 flash_attention = ["flash-attn>=2.6.3"]
 openai = ["openai>=1.41.0", "tiktoken>=0.8.0"]
 model2vec = ["model2vec>=0.3.0"]
-pylate = ["pylate>=1.3.1; python_version < '3.13'"]  # Required for sentence-transformers 5.0.0 compatibility (otherwise 1.1.6), pylate requires voyager, which in turn requires <=3.12
+pylate = ["pylate>=1.3.1; python_full_version < '3.13'"]  # Required for sentence-transformers 5.0.0 compatibility (otherwise 1.1.6), pylate requires voyager, which in turn requires <=3.12
```
KennethEnevoldsen, line 82: Should we add an issue here to track it

Makefile
Samoed, line 1: Probably `--no-sync` should be added to uv run before running scripts/linters, because syncing can take a lot of time and delete some dependencies. Also, pip users won't be able to use our makefile

isaac-chung: Oh right. Is there a way to be pip compatible as well? Or would we be forcing contributors to use uv?

Samoed: You can remove `uv run` and it should be compatible (at least I use it like this)

Makefile
isaac-chung: Doesn't it contradict https://github.com/embeddings-benchmark/mteb/pull/3702#discussion_r2647974872 ? "takes a lot of time"

Makefile
isaac-chung: When I removed the install block, it failed with `ModuleNotFoundError: No module named 'mteb'`

isaac-chung: https://github.com/embeddings-benchmark/mteb/actions/runs/20526721792/job/58971311792?pr=3702

Makefile
KennethEnevoldsen: > because syncing can take a lot of time and delete some dependencies

Really? I don't think I have ever seen this, and with a hot cache, it should be simply adding pointers

Makefile
KennethEnevoldsen: I don't think it does (maybe the very first time you resolve dependencies) - adding the uv lock file should basically fix that

Makefile
KennethEnevoldsen: That is because you have the `--no-sync` flag where you turn off the installation

isaac-chung: Still fails like this: https://github.com/embeddings-benchmark/mteb/actions/runs/20527875542/job/58974400744?pr=3702

Makefile
isaac-chung: Done.

.github/workflows/leaderboard_build.yml
```diff
@@ -33,15 +33,19 @@ jobs:
...
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
           python-version: '3.10'
-          cache: 'pip'
 
       - name: Install dependencies (incl. leaderboard extra)
         run: |
-          pip install ".[leaderboard]" --group dev
+          uv sync --extra leaderboard --group dev
```
KennethEnevoldsen: can be moved to `make leaderboard-build-test`

.github/workflows/lint.yml
```diff
@@ -33,15 +33,24 @@ jobs:
...
+        run: uv sync --group lint
+
+      - name: Install pre-commit hooks
+        run: uv run --no-sync --group lint pre-commit install
 
       - name: Lint
         id: lint
         run: |
-          make lint-check
+          uv run --no-sync --group lint ruff format . --check
```
KennethEnevoldsen: shouldn't this be `make lint-check` (we want to to be reproducible locally)

.github/workflows/lint.yml
```diff
@@ -33,15 +33,24 @@ jobs:
...
+          enable-cache: true
+
       - uses: actions/setup-python@v6
         with:
           python-version: "3.10"
-          cache: "pip"
 
       - name: Install dependencies
-        run: make install
+        run: uv sync --group lint
```
KennethEnevoldsen: can be moved to `lint-check`

.github/workflows/lint.yml
```diff
@@ -33,15 +33,24 @@ jobs:
...
+
+      - name: Install pre-commit hooks
+        run: uv run --no-sync --group lint pre-commit install
```
KennethEnevoldsen: I don't think this is needed

Makefile
```diff
@@ -37,56 +35,57 @@ pr:
...
 leaderboard-build-test:
 	@echo "--- 🚀 Running leaderboard build test ---"
-	pytest -n auto -m leaderboard_stability
+	uv sync --extra leaderboard --group dev
+	uv run --no-sync --group test pytest -n auto -m leaderboard_stability
 
 run-leaderboard:
 	@echo "--- 🚀 Running leaderboard locally ---"
-	python -m mteb.leaderboard.app
+	uv run --no-sync --extra leaderboard python -m mteb.leaderboard.app
```
isaac-chung: @KennethEnevoldsen maybe we should also run without `--no-sync` here?

uv.lock
Samoed, line 1: Lock file should be added to ignore, to not pin our dependencies to specific versions

> It depends. If your goal is to make CI use the same dependency versions as local development and avoid unexpected failures, you should add the pdm.lock file to version control. Otherwise, if your project is a library and you want CI to mimic the installation on user site to ensure that the current version on PyPI doesn't break anything, then do not submit the pdm.lock file.
https://pdm-project.org/latest/usage/lockfile/

isaac-chung: It's already in main :) https://github.com/embeddings-benchmark/mteb/blob/main/.gitignore#L146

Samoed: Yes, but how then this file appeared?

Makefile
```diff
@@ -6,16 +6,18 @@ install:
...
 lint-check:
 	@echo "--- 🧹 Check is project is linted ---"
 	# Required for CI to work, otherwise it will just pass
-	ruff format . --check						    # running ruff formatting
-	ruff check .    						        # running ruff linting
-	typos --diff
+	uv sync --group lint
+	uv run --no-sync --group lint ruff format . --check
+	uv run --no-sync --group lint ruff check .
+	uv run --no-sync --group lint typos --diff
```
KennethEnevoldsen:
```suggestion
	uv run --group lint ruff format . --check
	uv run --group lint ruff check .
	uv run --group lint typos --diff
```

uv.lock
isaac-chung: Intentionally added, see https://github.com/embeddings-benchmark/mteb/pull/3702#discussion_r2648561201

Makefile
```diff
@@ -65,6 +67,7 @@ dataset-load-test-pr:
 
 leaderboard-build-test:
 	@echo "--- 🚀 Running leaderboard build test ---"
+	uv sync --extra leaderboard --group dev
 	uv run --no-sync --group test pytest -n auto -m leaderboard_stability
```
KennethEnevoldsen:
```suggestion
	uv run pytest -n auto -m leaderboard_stability --group test --group dev --group leaderboard
```

Makefile
```diff
@@ -6,16 +6,18 @@ install:
 
 lint:
 	@echo "--- 🧹 Running linters ---"
-	ruff format . 			# running ruff formatting
-	ruff check . --fix --exit-non-zero-on-fix  	# running ruff linting # --exit-non-zero-on-fix is used for the pre-commit hook to work
-	typos
+	uv sync --group lint
+	uv run --no-sync --group lint ruff format . 			# running ruff formatting
+	uv run --no-sync --group lint ruff check . --fix --exit-non-zero-on-fix  	# running ruff linting # --exit-non-zero-on-fix is used for the pre-commit hook to work
+	uv run --no-sync --group lint typos
```
KennethEnevoldsen:
```suggestion
	uv run --group lint ruff format . 			# running ruff formatting
	uv run --group lint ruff check . --fix --exit-non-zero-on-fix  	# running ruff linting # --exit-non-zero-on-fix is used for the pre-commit hook to work
	uv run --group lint typos
```

uv.lock
KennethEnevoldsen: I would keep the lock file as well - both for speed but also acts as a way to keep note which versions things were run with so if something breaks you know that the versions in the lock file should work

isaac-chung:
I think that's all the comments - hope I didn't miss anything?

Makefile
Samoed: > Really? I don't think I have ever seen this, and with a hot cache, it should be simply adding pointers

Yes. It's trying to resolve all conflict dependencies in `conflicts` and I tried to do this once in maeb branch and took 2 hours and I got OOM on 15gb. Similar problem to ours https://github.com/astral-sh/uv/issues/16779

Samoed:
I still don’t think we should add a lock file, because in CI we would only have pinned dependencies, and we wouldn’t know when something breaks. I think we can remove all `uv run` commands from make file and use ruff, pytest etc directly

Makefile
KennethEnevoldsen: But that shouldn't be a problem on the sync (once the uv.lock is created)

Samoed: Yes, but we shouldn't add lock file because otherwise we won't test newer version of our dependencies https://github.com/embeddings-benchmark/mteb/pull/3702#issuecomment-3693901273

Samoed: I've tested again, and it seems that on maeb this now takes only 5 minutes (but this is on different computer, then previously was) and we can remove lock file

isaac-chung:
Think the only thing left is the uv.lock file: agree on whether to commit it or not. I'll leave it for you two to decide.

Makefile
```diff
@@ -1,34 +1,30 @@
 install:
 	@echo "--- 🚀 Installing project dependencies ---"
-	pip install -e ".[image]" --group dev
-	pre-commit install
-
-install-for-tests:
-	@echo "--- 🚀 Installing project dependencies for test ---"
 	@echo "This ensures that the project is not installed in editable mode"
```
isaac-chung, line 8: @KennethEnevoldsen also, `install` is in editable mode, whereas `install-for-tests` isn't. Which one do we want to keep? Editable no?

isaac-chung: Kept these separate.

isaac-chung:
> Think the only thing left is the uv.lock file: agree on whether to commit it or not. I'll leave it for you two to decide.

@KennethEnevoldsen I'll keep the file in. Feel free to settle this in a diff PR.

isaac-chung:
Found [this](https://github.com/astral-sh/uv/issues/9797) which might help: lock file should be committed to source control.

tomaarsen:
I believe this might have broken the Leaderboard, which now errors that  `uv` does not exist:

```
--- 🚀 Running leaderboard locally ---
uv run --extra leaderboard python -m mteb.leaderboard.app
make: uv: No such file or directory
make: *** [Makefile:80: run-leaderboard] Error 127
--- 🚀 Running leaderboard locally ---
uv run --extra leaderboard python -m mteb.leaderboard.app
make: uv: No such file or directory
make: *** [Makefile:80: run-leaderboard] Error 127
```

https://huggingface.co/spaces/mteb/leaderboard?logs=container

- Tom Aarsen

isaac-chung:
Thanks, @tomaarsen ! We'll need to install uv wherever we call that make command

isaac-chung:
https://huggingface.co/spaces/mteb/leaderboard/discussions/173

isaac-chung:
Doesn't seem like I have rights to merge.

PR #3710, opened 2025-12-11
[MAEB] CLAP Token Length Error from Fleurs

AdnanElAssadi56:
- Added manual truncation of 512 tokens in text processor.

Error:
[long_text_2e372dd9-4045-493f-96af-33117e7f99a3.txt](https://github.com/user-attachments/files/24093461/long_text_2e372dd9-4045-493f-96af-33117e7f99a3.txt)

### DIFF
--- a/mteb/models/model_implementations/clap_models.py
+++ b/mteb/models/model_implementations/clap_models.py
@@ -100,7 +100,6 @@ class ClapZeroShotWrapper(AbsEncoder):
                 return_tensors="pt",
                 padding=True,
                 truncation=True,
-                max_length=512,
             )
             inputs = {k: v.to(self.device) for k, v in inputs.items()}
 


Samoed:
I'm not sure that error is related to max length, because it setted in tokenizer config https://huggingface.co/laion/clap-htsat-fused/blob/main/tokenizer_config.json#L8

mteb/models/model_implementations/clap_models.py
```diff
@@ -95,7 +95,13 @@ def get_text_embeddings(
             inputs, disable=not show_progress_bar, desc="Processing text batches"
         ):
             texts = batch["text"]
-            inputs = self.processor(text=texts, return_tensors="pt", padding=True)
+            inputs = self.processor(
+                text=texts,
+                return_tensors="pt",
+                padding=True,
+                truncation=True,
+                max_length=512,
```
Samoed: I'm unsure if it will work correctly https://github.com/huggingface/transformers/blob/8ebfd84fa7f4d6c59f5059a439fad10ada26b3ff/src/transformers/models/clap/feature_extraction_clap.py#L217

Max length for tokenizer is set already https://huggingface.co/laion/clap-htsat-fused/blob/main/tokenizer_config.json#L8

AdnanElAssadi: That's what I initially thought as well because it worked before when fleurs was first added:
https://github.com/embeddings-benchmark/mteb/pull/2976

For some reason, it is not working now, will have to dig deeper into this. Only problem is that the dataset is too big to run locally and debug.

Samoed: Does this happen on other datasets? You can try to generate long array with numpy and try to reproduce

AdnanElAssadi: @Samoed I've tried it without truncation=true and it apparently only gives a warning without enforcing truncation:

```
long_text = "This is a test sentence. " * 200  # ~1000 words
inputs_long = processor(text=long_text, return_tensors="pt", padding=True)
print(f"  Input text length: {len(long_text)} chars")
print(f"  Token IDs shape: {inputs_long['input_ids'].shape}")
```

It gave:

```
Token indices sequence length is longer than the specified maximum sequence length for this model (1203 > 512). Running this sequence through the model will result in indexing errors
  Input text length: 5000 chars
  Token IDs shape: torch.Size([1, 1203])
 ```
 
 So, I guess passing truncation=true is necessary here.

Samoed: I agree on truncation. My comment was referring to `max_lenth`

AdnanElAssadi: Yeah sure, but before we didn't even have ```truncation=true```. I don't think it would hurt if we explicitly also state max_length.

Samoed: I'm not sure how `max_lenth` is working, so I think we can just leave `truncation=True`

Issue #3283, opened 2025-10-07
Investigate performance of `lightonai/GTE-ModernColBERT-v1`


Samoed:
Ok I think I might have a good lead
When I [do not convert to tensor to then cast to cpu/numpy](https://github.com/embeddings-benchmark/mteb/blob/55de95eff3503108f7f2a16cb5ef9976766d6063/mteb/models/model_implementations/pylate_models.py#L330) (which I do not do in my usual scripts), the results are way closer:
76,35 37.92 vs 76,34 38,03
Is there any reason why you did this casting or we can remove it (as it is handled by PyLate indexes and just add rounding noise)?
It is still not 100% equals, which is frustrating, but at this point I am not sure what could be different tbh

It should also be noted that the batch size has a slight impact, when I remove the batch_size=10 from the kwargs, I get 76,39 and 37,84. It's fairly minor and in any case the evaluation batch size should always be the same

_Originally posted by @NohTow in https://github.com/embeddings-benchmark/mteb/issues/3183#issuecomment-3376563745_

NohTow:
Yeah at this point I am not sure why there could be a difference between the two code. Thankfully without the casting the difference is not as big as before, but it is still puzzling that this is not totally equivalent

Samoed:
I'm looking now too

Samoed:
I run `colbert-ir/colbertv2.0` and got these results
| Task name | score (main) | score (v2) | score (script) | score (v2 without conversion to tensor)
| - | - | - | - | - |
| SciFact | DNF | 0.64199 | 0.6416511 | 0.64255 |
| NanoSciFactRetrieval | 0.67213 | 0.66351 | 0.66664 | 0.66664 |

NohTow:
From the results I reported, the results on SciFact are rather close, NFCorpus has a bigger gap

Samoed:
I will rerun then

Samoed:
| Task name | score (main) | score (v2) | score (script) | score (v2 without paddings)| score (v2 without tensors)|
| - | - | - | - | - | - |
| NFCorpus | DNF | 0.33083 | 0.3298156 | 0.32907 | 0.32907 |
| NanoSciFactRetrieval | 0.32085 | 0.3216 | 0.31998 | 0.32131 | 0.32131 |

Samoed:
[...]

PR #3711, opened 2025-12-11
Remove padding after pylate encoding

NohTow:
Hello,

As pointed out in the [original merging of PyLate ](https://github.com/embeddings-benchmark/mteb/pull/3183) and [the following discussion](https://github.com/embeddings-benchmark/mteb/issues/3283), after the encoding through PyLate models, there is an unnecessary padding operation because PyLate indexes handle padding on their own.

Originally, I observed some deltas, but lately when running new benches with new models, I actually have scores of 0, while with my scripts I had normal results. I checked the code and tested running with MTEB but without the padding and the scores are normal again. I cannot share a reproducible minimal code because the models are unfortunately not public for now but given that the padding is not needed and the code works great without it, I propose to remove it anyways.

I am not totally sure why this can totally _break_ some results, maybe it's the fact that we double cast it (here and in PyLate) or maybe I missed one effect of this particular padding call.

Close https://github.com/embeddings-benchmark/mteb/issues/3283

### DIFF
--- a/mteb/models/model_implementations/pylate_models.py
+++ b/mteb/models/model_implementations/pylate_models.py
@@ -328,13 +328,10 @@ class MultiVectorModel(AbsEncoder, PylateSearchEncoder):
             inputs,
             prompt_name=prompt_name,
             is_query=prompt_type == PromptType.query,
-            convert_to_tensor=True,
             **kwargs,
         )
 
-        # encode returns a list of tensors shaped (x, token_dim), pad to uniform length
-        pred = torch.nn.utils.rnn.pad_sequence(pred, batch_first=True, padding_value=0)
-        return pred.cpu().numpy()
+        return pred
 
 
 colbert_v2 = ModelMeta(


mteb/models/model_implementations/pylate_models.py
```diff
@@ -328,13 +328,10 @@ def encode(
...
             prompt_name=prompt_name,
             is_query=prompt_type == PromptType.query,
-            convert_to_tensor=True,
             **kwargs,
         )
 
-        # encode returns a list of tensors shaped (x, token_dim), pad to uniform length
-        pred = torch.nn.utils.rnn.pad_sequence(pred, batch_first=True, padding_value=0)
-        return pred.cpu().numpy()
+        return pred
```
Samoed, line 334: Is it numpy be default, right?

NohTow: A list of numpy arrays, yes!

Samoed:
From my test in https://github.com/embeddings-benchmark/mteb/issues/3283 padding didn't play any role. Only double casting caused difference

NohTow:
Yeah probably, the second padding should be a no-op.
Both were not required in any case, thanks for merging!

PR #3713, opened 2025-12-11
fix: Make `PIL` optional

Samoed:
After https://github.com/huggingface/sentence-transformers/pull/3589 PIL is optional dependency for `SentenceTransformers`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,7 +59,7 @@ mteb = "mteb.cli:main"
 [project.optional-dependencies]
 image = [
     "torchvision>0.2.1",
-    "transformers[torch-vision,vision]>=4.57.3",
+    "transformers[torch-vision,vision]",
 ]
 codecarbon = ["codecarbon>=2.0.0,<3.0.0"]
 leaderboard = [


pyproject.toml
```diff
@@ -67,34 +70,34 @@ leaderboard = [
...
-nomic = ["einops>=0.8.1"]
-ark = ["volcengine-python-sdk[ark]==3.0.2", "tiktoken>=0.8.0"]
-colpali_engine = ["colpali_engine>=0.3.12"]
-colqwen3 = ["transformers>=4.57", "torchvision>=0.22.1"]
-xet = ["huggingface_hub>=0.32.0"]
-youtu = ["tencentcloud-sdk-python-common>=3.0.1454", "tencentcloud-sdk-python-lkeap>=3.0.1451"]
-llama-embed-nemotron = ["transformers==4.51.0"]
-faiss-cpu = ["faiss-cpu>=1.12.0"]
-eager_embed = ["qwen_vl_utils>=0.0.14"]
+#peft = ["peft>=0.11.0"]
```
KennethEnevoldsen: I assume this is a mistake

mteb/models/model_implementations/voyage_v.py
```diff
@@ -12,10 +11,13 @@
...
 
+if TYPE_CHECKING:
+    from PIL import Image
+
 
 def _downsample_image(
-    image: Image.Image, max_pixels: int = 16000000, target_longest_side: int = 4000
-) -> Image.Image:
+    image: "Image.Image", max_pixels: int = 16000000, target_longest_side: int = 4000
+) -> "Image.Image":
```
KennethEnevoldsen: we can use future instead of "

KennethEnevoldsen: general comment (can we set a rule for this in ruff)

pyproject.toml
KennethEnevoldsen: it seems like there are some conflicts

Samoed: Yes, I'm doing this because `uv sync`/add takes a lot of time with all dependencies. Once I tried this on maeb and it took 1.5 hours and I got OOM

mteb/models/model_implementations/voyage_v.py
Samoed: It have only for 3.9 without autofixes. I'm not sure what is best approach, because it's a bit hacky with `__future__`

Samoed: I'll convert to future still

pyproject.toml
KennethEnevoldsen: Hmm, you should really make an issue on that. I suspect they would love to know of these extreme cases

KennethEnevoldsen (approved):

PR #3714, opened 2025-12-12
Modify summary table column's type

q275343119:
Set the field type of `model` to `markdown`, and set the field type of other fields to `auto`

Preview: https://huggingface.co/spaces/SmileXing/leaderboard

### DIFF
--- a/mteb/leaderboard/table.py
+++ b/mteb/leaderboard/table.py
@@ -26,16 +26,6 @@ def _format_scores(score: float) -> float:
     return round(score * 100, 2)
 
 
-def _get_column_types(df: pd.DataFrame) -> list[str]:
-    types = []
-    for column_name in df.columns:
-        if is_numeric_dtype(df[column_name]):
-            types.append("number")
-        else:
-            types.append("str")
-    return types
-
-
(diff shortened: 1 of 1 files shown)


mteb/leaderboard/table.py
```diff
@@ -226,7 +226,7 @@ def _apply_summary_table_styling(joint_table: pd.DataFrame) -> gr.DataFrame:
                     gmap=gmap_values.loc[mask],
                 )
 
-    column_types = _get_column_types(joint_table_style.data)
+    column_types = ["auto" for _ in joint_table_style.data.columns]
```
KennethEnevoldsen, line 219: is `_get_column_types` used elsewhere?

KennethEnevoldsen (approved):
Looks great!

mteb/leaderboard/table.py
Samoed: I checked and we don't use it in other places. I wanted to delete in this PR, but I can't change code

KennethEnevoldsen: @q275343119 once we have removed this I think we are good to merge

q275343119: I've removed it

Issue #3287, opened 2025-10-07
nvidia/NV-Embed-v2 fails during evaluation


KennethEnevoldsen:
### Describe the bug

I get the following error:

Error with model nvidia/NV-Embed-v2: cannot import name 'MISTRAL_INPUTS_DOCSTRING' from 'transformers.models.mistral.modeling_mistral' (/home/kenneth/githubs/mteb/.venv/lib/python3.12/site-packages/transformers/models/mistral/modeling_mistral.py)

### To reproduce

-

### Are you interested to contribute a fix for this bug?

No

Samoed:
I think we need to specify max transformers version for them. From their model card
```
pip uninstall -y transformer-engine
pip install torch==2.2.0
pip install transformers==4.42.4
pip install flash-attn==2.2.0
pip install sentence-transformers==2.7.0
```

isaac-chung:
Closing since the results have been merged.

Samoed:
Results for `nvidia/NV-Embed-v2` wasn't merged

ayush1298:
@Samoed Should we close this as results are merged: https://github.com/embeddings-benchmark/results/tree/main/results/nvidia__NV-Embed-v2

Samoed:
No, this is different results, and we should fix loader for the model

ayush1298:
> No, this is different results, and we should fix loader for the model

So, Is it correct that I have to write custom class for it, similar to LlamaEmbedNemotron written here: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_implementations/nvidia_models.py

Or, is there a way to add this without writing whole custom class?

PR #3715, opened 2025-12-12
fix: nv embed version

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3287


mteb/models/model_implementations/nvidia_models.py
```diff
@@ -100,8 +100,24 @@ def instruction_template(
     "MrTidyRetrieval",
 }
 
+
+def _nvembed_v2_wrapper(
```
KennethEnevoldsen: shouldn't we do this is a required optional dependency for nvembed?

Samoed:
It's working now, but I'm unsure if we want to add to pyproject these versions, because we have minimum version of `sentence_transformers>=3`, but nvembed requires 2. I tried to run with 5.2, but it produceses error.

PR #3717, opened 2025-12-12
fix: Added feature to fetch release_date in model_meta

### LINKED ISSUE #3696: Add release date to get_model_meta

It should be possible to get this from the git history of the repository.

ayush1298:
Close #3696.
If the release date is None, then this will add the release date as 1st commit in hf repository of the model.

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -335,22 +335,23 @@ class ModelMeta(BaseModel):
         return round(model_memory_mb)
 
     @staticmethod
-    def fetch_release_date(self) -> StrDate | None:
+    def fetch_release_date(model_name: str) -> StrDate | None:
         """Fetches the release date from HuggingFace Hub based on the first commit.
 
         Returns:
             The release date in YYYY-MM-DD format, or None if it cannot be determined.
         """
         try:
-            commits = list_repo_commits(repo_id=self.name, repo_type="model")
+            commits = list_repo_commits(repo_id=model_name, repo_type="model")
             if commits:
                 initial_commit = commits[-1]
                 release_date = initial_commit.created_at.strftime("%Y-%m-%d")
                 return release_date
         except RepositoryNotFoundError:
-            logger.warning(f"Model repository not found for {self.name}.")
+            logger.warning(f"Model repository not found for {model_name}.")
 
         return None
+
     def to_python(self) -> str:
         """Returns a string representation of the model."""
         return _pydantic_instance_to_code(self)


mteb/models/model_meta.py
```diff
@@ -183,6 +183,13 @@ def _check_name(cls, v: str | None) -> str | None:
                 "Model name must be in the format 'organization/model_name'"
             )
         return v
+    
+    @model_validator(mode='after')
+    def _auto_populate_release_date(self) -> 'ModelMeta':
```
Samoed: I think you shouldn't add this directly to `ModelMeta`. It's better to add to https://github.com/embeddings-benchmark/mteb/blob/b797d30580aacf4a58eaa1ac78e52137bd21f66d/mteb/models/get_model_meta.py#L176

ayush1298: > I think you shouldn't add this directly to `ModelMeta`. It's better to add to

Could you tell me what the benefit of doing it this way?

Samoed: Your method would be tritgered on each model construction, but this would be very slow and can modify existing metadata, but we want to get this only for new metadata

ayush1298: Can you check the updated one? I am not sure is still fully correct and what you want.

Samoed: No, I meant that it's better to move to function `_model_meta_from_hf_hub` that in file `get_model_meta`. Also, you can add this method to https://github.com/embeddings-benchmark/mteb/blob/b797d30580aacf4a58eaa1ac78e52137bd21f66d/mteb/models/get_model_meta.py#L215 and https://github.com/embeddings-benchmark/mteb/blob/b797d30580aacf4a58eaa1ac78e52137bd21f66d/mteb/models/get_model_meta.py#L237

ayush1298:
@Samoed , I have added in get_model_meta file as you have asked. For cross-encoder and sentence-transformer, it will use the same function of fecthing date from hf hub, right?

Samoed:
Yes, that's right

mteb/models/model_meta.py
```diff
@@ -332,6 +332,23 @@ def calculate_memory_usage_mb(self) -> int | None:
...
         return round(model_memory_mb)
 
+    def fetch_release_date(self) -> StrDate | None:
+        """Fetches the release date from HuggingFace Hub based on the first commit.
+
+        Returns:
+            The release date in YYYY-MM-DD format, or None if it cannot be determined.
+        """
+        try:
+            commits = list_repo_commits(repo_id=self.name, repo_type="model")
```
Samoed: Can you add check that repository exists?. you can catch here exception `RepositoryNotFoundError`

ayush1298: https://github.com/huggingface/huggingface_hub/blob/v1.2.3/src/huggingface_hub/hf_api.py#L3304-L3388, as per this list_repo_commits, will produce the error, so I have just added it in except.

mteb/models/get_model_meta.py
```diff
@@ -210,10 +210,13 @@ def _model_meta_from_hf_hub(model_name: str) -> ModelMeta:
         public_training_data=None,
         use_instructions=None,
     )
+    if meta.release_date is None:
+        meta.release_date = meta.fetch_release_date()
```
KennethEnevoldsen: you can move this to line 198 and avoid the if statement and just return the model meta

mteb/models/model_meta.py
```diff
@@ -332,6 +333,25 @@ def calculate_memory_usage_mb(self) -> int | None:
...
+            if commits:
+                initial_commit = commits[-1]
+                release_date = initial_commit.created_at.strftime("%Y-%m-%d")
+                return release_date
+        except RepositoryNotFoundError:
+            logger.warning(f"Model repository not found for {self.name}.")
+        except Exception as e:
+            logger.warning(f"Could not fetch release date for {self.name}: {e}")
```
KennethEnevoldsen: Can we avoid the generic exception here? This might hide if we introduce an error in the code

mteb/models/get_model_meta.py
```diff
@@ -232,6 +235,9 @@ def _model_meta_from_cross_encoder(model: CrossEncoder) -> ModelMeta:
         use_instructions=None,
         training_datasets=None,
     )
+    if meta.release_date is None:
+        meta.release_date = meta.fetch_release_date()
+    return meta
```
KennethEnevoldsen: Unsure if we want to fetch the release date from cross-encoders and sentence transformers. Here we just work on the object

mteb/models/get_model_meta.py
ayush1298: But how I could call fetch_release_date without a meta object? I can remove the if statement, and keep line 214 as it is, and then return model meta.

mteb/models/get_model_meta.py
Samoed: Maybe we can reuse model meta from `_model_meta_from_hf_hub` here and then just change loader and `is_cross_encoder`?

mteb/models/get_model_meta.py
KennethEnevoldsen:
```
ModelMeta._fetch_release_date(**kwargs)
```
Turn it into a static method

mteb/models/get_model_meta.py
KennethEnevoldsen: Yea that might be an option I suggested a refactor in #3735. Here it we could potentially add "fetch_from_hf", "compute_from_model" arguments to specify the extent we want to go to fetch the data.

ayush1298: Should I make this refactor in this PR only?

Samoed: This is a separate issue. You can resolve comments by creating static method for now

mteb/models/get_model_meta.py
```diff
@@ -210,14 +210,16 @@ def _model_meta_from_hf_hub(model_name: str) -> ModelMeta:
         public_training_data=None,
         use_instructions=None,
     )
+    return meta
 
 
 def _model_meta_from_cross_encoder(model: CrossEncoder) -> ModelMeta:
-    return ModelMeta(
+    model_name = (model.model.name_or_path,)
```
```suggestion
    model_name = model.model.name_or_path
```

ayush1298:
@Samoed Can you review again?

Samoed:
I think we can merge to main and then I will refactor a bit

Issue #3735, opened 2025-12-13
refactor `_model_meta_from_hf()` and similar to be methods on `ModelMeta`


KennethEnevoldsen:
### Description of the feature

I would refactor them to be methods like:

```python
meta.from_hf_hub("model_name", compute_metadata=True)
meta.from_sentence_transformers(sentence_trf_model, compute_metadata=True)
meta.from_cross_encoder(cross_enc_model, compute_metadata=True)
```

@Samoed wdyt?

edit: potentially add `fetch_from_hf: bool` and `compute_from_model: bool` to allow users to specify the desired directions.

Samoed:
Agree, I think this is better

Issue #3695, opened 2025-12-08
Add reference extraction to get meta


KennethEnevoldsen:
### Description of the feature

Currently it is None - could easily be the hf link

ayush1298:
@KennethEnevoldsen So, here if we are giving model-id, then its corresponding hf-link will be: 
https://huggingface.co/model_id, can we add these instead of default None

KennethEnevoldsen:
yep

PR #3737, opened 2025-12-13
feat: Add `.from_hf_hub()` and similar methods to `ModelMeta`

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3735
Close https://github.com/embeddings-benchmark/mteb/issues/3695
Close https://github.com/embeddings-benchmark/mteb/issues/3734

Move `_model_meta_from_hf_hub`, `_model_meta_from_cross_encoder`, `_model_meta_from_sentence_transformer` to `ModelMeta` class

```python

model = SentenceTransformer("Qwen/Qwen3-Embedding-0.6B", device="cpu")
meta = ModelMeta.from_sentence_transformer_model(model)
print(meta.to_dict())
# {'loader_kwargs': {}, 'name': 'Qwen/Qwen3-Embedding-0.6B', 'revision': 'c54f2e6e80b2d7b7de06f51cec4959f6b3e03418', 'release_date': None, 'languages': None, 'n_parameters': 595776512, 'memory_usage_mb': 1136, 'max_tokens': 32768, 'embed_dim': 1024, 'license': 'apache-2.0', 'open_weights': True, 'public_training_code': None, 'public_training_data': None, 'framework': ['Sentence Transformers'], 'reference': None, 'similarity_fn_name': <ScoringFunction.COSINE: 'cosine'>, 'use_instructions': None, 'training_datasets': None, 'adapted_from': None, 'superseded_by': None, 'modalities': ['text'], 'is_cross_encoder': None, 'citation': None, 'contacts': None, 'loader': 'sentence_transformers_loader'}

```

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -356,20 +356,19 @@ class ModelMeta(BaseModel):
     @classmethod
     def from_hub(
         cls,
-        model: str | None,
+        model: str,
         revision: str | None = None,
         compute_metadata: bool = True,
     ) -> Self:
-        """Generates a ModelMeta for SentenceTransformer model from hub.
+        """Generates a ModelMeta for model from HuggingFace hub.
 
         Args:
-            model: SentenceTransformer model.
+            model: Name of the model from HuggingFace hub. For example, `intfloat/multilingual-e5-large`
             revision: Revision of the model
             compute_metadata: Add metadata based on model card
 
         Returns:
             The generated ModelMeta.
-
         """
         meta = cls._from_hub(model, revision, compute_metadata)
         if _SENTENCE_TRANSFORMER_LIB_NAME not in meta.framework:


mteb/models/model_meta.py
```diff
@@ -213,9 +226,115 @@ def model_name_as_path(self) -> str:
...
+        Returns:
+            The generated ModelMeta.
+        """
+        from mteb.models import sentence_transformers_loader
+
+        loader = sentence_transformers_loader
+        revision = None
+        frameworks: list[FRAMEWORKS] = ["PyTorch"]
+        model_license = None
+        reference = "https://huggingface.co/" + model_name
```
ayush1298: Should we also check whether the link exists or not by requesting it?

Samoed: Added

mteb/deprecated_evaluator.py
```diff
@@ -39,6 +29,11 @@
 from mteb.results import TaskResult
 from mteb.types import ScoresDict
 
+if sys.version_info >= (3, 13):
+    from warnings import deprecated
+else:
+    from typing_extensions import deprecated
```
Samoed, line 35: Moved here to make other imports sortable by ruff

mteb/models/model_meta.py
```diff
@@ -213,9 +228,131 @@ def model_name_as_path(self) -> str:
...
 
-    def is_zero_shot_on(
-        self, tasks: Sequence["AbsTask"] | Sequence[str]
-    ) -> bool | None:
+    @classmethod
+    def from_hf_hub(
+        cls,
+        model_name: str | None,
+        revision: str | None = None,
+        compute_metadata: bool = True,
```
KennethEnevoldsen, line 244: Maybe refactor to `load_model=False`. The load model is to compute e.g. number of parameters from the model if it is not fetchable from the hub.

We can add `fetch_from_hub=True` for the other models

mteb/models/model_meta.py
```diff
@@ -291,18 +428,38 @@ def zero_shot_percentage(
...
+            SafetensorsParsingError,
+            GatedRepoError,
+            RepositoryNotFoundError,
+        ) as e:
+            logger.warning(
+                f"Can't calculate number of parameters for {model_name}. Got error {e}"
+            )
+            return None
+
+    def calculate_num_parameters_from_hub(self) -> int | None:
```
Samoed, line 523: Separated to public and private static to reuse during meta creation

mteb/models/model_meta.py
```diff
@@ -213,9 +228,131 @@ def model_name_as_path(self) -> str:
...
+            card = ModelCard.load(model_name)
+            card_data: ModelCardData = card.data
+            sentence_transformer_name = "sentence-transformers"
+            if (
+                card_data.library_name == sentence_transformer_name
+                or sentence_transformer_name in card_data.tags
+            ):
+                frameworks.append("Sentence Transformers")
+            else:
+                msg = "Model library not recognized, defaulting to Sentence Transformers loader."
```
KennethEnevoldsen, line 285: Should we add a link to the docs here

mteb/models/model_meta.py
```diff
@@ -213,9 +228,131 @@ def model_name_as_path(self) -> str:
...
+            loader=loader,
+            name=model_name,
+            revision=revision,
+            reference=reference,
+            release_date=None,
+            languages=None,
+            license=model_license,
+            framework=frameworks,
+            training_datasets=None,
+            similarity_fn_name=None,
```
KennethEnevoldsen, line 312: We can fetch similarity function from: https://huggingface.co/sentence-transformers/embeddinggemma-300m-medical/blob/main/config_sentence_transformers.json#L24

but if not I would be ok with assuming cosine? (I suspect it is also in the model data).

Feel free to make this a seperate issue

mteb/models/model_meta.py
```diff
@@ -213,9 +228,131 @@ def model_name_as_path(self) -> str:
...
+            n_parameters=n_parameters,
+            memory_usage_mb=memory_usage_mb,
+            max_tokens=None,
+            embed_dim=None,
```
KennethEnevoldsen: could also be fetched from metadata

mteb/models/model_meta.py
```diff
@@ -213,9 +228,131 @@ def model_name_as_path(self) -> str:
...
+            model: The CrossEncoder model
+            revision: Revision of the model
+            compute_metadata: Add metadata based on model card
+
+        Returns:
+            The generated ModelMeta
+        """
+        from mteb.models import CrossEncoderWrapper
+
+        meta = cls.from_hf_hub(model.model.name_or_path, revision, compute_metadata)
```
```suggestion
        meta = cls.from_hf_hub(model.model.name_or_path, revision, compute_metadata)
```
maybe worth splitting this into an inner `_fetch_metadata_from_hub`
and a public `fetch_from_hub`.

mteb/models/model_meta.py
```diff
@@ -321,18 +478,36 @@ def calculate_memory_usage_mb(self) -> int | None:
...
         return round(model_memory_mb)
 
+    def calculate_memory_usage_mb(self) -> int | None:
+        """Calculates the memory usage of the model in MB.
+
+        Returns:
+            The memory usage of the model in MB, or None if it cannot be determined.
+        """
+        if "API" in self.framework or self.name is None:
+            return None
```
KennethEnevoldsen, line 582: 1) Let us add a warning here
2) Shouldn't this be based on the number of parameters? It could have API tag while also having public weights

mteb/models/model_meta.py
KennethEnevoldsen: might be easier to just fetch from the model - but if we could avoid loading/downloading the model that would be great

KennethEnevoldsen (approved):

mteb/models/model_meta.py
Samoed: To https://embeddings-benchmark.github.io/mteb/contributing/adding_a_model/?

KennethEnevoldsen: I was thinking to `sentence_transformers_loader`, but I see that we don't have docs for that (we should probably add that, but that is another PR

mteb/models/model_meta.py
Samoed: I don't think we want to add `load_model` to `from_hf_hub`, because if not always possible to load it probably. Also, it's better to add it to `calculate_num_parameters` then, but I think we can update it later. 

I think we can rename `compute_metadata` to `fetch_from_hub`

mteb/models/model_meta.py
Samoed: Do you mean from `config.json`?

mteb/models/model_meta.py
KennethEnevoldsen: Actually, why don't we just use `SentenceTransformerEncoderWrapper` (probably what should be documented) and then remove `sentence_transformers_loader`?

mteb/models/model_meta.py
Samoed: Currently, we have api tag only for private models, so I'm not sure what to do here

mteb/models/model_meta.py
Samoed: Yes, we can just do this

mteb/models/model_meta.py
KennethEnevoldsen: Hmm, but my imagined behaviour is that if a I want to add a models then I would basically do:

```
model_meta = model.load_from_hub(...)  
```
And assuming that that should do as much as possible to make it very easy to add a model

Samoed: Overall I agree. I just don't think that would could handle this correctly in `fetch_from_hub`, because I think it should be model agnostic and handle both SentenceTransformer and CrossEncoder and work with just metadata on hub. We can create different variations as now `from_sentence_transformer` and `from_cross_encoder`. Currently if you do 

```py

model = SentenceTransformer("Qwen/Qwen3-Embedding-0.6B", device="cpu")
meta = ModelMeta.from_sentence_transformer_model(model)
print(meta.to_dict())
```
You would get correct metadata.


I just don't think that it's possible to have one method that would handle each model type correctly before loading model

mteb/models/model_meta.py
KennethEnevoldsen: Added issue on this

mteb/models/model_meta.py
KennethEnevoldsen: Yeah, I would essentially just assume SentenceTransformer, but I see that this might be better.

KennethEnevoldsen: We should probably update, the adding models docs here as well

mteb/models/model_meta.py
KennethEnevoldsen: But if it is private, then the number of parameters is None? (We have a few private functions where the number of parameters is public, but then I suppose we could also estimate the memory usage?). Though I see that might be a bit odd

Samoed:
Added to the doc
image

mteb/models/model_meta.py
Samoed: Added caclulation

mteb/models/model_meta.py
Samoed: Added `from_hub_for_sentence_transformer` maybe not best naming

mteb/models/model_meta.py
Samoed: Added to `adding a model`

Samoed:
FYI This would also close #3695 #3734 (added to `devolompent` section)

Samoed:
Problem with HF limits again or HF have some problems, but there is nothing in their status page (aws seems fine too)

mteb/models/model_meta.py
```diff
@@ -214,9 +236,199 @@ def model_name_as_path(self) -> str:
...
+            public_training_code=None,
+            public_training_data=None,
+            use_instructions=None,
+            modalities=[],
+        )
+
+    @classmethod
+    def from_sentence_transformer_model(
+        cls,
+        model: SentenceTransformer,
```
KennethEnevoldsen, line 327: could we allow just the model card here as well?

mteb/models/model_meta.py
KennethEnevoldsen: Hmm I don't think that is what we want. Why can't that just be the default for from_hub?

mteb/models/model_meta.py
Samoed: We can, but I don't think that anybody will use method with `ModelCard`. I think it's easier to use `from_hub` method.

mteb/models/model_meta.py
Samoed: Renamed

mteb/models/model_meta.py
```diff
@@ -356,7 +356,7 @@ def from_sentence_transformer_model(
     @classmethod
     def from_hub(
         cls,
-        model: str,
+        model: str | None,
```
KennethEnevoldsen: empty?

Samoed: This needed to handle cases where we can't retrive name from model. For example from `SentenceTransformers` https://github.com/embeddings-benchmark/mteb/blob/7a2724f0c4ff75dd512ff2b4811b09dfff81e85b/mteb/models/model_meta.py#L341-L345

KennethEnevoldsen: Alright let ud update the docstring to include this case, but otherwise feel free to merge

Samoed: I've made only possible to pass None only to private `_from_hub`

PR #3741, opened 2025-12-14
docs: Added changelog

KennethEnevoldsen:
- Clean up docs to prepare for adding the changelog. By adding missing links and removing references to documentation that does not exist
- Added whats new section
- Added changes from 2.0 upwards. I might be missing some

I think going forward we can just update this as well go.

### DIFF
--- a/docs/whats_new.md
+++ b/docs/whats_new.md
@@ -66,7 +66,7 @@ print(benchmark_scores_df)
 
 To make it easier to generate high quality metadata from models we created `.from_hf_hub`, `.from_sentence_transformer_model` and `.from_cross_encoder`.
 
-This does not fill out everythign, but it fills out everything that can be automated. 
+This does not fill out everything, but it fills out everything that can be automated. 
 ```
 from sentence_transformers import SentenceTransformer
 
@@ -103,7 +103,7 @@ This leads to a slight increase in performance, for example running `minishlab/p
 
 ## New in v2.2
 
-### Support for Assymetric embeddings in STS and `PairClassification`
+### Support for Asymmetric embeddings in STS and `PairClassification`
 MTEB v2.2 adds support for `prompt_type` for `STS` and `PairClassification` thus allowing for asymmetric embeddings.
 
 E.g. for `TERRa`, this allow us to add `TERRa.v2`, 


docs/whats_new.md
Samoed, line 1: Should we name it `CHANGELOG.md`?

Samoed (commented):
Should we make it autogenerated based on prs?

docs/whats_new.md
KennethEnevoldsen: But isn't this already here:

https://github.com/embeddings-benchmark/mteb/releases

I was thinking more of a way for us to ensure that we document changes relevant to the users.

Samoed: I thought about it as done in semantic release https://github.com/python-semantic-release/python-semantic-release/blob/master/CHANGELOG.rst

KennethEnevoldsen: Yeah I did generate that one (you can do it from the CLI). I think it targets two different users

isaac-chung:
Wondering how "what's new" would be accessible in docs?

And how often we want to update this?

Is what's new just a more concise subset from CHANGELOG?

KennethEnevoldsen:
@isaac-chung like this:

Screenshot 2025-12-14 at 14 26 02

KennethEnevoldsen:
> And how often we want to update this?

I would do it on relevant updates - WDYT?

isaac-chung:
We could but it seems to be manual for now.
Wondering if we could automate this as well? E.g. gather only certain tags. Idk what's relevant and what's not

CHANGELOG.md
KennethEnevoldsen, line 1: @isaac-chung and @Samoed added the autogenerated changelog

KennethEnevoldsen: We can customize then if we want to:
https://python-semantic-release.readthedocs.io/en/latest/concepts/changelog_templates.html#custom-changelogs

Then we could keep the changelog for patch updates and only use the what's new for minor updates?

isaac-chung: We can keep CHANGELOG to contain absolutely everything. 

Would you like "what's new" to be like [a release note](https://github.com/huggingface/sentence-transformers/releases/tag/v5.2.0) of some sort?

KennethEnevoldsen: but I kinda feel like this only gives the illusion of automation. You still have to write the changes, it just happens in the PR instead of in the docs, which makes it hard to change if you spot mistakes.

A PR message might also be perfectly valid to communicate a fix to a specific problem, but might not be relevant in the docs or maybe relevant, but you need to add additional context

KennethEnevoldsen: One solution for now could be to remove the overview of patch notes and then in a future PR add the automated changelog once we have figured out the automation (I tried a bit, but couldn't get it to work)

Samoed: Maybe it would be easier just to write release notes for minor versions, because I don't think that change log would be informative for other people. I'm following some repo with amount of releases as we have, and they have in releases titles of PRs, but this is very hard to understand without deep context

KennethEnevoldsen: So, remove the overview of patch notes and keep it focused on minor versions?

Samoed: Yes, I think so

KennethEnevoldsen: What are your thoughts @isaac-chung?

isaac-chung: Ya automate CHANGELOG, and write minor release notes sounds good.

KennethEnevoldsen:
Alright added autogenerated changelog and removed patch notes from `whats_new.md`

Samoed (approved):
Great!

KennethEnevoldsen:
Ahh forgot about this - will update it an finish it up

PR #3759, opened 2025-12-17
fix `bflhc/MoD-Embedding` loader

Samoed:
CC @bflhc


KennethEnevoldsen (approved):
nit: We should probably move these down to loader_kwargs
otherwise good

PR #3769, opened 2025-12-18
Add alias field in benchmark

### LINKED ISSUE #3225: Allow aliases in benchmark names

Currently, we have this dict of benchmark mapping to make it backward compatible. I think we can add alliases field to benchmark class where we can specify other benchmark names
https://github.com/embeddings-benchmark/mteb/blob/e2993452a0b9b1b9addb71f8ff7ed25f79d50825/mteb/benchmarks/get_benchmark.py#L38

ayush1298:
closes #3225


ayush1298:
@Samoed Should I remove this dict or not? 
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/benchmarks/get_benchmark.py#L47

ayush1298:
Will close this PR. Just saw that @Samoed had open a PR for this.

Issue #3545, opened 2025-11-11
Check implementation of `wav2clip`


Samoed:
`Clip` here used only in get_text_embeddings. Clip for wav2clip added automatically in their library

This is not clip that they're using in their library. They're downloading it from https://github.com/descriptinc/lyrebird-wav2clip/blob/1864b3924be5a785e2d49d975b8a26ff93f62951/wav2clip/__init__.py#L7
and it seems that they don't have a method to extract text embedding https://github.com/descriptinc/lyrebird-wav2clip/blob/1864b3924be5a785e2d49d975b8a26ff93f62951/wav2clip/model/encoder.py#L30

_Originally posted by @Samoed in https://github.com/embeddings-benchmark/mteb/pull/3497#discussion_r2476805837_

AdnanElAssadi:
> it seems that they don't have a method to extract text embedding

Yeah, I remember not being able to get it through their API, so I've followed the paper implementation. 
Why exactly do you think this is problematic?

Samoed:
That we're using CLIP to get text representation, not wav2clip

AdnanElAssadi:
But isn't Wav2Clip just using CLIP under the hood? I don't think they have their own text encoder.

Samoed:
> But isn't Wav2Clip just using CLIP under the hood

It’s tuned. I don’t think we should use CLIP and try to add text modality if it isn’t supported in the library.

Samoed:
Fixed in https://github.com/embeddings-benchmark/mteb/pull/3849

PR #3781, opened 2025-12-22
[MAEB] Wav2Clip Text Encoder

AdnanElAssadi56:
Related to https://github.com/embeddings-benchmark/mteb/issues/3545

@Samoed  I've double-checked the paper (arXiv:2110.11499v2), and it seem to confirm that we can use the standard CLIP text encoder.

1. The CLIP model is not tuned. The paper explicitly states in Section 2: "Throughout distillation, the original CLIP model weights are kept frozen". 
2. The authors note that "we get image and text modality for free". 
3. In their own evaluation (Section 3.2), they describe the process as extracting "CLIP text and Wav2CLIP audio embeddings".

Since the text encoder is identical to the standard CLIP encoder, I think we can safely get the text embeddings from the original CLIP model, and they will be mathematically aligned with the audio embeddings from wav2clip.

### DIFF
--- a/mteb/models/model_implementations/wav2clip_model.py
+++ b/mteb/models/model_implementations/wav2clip_model.py
@@ -78,13 +78,13 @@ class Wav2ClipZeroShotWrapper(AbsEncoder):
                 if wav.shape[-1] < max_length:
                     # Pad with zeros
                     pad_length = max_length - wav.shape[-1]
-                    padded_wav = torch.nn.functional.pad(wav, (0, pad_length))
+                    padded_wav = np.pad(wav, (0, pad_length), mode="constant")
                 else:
                     padded_wav = wav
                 padded_wavs.append(padded_wav)
 
-            # Stack into batch tensor
-            batch_tensor = torch.stack(padded_wavs).cpu().detach().numpy()
+            # Stack into batch array
+            batch_tensor = np.stack(padded_wavs)
 
             # Process entire batch at once
             batch_embeds = self.embed_audio(batch_tensor, self.audio_model)


mteb/models/model_implementations/wav2clip_model.py
```diff
@@ -34,12 +34,11 @@ def __init__(
         self.sampling_rate = 16_000
         self.max_audio_length_s = max_audio_length_s
 
-        # text side (CLIP)
-        self.clip = CLIPModel.from_pretrained(model_name, revision=revision).to(device)
+        # text side (CLIP) - we use the standard OpenAI CLIP model as mentioned in paper
+        clip_model_name = "openai/clip-vit-base-patch32"
```
Samoed, line 38: Can we process this without loading other model?

AdnanElAssadi: If you mean this way:

```self.clip = CLIPModel.from_pretrained(model_name, revision=revision).to(device)```

Then, no, because there is no huggingface revision for the model, and it was giving a 404 error during evaluation.

Samoed: I mean to retrieve from existing model somehow, but they're using this method for encoding and this is hard to tell how to do this with this model https://github.com/descriptinc/lyrebird-wav2clip/blob/1864b3924be5a785e2d49d975b8a26ff93f62951/wav2clip/__init__.py#L34

https://github.com/descriptinc/lyrebird-wav2clip/blob/1864b3924be5a785e2d49d975b8a26ff93f62951/wav2clip/model/encoder.py#L117-L122

Samoed, line 38: Did they state anywhere that they used this model for initialization? I found this issue https://github.com/descriptinc/lyrebird-wav2clip/issues/23 and there is no answer I think, so I'm more unsure what to do with this

AdnanElAssadi: I concluded that because they referenced: github.com/openai/CLIP in the footnotes. Also, it has 512 dimensions,  which matches the paper.

Samoed: It still can be `patch16` or `patch32`, but I think we can leave as is for now

PR #3782, opened 2025-12-22
model: add c2llm & fix f2llm

elvis-t9:

### DIFF
--- a/mteb/models/model_implementations/codefuse_models.py
+++ b/mteb/models/model_implementations/codefuse_models.py
@@ -1,6 +1,6 @@
 from mteb.models import ModelMeta
-from mteb.models.model_meta import ScoringFunction
 from mteb.models.instruct_wrapper import InstructSentenceTransformerModel
+from mteb.models.model_meta import ScoringFunction
 from mteb.types import PromptType
 
 F2LLM_CITATION = """@article{2025F2LLM,


mteb/models/model_implementations/codefuse_models.py
```diff
@@ -218,3 +345,59 @@ def instruction_template(
     training_datasets=training_datasets,
     citation=F2LLM_CITATION,
 )
+
+C2LLM_0B5 = ModelMeta(
+    loader=C2LLMWrapper,
```
Samoed: I think it's better to use `InstructSentenceTransformerModel`. You can pass your custom prompts as `prompts_dict`

elvis-t9: Sure. I have replaced C2LLMWrapper with InstructSentenceTransformerModel.

mteb/models/model_implementations/codefuse_models.py
```diff
@@ -218,3 +295,59 @@ def instruction_template(
...
+    contacts=None,
+)
+
+C2LLM_7B = ModelMeta(
+    loader=InstructSentenceTransformerModel,
+    loader_kwargs=c2llm_loader_kwargs,
+    name="codefuse-ai/C2LLM-7B",
+    revision="c1dc16d6d64eb962c783bfb36a6d9c2f24a86dca",
+    release_date="2025-12-22",
+    languages=None,
```
Samoed: Can you add supported languages?

elvis-t9: Sure! I have added the supported languages now.

PR #3785, opened 2025-12-22
save kwargs passed to get_model in model_meta

### LINKED ISSUE #3369: Save kwargs passed to `get_model` to model meta

We have some models that changing prompts and other params. This could solve an issue a bit

ayush1298:
closes #3369

### DIFF
--- a/tests/test_models/test_model_meta.py
+++ b/tests/test_models/test_model_meta.py
@@ -129,15 +129,15 @@ def test_check_training_datasets_can_be_derived(model_meta: ModelMeta):
 
 def test_loader_kwargs_persisted_in_metadata():
     model = mteb.get_model(
-        "sentence-transformers/all-MiniLM-L6-v2",
-        device="cpu",
+        "baseline/random-encoder-baseline",
+        not_existing_param=123,
     )
 
     assert hasattr(model, "mteb_model_meta")
     meta = model.mteb_model_meta
 
-    assert "device" in meta.loader_kwargs
-    assert meta.loader_kwargs["device"] == "cpu"
+    assert "not_existing_param" in meta.loader_kwargs
+    assert meta.loader_kwargs["not_existing_param"] == 123
 
 
 def test_model_to_python():


ayush1298:
@Samoed Can you tell me any model where I can test these? and also, should we update model_meta.json of all such models, or they will get updated automatically?

mteb/models/get_model_meta.py
```diff
@@ -91,6 +92,9 @@ def get_model(
         A model object
     """
     meta = get_model_meta(model_name, revision)
+    if kwargs:
+        meta = deepcopy(meta)
+        meta.loader_kwargs = {**meta.loader_kwargs, **kwargs}
```
Samoed: I think it's better to do here https://github.com/embeddings-benchmark/mteb/blob/0a0e3983df15be0eb947ef1314fe451e4b0febde/mteb/models/model_meta.py#L222

Samoed:
> Can you tell me any model where I can test these?

You can test this with any model. Just pass kwargs to the `get_model`

mteb/models/get_model_meta.py
ayush1298: Isnt that same thing only? Like updating `loader_kwargs` inside `load_model` or in `get_model`, as `get_model `eventually calls `load_model` only

Samoed: Same thing, but I think this is a better place to do this

mteb/models/get_model_meta.py
```diff
@@ -91,6 +92,8 @@ def get_model(
         A model object
     """
     meta = get_model_meta(model_name, revision)
+    if kwargs:
+        meta = deepcopy(meta)
```
```suggestion
```

ayush1298:
> > Can you tell me any model where I can test these?
> 
> You can test this with any model. Just pass kwargs to the `get_model`

Its working.

ayush1298:
@Samoed I am not getting why tests are failing here. Results are coming different, but these change has nothing to do with results

Samoed:
Between runs https://github.com/embeddings-benchmark/mteb/actions/runs/20440979195/job/58733671183 https://github.com/embeddings-benchmark/mteb/actions/runs/20440792498/job/58733033128
these libraries were updated:
- numpy 2.2.6 -> 2.4.0 (was released 2 days ago)
- networkx 3.4.2 -> 3.6.1
- scikit-learn 1.7.2 -> 1.8.0 (was released 2 weeks ago)
- scipy 1.15.3 -> 1.16.3 (was released 2 months ago)

https://www.diffchecker.com/jyEQQUaI/

I think something in scikit-learn was changed

mteb/models/model_meta.py
```diff
@@ -221,6 +221,9 @@ def load_model(self, **kwargs: Any) -> MTEBModels:
         _kwargs = self.loader_kwargs.copy()
         _kwargs.update(kwargs)
 
+        if kwargs:
+            self.loader_kwargs = _kwargs
```
```suggestion
        #if kwargs:
        #    self.loader_kwargs = _kwargs
```

Samoed:
Without kwargs everything seems normal

Samoed:
Checked again dependencies in https://github.com/embeddings-benchmark/mteb/actions/runs/20447872929/job/58754867706 and https://github.com/embeddings-benchmark/mteb/actions/runs/20443223889/job/58740834353 and they're the same

ayush1298:
Its fixed now, right?

Samoed:
Yes

KennethEnevoldsen (approved):

KennethEnevoldsen:
Should we add a test here before merging?

Samoed:
Yes, we need. I've mentioned this too in https://github.com/embeddings-benchmark/mteb/pull/3785#issuecomment-3683357569

ayush1298:
Should we add tests here? 
https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_models/test_model_meta.py

Samoed:
Yes, you can add it here

tests/test_models/test_model_meta.py
```diff
@@ -127,6 +127,19 @@ def test_check_training_datasets_can_be_derived(model_meta: ModelMeta):
     model_meta.get_training_datasets()
 
 
+def test_loader_kwargs_persisted_in_metadata():
+    model = mteb.get_model(
+        "sentence-transformers/all-MiniLM-L6-v2",
+        device="cpu",
```
Samoed: Can you add this check rather then device
```suggestion
        not_existing_param=123,
```

ayush1298: getting these: 
```

E           TypeError: SentenceTransformer.__init__() got an unexpected keyword argument 'not_existing_param'

mteb/models/sentence_transformer_wrapper.py:68: TypeError
```

Isn't that kwargs should be valid one?

Samoed: I think it's better to use `mteb.get_model("baseline/random-encoder-baseline")`

ayush1298: Changed to this, and its working. Isn't it should also worked with sentence_transformer, or is it just because that particular parameter was invalid?

Samoed: > is it just because that particular parameter was invalid?

Yes, problem with it

PR #3786, opened 2025-12-22
Optimize slow CLI imports

isaac-chung:
Implemented lazy loading for MTEB module imports to significantly reduce initialization time, resulting in 70% faster CLI test execution.

  ### Performance Impact

  | Metric | Before | After | Improvement |
  |--------|--------|-------|-------------|
  | Module import time | ~4.4s | ~2.0s | 55% faster |
  | test_available_tasks | ~5.1s | ~1.4s | 72% faster |
  | Overall CLI tests | ~30s | ~9s | 70% faster |

  ### Key Changes

  - Lazy loading via __getattr__ in mteb/__init__.py - Heavy dependencies only imported when accessed
  - _ModuleWrapper class - Solves module/function shadowing edge cases
  - Updated internal imports - 9 files now use from mteb import X pattern
  - Documentation compatibility - Added MTEB_BUILD_DOCS env var for mkdocstrings static analysis

  ### Test It
```python
  # Verify import time improvement
  import time
  start = time.time()
  import mteb
  print(f"Import time: {time.time() - start:.2f}s")  # Should be ~2.0s

  # Verify functionality
  tasks = mteb.get_tasks()
  print(f"Found {len(tasks)} tasks")
```

### DIFF
--- a/mteb/benchmarks/get_benchmark.py
+++ b/mteb/benchmarks/get_benchmark.py
@@ -12,6 +12,19 @@ logger = logging.getLogger(__name__)
 def _build_registry() -> dict[str, Benchmark]:
     import mteb.benchmarks.benchmarks as benchmark_module
 
+    # Explicitly import rteb_benchmarks module to ensure RTEB benchmarks are registered
+    from mteb.benchmarks.benchmarks import rteb_benchmarks
+
+    # Add RTEB benchmarks to the module's __dict__ if they're not already there
+    for attr_name in dir(rteb_benchmarks):
+        if not attr_name.startswith("_"):
+            attr = getattr(rteb_benchmarks, attr_name)
+            if (
+                isinstance(attr, Benchmark)
+                and attr_name not in benchmark_module.__dict__
+            ):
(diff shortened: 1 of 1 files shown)


Samoed (commented):
I think this change prevent importing everything that not in `_module_map`. For example, this code would fail in this branch

```python
import mteb

cache = mteb.cache.ResultCache()
```

isaac-chung:
Hmm true. I guess if we'd like users to do `import mteb`, then we cannot make these changes 🤔

isaac-chung:
I'm not too attached to this solution. This came to light when I was looking at slow CLI tests. We can close if this is not aligned with our UX goals.

KennethEnevoldsen (commented):
I don't think a cache here is the right solution. It increasing complexity and with minimal gain. I got quite decent gains on import times in a previous PR and I recall that there was more that could be done.

It is not like MTEB is an extremely huge package so I think we are over-engineering the problem here (figuring out that takes times and fixing that is the problem, not the number of files we are importing)

isaac-chung:
Let's close this puppy then.

PR #3789, opened 2025-12-23
model: add octen_models

bflhc:

### DIFF
--- a/mteb/models/model_implementations/octen_models.py
+++ b/mteb/models/model_implementations/octen_models.py
@@ -6,7 +6,9 @@ from mteb.models.models_protocols import PromptType
 def instruction_template(
     instruction: str, prompt_type: PromptType | None = None
 ) -> str:
-    if prompt_type == PromptType.document:
+    if (
+        prompt_type == PromptType.document
+    ):  # to avoid this issue: https://huggingface.co/Qwen/Qwen3-Embedding-8B/discussions/21
         return " "
     if not instruction:
         return ""


mteb/models/model_implementations/octen_models.py
```diff
@@ -0,0 +1,193 @@
+from mteb.models.instruct_wrapper import InstructSentenceTransformerModel
+from mteb.models.model_meta import ModelMeta
+from mteb.models.models_protocols import PromptType
+
+
+def instruction_template(
+    instruction: str, prompt_type: PromptType | None = None
+) -> str:
+    if prompt_type == PromptType.document:
+        return " "
```
Samoed, line 12: Is this correct that for documents empty space is added?

Samoed:
Your example in readme is a bit incorrect. You're adding `trust_remote_code=True` to `AutoModel` and `SentenceTransformers` model, but you're not using custom code for model

mteb/models/model_implementations/octen_models.py
bflhc: Yes， becase of this issue： https://huggingface.co/Qwen/Qwen3-Embedding-8B/discussions/21

Samoed: Wow. Can you add link this line then

bflhc:
> Your example in readme is a bit incorrect. You're adding `trust_remote_code=True` to `AutoModel` and `SentenceTransformers` model, but you're not using custom code for model

Good suggestion, It has been fixed

mteb/models/model_implementations/octen_models.py
bflhc: Of course! I have update this issue link

Issue #2580, opened 2025-04-26
Initial load and refresh of the leaderboard takes more than 30s


KennethEnevoldsen:
I tested without it for over a couple of days, but it consistently takes a while to reload. @x-tabdeveloping, do you have any ideas on what is happening here?

isaac-chung:
Hey, I've been tinkering a bit with this. Here is the rough breakdown with the current `main` branch when creating a fresh LB from cached results from disk:

```
INFO:mteb.leaderboard.app:=== Starting leaderboard app initialization ===
INFO:mteb.leaderboard.app:Step 1/7: Loading all benchmark results...
INFO:mteb.leaderboard.app:Loading cached results from disk...
INFO:mteb.leaderboard.app:Loaded cached results in 19.62s
INFO:mteb.leaderboard.app:Step 1/7 complete: Loaded results in 19.62s
INFO:mteb.leaderboard.app:Step 2/7: Fetching benchmarks...
INFO:mteb.leaderboard.app:Step 2/7 complete: Fetched 59 benchmarks in 0.00s
INFO:mteb.leaderboard.app:Step 3/7: Processing all benchmarks (select_tasks + join_revisions)...
INFO:mteb.leaderboard.app:Step 3/7 complete: Processed 59 benchmarks in 121.28s (avg 2.06s/benchmark)
INFO:mteb.leaderboard.app:Step 4/7: Filtering models...
INFO:mteb.leaderboard.app:Step 4/7 complete: Filtered 23053 models in 1.56s
INFO:mteb.leaderboard.app:Step 5/7: Generating tables...
INFO:mteb.leaderboard.app:Step 5/7 complete: Generated tables in 111.13s
INFO:mteb.leaderboard.app:Step 6/7: Creating Gradio components...
INFO:mteb.leaderboard.app:Step 6/7 complete: Created Gradio components in 18.06s
INFO:mteb.leaderboard.app:Step 7/7: Building Gradio interface and callbacks...
INFO:mteb.leaderboard.app:Step 7/7 complete: Built Gradio interface in 0.33s
INFO:mteb.leaderboard.app:Starting prerun on all benchmarks to populate caches...
INFO:mteb.leaderboard.app:Prerun complete: Processed 59 benchmarks in 15.70s
INFO:mteb.leaderboard.app:=== Leaderboard app initialization complete in 287.69s ===
```

Specifically during step 3 "benchmark processing" (in a separate, slower test run):
```
[...]

PR #3790, opened 2025-12-23
Add leaderboard timing logs and join_revisions() speedups

isaac-chung:
First step to fix #2580

1. Added timing logs for each step of the leaderboard app building
2. join_revisions() optimizations
  a. Cached model metas
  b. Pre-parse versions
  c. Optimize groupby with vectorized operations

## Key Highlights:

  - 30% overall leaderboard initialization improvement (287s
  → 200s)
  - 66% faster benchmark processing (131s → 45s)
  - 49x speedup for join_revisions() method (85s → 1.7s)

##  Performance Summary

  | Metric | Before | After | Improvement |
  |--------|--------|-------|-------------|
  | Benchmark processing | 131.2s | 44.7s | 2.93x faster |
  | join_revisions() calls | 85.0s | 1.7s | 49x faster |
  | Overall initialization | ~287s | ~200s | 30% faster |

##  Test Snippet:

```python
  import logging
  from mteb.leaderboard.app import get_leaderboard_app

  # Enable timing logs
  logging.basicConfig(level=logging.INFO, format='%(asctime)s
  - %(message)s')

  # Build leaderboard and launch Gradio interface
  app = get_leaderboard_app()
  app.launch()
```

PR #3792, opened 2025-12-23
Optimize validate filter scores only

isaac-chung:
Second step to fix https://github.com/embeddings-benchmark/mteb/issues/2580.
Follow up from #3790

1. Consolidated set conversion to single line 
2. Replaced append-in-loop with list comprehension
3. Used model_copy() instead of expensive to_dict() + from_validated() roundtrip

### DIFF
--- a/mteb/results/task_result.py
+++ b/mteb/results/task_result.py
@@ -666,8 +666,9 @@ class TaskResult(BaseModel):
             logger.warning(
                 f"{task.metadata.name}: Missing splits {set(splits) - seen_splits}"
             )
-        # Avoid to_dict() roundtrip - use model_copy for better performance
-        return self.model_copy(update={"scores": new_scores})
+        data = self.model_dump()
+        data["scores"] = new_scores
+        return type(self).model_construct(**data)
 
     def is_mergeable(
         self,


Samoed (approved):

mteb/results/task_result.py
```diff
@@ -633,21 +633,23 @@ def validate_and_filter_scores(self, task: AbsTask | None = None) -> Self:
...
-                if _scores["hf_subset"] not in hf_subsets:
-                    continue
-                new_scores[split].append(_scores)
+            # Use list comprehension for better performance
+            new_scores[split] = [
+                _scores
+                for _scores in self.scores[split]
+                if _scores["hf_subset"] in hf_subsets
+            ]
+            for _scores in new_scores[split]:
```
KennethEnevoldsen, line 650: this just does the loop twice - is that really faster?

mteb/results/task_result.py
```diff
@@ -664,9 +666,9 @@ def validate_and_filter_scores(self, task: AbsTask | None = None) -> Self:
             logger.warning(
                 f"{task.metadata.name}: Missing splits {set(splits) - seen_splits}"
             )
-        new_res = {**self.to_dict(), "scores": new_scores}
-        new_res = TaskResult.from_validated(**new_res)
-        return new_res
+        data = self.model_dump()
+        data["scores"] = new_scores
+        return type(self).model_construct(**data)
```
KennethEnevoldsen, line 671: I am not at all sure that this is faster - we wrote `from_validated` specifically to speed up data load and avoid validation. I would like like us to double check this PR

Samoed: From validated would just use model construct too https://github.com/embeddings-benchmark/mteb/blob/d3bc4cc043ed184d151d59dbed3a945ca4bb26a5/mteb/results/task_result.py#L590

But we can keep `from_validated` for consistency

KennethEnevoldsen: Right it is the same - it is only for ModelResult and BenchmarkResult that it is different. So we assume that self.model_dump is faster than `to_dict()`

Samoed: `to_dict` is `model_dump` too https://github.com/embeddings-benchmark/mteb/blob/d3bc4cc043ed184d151d59dbed3a945ca4bb26a5/mteb/results/task_result.py#L280-L286

In this part main change dict merging `{**self.to_dict(), "scores": new_scores}` and just `data["scores"] = new_scores`

KennethEnevoldsen: I would suspect as much, which is why I am not sure if this PR actually leads to a time saving that is worth the edit (it might, but I am not at all sure and if it does I would love to learn where the time saving is)

isaac-chung:
The savings are 121.28s -> 20.47s in step 3. (I'll get more detailed breakdown in a sec)

So the previous timings were [also listed here](https://github.com/embeddings-benchmark/mteb/issues/2580#issuecomment-3686082232).
```
INFO:mteb.leaderboard.app:Step 3/7 complete: Processed 59 benchmarks in 121.28s (avg 2.06s/benchmark)

  Total Processing Time: 131.17 seconds for 59 benchmarks

  1. join_revisions() - PRIMARY BOTTLENECK (65% of time, 84.96s)
      - Fetches all ModelMeta objects 59 times (should cache once)
      - Uses deprecated pandas groupby().apply() pattern (very slow)
      - Parses version strings repeatedly for same values
      - Average: 1.44s per benchmark
  2. select_tasks() - Secondary (35% of time, 46.20s)
      - Calls validate_and_filter_scores() for every task result
      - Heavy for multi-language benchmarks (MTEB Multilingual: 15s!)
      - Average: 0.78s per benchmark
```

Current main after this PR gives:
```
2025-12-25 21:46:51,848 - PID:36445 - __main__ - INFO - Step 3/7 complete: Processed 59 benchmarks in 20.47s (avg 0.35s/benchmark)

  Total Processing Time: 23.95 seconds for 59 benchmarks

  1. select_tasks() - PRIMARY BOTTLENECK (96% of time, 23.07s)
    - Calls validate_and_filter_scores() for every task result
    - Heavy for multi-language benchmarks (MTEB Multilingual v1: 14.38s!)
    - Average: 0.39s per benchmark
  2. join_revisions() - Secondary (4% of time, 0.88s)
    - Uses pandas DataFrame operations and complex groupby logic
    - Average: 0.01s per benchmark
```

KennethEnevoldsen:
Thanks @isaac-chung! That is great to see, but I am still not sure why we get the speed-up here (from the discussion it seems like many of the changes are essentially equivalent)

PR #3794, opened 2025-12-24
feat: Add vLLM support

### LINKED ISSUE #3747: Add vLLM wrapper support

Sometimes this can be helpful to evaluate using `vLLM`, because `setence-transformers` don't support multi-gpu eval.

This can be used as
```python
from vllm import LLM

encoder = LLM(...)
mteb.run(VLLMEncoder(encoder), ...)
```

CC @noooop

noooop:
MTEB testing helps align vllm with the implementation of sentence-transformers, and identifying potential numerical precision issues is becoming increasingly important.

It's great to have MTEB support vLLM as a backend.

Fix #3747

### DIFF
--- a/docs/advanced_usage/vllm_wrapper.md
+++ b/docs/advanced_usage/vllm_wrapper.md
@@ -37,7 +37,7 @@ To use vLLM with MTEB you have to wrap the model with its respective wrapper.
         llm = vllm.LLM(...)
     ```
 
-    See more https://docs.vllm.ai/en/latest/usage/troubleshooting/#python-multiprocessing
+    See more [troubleshooting](https://docs.vllm.ai/en/latest/usage/troubleshooting/#python-multiprocessing)
 
 === "Embedding models"
     ```python


noooop:
@Samoed

I’m not very familiar with the MTEB framework—I’m unsure where the relevant code should be placed and how the interface should be designed. So for now, I’ve implemented a very, very dirtyversion in order to get feedback as early as possible.

VLLM has many, many parameters. I have highlighted the most commonly used ones for MTEB users and added a lot of comments, which may require optimization.

mteb/models/vllm_wrapper.py
```diff
@@ -0,0 +1,199 @@
...
+from mteb.abstasks.task_metadata import TaskMetadata
+from mteb.models.abs_encoder import AbsEncoder
+from mteb.types import Array, BatchedInput, PromptType
+
+logger = logging.getLogger(__name__)
+
+Dtype = Literal["half", "float16", "float", "float32", "bfloat16"]
+
+
+def vllm_loader(
```
Samoed: You can delete this, because classes can be used directly

mteb/models/vllm_wrapper.py
```diff
@@ -0,0 +1,199 @@
+import logging
+from typing import Any, Literal
+
+import torch
+from torch.utils.data import DataLoader
+from vllm.config import PoolerConfig
```
Samoed: It's better to import inside class to make vllm as optional

mteb/models/vllm_wrapper.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # TODO: support task prompt
+        prompts = [text for batch in inputs for text in batch["text"]]
+        outputs = self.llm.encode(
+            prompts, pooling_task="embed", truncate_prompt_tokens=-1
+        )
+        embeddings = torch.stack([output.outputs.data for output in outputs])
+        return embeddings
+
+    def cleanup(self):
```
Samoed: We don't use such methods, but you can leave it for now

Samoed (commented):
Overall, this looks good! I think we can extend and change further if some models would need to update something. Thank you for adding!

mteb/models/vllm_wrapper.py
noooop: VLLM v1 requires calling this cleanup function—improper cleanup could lead to a disaster.

VLLM v1 requires WORKER_MULTIPROC_METHOD = spawn, and this will also be a disaster.

(╯‵□′)╯︵┻━┻

mteb/models/vllm_wrapper.py
noooop: I haven't fully mastered Python typing yet. LOL...

mteb/models/vllm_wrapper.py
Samoed: I see. I think users can call this manually then

test/embed.py
```diff
@@ -0,0 +1,31 @@
+import os
+
+os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
+
+
+if __name__ == "__main__":
+    import mteb
+    from mteb.models.vllm_wrapper import VllmEncoderWrapper
+
```
noooop: I'm trying to figure out how to get this script running without using the strange os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn" and if __name__ == "__main__":.

Samoed: I think you can put `VLLM_WORKER_MULTIPROC_METHOD` inside init of `VllmEncoderWrapper`

noooop:
How should documents, examples, and tests be organized?

Samoed:
I don't think you need to add tests, because this is wrapper and for now no models use it. 

As for docs, you can keep as docstrings without additional documentation

mteb/models/vllm_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+        # TODO: support task prompt
+        prompts = [text for batch in inputs for text in batch["text"]]
+        outputs = self.llm.encode(
+            prompts, pooling_task="embed", truncate_prompt_tokens=-1
+        )
+        embeddings = torch.stack([output.outputs.data for output in outputs])
+        return embeddings
+
+
+class VllmCrossEncoderWrapper(CrossEncoderProtocol, VllmWrapperBase):
```
Samoed: It's better not to inherit from protocols (at least in our repo, because this can cause some checks failing, but in `vllm` tests this is solution to pass our internal checks)

```suggestion
class VllmCrossEncoderWrapper(VllmWrapperBase):
```

mteb/models/vllm_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+    """vLLM wrapper for CrossEncoder models."""
+
+    def __init__(
+        self, model_name: str, revision: str | None = None, **kwargs: Any
+    ) -> None:
+        CrossEncoderProtocol.__init__(self, model_name=model_name, revision=revision)
+        VllmWrapperBase.__init__(
```
Samoed: In `VllmWrapperBase` will be inited with 
```
runner="pooling",
convert="embed",
```
is this fine?

Samoed (commented):
Wrapper looks good. You can add `vllm` as optional dependency to `pyproject` and add check to `VllmWrapperBase` like here https://github.com/embeddings-benchmark/mteb/blob/d3bc4cc043ed184d151d59dbed3a945ca4bb26a5/mteb/models/search_encoder_index/search_indexes/faiss_search_index.py#L28-L33

mteb/models/vllm_wrapper.py
noooop: There’s a  internal check for the class in the code.
It won’t work unless inherit it.

mteb/models/vllm_wrapper.py
Samoed: I just don't know which args is required for cross-encodres. Are they the same as for encoders?

mteb/models/vllm_wrapper.py
Samoed: Yes, this is because model don't have `mteb_model_meta` attribute. If model would have `ModelMeta` for this wrappper, then this will pass check. We're adding this attribute in `mteb.get_model` https://github.com/embeddings-benchmark/mteb/blob/d3bc4cc043ed184d151d59dbed3a945ca4bb26a5/mteb/models/get_model_meta.py#L96

```python
from mteb.models.vllm_wrapper import VllmCrossEncoderWrapper
m = VllmCrossEncoderWrapper("cross-encoder/ms-marco-MiniLM-L-6-v2")
isinstance(m, mteb.CrossEncoderProtocol)
# False
meta = mteb.get_model_meta("cross-encoder/ms-marco-MiniLM-L-6-v2")
m.mteb_model_meta = meta
isinstance(m, mteb.CrossEncoderProtocol)
# True
```

mteb/models/vllm_wrapper.py
noooop: I will fix it in the next commit.

KennethEnevoldsen:
> I don't think you need to add tests, because this is wrapper and for now no models use it.

I think we should at least write one test wrapping a smallish model (even if it is not the official reference implementation)
This test can also act as the example used in the documentation for the vLLM wrapper.

Samoed:
Yes, but I'm not sure if it would install and run correctly in CI

demo/embed.py
KennethEnevoldsen, line 1: I think we should add these two examples to the documentation. My suggestion would be here:

https://embeddings-benchmark.github.io/mteb/usage/defining_the_model/#using-a-sentence-transformer-model

under the title `## using vLLM`

KennethEnevoldsen: of course, I would rewrite it without the function

mteb/models/vllm_wrapper.py
```diff
@@ -0,0 +1,267 @@
...
+            **kwargs: Additional arguments to pass to the vllm serving engine model.
+        """
+        requires_package(
+            self,
+            "vllm",
+            "Wrapper for vllm serving engine",
+            install_instruction="pip install mteb[vllm]",
+        )
+
+        import os
```
KennethEnevoldsen: move to top level import

mteb/models/vllm_wrapper.py
```diff
@@ -0,0 +1,267 @@
...
+            and len(self.model_prompts) <= 2
+            and (
+                PromptType.query.value not in self.model_prompts
+                or PromptType.document.value not in self.model_prompts
+            )
+        ):
+            logger.warning(
+                "Encode models use prompts most often need to be configured with at least 'query' and"
+                f" 'document' prompts to ensure optimal performance. Received {self.model_prompts}"
+            )
```
KennethEnevoldsen: Hmm I think that there are quite a few models that have no prompts (typically older or low-resource models)

pyproject.toml
```diff
@@ -98,6 +98,7 @@ youtu = ["tencentcloud-sdk-python-common>=3.0.1454", "tencentcloud-sdk-python-lk
 llama-embed-nemotron = ["transformers==4.51.0"]
 faiss-cpu = ["faiss-cpu>=1.12.0"]
 eager_embed = ["qwen_vl_utils>=0.0.14"]
+vllm = ["vllm"]
```
KennethEnevoldsen: Should we set a lower bound here?

demo/embed.py
```diff
@@ -0,0 +1,31 @@
...
+    MAIN_SCORE = 0.7422994752439667
+    MTEB_EMBED_TOL = 1e-4
+
+    results = get_results(model=MODEL_NAME, tasks=MTEB_EMBED_TASKS)
+
+    vllm_main_score = results[0].scores["test"][0]["main_score"]
+    print("ST:", MAIN_SCORE)
+    print("vllm:", vllm_main_score)
+
+    assert MAIN_SCORE - vllm_main_score < MTEB_EMBED_TOL
```
KennethEnevoldsen: I think I would just convert this to a test reusing one of the already downloaded models

KennethEnevoldsen: same with the case below - for both cases we can probably use dummy datasets though

KennethEnevoldsen (commented):
Great work on this! I would love to add the tests + documentation. 

It would even be great to add a blog post entry on comparing the speed between the sentence transformers implementation and the vLLM implementation, however that is def. optional.

mteb/models/vllm_wrapper.py
Samoed: This copied from `SentenceTransformerEncoderWrapper`, so I think this is fine

KennethEnevoldsen: Hmm - I would probably just delete it here

demo/embed.py
Samoed: Do we want to add like separate implementation of models? They probably won't be 1-to-1 equivalent models to `SentenceTransformers` in terms of working with prompts.

KennethEnevoldsen: No I think we should keep the reference implementation as is (only one source of truth), but we should still test the wrapper

Samoed: I'm just not sure what you want to add to the documentation. Just notify users that we have a wrapper with vllm?

KennethEnevoldsen: Yea, I could imagine people would want to see how big of a difference there is in performance when using vLLM and sentence transformers with their specific model. 

That is not something we want to show on the leaderboard, but it's still a relevant use case of`mteb`.

Samoed: So, you want example like 
```python
meta = mteb.get_model("intfloat/e5-small")
model = VllmEncoderWrapper("intfloat/e5-small")
model.mteb_model_meta = meta

mteb.evaluate(model, tasks)
```
We can adapt wrapper to use directly `ModelMeta`, but I still not sure if we want this

KennethEnevoldsen: Yep (though I don't think `model.mteb_model_meta` is needed)

Samoed: It's needed for compatibility with our protocols https://github.com/embeddings-benchmark/mteb/pull/3794#discussion_r2646884064

KennethEnevoldsen: Yeah, but eval code is just fine without it

Samoed: It's needed for retrieval task to wrap correctly https://github.com/embeddings-benchmark/mteb/blob/d3bc4cc043ed184d151d59dbed3a945ca4bb26a5/mteb/abstasks/retrieval.py#L360-L369

KennethEnevoldsen: Ohh, didn't know that. This is probably also the reason why we always get the type error on the model loader as it not specified beforehand.

This is a bit inconvenient for testing (common pattern to just use the raw wrapper)

btw. The error here seems to be outdated; it should be `EncoderProtocol,` and it probably should be rephrased to `RetrievalEvaluator expects the model to implement one of the following protocols ...`

Samoed: It would work fine if it is inherited from `AbsEncoder` or just have `mteb_model_meta` as attribute.

```python
from mteb.models import IndexEncoderSearchProtocol
from mteb.models.instruct_wrapper import InstructSentenceTransformerModel

m = InstructSentenceTransformerModel(  # Instruct wrapper don't explicitly set `mteb_model_meta` 
    "intfloat/e5-small",
    "e272f3049e853b47cb5ca3952268c6662abda68f"
)
isinstance(m, EncoderProtocol)
# True

```

Samoed: I've run typecking (mypy) for model implementations and got few related to `Protocols` for rerankers and search protocols (bm25 for example), because they're not inherited from `AbsEncoder`. I think you're using default pyright from vscode and on my experience it creates a lot of false positives and I ignore them most of the time

KennethEnevoldsen: Hmm ahh then I would just set it to `None` as it is in the `AbsEncoder` when defining the class

> pyright from vscode

Yeah, I haven't bothered changing, but saw that `ty` was out in beta so might be the time to switch over

mteb/models/vllm_wrapper.py
```diff
@@ -0,0 +1,267 @@
...
+
+logger = logging.getLogger(__name__)
+
+Dtype = Literal["half", "float16", "float", "float32", "bfloat16"]
+
+
+class VllmWrapperBase:
+    """Wrapper for vllm serving engine."""
+
+    convert = "auto"
```
KennethEnevoldsen, line 33:
```suggestion
    convert = "auto"
    mteb_model_meta: ModelMeta | None = None
```
From discussion with @Samoed above

demo/embed.py
Samoed: I tried to test run ty check on mteb, but this hard to tell how accurate it is https://github.com/embeddings-benchmark/mteb/pull/3550#issuecomment-3684549797

mteb/models/vllm_wrapper.py
noooop: vLLM does not currently support task prompts. I plan to add this feature to vLLM, and there may be a compatibility patch. 

My goal is to align vLLM as closely as possible with the SentenceTransformer implementation,  and get comparable results.

demo/embed.py
noooop: I’m not very familiar with the MTEB CI system and need some help integrating the vLLM test.
It’s estimated that vLLM will require an additional 1~2GB of dependencies. 
Installing from pip every time, the CI runs would be very slow.

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,115 @@
+## vLLM Wrapper
+
+!!! note 
+    vLLM currently only supports a limited number of models, with many model implementations having subtle differences compared to the default implementations in mteb. We are working on it.
+
+## Installation
+
+Reference: https://docs.vllm.ai/en/latest/getting_started/installation/
+
+=== "uv"
```
noooop: > I think we should add these two examples to the documentation. My suggestion would be here:
> 
> https://embeddings-benchmark.github.io/mteb/usage/defining_the_model/#using-a-sentence-transformer-model
> 
> under the title `## using vLLM`


how about docs/advanced_usage/vllm_wrapper.md

I'm not very good at naming things or writing documents (as I'm a non-native English speaker). It would be great if someone could help rename things and optimize the documentation.  Make the content of this PR fit naturally with the other parts.

demo/benchmark_vllm.py
```diff
@@ -0,0 +1,34 @@
...
+        tasks,
+        cache=None,
+        show_progress_bar=False,
+    )
+    end = time.perf_counter()
+    elapsed_time = end - start
+    main_score = results[0].scores["dev"][0]["main_score"]
+    # elapsed_time 372.5035745330001 main_score 0.66899
+    print("elapsed_time", elapsed_time, "main_score", main_score)
+    return results
```
noooop: > It would even be great to add a blog post entry on comparing the speed between the sentence transformers implementation and the vLLM implementation, however that is def. optional.

I'd like to share an example from my personal experience. 

When I tested the BAAI/bge-m3 model using T2Reranking half a year ago. 
（and fix this https://huggingface.co/datasets/mteb/T2Reranking/discussions/1 ）

I don't know why, the speed was incredibly slow.

When using batch size is 12, the gpu memory usage is 18.669 GB, and it is estimated to take 13 hours to complete.
There is not much speed improvement even when using float16. it is estimated to take 12 hours to complete.

As a result, I haven’t ever managed to finish a single run of the test.

When I use the VllmEncoderWrapper, it requires only 2.725 GB of GPU memory and completes the computation in just six minutes.

demo/embed.py
Samoed: I'll help with testing, but you can add test similarly to https://github.com/embeddings-benchmark/mteb/blob/d94da97dbc802a309a84876d9c9491a357fa0bac/tests/test_models/test_colbert.py using `sentence-transformers/all-MiniLM-L6-v2`

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,115 @@
...
+    ```bash
+    uv pip install vllm mteb
```
Samoed:
```suggestion
    uv pip install mteb[vllm]
```

demo/benchmark_vllm2.py
```diff
@@ -0,0 +1,95 @@
...
+            n_step = llm.n_step
+            elapsed_time = end - start
+            delay = elapsed_time / n_step
+
+            print(
+                f"Batchsize {batchsize}, Input_len {input_len} Throughput: "
+                f"{len(prompts) / elapsed_time:.4f} requests/s, "
+                f"{len(prompts * input_len) / elapsed_time:.4f} tokens/s, "
+                f"Latency {delay * 1000:0.2f} ms, n_step {n_step}"
+            )
```
noooop: If the comparison is fair and the methodology is reasonable, the gap between the ST implementation and the VLLM implementation is not significant.

image

X-axis: Throughput (request/s)
Y-axis: Latency, Time needed for one step (ms) <- logarithmic scale
The curve lower right is better ↘

- The throughput using float16 is approximately four times that of float32. VLLM has almost no optimization for float32 inference.
- VLLM does not use padding, while ST uses padding. For a batch of requests, if there is a significant disparity in sequence lengths, ST's performance will undoubtedly be very poor.

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,115 @@
+## vLLM Wrapper
+
+!!! note 
+    vLLM currently only supports a limited number of models, with many model implementations having subtle differences compared to the default implementations in mteb. We are working on it.
```
Samoed:
```suggestion
    vLLM currently only supports a limited number of models, with many model implementations having subtle differences compared to the default implementations in mteb. We are working on it. For full list of supported models you can refer to [vllm documentation](https://docs.vllm.ai/en/stable/models/supported_models/#pooling-models).
```

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,115 @@
...
+
+if __name__ == "__main__":
+    MODEL_NAME = "intfloat/e5-small"
+    MTEB_EMBED_TASKS = ["STS12"]
+
+    results = get_results(model=MODEL_NAME, tasks=MTEB_EMBED_TASKS)
+    print(results)
+```
+
+## Two stage reranking
```
Samoed: I think it's better to change with only cross-encoder examle

demo/benchmark_vllm.py
Samoed: I would try to test with qwen embeddings, because I have a lot of problems running it.  E.g. https://github.com/embeddings-benchmark/results/pull/369. Also one of benefit of vLLM that it supports multi-gpu evaluation, because `SentenceTransformers` still use only one gpu for evaluation

demo/benchmark_vllm2.py
Samoed: Great illustration! I think you can add labels to axis and add to docs

docs/advanced_usage/vllm_wrapper.md
KennethEnevoldsen: I can go over the docs once we have settled on content

demo/benchmark_vllm2.py
KennethEnevoldsen: Do we want to add this as a blog post (I think I would prefer that) - so it is more of a "how to compare" rather that "this is how it compares", which is harder to maintain

docs/advanced_usage/vllm_wrapper.md
KennethEnevoldsen, line 1: Note to myself for when I rewrite this:

1) add link to speeding up mteb section
2) split into API docs and usage documentation
3) remove function scope
4) rewrite install section into admonition
5) refactor heading

demo/benchmark_vllm2.py
Samoed: > blog post

Do you mean as HF post?

KennethEnevoldsen: I was thinking of doing a blog post like this:

https://squidfunk.github.io/mkdocs-material/blog/2022/09/12/blog-support-just-landed/

(easy to copy to HF if we want to)

Samoed: Oh, nice! we can add this

KennethEnevoldsen: Yeah - it is a pretty good way to e.g. announce stuff or showcase an example, while not enforcing us to keep updating it

docs/advanced_usage/vllm_wrapper.md
noooop: I’ve submitted anything I found useful. It’s best to just update this PR to get it ready.

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,176 @@
...
+    |--------+------+----------+----------|
+
+    If the model weights are stored in float32:
+    - VLLM uses float16 for inference by default to inference a float32 model, it will keep numerical precision in most cases, for it have retains relatively more Fraction bits. However, due to the smaller Exponent part (only 5 bits), some models (e.g., the Gemma family) may risk producing NaN. VLLM maintains a list models that may cause NaN values and uses bfloat16 for inference by default.
+    - Using bfloat16 for inference avoids NaN risks because its Exponent part matches float32 with 8 bits. However, with only 7 Fraction bits, numerical precision decreases noticeably.
+    - Using float32 for inference incurs no precision loss but is about four times slower than float16/bfloat16.
+    If model weights are stored in float16 or bfloat16, vLLM defaults to using the original dtype for inference.
+    Quantization: With the advancement of open-source large models, fine-tuning of larger models for tasks like embedding and reranking is increasing. Exploring quantization methods to accelerate inference and reduce GPU memory usage may become necessary.
+
+### Unpadding
```
noooop, line 117: I uploaded the Unpadding performance test results.

noooop:
I think the content of this PR is pretty solid. Let’s polish it up so it’s ready to merge.

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,176 @@
...
+The throughput using float16 is approximately four times that of float32.
+
+!!! note 
+    |--------+------+----------+----------|
+    | Format | Bits | Exponent | Fraction |
+    |--------+------+----------+----------|
+    | float32 | 32 | 8 | 23 |
+    | float16 | 16 | 5 | 10 |
+    | bfloat16 | 16 | 8 | 7 |
+    |--------+------+----------+----------|
```
Samoed: Are you using Org mode?)

Samoed:
Too many conflicts in dependencies. Probably we can move out model specific tests to different actions to simplify everything

docs/advanced_usage/vllm_wrapper.md
noooop: I’m not quite sure what you mean by Org mode.

noooop:
> Too many conflicts in dependencies. Probably we can move out model specific tests to different actions to simplify everything

vLLM has very complex dependencies; that’s why I haven’t added tests to CI.

docs/advanced_usage/vllm_wrapper.md
Samoed: This is a markdown-like language mainly used in emacs. Your tables just happen to use a similar format to it, rather than standard markdown https://orgmode.org/worg/org-tutorials/tables.html

demo/benchmark_st.py
```diff
@@ -0,0 +1,30 @@
+import time
```
noooop: Thanks for your help. This PR looks much better now. 

I’m not sure how to handle these demos. Should I delete them, or keep some for reference?

Samoed: I think you can remove them

noooop:
Is there anything I can do?

Samoed:
I don't think so. I'm waiting for @KennethEnevoldsen review

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
+## vLLM Wrapper
+
+!!! note
+    vLLM currently only supports a limited number of models, with many model implementations having subtle differences compared to the default implementations in mteb. We are working on it. For full list of supported models you can refer to [vllm documentation](https://docs.vllm.ai/en/stable/models/supported_models/#pooling-models).
```
KennethEnevoldsen:
```suggestion
    vLLM currently supports only a limited number of models, and many implementations have subtle differences compared to the default implementations in mteb (see the [overview issue](add me) for more information). For the full list of supported models, refer to the [vllm documentation](https://docs.vllm.ai/en/stable/models/supported_models/#pooling-models).
```
Adding the issue of course

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
+## vLLM Wrapper
```
KennethEnevoldsen:
```suggestion
## Running models with vLLM
```
depending on how it looks in the docs

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+=== "pip"
+    ```bash
+    pip install "mteb[vllm]"
+    ```
+=== "uv"
+    ```bash
+    uv pip install "mteb[vllm]"
+    ```
+
+For other architectures, please refer to the [vllm installation guide](https://docs.vllm.ai/en/latest/installation/).
```
KennethEnevoldsen: I am not sure what is meant here?

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+            mteb.get_task("AskUbuntuDupQuestions"),
+        )
+
+
+    if __name__ == "__main__":
+        results = run_vllm_crossencoder()
+        print(results)
+    ```
+
+## vLLM is fast with
```
KennethEnevoldsen:
```suggestion
## Why is vLLM fast?
```

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+
+By default, vLLM uses Flash Attention, which only supports float16 and bfloat16 but not float32. vLLM does not optimize inference performance for float32.
+
+<figure markdown="span">
+    ![](../images/visualizations/half_precision_inference.png)
+    <figcaption>The throughput using float16 is approximately four times that of float32.
+
+X-axis: Throughput (request/s)
+Y-axis: Latency, Time needed for one step (ms) <- logarithmic scale
+The curve lower right is better ↘
```
KennethEnevoldsen, line 91: Add short description on labels (st=sentence transformers)

KennethEnevoldsen: also for the other figure

docs/advanced_usage/vllm_wrapper.md
KennethEnevoldsen:
```suggestion
## vLLM Inference
```

KennethEnevoldsen: or just:
```suggestion
## vLLM
```

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+    def run_vllm_encoder():
+        """Evaluate a model on specified MTEB tasks using vLLM for inference."""
+        encoder = VllmEncoderWrapper(model="intfloat/e5-small")
+        return mteb.evaluate(
+            encoder,
+            mteb.get_task("STS12"),
+        )
+
+
+    if __name__ == "__main__":
```
KennethEnevoldsen, line 55: is this scope required? If so we should explain why

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+    pip install "mteb[vllm]"
+    ```
+=== "uv"
+    ```bash
+    uv pip install "mteb[vllm]"
+    ```
+
+For other architectures, please refer to the [vllm installation guide](https://docs.vllm.ai/en/latest/installation/).
+## Usage
+
```
KennethEnevoldsen, line 21: To use vLLM with MTEB you have to wrap the model with its respective wrapper.

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+For causal attention models, such as the Qwen3 reranker, the following optimizations are also applicable:
+
+- Efficient KV cache memory management via PagedAttention
+- Chunked prefill for improved memory handling during long-context processing
+- Prefix caching to accelerate repeated prompt processing
+
+VLLM’s optimizations are primarily designed for and most effective with causal language models (generative models). For full list of features you can refer to [vllm documentation](https://docs.vllm.ai/en/latest/features/).
+
+
+## API Reference
```
KennethEnevoldsen, line 153: Shouldn't we move the to the API section

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+<figure markdown="span">
+    ![](../images/visualizations/unpadding.png)
+    <figcaption>X-axis: Throughput (request/s)
+Y-axis: Latency, Time needed for one step (ms) <- logarithmic scale
+The curve lower right is better ↘
+</figcaption>
+</figure>
+
+
+Sentence Transformers suffers a noticeable drop in speed when handling requests with varied input lengths, whereas vLLM does not.
```
KennethEnevoldsen:
```suggestion
Sentence Transformers (st) suffers a noticeable drop in speed when handling requests with varied input lengths, whereas vLLM does not.
```

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+
+    If model weights are stored in float16 or bfloat16, vLLM defaults to using the original dtype for inference.
+
+    Quantization: With the advancement of open-source large models, fine-tuning of larger models for tasks like embedding and reranking is increasing. Exploring quantization methods to accelerate inference and reduce GPU memory usage may become necessary.
+
+
+
+### Unpadding
+
+By default, Sentence Transformers pads all inputs in a batch to the length of the longest one, which is undoubtedly very inefficient. VLLM avoids the use of padding entirely during inference.
```
KennethEnevoldsen:
```suggestion
By default, Sentence Transformers (st) pads all inputs in a batch to the length of the longest one, which is undoubtedly very inefficient. VLLM avoids padding entirely during inference.
```

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+Sentence Transformers suffers a noticeable drop in speed when handling requests with varied input lengths, whereas vLLM does not.
+
+### Others
+
+For models using bidirectional attention, such as BERT, VLLM offers a range of performance optimizations:
+
+- Optimized CUDA kernels, including FlashAttention and FlashInfer integration
+- CUDA Graphs and torch.compile support to reduce overhead and accelerate execution
```
KennethEnevoldsen:
```suggestion
- CUDA Graphs and `torch.compile` support to reduce overhead and accelerate execution
```

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+- Multiple quantization schemes—GPTQ, AWQ, AutoRound, INT4, INT8, and FP8—for efficient deployment
+- Continuous batching of incoming requests to maximize throughput
+
+For causal attention models, such as the Qwen3 reranker, the following optimizations are also applicable:
+
+- Efficient KV cache memory management via PagedAttention
+- Chunked prefill for improved memory handling during long-context processing
+- Prefix caching to accelerate repeated prompt processing
+
+VLLM’s optimizations are primarily designed for and most effective with causal language models (generative models). For full list of features you can refer to [vllm documentation](https://docs.vllm.ai/en/latest/features/).
```
KennethEnevoldsen:
```suggestion
vLLM’s optimizations are primarily designed for and most effective with causal language models (generative models). For the full list of features, refer to the [vllm documentation](https://docs.vllm.ai/en/latest/features/).
```

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+
+
+## API Reference
+
+vLLM has a large number of parameters; here are some commonly used ones:
```
KennethEnevoldsen:
```suggestion
```

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+
+:::mteb.models.vllm_wrapper.VllmWrapperBase
+
+For all vLLM parameters, please refer to https://docs.vllm.ai/en/latest/configuration/engine_args/.
```
KennethEnevoldsen:
```suggestion
!!! info
    For all vLLM parameters, please refer to https://docs.vllm.ai/en/latest/configuration/engine_args/.
```

tests/test_models/test_vllm.py
KennethEnevoldsen, line 1: These will be skipped in our current tests - I assume that is what we want.

KennethEnevoldsen (approved):
Sorry for delaying this - I mostly updated on the documentation.

Feel free to make the edits and merge this

docs/advanced_usage/vllm_wrapper.md
Samoed: By default, vllm build for cuda. I think it means refer to documentation for not cuda installation

KennethEnevoldsen: Ahh makes sense - let us specify that

docs/advanced_usage/vllm_wrapper.md
```diff
@@ -0,0 +1,141 @@
...
+=== "pip"
+    ```bash
+    pip install "mteb[vllm]"
+    ```
+=== "uv"
+    ```bash
+    uv pip install "mteb[vllm]"
+    ```
+
+For other architectures, please refer to the [vllm installation guide](https://docs.vllm.ai/en/latest/installation/).
```
Samoed:
```suggestion
For other architectures, please refer to the [vllm installation guide](https://docs.vllm.ai/en/latest/getting_started/installation/).
```

Samoed (approved):
Great work! Thank you for integration!

PR #3795, opened 2025-12-24
fix: Download cached results zip from cached-data branch

isaac-chung:
## Summary

Part of Optimizing leaderboard startup times. #2580 

### Changes
- Modify `_load_results()` to first try downloading `__cached_results.json.gz` from the cached-data branch
- Only fallback to full repository clone if the direct download fails
- Add gzip decompression to handle the compressed cache file
- Added comprehensive logging to track download progress and fallback behavior
- Added tests for each new component

### Benefits
- **Significantly faster startup**: Avoids expensive full repository cloning when possible
- **Reduced bandwidth**: Downloads only the compressed cache file (~MB) instead of entire results repo (~GB)
- **Better reliability**: Graceful fallback to original behavior if direct download fails
- **Enhanced debugging**: Detailed logging for troubleshooting download issues


mteb/leaderboard/app.py
```diff
@@ -39,30 +42,53 @@ def _load_results(cache: ResultCache) -> BenchmarkResults:
     start_time = time.time()
     results_cache_path = Path(__file__).parent.joinpath("__cached_results.json")
     if not results_cache_path.exists():
```
Samoed, line 70: I think you can delete check on `__cached_results.json`, because it's no longer used (since v2 I think)

isaac-chung: Which line suggests that? I must have missed it

isaac-chung: I'm actually not following. Do you mean `__cached_results.json` is not used, or something else?

Samoed: https://github.com/embeddings-benchmark/mteb/blob/6bbe2c58ead1e2046ca79f4630e6a1cb457a524d/mteb/leaderboard/app.py#L43-L44

We no longer use this cache format

Samoed: Ah, sorry. No this still can be used. That's correct

isaac-chung:
I'm not sure why leaderboard build test is failing. Investigating after Christmas 🎄

.gitignore
```diff
@@ -161,3 +161,5 @@ powermetrics_log.txt
 /docs/overview/available_models/
 /docs/overview/available_tasks/
 /docs/overview/available_benchmarks.md
+
+CLAUDE.md
```
Samoed, line 165: I think you can include it (as a separate PR), because some users would want to add something. Maybe you can add it as `AGENTS.md`, but all providers uses different templates (at least file name), so this is hard to tell what to use

isaac-chung:
Yeah! I wanted to include a few of the changes from the copilot review, nothing major.

.gitignore
isaac-chung: Will do.

mteb/leaderboard/app.py
```diff
@@ -32,37 +35,389 @@
 
 logger = logging.getLogger(__name__)
 
+
+class LogFlusher:
+    """Optimized log flushing utility with batching and conditional flushing."""
```
KennethEnevoldsen: I am not sure exactly why we want this

KennethEnevoldsen (commented):
regarding using the pre-computed `__cached_results.json.gz` and the expensive download. We recently added the `--depth=1` as default behaviour when cloning. This is of course not compressed, but I am not sure how big the difference is

Overall we add a lot of complexity here - I am unsure if it is worth it - might be, but I would want us to test for the difference and simplify down to just "default to `.json.gz`

mteb/leaderboard/app.py
```diff
@@ -32,37 +35,389 @@
...
+               - Final function completion
+               - Critical errors where immediate user feedback is needed
+    """
+    _log_flusher.flush_if_needed(force=force)
+
+
 LANGUAGE: list[str] = list({l for t in mteb.get_tasks() for l in t.metadata.languages})
 
 
+def _validate_benchmark_json(json_string: str) -> None:
```
KennethEnevoldsen: Why do we need to validate a file we made ourselves?

mteb/leaderboard/app.py
isaac-chung: Yeah, I'm skeptical about this as well. Might remove.

isaac-chung:
In a _fresh build_, we'd offload the json build in the results repo. The difference is quite huge for step 1: 57.70s vs 391.95s. So in this PR, we 1) already build the json file, 2) download less, 3) don't clone when not necessary.

This branch: Step 1: 57.70s
```
2025-12-25 20:50:17,006 - PID:77120 - __main__ - INFO - Starting leaderboard app in process 77120
2025-12-25 20:50:17,006 - PID:77120 - __main__ - INFO - === Starting leaderboard app initialization ===
2025-12-25 20:50:17,006 - PID:77120 - __main__ - INFO - Step 1/7: Loading all benchmark results...
2025-12-25 20:50:17,006 - PID:77120 - __main__ - INFO - Cached results not found, trying to download from cached-data branch...
2025-12-25 20:50:17,006 - PID:77120 - __main__ - INFO - Attempting to download from: https://raw.githubusercontent.com/embeddings-benchmark/results/cached-data/__cached_results.json.gz
2025-12-25 20:50:50,175 - PID:77120 - __main__ - INFO - HTTP request successful, content length: 92706156 bytes
2025-12-25 20:50:50,176 - PID:77120 - __main__ - INFO - Attempting gzip decompression...
2025-12-25 20:50:50,953 - PID:77120 - __main__ - INFO - Decompression successful, data length: 707225132 chars
2025-12-25 20:50:50,953 - PID:77120 - __main__ - INFO - Validating JSON structure...
2025-12-25 20:50:56,258 - PID:77120 - __main__ - WARNING - Missing expected keys in JSON: ['results']
2025-12-25 20:50:56,258 - PID:77120 - __main__ - INFO - JSON validation successful
2025-12-25 20:50:56,704 - PID:77120 - __main__ - INFO - Attempting to write to: /Users/isaac/work/mteb/mteb/leaderboard/__cached_results.json
2025-12-25 20:50:56,885 - PID:77120 - __main__ - INFO - File write successful, size: 707225132 bytes
2025-12-25 20:50:56,885 - PID:77120 - __main__ - INFO - Downloaded cached results from cached-data branch in 39.88s
2025-12-25 20:50:56,886 - PID:77120 - __main__ - INFO - Loading cached results from disk...
2025-12-25 20:50:56,886 - PID:77120 - __main__ - INFO - Opening file: /Users/isaac/work/mteb/mteb/leaderboard/__cached_results.json
2025-12-25 20:50:56,886 - PID:77120 - __main__ - INFO - File exists, size: 707225132 bytes
2025-12-25 20:50:56,886 - PID:77120 - __main__ - INFO - File opened successfully, attempting JSON parse...
2025-12-25 20:51:02,383 - PID:77120 - __main__ - INFO - JSON parsed successfully, keys: ['model_results', 'benchmark']
2025-12-25 20:51:02,383 - PID:77120 - __main__ - INFO - Attempting BenchmarkResults.from_validated...
2025-12-25 20:51:14,689 - PID:77120 - __main__ - INFO - BenchmarkResults.from_validated successful
2025-12-25 20:51:14,689 - PID:77120 - __main__ - INFO - Loaded cached results in 57.68s
2025-12-25 20:51:14,705 - PID:77120 - __main__ - INFO - Step 1/7 complete: Loaded results in 57.70s
2025-12-25 20:51:14,705 - PID:77120 - __main__ - INFO - Step 2/7: Fetching benchmarks...
2025-12-25 20:51:14,706 - PID:77120 - __main__ - INFO - Step 2/7 complete: Fetched 59 benchmarks in 0.00s
2025-12-25 20:51:14,706 - PID:77120 - __main__ - INFO - Step 3/7: Processing all benchmarks (select_tasks + join_revisions)...
```


main: Step 1: 391.95s
```
2025-12-25 20:53:22,919 - PID:80486 - __main__ - INFO - Starting leaderboard app in process 80486
2025-12-25 20:53:22,919 - PID:80486 - __main__ - INFO - === Starting leaderboard app initialization ===
2025-12-25 20:53:22,920 - PID:80486 - __main__ - INFO - Step 1/7: Loading all benchmark results...
2025-12-25 20:53:22,920 - PID:80486 - __main__ - INFO - Cached results not found, downloading from remote...
2025-12-25 20:53:22,920 - PID:80486 - mteb.cache - INFO - No results repository found in /Users/isaac/.cache/mteb/remote, cloning it from https://github.com/embeddings-benchmark/results
Cloning into 'remote'...
remote: Enumerating objects: 85603, done.
remote: Counting objects: 100% (85603/85603), done.
remote: Compressing objects: 100% (10766/10766), done.
remote: Total 85603 (delta 78748), reused 80737 (delta 74766), pack-reused 0 (from 0)
Receiving objects: 100% (85603/85603), 383.14 MiB | 1.45 MiB/s, done.
Resolving deltas: 100% (78748/78748), done.
Updating files: 100% (84641/84641), done.
2025-12-25 20:58:12,171 - PID:80486 - __main__ - INFO - Downloaded remote results in 289.25s
2025-12-25 20:59:54,868 - PID:80486 - __main__ - INFO - Loaded results from cache in 102.70s
2025-12-25 20:59:54,868 - PID:80486 - __main__ - INFO - Step 1/7 complete: Loaded results in 391.95s
2025-12-25 20:59:54,868 - PID:80486 - __main__ - INFO - Step 2/7: Fetching benchmarks...
2025-12-25 20:59:54,868 - PID:80486 - __main__ - INFO - Step 2/7 complete: Fetched 59 benchmarks in 0.00s
2025-12-25 20:59:54,868 - PID:80486 - __main__ - INFO - Step 3/7: Processing all benchmarks (select_tasks + join_revisions)...
```

Note that this is by no means replacing the existing git clone method - if any part of this fails, it goes back to git cloning with --depth=1 anyway.

Samoed:
Can we keep version as was in https://github.com/embeddings-benchmark/mteb/pull/3795/commits/186a90a05e01f7f9ce13fa1ef06b6463af8fbd04 I think it was nearly good here already? I don't think we need more functions for data downloading in `app.py`.

Also, I'm not sure if we need new tests for leaderboard? Especially with so much patching

isaac-chung:
All changes after that commit are for readability and maintainability. I don't think one huge function is best practice, nor adding functionality without tests is either. The tests are there to convince us that these new components actually work as expected. I expect unit tests for new components, if we don't do that already.

Samoed:
I don't think we need tests for data validation and http operation. I think tests other than `test_leaderboard` mock too much and don't actually doing tests. 

I don't think we need `_validate_benchmark_json`, `_write_cache_file` functions, because pydantic will validate everything and we don't need to store this json file

isaac-chung:
I can for sure remove `_validate_benchmark_json`, but I'd prefer to keep the exception handling in `_write_cache_file`, unless there is existing checks (which there isn't).

Samoed:
We can just skip saving to file system json file and just do `Model.model_validate_json`

isaac-chung:
You're right about omitting validation. The validation should have been done in the results repo. 
However, I believe `_write_cache_file` is needed for parity with current main, to write `__cached_results.json` from the zip file.

Samoed:
I digged a bit deeper and it seems that `__cached_results.json` was dumped only in `v1` version, but we removed this in `v2`

(Version from in `1.39.7`)
https://github.com/embeddings-benchmark/mteb/blob/d2c704c15a6312264822be11986372cc1f7e6c6b/mteb/leaderboard/app.py#L41-L47

So, we can remove this. We can save gzip file for some optimization, but this is not clear how to validate cache then and how to update it. 

I think we can just load it with flag to download or not and activate it on spaces

isaac-chung:
If anything, this reveals a bug in our code - why do we bother checking for the cache file is we don't generate it at all? This means we _always_ download_from_remote(), and we always read all result JSON files and rebuild the full BenchmarkResults object, and this sounds like the opposite of an upgrade to me.

Any idea why we removed the file dump? Maybe we should add it back.

Samoed:
Agree, we can add it back

KennethEnevoldsen:
The reason the cache was removed was that the cache invalidation makes leaderboard development hard (but it has been a while so I am a bit uncertain - could also just be a merge)

Example workflow:
1) Someone reports a bug (model X does not appear)
2) I run the leaderboard locally to confirm (creating a cache file)
3) I fix the bug
4) I rerun the leaderboard, which loads from cache and then hides the bug fix

Seems like we have two changes here that are meaningful: 1) faster loading using a pre-computed file, and 2) faster download by reducing the file size.

To me it seems like this could be incorporated in the current `_load_results` as:

```py
# suggest that we rename to: _load_results_with_remote_cache
def _load_results(cache: ResultCache) -> BenchmarkResults:
    # check if ResultCache has a remote with a valid cached result
      # attempt to download the remote 
      # load and return
    # if not available load results and usual 
```

Problems I see from this are how we ensure that the loaded results are equivalent to what would be generated. An easy fix here would be to add `_load_results` (without cache), which just loads the results. This function is then used in the results repo to generate the cache as well. For debugging, we add the flag `--no-cache`.

I would ideally want these changes added to the `ResultCache` rather than to the leaderboard, as they are not leaderboard-specific and could benefit us elsewhere, but I am not entirely sure where to work them in. Maybe a general concern is that we could speed it up further down the chain without (reducing download size + loading time), and that would be a simpler solution that tackles the root problems (we e.g. have a PR open here: https://github.com/embeddings-benchmark/results/pull/364)

I am unsure how far we can get with these. I might be leaning towards adding the above-mentioned `_load_results_with_remote_cache` for now  - would love to hear your thoughts here as well.

Samoed:
> I would ideally want these changes added to the ResultCache rather than to the leaderboard, as they are not leaderboard-specific and could benefit us elsewher

Agree, that's better place

isaac-chung:
To fully address the cache invalidation problem (separate PR), I agree that a `_load_results` (without cache) function can be added, which just loads the results. In the meantime, I'll apply these changes to `ResultCache`.

isaac-chung:
I've moved the changes to `ResultCache` as `ResultCache.download_cached_results_from_branch()`

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
...
+        """
+        if output_path is None:
+            # Default to saving in mteb/leaderboard/__cached_results.json
+            # Get the mteb package directory (parent of this file)
+            mteb_package_dir = Path(__file__).parent
+            output_path = mteb_package_dir / "leaderboard" / "__cached_results.json"
+
+        # Extract repository owner and name from the remote URL
+        # e.g., "https://github.com/embeddings-benchmark/results" -> "embeddings-benchmark/results"
+        repo_path = remote.replace("https://github.com/", "").replace(
```
Samoed, line 340: Maybe pass repo name instead of full url?

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
 
         return results_directory
 
+    def download_cached_results_from_branch(
```
Samoed: Maybe this should return BenchmarkResult object?

isaac-chung: No why should it? This follows `download_from_remote()` above which returns a Path. This also writes the cache file. Whatever function reads the cache file will return the `BenchmarkResult`

mteb/cache.py
isaac-chung: If a user wants to specify a completely different repo (self hosted), then your suggestion blocks it.

mteb/cache.py
KennethEnevoldsen:
```suggestion
    def _download_cached_results_from_branch(
```
Let us keep it private for now since it is only intended for internal use in the leaderboard

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
...
+        self,
+        branch: str = "cached-data",
+        filename: str = "__cached_results.json.gz",
+        output_path: Path | None = None,
```
KennethEnevoldsen, line 287: shouldn't this be specified within the cache structure which is currently
```
cache_name
|– results # local results
|– remote
  |– results # remote results
  |– ... # other stuff in the remote branch
```

We could add it to `remote`, but then we have to make sure that you could still git clone the remote in the future, and make sure that they don't collide

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
...
+        remote: str = "https://github.com/embeddings-benchmark/results",
+        timeout: int = 60,
+        max_size_mb: int = 500,
```
KennethEnevoldsen, line 290: why is there a max size and timeout?

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
...
+                ct in content_type
+                for ct in [
+                    "application/gzip",
+                    "application/octet-stream",
+                    "application/x-gzip",
+                ]
+            ):
+                logger.warning(
+                    f"Unexpected content-type: {content_type}, continuing anyway..."
+                )
```
KennethEnevoldsen: But why would it continue anyway?

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
 
         return results_directory
 
+    def download_cached_results_from_branch(
+        self,
+        branch: str = "cached-data",
+        filename: str = "__cached_results.json.gz",
+        output_path: Path | None = None,
+        remote: str = "https://github.com/embeddings-benchmark/results",
```
KennethEnevoldsen, line 288: do we want this to work on any other repo (it requires the exact setup that we have)

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
...
+
+        except gzip.BadGzipFile as e:
+            logger.error(f"Invalid gzip file: {e}")
+            raise
+        except UnicodeDecodeError as e:
+            logger.error(f"UTF-8 decode error during decompression: {e}")
+            raise
+        except Exception as e:
+            logger.error(f"Unexpected decompression error: {type(e).__name__}: {e}")
+            raise
```
KennethEnevoldsen: why the try except here? I would just not catch the error

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
...
+            )
+        except PermissionError as e:
+            logger.error(f"Permission denied writing to {output_path}: {e}")
+            raise
+        except OSError as e:
+            logger.error(f"OS error writing to {output_path}: {e}")
+            raise
+        except Exception as e:
+            logger.error(f"Unexpected file write error: {type(e).__name__}: {e}")
+            raise
```
KennethEnevoldsen: same here

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
...
+
+        # Step 3: Write to disk
+        logger.info(f"Attempting to write to: {output_path}")
+
+        # Check parent directory exists and is writable
+        parent_dir = output_path.parent
+        if not parent_dir.exists():
+            logger.info(f"Creating parent directory: {parent_dir}")
+            parent_dir.mkdir(parents=True)
```
KennethEnevoldsen: can be simplified to simply 
```suggestion
        parent_dir = output_path.parent.mkdir(parents=True, exist_ok=True)
```

mteb/cache.py
```diff
@@ -275,6 +279,160 @@ def download_from_remote(
...
+            raise
+        except requests.exceptions.ConnectionError as e:
+            logger.error(f"HTTP connection error: {e}")
+            raise
+        except requests.exceptions.HTTPError as e:
+            logger.error(f"HTTP error {response.status_code}: {e}")
+            raise
+        except Exception as e:
+            logger.error(f"Unexpected HTTP error: {type(e).__name__}: {e}")
+            raise
```
KennethEnevoldsen: if we just raise the error directly after I think we gain a lot from catching the error

mteb/cache.py
isaac-chung: Explainability. I'd like to know where it went wrong, if it went wrong.

KennethEnevoldsen: Totally but you write out the error type - I don't think it adds new information

mteb/cache.py
isaac-chung: could work with maeb-results, for instance. of the human-benchmark-results.

isaac-chung: It might be better if I unify `remote: str = "https://github.com/embeddings-benchmark/results"` from `download_from_remote` ?

mteb/leaderboard/app.py
isaac-chung: Friendly reminder that this no longer exists.

mteb/cache.py
isaac-chung: We could, but I don't think this is the place to discuss this. I'm just trying to clean up and separate out the left-over stuff from the cache file write removal.

mteb/cache.py
isaac-chung: Guardrails.

mteb/cache.py
isaac-chung: It can raise an error instead

mteb/cache.py
isaac-chung: Not aware of catching error in Python. Do you mean you'd like to simply raise the error here?

mteb/cache.py
isaac-chung: IMO we're better off with some limits instead of none. But if you really don't like them, we can remove them.

mteb/cache.py
isaac-chung: Let's just raise the error then?

mteb/cache.py
Samoed: I mean we can pass `embeddings-benchmark/results` without github http prefix

mteb/cache.py
Samoed: Should we add a method to load it directly from cache path then?

mteb/cache.py
isaac-chung: We can, but I don't think it's the easiest to copy paste

isaac-chung: We can, but I don't think it's the easiest to copy paste

mteb/cache.py
isaac-chung: We could - would you prefer to review a bigger PR? or want it separately?

mteb/cache.py
KennethEnevoldsen: Hmm, I don't know what it prevents. (I might be missing something obvious here) If we want to be notified if the file grows too large them shouldn't be simply add it to the results repo tests?

mteb/cache.py
KennethEnevoldsen: but don't we at least want a default here or do we not want to save the file locally?

mteb/cache.py
KennethEnevoldsen: I think that would be better (sorry see that it was stated a bit agressive that wasn't my intention)

mteb/cache.py
KennethEnevoldsen: > could work with maeb-results, for instance, of the human-benchmark-results.

but HUME is merged and MAEB will probably also be added here - it would still require a us to setup the CI for those repos.


> It might be better if I unify

not entirely sure what you mean here

mteb/cache.py
KennethEnevoldsen: I would prefer that

mteb/cache.py
KennethEnevoldsen: Yea same as above - I think it is cleaner to not catch the error and just let it be raised

mteb/cache.py
isaac-chung: The default is set on L331

mteb/cache.py
isaac-chung: No - more like if the file size or the download process is out of expected nominal range. The branch can be corrupted (or hacked) to have a huge file. This way we have an added check on this side of the code as well. 
Re: timeout, here is a [relevant thread](https://github.com/psf/requests/issues/3070#issuecomment-2121302741).

mteb/cache.py
isaac-chung: Yep no worries

mteb/cache.py
isaac-chung: Yep no worries

mteb/cache.py
isaac-chung: I'm referring to the `download_from_remote()` method also referencing the same remote URL. If we don't care to offer the option to read from a different results repo (think of users who hosts this in their own org or fork), then we can set a single `remote_url` param and share between the 2 methods.

mteb/cache.py
Samoed: I think we can put in our default cache folder under remote folder, because it already has this structure. Our at least we can use `~/.cache/mteb` as a default

isaac-chung: I think we should aim to match where it used to be (in the leaderboard folder), and make other changes in a separate PR if we agree on it.

mteb/cache.py
isaac-chung: Leaving this as is for now.

mteb/cache.py
isaac-chung: We could, in a separate PR, update `ResultCache.load_results()` or add a separate `ResultCache.load_results_from_cache_file()`. Similar to https://github.com/embeddings-benchmark/mteb/issues/3798

isaac-chung:
Let's avoid scope creep in this PR and check if previous comments have been resolved? There are a few good ideas and they are recorded in https://github.com/embeddings-benchmark/mteb/issues/3798 for next steps. Remember that [this PR as-is already speeds up leaderboard build ~7x](https://github.com/embeddings-benchmark/mteb/pull/3795#issuecomment-3691676698).

tests/test_leaderboard/conftest.py
```diff
@@ -0,0 +1,14 @@
+"""Shared test fixtures and configuration for leaderboard tests."""
+
+import pytest
+
+
+@pytest.fixture(scope="session")
+def leaderboard_test_config():
```
Samoed: Seems that this conftest is not used

tests/conftest.py
Samoed, line 1: Maybe move to `test_leaderboard/conftest.py`

tests/test_leaderboard/conftest.py
isaac-chung: Will remove.

tests/conftest.py
isaac-chung: They're shared between test_leaderboard and test_result_cache. Should stay put.

isaac-chung:
@Samoed @KennethEnevoldsen do you have more feedback? or is this good to go?

mteb/cache.py
```diff
@@ -275,6 +279,136 @@ def download_from_remote(
...
+                )
+
+            logger.info(
+                f"HTTP request successful, content length: {content_length} bytes"
+            )
+            content = response.content
+
+        except Exception as e:
+            logger.error(f"Unexpected HTTP error: {type(e).__name__}: {e}")
+            raise e
```
Samoed, line 382: I think we can remove this try catch, because this will just raise error

mteb/cache.py
```diff
@@ -275,6 +279,136 @@ def download_from_remote(
...
+        logger.info("Attempting gzip decompression...")
+
+        try:
+            with gzip.open(io.BytesIO(content), "rt", encoding="utf-8") as gz_file:
+                data = gz_file.read()
+            logger.info(f"Decompression successful, data length: {len(data)} chars")
+
+        except Exception as e:
+            logger.error(f"Unexpected decompression error: {type(e).__name__}: {e}")
+            raise e
```
Samoed, line 394: This can be removed too

mteb/cache.py
```diff
@@ -275,6 +279,136 @@ def download_from_remote(
...
+
+        # Check parent directory exists and is writable
+        output_path.parent.mkdir(parents=True, exist_ok=True)
+
+        try:
+            output_path.write_text(data, encoding="utf-8")
+            logger.info(
+                f"File write successful, size: {output_path.stat().st_size} bytes"
+            )
+        except Exception as e:
```
Samoed, line 407: And this

mteb/leaderboard/app.py
```diff
@@ -32,37 +32,127 @@
...
+            raise
+        except Exception as e:
+            logger.error(f"Unexpected error in from_validated: {type(e).__name__}: {e}")
+            raise
+
+    except Exception as e:
+        logger.error(
+            f"Failed to load cached results from disk: {type(e).__name__}: {e}"
+        )
+        raise
```
Samoed: Here I'm not sure too if we need try execpt, because we will just adding error message to log

tests/test_leaderboard/test_leaderboard.py
```diff
@@ -0,0 +1,209 @@
...
+        level=logging.INFO,
+        format="%(asctime)s - TEST_PID:%(process)d - %(name)s - %(levelname)s - %(message)s",
+        force=True,
+    )
+    logger = logging.getLogger(__name__)
+
+    # Suppress WARNING messages in the main test process too
+    logging.getLogger("mteb.results.task_result").setLevel(logging.ERROR)
+    logging.getLogger("mteb.models.model_meta").setLevel(logging.ERROR)
+    logging.getLogger("mteb.results.benchmark_results").setLevel(logging.ERROR)
```
Samoed, line 76: I think this is duplicating logging setup from `run_leaderboard_app`

isaac-chung: I wanted to tests to also show info logs, and needed these to achieve that.

isaac-chung:
@Samoed I feel the last review neglected how the result cache and app.py `_load_results()` work together. Specifically, if there are any errors raised from calling `cache._download_cached_results_from_branch()` in app.py, it falls back to downloading full remote repository. 

The only change I'll make is the reduction of excepts in app.py.

mteb/leaderboard/app.py
isaac-chung: In future, we will need to handle the case somehow - e.g. gracefully handle these exceptions, and not just error out. That's just bad engineering.

Samoed:
Let's do that

isaac-chung:
Thanks. Done.

KennethEnevoldsen:
Sorry @isaac-chung for this PR taking so long. I agree that we have had a lot of scope creep in this PR and it partly stems from us not having had an initial discussion on the implementation, but I think we through this seems to settled on three parts which I would split into separate PRs: 1) for the ResultCache changes, 2) for the leaderboards result loading, and 3) for any changes to the leaderboard tests

isaac-chung:
@KennethEnevoldsen no worries. I'm glad we got the ball rolling with this PR, and I felt it was more concrete than if we had a discussion without. 

Any changes without tests shouldn't be merged anyway, so it'd make sense to combine the tests the with changes. Seeing that we'll exceed 100 comments shortly, instead of splitting this PR, perhaps at this time we can assess whether all your _existing_ comments have been addressed, and introduce follow-up PRs to address any outstanding items?

PR #3796, opened 2025-12-25
update generate_model_card with get_benchmark_result()

### LINKED ISSUE #3776: Add to `generate_model_card` results on benchmarks

Currently tasks scores only added to task metadata, but we can add results on benchmarks with model comparison to readme directly

ayush1298:
closes #3776

### DIFF
--- a/tests/test_results/test_benchmark_results.py
+++ b/tests/test_results/test_benchmark_results.py
@@ -5,7 +5,6 @@ from pathlib import Path
 
 import pandas as pd
 import pytest
-import yaml
 
 import mteb
 from mteb import Benchmark
@@ -212,7 +211,6 @@ def test_generate_model_card_with_table_and_benchmarks(
         ),
     ]
 
-    # Generate model card with 2 models to compare
     generate_model_card(
         model_name="sentence-transformers/all-MiniLM-L6-v2",
         benchmarks=benchmarks,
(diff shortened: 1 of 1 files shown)


mteb/cli/generate_model_card.py
```diff
@@ -5,14 +5,15 @@
...
+from mteb.benchmarks.benchmark import Benchmark
 from mteb.cache import ResultCache
 
 logger = logging.getLogger(__name__)
 
 
 def generate_model_card(
     model_name: str,
-    tasks: list[AbsTask] | None = None,
+    tasks: list[AbsTask] | Benchmark | list[Benchmark] | None = None,
```
ayush1298: @Samoed I am not sure when this is useful exactly because only place where this is used is:

https://github.com/embeddings-benchmark/mteb/blob/main/mteb/cli/build_cli.py#L288-L302, but here we are adding both individual tasks and tasks from benchmark, so we cant change it to pass benchmark directly

Samoed: You don't handle in `_add_table_to_model_card` `list[Benchmark]`. Also, no benchmark name won't be added if passed here. CLI function also should be updated

Samoed: Yes, you need to udate this too

mteb/cli/build_cli.py
```diff
@@ -285,17 +285,15 @@ def _create_meta(args: argparse.Namespace) -> None:
...
-    tasks = []
+    tasks_or_benchmarks = None
     if tasks_names is not None:
-        tasks = mteb.get_tasks(tasks_names)
+        tasks_or_benchmarks = mteb.get_tasks(tasks_names)
     if benchmarks is not None:
-        benchmarks = mteb.get_benchmarks(benchmarks)
-        for benchmark in benchmarks:
-            tasks.extend(benchmark.tasks)
+        tasks_or_benchmarks = mteb.get_benchmarks(benchmarks)
```
Samoed: I think it's better to split those parameters and pass them separately to `generate_model_card`

ayush1298: The problem is, load_results() from cache https://github.com/embeddings-benchmark/mteb/blob/main/mteb/cache.py#L467-L475, doesnt take list[Benchmark], so I have to anyhow convert them, so isn't its better to do that conversion at 1 place only.

Samoed: > doesnt take list[Benchmark], so I have to anyhow convert them

You shouldn't convert them. You can simply iterate over them

ayush1298: So, is it correct that we should get tasks to pass to `load_results()` from both parameters: `tasks` as well as `benchmark`.
While for `_add_table_to_model_card`, we should use benchmark and then `get_benchmark_results() `directly?

Samoed: Yes, you can do like this

ayush1298: Can you check it now?

Samoed:
I've updated a bit to make it work correctly. Can you add a test for this?

mteb/cache.py
```diff
@@ -472,7 +472,7 @@ def _filter_paths_by_task(
     def load_results(
         self,
         models: Sequence[str] | Iterable[ModelMeta] | None = None,
-        tasks: Sequence[str] | Iterable[AbsTask] | str | None = None,
+        tasks: Sequence[str] | Iterable[AbsTask] | Benchmark | str | None = None,
```
Samoed, line 475: Added benchmark just for documentation purposes that this is possible and because later we're checking if tasks are benchmark or not

ayush1298:
> I've updated a bit to make it work correctly. Can you add a test for this?

Both current tests in https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_cli.py,  passed None for both tasks and benchmarks. So, should I change them with some task and benchmark?

Samoed:
You can create new test using benchmark results like in https://github.com/embeddings-benchmark/mteb/blob/42dea0150cb270f0d9485c6d4b40ea678391259d/tests/test_results/test_benchmark_results.py#L166

ayush1298:
Added tests

Samoed:
I think it's better to add "golden" file and check if test output matching or not like in https://github.com/embeddings-benchmark/mteb/blob/42dea0150cb270f0d9485c6d4b40ea678391259d/tests/test_cli.py#L145

tests/test_results/test_benchmark_results.py
```diff
@@ -187,3 +188,101 @@ def test_benchmark_results(cache_path: Path) -> None:
...
+    assert temp_output.exists(), "Model card file not created"
+    content = temp_output.read_text()
+
+    assert "# MTEB Results" in content or "# MTEB results" in content, (
+        "Results table heading not found"
+    )
+    assert "STS12" in content or "Banking77" in content, "Task names not in content"
+
+
+def test_generate_model_card_with_table_and_benchmark(
```
Samoed: I think you can use this test, but with 2 models and remove other tests, because we already have tests for this functionality

tests/test_results/test_benchmark_results.py
```diff
@@ -187,3 +189,122 @@ def test_benchmark_results(cache_path: Path) -> None:
...
+    golden_readme = golden_content[end_yaml_gold + len(yaml_end_sep) :]
+
+    # Validate frontmatter structure
+    assert output_frontmatter is not None, "Output frontmatter should not be empty"
+    assert golden_frontmatter is not None, "Golden frontmatter should not be empty"
+
+    # Validate key frontmatter fields exist
+    assert "model-index" in output_frontmatter, "model-index not found in output"
+    assert "tags" in output_frontmatter, "tags not found in output"
+    assert "mteb" in output_frontmatter.get("tags", []), "mteb tag not found"
```
Samoed: You don't need to check this
```suggestion
```

tests/test_results/test_benchmark_results.py
```diff
@@ -187,3 +189,122 @@ def test_benchmark_results(cache_path: Path) -> None:
...
+        f"Number of models in model-index should match: "
+        f"expected {len(golden_model_index)}, got {len(output_model_index)}"
+    )
+
+    # Verify model names match
+    output_models = [m["name"] for m in output_model_index]
+    golden_models = [m["name"] for m in golden_model_index]
+    assert output_models == golden_models, (
+        f"Model names should match: expected {golden_models}, got {output_models}"
+    )
```
Samoed:
```suggestion
```

ayush1298: what I have to check in test, then?

Samoed: These checks are for frontmatter only. You should check that tables are the same

ayush1298:
@Samoed Can you check this one, I have updated tests

Issue #3784, opened 2025-12-22
[`TC`](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) - flake8-type-checking


Samoed:
Also, can be activated with

Ref https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/#options
```toml
[tool.ruff.lint]
future-annotations = true  # For TC rules
typing-modules = ["mteb.types", "collections.abc"]

[tool.ruff.lint.flake8-type-checking]
strict = true
runtime-evaluated-base-classes = ["pydantic.BaseModel"]
```

Should be done after https://github.com/embeddings-benchmark/mteb/pull/3550

Samoed:
I found strange behavior of ruff https://github.com/astral-sh/ruff/issues/22223, so I think we need to wait to activate this

PR #3800, opened 2025-12-27
refactor: Activate `TC`

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3784

### DIFF
--- a/mteb/models/vllm_wrapper.py
+++ b/mteb/models/vllm_wrapper.py
@@ -23,8 +23,6 @@ if TYPE_CHECKING:
     from mteb.abstasks.task_metadata import TaskMetadata
     from mteb.types import Array, BatchedInput
 
-else:
-    PoolerConfig = dict[str, Any]
 
 logger = logging.getLogger(__name__)
 


docs/overview/create_available_benchmarks.py
```diff
@@ -41,7 +43,7 @@ def pretty_long_list(items: list[str], max_items: int = 5) -> str:
 def create_table(benchmark: mteb.Benchmark) -> str:
     """Create a markdown table of tasks in the benchmark."""
     tasks = benchmark.tasks
-    tasks = cast(MTEBTasks, tasks)
+    tasks = cast("MTEBTasks", tasks)
```
KennethEnevoldsen, line 43: Why the quotes? Seems like that would be worse

Samoed: This was done due to https://docs.astral.sh/ruff/rules/runtime-cast-value/#flake8-type-checking-tc

Overall `cast` doing nothing and typecheckers can deal with type in string with cast

Samoed:
Thanks looks interesting, but for now I don't think we have problems with them

docs/overview/create_available_benchmarks.py
KennethEnevoldsen: But it also completely breaks navigation in the code. I am not sure I agree with this one. I would suggest disabling it

KennethEnevoldsen:
> but for now I don't think we have problems with them

I think we do, but we just fix it by moving it into the function, but it is still a circular import (I agree that it is not a big problem atm.)

docs/overview/create_available_benchmarks.py
Samoed: We still have these types imported under `TYPE_CHECKING`, otherwise, we would need to include them at the top level. I’m not totally against disabling this rule, but I think it helps simplify our imports. 

I tested this in VS Code, PyCharm, Zed can resolve these imports correctly and navigate to the definitions without any issues.

Zed
image
Vscode
image

KennethEnevoldsen: Alright let us keep it as is

PR #3801, opened 2025-12-27
fix: Add `fill_missing` parameter in `get_model_meta`

### LINKED ISSUE #3693: Add `compute_missing` to `mteb.get_model_meta`

```
meta = mteb.get_model_meta(
  "name",
  fetch_from_hf=True,
  compute_missing=True, # default should be false
)
"""
compute_missing: Computes missing attributes from the metadata including number of parameters and memory usage. 
```

ayush1298:
closes #3693

### DIFF
--- a/mteb/models/get_model_meta.py
+++ b/mteb/models/get_model_meta.py
@@ -108,7 +108,7 @@ def get_model_meta(
     model_name: str,
     revision: str | None = None,
     fetch_from_hf: bool = True,
-    compute_missing: bool = False,
+    fill_missing: bool = False,
 ) -> ModelMeta:
     """A function to fetch a model metadata object by name.
 
@@ -116,7 +116,7 @@ def get_model_meta(
         model_name: Name of the model to fetch
         revision: Revision of the model to fetch
         fetch_from_hf: Whether to fetch the model from HuggingFace Hub if not found in the registry
-        compute_missing: Computes missing attributes from the metadata including number of parameters and memory usage.
+        fill_missing: Computes missing attributes from the metadata including number of parameters and memory usage.
 
     Returns:
         A model metadata object
@@ -129,7 +129,7 @@ def get_model_meta(
                 f"Model revision {revision} not found for model {model_name}. Expected {model_meta.revision}."
             )
 
-        if compute_missing and fetch_from_hf:
+        if fill_missing and fetch_from_hf:
             original_meta_dict = model_meta.model_dump()
             new_meta = ModelMeta.from_hub(model_name)
             new_meta_dict = new_meta.model_dump(exclude_none=True)
--- a/tests/test_models/test_model_meta.py
+++ b/tests/test_models/test_model_meta.py
@@ -143,7 +143,7 @@ def test_loader_kwargs_persisted_in_metadata():
 def test_compute_missing_parameter():
     """Test that compute_missing parameter fetches missing metadata from HuggingFace Hub"""
     model_name = "sentence-transformers/all-MiniLM-L6-v2"
-    meta_with_compute = mteb.get_model_meta(model_name, compute_missing=True)
+    meta_with_compute = mteb.get_model_meta(model_name, fill_missing=True)
 
     assert meta_with_compute.n_parameters is not None
     assert meta_with_compute.memory_usage_mb is not None


mteb/models/get_model_meta.py
```diff
@@ -124,7 +128,20 @@ def get_model_meta(
...
+                    f"Number of parameters is missing for model '{model_name}'. Computing them now using Huggingface Hub."
+                )
+                model_meta.n_parameters = model_meta.calculate_num_parameters_from_hub()
+            if model_meta.memory_usage_mb is None:
+                logger.info(
+                    f"Memory usage is missing for model '{model_name}'. Computing them now using Huggingface Hub."
+                )
+                model_meta.memory_usage_mb = model_meta.calculate_memory_usage_mb()
         return model_meta
+
```
ayush1298, line 146: @Samoed Just a doubt,  `if compute_missing `, should under `if model_name in MODEL_REGISTRY:`
or its should be out of this if statement just like` if fetch_from_hf:`

Samoed: I think you can add `if compute_missing and fetch_from_hf`, but then this seems a bit strange

ayush1298: yes, then there is no use of `compute_missing` as a separate parameter as anyway `fetch_from_hf` will call it.

mteb/models/get_model_meta.py
```diff
@@ -124,7 +128,20 @@ def get_model_meta(
...
+            if model_meta.n_parameters is None:
+                logger.info(
+                    f"Number of parameters is missing for model '{model_name}'. Computing them now using Huggingface Hub."
+                )
+                model_meta.n_parameters = model_meta.calculate_num_parameters_from_hub()
+            if model_meta.memory_usage_mb is None:
+                logger.info(
+                    f"Memory usage is missing for model '{model_name}'. Computing them now using Huggingface Hub."
+                )
+                model_meta.memory_usage_mb = model_meta.calculate_memory_usage_mb()
```
Samoed: I think you can do like 

```python
new_meta = ModelMeta.from_hf(model_name)
return ModelMeta(**model_meta.model_dump(), **new_meta.model_dump(exclude_none=True))
```

ayush1298: Its same only, right?

Samoed: Not sure what you mean

ayush1298: So, I am saying we are 1st calling .from_hf() method, and then finally, in ModelMeta taking all things from these new_meta which are not None, but I think we have to calculate only the number of parameters and memory usage?

Samoed: Metas can miss any parameters. I think it's easier to do like this

ayush1298: okay. will change

mteb/models/get_model_meta.py
```diff
@@ -124,10 +128,15 @@ def get_model_meta(
             raise ValueError(
                 f"Model revision {revision} not found for model {model_name}. Expected {model_meta.revision}."
             )
-        return model_meta
+
+        new_meta = ModelMeta.from_hf(model_name)
+        return ModelMeta(
+            **model_meta.model_dump(), **new_meta.model_dump(exclude_none=True)
+        )
```
```suggestion
    if compute_missing and fetch_from_hf:
        new_meta = ModelMeta.from_hf(model_name)
        return ModelMeta(
            **model_meta.model_dump(), **new_meta.model_dump(exclude_none=True)
        )
    return model_meta
```

ayush1298:
@Samoed How the typechecking tests suddenly failing in all PRs?
getting this error: 

`mteb/models/get_model_meta.py:133: error: "type[ModelMeta]" has no attribute "from_hf"  [attr-defined]`

Samoed:
It was added to main. It's not failing in all prs, but if there is an error it will fail. Currently, the reason is that you call method that not existing. I fixed it

ayush1298:
@Samoed Do we have to add tests for this, or is it good?

Samoed:
Yes, it's better to add a test

mteb/models/get_model_meta.py
```diff
@@ -124,10 +128,24 @@ def get_model_meta(
...
+            new_meta_dict = new_meta.model_dump(exclude_none=True)
+
+            updates = {
+                k: v
+                for k, v in new_meta_dict.items()
+                if model_meta.model_dump().get(k) is None
+            }
+
+            if updates:
+                return model_meta.model_copy(update=updates)
```
ayush1298, line 144: Changed to this, because previously while merging original and new_meta there are conflicts because same filed are present in both original and new_meta.  At that time, getting error like: 

`TypeError: mteb.models.model_meta.ModelMeta() got multiple values for keyword argument 'name'`

mteb/models/get_model_meta.py
```diff
@@ -124,10 +128,24 @@ def get_model_meta(
...
             )
+
+        if compute_missing and fetch_from_hf:
+            new_meta = ModelMeta.from_hub(model_name)
+            new_meta_dict = new_meta.model_dump(exclude_none=True)
+
+            updates = {
+                k: v
+                for k, v in new_meta_dict.items()
+                if model_meta.model_dump().get(k) is None
```
Samoed: Can you move `model_meta.model_dump()` to separate variable?

mteb/models/get_model_meta.py
Samoed: Agree, that's better

tests/test_models/test_model_meta.py
```diff
@@ -140,6 +140,15 @@ def test_loader_kwargs_persisted_in_metadata():
     assert meta.loader_kwargs["not_existing_param"] == 123
 
 
+def test_compute_missing_parameter():
+    """Test that compute_missing parameter fetches missing metadata from HuggingFace Hub"""
+    model_name = "sentence-transformers/all-MiniLM-L6-v2"
+    meta_with_compute = mteb.get_model_meta(model_name, compute_missing=True)
+
+    assert meta_with_compute.n_parameters is not None
```
Samoed, line 148: Can you add check that existing meta is none?

ayush1298: I tried changing test to: 

```
def test_compute_missing_parameter():
    """Test that compute_missing parameter fetches missing metadata from HuggingFace Hub"""
    model_name = "nomic-ai/nomic-embed-text-v1"
    meta_without_compute = mteb.get_model_meta(model_name, compute_missing=False)
    meta_with_compute = mteb.get_model_meta(model_name, compute_missing=True)

    assert meta_without_compute.n_parameters is None
    assert meta_without_compute.memory_usage_mb is None

    assert meta_with_compute.n_parameters is not None
    assert meta_with_compute.memory_usage_mb is not None
```
    
There is no model with both `n_parameters` and `memory_usage_mb` None at same time and can be retrieved by using `from_hub()` method. There are some models with both none, but for them safetensors are not present on hf, so `from_hub()`, will not be able to retrieve anything in that case

ayush1298: I can try to add 2 different models to test, 1 for `n_parameter` and another for `memory_usage_mb`

Samoed: Agree, let's keep as is, for now then

mteb/models/get_model_meta.py
ayush1298: made these change

Samoed (approved):

ayush1298:
@KennethEnevoldsen Just tagging you to review this one whenever you get time

mteb/models/get_model_meta.py
```diff
@@ -124,10 +128,25 @@ def get_model_meta(
...
+
+            updates = {
+                k: v
+                for k, v in new_meta_dict.items()
+                if original_meta_dict.get(k) is None
+            }
+
+            if updates:
+                return model_meta.model_copy(update=updates)
         return model_meta
```
KennethEnevoldsen: isn't this the same as line 151:

```
meta = ModelMeta.from_hub(model_name, revision)
```
Does it compute anything else?

KennethEnevoldsen:
I think the initial intention with the issue was that we wanted to load the model to compute parameters if needed.

mteb/models/get_model_meta.py
Samoed: If `compute_missing` is activated then None values would be changed to new from hub

ayush1298:
> I think the initial intention with the issue was that we wanted to load the model to compute parameters if needed.

We are doing that only, but for that, both `compute_missing` and `fetch_from_hf` should be true. Also, its different from line 151, because after fetching from hf, we want to replace only those things which are None before, not to replace all the things.

mteb/models/get_model_meta.py
KennethEnevoldsen: Right so fetch_hf is more "you are allowed to fetch from the web if the models does not exist", whereas compute missing is more "you should to fill in missing values if you can".

Should we could it `compute_missing_from_hf` as to make it clear that we do not load the model.

KennethEnevoldsen:
but we are not loading the models using sentence transformers if that is required for some metrics

mteb/models/get_model_meta.py
Samoed: `from_hub` doesn’t download the model, so I think that was already clear, wasn’t it?

ayush1298:
> but we are not loading the models using sentence transformers if that is required for some metrics

I think we don't need it to load models. For `embedding_parameter` calculation only need it I guess.

mteb/models/get_model_meta.py
KennethEnevoldsen: hmm maybe `compute` seems to suggest something else than `fetch_unfilled_metadata`.

KennethEnevoldsen:
Yep. I think it is perfectly fine that we don't compute n embedding, but I would probably rename it (see discussion above)

mteb/models/get_model_meta.py
Samoed: We can rename to `fill_missing`

KennethEnevoldsen: let us do that

ayush1298:
> Yep. I think it is perfectly fine that we don't compute n embedding, but I would probably rename it (see discussion above)

Okay, rename it to `fill_missing`

KennethEnevoldsen (approved):

Issue #3725, opened 2025-12-12
Add model: `sentence-transformers/static-retrieval-mrl-en-v1`, `sentence-transformers/multi-qa-mpnet-base-dot-v1`


Samoed:
### Model link on Hugging Face

https://huggingface.co/sentence-transformers/static-retrieval-mrl-en-v1

Issue #3724, opened 2025-12-12
Add model: jinaai/jina-clip-v2


Samoed:
### Model link on Hugging Face

https://huggingface.co/jinaai/jina-clip-v2

### Description of the model

Ref https://github.com/embeddings-benchmark/results/pull/362#issuecomment-3646532702

PR #3808, opened 2025-12-29
model: add missing sentence transformers and jina models

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3725
Close https://github.com/embeddings-benchmark/mteb/issues/3724

I've run both models and got same results as in results repo


ayush1298:
@Samoed Could you merge these? Then we can also merge where I have excluded these models: https://github.com/embeddings-benchmark/results/pull/383

PR #3809, opened 2025-12-29
docs: Fix docs build strict mode errors

### LINKED ISSUE #3501: CI: Docs should fails if there is a broken link

This should be doable by running it in `--strict` mode

docs: https://www.mkdocs.org/user-guide/configuration/#strict

isaac-chung:
Fixes #3501

### DIFF
--- a/mteb/abstasks/multilabel_classification.py
+++ b/mteb/abstasks/multilabel_classification.py
@@ -70,10 +70,10 @@ class AbsTaskMultilabelClassification(AbsTaskClassification):
         input_column_name: Name of the column containing the input text.
         label_column_name: Name of the column containing the labels.
         samples_per_label: Number of samples to use pr. label. These samples are embedded and a classifier is fit using the labels and samples.
-        multilabel_classifier: Classifier to use for evaluation. Must be sklearn-compatible.
+        evaluator_model: Classifier to use for evaluation. Must implement the SklearnModelProtocol.
     """
 
-    multilabel_classifier: SklearnModelProtocol = KNeighborsClassifier(n_neighbors=5)
+    evaluator_model: SklearnModelProtocol = KNeighborsClassifier(n_neighbors=5)
     input_column_name: str = "text"
     label_column_name: str = "label"
 
@@ -169,7 +169,7 @@ class AbsTaskMultilabelClassification(AbsTaskClassification):
             y_train = train_split.select(sample_indices)[self.label_column_name]
             y_train = binarizer.transform(y_train)
             y_pred, current_classifier = _evaluate_classifier(
-                X_train, y_train, X_test, self.multilabel_classifier
+                X_train, y_train, X_test, self.evaluator_model
             )
             if prediction_folder:
                 all_predictions.append(y_pred.tolist())


docs/whats_new.md
```diff
@@ -79,7 +79,7 @@ batch_input: BatchedInput = {
 
 Where `text` is a batch of texts and `list[images]` is a batch for that texts. This e.g. allows markdown documents with multiple figures like so:
 
-> As you see in the following figure [figure 1](image_1) there is a correlation between A and B.
+> As you see in the following figure figure 1 there is a correlation between A and B.
```
isaac-chung: Note: This is meant to be an example snippet, not an actual reference to an "image_1"

docs/usage/defining_the_model.md
```diff
@@ -41,7 +41,7 @@ results = mteb.evaluate(model, tasks=tasks)
 However, we do recommend checking if mteb includes an implementation of the model before using sentence transformers since some models (e.g. the [multilingual e5 models](https://huggingface.co/collections/intfloat/multilingual-e5-text-embeddings-67b2b8bb9bff40dec9fb3534)) require a prompt and not specifying it may reduce performance.
 
 !!! note
-    If you want to evaluate a cross encoder for reranking, see the section on [running cross encoders for reranking](./running_the_evaluation.md#running-cross-encoders-on-reranking).
+    If you want to evaluate a cross encoder for reranking, see the section on [running cross encoders for reranking](./running_the_evaluation.md).
```
Samoed:
```suggestion
    If you want to evaluate a cross encoder for reranking, see the section on [running cross encoders for reranking](../andvanced_usage/two_stage_reranking.md).
```

scripts/build_docs_strict_filtered.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python3
```
KennethEnevoldsen:
```suggestion
```

scripts/build_docs_filtered.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python
```
isaac-chung: Seems unused

docs/whats_new.md
```diff
@@ -152,7 +152,7 @@ class SearchProtocol(Protocol):
         ...
 ```
 
-We're automatically wrapping `Encoder` and `CrossEncoder` models support `SearchProtocol`. However, if your model needs a custom index you can implement this protocol directly, like was done for colbert-like models. <!-- [colbert-like models][mteb.models.model_implementations.pylate_models.PylateSearchEncoder]. -->
```
isaac-chung, line 155: Check where this could be

scripts/build_docs_strict_filtered.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
isaac-chung: Can also be removed

mteb/abstasks/regression.py
```diff
@@ -84,7 +84,7 @@ class AbsTaskRegression(AbsTaskClassification):
         n_samples: Number of samples to use for training the regression model. If the dataset has fewer samples than n_samples, all samples are used.
         abstask_prompt: Prompt to use for the task for instruction model if not prompt is provided in TaskMetadata.prompt.
         evaluator_model: The model to use for evaluation. Can be any sklearn compatible model. Default is `LinearRegression`.
-            Full details of api in [`SklearnModelProtocol`][mteb._evaluators.sklearn_evaluator.SklearnModelProtocol].
```
isaac-chung, line 87: Does this exist somewhere else?

Samoed: Link is valid, but we dont export this module https://github.com/embeddings-benchmark/mteb/blob/1c873b458b70d75e2540759e665ba540d163a304/mteb/_evaluators/sklearn_evaluator.py#L19

docs/whats_new.md
Samoed: Here https://github.com/embeddings-benchmark/mteb/blob/1c873b458b70d75e2540759e665ba540d163a304/mteb/models/model_implementations/pylate_models.py#L31

But we don't export for docs

isaac-chung: Thanks! Maybe we need to export it?

mteb/abstasks/regression.py
isaac-chung: Thanks! Maybe we need to export it?

Samoed: Yes, but don't have ideas where it is suitable

docs/whats_new.md
isaac-chung: Actually this looks like a HTML comment, so it won't be rendered anyway.

Samoed: Yes. I've made this as comment to remove for some time, but I think we can just remove this

mteb/abstasks/regression.py
isaac-chung: Figured something out.

isaac-chung:
@Samoed could you give this another look please?

docs/api/evaluators.md
Samoed, line 1: Not sure if we want to add this page with only protocol for sklearn model

mteb/__init__.py
```diff
@@ -49,3 +50,12 @@
     "load_results",
     "types",
 ]
+
+
+# Deferred import to avoid circular dependency
```
Samoed: I don't think we want to export this at top level

isaac-chung:
Can do. Is that all?

Samoed:
Yes

Makefile
```diff
@@ -37,7 +37,8 @@ pr:
 
 build-docs: build-docs-overview
 	@echo "--- 📚 Building documentation ---"
-	uv run --group docs python -m mkdocs build
+	@echo "--- 🔍 Building with strict mode (BibTeX warnings filtered) ---"
+	uv run --group docs python scripts/build_docs_strict_filtered.py
```
Samoed: Can we keep using mkdocs directly here?

isaac-chung: 1. Using mkdocs directly with strict mode fails with warnings. This script ignore those warnings. 
2. I would really appreciate [complete reviews](https://github.com/embeddings-benchmark/mteb/pull/3809#issuecomment-3706130648) as much as possible moving forward.

Samoed: 1. Overall this is fine, I just think we should keep an issue open then
2. I know, just missed this.

isaac-chung: This is the [crux of the problem](https://github.com/shyamd/mkdocs-bibtex?tab=readme-ov-file#debugging):
> You can run mkdocs with the --strict flag to fail building on any citations that don't exist in the bibtex file.

We only put MTEB-related citations in `docs/references.bib`, so mkdocs consider all task citations to be missing.

Here is what I tried before (all failed):

1. Hook-based filtering - Attempted to disable bibtex plugin for specific pages using mkdocs hooks, but this doesn't work with mkdocstrings which extracts docstrings directly from Python
  files
  2. Bibtex plugin configuration - Searched for plugin options to ignore missing references, but the bibtex plugin has no configuration option to suppress warnings for undefined citations
  3. Pre-processing docstrings - Tried to modify docstrings before mkdocstrings processes them, but mkdocstrings extracts directly from Python source files, bypassing any pre-processing
  attempts
  4. Mkdocstrings configuration - Looked for options to filter or ignore citations within mkdocstrings config, but no such option exists to selectively process or ignore BibTeX references
  5. Disabling bibtex for specific paths - Attempted to conditionally disable the bibtex plugin for API documentation paths, but mkdocs doesn't support conditional plugin loading based on file
  paths

isaac-chung:
Opened the above issue for tracking.

Makefile
Samoed: I've digged a bit deeper, and it seems that bibtex plugin tried to process all text staring with `@` as citations, instead of real link (text inside `[]`), so we can't do anything with this (especially since project in archive). We can just disable this puling and find a way what to do with existing citations (probably we can just make a link to list of benchmarks instead). Previously, I thought the main reason why we can't use strict mode is that we have some broken links 
```
Doc file 'overview/available_benchmarks.md' contains a link './available_tasks/retrieval.md#englishfinance2retrieval', but the doc
'overview/available_tasks/retrieval.md' does not contain an anchor '#englishfinance2retrieval'.
```
but it seems that this is not a case

I apologize for the inconvenience caused by the review. I think we can merge this like this then

isaac-chung: Oh those are INFO, not warnings. I might have a solution to use mkdocs directly after all. Let me try it out.

isaac-chung: Turns out there is a setting we can use after all. Now it uses mkdocs build directly.

Samoed: Oh, great!

mteb/abstasks/multilabel_classification.py
```diff
@@ -70,10 +70,10 @@ class AbsTaskMultilabelClassification(AbsTaskClassification):
         input_column_name: Name of the column containing the input text.
         label_column_name: Name of the column containing the labels.
         samples_per_label: Number of samples to use pr. label. These samples are embedded and a classifier is fit using the labels and samples.
-        evaluator: Classifier to use for evaluation. Must implement the SklearnModelProtocol.
+        multilabel_classifier: Classifier to use for evaluation. Must be sklearn-compatible.
     """
 
-    evaluator: SklearnModelProtocol = KNeighborsClassifier(n_neighbors=5)  # type: ignore[assignment]
+    multilabel_classifier: SklearnModelProtocol = KNeighborsClassifier(n_neighbors=5)
```
Samoed: Can we keep typehint and keep name? I think this change could break compatibility

isaac-chung: Yeah - let me fix that.

isaac-chung: Seems like there was a bug - the parent class AbsTaskClassification defines both:
  - evaluator: type[SklearnEvaluator] (a class)
  - evaluator_model: SklearnModelProtocol (an instance)

  The multilabel classification was trying to override evaluator with a model instance, causing the mypy type error. (the parent class defined `evaluator` as `type[SklearnEvaluator]` (a class), but the multilabel classification was trying to assign `KNeighborsClassifier(n_neighbors=5)` (an instance of `SklearnModelProtocol`) to it.)

Samoed: This would be a bit breaking change, but I think this is probably fine

isaac-chung: You agree that it was a bug though and should be fixed?

Samoed: Yes, attribute was incorrect

isaac-chung: ✅?

PR #3812, opened 2025-12-30
fix: Allow passing device to model

### LINKED ISSUE #3773: Allow passing device to model

For not this is not possible to select device when getting a model

ayush1298:
Closes #3773 

I have updated EncoderProtocol, CrossEncoderProtocol and all wrapper classes using it.

### DIFF
--- a/mteb/models/model_implementations/cde_models.py
+++ b/mteb/models/model_implementations/cde_models.py
@@ -50,11 +50,16 @@ class CDEWrapper(SentenceTransformerEncoderWrapper):
     )
 
     def __init__(
-        self, model: str, device: str | None = None, *args, **kwargs: Any
+        self,
+        model: str,
+        revision: str | None = None,
+        device: str | None = None,
+        *args,
+        **kwargs: Any,
     ) -> None:
         from transformers import AutoConfig
 
-        super().__init__(model, device=device, *args, **kwargs)
+        super().__init__(model, revision=revision, device=device, *args, **kwargs)
         model_config = AutoConfig.from_pretrained(model, trust_remote_code=True)
         self.max_sentences = model_config.transductive_corpus_size
 


mteb/models/abs_encoder.py
```diff
@@ -52,7 +52,7 @@ def get_prompt_name(
         task_metadata: TaskMetadata,
         prompt_type: PromptType | None,
     ) -> str | None:
-        """A wrapper function around the model.encode method that handles the prompt_name argument and standardizes the output to a numpy array.
+        """A wrapper function around the mol.encode method that handles the prompt_name argument and standardizes the output to a numpy array.
```
Samoed: I don't think that was needed

ayush1298:
I have a doubt. Do I have to update `AbsEncoder` and `RerankerWrapper`, and then all models using it or not?
Also for `SearchProtocol`, there is no `__init__()` method, so I haven't changed anything there.

mteb/models/abs_encoder.py
ayush1298: I think this was made by `make lint`, will revert this.

mteb/models/instruct_wrapper.py
```diff
@@ -40,6 +40,7 @@ def __init__(
             self,
             model_name_or_path: str,
             mode: str,
+            device: str | None = None,
```
Samoed, line 45: It should add it as a last arg, otherwise you'd break compatibility

mteb/models/model_implementations/bmretriever_models.py
```diff
@@ -52,6 +53,7 @@ def __init__(
 
         transformer = Transformer(
             model_name,
+            device,
```
KennethEnevoldsen:
```suggestion
            device=device,
```

mteb/models/model_implementations/nvidia_models.py
```diff
@@ -355,7 +356,9 @@ def __init__(
         self.attn_implementation = (
             "flash_attention_2" if torch.cuda.is_available() else "eager"
         )
-        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        self.device = device or torch.device(
+            "cuda" if torch.cuda.is_available() else "cpu"
+        )
```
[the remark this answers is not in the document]
KennethEnevoldsen: yeah device probably has to be converted here

Samoed:
I will check later other implementations

ayush1298:
> I have a doubt. Do I have to update `AbsEncoder` and `RerankerWrapper`, and then all models using it or not? Also for `SearchProtocol`, there is no `__init__()` method, so I haven't changed anything there.

@Samoed Whenever you get time, please answer this one also

Samoed:
I don't think you need to update AbsEncoder, because it does not implement a model and the same for the rerankerwrapper (as I remember)

ayush1298:
> I don't think you need to update AbsEncoder, because it does not implement a model and the same for the rerankerwrapper (as I remember)

I am saying about models which inherit AbsEncoder or rerankerwrapper like https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_implementations/listconranker.py#L20

Samoed:
Yeah, it's better to add to them too

ayush1298:
> Yeah, it's better to add to them too

For model using AbsEncoder, I have checked all, in most of them, device is already added, and where its not added, I have added.
For RerankerWrapper, its same, but still I want anyone to check it once.

ayush1298:
@Samoed Could you check this one when you have time

Samoed (approved):

Samoed:
Great work!

PR #3813, opened 2025-12-30
fix: repo exists check

Samoed:
Previously if run 
```python
import mteb

mteb.get_model("/path/to/model")
```
Would rise an error 
```
huggingface_hub.errors.HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': '/path/to/model'. Use `repo_type` argument if needed.
```


isaac-chung (approved):
Looks good. Would be even better to add a test case.

Issue #2785, opened 2025-06-07
Refactor `gme-Qwen` to use naitve `SentenceTransformer`


Samoed:
They were integrated with `SentenceTransformers`  https://huggingface.co/Alibaba-NLP/gme-Qwen2-VL-2B-Instruct

ethanvillalovoz:
Hi! I’d be interested in revisiting this issue.

I reviewed the previous attempt in #3818, which was closed after the native SentenceTransformers path failed on MTEB’s PIL-image batches. Since then, #4785 merged multimodal input handling into SentenceTransformerEncoderWrapper, and #4923 refined its modality filtering, so the underlying compatibility situation has changed.

My proposed scope would be to:

1. Update the 2B and 7B model revisions to their SentenceTransformers-integrated revisions.

2. Replace the duplicated model and image-processing implementation with SentenceTransformerEncoderWrapper, using a minimal GME-specific adapter only if needed for its prompt-in-input-dictionary behavior.

3. Preserve GME’s query-instruction formatting and document/default-instruction behavior.

4. Add focused tests for text-only, image-only, fused image-text, and prompt handling.

5. Run representative evaluations with the 2B model and compare its outputs and scores with the existing implementation before removing the custom code.

Is that the intended scope, and may I take this?

PR #3818, opened 2025-12-30
refactor gme-Qwen to use SentenceTransformer

ayush1298:
closes https://github.com/embeddings-benchmark/mteb/issues/2785

### DIFF
--- a/mteb/models/model_implementations/gme_v_models.py
+++ b/mteb/models/model_implementations/gme_v_models.py
@@ -76,21 +76,21 @@ class GmeQwen2VL(AbsEncoder):
                 # Fused-modal: both text and image
                 for text, image in zip(batch["text"], batch["image"]):
                     item = {"text": text, "image": image}
-                    if instruction and prompt_type == PromptType.query:
+                    if instruction:
                         item["prompt"] = instruction
                     batch_inputs.append(item)
             elif has_text:
                 # Text-only
                 for text in batch["text"]:
                     item = {"text": text}
-                    if instruction and prompt_type == PromptType.query:
+                    if instruction:
                         item["prompt"] = instruction
                     batch_inputs.append(item)
             elif has_image:
                 # Image-only
                 for image in batch["image"]:
                     item = {"image": image}
-                    if instruction and prompt_type == PromptType.query:
+                    if instruction:
                         item["prompt"] = instruction
                     batch_inputs.append(item)
             else:


mteb/models/model_implementations/gme_v_models.py
```diff
@@ -15,7 +13,7 @@
 from mteb.types import Array, BatchedInput, PromptType
 
 if TYPE_CHECKING:
-    from PIL import Image
+    pass
```
Samoed: Can you delete?

mteb/models/model_implementations/gme_v_models.py
```diff
@@ -30,144 +28,21 @@
...
-        min_pixels = min_image_tokens * 28 * 28
-        max_pixels = max_image_tokens * 28 * 28
-        processor = AutoProcessor.from_pretrained(
-            model_name,
-            revision=revision,
-            min_pixels=min_pixels,
-            max_pixels=max_pixels,
-            **kwargs,
+        self.model: SentenceTransformer = SentenceTransformer(
+            model_name, revision=revision, trust_remote_code=True, **kwargs
```
Samoed: Trust remote code better to add in loader kwargs
```suggestion
            model_name, revision=revision, **kwargs
```

mteb/models/model_implementations/gme_v_models.py
```diff
@@ -181,145 +56,60 @@ def encode(
...
+            has_image = "image" in batch
+
+            # Prepare batch inputs for SentenceTransformer
+            batch_inputs = []
+
+            if has_text and has_image:
+                # Fused-modal: both text and image
+                for text, image in zip(batch["text"], batch["image"]):
+                    item = {"text": text, "image": image}
+                    if instruction and prompt_type == PromptType.query:
```
Samoed: Otherwise, results on other task types, e.g. classification won't be the same 
```suggestion
                    if instruction:
```

mteb/models/model_implementations/gme_v_models.py
```diff
@@ -360,7 +150,7 @@ def fetch_image(image: Image.Image, size_factor: int = IMAGE_FACTOR) -> Image.Im
     max_tokens=32768,
     reference="https://huggingface.co/Alibaba-NLP/gme-Qwen2-VL-2B-Instruct",
```
Samoed, line 152: You need to update revision of the models to run with newer implementation

ayush1298:
> Can you run it to verify that scores are matching?

yes, doing that

mteb/models/model_implementations/gme_v_models.py
ayush1298: Can you tell me, why is it so?

ayush1298:
@Samoed Should I remove `GmeQwen2VL` class and use `SentenceTransformerMultimodalEncoderWrapper` directly?

Samoed:
You can try, but I'm unsure that they're fully compatible. At least in gme have some custom logic for insturictions with adding `.`

mteb/models/model_implementations/gme_v_models.py
Samoed: This wasn't in original implementation. Prompt type is applied only for reranking/retrieval tasks

ayush1298:
Can you once try running it? I am not getting any error when I am running it locally, but on kaggle I am getting errors.

mteb/models/model_implementations/gme_v_models.py
```diff
@@ -181,145 +54,60 @@ def encode(
...
+
+            # Prepare batch inputs for SentenceTransformer
+            batch_inputs = []
+
+            if has_text and has_image:
+                # Fused-modal: both text and image
+                for text, image in zip(batch["text"], batch["image"]):
+                    item = {"text": text, "image": image}
+                    if instruction:
+                        item["prompt"] = instruction
```
Samoed, line 80: You shouldn't add prompt for each item. SentenceTransformers don't support this format

ayush1298:
> I've run and got `TypeError: object of type 'PngImageFile' has no len()`. I think you can leave gme implementation as is for now.

okay, will close this PR, then

PR #3819, opened 2025-12-31
fix: Extend framework annotations for `ModelMeta`

### LINKED ISSUE #1622: Leaderboard: Add ONNX and GGUF to frameworks to allow us to filter based on these
See #1579

I believe a way to do this automatically would be to consider Frameworks as a combined set from mteb and those derived from huggingface (we can create a whitelist of tags to look for)

e.g. [Me5-large-instruct](https://huggingface.co/intfloat/multilingual-e5-large-instruct) would have [sentence-transformers, Transformers, ONNX, safetensors]. 
We could potentially rename frameworks to "tags" or split it up into "frameworks" and "formats".

I am unsure if this is valid to fetch during the setup (if we run it once a day I believe it would be) - @x-tabdeveloping WDYT?

ayush1298:
closes #1622


ayush1298:
@Samoed Somehow, when I am trying to add gme_v_models.py and kalm_models.py, git is not allowing me for that, and giving me error:

```
> git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file -
Validate pyproject.toml..............................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
check json...........................................(no files to check)Skipped
pretty format json...................................(no files to check)Skipped
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
check toml...........................................(no files to check)Skipped
typos....................................................................Failed
- hook id: typos
- exit code: 2
```

This was also happening when I was trying to add all files at a time, so I had to add 4-5 files in a commit. Do you have any idea why this is happening?

Samoed:
You can add flag `--no-verify` to skip pre-commit. I tried to run typos check on your files, but I don't have any issues

mteb/models/model_meta.py
```diff
@@ -56,6 +56,10 @@
     "PyLate",
     "ColBERT",
     "ColPali",
+    "GGUF",
+    "safetensors",
+    "ONNX",
+    "Transformers",
```
Samoed, line 63: Can you add automatization for this too to ModelMeta?

ayush1298: Doesn't get these exactly? What automation here?

ayush1298:
I will remove script and csv before merge.
Should I add these filter under Advance Filters in this PR only, or in a separate one?

mteb/models/model_meta.py
Samoed: Extracting tags from model card in `ModelMeta.from_hub`

Samoed:
It's better to add this in separate PR. I'm not sure if we need this filter on leaderboard for now

ayush1298:
> It's better to add this in separate PR. I'm not sure if we need this filter on leaderboard for now

Okay, will wait for @KennethEnevoldsen opinion on this

mteb/models/model_meta.py
ayush1298: I have added these.

mteb/models/model_meta.py
```diff
@@ -646,6 +632,63 @@ def fetch_release_date(model_name: str) -> StrDate | None:
...
+                return []
+
+            # Master framework list to match against HuggingFace tags
+            master_framework_list = {
+                "sentence-transformers",
+                "transformers",
+                "onnx",
+                "safetensors",
+                "gguf",
+            }
```
Samoed: You can just use tag_to_framework
```suggestion
```

mteb/models/model_meta.py
```diff
@@ -646,6 +632,63 @@ def fetch_release_date(model_name: str) -> StrDate | None:
...
+
+            for tag in info.tags:
+                tag_lower = tag.lower()
+                if tag_lower in master_framework_list:
+                    framework_name = tag_to_framework.get(tag_lower)
+                    if framework_name and framework_name not in frameworks:
+                        frameworks.append(framework_name)
+
+            return frameworks
+        except Exception as e:
```
Samoed, line 651: I think it's better to include in try catch only getting model_info

mteb/models/model_meta.py
```diff
@@ -656,41 +656,32 @@ def _get_frameworks_from_hf_tags(model_name: str) -> list[FRAMEWORKS]:
...
+        }
+
+        frameworks: list[FRAMEWORKS] = []
+
+        for tag in info.tags:
+            tag_lower = tag.lower()
+            if tag_lower in tag_to_framework.keys():
+                framework_name = tag_to_framework.get(tag_lower)
+                if framework_name and framework_name not in frameworks:
+                    frameworks.append(framework_name)
```
```suggestion
                frameworks.add(tag_to_framework[tag_lower])
```

mteb/models/model_meta.py
```diff
@@ -656,41 +656,32 @@ def _get_frameworks_from_hf_tags(model_name: str) -> list[FRAMEWORKS]:
...
+        # Mapping from HuggingFace tags to MTEB framework names
+        tag_to_framework: dict[str, FRAMEWORKS] = {
+            "sentence-transformers": "Sentence Transformers",
+            "transformers": "Transformers",
+            "onnx": "ONNX",
+            "safetensors": "safetensors",
+            "gguf": "GGUF",
+        }
+
+        frameworks: list[FRAMEWORKS] = []
```
Samoed, line 666:
```suggestion
        frameworks: set[FRAMEWORKS] = set()
```

Samoed (commented):
Overall looks good

mteb/models/model_meta.py
ayush1298: Why set and not list? There will not be duplicate items

Samoed: You have check that framework is not appears in list. I think you can simplify to set.

ayush1298: Will remove that checking, it's not required as . But, will prefer to keep this as list as frameworks is defined as list everywhere.

ayush1298: Simplified logic so as to avoid duplication.

mteb/models/model_meta.py
```diff
@@ -642,6 +632,53 @@ def fetch_release_date(model_name: str) -> StrDate | None:
             return release_date
         return None
 
+    def calculate_frameworks_from_hub(self) -> list[FRAMEWORKS]:
```
```suggestion
    def _fetch_frameworks_from_hub(self) -> list[FRAMEWORKS]:
```

KennethEnevoldsen: Assuming we don't want it to be user-facing

KennethEnevoldsen: I would actually just delete this function and directly call the static one:
```
frameworks = fn(self.name) if self.name else []
```

KennethEnevoldsen (approved):
I think this looks good. Only a few minor changes. We could argue if `safetensors` is a framework, but I think people would appreciate being able to filter based on it

mteb/models/model_meta.py
ayush1298: okay will do it.

ayush1298:
> It's better to add this in separate PR. I'm not sure if we need this filter on leaderboard for now

@KennethEnevoldsen Do we have to add this filter or not on LB for now?

ayush1298:
I have kept only the static method and removed user facing one. Also deleted both script and csv.

ayush1298:
Anything left here @Samoed ?

PR #3820, opened 2026-01-01
activate `LOG` rule

Samoed:
Activated `LOG` rule


mteb/abstasks/retrieval_dataset_loaders.py
```diff
@@ -136,7 +136,7 @@ def _load_corpus(self) -> CorpusDatasetType:
                 "_id", "id"
             )
         logger.info("Loaded %d %s Documents.", len(corpus_ds), self.split.upper())
-        logger.info("Doc Example: %s", corpus_ds[0])
+        logger.debug("Doc Example: %s", corpus_ds[0])
```
Samoed, line 139: Changed level here, because for image and audio tasks, this not gives helpful information

PR #3828, opened 2026-01-02
fix: Add leaderboard docker workflow

### LINKED ISSUE #3825: Add Leaderboard Dockerfile in mteb repo

The [leaderboard dockerfile](https://huggingface.co/spaces/mteb/leaderboard/blob/main/Dockerfile) is currently not in the mteb repo, which means new changes are not completely tested before merging. This led to [this issue](https://github.com/embeddings-benchmark/mteb/pull/3702#issuecomment-3702330726) and [this issue](https://github.com/embeddings-benchmark/mteb/issues/3821).

isaac-chung:
Fixes #3825 and #3829

Add the leaderboard Dockerfile here and test it properly.

  - New Dockerfile: Creates a Python 3.12 container that clones the MTEB repo, installs dependencies with uv, and runs the leaderboard via make run-leaderboard
  - CI workflow: Adds .github/workflows/leaderboard_docker.yml to automatically test Docker builds with smart startup detection and timeout handling. This is only triggered by changes to specific files.
  - Build the LB docker image and push to GHCR (for caching and loading elsewhere)
  - Dependencies: Adds typer>=0.20.0 to leaderboard extras in pyproject.toml

The plan is for us to use the tested and built Dockerfile here in the [HF space](https://huggingface.co/spaces/mteb/leaderboard/blob/main/Dockerfile) in a separate PR.

## [Workflow Logs](https://github.com/embeddings-benchmark/mteb/actions/runs/20659988853/job/59320239884?pr=3828)
```
Starting leaderboard container...
✅ Detected leaderboard app startup
Container still running after 30s... (startup: true, init_complete: false)
2026-01-02 14:37:35,638 - PID:22 - __main__ - INFO - === Starting leaderboard app initialization ===
2026-01-02 14:37:35,638 - PID:22 - __main__ - INFO - Step 1/7: Loading all benchmark results...
Container still running after 60s... (startup: true, init_complete: false)
2026-01-02 14:37:50,904 - PID:22 - __main__ - INFO - Step 2/7 complete: Fetched 59 benchmarks in 0.00s
2026-01-02 14:37:50,904 - PID:22 - __main__ - INFO - Step 3/7: Processing all benchmarks (select_tasks + join_revisions)...
✅ Detected full app initialization complete!
🔄 Testing if server is responding...
✅ Server is responding - container fully operational!
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ ENV PATH="/home/user/.local/bin:$PATH"
 # Install uv
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
 
-RUN git clone https://github.com/embeddings-benchmark/mteb.git
-RUN chown -R user:user /mteb
+# Copy the current directory contents into the container
+COPY --chown=user:user . /mteb
 
 USER user
 WORKDIR /mteb


pyproject.toml
```diff
@@ -69,6 +69,7 @@ leaderboard = [
     "plotly>=5.24.0,<6.0.0",
     "cachetools>=5.2.0",
     "matplotlib>=3.9.4",
+    "typer>=0.20.0",
```
Samoed: I can't find where we use it

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:3.12-bookworm
+
+RUN apt update && apt install -y git make curl
+RUN useradd -m -u 1000 user
+ENV PATH="/home/user/.local/bin:$PATH"
+
+# Install uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
+
+RUN git clone https://github.com/embeddings-benchmark/mteb.git
```
Samoed: Wouldn't it be mounted as a dir? We can then do `COPY . /mteb`

.github/workflows/leaderboard_docker.yml
```diff
@@ -0,0 +1,166 @@
...
+      uses: docker/setup-buildx-action@v3
+
+    - name: Log in to GitHub Container Registry
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Build and push Docker image
```
Samoed, line 57: I think we should publish image only from main

Samoed (commented):
Overall this looks good

pyproject.toml
isaac-chung: Needed for gradio in my local env. Can remove to check it in CI

.github/workflows/leaderboard_docker.yml
isaac-chung: Oh yes, good catch.

Dockerfile
isaac-chung: Maybe I'm missing what the issue is here?

Samoed: This clone won't work correctly from different branches. I think it's easier to just use copy instead

isaac-chung: Good point. My hopes are to test the Dockerfile that will be in the HF space. Can we do `COPY . /mteb` for that purpose?

Samoed: No, but if we're building and publishing an image here, then we can simply reuse it on spaces

isaac-chung: Oh true! There we could do smth like:
```
FROM ghcr.io/embeddings-benchmark/mteb/leaderboard
```

pyproject.toml
isaac-chung: Removed.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ENV PATH="/home/user/.local/bin:$PATH"
+
+# Install uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
+
+# Copy the current directory contents into the container
+COPY --chown=user:user . /mteb
+
+USER user
+WORKDIR /mteb
```
Samoed, line 14:
```suggestion
USER user
WORKDIR /mteb

COPY . .
```

isaac-chung: This suggestion fails: Permission denied error when uv sync tries to access .venv directory. I'll merge as is.

PR #3831, opened 2026-01-03
Add typehint for encode kwargs

Samoed:
Created typehint for `EncodeKwargs`
```python
class EncodeKwargs(TypedDict):
    batch_size: NotRequired[int]
    show_progress_bar: NotRequired[bool]

```

### DIFF
--- a/mteb/_create_dataloaders.py
+++ b/mteb/_create_dataloaders.py
@@ -71,7 +71,6 @@ def _create_dataloader_for_retrieval_corpus(
         dataset: Corpus
         batch_size: Batch size for the dataloader.
 
-
     Returns:
         A dataloader with the corpus.
     """
@@ -101,7 +100,6 @@ def _create_text_dataloader_for_queries(
     Args:
         queries: A list of queries.
         batch_size: Batch size for the dataloader.
-
 
     Returns:
         A dataloader with the queries.
@@ -187,7 +185,6 @@ def _create_dataloader_for_queries_conversation(
     Args:
         queries: A list of queries.
         batch_size: Batch size for the dataloader.
-
 
     Returns:
         A dataloader with the queries.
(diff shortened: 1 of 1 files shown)


mteb/_evaluators/image/imagetext_pairclassification_evaluator.py
```diff
@@ -105,8 +109,8 @@ def __call__(  # type: ignore[override]
                     texts.append(row[col])
 
         text_embeddings = model.encode(
-            DataLoader(
-                Dataset.from_dict({"text": texts}),
+            _create_dataloader_from_texts(
```
isaac-chung, line 112: This doesn't seem related to type hints?

mteb/_create_dataloaders.py
```diff
@@ -297,6 +301,7 @@ def _create_image_dataloader(
         transform: A transformation function to apply to each image (e.g., converting to tensor).
         collate_fn: A custom collate function to handle batching.
 
+
```
isaac-chung: No op.

isaac-chung (commented):
Not sure I understand fully, if we can still pass in `dict[str, Any]` as encode_kwargs. Would love to see a test case for this as well.

Samoed:
> if we can still pass in dict[str, Any] as encode_kwargs

Yes, we still can do this without problem. Typehint won't remove any parameters from kwargs. Just helping to find possible problems

mteb/_evaluators/image/imagetext_pairclassification_evaluator.py
Samoed: It's a bit related. If user pass `show_progress_bar` or any unsupported kwarg of DataLoader here exception will be raised. I've changed to `_create_dataloader_from_texts` because it can handle any kwargs

Samoed:
I don't know, but I didn't change anything in pyproject or lock in this PR

Issue #3534, opened 2025-11-06
Add `num_proc` to dataset loading


Samoed:
no I did not, however on **all** datasets (including previous vidore benchmark, for which I didn't change any code 
) RGB conversion was quite slow on a few images, I manually added a `num_proc = 4` kwarg to speed up the process.

What really struck me is that it does not appear to be some abnormal images in one dataset, but rather a common processing issue for all datasets.

_Originally posted by @QuentinJGMace in https://github.com/embeddings-benchmark/mteb/pull/3514#discussion_r2498360267_

PR #3832, opened 2026-01-03
fix: use `num_proc` for dataset processing

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3534

Add `num_proc` for:
- `load_data`
- `dataset_transform`
- `push_to_hub`
- `create_dataloader`

This also would be essential for `MAEB`, because dataset processing can take a lot of time.

It will be passed through `mteb.evaluate(..., num_proc=1)`


mteb/types/_encoder_io.py
```diff
@@ -19,10 +19,12 @@ class EncodeKwargs(TypedDict):
     Attributes:
         batch_size: The batch size to use for encoding.
         show_progress_bar: Whether to show a progress bar during encoding.
+        num_proc: The number of processes to use for dataset processing.
     """
 
     batch_size: NotRequired[int]
     show_progress_bar: NotRequired[bool]
+    num_proc: NotRequired[int]
```
Samoed: Sentence Transformers doesn’t support this kwarg, which is why the tests are failing. We can either create a separate set of kwargs for processing or just pass `num_proc` everywhere (it would also be required when creating dataset loaders).

Samoed: Moved out from encoding kwargs to use directly in `evaluate`

Samoed:
Seems that problem was in dataloader, because 

```
num_workers (int, optional) – how many subprocesses to use for data loading. 0 means that the data will be loaded in the main process. (default: 0)
```
and even set it to 1, then it would try to load in different process. Changed to `num_workers=num_proc if num_proc > 1 else 0,`

mteb/abstasks/abstask.py
```diff
@@ -308,11 +315,14 @@ def stratified_subsampling(
             )  # only take the specified test split.
         return dataset_dict
 
-    def load_data(self) -> None:
+    def load_data(self, num_proc: int = 1) -> None:
```
Samoed: Also, should we add `kwargs` back to `load_data`? I think this is currently a bit of a breaking change for forks - if `load_data` doesn’t expect parameters and is called with a newer version of MTEB, it will break. Adding `kwargs` back would help with future compatibility.

KennethEnevoldsen (approved):

mteb/abstasks/abstask.py
KennethEnevoldsen: Ahh right, so older task implementations will break. I would be fine adding kwargs in.

KennethEnevoldsen:
> and even set it to 1, then it would try to load in different process. Changed to num_workers=num_proc if num_proc > 1 else 0,

I think this change is fine - alternatively we could do `None` as the default (or `0` to align with dataloaders, but it seems misleading)

Samoed:
Added kwargs to load_data

PR #3834, opened 2026-01-03
docs: Resolve problems with missing documentation links

Samoed:
- Added private and aggregated tasks to documentation
- Fixed problem with missing tasks on benchmark generation

### DIFF
--- a/docs/overview/prettify_list.py
+++ b/docs/overview/prettify_list.py
@@ -1,11 +1,3 @@
-import re
-
-
-def slugify_anchor(value: str) -> str:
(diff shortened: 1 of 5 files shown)


docs/overview/prettify_list.py
KennethEnevoldsen, line 1: I would just include this in the two - generally not a fan of utils files as it is unclear what it contains

KennethEnevoldsen: You can also split it in two scripts e.g. `slugify.py`

Samoed: Split into `slugify` and `pretty_list`

PR #3835, opened 2026-01-03
fix dataset generation tags

Samoed:
Remove `hf_subtypes` from `task_categories` to prevent warnings like 
image


KennethEnevoldsen (approved):
can we fix an example just to see that it works as intended?

Samoed:
I updated https://huggingface.co/datasets/mteb/JaCWIRRetrievalLite and warning disappeared

PR #3837, opened 2026-01-03
fix: Update metadata to include active number of parameter to `ModelMeta`

### LINKED ISSUE #3258: Add active/embedding params separately

Hello!

I'd like to propose to add 2 columns to the MTEB leaderboard. In addition to total number of parameters, the leaderboard could also display the number of _active_ parameters and number of _embedding_ parameters. I think this is useful because total number of parameters actually does not reflect runtime really well, only peak RAM usage.

Two motivating examples:
* Multilingual models tend to have very large embedding tables, but similar number of active parameters to their monolingual counterparts, making them look larger than they are. Example: [google/embeddinggemma-300m](https://huggingface.co/google/embeddinggemma-300m) has 200M embedding parameters, and only 100M active parameters. So the whole model is actually BERT-sized in throughput.
* Static models have 0 active parameters, but can still have millions of parameters. The speed of these models actually does not fluctuate as a function of the number of parameters. Example: [minishlab/potion-multilingual-128M](https://huggingface.co/minishlab/potion-multilingual-128M) has 128M parameters, but is approximately as fast as [minishlab/potion-base-8M](https://huggingface.co/minishlab/potion-base-8M)

In these cases, the total number of parameters is not informative to users, and sorting/filtering by number of parameters actually leads users to exclude models that could be useful to them. 

# Implementation

I think the implementation could be relatively straightforward, if the model can be loaded using transformers. Here's a sketch:

```python
import numpy as np

model_name = "google/embeddinggemma-300m"
model = AutoModel.from_pretrained(model_name)

input_params = np.prod(model.get_input_embeddings().weight.shape)
total_params = np.sum([np.prod(x.shape) for x in list(model.parameters())])
active_params = total_params - input_params
```

What I suggest is that we make a script that:

1. downloads the model
2. runs this code
3. optionally deletes the model from the cache

The addition of step 3 allows this script to be run on some low-end machine without it running out of disk space.

For models that can't be loaded using transformers (for example, the static models), users should be able to submit their own estimates.

ayush1298:
closes #3258


ayush1298:
Just added csv for now. Will remove memory usage column, add active embedding to ModelMeta of each model and then add column to LB.

ayush1298:
@Samoed @KennethEnevoldsen @isaac-chung 
1 doubt that I have is, value of `n_parameters` (total number of parameters) which we have in ModelMeta, is almost equal to(though not exact) same as `total_params + embedding_params` counted using code. 

For model: `Alibaba-NLP/gte-Qwen2-7B-instruct`

```
  ModelMeta:  7,613,000,000
  In CSV:  total_params - 7069121024, active_params - 6525621760, embedding_params - 543499264 
so, total_params + embedding_params = 7,612,620,288 which is almost equal to what we have in ModelMeta.
```

Also in csv these 3 parameters came from below code. 

```python
import numpy as np

model_name = "google/embeddinggemma-300m"
model = AutoModel.from_pretrained(model_name)

input_params = np.prod(model.get_input_embeddings().weight.shape)
total_params = np.sum([np.prod(x.shape) for x in list(model.parameters())])
active_params = total_params - input_params
```

So, what exactly, I should add in ModelMeta? Should I add 2 new field in ModelMeta with name `active_parameter`, and `embedding_parameter`, and should we show both of these on LB, and remove `memory_usage` and `number of parameters`?

Samoed:
Let's modify leaderboard in separate PR

>  1 doubt that I have is, value of n_parameters (total number of parameters) which we have in ModelMeta, is almost equal to(though not exact) same as total_params + embedding_params counted using code.

Yes, this is expected that total parameters = active parameters + embedding paramters

> Should I add 2 new field in ModelMeta with name active_parameter, and embedding_parameter

Probably yes 

> should we show both of these on LB, and remove memory_usage and number of parameters?

Let's discuss this in issue

KennethEnevoldsen:
Converting this to a draft given the current state, but great to see it!

- Should we have a script to compute active parameters (e.g. the automatic metadata). Should we consider how we handle MoE? (this is where active != total - embedding) 
- Leaderboard: Let us deal with that in a seperate PR
- embedding parameters: If we define active parameters as total - embedding (might be a good enough assumption) I would probably just add embedding parameters and let the other one be a property. We could potentially allow an overwrite using e.g. `self._active_parameters`

ayush1298:
> * Should we have a script to compute active parameters (e.g. the automatic metadata). Should we consider how we handle MoE? (this is where active != total - embedding)

Yes, I will be adding the logic to calculate automatically in from_hub method. I think MOE are case, where I was not able to calculate it. Hence I think atleast in ModelMeta we can keep n_parameters and add both active and embedding. Also, as I have already calculated these parameters for around 315 models, and around 60 models are propritory out of 498 models we have currently. 

> * embedding parameters: If we define active parameters as total - embedding (might be a good enough assumption) I would probably just add embedding parameters and let the other one be a property. We could potentially allow an overwrite using e.g. `self._active_parameters`

For now, we can use csv. But, I think its better to keep both in ModelMeta as then we could also handle it for MOE case, where these property will not be useful.

ayush1298:
I have updated files with these meta and also added it to `.from_hub()` method. I have also added script that I have used to update ModelMeta. Will remove these script at end.

ayush1298:
Previously when I was doing `make lint`, it was fixing errors itself (before adding uv everywhere), but now somehow it was just giving error, but not fixing them. Why this was happening?

Samoed:
Strange. I don't have this behavior

KennethEnevoldsen:
Hmm do we want to use the csv? I think we should rather use the ModelMeta object - if not then I think we should use the json format similar to what we do with descriptive statistics

Samoed:
I think csv just for demonstration

ayush1298:
> Hmm do we want to use the csv? I think we should rather use the ModelMeta object - if not then I think we should use the json format similar to what we do with descriptive statistics

Its just for demonstration of all results I have calculated, and I had filled ModelMeta of models using that csv. 
For future if someone wants to calculate or add it for models where I am not able to calculate, then can just do it in below way: 

```
model_meta = mteb.get_model_meta("model_name")
active_parameters, embedding_parameters  = model_meta.extract_parameter_breakdown_from_hub
```

mteb/models/model_implementations/andersborges.py
```diff
@@ -43,6 +45,8 @@
     revision="cb576c78dcc1b729e4612645f61db59929d69e61",
     release_date="2025-11-21",
     n_parameters=48578560,
+    active_parameters=None,
+    embedding_parameters=None,
```
KennethEnevoldsen: Do we want `None` to be valid?

mteb/models/model_meta.py
```diff
@@ -575,6 +587,40 @@ def calculate_num_parameters_from_hub(self) -> int | None:
...
+
+    def extract_parameter_breakdown_from_hub(self) -> tuple[int | None, int | None]:
+        """Extracts the breakdown of active and embedding parameters from the model.
+
+        Returns:
+            A tuple containing the number of active parameters and embedding parameters.
+        """
+        if not self.name:
+            return None, None
+        return self._extract_parameter_breakdown_from_hub(self.name)
```
KennethEnevoldsen: I would delete this
```suggestion
```

mteb/models/model_meta.py
```diff
@@ -93,7 +94,9 @@ class ModelMeta(BaseModel):
         loader_kwargs: The keyword arguments to pass to the loader function.
         name: The name of the model, ideally the name on huggingface. It should be in the format "organization/model_name".
         n_parameters: The number of parameters in the model, e.g. 7_000_000 for a 7M parameter model. Can be None if the number of parameters is not known (e.g. for proprietary models) or
-            if the loader returns a SentenceTransformer model from which it can be derived.
+            if the loader returns a SentenceTransformer model from which it can be derived. These are total number of parameter of model (active_parameters + embedding_parameters).
```
```
      n_parameters: The total number of parameters in the model, e.g. `7_000_000` for a 7M parameter model. Can be none in case the number of parameters is unknown.

File: mteb/models/model_meta.py [hunk 4]
```diff
@@ -93,7 +94,9 @@ class ModelMeta(BaseModel):
...
+        active_parameters: The number of active (trainable) parameters in the model. Can be None if the number of active parameters is not known (e.g. for proprietary models). To calculate it use the `extact_parameter_breakdown_from_hub` method
```
```suggestion
```

mteb/models/model_meta.py
```diff
@@ -93,7 +94,9 @@ class ModelMeta(BaseModel):
...
+        embedding_parameters: The number of parameters used for embedding generation. Can be None if the number of embedding parameters is not known (e.g. for proprietary models). To calculate it use the `extract_parameter_breakdown_from_hub` method.
```
```suggestion
        embedding_parameters: The number of parameters used for the embedding layer. Can be None if the number of embedding parameters is not known (e.g. for proprietary models).
```
I think we want people to use the `from_hub` or `from_sentence_transformers` etc. entry points

mteb/models/model_meta.py
```diff
@@ -132,6 +135,8 @@ class ModelMeta(BaseModel):
     release_date: StrDate | None
     languages: list[ISOLanguageScript] | None
     n_parameters: int | None
+    active_parameters: int | None = None
```
KennethEnevoldsen: Would replace with:


```
_active_parameters: int | None = None

@property:
def n_active_parameters(self):
    """Number of active parameters. Assumed to be `n_parameters - n_embedding_parameters`. Can be overwritten using `_n_active_parameters` e.g. for MoE models."""
    if self._n_active_parameters:
      return self._n_active_parameters
    return self.n_parameters - self.n_embedding_parameters
```

mteb/models/model_meta.py
```diff
@@ -575,6 +587,40 @@ def calculate_num_parameters_from_hub(self) -> int | None:
         """
         return self._calculate_num_parameters_from_hub(self.name)
 
+    @staticmethod
+    def _extract_parameter_breakdown_from_hub(
```
KennethEnevoldsen: would convert to just extracting embedding parameters. Also needs to be added for the sentence transformers models (no need to load the models twice)

mteb/models/model_implementations/andersborges.py
KennethEnevoldsen: (for the model2vec models `n_embedding_parameters==n_parameters`.

mteb/models/model_meta.py
```diff
@@ -132,6 +135,8 @@ class ModelMeta(BaseModel):
     release_date: StrDate | None
     languages: list[ISOLanguageScript] | None
     n_parameters: int | None
+    active_parameters: int | None = None
+    embedding_parameters: int | None = None
```
```suggestion
    n_embedding_parameters: int | None = None
```
for consistency

mteb/models/model_meta.py
ayush1298: This is added under _from_hub(), and sentence transformer calls the same thing: 
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_meta.py#L386

mteb/models/model_meta.py
ayush1298: yes they can, but if they just want to calculate these, similar to `calculate_memory_usage` thing then they can directly call these like: 

```
model_meta = mteb.get_model_meta("model_name")
active_parameters, embedding_parameters  = model_meta.extract_parameter_breakdown_from_hub
```
Will convert active parameter to property.

mteb/models/model_implementations/andersborges.py
ayush1298: So, here I have set None because I saw n_parameter is also set to None when the value is not known. Now, in my case, for some of the models, I was not able to calculate values (leave proprietary models) because of either the very large size of the models, and hence for them also I have set to None for these parameter values in their ModelMeta.

mteb/models/model_meta.py
KennethEnevoldsen: Hmm but I `from_hub` currently doesn't load the model at all (I am not sure we want it to). Maybe it is better just to add it to the two others? (@Samoed WDYT?)

mteb/models/model_meta.py
KennethEnevoldsen: Let us solve the other ones and get back to this one afterwards

mteb/models/model_implementations/andersborges.py
ayush1298: > (for the model2vec models `n_embedding_parameters==n_parameters`.

Okay, will set these for all model2vec models, somehow I was not able to run them, I was getting error: 

`error: The checkpoint you are trying to load has model type `model2vec` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.`

mteb/models/model_meta.py
ayush1298: I agree, Its better to add separately. As it's correct that this has nothing to do with `from_hub()`

mteb/models/model_meta.py
ayush1298: Just 1 doubt here, after adding this property, I have to completely remove active_parameters from individual ModelMeta as well, right? While for all MOE models, I need to change `active_parameters` to `_active_parameters` in ModelMeta?

mteb/models/model_meta.py
Samoed: Agree, this is better to add as separat method

ayush1298:
Converted `active_parameters` to property. I will check for which models `n_parameter != embedding + active` based on results in csv, and only for them will keep `_n_active_parameters` field in ModelMeta and will remove these field for rest of all models.

mteb/models/model_implementations/andersborges.py
Samoed: Yes, this should be valid for private api only models

ayush1298: > Yes, this should be valid for private api only models

Are you saying error is valid?

mteb/models/model_meta.py
KennethEnevoldsen: yes but we would need to calculate them differently

mteb/models/model_implementations/andersborges.py
Samoed: I'm saying that `None` is valid. What about to error, you need to use model2vec library https://github.com/MinishLab/model2vec to load them. You can do this by `mteb.get_model`

ayush1298:
@KennethEnevoldsen @Samoed How do I actually check for MOE, here problem is,` n_parameters` which we already have will not be exactly matching with `embedding + active` as n_parameters was reported and not excatly calculated by loading model, So, how I should decide which one is MOE, I tried keeping some threshold between difference of `embedding + active` value and `n_parameters` value, but I am not sure how correct it is to consider MOE by that way?

Samoed:
> n_parameters which we already have will not be exactly matching with embedding + active

Yes, but you can set `_n_active_parameters` and it will be returned correctly.

> So, how I should decide which one is MOE, I tried keeping some threshold between difference of embedding + active value and n_parameters value, but I am not sure how correct it is to consider MOE by that way?

I'm not sure about automating it. I think model authors can submit their score by themselves. You can try to test this on `nomic-ai/nomic-embed-text-v2-moe`

ayush1298:
> > n_parameters which we already have will not be exactly matching with embedding + active
> 
> Yes, but you can set `_n_active_parameters` and it will be returned correctly.
> 
But, kenneth want to have it only for MOE models in ModelMeta, and for rest of models it should not be present. So, I was thinking how to get MOE. I am okay with keeping it for all models in ModelMeta, but then what's point of keeping it as a property? though it will be used for future models where results are not reported.

> > So, how I should decide which one is MOE, I tried keeping some threshold between difference of embedding + active value and n_parameters value, but I am not sure how correct it is to consider MOE by that way?
> 
> I'm not sure about automating it. I think model authors can submit their score by themselves. You can try to test this on `nomic-ai/nomic-embed-text-v2-moe`

So, I think the best way is to keep `_n_active_parameters` in all ModelMeta for now. In the future, if results are not reported then property will be used in that case.

ayush1298:
1 more question in ModelMeta, it can't be named as `_n_active_parameters`, right as it becomes private variable, we have to name it as `active_parameters` or `n_active_parameters`?

Samoed:
> But, kenneth want to have it only for MOE models in ModelMeta

Yes, you shouldn't fill it for other models, I think. I don't see a problem here.

> we have to name it as active_parameters or n_active_parameters?

I think you can name change `_n_active_parameters` to `n_active_parameters` and name property as `active_parameters`

ayush1298:
> > But, kenneth want to have it only for MOE models in ModelMeta
> 
> Yes, you shouldn't fill it for other models, I think. I don't see a problem here.
> 

Yes, problem is just to identify which models are MOE.

ayush1298:
I have added `_calculate_embedding_parameters` to `from_cross_encoder` and `from_senetence_tranformer`. Do we have to keep `calculate_embedding_parameters` to access it directly for someone or not is yet to decide.
I have also updated model2vec models with `n_embedding_parameters=n_parameters`

As per my analysis, below are list of model2vec models, let me know if there is any model which is not, or are there any other model2vec models: 

1. NeuML/pubmedbert-base-embeddings-100K
2. NeuML/pubmedbert-base-embeddings-1M
3. NeuML/pubmedbert-base-embeddings-2M
4. NeuML/pubmedbert-base-embeddings-500K
5. NeuML/pubmedbert-base-embeddings-8M
6. minishlab/M2V_base_glove
7. minishlab/M2V_base_glove_subword
8. minishlab/M2V_base_output
9. minishlab/M2V_multilingual_output
10. minishlab/potion-base-2M
11. minishlab/potion-base-4M
12. minishlab/potion-base-8M
13. minishlab/potion-multilingual-128M
14. rasgaard/m2v-dfm-large

mteb/models/model_meta.py
```diff
@@ -384,6 +397,7 @@ def from_sentence_transformer_model(
             else model.model_card_data.base_model
         )
         meta = cls._from_hub(name, revision, compute_metadata)
+        meta.n_embedding_parameters = cls._calculate_embedding_parameters(name)
```
Samoed: Here you already have SentenceTransformer model. Can you compute n parameters without creating another instance of the model?

ayush1298: Changed these one

mteb/models/model_meta.py
```diff
@@ -585,13 +585,17 @@ def calculate_num_parameters_from_hub(self) -> int | None:
...
+    def _calculate_embedding_parameters(
+        model: SentenceTransformer | CrossEncoder, model_name: str | None = None
+    ) -> int | None:
+        if model is None and model_name is None:
             return None
         n_embedding_parameters = None
         try:
-            model = AutoModel.from_pretrained(model_name, trust_remote_code=True)
-            emb = model.get_input_embeddings()
+            if hasattr(model, 'auto_model'):  # For SentenceTransformer
```
```suggestion
            if isinstance(model, SentenceTransformer):
```

mteb/models/model_meta.py
```diff
@@ -567,6 +584,27 @@ def calculate_num_parameters_from_hub(self) -> int | None:
         """
         return self._calculate_num_parameters_from_hub(self.name)
 
+    @staticmethod
+    def _calculate_embedding_parameters(
```
Samoed: Maybe this would be better to calculate directly in `from_cross_encoder` and `from_sentence_transformer_model`?

ayush1298: okay, will remove function and add there

ayush1298:
I don't know why a typecheck error is coming; I have also added explicit conditions for that. And I have also tested code for both sentence-transformer and cross-encoder models.  Also, there are some conflicts in dependencies.

mteb/models/model_meta.py
```diff
@@ -455,6 +479,16 @@ def from_cross_encoder(
         from mteb.models import CrossEncoderWrapper
 
         meta = cls._from_hub(model.model.name_or_path, revision, compute_metadata)
+        try:
+            if isinstance(model.model, PreTrainedModel):
```
Samoed: I don't think this check is necessary because you'll get valid cross-encoder model

mteb/models/model_meta.py
```diff
@@ -384,6 +397,17 @@ def from_sentence_transformer_model(
             else model.model_card_data.base_model
         )
         meta = cls._from_hub(name, revision, compute_metadata)
+        try:
+            first_module = model[0]
+            if hasattr(first_module, "auto_model"):
```
Samoed: Why do you need these checks?

mteb/models/model_meta.py
ayush1298: I though typecheck is failing because it was not able to identify this one. So, I have added explicit checks.

mteb/models/model_meta.py
ayush1298: How to avoid typecheck error, then?

ayush1298:
@Samoed @KennethEnevoldsen I am still not getting how to find out MOE models for which we have to keep `active_parameter`? as parameters are not matching exactly.

Samoed:
I don't think that we need to automate this for now

ayush1298:
> I don't think that we need to automate this for now

I am not talking about automation. Currently, n_active_parameters is present for all models in their ModelMeta, but we want it only for MOE.

Samoed:
I don't think we can do anything with it

mteb/models/model_meta.py
```diff
@@ -134,6 +137,8 @@ class ModelMeta(BaseModel):
     release_date: StrDate | None
     languages: list[ISOLanguageScript] | None
     n_parameters: int | None
+    n_active_parameters: int | None = None
```
```suggestion
    _n_active_parameters: int | None = None
```
I would set this as hidden and instead use the property.

mteb/models/model_meta.py
```diff
@@ -192,6 +197,16 @@ def is_cross_encoder(self) -> bool:
         """
         return "cross-encoder" in self.model_type
 
+    @property
+    def active_parameters(self):
```
```suggestion
    def n_active_parameters(self):
```

mteb/models/model_meta.py
```diff
@@ -192,6 +197,16 @@ def is_cross_encoder(self) -> bool:
...
+        """Number of active parameters. Assumed to be `n_parameters - n_embedding_parameters`. Can be overwritten using `active_parameters` e.g. for MoE models."""
```
```suggestion
        """Number of active parameters. Assumed to be `n_parameters - n_embedding_parameters`. Can be overwritten using `_n_active_parameters` e.g. for MoE models."""
```

mteb/models/model_meta.py
```diff
@@ -192,6 +197,16 @@ def is_cross_encoder(self) -> bool:
...
+        if self.n_active_parameters is not None:
+            return self.n_active_parameters
```
```suggestion
        if self._n_active_parameters is not None:
            return self._n_active_parameters
```

mteb/models/model_implementations/yuan_models_en.py
```diff
@@ -43,6 +43,8 @@ def instruction_template(
     revision="b2fd15da3bcae3473c8529593825c15068f09fce",
     release_date="2025-11-27",
     n_parameters=595776512,
+    n_active_parameters=None,
```
KennethEnevoldsen: delete throughout (create a seperate issue for MoEs)

mteb/models/model_meta.py
Samoed: Proprieties starting with `_` won't be treated by pydantic as fields https://docs.pydantic.dev/latest/concepts/models/#private-model-attributes

KennethEnevoldsen: Ahh, didn't know.

Maybe `n_active_parameters_override`?

Samoed: Renamed to `n_active_parameters_override`

mteb/models/model_implementations/yuan_models_en.py
Samoed: Deleted

model total and active parameters - model_parameters.csv
KennethEnevoldsen, line 1: I think can delete this now

mteb/models/model_implementations/amazon_models.py
```diff
@@ -8,6 +8,7 @@
     release_date="2024-04-30",
     languages=["eng-Latn"],
     n_parameters=None,
+    n_embedding_parameters=None,
```
KennethEnevoldsen, line 11: let us create an issue on filling these in and adding a test to prevent None

mteb/models/model_implementations/bge_models.py
```diff
@@ -502,6 +514,8 @@
     revision="4778d71a06863076696b03fd2777eb118712cad8",
     release_date="2023-08-05",  # initial commit of hf model.
     n_parameters=33_400_000,
+    n_active_parameters_override=21_639_552,
```
KennethEnevoldsen: can be removed throughout
```suggestion
```

tests/test_models/test_model_meta.py
```diff
@@ -160,6 +160,8 @@ def test_model_to_python():
     release_date='2021-08-30',
     languages=['eng-Latn'],
     n_parameters=22700000,
+    n_active_parameters_override=10992768,
```
```suggestion
```

mteb/models/model_implementations/amazon_models.py
ayush1298: @KennethEnevoldsen There are around I think 100 models for which I was not able to calculate these. So, do you want me to add list of all of them in a issue?

mteb/models/model_implementations/bge_models.py
ayush1298: Is it correct that you want me to remove these for all models?

KennethEnevoldsen: yeah it should be the same as `n_active_parameters` so there is not reason to add it

mteb/models/model_implementations/amazon_models.py
KennethEnevoldsen: That would be great!

ayush1298: Created issue for this, #3947

model total and active parameters - model_parameters.csv
ayush1298: Done

ayush1298:
I have deleted csv and created issue to fill n_embedding_parameters (#3947 ) and to calculate active_parameters for MOE models (#3948)

KennethEnevoldsen:
@ayush1298 will to resolve the merge - I think this is good to go. @Samoed I marked it as fix since it is only an update on the metadata, but would be fine with doing a feat as well.

PR #3841, opened 2026-01-04
Don't sync make lint

Samoed:
Currently with each lint syncs lock file. If I'm switching between prs it tries to update dependencies and this takes a lot of time

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -44,7 +44,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          make install-for-tests
+          make install
 
       - name: Lint
         id: lint
--- a/.github/workflows/typechecking.yml
+++ b/.github/workflows/typechecking.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          make install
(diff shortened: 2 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -44,7 +44,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          uv sync --group lint
+          make install
+          uv sync --group linting
```
KennethEnevoldsen: I would include in make install

.github/workflows/typechecking.yml
```diff
@@ -43,6 +43,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          make install
           uv sync --group typing --extra leaderboard
```
KennethEnevoldsen: same here

.github/workflows/lint.yml
Samoed: Removed

isaac-chung:
I guess this takes [a few trial and error](https://github.com/embeddings-benchmark/mteb/pull/3702#discussion_r2649079092), and at least we don't have 60+ comments this time 😅

PR #3845, opened 2026-01-05
[MAEB] init imports fix

AdnanElAssadi56:


KennethEnevoldsen:
@AdnanElAssadi56, tests are failing. Is this expected?

Samoed:
Yes. This is problem for all maev prs, because in cache dataset v4 version, but in maeb datasets v3 and datasets can't load them correctly

PR #3846, opened 2026-01-05
[MAEB] Ensure that Yamnet handles empty audio snippets

AdnanElAssadi56:


mteb/models/model_implementations/yamnet_models.py
```diff
@@ -72,7 +72,7 @@ def _normalize_audio(self, audio):
                 audio = audio[..., :max_length]
 
             # Normalize to [-1.0, 1.0]
-            if audio.abs().max() > 1.0:
+            if audio.numel() > 0 and audio.abs().max() > 1.0:
```
Samoed, line 75: Can you add description why this change is required?

AdnanElAssadi: It is again related to possible empty audio samples. If filtered properly, it shouldn't happen but it triggered for voice gender clustering (afair) when empty samples were not being filtered. It is just a safety check.

PR #3851, opened 2026-01-05
[MAEB] Make seemlessm45 model more efficient

AdnanElAssadi56:

### DIFF
--- a/mteb/models/model_implementations/seamlessm4t_models.py
+++ b/mteb/models/model_implementations/seamlessm4t_models.py
@@ -68,8 +68,9 @@ class SeamlessM4TWrapper(AbsEncoder):
                         orig_freq=sr, new_freq=self.sampling_rate
                     )
                     array = resampler(array)
-                # truncate audio and convert to numpy for processor
-                array = array.squeeze().cpu().detach().numpy()[:max_samples]
+
+                # Squeeze to 1D and truncate
+                array = array.squeeze()[:max_samples]
                 audio_arrays.append(array)
 
             # Process the entire batch at once


mteb/models/model_implementations/seamlessm4t_models.py
```diff
@@ -137,7 +139,7 @@ def get_audio_embeddings(
 
                 all_embeddings.append(embeddings.cpu())
 
-        return torch.cat(all_embeddings, dim=0)
+        return torch.cat(all_embeddings, dim=0).numpy()
```
KennethEnevoldsen, line 143: I don't think we need the numpy conversion

mteb/models/model_implementations/seamlessm4t_models.py
```diff
@@ -68,24 +68,26 @@ def get_audio_embeddings(
                         orig_freq=sr, new_freq=self.sampling_rate
                     )
                     array = resampler(array)
-                # truncate audio
-                array = array.cpu().detach().numpy()[:max_samples]
+                # truncate audio and convert to numpy for processor
```
KennethEnevoldsen: seems odd to convert it to numpy to convert it back to torch - is that really needed? (if so let us add a comment on why it is needed)

KennethEnevoldsen (approved):

mteb/models/model_implementations/seamlessm4t_models.py
AdnanElAssadi: I think torch padding and numpy padding differ in higher dimensions i.e if audio is 2D. For most cases, it is not needed, but I did this to avoid an error.

KennethEnevoldsen: It seems like we are avoiding the error rather than fixing it - not sure why numpy and torch would differ here, are you sure that it is not the `squeeze` that makes it work?

AdnanElAssadi:
@KennethEnevoldsen You are right, I've reviewed configs and processor accepts tensors. Should be good now.

PR #3858, opened 2026-01-05
[MAEB] Add safety checks for encodec model

AdnanElAssadi56:


mteb/models/model_implementations/encodec_model.py
```diff
@@ -46,8 +46,10 @@ def get_audio_embeddings(
         **kwargs: Any,
     ) -> Array:
         import torchaudio
+        import numpy as np
```
KennethEnevoldsen: top-level

mteb/models/model_implementations/encodec_model.py
```diff
@@ -76,22 +78,46 @@ def get_audio_embeddings(
                     array = resampler(array)
 
                 array = array.squeeze()
+                
+                # Handle edge case where squeeze results in 0-dim tensor
+                if array.dim() == 0:
+                    array = array.unsqueeze(0)
+                
+                # Ensure we have at least some samples
+                if len(array) == 0:
```
Samoed: I think it's better to filter out such audios in tasks

mteb/models/model_implementations/encodec_model.py
```diff
@@ -76,22 +78,46 @@ def get_audio_embeddings(
...
+                # Need enough samples to produce at least 1 frame in the latent space
+                # Being more conservative with 1 second minimum to ensure encoder works
+                min_samples = self.sampling_rate  # 1 second minimum
+                if len(array) < min_samples:
+                    padding = torch.zeros(min_samples - len(array))
+                    array = torch.cat([array, padding])
+                
+                # Pad or truncate to fixed length for consistent batching
+                if len(array) > max_samples:
+                    array = array[:max_samples]
```
Samoed, line 94: You can process max size and padding with processor https://github.com/huggingface/transformers/blob/a7f29523361b2cc12e51c1f5133d95f122f6f45c/src/transformers/models/encodec/feature_extraction_encodec.py#L87-L91

mteb/models/model_implementations/encodec_model.py
```diff
@@ -76,22 +78,46 @@ def get_audio_embeddings(
...
+                
+                # Encodec has multiple downsampling layers with total stride ~320
+                # Need enough samples to produce at least 1 frame in the latent space
+                # Being more conservative with 1 second minimum to ensure encoder works
+                min_samples = self.sampling_rate  # 1 second minimum
+                if len(array) < min_samples:
+                    padding = torch.zeros(min_samples - len(array))
+                    array = torch.cat([array, padding])
+                
+                # Pad or truncate to fixed length for consistent batching
```
Samoed: I think you'll truncate total number of samples in batch rather than length

mteb/models/model_implementations/encodec_model.py
KennethEnevoldsen: I would be fine with handling it, but we should probably add a warning log + warning

mteb/models/model_implementations/encodec_model.py
AdnanElAssadi: I don't think processor allows setting both truncation and padding to True.

Samoed: It will allow. 

```
padding (`bool`, `str` or [`~utils.PaddingStrategy`], *optional*, defaults to `True`):
                Select a strategy to pad the returned sequences (according to the model's padding side and padding
                index) among:

                - `True` or `'longest'`: Pad to the longest sequence in the batch (or no padding if only a single
                  sequence if provided).
                - `'max_length'`: Pad to a maximum length specified with the argument `max_length` or to the maximum
                  acceptable input length for the model if that argument is not provided.
            truncation (`bool`, *optional*, defaults to `False`):
                Activates truncation to cut input sequences longer than `max_length` to `max_length`.
```

AdnanElAssadi: And padding to max_length wouldn't exactly work because we need to also ensure a min_length when audio is too short. If the audio samples were too short, and after Encodec's multiple downsampling layers (stride), there will not be enough samples to produce even a single latent frame, resulting in a tensor with 0 elements in the time dimension.

And, by the way, it seems that the dataloader is passing empty batches when empty audio samples are not filtered. This can be triggered with the VoiceGenderClustering Dataset.

AdnanElAssadi: Lines 138:
`
        if padding and truncation:
            raise ValueError("Both padding and truncation were set. Make sure you only set one.")
`

Samoed: I see. I think you can use `padding` `True` by default and it would be enough

AdnanElAssadi: But we need both padding and truncation and a min_length. I think it is okay now.

Samoed:
Can you fix linting issue? Seems that you're not using numpy now

AdnanElAssadi:
Ran lint

PR #3859, opened 2026-01-05
Fix sound_desc dataset name typo

AdnanElAssadi56:


KennethEnevoldsen (commented):
Do we have current `SoundDescsA2TRetrieval` results, which should be deleted?

AdnanElAssadi:
I don't think we have the wrong results because the wrong dataset name with that revision probably don't exist on huggingface, so there will be an error before evaluation takes place.

isaac-chung:
@AdnanElAssadi56 would you mind running one small model like yamnet and check against existing results, just to be safe?

KennethEnevoldsen:
Merging do reduce number of active PRs, but I would still love to see a model run on this just to confirm that it works

Issue #3499, opened 2025-10-26
Reupload `MAEB` tasks with `trust_remote_code`


Samoed:
Now we have 3 tasks with `trust_remote_code`:
- [x] SpeechCommandsZeroshotv0.01 https://huggingface.co/datasets/mteb/SpeechCommandsZeroshotv0.01
- [x] SpeechCommandsZeroshotv0.02
- [x] VoxPopuliGenderClustering
- [x] BirdSet (blocked by https://github.com/embeddings-benchmark/mteb/pull/3863, because not clear split selection)
- [x] RavdessZeroshot
- [x] FleursA2TRetrieval
- [x] FleursT2ARetrieval

We need to reupload them. To do this you can run `task.push_dataset_to_hub(f"mteb/{task.metadata.name}")`

Samoed:
Reuploaded versions:

- SpeechCommandsZeroshotv0.01 https://huggingface.co/datasets/mteb/SpeechCommandsZeroshotv0.01
- SpeechCommandsZeroshotv0.02 https://huggingface.co/datasets/mteb/SpeechCommandsZeroshotv0.02
- VoxPopuliGenderClustering https://huggingface.co/datasets/mteb/VoxPopuliGenderClustering
- BirdSet https://huggingface.co/datasets/mteb/BirdSet
- RavdessZeroshot https://huggingface.co/datasets/mteb/RavdessZeroshot
- FleursA2TRetrieval https://huggingface.co/datasets/mteb/fleurs
- FleursT2ARetrieval https://huggingface.co/datasets/mteb/fleurs

PR #3863, opened 2026-01-05
Make birdset dataset handling more efficient

AdnanElAssadi56:


KennethEnevoldsen:
@Samoed can I ask to take this one?

mteb/tasks/multilabel_classification/zxx/bird_set.py
```diff
@@ -18,7 +20,7 @@ class BirdSetMultilabelClassification(AbsTaskMultilabelClassification):
         },
         type="AudioClassification",
         category="a2t",
-        eval_splits=["test_5s", "test"],
+        eval_splits=["test_5s"],
```
Samoed, line 21: Why do you remove test split?

mteb/tasks/multilabel_classification/zxx/bird_set.py
```diff
@@ -81,22 +88,47 @@ def dataset_transform(self):
...
-            seed=self.seed,
-            splits=self.eval_splits,
-            label=self.label_column_name,
-            n_samples=2048,
+        # Filter out empty labels
+        self.dataset = self.dataset.filter(
+            lambda x: x[self.label_column_name] is not None and len(x[self.label_column_name]) > 0
+        )
+
+        # Only subsample splits that are larger than n_samples to avoid division by zero
```
Samoed, line 105: Not sure why this is required

mteb/tasks/multilabel_classification/zxx/bird_set.py
AdnanElAssadi: It does not have multilabel classficistion format.

https://huggingface.co/datasets/DBD-research-group/BirdSet

Gave error during evaluation.

mteb/tasks/multilabel_classification/zxx/bird_set.py
AdnanElAssadi: Label had wrong type, and handling was taking some time.

Samoed:
Can you reupload this dataset? You can do it by `task.push_dataset_to_hub(f"mteb/{task.metadata.name}")`. I tried to do this on kaggle, but casting took too much memory. This would help to close https://github.com/embeddings-benchmark/mteb/issues/3499

Samoed:
By the way do you know why we're using only `HSN` subset?

AdnanElAssadi:
@imadtyx I see that you've first integrated this.  Can you please tell us about this?

>By the way do you know why we're using only HSN subset?

Samoed:
It seems that `HSN` is the smallest split of this task. 

image


I'm currently reuploading this dataset with all splits, but test split creates a lot of shards for some reason, I tried to set it to some value, but got datasets error, but I don't think that this would be a problem for us

AdnanElAssadi:
@Samoed Is the dataset currently on HF containing all splits?

Also, I wonder how feasible it is to run BirdSet evaluation once again on all the splits, considering we've already evaluated most (if not all) models already on HSN. 

The stats from the screenshot look heavy, did you subsample in a specific way?

Samoed:
I'm uploading now. I didn't do anything for processing, I just reupload it without trust remote code

AdnanElAssadi:
Maybe we can try including all splits in HF, and depending on whether we have time to evaluate, we indicate in paper what split was used.

Samoed:
I'm reuploading all splits, but for now I've reuploaded HSN https://huggingface.co/datasets/mteb/BirdSet

AdnanElAssadi:
Great @Samoed!

Do you have a specific idea in mind on how we want to handle the different subsets? This will probably require us to specify the subset config when loading dataset. 

For eval splits, I think we should still stick to test_5s if we are doing multilabel classification.

Samoed:
> For eval splits, I think we should still stick to test_5s if we are doing multilabel classification.

Yes, test audio for 10 min long and I think this is too much.

> Do you have a specific idea in mind on how we want to handle the different subsets? This will probably require us to specify the subset config when loading dataset.

Yes, this can be done as other multilingual classification tasks, e.g. MTOP https://github.com/embeddings-benchmark/mteb/blob/8c96b79985a038924dd89eb7d982f01ee9d8967d/mteb/tasks/classification/multilingual/mtop_domain_classification.py#L14-L27

Samoed:
@AdnanElAssadi56 Do you want to merge this?

isaac-chung:
Can this be merged after resolving conflicts? We can probably discuss the other subsets separately?

AdnanElAssadi:
@Samoed @isaac-chung 
No more conflicts, I think this can be merged for now. Adding other splits can be done separately.

PR #3864, opened 2026-01-05
Filter out empty samples in voice_gender_clustering

AdnanElAssadi56:

### DIFF
--- a/mteb/tasks/audio/clustering/eng/voice_gender.py
+++ b/mteb/tasks/audio/clustering/eng/voice_gender.py
@@ -38,6 +38,12 @@ class VoiceGenderClustering(AbsTaskClustering):
     input_column_name: str = "audio"
 
     def dataset_transform(self):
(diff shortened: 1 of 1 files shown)


AdnanElAssadi:
> This will break reproducibility of our current results (which is fine, but then we need to be able to rerun it)

I didn't quite get what you mean.

Samoed:
> I didn't quite get what you mean.

We will need to reevaluate models for this task

AdnanElAssadi:
I see, how do you suggest we handle this?

We should either fix it here, or in the dataloader. This dataset seems to be giving empty batches, and I guess some models were able to handle this and others not.

Samoed:
Fix here is fine. I don't see problems here

isaac-chung:
Yeah - just need to be rerun

isaac-chung (approved):

PR #3868, opened 2026-01-05
[MAEB] Model Meta Revision Fix

AdnanElAssadi56:
The "N/A" format in revision messes up the directory structure in the Results Repo as it gets intepreted as two subdirectories.

### DIFF
--- a/mteb/models/model_implementations/speecht5_models.py
+++ b/mteb/models/model_implementations/speecht5_models.py
@@ -223,7 +223,7 @@ class SpeechT2Multimodal(AbsEncoder):
     ):
         # Revision is combined as "asr_revision-tts_revision"
         asr_revision, tts_revision = revision.split("-")
-        
+
         self.asr_encoder = SpeechT5Audio(
             model_name=model_name,
             revision=asr_revision,


mteb/models/model_implementations/speecht5_models.py
```diff
@@ -332,7 +332,7 @@ def encode(
     name="microsoft/speecht5_multimodal",
     languages=["eng-Latn"],
     open_weights=True,
-    revision="N/A",
+    revision="no_revision",
```
Samoed: I think this was added in other pr too

AdnanElAssadi: I don't think so. I did the changes for the other models in their active PRs. These were already merged so created new one.

KennethEnevoldsen: isn't the revision: `30fcde30f19b87502b8435427b5f5068e401d5f6` and `53615c10408485422e09a12cda191a747f4bbe34`
should we instead write:

`30fcde30f19b87502b8435427b5f5068e401d5f6-53615c10408485422e09a12cda191a747f4bbe34` and just split the string on init?

Samoed: > I don't think so. I did the changes for the other models in their active PRs. These were already merged so created new one.

I confused this with clap


> isn't the revision: `30fcde30f19b87502b8435427b5f5068e401d5f6` and `53615c10408485422e09a12cda191a747f4bbe34` should we instead write:
> 
> `30fcde30f19b87502b8435427b5f5068e401d5f6-53615c10408485422e09a12cda191a747f4bbe34` and just split the string on init?

Agree, this is good solution

AdnanElAssadi:
Done. I've also fixed directory structure in the Results Repo.

PR #3869, opened 2026-01-06
model: Update bflhc/Octen-Embedding-8B revision

bflhc:


Samoed:
Can you describe what was changed?

bflhc:
Sure, the new revision of the model utilized more GPU resources for training. We also applied more refined data cleaning to the synthetic training datasets across domains.

KennethEnevoldsen:
Looks like a solid improvement @bflhc!

PR #3870, opened 2026-01-06
fix: Add --rebuild flag to bypass cache and force full rebuild

isaac-chung:
Part of #3798

##  What
Adds a `--rebuild` CLI flag to the MTEB leaderboard command that bypasses pre-computed JSON caches and forces a full rebuild from the results repository.

 ## Why
Users need a way to guarantee they're working with the freshest data from source when debugging cache issues or validating changes. The existing cache strategy prioritizes speed, but sometimes you need absolute certainty that you're seeing the latest state of individual model files.

##  How it works:

  - When `--rebuild` is passed, the system skips the quick JSON cache entirely
  - Directly clones/updates the full results repository
  - Builds BenchmarkResults from scratch by parsing individual model files
  - Overwrites the existing JSON cache with fresh data

##  Key changes:

  - Added load_from_cache() method to ResultCache class with rebuild parameter - this replaces the need for a `_load_results()` function in `app.py`
  - Added --rebuild CLI argument to leaderboard command
  - Simplified app.py by removing redundant _load_results() wrapper - now directly calls cache.load_from_cache()
  - Added comprehensive test coverage for all loading strategies and edge cases
  - Updated relevant docs sections

##  Usage
```
mteb leaderboard --rebuild
````

### CODE REVIEW

File: mteb/cache.py [hunk 1]
```diff
@@ -432,6 +432,107 @@ def clear_cache(self) -> None:
             logger.warning(msg)
             warnings.warn(msg)
 
+    def load_from_cache(
+        self,
+        quick_cache_path: Path | None = None,
```
Samoed: Maybe we can use `self.cache_path` as default path?

mteb/cache.py
```diff
@@ -432,6 +432,107 @@ def clear_cache(self) -> None:
...
+            for model_meta in mteb.get_model_metas()
+            if model_meta.name is not None
+        ]
+
+        all_results = self.load_results(
+            models=all_model_names,
+            only_main_score=True,
+            require_model_meta=False,
+            include_remote=True,
+        )
```
Samoed: Maybe these parameters can be passed as parameters to this function?

mteb/cache.py
isaac-chung: It cannot, as they are completely separate things. 
- `self.cache_path` points to the path where we run git clone [results repo]
- quick_cache_path is where we store the prebuilt  `__cached_results.json`. Before v2, it was in the `mteb/leaderboard` dir.

mteb/cache.py
isaac-chung: Not sure I follow. How will that look like, in a diff?

mteb/cache.py
Samoed: cache path have now 2 folders:
- results (with locally run tasks and models)
- remote (where done git clone)

I think we can create a third folder or just put it in root

isaac-chung: Oh I see. Then a 3rd folder is a pretty good option. I'll do that.

isaac-chung: Should we keep `quick_cache_path` as an input or not?

Samoed: I think we can change it to name of cached file to load from with default of `__cached_results.json`

mteb/cache.py
Samoed: We can add like 

```
def load_from_cache(
    models: Sequence[str] | Iterable[ModelMeta] | None = None,
    tasks: Sequence[str] | Iterable[AbsTask] | Benchmark | str | None = None,
    require_model_meta: bool = True,
    include_remote: bool = True,
    validate_and_filter: bool = False,
    only_main_score: bool = False,
)
```

But it seems that this is a bit complicated and would of parameters from `load_results` and this probably would work a bit unexpected to users when they are trying to load from cache, because we don't filter anything from it. 

I think other solution would be to remove this part from `load_from_cache`, because if we're fully rebuilding it we can just use `load_results` and here raise an error or return empty benchmark result if no cache exists

mteb/cache.py
isaac-chung: Done.

mteb/cache.py
isaac-chung: Hmm in the context of the LB, which is important, this method is to load all results, so adding all these params would really complicate things. Is this what you mean by "removing this part", would this work?

Option 1:
```diff
all_results = self.load_results(
-    models=all_model_names,
    only_main_score=True,
    require_model_meta=False,
    include_remote=True,
)
```

or option 2:
```diff
all_results = self.load_results(
-    models=all_model_names,
-    only_main_score=True,
-    require_model_meta=False,
-    include_remote=True,
)
```

Samoed: > Is this what you mean by "removing this part",

I mean we can remove this download step and fall back to it only if cache is not exists. 

It's can be a bit unclear for users why we choose these parameters for `load_results`

isaac-chung: It is already a fall back step, as this is within `_rebuild_from_full_repository()`, which is only called in `load_from_cache()` in strategy 3 only if strategy 1 and 2 fail. 

To me, it seems you lean towards option 2 then. Is that right?

Samoed: I meant that we can remove the third strategy from this function. I don’t think it should use `load_results`, because there are two possible cases:

1. We pass parameters to customize `load_results`. In that case, the first and second strategies wouldn’t work correctly, since loading from a JSON file can’t support this directly — it would only be possible to apply filters after loading from the cache.
2. We don’t pass parameters. Then, if the user wants to customize something and we fall back to the third option, we can’t support it anyway.

I think we should only handle loading from a file or a remote URL, and otherwise raise an error indicating that this isn’t supported.

isaac-chung: [Tagged you](https://github.com/embeddings-benchmark/mteb/pull/3870/changes#r2680249562) where the original call was. Here, the code loads all results. Removing the 3rd strategy seems reckless when you remove the fallback option.

Did you run the LB locally in this branch?

mteb/leaderboard/app.py
```diff
@@ -38,108 +35,6 @@
...
-
-            load_start = time.time()
-            all_model_names = [model_meta.name for model_meta in mteb.get_model_metas()]
-
-            all_results = cache.load_results(
-                models=all_model_names,
-                only_main_score=True,
-                require_model_meta=False,
-                include_remote=True,
-            )
```
isaac-chung, line 135: @Samoed here is the original call. We load all results.

Samoed: Yes, I know. Let's keep it with these parameters for now

isaac-chung:
Can we see what's missing still and wrap up this PR?

mteb/cache.py
```diff
@@ -438,6 +436,105 @@ def clear_cache(self) -> None:
             logger.warning(msg)
             warnings.warn(msg)
 
+    def load_from_cache(
```
KennethEnevoldsen: do we want this to be public?

KennethEnevoldsen (commented):
I would make the method private but otherwise I think it is good to merge

mteb/cache.py
isaac-chung: Making private.

Issue #3538, opened 2025-11-07
[MAEB] Migrate to `torchcodec`


Samoed:
### Description of the feature

Since 2.8.0 torchaudio in maintenance mode as alternative they suggested `torchcodec`. Datasets v4 already switched to them, but transformers requires torchaudio for now, but they've plans to add support in v5 version https://github.com/huggingface/transformers/issues/40986#issuecomment-3311494106 I've created issue about resampling with torchcodec https://github.com/meta-pytorch/torchcodec/issues/1029

Also we can't use it directly because of this issue https://github.com/huggingface/transformers/issues/42103

Samoed:
I tried running it with the current `main` branch of Transformers, and it seems the issue is resolved. We probably need to wait for Transformers v5 to be fully compatible with Datasets v4.

PR #3873, opened 2026-01-06
[MAEB] merge from main again

isaac-chung:
Update `maeb` branch with mteb 2.6.5

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest] #, macos-latest, windows-latest]
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"] #, "3.14"]
         include:
           # Add Windows with Python 3.10 only to avoid tests taking too long
           - os: windows-latest


Samoed:
`uv` tried to resolve dependencies for 23 minutes. Maybe we should roll back to pip to not resolve dependencies that we don't need?

isaac-chung:
Actually I'm curious whether a main merge would solve [this issue](https://github.com/embeddings-benchmark/mteb/pull/3845#issuecomment-3709756698).

Samoed:
No, It won't. It will be fixed only if maeb would use datasets v4

isaac-chung:
> No, It won't. It will be fixed only if maeb would use datasets v4

So, can it? Looks like https://github.com/huggingface/datasets/issues/7707 is solved now.

Samoed:
We're limited by transformers now https://github.com/embeddings-benchmark/mteb/issues/3538 or more specific https://github.com/huggingface/transformers/issues/42103

isaac-chung:
I see... maybe we don't merge main for now then. Is the fix now to invalidate the dependency cache in CI or what?

Samoed:
> maybe we don't merge main for now then

Same happening in all maeb related branches. (Except syncing for 20 mins, but you can update lock file for this)

> Is the fix now to invalidate the dependency cache in CI or what?

This is not dependency cache, but huggingface cache. You can delete all caches from CI (to delete huggingface cache you can delete only that ends on `-hf`)

image

isaac-chung:
You can't delete the cache, or what's the problem?

Samoed:
There is no big problem. I've deleted it, but each run on main will overwrite overwrite it

isaac-chung:
So what's the better fix?

Samoed:
You can remove HF cache from tests in maeb branch, but we would need to get back it before merge to main https://github.com/embeddings-benchmark/mteb/blob/1e7879300fd27e57973a01f31e171e2c29df4580/.github/workflows/test.yml#L52-L57

isaac-chung:
We can change the cache key to `${{ runner.os }}-hf-maeb` for now?

Samoed:
Yes, this would also work

isaac-chung:
@Samoed I was hoping to verify my last commit's CI runs. Would you mind holding off pushing commit to this branch please?

Samoed:
Yes, of course. I've just noticed that you tried to fix `random-encoder` and I fixed it a bit

isaac-chung:
> We're limited by transformers now #3538 or more specific [huggingface/transformers#42103](https://github.com/huggingface/transformers/issues/42103)

I think this point is running into the issue you mentioned earlier. We might have to just temporarily remove 3.14 in maeb if we want all tests to pass.

isaac-chung:
> > We're limited by transformers now #3538 or more specific [huggingface/transformers#42103](https://github.com/huggingface/transformers/issues/42103)
> 
> I think this point is running into the issue you mentioned earlier. We might have to just temporarily remove 3.14 in maeb if we want all tests to pass.

wdyt @Samoed ?

Samoed:
We can do this, but we need to not forget to get back it before merge

PR #3875, opened 2026-01-07
Simplify conflicts

Samoed:
Removed pair of conflict dependencies and combined them together. Now lock file takes only 5mb


KennethEnevoldsen:
great fix @Samoed!!

PR #3898, opened 2026-01-08
Remove maeb results folder (sym link)

Samoed:
Remove maeb results folder (sym link), because this cannot be used directly


isaac-chung (approved):
I guess this was meant to be a submodule?

Samoed:
Maybe, but I don't think we want to include this

Issue #3498, opened 2025-10-26
Calculate descriptive statistics for `MAEB`


Samoed:
### Description of the feature

Calculate descriptive statistics for `MAEB`. You can do this by 
```python
import mteb

task = mteb.get_task(...)
task.calculate_descriptive_statistics()
```

TODO:
- [x] ClothoA2TRetrieval
- [x] ClothoT2ARetrieval
- [x] CommonVoice17A2TRetrieval
- [x] CommonVoice17T2ARetrieval
- [x] CommonVoice21A2TRetrieval
- [x] CommonVoice21T2ARetrieval
- [x] VocalSound
- [x] SIBFLEURS
- [x] RavdessZeroshot
- [x] VoiceGenderClustering
- [x] UrbanSound8kZeroshot
- [x] FSD2019Kaggle
- [x] GoogleSVQT2ARetrieval
- [x] GoogleSVQA2TRetrieval
- [ ] AudioSet
- [ ] FleursA2TRetrieval
- [ ] FleursT2ARetrieval
- [ ] SoundDescsA2TRetrieval
- [ ] SoundDescsT2ARetrieval
- [ ] BirdSet (blocked by https://github.com/embeddings-benchmark/mteb/pull/3863, because not clear split selection)

ayush1298:
@Samoed What exactly `calculate_descriptive_statistics()` will show? Is it similar to how this is implemented:
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/abstasks/classification.py#L314

Samoed:
We need to update code for ot a bit (I'll add soon) also we need to think about what to do with https://github.com/embeddings-benchmark/mteb/issues/3471

ayush1298:
okay, will wait untill then.

Samoed:
@gowitheflow-1998 @AdnanElAssadi56 Can you help with calculating statics of left tasks? I can't do this, because tasks are too big. Fleurs can take a lot of time to load you can use fixes from https://github.com/embeddings-benchmark/mteb/pull/3971 (speedup loading) and https://github.com/embeddings-benchmark/mteb/pull/3961 (reupload tasks to remove trust remote code)

```python
import mteb

for task_name in [
    "AudioSet",   # reqires 2TB!!!!!!!! of space
    # after 3961 fleurs reqires 300gb (~4 gb per subset)
    "FleursA2TRetrieval",
    "FleursT2ARetrieval",
    # requires 100gb 
[...]

PR #3880, opened 2026-01-07
Use zxx-Zxxx language code for non-human audio tasks

isaac-chung:
Updates audio tasks that don't involve human speakers to use "zxx-Zxxx" (No linguistic content / Unwritten) instead of "eng-Latn":

- GTZAN (music classification, reranking, clustering)
- GunshotTriangulation
- MridinghamTonic and MridinghamStroke
- NSynth
- BirdSet and BirdCLEF
- BeijingOpera
- ESC50 (classification, pair classification, reranking, clustering, zeroshot)
- MusicCaps and MusicGenreClustering
- SoundDescs
- TUTAcousticScenes
- UrbanSound8K (classification, retrieval, reranking, zeroshot)
- VehicleSoundClustering
- AudioCaps (adds zxx-Zxxx alongside eng-Latn for audio-text tasks)

Fixes #3872

PR #3903, opened 2026-01-09
Compute statistics

Samoed:
- Changed category for classification datasets from `a2t` to `a2c`
- Added `trust_remote_code` to `RavdessZeroshot`, `FleursA2TRetrieval`, `FleursT2ARetrieval`
- Changed language to `zxx` of `AmbientAcousticContext`
- Changed revision of `AudioSet` to not require `trust_remote_code`, because I can't download it (was updated by author 3 months ago)
- Computed statistics for other tasks

ref https://github.com/embeddings-benchmark/mteb/issues/3498

Now we need to compute statistics for (added to #3498). They require more than 30gb of ram and/or very big:
- [x] ClothoA2TRetrieval
- [x] ClothoT2ARetrieval
- [x] CommonVoice17A2TRetrieval
- [x] CommonVoice17T2ARetrieval
- [x] CommonVoice21A2TRetrieval
- [x] CommonVoice21T2ARetrieval
- [x] VocalSound
- [x] SIBFLEURS
- [x] RavdessZeroshot
- [x] VoiceGenderClustering
- [x] UrbanSound8kZeroshot
- [x] FSD2019Kaggle
- [x] GoogleSVQT2ARetrieval
- [x] GoogleSVQA2TRetrieval
- [ ] AudioSet
- [ ] FleursA2TRetrieval
- [ ] FleursT2ARetrieval
- [ ] SoundDescsA2TRetrieval
- [ ] SoundDescsT2ARetrieval
- [ ] BirdSet (blocked by https://github.com/embeddings-benchmark/mteb/pull/3863, because not clear split selection)

### DIFF
--- a/mteb/descriptive_stats/AudioClassification/CSTRVCTKAccentID.json
+++ b/mteb/descriptive_stats/AudioClassification/CSTRVCTKAccentID.json
@@ -1,0 +1,66 @@
+{
+    "train": {
+        "num_samples": 12000,
+        "number_texts_intersect_with_train": null,
+        "text_statistics": null,
+        "image_statistics": null,
+        "audio_statistics": {
+            "total_duration_seconds": 40296.910437499966,
+            "min_duration_seconds": 1.3244166666666666,
+            "average_duration_seconds": 3.358075869791664,
+            "max_duration_seconds": 15.546083333333334,
+            "unique_audios": 12000,
+            "average_sampling_rate": 48000.0,
+            "sampling_rates": {
+                "48000": 12000
+            }
+        },
+        "label_statistics": {
+            "min_labels_per_text": 1,
+            "average_label_per_text": 1.0,
+            "max_labels_per_text": 1,
+            "unique_labels": 13,
+            "labels": {
+                "English": {
+                    "count": 3712
+                },
+                "Scottish": {
+                    "count": 2072
+                },
+                "NorthernIrish": {
+                    "count": 703
+                },
+                "Irish": {
+                    "count": 986
+                },
+                "Indian": {
+                    "count": 317
+                },
+                "Welsh": {
+                    "count": 102
+                },
+                "Unknown": {
+                    "count": 56
+                },
+                "American": {
+                    "count": 2288
+                },
+                "Canadian": {
+                    "count": 858
+                },
+                "SouthAfrican": {
+                    "count": 459
+                },
+                "Australian": {
+                    "count": 224
+                },
+                "NewZealand": {
+                    "count": 115
+                },
+                "British": {
+                    "count": 108
+                }
+            }
+        }
+    }
+}
--- a/mteb/descriptive_stats/AudioClassification/CSTRVCTKGender.json
+++ b/mteb/descriptive_stats/AudioClassification/CSTRVCTKGender.json
@@ -1,0 +1,33 @@
+{
+    "train": {
+        "num_samples": 12000,
+        "number_texts_intersect_with_train": null,
+        "text_statistics": null,
+        "image_statistics": null,
+        "audio_statistics": {
+            "total_duration_seconds": 40402.33970833344,
+            "min_duration_seconds": 1.3237083333333333,
+            "average_duration_seconds": 3.36686164236112,
+            "max_duration_seconds": 16.556875,
+            "unique_audios": 12000,
+            "average_sampling_rate": 48000.0,
+            "sampling_rates": {
+                "48000": 12000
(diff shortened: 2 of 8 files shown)


mteb/tasks/audio/audio_classification/zxx/ambient_acoustic_context.py
```diff
@@ -12,9 +12,9 @@ class AmbientAcousticContextClassification(AbsTaskClassification):
             "revision": "8f4de158d4162de768ebb4dc0594429d785077da",
         },
         type="AudioClassification",
-        category="a2t",
+        category="a2c",
         eval_splits=["test"],
-        eval_langs=["eng-Latn"],
+        eval_langs=["zxx-Zxxx"],
```
KennethEnevoldsen, line 17: Ahh great to use the zxx tag didn't know that this existed.

We could of course also just use an empty list `[]`?

mteb/descriptive_stats/AudioClassification/AmbientAcousticContext.json
```diff
@@ -0,0 +1,193 @@
...
+        "num_samples": 1036,
+        "number_texts_intersect_with_train": null,
+        "text_statistics": null,
+        "image_statistics": null,
+        "audio_statistics": {
+            "total_duration_seconds": 1045.632,
+            "min_duration_seconds": 0.896,
+            "average_duration_seconds": 1.0092972972972973,
+            "max_duration_seconds": 1.024,
+            "unique_audios": 1009,
```
KennethEnevoldsen, line 12: seems like we have some duplicates - is that something we want to handle or just create an issue on?

Samoed: I think it's better to create separate issue for now

KennethEnevoldsen: https://github.com/embeddings-benchmark/mteb/issues/3922

mteb/tasks/audio/audio_classification/zxx/ambient_acoustic_context.py
Samoed: We can, but then this would be unclear if this is missing annotation or unclassified language. Originally this was done in https://github.com/embeddings-benchmark/mteb/pull/3880

KennethEnevoldsen: Yeah I think this is indeed better.

Do we have cases for images where we should add this? If so let us add an issue on it

Samoed: I think we've changed in all places, but we need to check tasks again

Samoed:
@isaac-chung Do you have scripts for statistics table generation from mieb? I want to add something similar 
image

isaac-chung:
I use a script to generate the maeb task overview table. If you can push the stats I can add the columns in my scripts.

Samoed:
Statistics for almost all tasks are already here. I'll try to compute for the rest today

isaac-chung:
Oh don't forget to merge in the latest changes from `maeb`, e.g. fleurs was moved into the correct folder.

diffunity:
`calculate_descriptive_statistics()` currently puts `Any2AnyRetrieval` under `descriptive_stats/Image` folder. This seems to be due to [this line of code](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/abstasks/task_metadata.py#L399-L400).

Looking at this code also tells us that while all Image tasks go under `descriptive_stats/Image` folder, all audio tasks go under the parent folder `descriptive_stats/`

I think we should make `descriptive_stats/Audio` and figure out a way to differentiate audio from image tasks for `Any2AnyRetrieval`

Samoed:
I know, but I don't think this is a much problem. I think we can keep as is now. 

With this structure I tried to reflect our structure (it was like this before v2), but we haven't decided what to next. I think we can just move content `Image` stats folder

isaac-chung:
Are we not going to follow the same V2 structure as tasks?

Samoed:
For now not, but we can change this

isaac-chung:
It would be great to align fully with v2

Samoed:
I think we can move this later. I don't want to do this in this PR

KennethEnevoldsen:
Perfectly fine with me - I think we will need to do some general reviews before we merge into main so as long we have an issue on it then we will get to it

PR #3914, opened 2026-01-11
fix: model load test

Samoed:
Currently, it’s not possible to install all dependencies by groups, because after https://github.com/embeddings-benchmark/mteb/pull/3875, `pylate` and `model2vec` are in the same group and their dependencies cannot be resolved together.


isaac-chung:
Ah - model load test was skipped in #3875. We should trigger that test if the dependencies changed as well.

isaac-chung (approved):

Issue #3908, opened 2026-01-11
Cannot get result for SynPerChatbotConvSAClassification


Geralt-Targaryen:
### Describe the bug

When evaluating a model on the task SynPerChatbotConvSAClassification, the following result files are created:
```
['SynPerChatbotConvSAAnger.json', 'SynPerChatbotConvSAFear.json', 'SynPerChatbotConvSAFriendship.json', 'SynPerChatbotConvSAHappiness.json', 'SynPerChatbotConvSAJealousy.json', 'SynPerChatbotConvSALove.json', 'SynPerChatbotConvSASadness.json', 'SynPerChatbotConvSASatisfaction.json', 'SynPerChatbotConvSASurprise.json']
```
But no file named SynPerChatbotConvSAClassification.json is created.

And when I load the results for benchmark MTEB(fas, v2), I can't get a score for this task.

### To reproduce

```
benchmark = mteb.get_benchmark('MTEB(fas, v2)')
cache = ResultCache(result_path)
results = cache.load_results(models=["no_model_name__available"], tasks=benchmark)
```

### Are you interested to contribute a fix for this bug?

No

Samoed:
`SynPerChatbotConvSAClassification` is aggregated tasks that consist of other different task. We don't create json for this task directly. 

I can't reproduce your problem, because when I'm loading with this snippet I get results without problems
```python
import mteb

cache = mteb.cache.ResultCache()
task = mteb.get_task("SynPerChatbotConvSAClassification")
res = cache.load_results(tasks=[task])
print(res[0][0])
# TaskResult(task_name=SynPerChatbotConvSAClassification, scores=...)
print(res[0][0].get_score())
# np.float64(0.52419)
print(res[0][0].task_name)
# 'SynPerChatbotConvSAClassification'

benchmark = mteb.get_benchmark('MTEB(fas, v2)')
res = cache.load_results(tasks=benchmark)
print(res.to_dataframe().head())
# model_name                                     task_name  ...  zeta-alpha-ai/Zeta-Alpha-E5-Mistral
# 0                                          ArguAna-Fa.v2  ...                                  NaN
# ... 
# 35                     SynPerChatbotConvSAClassification  ...                                  NaN
# ...
df = res.to_dataframe()
df = df.set_index("task_name")
print(df.loc["SynPerChatbotConvSAClassification", "google/embeddinggemma-300m"])
# np.float64(0.697946)
```

Geralt-Targaryen:
[...]

PR #3915, opened 2026-01-11
fix: saving aggregated tasks

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3908


Samoed:
Agree, I think this would be nice to do. Do you want to merge this pr or I should implement your suggestion?

KennethEnevoldsen:
I think that solution is better (as we do not need to recompute the file) - I am fine with merging this in as a temporary fix though

PR #3930, opened 2026-01-14
Update references and citations for ViDoRe V3 benchmark

antoineedy:
Preprint of ViDoRe V3 has been published to arXiv, so we would like to update the citations :)


antoineedy:
thanks @Samoed 🤗

PR #3932, opened 2026-01-14
fix: temporarily remove private column from RTEB

### LINKED ISSUE #3902: RTEB private track is effectively private only to non-Voyage participants
I am opening this issue to raise serious concerns about the current state of the RTEB private track and its impact on fairness and credibility.

1. The current “private” datasets are not truly private

At present, the so-called private datasets are effectively private only to participants other than Voyage.
Voyage already has access to these datasets, while all other participants do not.
https://github.com/embeddings-benchmark/results/pull/391

2. This is not an accusation of misconduct, but a structural fairness problem

I want to be explicit that this is not an accusation of academic misconduct or bad faith on Voyage’s part.
However, from a technical and methodological perspective, the possibility of implicit optimization against private benchmarks absolutely exists when a participant has access to the evaluation tasks.

Empirically, we can already observe that the primary advantage of the Voyage model series comes from these “private” datasets, rather than from public benchmarks alone.
Image
This creates a structural advantage that no other participant can reasonably match.

3. There is clear historical precedent for this type of issue

This is not a hypothetical concern. Similar situations have occurred in the past in other benchmarks and competitions, where access to evaluation data — even without explicit misuse — led to disqualification or intervention by organizers to preserve fairness.
https://www.bbc.com/news/technology-33005728

4. Proposed actions for the community

Given the above, I strongly suggest the following:
  - Temporarily suspend the private track until truly private datasets — inaccessible to all participants — can be prepared and enforced.
  - Until then, retain only the public retrieval/search tasks for evaluation and comparison.
  - Ideally, no single organization outside the community should act as both participant and evaluator.
  - If full separation is not feasible, then a committee composed of multiple independent stakeholders should be formed to oversee private evaluations and governance.

Without such safeguards, RTEB risks becoming effectively a single-organization showcase rather than a trusted community benchmark, which removes any real incentive for independent developers to continue participating.

5. Call to the community

Until the fairness issues around the private track are properly addressed, we urge all developers and contributors to carefully consider whether participation in the private track is appropriate.

Benchmark credibility depends not only on results, but on the trust that those results are produced under equal and transparent conditions.

KennethEnevoldsen:
fixes #3902

here is the current leaderboard:

Screenshot 2026-01-14 at 20 50 17

Link to decision: #3934

### DIFF
--- a/mteb/benchmarks/_create_table.py
+++ b/mteb/benchmarks/_create_table.py
@@ -303,6 +303,7 @@ def _create_per_language_table_from_benchmark_results(
 
 def _create_summary_table_mean_public_private(
     benchmark_results: BenchmarkResults,
+    exclude_private_from_borda: bool = False,
 ) -> pd.DataFrame:
     """Create summary table from BenchmarkResults.
 
@@ -311,6 +312,7 @@ def _create_summary_table_mean_public_private(
 
     Args:
         benchmark_results: BenchmarkResults object containing model results
+        exclude_private_from_borda: If True, calculate Borda rank using only public tasks
 
     Returns:
         DataFrame with model summaries, ready for styling in the leaderboard
@@ -356,7 +358,11 @@ def _create_summary_table_mean_public_private(
     joint_table = joint_table.drop(models_to_remove, axis=0)
     joint_table.insert(0, "mean(public)", public_mean)
     joint_table.insert(1, "mean(private)", private_mean)
-    joint_table["borda_rank"] = _get_borda_rank(per_task)
+    if exclude_private_from_borda:
+        borda_per_task = per_task[public_task_name]
+    else:
+        borda_per_task = per_task
+    joint_table["borda_rank"] = _get_borda_rank(borda_per_task)
     joint_table = joint_table.sort_values("borda_rank", ascending=True)
     joint_table = joint_table.reset_index()
 
--- a/mteb/benchmarks/benchmark.py
+++ b/mteb/benchmarks/benchmark.py
@@ -123,7 +123,9 @@ class RtebBenchmark(Benchmark):
             _create_summary_table_mean_public_private,
         )
 
-        joint_table = _create_summary_table_mean_public_private(benchmark_results)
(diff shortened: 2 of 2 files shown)


bflhc:
Hi, I noticed that the current implementation only removes the "Mean (Private)" column but doesn't exclude private tasks from the Borda rank calculation. I've added a fix that introduces an `exclude_private_from_borda` parameter to `_create_summary_table_mean_public_private`, so the Rank(Borda) is now calculated using only public tasks for RTEB.

You can find my changes here: https://github.com/bflhc/mteb/tree/remove-private

KennethEnevoldsen:
Good catch! - I have merged it in

bflhc:
I think the current implementation is solid now. Are we good to merge this PR? @Samoed @KennethEnevoldsen

bflhc:
The changes don’t seem to be reflected on the web UI yet. I also noticed that the latest run of the workflow
`leaderboard_docker.yml` on https://github.com/embeddings-benchmark/mteb/actions/workflows/leaderboard_docker.yml
 has issues related to Docker.

Could you please help take a look and investigate what’s going wrong there? @KennethEnevoldsen

Samoed:
This workflow fail is not fully related that leaderboard not updating. Currently, leaderboard pinned to specific revision until fix https://github.com/embeddings-benchmark/mteb/issues/3959

I saw that this workflow failed earlier, but I don't have ideas why this happened

KennethEnevoldsen:
Agree with @Samoed here. https://github.com/embeddings-benchmark/mteb/issues/3959 should be close to being solved (just ok'ed the PR)

@Samoed can I ask you to merge it in and then update the leaderboard revision at a time where we can ensure that everything works as intended (I will be out of office for the full day tomorrow)

As for the Docker crash, it doesn't seem like the script failed without any logs (this sometimes happens with GitHub actions). If we see this again we will def. look into it, but it seems like future actions work without issue.

Samoed:
@KennethEnevoldsen @bflhc Fixed leaderboard version was deployed

PR #3941, opened 2026-01-15
Adding nvidia/nemotron-colembed models

rnyak:

https://huggingface.co/nvidia/llama-nemotron-colembed-vl-3b-v2
https://huggingface.co/nvidia/nemotron-colembed-vl-4b-v2
https://huggingface.co/nvidia/nemotron-colembed-vl-8b-v2


mteb/models/model_implementations/nvidia_llama_nemoretriever_colemb.py
```diff
@@ -199,3 +199,30 @@ def encode(
     training_datasets=TRAINING_DATA,
     citation=LLAMA_NEMORETRIEVER_CITATION,
 )
+
+llama_nemotron_colembed_3b_v2 = ModelMeta(
+    loader=LlamaNemoretrieverColembed,
+    loader_kwargs=dict(
+        trust_remote_code=True,
+    ),
+    name="nvidia/llama-nemotron-colembed-3b-v2",
```
Samoed: Are you from nvidia? I can't find this model on HF

rnyak: @Samoed hello. thanks for review, yes I am from NV. we will make the repo public shortly. I will also add two more model meta here.

mteb/models/model_implementations/nvidia_llama_nemoretriever_colemb.py
```diff
@@ -226,3 +244,59 @@ def encode(
     training_datasets=TRAINING_DATA,
     citation=LLAMA_NEMORETRIEVER_CITATION,
 )
+
+nemotron_colembed_vl_4b_v2 = ModelMeta(
+    loader=LlamaNemoretrieverColembed,
+    loader_kwargs=dict(
+        trust_remote_code=True,
+        transformers_version_constraint="==5.0.0rc0",
```
Samoed, line 258: I think it's better to change like this to be able to run on release version or other rc versions
```suggestion
        transformers_version_constraint=">=5.0.0rc0",
```

rnyak: @Samoed thanks for the suggestion, yes we can do that, but we did not do tests with other `rc` versions. may be safer to keep this a fixed version?

Samoed: I understand that, but later it would be better to have not rc version, but I think we can keep like this

Samoed: Can you try to run on released version? https://github.com/huggingface/transformers/releases/tag/v5.0.0

rnyak:
@Samoed hello, the three models are public on HF now,  can we merge this PR? thanks.

mteb/models/model_implementations/nvidia_llama_nemoretriever_colemb.py
```diff
@@ -205,3 +222,86 @@ def encode(
...
+    revision="75f03c712cb3a252e062295f9a0966e5d95d6156",
+    release_date="2026-01-21",
+    modalities=["image", "text"],
+    n_parameters=4_407_000_000,
+    memory_usage_mb=8403,
+    max_tokens=8192,
+    embed_dim=3072,
+    license="https://huggingface.co/nvidia/llama-nemotron-colembed-vl-3b-v2/blob/main/LICENSE",
+    open_weights=True,
+    public_training_code="Proprietary Code",
```
Samoed:
```suggestion
    public_training_code=None,
```

rnyak: @Samoed hello, may I ask why did you change this as "None"? thanks.

Samoed: This field is expected to be a URL to the code, and `None` means that it’s private.
https://github.com/embeddings-benchmark/mteb/blob/7c9bab2f8ed4f3992c681098c232524eb6043621/mteb/models/model_meta.py#L116

mteb/models/model_implementations/nvidia_llama_nemoretriever_colemb.py
rnyak: @Samoed hello. thanks for your suggestion. I did a quick check and 5.0.0 does not work out of the box, it'd need a fix, can we merge this MR with `transformers_version_constraint="==5.0.0rc0"` for now.

boliu61:
Hi @Samoed, thanks for the review! We understand that the stable 5.0.0 is preferred long term. However, we tested pinning to 5.0.0 (just released yesterday) and it unfortunately introduces breaking changes that will require some code fixes and tests to resolve. Since 5.0.0rc0 is fully verified and working for these models, could we proceed with merging this PR as-is?

We're happy to do a quick follow-up PR after this to support 5.0.0. This way, we can get the models out to users without blocking on the version upgrade debugging. Thanks for your consideration.

Samoed:
Sadly, you ran into an error with the new transformer. I agree, I’ll merge this.

Issue #3926, opened 2026-01-13
`mteb.evaluate` runs on all languages and splits


meetdoshi90:
### Describe the bug

mteb.evaluate runs evaluation on all languages, even when specific languages or splits are specified

> TASK_NAME='MIRACLRetrievalHardNegatives'
> mteb.get_tasks([TASK_NAME], languages=['eng'])
-> MTEBTasks(MultiLongDocRetrieval(name='MultiLongDocRetrieval', languages=['eng']),)
> mteb.get_tasks([TASK_NAME], languages=['eng'])[0].metadata.eval_langs
-> {'ar': ['ara-Arab'], 'bn': ['ben-Beng'], 'de': ['deu-Latn'], 'en': ['eng-Latn'], 'es': ['spa-Latn'], 'fa': ['fas-Arab'], 'fi': ['fin-Latn'], 'fr': ['fra-Latn'], 'hi': ['hin-Deva'], 'id': ['ind-Latn'], 'ja': ['jpn-Jpan'], 'ko': ['kor-Kore'], 'ru': ['rus-Cyrl'], 'sw': ['swa-Latn'], 'te': ['tel-Telu'], 'th': ['tha-Thai'], 'yo': ['yor-Latn'], 'zh': ['zho-Hans']}`

When the above snippet is passed to mteb.evaluate it runs evaluation on all languages and I do not see any argument that can be passed to evaluate function to run only certain language or data splits.

The old way works fine, but it's now deprecated

>  tasks = mteb.get_tasks(tasks=["MultiLongDocRetrieval"], eval_splits=['test'], languages=[args.lang])
    evaluation = mteb.MTEB(tasks=tasks)
    results = evaluation.run(model, output_folder=args.output_dir, batch_size=16)

I even tried the approach specified in https://github.com/embeddings-benchmark/mteb/issues/119 but doesnt seem to work.

> from mteb.tasks import MultiLongDocRetrieval, MIRACLRetrievalHardNegatives
MultiLongDocRetrieval(languages=['eng'])
-> MultiLongDocRetrieval(name='MultiLongDocRetrieval', languages=['ara', 'cmn', 'deu', '...'])
MultiLongDocRetrieval(languages=['en'])
-> MultiLongDocRetrieval(name='MultiLongDocRetrieval', languages=['ara', 'cmn', 'deu', '...'])
MultiLongDocRetrieval(langs=['en'])
[...]

PR #3946, opened 2026-01-16
fix: Ensure that retrieval tasks only evaluate on specified subsets instead of all

Samoed:
Previously all defined languages and subsets would be run. Now only that passed thought `mteb.get_task`. Ref https://github.com/embeddings-benchmark/mteb/issues/3926

### DIFF
--- a/tests/test_tasks/test_load_data.py
+++ b/tests/test_tasks/test_load_data.py
@@ -1,0 +1,52 @@
+from unittest.mock import patch
+
+import pytest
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen:
Ohh feels like this error keeps popping up (I think it is because someone thinks it is incorrect) - should we add a test on this as well so ensure that it isn't changed in the future

Samoed:
Added test and verified that without this fix retrieval loading test would fail

PR #3950, opened 2026-01-16
feat: Evaluate Compression Methods on Retrieval Tasks

### LINKED ISSUE #3949: Evaluate Compression Methods on RTEB

As discussed with @KennethEnevoldsen, @isaac-chung and @Samoed in December, my colleague @michaeldinzinger and I propose to add an evaluation of simple compression methods like quantization to RTEB based on our recent work[1]. To start the discussion, I've implemented a first (partial) draft to add quantization (see PR #3950). Based on your feedback, we can extend and update the current implementation.

**Points of discussion**

From our perspective, there are a few things that still need to be implemented/discussed:

**Compression with FAISS:** *(missing)* So far, quantization is only supported when running retrieval tasks without an `index_backend`. FAISS has its own support for binary and scalar indices, so it's possible to add quantization support in a similar manner. However, retrieval performance will differ between FAISS quantization and the other quantization approach.

**Displaying results on leaderboard:** *(missing)* To avoid cluttering the leaderboard, we propose to add an additional selection menu (using radio buttons, dropdowns or similar) per quantization level, so users can select which results should be displayed.

**Check existing results:** *(missing)* With multiple quantization levels, the check for existing results in the `evluate` method needs to consider if results for quantized embeddings are present and compute them if not.

**Compressing embeddings through truncation:** *(optional, missing)* Another easy way to compress embeddings is truncating the vectors, which is facilitated by MRL training. We could extend the code to also support this, as it is very easy to implement. However, choosing which dimensions to truncate to might be trickier due to differences in embedding dimensions and the chosen MRL truncation points.

**Result storage:** The current implementation stores retrieval performance with quantization as part of the regular result JSON file. I think it would be cleaner to split this into one result file per quantization level, named accordingly.

**Handling model-internal compression:** There are a few models like some from Cohere and VoyageAI which support different output data types by passing a corresponding parameter at inference. In such cases, the regular quantization pipeline should ignore dtypes the model natively supports as currently implemented. However, this means repeated evaluation for different dtypes.

What are your thoughts on this? We're happy to discuss any issues you see with the code and update it accordingly. As adding this means re-computing results for a lot of models, we can also offer help with this.

[1] [https://arxiv.org/abs/2510.19340](https://arxiv.org/abs/2510.19340) (accepted at ECIR'26)

casparil:
This PR adds the option to evaluate MTEB tasks on compressed embeddings (see Issue #3949).

**What the code does**

- Introduces a new (command-line) parameter that computes quantization performance at different levels (float8, int8, int4 and binary) when set accordingly.
- The model is then wrapped in a `CompressionWrapper` class that handles compression.
- Computes embeddings as normal, then compresses embeddings per batch and computes results as normal.
- Store the result JSON file in a folder named after model and compression level.

Closes #3949.

### DIFF
--- a/tests/test_models/test_compression_wrapper.py
+++ b/tests/test_models/test_compression_wrapper.py
@@ -1,4 +1,3 @@
-
 import pytest
 import torch
 from datasets import Dataset


Samoed:
I think it would be better to add support for any model, not just for retrieval by creating wrapper for encoder model like Cache wrapper https://github.com/embeddings-benchmark/mteb/blob/4c4ab2b5bf67c693d10609726c220e1ff531d33a/mteb/models/cache_wrappers/cache_wrapper.py#L23 rather than create specific implementation for retrieval models

KennethEnevoldsen:
thanks for the outline! It always great to something to work from.

If we want to have these displayed as different models I am probably leaning a slightly different direction (moving the compression logic to the model)

```py
int8_mdl = CompressionWrapper(model, dtype="int8") 
# also manipulates task metadata to such that the name is "{name} (dtype=int8)"

res = mteb.evaluate(int8_mdl, task)
```

This is of course inefficient, which is why I would probably use:
```py
cached_mdl = CachedEmbeddingWrapper(model)
int8_mdl = CompressionWrapper(cached_mdl, dtype="int8") 
# also manipulates task metadata to such that the name is "{name} (dtype=int8)"

res = mteb.evaluate(int8_mdl, task)
```
Which would allow for fast iteration over different compression levels without any changes to the `core` evaluation loop. This would also automatically make this approach applicable to any task within MTEB.

Samoed:
> This is of course inefficient, which is why I would probably use:

I've mentioned `CachedEmbeddingWrapper` as example of wrapper which is better approach of integration. I don't think that this is required to use them both

> If we want to have these displayed as different models I am probably leaning a slightly different direction (moving the compression logic to the model)

I don't think that we should move this logic to the model. I think this can be solved by https://github.com/embeddings-benchmark/mteb/issues/1211

KennethEnevoldsen:
> I don't think that we should move this logic to the model. I think this can be solved by https://github.com/embeddings-benchmark/mteb/issues/1211

but then we have to implement compression metrics for all tasks? (maybe that is easier, but out results are already quite big)

the issue would also be solved by my suggestion though (though I am not sure it is the best approach)

Samoed:
> but then we have to implement compression metrics for all tasks? (maybe that is easier, but out results are already quite big)

I don't think we need to implement additional metrics for them. We can measure the same metrics, but on a quantized embedding.

Your approach is similar to #1211 overall. Maybe I misunderstood part of `moving the compression logic to the model`, because seems from your comment this is just a wrapper around our implementations and I don't think we need to create separate instances of models for this

KennethEnevoldsen:
Yea, the compression would just happen in the wrapper (so no need to create a new wrapper), but for models that require it, we could create their own custom wrappers.

Samoed:
Yes, agree

KennethEnevoldsen:
Potentially raise a warning for cases like:

```
mdl = mteb.get_model("voyageai/voyage-3.5") 
int8_mdl = CompressionWrapper(mdl, dtype="int8")
# Warning: The model `voyageai/voyage-3.5 (output_dtype=int8)` already exists. Model name of will be set to `voyageai/voyage-3.5 (output_dtype=int8*)` to avoid conflicts.
```

casparil:
Thanks to both of you for the feedback!

Using a wrapper class as you suggested sounds like a better approach. We'll update the code accordingly and try to integrate your comments.

mteb/abstasks/retrieval.py
```diff
@@ -361,16 +375,14 @@ def _evaluate_subset(
...
 
-        if isinstance(model, EncoderProtocol) and not isinstance(model, SearchProtocol):
-            search_model = SearchEncoderWrapper(model, quantize=self._quantize)
-        elif isinstance(model, CrossEncoderProtocol):
-            search_model = SearchCrossEncoderWrapper(model)
-        elif isinstance(model, SearchProtocol):
-            search_model = model
+        from mteb.models import CachedEmbeddingWrapper, CompressionWrapper
+
+        if isinstance(model, CompressionWrapper):
```
Samoed: You don't need to add `CompressionWrapper` to the retrieval evaluator. You can wrap model and then pass it to mteb without changing evaluation code

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+            "int8",
+            "int4",
+        ]:
+            # Otherwise, compute thresholds for int8/int4 quantization on documents first, then apply them on queries
+            logger.info("Query embeddings will be quantized on similarity calculation.")
+            self.query_embeds = embeddings
+            return embeddings
+        else:
+            logger.info(f"Quantizing embeddings to {self._quantization_level}")
+            return self._quantize_embeddings(embeddings, prompt_type)
```
Samoed: Why not always quantize embeddings?

casparil: In a lot of datasets, the number of queries is relatively small, while the number of documents is much larger. For integer quantization, we need to estimate the thresholds that decide which range of floating points is mapped to which integer. Applying this to a relatively small number of embeddings might lead to a bad estimation, so we first compute those thresholds on the larger number of documents, then apply the thresholds to the queries. This also ensures that both queries and documents are quantized using the same thresholds.

mteb/abstasks/retrieval.py
casparil: The model is wrapped in the `CompressionWrapper` class before `mteb.evaluate()` is called, so it now works with other task types as well. There's just this specific code snippet during retrieval evaluation that performs the following checks:

```
 if isinstance(model, EncoderProtocol) and not isinstance(model, SearchProtocol):
        return SearchEncoderWrapper(model)
elif isinstance(model, CrossEncoderProtocol):
        return SearchCrossEncoderWrapper(model)
 elif isinstance(model, SearchProtocol):
        return model
else:
        raise TypeError(
            f"RetrievalEvaluator expects a SearchInterface, Encoder, or CrossEncoder, got {type(model)}"
        )
```

As the model is wrapped in the `CompressionWrapper` class, this will raise the error, so I've adapted the code accordingly. If you prefer to handle this differently, I'm open for suggestions.

casparil:
@Samoed any update on my comments? From my perspective, the code is ready to be merged.

mteb/cli/build_cli.py
```diff
@@ -80,6 +81,13 @@ def run(args: argparse.Namespace) -> None:
         )
         overwrite_strategy = OverwriteStrategy.ALWAYS.value
 
+    quantization_level = args.quantization
```
Samoed: Can you remove this? I don't think we should add quantization to cli

mteb/models/compression_wrappers/compression_wrapper.py
Samoed: The problem is that we don't guarantee that queries will be before documents. If users used `search_backend` then order would change. 


You're using `self.query_embeds` only in similarity function and I don't see that it's using for estimations

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+            embeddings: The embeddings to quantize.
+            prompt_type: The name type of prompt. (query or passage)
+
+        Returns:
+            The quantized embeddings.
+        """
+        quantiles = torch.tensor([0.025, 0.975])
+        if self._quantization_level == "float8":
+            # Cast to float8, then back to float16 using PyTorch as numpy doesn't support float8
+            quantized = embeddings.type(torch.float8_e4m3fn).type(torch.float16)
```
Samoed: Why use e4m3? Maybe add customization?

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+        elif self._quantization_level == "int8" or self._quantization_level == "int4":
```
Samoed:
```suggestion
        elif self._quantization_level in ("int8", "int4"):
```

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+        self.min_embeds = 10_000
+        self.query_embeds = None
+        self.hf_subset = None
+        self.task_name = None
+        embed_types = None
+        if model.mteb_model_meta:
+            embed_types = model.mteb_model_meta.embedding_types
+            model.mteb_model_meta.name = (
+                f"{model.mteb_model_meta.name} (output_dtype={quantization_level})"
+            )
```
Samoed: You shouldn't change name of the model. Previously we used this as temporary solution

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+        of) documents need to be embedded first.
+
+        Args:
+            embeddings: The embeddings to quantize.
+            prompt_type: The name type of prompt. (query or passage)
+
+        Returns:
+            The quantized embeddings.
+        """
+        quantiles = torch.tensor([0.025, 0.975])
```
Samoed: Why do you use these quantiles?

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+logger = logging.getLogger(__name__)
+
+
+class CompressionWrapper:
+    """Wraps a model to quantize the embeddings and compute results on the compressed vectors instead."""
+
+    def __init__(
+        self,
+        model: EncoderProtocol,
+        quantization_level: str,
```
Samoed: Can you add enum for levels?

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+            if not isinstance(embeddings1, torch.Tensor):
+                embeddings1 = torch.tensor(embeddings1)
+            if not isinstance(embeddings2, torch.Tensor):
+                embeddings2 = torch.tensor(embeddings2)
+            logger.info("Quantizing query embeddings.")
+            if torch.equal(embeddings1, self.query_embeds):
+                embeddings1 = self._quantize_embeddings(embeddings1, PromptType.query)
+            elif torch.equal(embeddings2, self.query_embeds):
+                embeddings2 = self._quantize_embeddings(embeddings2, PromptType.query)
+        self._reset_boundaries()
```
Samoed: I don't think that you need to quantize embeddings again here
```suggestion
```

mteb/models/compression_wrappers/compression_wrapper.py
casparil: I saw that depending on the task, queries are embedded first and would thus be compressed before documents. So I adapted the code to simply return the uncompressed query embeddings if we need to compute any embedding-dependent thresholds and only quantize them when computing similarity, at which point thresholds should have been computed on document embeddings. Looking at the code again, I think there's a better way to do this, though, without using `self.query_embeds`.

mteb/models/compression_wrappers/compression_wrapper.py
casparil: We chose it as this type performed best in our experiments, but I agree it is better to make it configurable.

mteb/models/compression_wrappers/compression_wrapper.py
casparil: Ok, I will change it back. Do you have any suggestions on how to handle naming the result files or folder structure when using compression such that they do not overwrite the regular results? Or would see that as something people would need to handle themselves outside of the MTEB code?

mteb/models/compression_wrappers/compression_wrapper.py
casparil: When computing thresholds for quantizing to 8/4 bits, we determine them by taking the minimum and maximum value per embedding dimension, then dividing the space in between into 2^8/2^4 equidistant bins. For example, if we wanted to quantize to 2 bits, we would effectively calculate the 25th, 50th and 75th percentile values, then map numbers up to the 25th percentile to 0, between the 25th and 50th to 1, etc.

The problem with this approach is that outliers can heavily impact the thresholds, i.e. if the maximum value of one dimension is much higher than the rest, that bin might contain very few values. To avoid this, we can crop the top/bottom x percent of values. In our experiments, we found that this notably increased performance. However, this should be a configurable option, so I will change it.

mteb/models/compression_wrappers/compression_wrapper.py
casparil: Sure, will do

mteb/models/compression_wrappers/compression_wrapper.py
Samoed: I think it will fit in experiments https://github.com/embeddings-benchmark/mteb/pull/4071

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+logger = logging.getLogger(__name__)
+
+
+class QuantizationLevel(Enum):
+    """Enum for valid compression levels."""
+
+    FLOAT8 = 1,
+    INT8 = 2,
+    INT4 = 3,
+    BINARY = 4,
```
Samoed:
```suggestion
class QuantizationLevel(HelpfulStrEnum):
    """Enum for valid compression levels."""

    FLOAT8 = "float8"
    INT8 = "int8"
    INT4 = "int4"
    BINARY = "binary"
```

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+
+class CompressionWrapper:
+    """Wraps a model to quantize the embeddings and compute results on the compressed vectors instead."""
+
+    def __init__(
+        self,
+        model: EncoderProtocol,
+        quantization_level: QuantizationLevel = QuantizationLevel.FLOAT8,
+        quantiles: tuple[float, float] | None = None,
+        float_type: torch.dtype = torch.float8_e4m3fn,
```
Samoed: Probably this can be moved to `QuantizationLevel`

mteb/models/compression_wrappers/compression_wrapper.py
Samoed: Or you can change to `FLOAT8 = torch.float8`, `FLOATE4M3FN = torch.float8_e4m3fn`

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+        self,
+        embeddings1: Array,
+        embeddings2: Array,
+    ) -> Array:
+        """Refer to [EncoderProtocol.similarity][mteb.models.EncoderProtocol.similarity_pairwise] for more details."""
+        if self.quantize_queries:
+            embeddings1 = self._quantize_queries(embeddings1)
+            embeddings2 = self._quantize_queries(embeddings2)
+        self._reset_boundaries()
+        return self.model.similarity_pairwise(embeddings1, embeddings2)
```
Samoed, line 232: These functions would receive only already quantized embeddings  
```suggestion
    def similarity(
        self,
        embeddings1: Array,
        embeddings2: Array,
    ) -> Array:
        """Refer to [EncoderProtocol.similarity][mteb.models.EncoderProtocol.similarity] for more details."""
        return self.model.similarity(embeddings1, embeddings2)

    def similarity_pairwise(
        self,
        embeddings1: Array,
        embeddings2: Array,
    ) -> Array:
        """Refer to [EncoderProtocol.similarity][mteb.models.EncoderProtocol.similarity_pairwise] for more details."""
        return self.model.similarity_pairwise(embeddings1, embeddings2)
```

mteb/models/model_meta.py
```diff
@@ -163,6 +164,7 @@ class ModelMeta(BaseModel):
     model_type: list[MODEL_TYPES] = ["dense"]
     citation: str | None = None
     contacts: list[str] | None = None
+    embedding_types: list[str] | None = None
```
```suggestion
    embedding_types: list[EmbeddingType] | None = None
```

mteb/models/compression_wrappers/compression_wrapper.py
Samoed: Also, maybe would be better to unify with `EmbeddingType`

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+        if prompt_type == PromptType.query and task_metadata.category in [
+            "t2i",
+            "i2t",
+            "it2i",
+            "i2it",
+        ]:
+            # With multimodal tasks, always quantize text and image embeddings separately
+            logger.info(f"Quantizing query embeddings to {self._quantization_level}")
+            embeddings = self._quantize_embeddings(embeddings)
+            self._reset_boundaries()
```
Samoed: Do you need to store min/max? You reset them in the end of each function

mteb/models/compression_wrappers/compression_wrapper.py
casparil: Not necessarily. If the quantization level is set to int8/int4, queries will not be quantized in the `encode` function. Subsequently, when document embeddings are calculated, they are quantized to int8/int4 and used to set `self.mins` and `self.maxs`. Then, when `similarity` is called with the quantized document embeddings and the float query embeddings, the queries are quantized to int8/int4 using the pre-computed `self.mins`, `self.maxs` before resetting them.

mteb/models/compression_wrappers/compression_wrapper.py
casparil: They are stored to apply the min/max values calculated on the document embeddings to the query embeddings.

Samoed: For big retrieval tasks there can be a problem. For example, if task have more than 100_000 examples in corpus, then:
1. Query encode (min/max set)
2. Document batch 1 encode 
3. Similarity between query and documents (min/max reset)
4. Document batch 2 encode (different min/max)

Also for classification tasks this can cause problems, because encode called multiple times

mteb/models/compression_wrappers/compression_wrapper.py
Samoed: This seems a bit unexpected. Can we quantize queries on first run?

mteb/models/model_meta.py
casparil: The `embedding_types` need to be defined per model and I don't think there's a standard way to do this yet. Only `cohere_models.py` specifically defines a list of valid compression levels. In `voyage_models.py`, there's a dictionary containing `float32` and `bf16`, but the model also supports compression to binary and int8. Would you try to make this consistent across the `model_implementations` files?

mteb/models/compression_wrappers/compression_wrapper.py
casparil: True, the quantized embeddings need to be assigned to a new variable.

mteb/models/compression_wrappers/compression_wrapper.py
casparil: Of course quantizing document and query embeddings separately would be easier. However, there are a few caveats here. From a practical perspective, when i.e. setting up a RAG system you have a set of documents you want to embed, but usually don't have a representative set of user queries. So you would compute thresholds based on the document embeddings and use them to quantize incoming queries. Other than that, most datasets contain rather few queries compared to the documents. Estimating thresholds on, i.e., only 50 queries might be very unstable.

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -13,13 +13,13 @@
 logger = logging.getLogger(__name__)
 
 
-class QuantizationLevel(Enum):
+class QuantizationLevel(StrEnum):
```
Samoed: Please, use `HelpfulStrEnum`. `StrEnum` is not existing in python3.10

mteb/models/compression_wrappers/compression_wrapper.py
casparil: Not sure about moving `float_type` to `QuatizationLevel`. This might clutter the enum unnecessarily with the many float8 types possible.

casparil: @Samoed anything you'd still like to have changed or that you think needs further discussion? I've resolved all discussions where I updated the code accordingly or which are duplicate to existing ones, so hopefully there's only a few points left to focus on.

Samoed: We need to resolve this issue https://github.com/embeddings-benchmark/mteb/pull/3950#discussion_r2816540190. Also during one task search method can be called multiple times and because of this boundaries would be reset multiple times. Overall, I'm not sure what do to do with approach when we apply quantizations based on `queries`

casparil: So the way it's currently implemented, document embeddings will be quantized per batch, with the thresholds being reset after each batch. Query embeddings are then quantized using the respective batch threshold. With tasks like classification, where multiple experiments are run to estimate performance, the documents are quantized in each run with new thresholds. Query quantization again uses the thresholds computed in the current run. The threshold calculation is only relevant when compressing to int8/int4. I don't see an issue with resetting thresholds in each batch and computing new ones and I'm in favor of quantizing query embeddings per batch as well for the reasons stated in [this comment](https://github.com/embeddings-benchmark/mteb/pull/3950#discussion_r2823193669).

mteb/abstasks/classification.py
```diff
@@ -260,7 +260,7 @@ def _evaluate_subset(
             k: (
                 float(np.mean(values))
                 if (values := [s[k] for s in scores if s[k] is not None])  # type: ignore[literal-required]
-                else np.nan
+                else None
```
KennethEnevoldsen: unsure why this is changed?

mteb/abstasks/retrieval.py
```diff
@@ -382,16 +397,14 @@ def _evaluate_subset(
...
+            search_model = self._get_search_model(model.model, model)
+        elif isinstance(model, CachedEmbeddingWrapper):
+            search_model = self._get_search_model(model._model, model)
```
KennethEnevoldsen: This part of the code should not care if the models is a `CachedEmbeddingWrapper` or `CompressionWrapper`. It should only care about what interfaces (protocols) it implements.

I would revert

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+    BINARY = "binary",
+
+
+class CompressionWrapper:
+    """Wraps a model to quantize the embeddings and compute results on the compressed vectors instead."""
+
+    def __init__(
+        self,
+        model: EncoderProtocol,
+        quantization_level: QuantizationLevel = QuantizationLevel.FLOAT8,
```
KennethEnevoldsen: is there a reason to both have `quantization_level` and `float_type`. I would just use float_type

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+        quantiles: tuple[float, float] | None = None,
+        float_type: torch.dtype = torch.float8_e4m3fn,
+    ) -> None:
+        """Init
```
KennethEnevoldsen: init?

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+            inputs: Batch of inputs to encode.
+            task_metadata: The metadata of the task.
+            hf_split: Split of current task
+            hf_subset: Subset of current task
+            prompt_type: The name type of prompt. (query or passage)
+            batch_size: Batch size
+            **kwargs: Additional arguments to pass to the encoder.
+
+        Returns:
+            The encoded and quantized input in a numpy array of the shape (Number of sentences) x (Embedding dimension).
```
KennethEnevoldsen:
```suggestion
            The encoded and quantized input in an array of the shape (Number of sentences) x (Embedding dimension).
```

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+                2**num_bits * 0.5
+            )
+            quantized = quantized.type(torch.int8)
+        elif self._quantization_level == QuantizationLevel.BINARY:
+            quantized = torch.where(embeddings > 0, 1.0, 0.0)
+        else:
+            raise ValueError(
+                f"Quantization method {self._quantization_level} is not supported!"
+            )
+        return quantized.numpy()
```
KennethEnevoldsen:
```suggestion
        return quantized
```
Numpy conversion is not needed on our side

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+
+        if self.hf_subset != hf_subset or task_metadata.name != self.task_name:
+            self.hf_subset = hf_subset
+            self.task_name = task_metadata.name
+            self._reset_boundaries()
+
+        if not isinstance(embeddings, torch.Tensor):
+            embeddings = torch.tensor(embeddings)
+
+        if prompt_type == PromptType.query and task_metadata.category in [
```
KennethEnevoldsen: I am not sure why this wrapper should care about the modality - the model already supplied the aggregated embeddings in a common space

mteb/models/compression_wrappers/compression_wrapper.py
KennethEnevoldsen: @casparil if you want to do something across the indexing and search steps like what you describe here, then the 'EncoderProtocol` is not the intended format, then you have to implement the [`SearchProtocol`](https://embeddings-benchmark.github.io/mteb/api/model/#mteb.models.SearchProtocol).

mteb/models/model_meta.py
KennethEnevoldsen: @casparil couldn't we use torch.dtypes here? (seems like a reasonable standard)

tests/test_models/test_compression_wrapper.py
```diff
@@ -0,0 +1,144 @@
...
+                "What is the effect of vitamin C on common cold?",
+                "How does exercise affect cardiovascular health?",
+            ]
+        }
+    )
+)
+
+
+def test_float8_compression():
+    model = RandomEncoderBaseline(model_name="dummy", revision=None)
```
KennethEnevoldsen: Should work with just:
```suggestion
    model = mteb.get_model("mteb/baseline-random-encoder")
```

tests/test_models/test_compression_wrapper.py
KennethEnevoldsen, line 1: we also need to ensure that it works across a representative task:

I would make a test here like this one: 

https://github.com/embeddings-benchmark/mteb/blob/b8fe9bd312f512bdf46d71b1665a3c79ba585501/tests/test_integrations/test_dtype_encoders.py#L34-L35

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+        self.hf_subset = None
+        self.task_name = None
+        embed_types = None
+        if quantiles is not None:
+            assert 0 < quantiles[0] < quantiles[1] < 1
+            self.quantiles = torch.tensor(quantiles)
+        if model.mteb_model_meta:
+            embed_types = model.mteb_model_meta.embedding_types
+        if embed_types and quantization_level in embed_types:
+            logger.warning(
```
KennethEnevoldsen: add a `warning.warn`

tests/test_models/test_compression_wrapper.py
```diff
@@ -0,0 +1,144 @@
...
+    wrapper = CompressionWrapper(model, QuantizationLevel.INT8)
+    embeddings = wrapper.encode(
+        task_texts,
+        task_metadata=task_metadata,
+        hf_split="test",
+        hf_subset="test",
+        prompt_type=PromptType.query,
+    )
+    assert wrapper.quantize_queries
+    assert embeddings.dtype == torch.float32
```
KennethEnevoldsen: I am not sure I understand this (see comment above)

tests/test_models/test_compression_wrapper.py
```diff
@@ -0,0 +1,144 @@
...
+    model = RandomEncoderBaseline(model_name="dummy", revision=None)
+    wrapper = CompressionWrapper(model, "full")
+    with pytest.raises(ValueError) as e:
+        wrapper.encode(
+            task_texts,
+            task_metadata=task_metadata,
+            hf_split="test",
+            hf_subset="test",
+        )
+    assert str(e.value) == "Quantization method full is not supported!"
```
KennethEnevoldsen: Don't check for exact string match - just check that a value error is raised and the method is mentioned

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,258 @@
...
+            steps = (maxs - mins) / (2**num_bits - 1)
+            quantized = torch.floor((embeddings - mins) / steps) - int(
+                2**num_bits * 0.5
+            )
+            quantized = quantized.type(torch.int8)
+        elif self._quantization_level == QuantizationLevel.BINARY:
+            quantized = torch.where(embeddings > 0, 1.0, 0.0)
+        else:
+            raise ValueError(
+                f"Quantization method {self._quantization_level} is not supported!"
```
KennethEnevoldsen:
```suggestion
                f"Quantization method '{self._quantization_level}' is not supported!"
```

mteb/models/compression_wrappers/compression_wrapper.py
casparil: Will update

mteb/abstasks/retrieval.py
casparil: Reverted

mteb/abstasks/classification.py
casparil: Sorry for that, I'll change it back. I changed it because I noticed that some results files contained `NAN` values, which technically isn't valid JSON (should be `null`). However, as multiple measures return `np.nan`, the clean way would be to replace the `np.nan` values when writing to file, should you want to have clean JSON.

mteb/models/model_meta.py
casparil: Sure, sounds like a good standard.

mteb/models/compression_wrappers/compression_wrapper.py
casparil: @KennethEnevoldsen thanks for pointing this out. Our initial idea was to enable compression for every model and task regardless of the specific protocol (which currently isn't the case), but this would require additional implementation to support all protocols. Would you limit the wrapper to a specific protocol, i.e. `SearchProtocol`, or change the compression logic to be work across protocols?

Samoed: You can implement `EncoderProtocol` and add methods for `index` and `search` to make your model compatible with `SearchProtocol`, but I don't think you need this

mteb/models/compression_wrappers/compression_wrapper.py
casparil: It would be consistent to also just use torch.dtypes here if they are used to indicate embeddings types natively supported by models. The only issue is that PyTorch currently doesn't have native support for `int4`, so there's no smooth way to cover all implemented quantization levels.

mteb/models/compression_wrappers/compression_wrapper.py
casparil: I've addressed everything @KennethEnevoldsen commented on, so hopefully the only remaining issue is the old one, i.e. whether to (a) quantize queries and documents separately or (b) not.  To avoid this further blocking the PR, let's just reach a decision. @KennethEnevoldsen, @Samoed if you think that option (a) is preferable, I'll adapt the code accordingly (will make code simpler). Or if option (b) makes sense to you, we have to reach a decision on implementing it, i.e. if the wrapper should be changed to use the `SearchProtocol` instead.

@michaeldinzinger @mmilunovic feel free to add your opinions on this.

Samoed: I think it's better to quantize them separately

michaeldinzinger: I also think that quantizing them separately is the preferable choice. I would not expect a significant benefit from quantizing queries and documents together.

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,196 @@
...
+        self.model = model
+        self._quantization_level = quantization_level
+        self.quantiles = None
+        self.min_embeds = 10_000
+        embed_types = None
+        if quantiles is not None:
+            assert 0 < quantiles[0] < quantiles[1] < 1
+            self.quantiles = torch.tensor(quantiles)
+        if model.mteb_model_meta:
+            embed_types = model.mteb_model_meta.embedding_types
```
Samoed: To activate experiments and set updated metadata
```suggestion
        embed_types = model.mteb_model_meta.embedding_types
        model.mteb_model_meta.experiment_kwargs = {"precision": str(quantization_level)}
        model.mteb_model_meta.embedding_types = quantization_level
```

mteb/models/model_meta.py
```diff
@@ -163,6 +164,7 @@ class ModelMeta(BaseModel):
     citation: str | None = None
     contacts: list[str] | None = None
     experiment_kwargs: Mapping[str, Any] | None = None
+    embedding_types: list | None = None
```
Samoed: I think we should specify type of list, but I don't have ideas rather than enum

tests/test_models/test_compression_wrapper.py
```diff
@@ -0,0 +1,103 @@
...
+        task_metadata=task_metadata,
+        hf_split="test",
+        hf_subset="test",
+    )
+    assert torch.max(embeddings) <= 1 and torch.min(embeddings) >= 0
+
+
+def test_invalid_compression():
+    model = mteb.get_model("mteb/baseline-random-encoder")
+    wrapper = CompressionWrapper(model, torch.bfloat16)
```
Samoed: Why this this is invalid?

casparil: Currently 16-bit floats aren't supported, but I'll add them.

mteb/models/model_meta.py
casparil: As Kenneth suggested, I've defined the `embedding_types` in the model files as lists of `torch.dtypes`, but that won't just work with Pydantic. Also, the dtypes don't cover every compression type , i.e. missing int4. It might be better to just define an enum that the model implementations, model meta and compression wrapper all use. @Samoed what is your opinion on this and where would you define the class?

Samoed: I think it's better to create enum for this. I think you can add it here https://github.com/embeddings-benchmark/mteb/blob/main/mteb/types/_encoder_io.py

casparil: All requested changes are implemented, do you see any more issues @Samoed?

tests/test_models/test_compression_wrapper.py
```diff
@@ -0,0 +1,130 @@
...
+    model = mteb.get_model("mteb/baseline-random-encoder")
+    wrapper = CompressionWrapper(model, QuantizationLevel.UINT8)
+    with pytest.raises(ValueError) as e:
+        wrapper.encode(
+            task_texts,
+            task_metadata=task_metadata,
+            hf_split="test",
+            hf_subset="test",
+        )
+    print(e)
```
Samoed:
```suggestion
```

tests/test_models/test_compression_wrapper.py
```diff
@@ -0,0 +1,130 @@
...
+        hf_split="test",
+        hf_subset="test",
+    )
+    assert torch.max(embeddings) <= 1 and torch.min(embeddings) >= 0
+
+
+def test_invalid_compression():
+    model = mteb.get_model("mteb/baseline-random-encoder")
+    wrapper = CompressionWrapper(model, QuantizationLevel.UINT8)
+    with pytest.raises(ValueError) as e:
```
Samoed: Can you add comment why this is invalid?

mteb/models/model_meta.py
```diff
@@ -165,6 +173,7 @@ class ModelMeta(BaseModel):
     citation: str | None = None
     contacts: list[str] | None = None
     experiment_kwargs: Mapping[str, Any] | None = None
+    embedding_types: list[QuantizationLevel] | None = None
```
```suggestion
    embedding_types: QuantizationLevel | list[QuantizationLevel] | None = None
```

mteb/types/_encoder_io.py
```diff
@@ -165,6 +165,27 @@ class MultimodalInput(  # type: ignore[misc]
...
 
 
+class QuantizationLevel(str, Enum):
+    """Enum for valid compression levels.
+
+    Used by the CompressionWrapper class and specified by models to indicate the dtypes of output embeddings they
+    support internally.
+    """
+
+    FLOAT16 = ("float16",)
```
Samoed: Can you change all values like this?
```suggestion
    FLOAT16 = "float16"
```

mteb/types/_encoder_io.py
```diff
@@ -165,6 +165,27 @@ class MultimodalInput(  # type: ignore[misc]
     pass
 
 
+class QuantizationLevel(str, Enum):
```
Samoed: I think we can rename like 
```suggestion
class OutputDType(str, Enum):
```

tests/test_models/test_compression_wrapper.py
```diff
@@ -0,0 +1,130 @@
...
+def test_bf16_compression():
+    model = mteb.get_model("mteb/baseline-random-encoder")
+    wrapper = CompressionWrapper(model, QuantizationLevel.BF16)
+    embeddings = wrapper.encode(
+        task_texts,
+        task_metadata=task_metadata,
+        hf_split="test",
+        hf_subset="test",
+    )
+    assert embeddings.dtype == torch.float32
```
Samoed, line 67:
```suggestion
    # Cast to bf16, then back to float32 using PyTorch as numpy doesn't support bf16
    assert embeddings.dtype == torch.float32
```

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class CompressionWrapper:
+    """Wraps a model to quantize the embeddings and compute results on the compressed vectors instead."""
+
+    def __init__(
+        self,
+        model: EncoderProtocol,
+        quantization_level: QuantizationLevel = QuantizationLevel.FLOAT8_E4M3FN,
+        quantiles: tuple[float, float] | None = None,
```
Samoed: I think this is more commonly called as clipping
```suggestion
        clipping_margin: tuple[float, float] | None = None,
```

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,205 @@
...
+        if self._quantization_level in [
+            QuantizationLevel.FLOAT8_E4M3FN,
+            QuantizationLevel.FLOAT8_E5M2,
+            QuantizationLevel.FLOAT8_E8M0FNU,
+            QuantizationLevel.FLOAT8_E4M3FNUZ,
+            QuantizationLevel.FLOAT8_E5M2FNUZ,
+            QuantizationLevel.FLOAT16,
+        ]:
+            # Cast to float8, then back to float16 using PyTorch as numpy doesn't support float8
+            float_type = getattr(torch, self._quantization_level)
```
Samoed: I think would be good to add method for enum like 
`QuantizationLevel.get_dtype(QuantizationLevel.FLOAT16)`

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,205 @@
...
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.models.model_meta import ModelMeta
+    from mteb.models.models_protocols import EncoderProtocol
+    from mteb.types import Array, BatchedInput, PromptType
+
+logger = logging.getLogger(__name__)
+
+
+class CompressionWrapper:
```
Samoed, line 22: Can you add this to https://github.com/embeddings-benchmark/mteb/blob/2e1e513da3b585efc08708808ebbfcbd79fab56a/docs/api/model.md to show in documentation?

tests/test_models/test_compression_wrapper.py
casparil: I can also add support for uint8/uint4 to the wrapper, it's just currently limited to int

docs/api/model.md
KennethEnevoldsen, line 1: We also need documentation on usage - as well as a entry on the [whats new section](https://embeddings-benchmark.github.io/mteb/whats_new/)

mteb/types/_encoder_io.py
```diff
@@ -165,6 +165,41 @@ class MultimodalInput(  # type: ignore[misc]
...
+        Output types that are not natively supported by PyTorch like 4-bit integers require specific mapping to the
+        desired dtype.
+        """
+        if self == OutputDType.UINT4:
+            return torch.uint8
+        elif self == OutputDType.INT4:
+            return torch.int8
+        elif self == OutputDType.BINARY:
+            return torch.bool
+        return getattr(torch, self)
```
KennethEnevoldsen: Does this work? Shouldn't it be self.value?

mteb/models/model_meta.py
```diff
@@ -132,6 +139,7 @@ class ModelMeta(BaseModel):
         modalities: A list of strings representing the modalities the model supports. Default is ["text"].
         contacts: The people to contact in case of a problem in the model, preferably a GitHub handle.
         experiment_kwargs: A dictionary of parameters used in the experiment that are not covered by other fields. This is used to create experiment names for ablation studies and similar experiments.
+        embedding_types: Output embedding data types (e.g. int8, binary, float) natively supported by the model. If None, it is assumed that the model only returns float embeddings.
```
KennethEnevoldsen: I would probably call it dtypes instead of embedding types

KennethEnevoldsen: or `output_dtypes`

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,214 @@
...
+            quantization_level: The quantization level to use. Has to be supported by the quantize_embeddings method.
+            clipping_margin: Optional lower and upper percentiles to crop embeddings before integer quantization.
+        """
+        self.model = model
+        self._quantization_level = quantization_level
+        self.clipping_margin = None
+        self.min_embeds = 10_000
+        embed_types = model.mteb_model_meta.embedding_types
+        model.mteb_model_meta.experiment_kwargs = {
+            "precision": quantization_level.value
```
KennethEnevoldsen: currently existing models use `output_dtype` - we should be consistent

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,214 @@
...
+        inputs: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        batch_size: int = 32,
+        **kwargs: Any,
+    ) -> Array:
+        """Encodes the given sentences using the encoder, then quantizes the embeddings.
```
KennethEnevoldsen: I would also add a reference to protocol here (similar to what you did below)

mteb/types/_encoder_io.py
casparil: It returns the correct dtype as is, but self.value is cleaner, so I'll change it.

mteb/models/compression_wrappers/compression_wrapper.py
Samoed: `precision` would match with sentence transformers arg name https://github.com/huggingface/sentence-transformers/blob/aebd46c05d3d967d1f2ece54295a162e4b3d5f88/sentence_transformers/SentenceTransformer.py#L877

docs/api/model.md
casparil: Naming should be consistent now and I've added documentation on how to use the wrapper. I can also add a first draft to [whats new](https://embeddings-benchmark.github.io/mteb/whats_new/). One other thing that might need documentation is the new `output_dtypes` parameter for the `ModelMeta` class. Should I add a short paragraph on this to [adding_a_model](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_model/)?

mteb/models/compression_wrappers/compression_wrapper.py
KennethEnevoldsen: Hmm, then we should change `output_dtype` to precision consistently throughout (sorry @casparil) - we will have to name the models, but we can do this in a follow-up PR

docs/api/model.md
KennethEnevoldsen: > Should I add a short paragraph on this to [adding_a_model](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_model/)?

I think you can add it to the admonition on the model meta as an example yes

mteb/models/compression_wrappers/compression_wrapper.py
Samoed: Maybe we can keep `output_dtype`, because we can't handle sentence transformers `precision` as an experiment, because it will be passed into encode only

casparil: So I change `precision` here to `output_dtype`? Or do you prefer `precision`?

KennethEnevoldsen: I am fine with either @casparil please just pick whatever is your preference

Samoed: I think we can keep `output_dtype` to make it more general

docs/whats_new.md
```diff
@@ -7,6 +7,29 @@ icon: lucide/newspaper
 
 This section is an overview of releases for more information check out the autogenerated [changelog](https://github.com/embeddings-benchmark/mteb/blob/main/pyproject.toml).
 
+## New in v2.12
+### Support for Embedding Compression
+
+MTEB v2.12 adds support for evaluating model performance on compressed embeddings, i.e. quantized to int8. Models that internally support compressed output embeddings can now specify them as part of the model's metadata using the `output_dtype` field.
+
+To generate compressed embeddings, the model is needs to be wrapped in the `CompressionWrapper` class as follows:
```
KennethEnevoldsen: link to docs

docs/get_started/advanced_usage/compress_embeddings.md
```diff
@@ -0,0 +1,45 @@
...
+
+Naturally, the process described above is prone to outliers, i.e. maximum and minimum values that rarely occur and are far outside the usual value range. As a result, the integer distribution of the compressed embeddings can be highly imbalanced. To ensure a better value distribution, the top and bottom x percentile of embedding values can be clipped:
+
+```python
+import mteb
+from mteb.models import CompressionWrapper
+from mteb.types import OutputDType
+
+model = mteb.get_model("intfloat/multilingual-e5-large-instruct")
+model_with_compression = CompressionWrapper(model, output_dtype=OutputDType.INT8, clipping_margin=(0.025, 0.975))
```
KennethEnevoldsen, line 43: is this recommended - if so I would make it the default and then add an info box (admonition) with the explanation

KennethEnevoldsen: (I would also make it the default in the whats new example if that is the case)

docs/whats_new.md
```diff
@@ -7,6 +7,29 @@ icon: lucide/newspaper
...
+task = mteb.get_task("AILACasedocs")
+model = mteb.get_model("intfloat/multilingual-e5-large-instruct")
+
+# wrap the model in the compression wrapper and specify the output dtype
+model_with_compression = CompressionWrapper(model, output_dtype=OutputDType.INT8)
+
+# evaluate as normal
+results = mteb.evaluate(model_with_compression, tasks=[task])
+```
+
```
KennethEnevoldsen, line 32:
```suggestion

To learn more about how to evaluate compressed embedding see [related documentation](...).
```

docs/get_started/advanced_usage/compress_embeddings.md
Samoed: I don't think this is recommended, because this would be model specific

docs/whats_new.md
Samoed: Added

docs/whats_new.md
Samoed: Added

KennethEnevoldsen:
I think this is ready to merge, great work @casparil @Samoed et al.!!

(@Samoed I will let you merge, in case you have some final additions)

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,228 @@
...
+        self.clipping_margin = None
+        self.min_embeds = 10_000
+        embed_types = model.mteb_model_meta.output_dtypes
+        model.mteb_model_meta.output_dtypes = [output_dtype]
+        if model.mteb_model_meta.experiment_kwargs is None:
+            model.mteb_model_meta.experiment_kwargs = {}
+        model.mteb_model_meta.experiment_kwargs["output_dtypes"] = output_dtype.value
+
+        if clipping_margin is not None:
+            assert 0 < clipping_margin[0] < clipping_margin[1] < 1
```
Samoed: Not sure that we need to limit clip margin, because different models could have different embedding spreads

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -0,0 +1,228 @@
...
+            self.clipping_margin = torch.tensor(clipping_margin)
+            model.mteb_model_meta.experiment_kwargs["clipping_margin"] = list(
+                clipping_margin
+            )
+        if embed_types and output_dtype in embed_types:
+            msg = (
+                f"The model {model.mteb_model_meta.name} internally supports quantization to "
```
Samoed: Not sure what `internally supports quantization` means. As I understand it, the intention was to notify the user that the model supports different quantizations out of the box (for now only API-only models), but we don’t have a mechanism to process this, and the user wouldn’t understand what they’re supposed to do.

Samoed (commented):
Looks good! I have some small comments

mteb/models/compression_wrappers/compression_wrapper.py
casparil: This is actually required by `torch.quantile`, so it should be checked.

Samoed: Then can you raise an error instead of assert?

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -50,20 +50,24 @@ def __init__(
         self.clipping_margin = None
         self.min_embeds = 10_000
         embed_types = model.mteb_model_meta.output_dtypes
-        model.mteb_model_meta.experiment_kwargs = {"output_dtypes": output_dtype.value}
         model.mteb_model_meta.output_dtypes = [output_dtype]
+        if model.mteb_model_meta.experiment_kwargs is None:
+            model.mteb_model_meta.experiment_kwargs = {}
+        model.mteb_model_meta.experiment_kwargs["output_dtypes"] = output_dtype.value
+
```
casparil, line 57: @Samoed  this currently makes type checking unhappy. Should `experiment_kwargs` be an immutable `Mapping[str, Any]` as defined in `ModelMeta` or is it intended to be  modified as is the case here?

Samoed: This is intended to be modified. Added to ignore

mteb/models/compression_wrappers/compression_wrapper.py
casparil: I've updated the message and hope it's more comprehensible now, but feel free to update it if you think it's still confusing.

casparil:
Great! Thanks @Samoed and @KennethEnevoldsen for your continuous feedback and help on this PR!

PR #3943, opened 2026-01-16
[MAEB] Msclap Eval Error Fix

AdnanElAssadi56:
- Checks for empty audio sample.
- Caches resampler so that it is not created with each sample.

Error:

```
 6%|█████████                                                                                                                                                         | 13/233 [00:48<13:47,  3.76s/it]
Error while evaluating VoxPopuliAccentPairClassification: cannot reshape tensor of 0 elements into shape [-1, 0] because the unspecified dimension size -1 can be any value and is ambiguous
Traceback (most recent call last):
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/run_retrieval4.py", line 169, in <module>
    results = evaluation.run(model, output_folder="/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/results", batch_size=8)
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/mteb/deprecated_evaluator.py", line 539, in run
    raise e
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/mteb/deprecated_evaluator.py", line 495, in run
    results, tick, tock = self._run_eval(
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/mteb/deprecated_evaluator.py", line 181, in _run_eval
    results = task.evaluate(
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/mteb/abstasks/abstask.py", line 183, in evaluate
    scores[hf_subset] = self._evaluate_subset(
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/mteb/abstasks/pair_classification.py", line 105, in _evaluate_subset
    similarity_scores = evaluator(model, encode_kwargs=encode_kwargs)
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/mteb/_evaluators/pair_classification_evaluator.py", line 91, in __call__
    embeddings1 = model.encode(
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/mteb/models/model_implementations/msclap_models.py", line 166, in encode
    audio_embeddings = self.get_audio_embeddings(inputs, **kwargs)
  File "/mnt/workspace/workgroup/chx/mteb-maeb-evaluaton-fixes/mteb/models/model_implementations/msclap_models.py", line 91, in get_audio_embeddings
[...]

PR #3960, opened 2026-01-18
[MAEB] VoiceGenderCluster Dataset Upload

AdnanElAssadi56:
I’ve uploaded the dataset to MTEB here:
https://huggingface.co/datasets/mteb/VoiceGenderClusteringMini

This is consistent with the discussion in:
https://github.com/embeddings-benchmark/mteb/pull/3943

I decided to apply the filtering and subsampling and upload the processed version directly to MTEB. This way, we avoid having to transform the dataset inside the task code and prevent the same issue from resurfacing in future models or evaluations.


mteb/tasks/audio/clustering/eng/voice_gender.py
```diff
@@ -6,11 +6,11 @@ class VoiceGenderClustering(AbsTaskClustering):
...
     metadata = TaskMetadata(
         name="VoiceGenderClustering",
-        description="Clustering audio recordings based on gender (male vs female).",
-        reference="https://huggingface.co/datasets/mmn3690/voice-gender-clustering",
+        description="Clustering audio recordings based on gender (male vs female). Subsampled to 2048 samples.",
+        reference="https://huggingface.co/datasets/mteb/VoiceGenderClusteringMini",
         dataset={
-            "path": "mmn3690/voice-gender-clustering",
-            "revision": "1b202ea7bcd0abd5283e628248803e1569257c80",
+            "path": "mteb/VoiceGenderClusteringMini",
```
Samoed, line 12: Maybe it would be better to keep original task as is and create new with mini prefix/suffix?

AdnanElAssadi: This is an option, but we didn't do it like this elsewhere, and original contains empty audio samples, so it might be problematic if someone tries to evaluate.

AdnanElAssadi:
Results file
[VoiceGenderClustering.json](https://github.com/user-attachments/files/24694386/VoiceGenderClustering.json)

mteb/tasks/audio/clustering/eng/voice_gender.py
Samoed: We already filtered out empty samples

> we didn't do it like this elsewhere

I think we need to rename some tasks

AdnanElAssadi: > We already filtered out empty samples

It was part of dataset transform.

> I think we need to rename some tasks

I did this mostly to remove empty samples, but original was about ~4k and this is ~2k, we can evaluate on original as well. Do we want to rename everything to mini whenever we do stratified subsampling?

Samoed: I don't think we need to add this for stratified task, but if we're reuploading them and doing something with them I think it's better to change name. It's just unclear why we have suffix in dataset name and task stays the same

AdnanElAssadi: Then, we can close this PR if you want, we can stick to filtering on transform.

Samoed: What did you do for undersample? Is it same as in dataset transform?

AdnanElAssadi: I filtered then ran stratified subsampling and pushed to hub (2048 samples).

Samoed: Then I think we should change name or otherwise reevaluate all models

AdnanElAssadi: I don't wanna overload evaluation. Main issue was empty audio samples, which without this PR are being filtered in dataset transform (most recent merge). We can stick to that then and not have to change names.

I still dont quite see why you're viewing doing stratified subsampling in dataset transform as different than having it already subsampled on huggingface. It is even the same random seed. 

We should discuss how we wanna unify this across tasks.

Samoed: I think we can keep tasks without mini for clustering and classification, but other task types, e.g. retrieval (common voice) we need to add it, because this is less common case when such tasks are downsampled. 


Do we want to rerun all evaluations later? Because some tasks were changed and this is hard to keep track of

Samoed:
By the way I reuploaded `VoiceGenderCluster` in https://github.com/embeddings-benchmark/mteb/pull/3961

AdnanElAssadi:
> By the way I reuploaded `VoiceGenderCluster` in #3961

I guess this PR is redundant then.

Issue #3945, opened 2026-01-16
[MAEB] Mark downsampled dataset


Samoed:
We have some downsampled tasks, e.g. `CommonVoice17A2TRetrieval`, but we don't mark them as downsampled or modified. I think we can rename them as `MiniCommonVoice17A2TRetrieval` or something similar and keep full tasks versions too, but without including to benchmarks

@AdnanElAssadi56 @isaac-chung

isaac-chung:
Please check existing maeb tasks. I think we put "mini" at the end of the name?

Samoed:
Not to all of them. As I mention in description we have commonvoice that have this only in dataset repository name, not in task name. Probably there are more tasks like this https://github.com/embeddings-benchmark/mteb/blob/edf141aac84d31b851eae0c2e7bba4bd7281a801/mteb/tasks/audio/any_2_any_retrieval/common_voice.py#L186-L193

This creates inconsistency when we're adding new dataset, e.g. https://github.com/embeddings-benchmark/mteb/pull/3904#discussion_r2689281124

isaac-chung:
I understand. Just meant that for those we added, it was added at the end, not at the beginning like in your example. We should unify the naming.

Samoed:
What do you mean in the end? In `load_data`/`dataset_transform`? If yes, then most of the time we're just downsampling classifications tasks with `stratified_subsampling` which is fine I think, but if we're doing something with retrieval, then we need to state this

diffunity:
@Samoed I think he means the convention for the datasets is "some-dataset-mini" instead of "mini-some-dataset"

Other tasks that are "-mini" prepended in the dataset name but not indicated in the task name:

`mteb/commonlanguage-age-mini`
`mteb/commonlanguage-gender-mini`
`mteb/commonlanguage-lang-mini`
`mteb/speech-commands-mini`
`mteb/voxpopuli-accent-mini`
`mteb/sib-fleurs-multilingual-mini`
`mteb/voxpopuli-mini`
`mteb/birdclef25-mini`
`mteb/nsynth-mini`
`mteb/tut-acoustic-scenes-mini`

AdnanElAssadi:
[...]

PR #3964, opened 2026-01-18
Add mini to common voice tasks

Samoed:
Overall I think this would close https://github.com/embeddings-benchmark/mteb/issues/3945 (at least my main concerns, other tasks fine, because they're classification or clustering)

I've added mini prefix for common voice, because for me this is very huge task, but in our implementation it's pretty small


AdnanElAssadi:
Would having it as a suffix be better?

I found a multilabel classification task doing this:
```AudioSetMiniMultilingualClassification```

Samoed:
Yes, we can name it like this

AdnanElAssadi:
Do you want to rename it then?

Samoed:
Updated

Issue #3959, opened 2026-01-18
Leaderboard not showing table for some languages


Samoed:
### Describe the bug

Leaderboard can't show summary table for mteb eng and some other language specific benchamarks

https://huggingface.co/spaces/mteb/leaderboard/discussions/176

### To reproduce

<img width="1290" height="787" alt="Image" src="https://github.com/user-attachments/assets/2d1a28b7-01fd-48b2-92da-289b2f3ef91f" />

In logs 
```
Traceback (most recent call last):
  File "/mteb/.venv/lib/python3.12/site-packages/gradio/queueing.py", line 763, in process_events
    response = await route_utils.call_process_api(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mteb/.venv/lib/python3.12/site-packages/gradio/route_utils.py", line 354, in call_process_api
    output = await app.get_blocks().process_api(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mteb/.venv/lib/python3.12/site-packages/gradio/blocks.py", line 2106, in process_api
    result = await self.call_function(
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mteb/.venv/lib/python3.12/site-packages/gradio/blocks.py", line 1588, in call_function
    prediction = await anyio.to_thread.run_sync(  # type: ignore
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mteb/.venv/lib/python3.12/site-packages/anyio/to_thread.py", line 61, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mteb/.venv/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 2525, in run_sync_in_worker_thread
    return await future
           ^^^^^^^^^^^^
  File "/mteb/.venv/lib/python3.12/site-packages/anyio/_backends/_asyncio.py", line 986, in run
    result = context.run(func, *args)
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mteb/.venv/lib/python3.12/site-packages/gradio/utils.py", line 1048, in wrapper
    response = f(*args, **kwargs)
[...]

PR #3965, opened 2026-01-19
fix: leaderboard Nan handling

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3959

Seems that we did multiple times drop on same table, but previously there wasn't `Nan` values, so nothing was in list for deletion. And for means per types nans were handled incorrectly

Now everything is working

MIEB
image

MTEB eng v2

image

### DIFF
--- a/mteb/results/task_result.py
+++ b/mteb/results/task_result.py
@@ -650,7 +650,6 @@ class TaskResult(BaseModel):
                 ]
                 seen_subsets = {"default"}
             else:
-                # Use list comprehension for better performance
                 new_scores[split] = [
                     _scores
                     for _scores in self.scores[split]


mteb/benchmarks/_create_table.py
```diff
@@ -58,7 +58,7 @@ def _get_means_per_types(per_task: pd.DataFrame):
                 dict(
                     model_name=model_name,
                     task_type=task_type,
-                    score=scores[tasks].mean(skipna=False),
+                    score=scores[tasks].mean(skipna=True),
```
KennethEnevoldsen: Should it be NaN if a value is missing?

KennethEnevoldsen: I assume this means that we have a bunch of unknown missing values?

Samoed: `skipna=True`
image



If I change to `skipna=False`, then `VisualSTS` column would contains only empty values. image
image



It table will show that models don't have scores on `VisualSTS17Eng` and `VisualSTS17Multilingual` which consist of only `STS17MultilingualVisualSTS` 

https://github.com/embeddings-benchmark/mteb/blob/eaf7c9147bc9bc3be21efb5f87437388d7bcbad7/mteb/tasks/aggregated_tasks/eng/sts17_multilingual_visual_sts_eng.py#L7-L16

but voyage have all (`en-en` for `VisualSTS17Eng`) splits for these tasks https://github.com/embeddings-benchmark/results/blob/82440b14b144482e22b26f59faba0347b7d01ca5/results/voyageai__voyage-multimodal-3/1/STS17MultilingualVisualSTS.json#L75

image

Also, table for ViDoRe shifts a bit 

image
image


I'll try to dig deeper, because other text benchmarks didn't change, but probably we can resolve this later to keep leaderboard without changes

KennethEnevoldsen: yeah would be great to figure out why it is dropped here (we don't want to display the incorrect value and shouldn't drop a correct one)

Samoed: Problem was in loading aggregated tasks, because they have only `default` subset, but their metadata they could have multiple subsets

https://github.com/embeddings-benchmark/mteb/blob/eaf7c9147bc9bc3be21efb5f87437388d7bcbad7/mteb/tasks/aggregated_tasks/multilingual/sts17_multilingual_vision_sts.py#L26-L61

https://github.com/embeddings-benchmark/results/blob/82440b14b144482e22b26f59faba0347b7d01ca5/results/voyageai__voyage-multimodal-3/1/VisualSTS17Multilingual.json#L1-L27

Samoed: Now leaderboard will look correct 
image

mteb/results/task_result.py
```diff
@@ -641,16 +641,27 @@ def validate_and_filter_scores(
...
-                if _scores["hf_subset"] in hf_subsets
-            ]
+            if task.is_aggregate:
+                # aggregate tasks only have the default subset, but in metadata can be multiple
+                new_scores[split] = [
+                    _scores
+                    for _scores in self.scores[split]
+                    if _scores["hf_subset"] == "default"
+                ]
+                seen_subsets = {"default"}
```
KennethEnevoldsen, line 651: I am fine with merging this in as is now, but would add an issue to remove it, as I suspect this should be solved in the `ResultCache`

mteb/results/task_result.py
```diff
@@ -641,16 +641,27 @@ def validate_and_filter_scores(
...
+            else:
+                # Use list comprehension for better performance
```
KennethEnevoldsen:
```suggestion
```

mteb/benchmarks/_create_table.py
KennethEnevoldsen: > Problem was in loading aggregated tasks, because they have only default subset, but their metadata they could have multiple subsets

It doesn't really have subsets.

mteb/results/task_result.py
Samoed: Probably we can solve this with https://github.com/embeddings-benchmark/mteb/issues/3974

KennethEnevoldsen: yes, I suspect so. So we can just merge it here and remove later

KennethEnevoldsen: I would probably add a comment in the code

Issue #3968, opened 2026-01-19
Leaderboard shows error


Samoed:
### Describe the bug

On leaderboard switching, sometimes an error appears. This often related to languages, sometimes to task types

<img width="1966" height="777" alt="Image" src="https://github.com/user-attachments/assets/abeee91a-82bd-40ba-87bc-a077395bad58" />

```
Traceback (most recent call last):
  File "/Users/samoed/Desktop/mteb/.venv/lib/python3.12/site-packages/gradio/queueing.py", line 763, in process_events
    response = await route_utils.call_process_api(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/samoed/Desktop/mteb/.venv/lib/python3.12/site-packages/gradio/route_utils.py", line 354, in call_process_api
    output = await app.get_blocks().process_api(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/samoed/Desktop/mteb/.venv/lib/python3.12/site-packages/gradio/blocks.py", line 2102, in process_api
    inputs = await self.preprocess_data(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/samoed/Desktop/mteb/.venv/lib/python3.12/site-packages/gradio/blocks.py", line 1749, in preprocess_data
    processed_value = block.preprocess(inputs_cached)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/samoed/Desktop/mteb/.venv/lib/python3.12/site-packages/gradio/components/checkboxgroup.py", line 133, in preprocess
    raise Error(
gradio.exceptions.Error: "Value: 'aai' (type: <class 'str'>) is not in the list of choices: ['eng']"
```

### To reproduce

On spaces this not often appears (probably due to caching), but when I'm running locally this error appears almost on every benchmark switch e.g. from `MTEB(Multilingual)` to `MTEB(eng, v2)` above. 

### Are you interested to contribute a fix for this bug?

Yes

KennethEnevoldsen:
Yea I have also seen it. 

@q275343119 any ideas why this might be? 

I suspect it might be that we have a language ("aai") in MMTEB which is not valid for MTEB(eng, v2), and then it errors out saying that this value is no longer valid. I suspect we don't properly clear out the selection when we switch benchmark (should we generally do that?)

Samoed:
[...]

PR #3969, opened 2026-01-19
fix: leaderboard errors

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3968

The leaderboard occasionally crashed when switching benchmarks/tabs because Gradio sent stale checkbox values that no longer existed in the new choice (e.g., aai when only eng was valid). Gradio’s CheckboxGroup preprocessing raises if any selected value isn’t in the current choices, so the UI would throw a server error.

This PR hardens the leaderboard callbacks by skipping Gradio preprocessing on those events, sanitizing selections against the active benchmark filters, and normalizing dict-shaped checkbox payloads before use. It also updates the task info/description to be driven by explicit change events and keeps task-list caching consistent.


KennethEnevoldsen (approved):
Changes look good - ran it locally to see if things run well.

I found that we now have loading times on top of the description?

Screenshot 2026-01-19 at 20 00 28

This seems a bit odd to me. Any way to can avoid it?

Samoed:
Previously description didn't reflect filtering, but I can remove this change

mteb/leaderboard/app.py
```diff
@@ -759,6 +760,19 @@ def on_benchmark_select(benchmark_name):
                 models,
             ],
         )
+        for trigger in [benchmark_select, lang_select, type_select, domain_select]:
+            trigger.change(
+                _update_description,
+                inputs=[benchmark_select, lang_select, type_select, domain_select],
+                outputs=[description],
+                preprocess=False,
+            )
```
Samoed: This part recreates description on filter change

Samoed:
Also progress on description can appear, because I've disabled precompute

KennethEnevoldsen:
> Also progress on description can appear, because I've disabled precompute

Is precomputing what causes the issue?

@q275343119, do you have the time to take a look at this? Would I love a second pair of eyes on this

Samoed:
Yes, I tried to remove this part https://github.com/embeddings-benchmark/mteb/blob/1ca5f3e66f2ff253b0ef052642b254c1cab6a589/mteb/leaderboard/app.py#L770-L776 and rollback description updating to the original format and errors are started appearing again

q275343119:
> > Also progress on description can appear, because I've disabled precompute
> 
> Is precomputing what causes the issue?
> 
> @q275343119, do you have the time to take a look at this? Would I love a second pair of eyes on this

Alright,I'll take a look

q275343119:
Hi @KennethEnevoldsen @Samoed   If we don't want to display the progress bar, we can use the parameter `show_progress="hidden"`.
https://www.gradio.app/main/docs/gradio/markdown#param-event-show-progress

And perhaps removing `benchmark_select` from `trigger_list` could make the description update appear faster, because changes to `benchmark_select` will inevitably cause changes to `lang_select`, `type_select`, and `domain_select`.
When `benchmark_select` is in `trigger_list`, after a change in the benchmark, the interface refreshes the description once, and then waits for subsequent refreshes. This makes it appear as if the description refresh takes a very long time.
```python
        for trigger in [lang_select, type_select, domain_select]:
            trigger.change(
                _update_description,
                inputs=[benchmark_select, lang_select, type_select, domain_select],
                outputs=[description],
                preprocess=False,
                show_progress="hidden"
            )
```

Samoed:
Updated as you suggested

KennethEnevoldsen:
@Samoed I am getting the following error when running `make run-leaderboard`:
```
--- 🚀 Running leaderboard locally ---
uv run --extra leaderboard python -m mteb.leaderboard.app
warning: The `extra-build-dependencies` option is experimental and may change without warning. Pass `--preview-features extra-build-dependencies` to disable this warning.
<frozen runpy>:128: RuntimeWarning: 'mteb.leaderboard.app' found in sys.modules after import of package 'mteb.leaderboard', but prior to execution of 'mteb.leaderboard.app'; this may result in unpredictable behaviour
2026-02-05 11:13:36,038 - PID:40687 - __main__ - INFO - Starting leaderboard app in process 40687
2026-02-05 11:13:36,038 - PID:40687 - __main__ - INFO - === Starting leaderboard app initialization ===
2026-02-05 11:13:36,038 - PID:40687 - __main__ - INFO - Step 1/7: Loading all benchmark results...
2026-02-05 11:13:36,038 - PID:40687 - __main__ - INFO - Loading cached results from disk...
2026-02-05 11:13:36,038 - PID:40687 - __main__ - INFO - Opening file: /Users/au561649/Github/mteb/mteb/leaderboard/__cached_results.json
2026-02-05 11:13:36,038 - PID:40687 - __main__ - INFO - File exists, size: 1000 bytes
2026-02-05 11:13:36,039 - PID:40687 - __main__ - INFO - File opened successfully, attempting JSON parse...
2026-02-05 11:13:36,040 - PID:40687 - __main__ - ERROR - Failed to load cached results from disk: JSONDecodeError
```
I suspect there is an error in the cache:

I would like to redownload it, but it is correctly understood that we don't have a setting for that?

```
❯ mteb leaderboard --help                                             
usage: mteb leaderboard [-h] [--cache-path CACHE_PATH] [--host HOST] [--port PORT] [--share]

options:
  -h, --help            show this help message and exit
  --cache-path CACHE_PATH
                        Path to the cache folder containing model results
  --host HOST           Host to run the leaderboard server on
  --port PORT           Port to run the leaderboard server on
  --share               Create a public URL for the leaderboard
```
---

For now I just manually deleted the cache

KennethEnevoldsen:
Other than the comment above (which I suspect we should just create an issue on), then I think this looks good

Samoed:
> I would like to redownload it, but it is correctly understood that we don't have a setting for that?

Yes. I think this would be solved in https://github.com/embeddings-benchmark/mteb/pull/3870

PR #3973, opened 2026-01-20
Add to_disk and from_disk to ModelResults

### LINKED ISSUE #3034: Add `to_disk` and `from_disk` to `ModelResults` to make JSON serialization easier
### Allow writing of `Results` object to JSON files

Given code like this, it fails because `TaskResults` is not JSON serializable. It would be nice to make this easier since people will want to save these to file.

```
import mteb
import json

# Define the sentence-transformers model name
model_name = "average_word_embeddings_komninos"

model = mteb.get_model(model_name) # if the model is not implemented in MTEB it will be eq. to SentenceTransformer(model_name)
with open("test_results.json", "w") as f:
    json.dumps(results, f)
```

ayush1298:
closes #3034 

We can use `to_disk` and `from_disk` in following way:

```
from mteb.results import ModelResult
res = ModelResult(model_name="all-MiniLM-L6-v2", model_revision="v2", task_results=[])

# Save to disk
res.to_disk("tmp.json")

# Load from disk
loaded_res = ModelResult.from_disk("tmp.json")
```

### DIFF
--- a/mteb/results/model_result.py
+++ b/mteb/results/model_result.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import logging
 import warnings
-from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, cast
 
 import numpy as np
@@ -18,6 +17,7 @@ from .task_result import TaskError, TaskResult
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterable
+    from pathlib import Path
 
     from mteb.abstasks.abstask import AbsTask
     from mteb.abstasks.task_metadata import (
@@ -425,7 +425,6 @@ class ModelResult(BaseModel):
         Args:
             path: The path to the file to save.
         """
-        path = Path(path)
         with path.open("w") as f:
             f.write(self.model_dump_json(indent=2))
 
@@ -439,6 +438,5 @@ class ModelResult(BaseModel):
         Returns:
             The loaded ModelResult object.
         """
-        path = Path(path)
         with path.open("r", encoding="utf-8") as f:
             return cls.model_validate_json(f.read())


mteb/results/model_result.py
```diff
@@ -417,3 +419,32 @@ def modalities(self) -> list[Modalities]:
...
             mods = self.default_modalities
         return list(set(mods))
+
+    def to_disk(self, path: Path) -> None:
+        """Save ModelResult to disk as JSON.
+
+        Args:
+            path: The path to the file to save.
+        """
+        from pathlib import Path
```
Samoed: You can import at high level
```suggestion
```

mteb/results/model_result.py
```diff
@@ -417,3 +419,32 @@ def modalities(self) -> list[Modalities]:
...
+
+        Returns:
+            The loaded ModelResult object.
+        """
+        from pathlib import Path
+
+        path = Path(path)
+        with path.open("r", encoding="utf-8") as f:
+            data = json.load(f)
+        return cls.model_validate(data)
```
Samoed: Maybe use `model_validate_json`?

ayush1298: I was using this as it was used in task_results also.
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/results/task_result.py#L360

Is there any benifit from `model_validate_json`?

Samoed: Probably it would parse json faster, but I'm not sure.

KennethEnevoldsen: @ayush1298 can I ask you to test it. Loading speed could actually matter since the loading results takes a while.

KennethEnevoldsen (approved):
Looks good otherwise.

mteb/results/model_result.py
ayush1298: I can check. Found this one: https://docs.pydantic.dev/latest/concepts/performance/
here its written most of the time `model_validate_json()` is generally faster as compare to `model_validate()`. So, I think we can change it.

KennethEnevoldsen: @ayush1298 will you change the other ones as well and compare the runtime of:

```
import mteb

cache = mteb.ResultCache()
cache.download_from_remote()
results = cache.load_results()
```
It is important that we don't make this slower - faster would be great

ayush1298: So, I have tried this one, 1st I got the results from the above code and then I had to check the runtime for both approaches by loading the results of 10 different models(to get a better idea of the runtime instead of doing it for just 1 model). Below are the results: 

```
Approach 1: json.load() + model_validate()
Time: 1.0361s
Average per file: 103.61ms

Approach 2: read() + model_validate_json()
Time: 0.9310s
Average per file: 93.10ms

Comparison
Speedup: 1.11x (Approach 2 faster)
Time difference: 0.1051s
```

I have changed `model_validate` to `model_validate_json` in the `from_disk` function in both `ModelResult` and `TaskResult.`

mteb/results/model_result.py
```diff
@@ -417,3 +418,27 @@ def modalities(self) -> list[Modalities]:
...
             mods = self.default_modalities
         return list(set(mods))
+
+    def to_disk(self, path: Path) -> None:
+        """Save ModelResult to disk as JSON.
+
+        Args:
+            path: The path to the file to save.
+        """
+        path = Path(path)
```
Samoed: It's already path
```suggestion
```

ayush1298: Sorry, forgot to remove this one.

Samoed (approved):

Issue #3977, opened 2026-01-21
Add check in statistics that we don't have empty audios


Samoed:
### Description of the feature

Currently there are some datasets with empty samples, e.g. https://github.com/embeddings-benchmark/mteb/blob/4a1e4c3808bd35486766c0612a73d5bd2b9759e5/mteb/descriptive_stats/AudioPairClassification/VoxPopuliAccentPairClassification.json#L10

PR #3976, opened 2026-01-20
[MAEB] Filtered VoxPopuliAccentPairCls Dataset

AdnanElAssadi56:
Removes Empty Audio Samples from VocPopuliAccentPairCls Dataset.

Addressing Discussion in https://github.com/embeddings-benchmark/mteb/pull/3943

PR #3978, opened 2026-01-21
[MAEB] Find tasks with empty samples

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3977

We have only 2 tasks with empty samples, but they will be fixed in:
- https://github.com/embeddings-benchmark/mteb/pull/3960
- https://github.com/embeddings-benchmark/mteb/pull/3976


AdnanElAssadi56 (approved):

AdnanElAssadi:
We can remove exceptions when statistics get updated.

Samoed:
Updated statistics

Issue #3982, opened 2026-01-22
Check support for pandas 3.0


Samoed:
https://github.com/pandas-dev/pandas/releases/tag/v3.0.0

Samoed:
Greadio uses `pandas<3` for now

https://github.com/gradio-app/gradio/blob/f4a14e7287850c505a1a88d34db795074bb52b7c/requirements.txt#L16

Samoed:
Gradio support pandas 3 https://github.com/gradio-app/gradio/releases/tag/gradio%406.5.0

PR #3983, opened 2026-01-22
Fix support for datsets 4.5 with pandas 3

Samoed:
I tried to run mteb with pandas 3.14, and we support it (at least main code, I haven't tried to run leaderboard) https://github.com/embeddings-benchmark/mteb/issues/3982. But when I tried to run test on newer version of venv with datasets v4.5 and pandas 3 test https://github.com/embeddings-benchmark/mteb/blob/12d0c17a35f85907184248ee2a2e2c1a9a8104c5/tests/test_integrations/test_integration_with_datasets.py#L19 failed on line https://github.com/embeddings-benchmark/mteb/blob/12d0c17a35f85907184248ee2a2e2c1a9a8104c5/mteb/abstasks/pair_classification.py#L355 for `FarsTail` because this task output `large_string` instead of `int` for labels (probably because pandas uses str type for strings instead of objects, idk). I verified that I get same results for this task

If I update lock it would still have old versions of `datasets` and `pandas`

Close https://github.com/embeddings-benchmark/mteb/issues/3982


Samoed:
Yes, this error appeared again in https://github.com/embeddings-benchmark/mteb/pull/3984

tests/test_search_index/test_search_index.py
```diff
@@ -30,6 +30,7 @@ def test_retrieval_backends(
     task: AbsTaskRetrieval, similarity: ScoringFunction, tmp_path: Path
 ):
     """Test different retrieval backends for retrieval and reranking tasks."""
+    pytest.importorskip("faiss", reason="faiss is not installed")
```
KennethEnevoldsen, line 33: shouldn't we just install it for tests?

KennethEnevoldsen:
Changes look fine, but doesn't seem to solve the issue

KennethEnevoldsen:
hmm tried to reproduce it locally:

```
import pandas
import mteb

assert pandas.__version__ == "3.0.0"

task = mteb.get_task("FarsTail")
res = mteb.evaluate(model, task)
# works!
```

tests/test_search_index/test_search_index.py
Samoed: It's installed in tests. This is when we're running locally and don't have installed it

Samoed:
This appears only when `datasets==4.5.0 pandas==3.0.0`
```bash
uv pip install datasets==4.5.0 pandas==3.0.0
mteb run -m baseline/random-encoder-baseline -t FarsTail
# TypeError: unsupported operand type(s) for +: 'int' and 'numpy.str_'
```

The error is the same as in https://github.com/embeddings-benchmark/mteb/pull/3984

KennethEnevoldsen:
Hmm, still works here:
```
import datasets
import pandas

import mteb

assert pandas.__version__ == "3.0.0"
assert datasets.__version__ == "4.5.0"

task = mteb.get_task("FarsTail")
res = mteb.evaluate(model, task, overwrite_strategy="always")
# works!
```

This is how it looks in debug mode:
Screenshot 2026-01-28 at 21 24 31

Tried to reset the cache in case that was the issue, but it doesn't seem like it.

But proposed a fix in this PR regardless. I suspect it should solve it but sadly could not check.

Samoed:
Did you run your tests from this branch or main? Because in this branch I've changed datasets base repository to fix the problem. Even with your changes I got same error on old dataset loading


Maybe numpy version plays role, but with 2.3 and 2.4 I got same results. You can try to recreate venv from https://github.com/embeddings-benchmark/mteb/pull/4004 and see the problem.

KennethEnevoldsen:
ran it on this branch

> Maybe numpy version plays role, but with 2.3 and 2.4 I got same results. You can try to recreate venv from https://github.com/embeddings-benchmark/mteb/pull/4004 and see the problem.

Yeah the error might be elsewhere - CI seems to work fine now though?

Samoed:
> ran it on this branch

Then you won't see exception, because it fixed by using dataset with specified types. You can use version from `main` to get the error
> Yeah the error might be elsewhere - CI seems to work fine now though?

Yes, because I'm using a different dataset to load

KennethEnevoldsen:
reproduced error in main and then changed

```diff
- np_labels = np.asarray(labels)
+ np_labels = np.asarray(labels, dtype=np.int64)
```

And then the problem is solved.

mteb/models/model_implementations/salesforce_models.py
```diff
@@ -111,7 +111,7 @@ def instruction_template(
     public_training_data=None,
     training_datasets=None,
     citation="""@article{liu2024codexembed,
-  title={CodeXEmbed: A Generalist Embedding Model Family for Multiligual and Multi-task Code Retrieval},
```
Samoed, line 114: This is their real title on arxiv, so I don't know if this counts as a typo or not

KennethEnevoldsen: ahh no then we can just revert that one

Samoed:
I was requesting your reevaluation, because for some reason I was running an old version of that was generated in `build` folder (because I've mistakenly added `no-editable` to sync) and still got this error

PR #4007, opened 2026-01-27
rename bm25s to baseline/bm25s

### LINKED ISSUE #3972: rename bm25s to baseline/bm25s

I fits better with the current baselines we have. We could do a similar thing for humans.

This would allow to remove to model exceptions for these e.g. `_check_name`. It would also be more consistent. 

Some care have to be taken to ensure that it is backwards compatible. E.g. one should still be able to load the model using `mteb.get_model("bm25s")`, though probably some sort of warning.

ayush1298:
closes #3972

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -266,7 +266,7 @@ class ModelMeta(BaseModel):
     @field_validator("name")
     @classmethod
     def _check_name(cls, v: str | None) -> str | None:
-        if v is None or v == "Human":
+        if v is None:
             return v
         if "/" not in v:
             raise ValueError(
--- a/mteb/models/model_implementations/human.py
+++ b/mteb/models/model_implementations/human.py
@@ -2,7 +2,7 @@ from mteb.models import ModelMeta
 
 human = ModelMeta(
     loader=None,
-    name="Human",
+    name="baseline/Human",
     model_type=["dense"],
     languages=["eng-Latn", "ara-Arab", "rus-Cyrl", "dan-Latn", "nob-Latn"],
     open_weights=True,


mteb/models/get_model_meta.py
```diff
@@ -139,6 +145,13 @@ def get_model_meta(
     Returns:
         A model metadata object
     """
+    if model_name in _MODEL_RENAMES:
+        new_name = _MODEL_RENAMES[model_name]
+        msg = f"The model '{model_name}' has been renamed to '{new_name}'. To prevent this warning use the new name."
+        logger.warning(msg)
```
Samoed: I don't think we need to add logging message
```suggestion
```

ayush1298: I thought we are adding both logging and warning in tasks also: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/get_tasks.py#L345-L349
so kept same here. Should I remove it?

Samoed: I think you can remove logger message here too

mteb/models/model_meta.py
```diff
@@ -266,7 +266,7 @@ def _languages_are_valid(
     @field_validator("name")
     @classmethod
     def _check_name(cls, v: str | None) -> str | None:
-        if v is None or v in ("bm25s", "Human"):
+        if v is None or v == "Human":
```
KennethEnevoldsen: should we name human to a baseline as well? (I think that would be reasonable)

ayush1298: Should we do it like: `baseline/Human`?

KennethEnevoldsen: That was what I was thinking yes

ayush1298: Sure, we can do that

KennethEnevoldsen: let us just double check that others agree (@Samoed)

Samoed: Agree, let's change this too. But we need to update name in results repository as well

ayush1298: Will change here, and also create PR in results repo to update names

PR #4010, opened 2026-01-28
fix fleurs citation

Samoed:
I can't find source of this citation, only found this paper https://arxiv.org/abs/2205.12446 (updated in paper)


AdnanElAssadi:
You're right

PR #4017, opened 2026-01-29
fix: Simplify dependencies

KennethEnevoldsen:
To address: https://github.com/embeddings-benchmark/mteb/discussions/4014#discussioncomment-15644719

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,8 +44,8 @@ dependencies = [
     "pytrec-eval-terrier>=0.5.6",
     "pydantic>=2.0.0",
     "polars>=0.20.22",
-    "torch; python_full_version < '3.14'",
-    "torch>=2.9.0; python_full_version >= '3.14'",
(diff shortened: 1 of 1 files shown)


Samoed (approved):
Seems that colpali changed their limits of torch, and we don't need this anymore https://github.com/illuin-tech/colpali/blob/350ea5ebc7c57d24abddc05aea5d0c51a04b5b7d/pyproject.toml#L41

But with newer python versions we would probably face same error

KennethEnevoldsen:
No agree that we could see it again - we should def. continue the discussion

Samoed:
We’ll run into this sooner or later. Next year, a new python version will be released. Our main dependencies may support python 3.15, but some optional dependencies might pin torch versions (like colpali with `torch<2.10`) that won’t support 3.15. Then we’ll have to start pinning versions like this again.

KennethEnevoldsen:
Let's take that point in the discussion thread.

PR #4018, opened 2026-01-30
fix: Improve array typing by also specifying dtype

### LINKED ISSUE #4016: Better array typing

Currently, our typing for arrays is quite poor often using `np.ndarray` instead of something like `NDArray[np.int64]` or more general like `NDArray[np.integer | np.floating].

An example bug that probably would have been easier to trace if we had the typing: #3983

ayush1298:
closes #4016

Some of the things that I followed: 

1. For Model Implementations, I have changed [Array](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/types/_encoder_io.py#L29) to `Array = NDArray[np.floating] | torch.Tensor`. And then I am using the same array in the encode method in all model implementations.
2. For all embedding-related things, I have added  `NDArray[np.floating] ` and not `NDArray[np.integer | np.floating]`, as embeddings will be float
3. For the evaluator, depending on what exactly array is doing, I have added types. Like for labels, it is either `NDArray[np.int64] `or `NDArray[np.integer]`

### DIFF
--- a/mteb/_evaluators/classification_metrics.py
+++ b/mteb/_evaluators/classification_metrics.py
@@ -8,7 +8,9 @@ if TYPE_CHECKING:
     from numpy.typing import NDArray
 
 
-def hamming_score(y_true: NDArray[np.integer], y_pred: NDArray[np.integer]) -> float:
+def hamming_score(
+    y_true: NDArray[np.integer], y_pred: NDArray[np.integer | np.floating]
+) -> float:
     """Compute the Hamming score (a.k.a. label-based accuracy) for multilabel classification.
 
     The Hamming score is the fraction of labels that are correctly predicted for each sample,
--- a/mteb/abstasks/classification.py
+++ b/mteb/abstasks/classification.py
@@ -273,7 +273,7 @@ class AbsTaskClassification(AbsTask):
     def _calculate_scores(
         self,
         y_test: NDArray[np.integer] | list[int],
-        y_pred: NDArray[np.integer] | list[int],
+        y_pred: NDArray[np.integer | np.floating] | list[int],
     ) -> ClassificationMetrics:
         scores = ClassificationMetrics(
             accuracy=accuracy_score(y_test, y_pred),
--- a/mteb/abstasks/multilabel_classification.py
+++ b/mteb/abstasks/multilabel_classification.py
@@ -37,7 +37,7 @@ def _evaluate_classifier(
     y_train: NDArray[np.integer],
     embeddings_test: Array,
     classifier: SklearnModelProtocol,
-) -> tuple[NDArray[np.integer], SklearnModelProtocol]:
+) -> tuple[NDArray[np.integer | np.floating], SklearnModelProtocol]:
     classifier_copy: SklearnModelProtocol = clone(classifier)
     classifier_copy.fit(embeddings_train, y_train)
     return classifier_copy.predict(embeddings_test), classifier_copy
@@ -211,7 +211,7 @@ class AbsTaskMultilabelClassification(AbsTaskClassification):
     def _calculate_scores(  # type: ignore[override]
         self,
         y_test: NDArray[np.integer],
-        y_pred: NDArray[np.integer],
+        y_pred: NDArray[np.integer | np.floating],
         x_test_embedding: Array,
         current_classifier: SklearnModelProtocol,
     ) -> MultilabelClassificationMetrics:
--- a/mteb/models/cache_wrappers/cache_backend_protocol.py
+++ b/mteb/models/cache_wrappers/cache_backend_protocol.py
@@ -5,8 +5,7 @@ from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 if TYPE_CHECKING:
     from pathlib import Path
 
-    import numpy as np
-    from numpy.typing import NDArray
+    from mteb.types import Array
 
 
 @runtime_checkable
@@ -27,7 +26,7 @@ class CacheBackendProtocol(Protocol):
             **kwargs: Additional backend-specific arguments.
         """
 
-    def add(self, item: list[dict[str, Any]], vectors: NDArray[np.floating]) -> None:
+    def add(self, item: list[dict[str, Any]], vectors: Array) -> None:
         """Add a vector to the cache.
 
         Args:
@@ -35,7 +34,7 @@ class CacheBackendProtocol(Protocol):
             vectors: Embedding vector of shape (dim,) or (1, dim).
         """
 
-    def get_vector(self, item: dict[str, Any]) -> NDArray[np.floating] | None:
+    def get_vector(self, item: dict[str, Any]) -> Array | None:
         """Retrieve the cached vector for the given item.
 
         Args:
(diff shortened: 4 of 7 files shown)


ayush1298:
A few doubts where I have not made changes: 
1. Should we do these in docs also like in [defining_the_model.md ](https://github.com/embeddings-benchmark/mteb/blob/main/docs/usage/defining_the_model.md?plain=1#L85) and [adding_the_model.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/contributing/adding_a_model.md?plain=1#L125)
2. When we have some kind of assertion for instance, checking like: https://github.com/embeddings-benchmark/mteb/blob/8cacef77c230d17efd8815a5047c63a45d085479/tests/test_evaluators/test_ClassificationEvaluator.py#L43

Samoed:
> Should we do these in docs also like in [defining_the_model.md ](https://github.com/embeddings-benchmark/mteb/blob/main/docs/usage/defining_the_model.md?plain=1#L85)and [adding_the_model.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/contributing/adding_a_model.md?plain=1#L125)

Yes, but it's better to change them to `Array` type

> When we have some kind of assertion for instance, checking like: https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_evaluators/test_ClassificationEvaluator.py#L43

NDArray is only for typing, not for runtime validation

ayush1298:
> > Should we do these in docs also like in [defining_the_model.md ](https://github.com/embeddings-benchmark/mteb/blob/main/docs/usage/defining_the_model.md?plain=1#L85)and [adding_the_model.md](https://github.com/embeddings-benchmark/mteb/blob/main/docs/contributing/adding_a_model.md?plain=1#L125)
> 
> Yes, but it's better to change them to `Array` type
> 

Added this.

Samoed (approved):

mteb/_evaluators/sklearn_evaluator.py
```diff
@@ -21,11 +22,11 @@
...
 
 class SklearnModelProtocol(Protocol):
-    def fit(self, X: Array, y: np.ndarray | list[int]) -> None: ...  # noqa: N803
-    def predict(self, X: Array) -> np.ndarray: ...  # noqa: N803
+    def fit(self, X: Array, y: NDArray[np.integer] | list[int]) -> None: ...  # noqa: N803
+    def predict(self, X: Array) -> NDArray[np.integer]: ...  # noqa: N803
     def get_params(self) -> dict[str, Any]: ...
     def set_params(self, random_state: int, **kwargs: dict[str, Any]) -> Self: ...
-    def score(self, X: Array, y: np.ndarray | list[int]) -> float: ...  # noqa: N803
+    def score(self, X: Array, y: NDArray[np.integer] | list[int]) -> float: ...  # noqa: N803
```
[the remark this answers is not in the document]
KennethEnevoldsen: @ayush1298 I would address this comment

mteb/abstasks/regression.py
```diff
@@ -123,8 +124,8 @@ def _undersample_data(
 
     def _calculate_scores(  # type: ignore[override]
         self,
-        y_test: np.ndarray | list[int],
-        y_pred: np.ndarray,
+        y_test: NDArray[np.floating] | list[int],
+        y_pred: NDArray[np.floating] | list[int],
```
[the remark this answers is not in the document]
KennethEnevoldsen: @ayush1298 is this ints, floats or both?

mteb/types/_encoder_io.py
```diff
@@ -26,7 +27,7 @@ class EncodeKwargs(TypedDict):
 
 
 # --- Output types ---
-Array = np.ndarray | torch.Tensor
+Array = NDArray[np.floating] | torch.Tensor
```
KennethEnevoldsen: Does this also work for:
`voyageai/voyage-3.5 (output_dtype=int8)` and `voyageai/voyage-3.5 (output_dtype=binary)`?

Samoed: I think we can make it just `NDArray[Any]` then

KennethEnevoldsen: Hmm but would be nice to know that it is not a string

Samoed: Then
```suggestion
Array = NDArray[np.floating | np.integer | np.bool] | torch.Tensor
```

mteb/abstasks/regression.py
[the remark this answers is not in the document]
ayush1298: I think it should be float only, as these are used for calculating `mean_squared_error`, don't know why it was kept as `list[int]`. Should I change it to `list[float]`?

mteb/_evaluators/sklearn_evaluator.py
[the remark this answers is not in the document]
ayush1298: @KennethEnevoldsen As this evaluator will be used for both classification and regression, so should I add something like: 

```
LabelType = TypeVar("LabelType", bound=np.integer | np.floating)
def fit(self, X: Array, y: NDArray[LabelType] | list[int] | list[float]) -> None: ...
```
Then it would correctly identify type from label, basically `integer` for classification tasks and `float` for regression tasks.

Or, I would simply make it as `NDArray[np.floating] | NDArray[np.integer]`

mteb/types/_encoder_io.py
KennethEnevoldsen: yeah I think that is what we should be using

mteb/_evaluators/sklearn_evaluator.py
```diff
@@ -20,13 +21,15 @@
 
 logger = logging.getLogger(__name__)
 
+LabelType = TypeVar("LabelType", np.integer, np.floating)
```
KennethEnevoldsen: I don't think there is a need to define this if we only use it here. I would juse define it below

KennethEnevoldsen:
@ayush1298, it seems like typechecking is failing

mteb/_evaluators/sklearn_evaluator.py
ayush1298: After adding it to below, like:
`y: NDArray[np.integer | np.floating] | list[int | float]`
, somehow don't know why but lintter is failing and giving error: 

```
N803 Argument name `X` should be lowercase
  --> mteb/_evaluators/sklearn_evaluator.py:26:15
   |
24 | class SklearnModelProtocol(Protocol):
25 |     def fit(
26 |         self, X: Array, y: NDArray[np.integer | np.floating] | list[int | float]
   |               ^^^^^^^^
27 |     ) -> None: ...
28 |     def predict(self, X: Array) -> NDArray[np.integer | np.floating]: ...  # noqa: N803
   |
```
So, I am adding "X" under https://github.com/embeddings-benchmark/mteb/blob/main/pyproject.toml#L284

ayush1298: After, changing these, its failing `typechecking` in `evaluator`. The issue is that NDArray[np.integer | np.floating] creates a union type that mypy cannot properly narrow down. So, we can use cast in that case something like: 

`scores_exp = self._calculate_scores(y_test, cast("NDArray[np.integer]", y_pred))`

ayush1298: @Samoed Can you help here, to solve this

Samoed: Fixed

ayush1298: Thanks @Samoed !

PR #4019, opened 2026-01-30
Merge main 3001

Samoed:
Merge main


Samoed:
Windows test is network problem `huggingface_hub.errors.HfHubHTTPError: 504 Server Error: Gateway Time-out for url`. Datasets available failes because in this pr changes `AfriSentiLangClassification` but we know that it not exists

PR #4024, opened 2026-01-31
ci: Pre-commit should not be forced

KennethEnevoldsen:
We want to support people using pre-commit, but we don't want to force it.


Samoed (approved):

KennethEnevoldsen:
Test fails is unrelated to PR. I made an issue for it here:
https://github.com/embeddings-benchmark/mteb/issues/4027

Issue #4027, opened 2026-01-31
Test fails due to `HfHubHTTPError`


KennethEnevoldsen:
E.g. the following is from https://github.com/embeddings-benchmark/mteb/pull/4024

```
[206](https://github.com/embeddings-benchmark/mteb/actions/runs/21544624436/job/62084135275?pr=4024#step:7:1207)
FAILED tests/test_evaluate.py::test_evaluate[mock_retrieval] - huggingface_hub.errors.HfHubHTTPError: 502 Server Error: Bad Gateway for url: https://huggingface.co/api/models/mock/MockSentenceTransformer
FAILED tests/test_models/test_model_meta.py::test_load_cross_encoder_via_get_model_meta - huggingface_hub.errors.HfHubHTTPError: 502 Server Error: Bad Gateway for url: https://huggingface.co/api/models/cross-encoder/ms-marco-TinyBERT-L-2-v2
```

Having tests fail due to unrelated issues can lead us to ignore them.

Samoed:
This happens, because we set https://github.com/embeddings-benchmark/mteb/blob/b45ac3a901ee22e0b35b82b97ebee943f69abb2f/mteb/models/model_meta.py#L421-L426 `compute_metadata` to True. Maybe we can switch to false to remove request to HF

KennethEnevoldsen:
Yeah I don't think we should have that be default true

KennethEnevoldsen:
Hmm no actually I do think that should be default true, but I don't think it should be called by default

Samoed:
Also we're computing metadata from sentence transformers even if we have its implementation https://github.com/embeddings-benchmark/mteb/blob/b45ac3a901ee22e0b35b82b97ebee943f69abb2f/mteb/models/sentence_transformer_wrapper.py#L80

PR #4029, opened 2026-01-31
fix: don't fill-missing metedata by default

KennethEnevoldsen:
This include three changes:
1) rename `compute_missing` to `fill_missing`, it seems like this change was only applied to `get_model_meta`, but not throughout the stack. This is technically breaking so I added a deprecation.
2) passed fill_missing to `from_hub`, I believe this was unintentionally missed
3) set `fill_missing` to false for `from_{model}` to avoid unexpected external calls

fixes https://github.com/embeddings-benchmark/mteb/issues/4027

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -575,7 +575,10 @@ class ModelMeta(BaseModel):
             The generated ModelMeta.
         """
         meta = cls._from_hub(
-            model, revision, fill_missing=fill_missing, compute_metadata=compute_metadata
+            model,
+            revision,
+            fill_missing=fill_missing,
+            compute_metadata=compute_metadata,
         )
         meta.modalities = ["text"]
 


mteb/models/model_meta.py
```diff
@@ -422,21 +422,32 @@ def _from_hub(
         cls,
         model_name: str | None,
         revision: str | None = None,
-        compute_metadata: bool = True,
+        fill_missing: bool = True,
+        compute_missing: bool | None = None,
```
KennethEnevoldsen: If we actually want to use `compute_missing` in the future, this will make it hard, but I think that it is better not have breaking changes.

mteb/models/get_model_meta.py
```diff
@@ -161,7 +161,7 @@ def get_model_meta(
 
         if fill_missing and fetch_from_hf:
             original_meta_dict = model_meta.model_dump()
-            new_meta = ModelMeta.from_hub(model_name)
```
Samoed, line 164: Maybe it would be easier to change in `get_model_meta` `fetch_from_hf` to False?

KennethEnevoldsen: Hmm but then we are breaking behaviour that has been there for a while

KennethEnevoldsen: basically `mteb.get_model("org/myname")` would stop working. Maybe it should but I feel that is a different discussion

Samoed: Thought again and I think disabling `fill_missing` would solve this

Samoed (approved):

mteb/models/model_meta.py
ayush1298: Can't we just keep `fill_missing` only? Why to add `compute_missing` just to give depreciation warning, or are you adding `compute_missing` inplace of `compute_metadata`

mteb/models/get_model_meta.py
KennethEnevoldsen: Agree

mteb/models/model_meta.py
KennethEnevoldsen: AAhh yes I wanted to keep support `compute_metadata` not `compute_missing`

PR #4030, opened 2026-01-31
fix: remove undocumented models or models with no model card

### LINKED ISSUE #3746: Remove models with empty model card
The models seems to have an empty model card:

https://huggingface.co/castorini/monobert-large-msmarco
https://huggingface.co/unicamp-dl/mt5-13b-mmarco-100k

I have added a comment for them to add it. Otherwise, we will probably remove these.

KennethEnevoldsen:
Also removed `scripts/generate_metadata.py`

fixes #3746

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -575,7 +575,10 @@ class ModelMeta(BaseModel):
             The generated ModelMeta.
         """
         meta = cls._from_hub(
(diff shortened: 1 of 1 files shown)


ayush1298:
@KennethEnevoldsen Should we also remove the results from results repo or keep it?

Samoed (approved):

Samoed:
Yes, we should

PR #4033, opened 2026-02-02
fix: Update ViDoRe3 to rank based on the mean

antoineedy:
Hi team!

For ViDoRe V3, we would rather use a model ranking based on the mean of all the tasks than the Borda Rank. I made these changes, tell me if they look fine to you.

By doing this, we want to give more credits to models that have been evaluated on both public and private tasks, notably to encourage model owners to do so :)

---

Before:
SCR-20260202-jleh

After:
SCR-20260202-jkze


KennethEnevoldsen:
Yeah - borda rank is good at handling missing values, which might not be a feature in this case

ManuelFay:
@PierreColombo end of an era :/

Issue #4008, opened 2026-01-27
Evaluate models: nvidia/nemotron-colembed-vl-{3b,4b,8b}-v2


rnyak:
### Model link on Hugging Face

https://huggingface.co/nvidia/llama-nemotron-colembed-vl-3b-v2; https://huggingface.co/nvidia/nemotron-colembed-vl-4b-v2;  https://huggingface.co/nvidia/nemotron-colembed-vl-8b-v2

### What do you want it to be evaluated on?

Vidore3 Private datasets:

[Vidore3TelecomRetrieval](https://embeddings-benchmark.github.io/mteb/overview/available_tasks/documentunderstanding/#vidore3telecomretrieval)
[Vidore3NuclearRetrieval](https://embeddings-benchmark.github.io/mteb/overview/available_tasks/documentunderstanding/#vidore3nuclearretrieval)

### Are you interested in contributing to the evaluation of this model?

No

### Does this model already exist in MTEB?

Yes

boliu61:
hello @Samoed, may I follow up on this private dataset evaluation request? Thanks

KennethEnevoldsen:
We typically aim to get an evaluation run up and running within a week. I tried to run this, but it seems like the model dependencies are not specified correctly [#4025](https://github.com/embeddings-benchmark/mteb/issues/4025).

We will probably get to that, but it is not our highest priority. If you want to fast-track this, you will have to resolve those dependencies. Note that the warning in the model implementation is incorrect: it should specify the optional dependency, not the transformers version. This is to ensure consistency across models.

Samoed:
Live on leaderboard!
<img width="1858" height="590" alt="Image" src="https://github.com/user-attachments/assets/04cf2dda-3eeb-4451-801d-7002a503c0c4" />

boliu61:
Thank you @Samoed @KennethEnevoldsen for the quick turnaround!

PR #4036, opened 2026-02-02
fix: Add Optional dependencies for NemotronColEmbed models as extras

### LINKED ISSUE #4025: Optional dependency for llama-embed-nemotron is not correct
in pyproject.toml it is currently:

```
```

but in code in can be multiple:
```
llama_nemotron_colembed_vl_3b_v2 = ModelMeta(
    loader=LlamaNemoretrieverColembed,
    loader_kwargs=dict(
        trust_remote_code=True,
        transformers_version_constraint="==4.49.0",
    ),
# ...
)

nemotron_colembed_vl_4b_v2 = ModelMeta(
    loader=LlamaNemoretrieverColembed,
    loader_kwargs=dict(
        trust_remote_code=True,
        transformers_version_constraint="==5.0.0rc0",
    ), 
# ...
)
```

Ideally, we should treat these as optional dependencies rather than include them in the code itself. I would split it up into two set of optional dependencies. (@Samoed let me know what you think)

---

Additionally, when I run it I also get the errors:

```
  File "/home/kenneth/.cache/uv/environments-v2/tmp-nvidia-colembed-35a57bb6b651c08b/lib/python3.12/site-packages/transformers/modeling_utils.py", line 3611, in from_pretrained
    raise ImportError(
ImportError: Using `low_cpu_mem_usage=True` or a `device_map` requires Accelerate: `pip install 'accelerate>=0.26.0'`
```

```
  File "/home/kenneth/.cache/uv/environments-v2/tmp-nvidia-colembed-35a57bb6b651c08b/lib/python3.12/site-packages/transformers/modeling_utils.py", line 1710, in _check_and_enable_flash_attn_2
    raise ImportError(f"{preface} the package flash_attn seems to be not installed. {install_message}")
ImportError: FlashAttention2 has been toggled on, but it cannot be used due to the following error: the package flash_attn seems to be not installed. Please refer to the documentation of https://huggingface.co/docs/transformers/perf_infer_gpu_one#flashattention-2 to install Flash Attention 2.
```

discovered when trying to run #4008

oliverholworthy:
Fixes #4025

This PR addresses the issues raised in #4025 and incorporates feedback from #4008 regarding dependency handling for the NVIDIA NemotronColEmbed model family.

### Changes

**1. Fixed optional dependencies in `pyproject.toml`**

Added new extras with required dependencies and version constraints:
```toml
llama-nemotron-colembed-vl = ["transformers[torch]==4.49.0", "torchvision>=0.22.0", "flash-attn>=2.6.3", "accelerate"]
nemotron-colembed-vl-v2 = ["transformers[torch]==5.0.0rc0", "torchvision>=0.22.0", "flash-attn>=2.6.3", "accelerate"]
```

This fixes the missing `accelerate`, `torchvision`, and `flash-attn` dependencies that were causing import errors.

**2. Refactored model implementation**

- Renamed `nvidia_llama_nemoretriever_colemb.py` to `nvidia_nemotron_colembed_vl.py`
- Consolidated `LlamaNemoretrieverColembed` into a single `NemotronColEmbedVL` class
- Added `extra_name` parameter to specify which optional dependency to use
- Version checking now points users to the correct optional dependency (addressing feedback: *"the warning should specify the optional dependency, not the transformers version"*)

### Version constraint handling

Different models in this family require different transformers versions:
- `llama-nemotron-colembed-vl` extra → `transformers==4.49.0` (for v1 models and llama-nemotron-colembed-vl-3b-v2)
- `nemotron-colembed-vl-v2` extra → `transformers>=5.0.0rc0` (for 4b-v2 and 8b-v2 models)

The code validates the transformers version at runtime and directs users to install the correct extra if there's a mismatch, ensuring correct embeddings are produced.

### Usage

```bash
# For v1 models and llama-nemotron-colembed-vl-3b-v2
pip install 'mteb[llama-nemotron-colembed-vl]'

# For nemotron-colembed-vl-4b-v2 and nemotron-colembed-vl-8b-v2
pip install 'mteb[nemotron-colembed-vl-v2]'
```


Samoed (commented):
Looks good but why revision and citations were changed?

oliverholworthy:
> revision and citations were changed

Revision change is for fixes to support transformers 5.0.0 (instead of rc0)
  - https://huggingface.co/nvidia/nemotron-colembed-vl-8b-v2/commit/b6db318a581be528d8a24c25f0857d7d833d5263
  - https://huggingface.co/nvidia/nemotron-colembed-vl-4b-v2/commit/6417b4ca925bfda7fc1415d3253ae8f8842ba8ee
  
Citation: I've reverted unrelated citation change.

KennethEnevoldsen:
Thanks for the PR @oliverholworthy - I will try to give these a spin

KennethEnevoldsen:
@oliverholworthy using the following code:

```
uv run tmp.py
```
where tmp.py is:

```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "mteb[nemotron-colembed-vl]>=2.7.14",
# ]
# ///
import mteb
tasks = mteb.get_tasks(...)

mdls = [
    "nvidia/llama-nemotron-colembed-vl-3b-v2",
]

for mdl in mdls:
    print("Evaluating model:", mdl)
    meta = mteb.get_model(mdl, device="cuda:0")
    results = mteb.evaluate(
        meta,
        tasks,
        cache=None, # don't use cache
        encode_kwargs={"batch_size": 16},
    )
```
I get a warning that accelerate is not installed. I suspect we need to add that one as well

(adding the dependency manually I get it to work)

pyproject.toml
```diff
@@ -99,6 +99,8 @@ sauerkrautlm-colpali = ["sauerkrautlm-colpali>=0.1.0; python_full_version < '3.1
 xet = ["huggingface_hub>=0.32.0"]
 youtu = ["tencentcloud-sdk-python-common>=3.0.1454", "tencentcloud-sdk-python-lkeap>=3.0.1451"]
 llama-embed-nemotron = ["transformers==4.51.0"]
+nemotron-colembed-vl = ["transformers[torch]==4.49.0", "torchvision>=0.22.0", "flash-attn>=2.6.3"]
+nemotron-colembed-vl-v2 = ["transformers[torch]>=5.0.0", "torchvision>=0.22.0", "flash-attn>=2.6.3"]
```
rnyak: we would need accelerate to be installed as well. `pip install 'accelerate>=1.1.0'`

pyproject.toml
```diff
@@ -99,6 +99,8 @@ sauerkrautlm-colpali = ["sauerkrautlm-colpali>=0.1.0; python_full_version < '3.1
 xet = ["huggingface_hub>=0.32.0"]
 youtu = ["tencentcloud-sdk-python-common>=3.0.1454", "tencentcloud-sdk-python-lkeap>=3.0.1451"]
 llama-embed-nemotron = ["transformers==4.51.0"]
+nemotron-colembed-vl = ["transformers[torch]==4.49.0", "torchvision>=0.22.0", "flash-attn>=2.6.3"]
```
rnyak: do u think better to name this as `llama-nemotron-colembed-vl-v2`?

oliverholworthy: this extra is for the v1 models and llama-nemotron-colembed-vl-v2 which was tested with 4.49.0. but maybe would be clearer to have a separate one for v1 and the llama-v2 variant as you suggest

pyproject.toml
oliverholworthy: accelerate is part of the `torch` extra of transformers so already included as part of this dependency specifier set

pyproject.toml
oliverholworthy: I've addd the llama prefix `llama-nemotron-colembed-vl` (without version suffix)  to make it clearer which models it corresponds to. (no version suffix because this applies to both v1 and v2 llama-nemotron-colembed models

oliverholworthy:
@KennethEnevoldsen Please could you try `uv run` with the `--extra` kwarg to ensure it's picking up the correct set of dependencies. My suspicion is the extra wasn't installed correctly (pip silently ignores unknown extras, while uv will raise an error if `--extra` doesn't exist. 

`accelerate` is included [as part of `transformers[torch]`](https://github.com/huggingface/transformers/blob/08810b1e278938278c50153ee1edfd7a20a759da/setup.py#L230) so should be picked up correctly with this:

```
uv run --extra llama-nemotron-colembed-vl tmp.py
```

oliverholworthy:
@Samoed @KennethEnevoldsen Note I've reverted back to specify transformers 5.0.0rc0 for the 4b/8b models because we still have some ongoing investigation into differences that could result in output discrepancies compared to the results we submitted with rc0. For consistency we should continue with rc0 while we continue to look at the implementation

boliu61:
Hi @Samoed @KennethEnevoldsen, this PR should fix the dependencies issue now. May I request the private evaluation results in the next 24 hours, if possible? We have some internal timeline we want to hit. Thanks!

Samoed:
Hi! Sorry for delay, I'll run evaluation

pyproject.toml
```diff
@@ -99,6 +99,8 @@ sauerkrautlm-colpali = ["sauerkrautlm-colpali>=0.1.0; python_full_version < '3.1
 xet = ["huggingface_hub>=0.32.0"]
 youtu = ["tencentcloud-sdk-python-common>=3.0.1454", "tencentcloud-sdk-python-lkeap>=3.0.1451"]
 llama-embed-nemotron = ["transformers==4.51.0"]
+llama-nemotron-colembed-vl = ["transformers[torch]==4.49.0", "torchvision>=0.22.0", "flash-attn>=2.6.3"]
+nemotron-colembed-vl-v2 = ["transformers[torch]==5.0.0rc0", "torchvision>=0.22.0", "flash-attn>=2.6.3"]
```
```suggestion
llama-nemotron-colembed-vl = ["transformers[torch]==4.49.0", "torchvision>=0.22.0", "flash-attn>=2.6.3", "accelerate"]
nemotron-colembed-vl-v2 = ["transformers[torch]==5.0.0rc0", "torchvision>=0.22.0", "flash-attn>=2.6.3", "accelerate"]
```

oliverholworthy: `accelerate` was already covered by [the `transformers[torch]` extra, which includes accelerate](https://github.com/huggingface/transformers/blob/08810b1e278938278c50153ee1edfd7a20a759da/setup.py#L230) However, I agree it is clearer to have it listed here too for clarity.

oliverholworthy:
Thanks for taking a look and running the evaluation, @Samoed  Please let me know if you run into any issues or if there’s anything I can help with on my end.

Samoed:
I've almost evaluated and everything is good. I'll finish run in an hour

KennethEnevoldsen:
> accelerate is included [as part of transformers[torch]](https://github.com/huggingface/transformers/blob/08810b1e278938278c50153ee1edfd7a20a759da/setup.py#L230) so should be picked up correctly with this:

Hmm I don't see this if I create the lock file using `uv lock --script tmp.py`:


```toml
[[package]]
name = "transformers"
version = "5.0.0"
source = { registry = "https://pypi.org/simple" }
dependencies = [
    { name = "filelock" },
    { name = "huggingface-hub" },
    { name = "numpy" },
    { name = "packaging" },
    { name = "pyyaml" },
    { name = "regex" },
    { name = "safetensors" },
    { name = "tokenizers" },
    { name = "tqdm" },
    { name = "typer-slim" },
]
sdist = { url = "https://files.pythonhosted.org/packages/bc/79/845941711811789c85fb7e2599cea425a14a07eda40f50896b9d3fda7492/transformers-5.0.0.tar.gz", hash = "sha256:5f5634efed6cf76ad068cc5834c7adbc32db78bbd6211fb70df2325a9c37dec8", size = 8424830, upload-time = "2026-01-26T10:46:46.813Z" }
wheels = [
    { url = "https://files.pythonhosted.org/packages/52/f3/ac976fa8e305c9e49772527e09fbdc27cc6831b8a2f6b6063406626be5dd/transformers-5.0.0-py3-none-any.whl", hash = "sha256:587086f249ce64c817213cf36afdb318d087f790723e9b3d4500b97832afd52d", size = 10142091, upload-time = "2026-01-26T10:46:43.88Z" },
]
```

We also don't have `transformers[torch]` as a dependency, neither does sentence transformers it seems.

Samoed:
> Hmm I don't see this if I create the lock file using uv lock --script tmp.py

I'm not sure that uv handles correctly `mteb[nemotron-colembed-vl]` in script, because I can't find in `tmp.py.lock` even accelerate. Probably uv try to fetch from pypi firstly, because in lock transformers have `5.0.0` version not `5.0.0rc0`. But when I change to (as in pyproject for `nemotron-colembed-vl`)
```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "transformers[torch]==5.0.0rc0", "torchvision>=0.22.0", "flash-attn>=2.6.3", "accelerate"
# ]
# ///
```
I can find in lock
```toml
[[package]]
name = "transformers"
version = "5.0.0rc0"
source = { registry = "https://pypi.org/simple" }
dependencies = [
    { name = "filelock" },
    { name = "huggingface-hub" },
    { name = "numpy" },
    { name = "packaging" },
    { name = "pyyaml" },
    { name = "regex" },
    { name = "requests" },
    { name = "safetensors" },
    { name = "tokenizers" },
    { name = "tqdm" },
    { name = "typer-slim" },
]
sdist = { url = "https://files.pythonhosted.org/packages/94/29/835a5522b672c394b1c808b4f94a121b95fc8a8c58dc61404d69668f9d3e/transformers-5.0.0rc0.tar.gz", hash = "sha256:bb427000caa4a88943704f80448b2323ad8c6a2f4f13c1433e27d0a1f690c975", size = 8181669, upload-time = "2025-12-01T18:11:45.336Z" }
wheels = [
    { url = "https://files.pythonhosted.org/packages/8e/a9/fc60ee518b57fefca1551b15b36a5e934a101148327003fd18741c732b5b/transformers-5.0.0rc0-py3-none-any.whl", hash = "sha256:1935f8b396891c93b8520d951d4385da1b1b778914e1d79ed151ddbd32d83a22", size = 9815900, upload-time = "2025-12-01T18:11:42.697Z" },
]

[package.optional-dependencies]
torch = [
    { name = "accelerate" },
    { name = "torch" },
]
```

> We also don't have transformers[torch] as a dependency, neither does sentence transformers it seems.

It was added to `nemotron-colembed-vl`

KennethEnevoldsen:
Hmm, I suspect it might be resolving from pypi and not throwing an error because it can't find it and I see that transformers[torch] was added here (which I missed).

I double checked from the branch and I can see accelerate just fine - sorry for the confusion

Samoed:
Why are you set batch size by one? https://github.com/embeddings-benchmark/mteb/blob/363a27e6fcfdbe907346c271e593cb39a82a1eeb/mteb/models/model_implementations/nvidia_nemotron_colembed_vl.py#L85-L87
https://github.com/embeddings-benchmark/mteb/blob/363a27e6fcfdbe907346c271e593cb39a82a1eeb/mteb/models/model_implementations/nvidia_nemotron_colembed_vl.py#L113

I didn't notice firstly when I was running, but after all results computed I found this. Even if `encode` receives `batch_size`

CC @boliu61 @rnyak

rnyak:
> Why are you set batch size by one?
> 
> https://github.com/embeddings-benchmark/mteb/blob/363a27e6fcfdbe907346c271e593cb39a82a1eeb/mteb/models/model_implementations/nvidia_nemotron_colembed_vl.py#L85-L87
> 
> 
> https://github.com/embeddings-benchmark/mteb/blob/363a27e6fcfdbe907346c271e593cb39a82a1eeb/mteb/models/model_implementations/nvidia_nemotron_colembed_vl.py#L113
> 
> I didn't notice firstly when I was running, but after all results computed I found this. Even if `encode` receives `batch_size`
> 
> CC @boliu61 @rnyak

@Samoed hello. by setting batch_size=1, our intention was to avoid OOM. thanks.

Samoed:
Users can change `batch_size` if they have OOM. I don't think you need to set it to 1

PR #4038, opened 2026-02-02
fix: set num proc to `None` by default

Samoed:
Datasets uses `None` by default, but seems `1` changes behavior


Samoed:
It used by for `datasets.map` and `torch.Dataloader`. I didn't add it to sklearn.

For dataloader it's 0 by default and handled with `num_workers=num_proc if num_proc is not None and num_proc > 1 else 0,`

PR #4031, opened 2026-01-31
fix: Fill in embedding and total parameters in ModelMeta

ayush1298:
This PR fills Total/Embedding parameters in ModelMeta for model listed under #3947 , so that we can calculate Active Parameters column on LB and merge PR https://github.com/embeddings-benchmark/mteb/pull/3979
Also, exact parameters are updated for each model, this will close #3811

PR #4043, opened 2026-02-03
fix docs links

Samoed:
Build was also failing due to `--strict` like in https://github.com/embeddings-benchmark/mteb/actions/runs/21631998035/job/62346912892 (ref #4031)


ayush1298:
@Samoed Can you fix linting also.

Samoed:
This comes from the changelog. @KennethEnevoldsen, do we want to include the changelog in this format in `main` or not? I don’t think releases listing all commits make a useful changelog. We might want to switch to using PR titles only. I think we can remove changelog for now.

KennethEnevoldsen (approved):
Yeah let us remove the changelog

PR #4044, opened 2026-02-03
fix docs deploy command

Samoed:
Previously we don't build documentation in strict mode on main and because of this it was passing on ci, but failed in PRs


KennethEnevoldsen (approved):
Ahh great - yes good edit!

Issue #2865, opened 2025-06-28
[maeb] Move all MAEB datasets into MTEB org on HF


isaac-chung:
Such as https://github.com/embeddings-benchmark/mteb/pull/2849.

isaac-chung:
Related to https://github.com/embeddings-benchmark/mteb/pull/2643

isaac-chung:
@Samoed was there a script that was used for a similar purpose in MMTEB or MIEB?

Samoed:
For mieb datasets weren't moved. We have function `push_dataset_to_hub`, but this would require download full repo and upload it (which is better in some cases). But easiest solution would be to create duplicates (forks) datasets without downloading or changing the data

isaac-chung:
Yeah, let's do that (fork). If there isn't already an existing issue for MIEB datasets, we should add one. 
Is there already a script to programmatically fork datasets?

Samoed:
> If there isn't already an existing issue for MIEB datasets, we should add one.

Seems we don't have, at least I can't find it

> Is there already a script to programmatically fork datasets?

Seems you can duplicate by this method https://huggingface-openapi.hf.space/#tag/repos/POST/api/{repoType}/{namespace}/{repo}/duplicate. Note that revision of duplicated repositories will change

Example what I got when duplicating https://huggingface.co/datasets/mteb/fsdkaggle2019-parquet from browser requests logs (removed headers and cookies part)

```
curl 'https://huggingface.co/api/datasets/confit/fsdkaggle2019-parquet/duplicate' \
  -X POST \
  --data-raw '{"repository":"mteb/fsdkaggle2019-parquet","private":false,"hardware":"cpu-basic","sleepTimeSeconds":3600,"secrets":[],"variables":[],"storageTier":null}'
```

isaac-chung:
If MIEB hasn't done it either, then this is not a blocker for release.

KennethEnevoldsen:
Yeah the duplicate feature is very convenient. I think this should be done before the release, at least, even if it wasn't for MIEB. We don't need to do it for all, but I think we should do it for the core MAEB, seems like we are only missing:

```
import mteb

b = mteb.get_benchmark("MAEB")

for t in b.tasks:
    if not t.metadata.dataset["path"].startswith("mteb/"):
        print(t.metadata.dataset["path"])

# arteemg/spoken-squad-t2a
# silky1708/CREMA-D
[...]

PR #4048, opened 2026-02-05
[maeb] Move datasets to mteb org

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/2865

Now only dataset that not in MTEB is https://huggingface.co/datasets/agkphysics/AudioSet which weights 5TB


KennethEnevoldsen:
> Now only dataset that not in MTEB is https://huggingface.co/datasets/agkphysics/AudioSet which weights 5TB

Shouldn't we just duplicate it?

PR #4067, opened 2026-02-07
fix: Remove task performance by type tab when there is only one type

### LINKED ISSUE #4046: Remove `task performance by type` when the plot can't be generated

Currently, for the retrieval-only benchmark, we try to make a task performance by type radar chart, but it fails. I would suggest removing the tab.

Currently it looks like:

Image

ayush1298:
closes #4046 
Added a check so that Radar Plot can't be generated when there is only 1 task type.

LB Image for RTEB: 

image

LB Image for Vidore(v3):

image

There are other benchmarks also, which have just 1 task type, so this tab is removed.

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -450,7 +450,11 @@ def get_leaderboard_app(cache: ResultCache = ResultCache()) -> gr.Blocks:
 
     default_benchmark = mteb.get_benchmark(DEFAULT_BENCHMARK_NAME)
     default_results = all_benchmark_results[default_benchmark.name]
-    default_task_types = set(default_results.task_types)
+    default_task_types = {
+        task_type
+        for task_type in default_results.task_types
+        if task_type != "InstructionRetrieval"
+    }
     display_radar_chart = len(default_task_types) > 1
 
     logger.info("Step 4/7: Filtering models...")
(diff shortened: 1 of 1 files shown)


mteb/leaderboard/app.py
```diff
@@ -450,6 +450,12 @@ def get_leaderboard_app(cache: ResultCache = ResultCache()) -> gr.Blocks:
 
     default_benchmark = mteb.get_benchmark(DEFAULT_BENCHMARK_NAME)
     default_results = all_benchmark_results[default_benchmark.name]
+    default_task_types = {
+        task_type
+        for task_type in default_results.task_types
+        if task_type != "InstructionRetrieval"
```
Samoed, line 456: Why do you skip `InstructionRetrieval`?

ayush1298: Didn't remove it at the start. But it's suggested by Copilot as having negative scores. also written here: 
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/leaderboard/figures.py#L190-L193

Samoed: I see. Maybe move to function then? Because you duplicate this 2 times

ayush1298: Didn't create a function as its just 1 liner statement.

ayush1298:
> Can you add a image showing that it does appear - otherwise I don't have anything to add besides what is already here

This one for MTEB(Multilingual, v2)

image

ayush1298:
@KennethEnevoldsen Can we merge these one also?

Issue #3546, opened 2025-11-11
Resample audio in dataloader


Samoed:
should this be done in the collate_fn to avoid stalling the GPU

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/pull/3497#discussion_r2477713592_

isaac-chung:
Did we run all paper results without a fix for this issue? Is it a blocker or an optimization?

Samoed:
We haven't implemented this yet. This is just optimization, I'll try to implement today

PR #4069, opened 2026-02-07
[MAEB] Use collator for processing

Samoed:
Changes processing in audio models to use collator. I've run some tasks on CLAP and got same scores

Close https://github.com/embeddings-benchmark/mteb/issues/3546

### DIFF
--- a/mteb/_create_dataloaders.py
+++ b/mteb/_create_dataloaders.py
@@ -317,7 +317,7 @@ def _custom_collate_fn(batch: list[dict[str, Any]]) -> BatchedInput:
             if any(item[key] is None for item in batch):
                 raise ValueError(f"Found None in batch for key '{key}'")
             collated[key] = default_collate([item[key] for item in batch])
-    return collated  # type: ignore[return-value]
+    return cast("BatchedInput", collated)
 
 
 def _create_image_dataloader(
@@ -602,10 +602,13 @@ class AudioCollator:
                 array=audio_array, sampling_rate=target_sampling_rate
             )
             collated_inputs.append(row)
-        return {  # type: ignore[return-value]
-            key: [row[key] for row in collated_inputs]
-            for key in collated_inputs[0].keys()
-        }
+        return cast(
+            "BatchedInput",
+            {
+                key: [row[key] for row in collated_inputs]
+                for key in collated_inputs[0].keys()
+            },
+        )
 
     @staticmethod
     def resample_audio(


AdnanElAssadi:
Does collator allow any empty sample slip up?

I am not sure why this happened before for us.

Samoed:
Collator doesn't allow empty values

mteb/_create_dataloaders.py
```diff
@@ -315,15 +318,15 @@ def _custom_collate_fn(batch: list[dict[str, Any]]) -> dict[str, Any]:
             if any(item[key] is None for item in batch):
                 raise ValueError(f"Found None in batch for key '{key}'")
             collated[key] = default_collate([item[key] for item in batch])
-    return collated
+    return collated  # type: ignore[return-value]
```
KennethEnevoldsen: is this not type compatible?

mteb/_create_dataloaders.py
```diff
@@ -552,3 +555,103 @@ def create_dataloader(
...
+        """Initialize the collator.
+
+        Args:
+            target_sampling_rate: The sampling rate to resample the audio to.
+            max_samples: The maximum number of samples to keep for each audio. If None, no truncation is applied.
+        """
+        self.target_sampling_rate = target_sampling_rate
+        self.max_samples = max_samples
+
+    def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
```
KennethEnevoldsen, line 574:
```suggestion
    def __call__(self, inputs: BatchedInput) -> BatchedInput:
```
assumed

KennethEnevoldsen: if so then also fix below

mteb/_create_dataloaders.py
```diff
@@ -552,3 +555,103 @@ def create_dataloader(
...
+            audio_array = AudioCollator.resample_audio(
+                row,
+                target_sampling_rate,
+                max_samples,
+            )
+            row["audio"] = AudioInputItem(
+                array=audio_array, sampling_rate=target_sampling_rate
+            )
+            collated_inputs.append(row)
+        return {  # type: ignore[return-value]
```
KennethEnevoldsen: probably the same typing error as above - something that we should look into?

mteb/_create_dataloaders.py
```diff
@@ -552,3 +555,103 @@ def create_dataloader(
...
+            audio_array = np.mean(audio_array, axis=0)
+
+        if max_samples is not None:
+            num_samples = audio_array.shape[-1]
+            if num_samples > max_samples:
+                audio_array = audio_array[..., :max_samples]
+        return audio_array
+
+
+@lru_cache(maxsize=10)
```
KennethEnevoldsen: does this need an LRU?

KennethEnevoldsen (commented):
Generally looks good, a few minor comments

Should we just do a quick speed comparison to make sure that things don't get slower?

mteb/_create_dataloaders.py
Samoed: Type is correct. `collate_fn` will receive elements like:
```
[
  {"text": "text1"},
  {"text": "text2"},
]
```
while `BatchedInput` should be
```
{
  "text": ["text1", "text2"],
}
```

mteb/_create_dataloaders.py
Samoed: We have this optimization in one model, but it seems that resample init doesn't do anything complex. Only some matrix initializations. Removed this

mteb/_create_dataloaders.py
Samoed: Compatible, but we pass here `list[dict[str, Any]]` with any key because of that type can't be checked

Samoed:
> Should we just do a quick speed comparison to make sure that things don't get slower?

I tested and running didn't get slower

mteb/_create_dataloaders.py
KennethEnevoldsen: hmm would be nice to make it checkable in some way would we wrap it in the type?

Otherwise I would probably cast it before the return instead of the type ignore

Samoed: Changed to cast. Wrapping in type would be a bit hard, because for creating typed dict names of should be explicit

PR #4071, opened 2026-02-08
feat: save and load experiments results

Samoed:
Allow saving experiments in the results.


Example:
```python
import mteb

task = mteb.get_task("HUMEEmotionClassification")
cache = mteb.ResultCache("test_exp")
mteb.evaluate(model, task, cache=cache)

model = mteb.get_model("baseline/random-encoder-baseline", a="test")
mteb.evaluate(model, task, cache=cache)

model = mteb.get_model("baseline/random-encoder-baseline", a="test", b="test2")
mteb.evaluate(model, task, cache=cache)
```
Will produce
```
test_exp/                                                                                                                                                                                                                                                                         
└── results
    └── baseline__random-encoder-baseline
        └── 1
            ├── HUMEEmotionClassification.json
            ├── experiments
            │   ├── a_test
            │   │   ├── HUMEEmotionClassification.json
            │   │   └── model_meta.json
            │   └── a_test__b_test2
            │       ├── HUMEEmotionClassification.json
            │       └── model_meta.json
            └── model_meta.json

```

Example loading results:
```python
import mteb
cache = mteb.ResultCache("test_exp")
base_res = cache.load_results()
exp_res = cache.load_results(load_experiments=True)
exp_res_named = cache.load_results(load_experiments=True, experiment_names=["a_test"])

base_res.model_results
# [ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, task_results=[...](#2))]
exp_res.model_results
# [ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, task_results=[...](#2)), ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, experiment_name=a_test__b_test2, task_results=[...](#2)), ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, experiment_name=a_test, task_results=[...](#2))]
exp_res_named.model_results
# [ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, task_results=[...](#2)), ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, experiment_name=a_test, task_results=[...](#2))]
```

Ref https://github.com/embeddings-benchmark/mteb/issues/1211

### DIFF
--- a/tests/test_evaluate.py
+++ b/tests/test_evaluate.py
@@ -314,7 +314,7 @@ def test_evaluate_experiment(tmp_path):
 
     expected_path = (
         Path("results")
-        / "baseline__random-encoder-baseline"
+        / "mteb__baseline-random-encoder"
         / "1"
         / "experiments"
         / "test_param_123__test_param2_abc"


mteb/models/model_meta.py
```diff
@@ -131,6 +132,7 @@ class ModelMeta(BaseModel):
         model_type: A list of strings representing the type of model.
         modalities: A list of strings representing the modalities the model supports. Default is ["text"].
         contacts: The people to contact in case of a problem in the model, preferably a GitHub handle.
+        experiment_params: **DO NOT SET** A dictionary of parameters used in the experiment that are not covered by other fields. This is used to create experiment names for ablation studies and similar experiments.
```
KennethEnevoldsen: Set it as a private attribute instead?

mteb/models/model_meta.py
```diff
@@ -302,6 +307,16 @@ def model_name_as_path(self) -> str:
             raise ValueError("Model name is not set")
         return self.name.replace("/", "__").replace(" ", "_")
 
+    @property
+    def experiment_name(self) -> str | None:
+        """Create a filesystem-safe string representation of the experiment parameters.
+
+        Uses deterministic serialization and hashing to ensure stable, bounded output.
+        """
```
KennethEnevoldsen, line 323: can you add an example in the docstrings?

tests/test_evaluate.py
KennethEnevoldsen, line 1: What about stuff like batch size?

tests/test_result_cache.py
```diff
@@ -265,6 +265,62 @@ def test_cache_load_different_subsets():
     assert result2.model_results[0].task_results[0].get_score() == 0.01035
 
 
+def test_load_experiment_results(tmp_path):
```
KennethEnevoldsen, line 268: I think I have a few things here:

1) What do we want to happen when I run:
```
model = mteb.get_model(
        "baseline/random-encoder-baseline",
        a="test",
        b="test2",
    )

results = cache.load_results(model.mteb_model_meta)
```
I assume it should be only the results for that model experiment (though if batch size is an experiment, then we have a problem, similar with a few other utility parameters)

1.1)  What happens here if `load_experiments=True`? I would say it is ignored, but a warning is displayed indicating that you are attempting to load an experiment (to disable set `load_experiments=True`

2) I am a bit unsure about the `experiment_names` as it sets two ways of doing the same thing. I think I prefer:

```
model = mteb.get_model(
        "baseline/random-encoder-baseline",
        a="test",
        b="test2",
    )

results = cache.load_results(model)
```

over:

```
results = cache.load_results(model, experiment_names=["a_test__b_test2"])

Comment (KennethEnevoldsen) [reply tests/test_result_cache.py:268]: For the utility parameters, we can probably just define a set that we will ignore as "not" experiments.

We might want to do that only on the loading side, though

File: mteb/models/model_meta.py (hunk 1 above)
Comment (Samoed) [reply mteb/models/model_meta.py]: It was private in some iteration, but if we make it private, then pydantic won't load it from json

File: tests/test_evaluate.py
Comment (Samoed) [reply tests/test_evaluate.py:1]: Batch size should be passed to evaluate, not to `get_model`

File: tests/test_result_cache.py (hunk 3 above)
Comment (Samoed) [reply tests/test_result_cache.py:268]: > I assume it should be only the results for that model experiment 

I agree that this behavior is better. Will switch to this

>  What happens here if load_experiments=True? I would say it is ignored

Yes, ignored

> but a warning is displayed indicating that you are attempting to load an experiment

I didn't add warning, not sure what you're referring to

> I am a bit unsure about the experiment_names as it sets two ways of doing the same thing. I think I prefer:

Your method would work only if you want to take experiments for only 1 model, but if you want to load for multiple models (for example, compare MRL/quantization performance for multiple models) this method won't work

**Comment by Samoed** [comment]
Changed load experiments from bool to 
```python
class LoadExperimentEnum(HelpfulStrEnum):
    """Enum to specify whether to load experiments when loading results from the cache.

    Attributes:
        ALWAYS: Will always attempt to load experiments, if available.
        NEVER: Will never attempt to load experiments, even if they are available.
        ONLY_EXPERIMENTS: Will only load experiments, and ignore non-experiment results.
    """

    ALWAYS = "always"
    NEVER = "never"
    ONLY_EXPERIMENTS = "only-experiments"

```

mteb/models/model_meta.py
KennethEnevoldsen: hmm, I just seems wierd to ask people not to set an attribute

tests/test_evaluate.py
KennethEnevoldsen: right - "device" then

tests/test_result_cache.py
KennethEnevoldsen: > I didn't add warning, not sure what you're referring to

This part here seems contradictory:
```
model = mteb.get_model(
        "baseline/random-encoder-baseline",
        a="test",
    )

results = cache.load_results(
  model, 
  load_experiments="never", # default
)
```

Essentially we are saying load an experiments, but also not to. 

Also for backward compatibility, this would previously have loaded the results from just the random-encoder-baseline, which now is no longer loaded, which I would like to make the user aware of.

> Your method would work only if you want to take experiments for only 1 model, but if you want to load for multiple models (for example, compare MRL/quantization performance for multiple models) this method won't work

Hmm but then I would just supply multiple model metas.

```
models = [mteb.get_model(name, dtype="int8") for name in model_names]
results = cache.load_results(models)
```

The problem with `a_test__b_test2` is that you have to know the serialization.

tests/test_evaluate.py
Samoed: `device` is explicitly added to `load_model`, so it won't be in experiment https://github.com/embeddings-benchmark/mteb/blob/50bd0fa8b7bebc51ec7f05f1dff02c447f541e49/mteb/models/model_meta.py#L277

mteb/models/model_meta.py
Samoed: I agree, but I can't find other solutions. This parameter shouldn't be set only when users adding `ModelMeta` and we have control for this

tests/test_result_cache.py
Samoed: > Essentially we are saying load an experiments, but also not to.

Yeah, this weird. For now, I made that `load_experiments` will be ignored. Maybe we can load experiments only if metadata is matching

> Also for backward compatibility, this would previously have loaded the results from just the random-encoder-baseline, which now is no longer loaded, which I would like to make the user aware of.

Agree. How we can do this? Just warning message?

> The problem with a_test__b_test2 is that you have to know the serialization.

I agree. We can add an option to pass parameters and convert them, but maybe removing this argument would be easier solution

KennethEnevoldsen: yeah "if-matching" sounds like a good category.

> Agree. How we can do this? Just warning message?

Of course, a warning message is not fully backward compatible, but it is at least better than nothing.

Alternatively, we could simply have `load_experiments="never"` as the default, and that will be backward compatible - then during eval we have change it to "if-matching" which is approximately backward compatible 

Another way it also to rework it to `loading_strategy:
- matching-revison: loads everything with the same revision (default, current behaviour)
- matching-kwargs: only with matching kwargs
- matching-name: also include other revisions of the same model

tests/test_evaluate.py
KennethEnevoldsen: Ahh great then I think we are basically good

tests/test_result_cache.py
Samoed: > Of course, a warning message is not fully backward compatible, but it is at least better than nothing.

There is already warning if models are `ModelMeta` that `load_experiments` will be ignored

> Another way it also to rework it to `loading_strategy:

I don't think we should split matching revision, name, kwargs, because this is too much complicated.

> then during eval we have change it to "if-matching" which is approximately backward compatible

Not sure when we want to change this

I think current implementation with different strategies that would work only when models would be passed as names if good enough

Samoed:
I can't reproduce test error locally

tests/test_result_cache.py
KennethEnevoldsen: > I don't think we should split matching revision, name, kwargs, because this is too much complicated.

what about just doing:
```
matching-revison: loads everything with the same revision (default, current behaviour)
matching-kwargs: only with matching kwargs
```

This is basically the current load_experiments, but reworked in a way that is backwards compatible.

Samoed: I think we can add like this. Should this be added in `LoadExperiment` or craete other parameter (I don't think so, but still)? Because these mathing kwargs will be supported when passing ModelMeta

KennethEnevoldsen: but currently we loose backward compatibility: 
```
model = mteb.get_model(
        "baseline/random-encoder-baseline",
        a="test",
    )

results = cache.load_results(
  model, 
  load_experiments="never", # default
)
```
Would load result now, but not after. I think we can avoid that

```
model = mteb.get_model(
        "baseline/random-encoder-baseline",
        a="test",
    )

results = cache.load_results(
  model, 
  loading_strategy="matching-revison" # default - backward compatible
  # load_experiments="never", # this is replaced by loading_strategy
)
```

Here is the mapping:

ONLY_EXPERIMENTS: I would remove not sure it makes sense
ALWAYS: I assume this is `loading_strategy.matching-revison`?
NEVER: I assume this is to ensure that we only get default for: "baseline/random-encoder-baseline", here I would instead add `loading_strategy.only-default`

We can still use the load_experiments naming, but the categories I want are:

default / backward compatible / all results for a specific model rev: It is possible so I think we should not break things we don't have to.
matching kwargs: for specific experiments - this is just using modelmeta
only default / no experiment: To only examine defaults.

Samoed: I mean that `matching-revision` and `match-experiments` are only ModelMeta related, but what should be when only model names as strings? I think current params `LoadExperiment` are good when str are passed, but matching kwargs/experiments are for modelmeta loading. Probably we can load experiments only when ModelMeta is passed

KennethEnevoldsen: Right, but I don't think we want two - I think we can phrase it in one.

> Probably we can load experiments only when ModelMeta is passed

but this would still break backward compatibility


Here is my suggestion
```py
class LoadExperimentEnum(HelpfulStrEnum):
    """Enum to specify whether to load experiments when loading results from the cache.

    Attributes:
        MATCH_NAME: Will load everything that matches the name of the model, including experiments. If a revision is supplied using`ModelMeta` the this will also match the revision.
        MATCH_KWARGS: Will load experiments that match the keyword arguments supplied in the`ModelMeta`. Assumes a `ModelMeta`s are supplied.
        ONLY_DEFAULT: Will only load models with default keyword arguments, meaning that it will not include any experiments.
    """

    MATCH_NAME = "match-name" # default
    MATCH_KWARGS = "match-kwargs"
    ONLY_DEFAULT = "only-default" # could be renamed to "no-experiments"
```

Alternatively we could allow `match_kwargs` to load using `cache.load_results(..., experiment_kwargs={"a": "test"}`, in which case the text would be:

```
MATCH_KWARGS: Will load experiments that match the keyword arguments (kwargs) supplied in the`ModelMeta` and `experiment_kwargs`. If both a `ModelMeta` and `experiment_kwargs` is supplied the `experiment_kwargs` will be ignored.
```
The problem here is of course that you use ModelMeta to denote revision.

Samoed: I think this is similar to what is existing now. What should be in case metadata passing? Ignoring `LoadExperimentEnum`?

KennethEnevoldsen: No this one interact with the passed metadata:

"match_name" ignores passed kwargs

Samoed:
I think implementation is ready, and I added a section in what's new

Samoed:
@KennethEnevoldsen can you review this once again?

docs/whats_new.md
```diff
@@ -2,7 +2,75 @@
 
 This section is an overview of releases for more information check out the autogenerated [changelog](https://github.com/embeddings-benchmark/mteb/blob/main/pyproject.toml).
 
+## New in v2.9 (Experiments)
```
```suggestion
## New in v2.9
```

docs/whats_new.md
```diff
@@ -2,7 +2,75 @@
...
+```python
+import mteb
+
+# base run
+model = mteb.get_model("baseline/random-encoder-baseline")
+task = mteb.get_task("HUMEEmotionClassification")
+mteb.evaluate(model, task)
+
+# experiment with different hyperparameters
+model = mteb.get_model("baseline/random-encoder-baseline", a="test")
```
KennethEnevoldsen: maybe we should do it with an actual hyperparameter (we could easily make embed-dim configurable). Make the example for understandable

we can also just use a model with a hyperparameter

docs/whats_new.md
```diff
@@ -2,7 +2,75 @@
...
+
+cache = mteb.ResultCache()
+
+# load default results
+results = cache.load_results()
+print(results.model_results)
+# [ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, task_results=[...](#1))]
+
+# getting results with specific experiment
+model_meta = mteb.get_model_meta("baseline/random-encoder-baseline", experiment_params={"a": "test"})
```
KennethEnevoldsen: should we rename to `experiment_kwargs`? (initialization_kwargs is also an option)

docs/whats_new.md
```diff
@@ -2,7 +2,75 @@
...
+print(results_without_experiment.model_results)
+# [ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, task_results=[...](#1))]
+
+# load all results for specific model
+results_without_experiment = cache.load_results(models=["baseline/random-encoder-baseline"], load_experiments="match_name")
+print(results_without_experiment.model_results)
+# [ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, task_results=[...](#1)), ModelResult(model_name=baseline/random-encoder-baseline, model_revision=1, experiment_name=a_test, task_results=[...](#1))]
+
+# only one result will be loaded as experiment name doesn't match, so it will load the default one
+results_without_experiment = cache.load_results(models=["baseline/random-encoder-baseline"], load_experiments="match_kwargs")
```
KennethEnevoldsen: I would say this make more sense if the you supply some kwargs

mteb/models/model_meta.py
```suggestion
        experiment_params: A dictionary of parameters used in the experiment that are not covered by other fields. This is used to create experiment names for ablation studies and similar experiments.
```
I think this would work. I would also add a test to ensure that it is not set in any of our model metas

mteb/models/model_meta.py
```diff
@@ -302,6 +311,23 @@ def model_name_as_path(self) -> str:
...
 
+    @property
+    def experiment_name(self) -> str | None:
+        """Create a filesystem-safe string representation of the experiment parameters.
+
+        Uses deterministic serialization and hashing to ensure stable, bounded output.
+
+        Examples:
+            >>> import mteb
+            >>> model = mteb.get_model("baseline/random-encoder-baseline", param1="test")
```
KennethEnevoldsen: Just to confirm want to model name on the leaderboard to be:

```
baseline/random-encoder-baseline (param1=test)
```

Right?

This property is intended for building paths - do we want it to be public?

KennethEnevoldsen: Should we also create a property to the other name?

mteb/models/model_meta.py
```diff
@@ -1022,3 +1052,59 @@ def _repo_exists(repo_id: str, repo_type: str | None = None) -> bool:
     except HFValidationError as e:
         logger.warning(f"Can't check existence of {repo_id}: {e}")
         return False
+
+
+def _get_experiment_name_from_params(
```
```suggestion
def _serialize_experiment_kwargs_to_name(
```
or
```suggestion
def _serialize_experiment_name_from_kwargs(
```

docs/whats_new.md
Samoed: Point of this example if we add `match_kwargs` without kwargs then default meta will return

docs/whats_new.md
Samoed: Changed

mteb/models/model_meta.py
Samoed: Created `model_name_with_experiment`

> Right?

I think so, but for now, I don't add integration to leaderboard. We can add this later

KennethEnevoldsen: agree!

mteb/models/model_implementations/random_baseline.py
```diff
@@ -155,10 +155,11 @@ def __init__(
         revision: str | None,
         array_framework: Literal["numpy", "torch"] = "numpy",
         dtype: torch.dtype | np.floating = np.float32,
+        truncate_dim: int = _EMBEDDING_DIM,
```
KennethEnevoldsen: don't we call this "embed_dim=" for other models (e.g. `voyageai/voyage-4-large (embed_dim=2048)`)
```suggestion
        embed_dim: int = _EMBEDDING_DIM,
```

Samoed: I added as in sentence transformers, but can rename https://github.com/huggingface/sentence-transformers/blob/73d700b8431cdcda04f2f0efbb156ea06d81ae90/sentence_transformers/SentenceTransformer.py#L91, but agree, let's add with embed_dim

PR #4082, opened 2026-02-10
docs: Outline for adding a task documentation

### LINKED ISSUE #4077: Improve documentation when adding tasks

Model PR generally seems be doing fine, but with our task PR, I generally get quite a few questions on how to set up a task. We could probably spell it out more, with a few more examples across tasks.

KennethEnevoldsen:
This is a suggested structure, PR is just to get feedback before I finish it up.

fixes #4077

@yevhenkost if you have anything that you feel is missing that I haven't addressed here let me know (otherwise we can add it later)

### DIFF
--- a/docs/contributing/adding_a_dataset.md
+++ b/docs/contributing/adding_a_dataset.md
@@ -283,42 +283,116 @@ Once we have decided on task, we can implement them as follows:
     While we do not recommend overwriting `load_data` it can often be useful, when developing tasks and can also be used in conjunction with
     [`push_dataset_to_hub`][mteb.AbsTask.push_dataset_to_hub] to push datasets to the hub in the correct format.
 
-    ```python
-    import mteb
-    from mteb.abstasks import AbsTaskClassification
+    === "Classification"
+         For our classification task we use a dummy dataset consisting of two samples with labels 0 and 1. We can implement the task as follows:
+        ```python
+        import mteb
+        from mteb.abstasks import AbsTaskClassification
 
 
-    class MyClassificationtask(AbsTaskClassification):
-        metadata = mteb.TaskMetadata(  # minimal metadata
-            name="MyClassificationTask",
-            description="A dummy classification just for testing",
-            main_score="accuracy",
-            eval_langs=["eng-Latn"],
-            eval_splits=["test"],
-            type="Classification",
-            dataset={"path": "na", "revision": "na"},
-        )
+        class MyClassificationtask(AbsTaskClassification):
+            metadata = mteb.TaskMetadata(  # minimal metadata
+                name="MyClassificationTask",
+                description="A dummy classification just for testing",
+                main_score="accuracy",
+                eval_langs=["eng-Latn"],
+                eval_splits=["test"],
+                type="Classification",
+                dataset={"path": "na", "revision": "na"},
+            )
 
-        label_column_name = "label"
-        input_column_name = "text"
+            label_column_name = "label"
+            input_column_name = "text"
 
-        def load_data(self, num_proc=None, **kwargs) -> None:
-            self.dataset = {
-                "test": [
-                    {"text": "sample text 1", "label": 1},
-                    {"text": "sample text 2", "label": 0},
-                ]
-            }
-            self.data_loaded = True
-    ```
+            def load_data(self, num_proc=None, **kwargs) -> None:
+                self.dataset = {
+                    "test": [
+                        {"text": "sample text 1", "label": 1},
+                        {"text": "sample text 2", "label": 0},
+                    ]
+                }
+                self.data_loaded = True
+        ```
+        which can then be run as follows:
+        ```py
+        task = MyClassificationtask()
+        mdl = mteb.get_model("baseline/random-encoder-baseline")
+        results = mteb.evaluate(mdl, task)
+        ```
+    === "Retrieval"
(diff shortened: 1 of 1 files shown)


Samoed:
I updated links to some classes and make docs buildable

docs/contributing/adding_a_dataset.md
```diff
@@ -2,118 +2,197 @@
...
+        # some processing
+        ...
+```
+
+### Select an appropriate task
+
+To add a dataset you first need to figure out which type of task is the best suited for the dataset. Below we will give you an overview of the most common,
+but do see [abstasks](../api/task.md#multimodal-tasks) for an overview of all the tasks available.
+
+| Task                     | Abstask                                                                       | Description                                                                                                                                                                                                                                                                                                                                                                                                                                  | Default Metric                                                                                      |
```
Samoed: 1. I think we can rename `Default Metric` to `Most common metric`, because we don't have "default" metrics
2. Should we add column about dataset format? How many columns etc. Mostly we have this information in AbsTask docstrings. Overall I think users just copy different classes and don't see docs much, so this is minor

docs/contributing/adding_a_dataset.md
```diff
@@ -2,118 +2,197 @@
...
+
+    We can then implement the task as follows:
+
+    ```py
+    import mteb
+    from mteb.abstasks import AbsTaskClassification
+
+
+    class MyClassificationtask(AbsTaskClassification):
+        metadata = mteb.TaskMetadata(  # minimal metadata
```
Samoed, line 81: Should we fill all values? Domains, subcategory etc.? Otherwise, users could submit without these fields

docs/contributing/adding_a_dataset.md
```diff
@@ -2,118 +2,197 @@
...
+| Semantic Text Similarity | [`AbsTaskSTS`][mteb.abstasks.sts.AbsTaskSTS]                                  | Compares the (semantic) similarity of pairs of documents. The goal is to embed documents such that semanticly similar statements appear close.                                                                                                                                                                                                                                                                                               | [Spearman](https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient)                 |
+
+!!! Note
+    While each task has a default score we compute multiple and it is possible to select any of these are the default metric for your task.
```
Samoed:
```suggestion
    While each task has a main score we compute multiple and it is possible to select any of these are the main metric for your task.
```

docs/contributing/adding_a_dataset.md
```diff
@@ -2,118 +2,197 @@
...
 
-??? example "DBpediaClassificationV2 Task"
+    MISSING
+
+=== "Semantic Similarity"
+
+    MISSING
+
+
+??? example "Overwriting `load_data`"
```
Samoed, line 281: I think we already have example in section `Implementing the task`

docs/contributing/adding_a_dataset.md
KennethEnevoldsen: I think this is mostly a quick overview (though I see that some of the docs on the individual abstasks can be improved - e.g. few of them describe the task at a high level)

change to "common metric"

docs/contributing/adding_a_dataset.md
KennethEnevoldsen: for the initial (and local) implementation I think it is good to have a minimal one - but I added a section on filling out the task metadata

docs/contributing/adding_a_dataset.md
KennethEnevoldsen: There is an overview and the one that overwrites the dataset_transform.

KennethEnevoldsen:
@Samoed saw that we both made some edits at the same time here - I overwrite most of yours with mine. I especially wanted the code to be runnable as-is so avoided adding any `...`.

Samoed:
In my changes, I add custom load_data for retrieval task to show how it different. Maybe it would be better to keep?

docs/contributing/adding_a_dataset.md
```diff
@@ -2,127 +2,403 @@
...
+    # {'id': '51829', 'title': 'How can show android tablet as a external storage to PC?', 'text': "I want to send files to android tablet ..."}
+    print(test_set["queries"][0])  # print the first query
+    # {'id': '11546', 'text': 'Android chroot ubuntu - is it possible to get ubuntu to recognise usb devices'}
+    test_set["relevant_docs"]["11546"]  # print the relevant documents for the first query
+    # {'18572': 1}
+
+    # ensure that we can evaluate the model on the task
+    mdl = mteb.get_model("baseline/random-encoder-baseline")
+    results = mteb.evaluate(mdl, task)
+    print(results[0].get_score())  # print the spearman score of the random baseline
```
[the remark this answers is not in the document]
KennethEnevoldsen:
```suggestion
    print(results[0].get_score())  # print the score of the random baseline
```

Samoed:
Should we add custom `load_data` for retrieval task because retrieval tasks have different format? I've added it in https://github.com/embeddings-benchmark/mteb/pull/4082/changes/5e0186a9935d64fbd16ce8ec090efcc7ab7c262e, but it was removed in merges. Maybe this would be nice to have too

KennethEnevoldsen:
Yeah I think you are right. I just wanted it to be runnable as is, but I think we should add a `load_data`

Samoed:
Updated example with `load_data`

image

PR #4084, opened 2026-02-11
model: Add Bayesian BM25 (bb25) baseline model

### LINKED ISSUE #4072: Add model: bb25

NA

### Arxiv link

https://zenodo.org/records/18514855

### Description of the model

Comparison with bm25: https://github.com/instructkr/bb25/pull/1

Also seems to work well for hybrid approaches

jaepil:
## Summary

Closes #4072

Add `baseline/bb25`, a Bayesian BM25 model that transforms BM25 scores into calibrated probability estimates in [0, 1] through Bayesian inference with a sigmoid likelihood model.

- Uses `bm25s` as the BM25 backend (same as `baseline/bm25s`)
- Default `prior_weight=0.0`: sigmoid likelihood is strictly monotonic, so BM25 rankings are preserved exactly while scores become calibrated probabilities suitable for hybrid search score fusion
- Setting `prior_weight > 0` enables the Composite Prior (Section 4.2 of the paper), which re-ranks based on document-level evidence (query term frequency + document length)

### NDCG@10 Comparison (11 Retrieval Tasks, prior_weight=0.0)

| Task | bm25s | bb25 | diff |
|------|------:|-----:|-----:|
| ArguAna | 0.49233 | 0.49233 | 0.000% |
| CQADupstackGamingRetrieval | 0.48474 | 0.48498 | +0.050% |
| CQADupstackUnixRetrieval | 0.29195 | 0.29195 | 0.000% |
| ClimateFEVERHardNegatives | 0.14019 | 0.14019 | 0.000% |
| DBPediaHardNegatives | 0.30270 | 0.30270 | 0.000% |
| FEVERHardNegatives | 0.51993 | 0.51993 | 0.000% |
| FiQA2018 | 0.25139 | 0.25139 | 0.000% |
| HotpotQAHardNegatives | 0.58882 | 0.58882 | 0.000% |
| NFCorpus | 0.32102 | 0.32102 | 0.000% |
| SciFact | 0.68630 | 0.68597 | -0.048% |
| Touche2020 | 0.32922 | 0.32922 | 0.000% |
| **Average** | **0.40078** | **0.40077** | **-0.002%** |

9/11 tasks produce exactly identical scores. The two minor differences (CQADupstackGaming +0.050%, SciFact -0.048%) are caused by floating-point tie-breaking when the sigmoid maps very close BM25 scores to the same float64 value.

### Changes

- `mteb/models/model_implementations/bb25.py` -- Bayesian BM25 model implementation (`model_type=["sparse"]`)
- `tests/test_models/test_bm25.py` -- e2e test on NFCorpus
- `tests/test_models/test_model_meta.py` -- add `baseline/bb25` to `_HISTORIC_MODELS`

### References

- [Bayesian BM25: A Probabilistic Framework for Hybrid Text and Vector Search](https://doi.org/10.5281/zenodo.18414941)
- [From Bayesian Inference to Neural Computation](https://doi.org/10.5281/zenodo.18512411)


tests/test_models/test_bb25.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+import pytest
+
+import mteb
+
+
+def test_bb25_matches_bm25s():
```
Samoed: Maybe would be better to reuse `test_bm25s_e2e` test

mteb/models/model_implementations/bb25.py
```diff
@@ -0,0 +1,311 @@
...
+            *,
+            task_metadata: TaskMetadata,
+            hf_split: str,
+            hf_subset: str,
+            encode_kwargs: EncodeKwargs,
+            num_proc: int | None = None,
+        ) -> None:
+            logger.info("Encoding Corpus...")
+            corpus_texts = [
+                "\n".join([doc.get("title", ""), doc["text"]]) for doc in corpus
```
Samoed, line 125: Seems that we have bug in `bm25` implementation, because `text` already includes `title`
```suggestion
                doc["text"] for doc in corpus
```

mteb/models/model_implementations/bb25.py
```diff
@@ -0,0 +1,311 @@
...
+            self.k1 = k1
+            self.b = b
+            self.alpha = alpha
+            self.prior_weight = prior_weight
+            self.stopwords = stopwords
+            self.stemmer = (
+                Stemmer.Stemmer(stemmer_language) if stemmer_language else None
+            )
+
+        def encode(self, texts: list[str]):
```
Samoed: I see that in `bm25` wrapper have same function, but I think would be better to rename to `_encode` to model not compatible with `EncodeWrapper`

mteb/models/model_implementations/bb25.py
jaepil: Thanks for the review! I checked the NFCorpus dataset -- `title` and `text` are separate fields, and `title` is not included in `text`:

```
columns: ['id', 'title', 'text']
docs with title: 3633/3633
docs where title in text: 2/3633 (coincidental matches)
```

The `"\n".join([doc.get("title", ""), doc["text"]])` pattern is intentional and identical to the existing `bm25s` implementation (`bm25.py:61`). Removing `title` would lose document title information for retrieval scoring.

mteb/models/model_implementations/bb25.py
jaepil: Good point. I renamed `encode` to `_encode` in both `bb25.py` and `bm25.py` to avoid confusion with `EncoderProtocol.encode()`.

tests/test_models/test_bb25.py
jaepil: Done. Merged into a parametrized `test_bm25_e2e` in `test_bm25.py` and removed `test_bb25.py`.

mteb/models/model_implementations/bb25.py
Samoed: Ah, yes. Here dataset passed directly without creating dataloader. Then this is correct

Samoed (approved):

jaepil:
Regarding the default configuration:

I have intentionally configured the default settings to strictly mirror the standard BM25 ranking behavior while providing probabilistic scores in [0, 1] for hybrid score fusion.

My goal was to provide a controlled baseline where the only variable is the output scale, ensuring full compatibility with existing research.

Separately, the model is capable of achieving higher performance with different configurations.

If MTEB wants to track an optimized setting in addition to the baseline, I can add it in a follow-up PR as a separate variant.

KennethEnevoldsen (approved):
Looks great

> If MTEB wants to track an optimized setting in addition to the baseline, I can add it in a follow-up PR as a separate variant.

I would love to see the optimized setting as well. Is there a reason why the optimized setting is not the default setting?

PR #4085, opened 2026-02-11
fix: Deprecate GritLM Wrapper and use Sentence Transformers

### LINKED ISSUE #3517: Wrong instructions for multilingual-e5-large-instruct
### Describe the bug

It seems that  multilingual-e5-large-instruct uses the query template to encode the corpus. However, it should be empty according to https://huggingface.co/intfloat/multilingual-e5-large-instruct.

Image

### To reproduce

```bash
mteb run \
     -m intfloat/multilingual-e5-large-instruct \
     -t "Touche2020-NL" \
     --output-folder results

```

### Additional information

_No response_

No

ayush1298:
closes #3517 

Change instruct_wrapper(GritLM Wrapper) to InstructSentenceTransformerModel. The list of models with these change are as follow: 

1. intfloat/multilingual-e5-large-instruct
2. intfloat/e5-mistral-7b-instruct
3. zeta-alpha-ai/Zeta-Alpha-E5-Mistral
4. GritLM/GritLM-7B
5. GritLM/GritLM-8x7B
6. Alibaba-NLP/gte-Qwen2-7B-instruct
7. Alibaba-NLP/gte-Qwen1.5-7B-instruct
8. Alibaba-NLP/gte-Qwen2-1.5B-instruct
9. Linq-AI-Research/Linq-Embed-Mistral
10. Salesforce/SFR-Embedding-2_R
11. Salesforce/SFR-Embedding-Mistral
12. NovaSearch/stella_en_400M_v5
13. NovaSearch/stella_en_1.5B_v5

I have checked it by loading `intfloat/multilingual-e5-large-instruct` with a new wrapper.

### DIFF
--- a/mteb/models/instruct_wrapper.py
+++ b/mteb/models/instruct_wrapper.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import sys
 import warnings
 from typing import TYPE_CHECKING, Any
 
@@ -19,10 +20,19 @@ if TYPE_CHECKING:
     from mteb.abstasks.task_metadata import TaskMetadata
     from mteb.types import Array, BatchedInput
 
+if sys.version_info >= (3, 13):
+    from warnings import deprecated
+else:
+    from typing_extensions import deprecated
+
 
 logger = logging.getLogger(__name__)
 
 
+@deprecated(
+    "`instruct_wrapper` is deprecated and will be removed in future versions. "
+    "Please use `InstructSentenceTransformerModel` instead."
+)
 def instruct_wrapper(
     model_name_or_path: str,
     mode: str,
@@ -52,6 +62,10 @@ def instruct_wrapper(
     )
     from gritlm import GritLM  # type: ignore[import]
 
+    @deprecated(
+        "`instruct_wrapper` is deprecated and will be removed in future versions. "
+        "Please use `InstructSentenceTransformerModel` instead."
+    )
     class InstructGritLMModel(GritLM, AbsEncoder):
         def __init__(
             self,


ayush1298:
> `torch_dtype` is deprecated! Use `dtype` instead!

So, I have changed it at all places to avoid warning.

mteb/models/model_implementations/e5_instruct.py
```diff
@@ -130,16 +121,12 @@
...
         instruction_template=E5_INSTRUCTION,
-        attn="cccc",
-        pooling_method="lasttoken",
-        mode="embedding",
-        torch_dtype=torch.bfloat16,
+        model_kwargs={"dtype": torch.bfloat16},
         # The ST script does not normalize while the HF one does so unclear what to do
         # https://huggingface.co/intfloat/e5-mistral-7b-instruct#transformers
-        normalized=True,
     ),
```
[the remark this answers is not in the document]
ayush1298: @Samoed Will we have this problem?

mteb/models/model_implementations/gritlm_models.py
```diff
@@ -31,11 +31,10 @@ def gritlm_instruction(instruction: str = "", prompt_type=None) -> str:
 
 
 gritlm7b = ModelMeta(
-    loader=instruct_wrapper,
+    loader=InstructSentenceTransformerModel,
```
Samoed, line 34: For gritlm models you should pass `include_prompt=False` https://github.com/huggingface/sentence-transformers/blob/d0df0cccfc5564ed24092ae3fe24290060a0f96e/sentence_transformers/models/Pooling.py#L73 like https://github.com/huggingface/sentence-transformers/blob/d0df0cccfc5564ed24092ae3fe24290060a0f96e/sentence_transformers/SentenceTransformer.py#L388-L403 but I don't see how we can explicitly pass this to pooler https://github.com/huggingface/sentence-transformers/blob/d0df0cccfc5564ed24092ae3fe24290060a0f96e/sentence_transformers/SentenceTransformer.py#L2112-L2123

ayush1298: okay, will change these. Should we also pass `pooling_mode`?

Samoed: I don't see where you want to pass `pooling_mode`. Probably the easiest solution for gritlm would be explicitly pass `Transformer` and `Pooling` to `SentenceTransformer` (or call `self.set_pooling_include_prompt(include_prompt=False)` after model loading

ayush1298: So, we can add `include_prompt` in `__init__` of `InstructSentenceTransformerModel`, and then add below thing after https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/instruct_wrapper.py#L181: 

```
if not include_prompt:
    self.model.set_pooling_include_prompt(include_prompt=False)
```

Samoed: We can do like this

ayush1298: Do we have to also somehow handle `normalized=True` as GritLM was specifically normalizing embeddings

Samoed: This can be passed to `SentenceTransformers.encode` https://github.com/huggingface/sentence-transformers/blob/d0df0cccfc5564ed24092ae3fe24290060a0f96e/sentence_transformers/SentenceTransformer.py#L869

ayush1298: 1 more thing, Is it correct that whenever the `pooling_method` is present in loader_kwargs, we don't have to add `include_prompt=False` irrespective of type of `pooling_method` like `lasttoken, mean, cls`?

Samoed: `pooling_method` won't be used in sentence transformers and should be removed. I checked all models with `pooling_method` and they all have correct config for sentence transformers

ayush1298: For models under https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_implementations/salesforce_models.py, I think instead of changing them to `InstructSentenceTransformerModel`, should we change them to `sentence_transformers_loader`, as currently, I am not able to run `Salesforce/SFR-Embedding-Code-2B_R` as it requires  `trust_remote_code=True`, which I think `InstructSentenceTransformerModel` doesn't allow.

Samoed: `InstructSentenceTransformerModel` support `trsut_remote_code`. You just should it to `loader_keargs`

ayush1298: Oh, okay. I just missed that they support it.

Samoed: It not explicitly added to `init`, but all kwargs are passed to `SentenceTransformer`, so this won't be a problem

ayush1298: Yes, will just add them to loader

mteb/models/model_implementations/e5_instruct.py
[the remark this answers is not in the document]
Samoed: I think we can just add `apply_instruction_to_passages=False`

mteb/models/model_implementations/e5_instruct.py
```diff
@@ -75,16 +70,12 @@
...
+    loader=InstructSentenceTransformerModel,
     loader_kwargs=dict(
         instruction_template=E5_INSTRUCTION,
-        attn="cccc",
-        pooling_method="lasttoken",
-        mode="embedding",
-        torch_dtype=torch.float16,
+        model_kwargs={"dtype": torch.float16},
         # The ST script does not normalize while the HF one does so unclear what to do
         # https://huggingface.co/intfloat/e5-mistral-7b-instruct#transformers
```
Samoed:
```suggestion
```

mteb/models/model_implementations/gritlm_models.py
ayush1298: > This can be passed to `SentenceTransformers.encode` https://github.com/huggingface/sentence-transformers/blob/d0df0cccfc5564ed24092ae3fe24290060a0f96e/sentence_transformers/SentenceTransformer.py#L869

Mostly model has a Normalize module under its modules.json, so in that case, emebddings gets-autonormalized. But, if it's not then how exactly we have to pass it in that case? as we can't just pass to `loader_kwargs`

mteb/models/model_implementations/e5_instruct.py
```diff
@@ -75,16 +70,12 @@
...
 e5_mistral = ModelMeta(
-    loader=instruct_wrapper,
+    loader=InstructSentenceTransformerModel,
     loader_kwargs=dict(
         instruction_template=E5_INSTRUCTION,
-        attn="cccc",
-        pooling_method="lasttoken",
-        mode="embedding",
-        torch_dtype=torch.float16,
+        model_kwargs={"dtype": torch.float16},
```
Samoed, line 77: Here also should be added `apply_instruction_to_passages=False`

mteb/models/model_implementations/e5_instruct.py
```diff
@@ -130,16 +121,12 @@
...
-    loader=instruct_wrapper,
+    loader=InstructSentenceTransformerModel,
     loader_kwargs=dict(
         instruction_template=E5_INSTRUCTION,
-        attn="cccc",
-        pooling_method="lasttoken",
-        mode="embedding",
-        torch_dtype=torch.bfloat16,
+        model_kwargs={"dtype": torch.bfloat16},
         # The ST script does not normalize while the HF one does so unclear what to do
```
Samoed:
```suggestion
    apply_instruction_to_passages=False,
```

mteb/models/model_implementations/gritlm_models.py
Samoed: I think we can skip this for now, because most of times this normalization not make much difference

mteb/models/model_implementations/gte_models.py
```diff
@@ -29,17 +29,13 @@ def instruction_template(
...
-        attn="bbcc",
-        pooling_method="lasttoken",
-        mode="embedding",
-        torch_dtype=torch.float16,
         # The ST script does not normalize while the HF one does so unclear what to do
         # https://huggingface.co/Alibaba-NLP/gte-Qwen2-7B-instruct#sentence-transformers
-        normalized=True,
-        embed_eos="<|endoftext|>",
+        model_kwargs={"dtype": torch.float16},
+        add_eos_token=True,
```
Samoed, line 37: In their examples they're not including eos, so I'm not sure if we need to add it

mteb/models/model_implementations/gte_models.py
```diff
@@ -29,17 +29,13 @@ def instruction_template(
 """
 
 gte_qwen2_7b_instruct = ModelMeta(
-    loader=instruct_wrapper,
+    loader=InstructSentenceTransformerModel,
     loader_kwargs=dict(
         instruction_template=instruction_template,
```
Samoed, line 34:
```suggestion
        instruction_template=instruction_template,
        apply_instruction_to_passages=False,
```

mteb/models/model_implementations/linq_models.py
```diff
@@ -27,14 +27,10 @@ def instruction_template(
 
 
 Linq_Embed_Mistral = ModelMeta(
-    loader=instruct_wrapper,
+    loader=InstructSentenceTransformerModel,
     loader_kwargs=dict(
         instruction_template=instruction_template,
```
Samoed, line 32:
```suggestion
        instruction_template=instruction_template,
        apply_instruction_to_passages=False,
```

mteb/models/model_implementations/salesforce_models.py
```diff
@@ -40,16 +39,13 @@ def instruction_template(
...
         instruction_template=instruction_template,
-        attn="cccc",
-        pooling_method="lasttoken",
-        mode="embedding",
-        torch_dtype="auto",
         # The ST script does not normalize while the HF one does so unclear what to do
         # https://huggingface.co/Salesforce/SFR-Embedding-2_R
-        normalized=True,
+        model_kwargs={"dtype": "auto"},
+        trust_remote_code=True,
```
Samoed, line 47:
```suggestion
        trust_remote_code=True,
        apply_instruction_to_passages=False,
```

mteb/models/model_implementations/stella_models.py
```diff
@@ -51,12 +51,10 @@
...
     # https://huggingface.co/dunzhang/stella_en_400M_v5/discussions/21#671a6205ac1e2416090f2bf4
-    loader=instruct_wrapper,
+    loader=InstructSentenceTransformerModel,
     loader_kwargs=dict(
-        attn="cccc",
-        pooling_method="lasttoken",
-        mode="embedding",
-        torch_dtype="auto",
+        model_kwargs={"dtype": "auto"},
+        trust_remote_code=True,
```
Samoed, line 57:
```suggestion
        trust_remote_code=True,
        apply_instruction_to_passages=False,
```

mteb/models/model_implementations/e5_instruct.py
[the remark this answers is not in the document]
ayush1298: Why do we have to add it everywhere, not just for E5-instruct models?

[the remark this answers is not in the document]
Samoed: Because all models don't include it to the documents. You can see in their model cards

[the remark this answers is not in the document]
ayush1298: So, need to add for GritLM also, right?

mteb/models/model_implementations/gte_models.py
ayush1298: I thought, `embed_eos` was there already, so we should use `add_eos_token` in place of it

mteb/models/model_implementations/e5_instruct.py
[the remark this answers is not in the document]
Samoed: Only gritlm don't require this I think

ayush1298:
I have also updated the wrapper for all models mentioned [here](https://github.com/embeddings-benchmark/mteb/issues/3517#issuecomment-3883499105) and added them in the PR description.

KennethEnevoldsen (commented):
Can you run a few experiments to compare the scores?

mteb/models/instruct_wrapper.py
```diff
@@ -149,6 +150,7 @@ def __init__(
             add_eos_token: Whether to add the eos token to each input example.
             prompts_dict: Dictionary of task names to prompt names. If task name is missing in the dict or prompts dict is None, prompt from task metadata or
                 AbsTask.abstask_prompt will be used.
+            include_prompt: Whether to include the prompt in the input.
```
tomaarsen: So we're on the same page: this parameter determines whether the prompt will be included in the (mean) pooling, not whether the prompt is used at all. It's exclusively used here: https://github.com/huggingface/sentence-transformers/blob/d0df0cccfc5564ed24092ae3fe24290060a0f96e/sentence_transformers/models/Pooling.py#L142-L163

Where I mask away the prompt tokens in the `attention_mask` so that the mean pooling ignores those tokens.

Samoed: Yes. Gritlm uses same logic for attention masking as instructor models https://github.com/ContextualAI/gritlm/blob/971068105a8508bca421841c59fddba7f6596402/gritlm/gritlm.py#L144-L153

tomaarsen: Indeed, the implementation on ST was designed to match both INSTRUCTOR and GRIT, but the comment here in the docstring is confusing. Perhaps this is clearer:

```suggestion
            include_prompt: Whether to include the prompt tokens in the pooling.
```

Samoed: Agree, that's better

ayush1298:
> Can you run a few experiments to compare the scores?

okay, will try to run for some small models

mteb/models/instruct_wrapper.py
ayush1298: Update docstring with these.

ayush1298:
@KennethEnevoldsen 
Results: 

Model name: intfloat/multilingual-e5-large-instruct, task: EmotionClassification, Results are matching exactly for both validation and test split. Below are the results for the test split.

Metric | old | new |
-------| ----- | ----- |
accuracy |  0.54985 | 0.54985 |  
f1 |  0.49156611889542523 | 0.49156611889542523 |
f1_weighted|  0.560848756305318 | 0.560848756305318 |
main_score | 0.54985 | 0.54985| 

Do you want me to check other models, mostly 7B, so I am not able to run the evaluation? But I was able to load all models.

Samoed:
It's better to test on retrieval/reranking tasks

ayush1298:
Model name: intfloat/multilingual-e5-large-instruct, task: AppsRetrieval. Matching results for all metrics

Metric | old | new |
-------| ----- | ----- |
recall_at_1 | 0.24382| 0.24382|  
ndcg_at_1 |  0.24382 | 0.24382 |
precision_at_1 | 0.24382 | 0.24382 |
map_at_1 | 0.24382 | 0.24382 |
mrr_at_1 | 0.24382470119521912 |  0.24382470119521912 | 
main_score | 0.34894 | 0.34894 | 

Model name: intfloat/multilingual-e5-large-instruct, task: AlloprofReranking. Matching results for all metrics

Metric | old | new |
-------| ----- | ----- |
mrr |  0.7580595306631576 | 0.7580595306631576 |
map | 0.7467772958086248 | 0.7467772958086248 |
main_score | 0.7467772958086248 | 0.7467772958086248 |

Samoed:
Thought prompt attention would make bigger difference

KennethEnevoldsen:
Yeah 100% thought that as well

KennethEnevoldsen:
Do we want additional tests for the larger models, or is this fine? If so I can try to give them a spin

ayush1298:
> Yeah 100% thought that as well

I think maybe for some other models, it may affect.  Though, yeah there should be change in results.

ayush1298:
> Do we want additional tests for the larger models, or is this fine? If so I can try to give them a spin

I think can you try 1st with just 1 of gritlm models, if they are good. Then you can run on all.

ayush1298:
@KennethEnevoldsen Can we merge this one? Or should you merge after you ran all models?

mteb/models/instruct_wrapper.py
KennethEnevoldsen, line 1: oh wait I see that we actually haven't remove the InstructGritLMModel and instruct_wrapper. We also need to remove the gritlm dependency in the pyproject.toml

ayush1298: Do we have to keep them for some backward compatibility in case?

KennethEnevoldsen: Yeah I suspect that other might be used them. We should probably just add a deprecation warning to them then.

ayush1298: Added the warning.

mteb/models/instruct_wrapper.py
```diff
@@ -40,6 +41,12 @@ def instruct_wrapper(
         device: Device used to load the model.
         **kwargs: Additional arguments to pass to the model.
     """
+    warnings.warn(
+        "`instruct_wrapper` is deprecated and will be removed in future versions. "
+        "Please use `InstructSentenceTransformerModel` instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
     requires_package(
```
KennethEnevoldsen, line 60: Delete an instead use:

```
@deprecated("msg")
def function()...
```

KennethEnevoldsen: Also add a deprecation on the class itself

ayush1298: So, to add depreciation on both before function and class? or only class?

KennethEnevoldsen: both

KennethEnevoldsen: (this will lead to two deprecation warnings that is fine)

ayush1298: Added it now.

ayush1298:
Can we merge this one?

PR #4086, opened 2026-02-11
model: add voyage-4-nano

### LINKED ISSUE #4073: Add model: voyage-4-nano

https://huggingface.co/voyageai/voyage-4-nano

### Arxiv link

_No response_

### Description of the model

There's base, lite and large, but not nano. 

Image

https://blog.voyageai.com/2026/01/15/voyage-4/

@KennethEnevoldsen I would take this as good first issue if that's ok?

mmilunovic:
Closes #4073 

`make test` and `make lint` are all passing
 
Tested on the following tasks with different values for `truncate_dim`:
 - NanoNFCorpusRetrieval (Retrieval)
 - NanoSciFactRetrieval (Retrieval)
 - STS12 (STS)
 - Banking77Classification (Classification)
 - SprintDuplicateQuestions (PairClassification)

### DIFF
--- a/mteb/models/model_implementations/voyage_models.py
+++ b/mteb/models/model_implementations/voyage_models.py
@@ -125,7 +125,6 @@ VOYAGE_TOTAL_TOKEN_LIMITS = {
     "voyage-3-lite": 120_000,
     "voyage-code-2": 120_000,
     "voyage-3-m-exp": 120_000,
-    "voyage-4-nano": 32_000,
 }
 
 


mteb/models/model_implementations/voyage_models.py
```diff
@@ -48,6 +125,7 @@
     "voyage-3-lite": 120_000,
     "voyage-code-2": 120_000,
     "voyage-3-m-exp": 120_000,
+    "voyage-4-nano": 32_000,
```
Samoed: Voyage nano using different wrapper
```suggestion
```

fzoll:
@KennethEnevoldsen @Samoed I think it is fine. I'm adding voyage-4-nano into the voyageai python package, but probably i won't touch this implementation even after that.

mmilunovic:
@KennethEnevoldsen @Samoed I noticed that this model is not visible on the leaderboard - is is because of the one failed check?

Samoed:
@mmilunovic We don't have results in results repo for this model, so we don't have data to show.

mmilunovic:
Got it, should I run it on what I can and check in the results?
Is there a minimal set of tasks that should be evaluated or is it all or none?

Samoed:
You can run on any number of tasks. For simplicity, you can select one of the benchmarks to run from https://embeddings-benchmark.github.io/mteb/overview/available_benchmarks

PR #3778, opened 2025-12-22
Feat: Add event logger

q275343119:
This is a PR for issue #3537 Added the event_logger module

Changes

* Added event_logger module
* Tracked events include:
* Homepage (leaderboard) load
* Benchmark switching
* Task filter changes
* Model filter changes

Data Storage
* Events are stored in MongoDB

Schema:
```json
{
  "event_name": "string",           // Event type identifier
  "timestamp": "2025-12-02T10:12:32Z",  // ISO 8601 UTC timestamp
  "session_id": "s_xyz789",         // Session ID (generated by frontend)
  "page_id": "leaderboard",          // Page identifier (default: "leaderboard")
  "benchmark": "MTEB-English",       // Current benchmark (optional)
  "filters": {                       // Current filter conditions snapshot (optional)
    "task_type": "classification",
    "domain": "all",
    "language": "en",
    ...
  },
  "properties": {                    // Event-specific properties (optional)
    "filter_name": "task_type",
    "old_value": "all",
    "new_value": "classification"
  }
}
```

PR #4093, opened 2026-02-12
Relax leaderboard_docker.yml timeout to 8min

isaac-chung:
https://github.com/embeddings-benchmark/mteb/pull/3778#issuecomment-3892513552


.github/workflows/leaderboard_docker.yml
isaac-chung:
```suggestion
      timeout-minutes: 9
```

PR #4050, opened 2026-02-05
Fill up n_embedding_parameters value in ModelMeta

ayush1298:
Closes #3947

Uses: For text-based models:  `n_embedding_parameters=vocab_size x hidden_dim` to fill up the `n_embedding_parameter` value for Models.

PR #4101, opened 2026-02-16
fix: refactor `from_*` function in model metas

KennethEnevoldsen:
refactor from_* function in model metas

PR to show the idea mentioned in #4050

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -521,8 +521,6 @@ class ModelMeta(BaseModel):
         This is based on the heuristic: `vocab_size * embedding_dim` where vocab_size and embedding_dim are extracted from the model's first
         Transformer module.
         """
-        # TODO: this seems rather arbitrary to me - could we combine the logic with that of cross-encoders?
-        # also why the try-except? would love to know when this fails
         logger.info(
             "Calculating number of embedding parameters for SentenceTransformer model."
         )


mteb/models/model_meta.py
```diff
@@ -417,13 +417,166 @@ def _detect_model_type_and_loader(
 
         return sentence_transformers_loader, "dense"
 
+    @classmethod
+    def _create_empty(cls, overwrites: dict | None = None) -> Self:
+        """Creates an empty ModelMeta with all fields set to None or empty."""
+        kwargs: dict[str, Any] = dict(
```
Samoed: Maybe use `ModelMeta` directly here?

mteb/models/model_meta.py
```diff
@@ -417,13 +417,166 @@ def _detect_model_type_and_loader(
...
+        name: str | None = (
+            model.model_card_data.model_name
+            if model.model_card_data.model_name
+            else model.model_card_data.base_model
+        )
+        n_parameters = cls._get_n_embedding_parameters_from_sentence_transformer_model(
+            model
+        )
+        return cls._create_empty(
+            overwrites=dict(
```
Samoed, line 503: Based on typing this should be `ModelMeta`

mteb/models/model_meta.py
```diff
@@ -439,76 +592,99 @@ def _from_hub(
...
+        embedding_dim = (
+            model_config.hidden_size
+            if model_config and hasattr(model_config, "hidden_size")
+            else None
+        )
+        max_tokens = (
+            model_config.max_position_embeddings
+            if model_config and hasattr(model_config, "max_position_embeddings")
+            else None
+        )
```
Samoed: I think we can simplify to
```suggestion
        max_tokens = getattr(model_config, "max_position_embeddings", None)
```

mteb/models/model_meta.py
```diff
@@ -526,93 +703,88 @@ def from_sentence_transformer_model(
...
+                "The fill_missing parameter is deprecated and will be removed in a future version. "
+                f"Use fetch_from_hf instead. Setting `fetch_from_hf={fill_missing}`.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            fetch_from_hf = fill_missing
+
+        meta = cls._from_sentence_transformer_model(model)
+        if fetch_from_hf:
+            name = cast("str", meta.name)
```
```suggestion
        if fetch_from_hf and meta.name is not None:
            name = meta.name
```

mteb/models/model_meta.py
```diff
@@ -417,13 +417,166 @@ def _detect_model_type_and_loader(
...
+    def _get_n_embedding_parameters_from_sentence_transformer_model(
+        model: SentenceTransformer,
+    ) -> int | None:
+        """Calculates the number of embedding parameters in a SentenceTransformer model
+
+        This is based on the heuristic: `vocab_size * embedding_dim` where vocab_size and embedding_dim are extracted from the model's first
+        Transformer module.
+        """
+        # TODO: this seems rather arbitrary to me - could we combine the logic with that of cross-encoders?
+        # also why the try-except? would love to know when this fails
```
ayush1298: For sentence-transformers and cross-encoders, we were using the code below to get the embedding values.

```py
### If the model can be loaded from transformers/senetence-transformers then load the model and use below code:

import numpy as np

model = AutoModel.from_pretrained(model_name, trust_remote_code=True)
emb = model[0].auto_model.get_input_embeddings()
meta.n_embedding_parameters = int(np.prod(emb.weight.shape))


###  If the model loaded using CrossEncoder then:

import numpy as np
from sentence_transformers import CrossEncoder

model = CrossEncoder("cross-encoder/ms-marco-MiniLM-L6-v2")
emb = model.model.get_input_embeddings()
n_embedding_parameters = int(np.prod(emb.weight.shape))
```
Here difference is just in 2nd line to get `emb`.
So, at that time, we don't have functionality of detecting loader, so I have added code separately to find these parameters in both `from_*` functions only, like https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_meta.py#L541-L548 and https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_meta.py#L636-L644

But now that we have a way to identify loader also using: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_meta.py#L362. We can use it to combine the logic with cross-encoders based on loader type.


Also, I think keeping a heuristic way to calculate embedding parameters in a separate function will be better as, and we can use this one directly: https://github.com/ayush1298/mteb/blob/a71b03d8af5c8d681482469dfc0359ac00e20ab0/mteb/models/model_meta.py#L808

Regarding `try-except`, it was added because many times models won't have `get_input_embeddings()` method implemented in their subclass, so it is used to catch that error.

mteb/models/model_meta.py
```diff
@@ -439,76 +592,99 @@ def _from_hub(
         loader: Callable[..., MTEBModels] | None
         model_type: MODEL_TYPES
 
-        if compute_metadata is not None:
+        loader, model_type = cls._detect_model_type_and_loader(model_name, revision)
+        reference = "https://huggingface.co/" + model_name
```
ayush1298: I think better to have both of these after checking `if not _repo_exists(model_name)`, because then we will be setting the wrong reference link.  Also, in that way you can also remove the check on line: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_meta.py#L392

mteb/models/model_meta.py
```diff
@@ -854,7 +1028,9 @@ def _get_frameworks_from_hf_tags(model_name: str) -> list[FRAMEWORKS]:
             "gguf": "GGUF",
         }
 
-        frameworks: list[FRAMEWORKS] = []
+        # Assume PyTorch support by default
+        # TODO: could be detected from repo as well
```
ayush1298: Yes, we can assume. Generally, `Pytorch` is not written explicitly in `hf model tags`, so not possible to extract from there.

mteb/models/model_meta.py
KennethEnevoldsen: it is? (`cls(**kwargs)`)

mteb/models/model_meta.py
KennethEnevoldsen: it does return a model meta? (Self is ModelMeta)

mteb/models/model_meta.py
KennethEnevoldsen: it is not None it (from create_empty) I would rather have the fail if it ends up as None

mteb/models/model_meta.py
KennethEnevoldsen: but it seems like it is there?


Screenshot 2026-02-17 at 09 30 08

ayush1298: Maybe then it's present for some of the models. But, if it's not present in tags, then we will not have Pytorch in tags if fetched from hf, which I assume will happen for many models (which though support PyTorch.)

mteb/models/model_meta.py
Samoed: Yes, but I think would be better to define `ModelMeta` instead of dict

mteb/models/model_meta.py
Samoed: Yes, missed

mteb/models/model_meta.py
Samoed: Should we add check then? I don't think casting here is good approach

mteb/models/model_meta.py
KennethEnevoldsen: It is not in the frontmatter on the readme. I suspect that they derive it from the files

KennethEnevoldsen: https://github.com/embeddings-benchmark/mteb/issues/4104

KennethEnevoldsen: added an issue on it

mteb/models/model_meta.py
KennethEnevoldsen: moved model_type - reference is used in the warning message beforehand

mteb/models/model_meta.py
```diff
@@ -439,76 +592,89 @@ def _from_hub(
...
+        loader, model_type = cls._detect_model_type_and_loader(model_name, revision)
+        reference = "https://huggingface.co/" + model_name
+
+        if not _repo_exists(model_name):
             warnings.warn(
-                "The compute_metadata parameter is deprecated and will be removed in a future version. "
-                f"Use fill_missing instead. Setting `fill_missing={compute_metadata}`.",
-                DeprecationWarning,
-                stacklevel=2,
+                f"Could not find model {model_name} on HuggingFace Hub repository ({reference}). Metadata will be limited."
```
ayush1298, line 592: I think we should not give a reference here. As it will be just incorrect in that case, as repo does not exists.

mteb/models/model_meta.py
KennethEnevoldsen: Yah I think adding a check is the best way to do it

KennethEnevoldsen: added

mteb/models/model_meta.py
KennethEnevoldsen: But here we don't have access to the loader, only the model itself. I redid the functions and combined them

mteb/models/model_meta.py
KennethEnevoldsen: But that is what we want to show? Cant find the model - this is where I looked

KennethEnevoldsen:
Reworked it one more time - incorporating your changes and fixing tests 

@ayush1298 reworked the computations emb size to (I believe) be slightly more general using vocab size and the `get_sentence_embedding_dimension`, but I might be wrong here.

mteb/models/model_meta.py
```diff
@@ -417,13 +415,163 @@ def _detect_model_type_and_loader(
...
+    def _get_n_embedding_parameters_from_sentence_transformers(
+        model: SentenceTransformer | CrossEncoder,
+    ) -> int | None:
+        """Calculates the number of embedding parameters in a SentenceTransformer model
+
+        This is based on the heuristic: `vocab_size * embedding_dim` where vocab_size and embedding_dim are extracted from the model's first
+        Transformer module.
+        """
+        # TODO: this seems rather arbitrary to me - could we combine the logic with that of cross-encoders?
+        # also why the try-except? would love to know when this fails
```
Samoed: I think we can remove this after refactor
```suggestion
```

Samoed (approved):
Great work!

mteb/models/model_meta.py
```diff
@@ -417,13 +415,163 @@ def _detect_model_type_and_loader(
...
+        )
+
+    @staticmethod
+    def _get_n_embedding_parameters_from_sentence_transformers(
+        model: SentenceTransformer | CrossEncoder,
+    ) -> int | None:
+        """Calculates the number of embedding parameters in a SentenceTransformer model
+
+        This is based on the heuristic: `vocab_size * embedding_dim` where vocab_size and embedding_dim are extracted from the model's first
+        Transformer module.
```
ayush1298, line 522: Can you also update this description as it 1st try to calculate by normal way and if fail then only uses heuristics.

ayush1298:
> @ayush1298 reworked the computations emb size to (I believe) be slightly more general using vocab size and the `get_sentence_embedding_dimension`, but I might be wrong here.

I am not sure about these one. But, similar to how `get_input_embeddings()` is not implemented. If this one is also not implemented then?
I think we anyway have access to `config.json`. So, we can use that.

mteb/models/model_meta.py
KennethEnevoldsen: I don't think that is true - it does it in this way in both cases

KennethEnevoldsen: this is always what happens in int(np.prod(emb.weight.shape))

KennethEnevoldsen:
> I think we anyway have access to config.json. So, we can use that.

How do we have access to this - is in the SentenceTransformer object?

Samoed:
> How do we have access to this - is in the SentenceTransformer object?

This is transformers config `AutoConfig`

KennethEnevoldsen:
@ayush1298 I think this is essentially ready to go - we might want to fix the config.json thing. but I think you can add your changes on top of this one. I would suggest that we just merge this one into main

mteb/models/model_meta.py
ayush1298: Sorry, I missed. That's the same only.

ayush1298:
> @ayush1298 I think this is essentially ready to go - we might want to fix the config.json thing. but I think you can add your changes on top of this one. I would suggest that we just merge this one into main

Sure, we can merge this one. Then I will merge this in #4050. Will see what to do for config.json thing, or we can use Transformers config.json in that case.

Samoed:
> Will see what to do for config.json thing, or we can use Transformers config.json in that case.

We already read the config in `_from_hub`, so we only need to add a small multiplication there https://github.com/embeddings-benchmark/mteb/blob/be2776c3adc46e19a23ec9e0c4db7d201af67fb5/mteb/models/model_meta.py#L606

ayush1298:
> > Will see what to do for config.json thing, or we can use Transformers config.json in that case.
> 
> We already read config in `_from_hub`, so only small multiplication is needed to be added
> 
> https://github.com/embeddings-benchmark/mteb/blob/be2776c3adc46e19a23ec9e0c4db7d201af67fb5/mteb/models/model_meta.py#L606

yes, I was suggesting same thing.  We can use these:
https://github.com/ayush1298/mteb/blob/a71b03d8af5c8d681482469dfc0359ac00e20ab0/mteb/models/model_meta.py#L808

Need to modify so as to directly take json if possible.

KennethEnevoldsen:
Sounds like this is a merge then

PR #4102, opened 2026-02-17
add: jina-embeddings-v5-text family

qherreros:
- [X] The model is public, i.e., is available either as an API or the weights are publicly available to download

Hi, here is the PR to add jina-embeddings-v5-text family into MTEB.

Thanks!

### DIFF
--- a/mteb/models/model_implementations/jina_models.py
+++ b/mteb/models/model_implementations/jina_models.py
@@ -732,23 +732,9 @@ class JinaV5TextWrapper(SentenceTransformerEncoderWrapper):
         model_prompts: dict[str, str] | None = None,
         **kwargs,
     ) -> None:
-        requires_package(self, "peft", model, "pip install 'mteb[peft]'")
         super().__init__(
             model, revision, device=device, model_prompts=model_prompts, **kwargs
         )
-
-        self.prompt_name_to_task = {
-            "Retrieval": "retrieval",
-            "Clustering": "clustering",
-            "Classification": "classification",
-            "STS": "text-matching",
-            "PairClassification": "text-matching",
-            "BitextMining": "text-matching",
-            "MultilabelClassification": "classification",
-            "Reranking": "retrieval",
-            "Summarization": "text-matching",
-            "InstructionReranking": "retrieval",
-        }
 
     def encode(
         self,
@@ -775,25 +761,37 @@ class JinaV5TextWrapper(SentenceTransformerEncoderWrapper):
 
         jina_task_name = None
         if self.model_prompts and prompt_name:
-            jina_task_name = self.prompt_name_to_task.get(prompt_name, None)
+            jina_task_name = self.model_prompts.get(prompt_name, None)
 
         task = jina_task_name if jina_task_name else "retrieval"
-        logger.info(f"Resolved task={task} for prompt_name={prompt_name}.")
-
-        prompt_name = (
-            "query" if prompt_type and prompt_type == PromptType.query else "document"
+        prompt = (
+            "Query: "
+            if prompt_type and prompt_type == PromptType.query
+            else "Document: "
         )
 
-        embeddings = self.model.encode(
-            sentences, task=task, prompt_name=prompt_name, **kwargs
-        )
+        embeddings = self.model.encode(sentences, task=task, prompt=prompt, **kwargs)
 
         return embeddings
 
 
 jina_embeddings_v5_text_small = ModelMeta(
     loader=JinaV5TextWrapper,
-    loader_kwargs=dict(trust_remote_code=True),
+    loader_kwargs=dict(
+        trust_remote_code=True,
+        model_prompts={
+            "Retrieval": "retrieval",
+            "Clustering": "clustering",
+            "Classification": "classification",
+            "STS": "text-matching",
+            "PairClassification": "text-matching",
+            "BitextMining": "text-matching",
+            "MultilabelClassification": "classification",
+            "Reranking": "retrieval",
+            "Summarization": "text-matching",
+            "InstructionReranking": "retrieval",
+        },
+    ),
     name="jinaai/jina-embeddings-v5-text-small",
     model_type=["dense"],
     modalities=["text"],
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,164 @@ def get_programming_task_override(
...
+        "PyTorch",
+        "Transformers",
+        "safetensors",
+    ],
+    use_instructions=True,
+    reference="https://huggingface.co/jinaai/jina-embeddings-v5-text-small",
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
+    adapted_from="Qwen3-0.6B",
```
Samoed:
```suggestion
    adapted_from="Qwen/Qwen3-0.6B",
```

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,164 @@ def get_programming_task_override(
...
+        "PyTorch",
+        "Transformers",
+        "safetensors",
+    ],
+    use_instructions=True,
+    reference="https://huggingface.co/jinaai/jina-embeddings-v5-text-nano",
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
+    adapted_from="EuroBERT",
```
Samoed:
```suggestion
    adapted_from="EuroBERT/EuroBERT-210m",
```

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,164 @@ def get_programming_task_override(
...
+    ),
+    name="jinaai/jina-embeddings-v5-text-nano",
+    model_type=["dense"],
+    modalities=["text"],
+    languages=multilingual_langs,
+    open_weights=True,
+    revision="4c1f9846bb85490df1d9e535834cf00ceb33823b",
+    release_date="2026-02-17",  # official release date
+    n_parameters=211766016,
+    n_embedding_parameters=None,
```
Samoed: Can you fill this?

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,164 @@ def get_programming_task_override(
...
+        "Sentence Transformers",
+        "PyTorch",
+        "Transformers",
+        "safetensors",
+    ],
+    use_instructions=True,
+    reference="https://huggingface.co/jinaai/jina-embeddings-v5-text-small",
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
```
Samoed, line 819: Can you fill this?

qherreros: Details regarding the training datasets cannot be disclosed for now.

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,164 @@ def get_programming_task_override(
...
+        prompt_name = (
+            "query" if prompt_type and prompt_type == PromptType.query else "document"
+        )
+
+        embeddings = self.model.encode(
+            sentences, task=task, prompt_name=prompt_name, **kwargs
+        )
+
+        if isinstance(embeddings, torch.Tensor):
+            embeddings = embeddings.cpu().detach().float().numpy()
```
KennethEnevoldsen:
```suggestion
```
shouldn't be required

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,164 @@ def get_programming_task_override(
...
+
+        jina_task_name = None
+        if self.model_prompts and prompt_name:
+            jina_task_name = self.model_prompts.get(prompt_name, None)
+
+        task = jina_task_name if jina_task_name else "retrieval"
+        prompt_name = (
+            "query" if prompt_type and prompt_type == PromptType.query else "document"
+        )
+
```
KennethEnevoldsen, line 772: I believe this should be possible with just supplying the prompts

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,164 @@ def get_programming_task_override(
...
+        **kwargs,
+    ) -> None:
+        from sentence_transformers import __version__ as st_version
+
+        current_sentence_transformers_version = tuple(map(int, st_version.split(".")))
+
+        if current_sentence_transformers_version < MIN_SENTENCE_TRANSFORMERS_VERSION:
+            raise RuntimeError(
+                f"sentence_transformers version {st_version} is lower than the required version 3.1.0"
+            )
```
KennethEnevoldsen: Hmm seems like there is no way to enforce the minimum version of sentence trf in the config for sentence transformer models. Maybe we should create an issue on that (Ideally that would be in the configs and not here)

mteb/models/model_implementations/jina_models.py
qherreros: The `task` argument allows the model to use the right LoRA adapter for the right task. We need to propagate the "task_type" information to the model. If I use prompts directly to pass "query" or "document", then I lose the task type information. It's also probably the reason we need a custom encode function.
Maybe I'm missing something though!

Samoed: Does `task` required for non retrieval tasks? Classification, bitext etc.

mteb/models/model_implementations/jina_models.py
Samoed: Will you provide them with paper, or it will be still private?

qherreros: For now, datasets are not provided in the paper either. If it changes, I will update them here too (if that's ok).

mteb/models/model_implementations/jina_models.py
qherreros: Yes it is required for non retrieval tasks

qherreros:
Do you see anything else I should rework?

Samoed:
@qherreros Can you add citation for your model?

qherreros:
Sure! Good catch!

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,169 @@ def get_programming_task_override(
...
+    """following the hf model card documentation."""
+
+    def __init__(
+        self,
+        model: str,
+        revision: str,
+        device: str | None = None,
+        model_prompts: dict[str, str] | None = None,
+        **kwargs,
+    ) -> None:
```
Samoed, line 734:
```suggestion
    ) -> None:
        requires_package(self, "peft", model, "pip install 'mteb[peft]'")
```

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,169 @@ def get_programming_task_override(
...
+        if self.model_prompts and prompt_name:
+            jina_task_name = self.model_prompts.get(prompt_name, None)
+
+        task = jina_task_name if jina_task_name else "retrieval"
+        prompt_name = (
+            "query" if prompt_type and prompt_type == PromptType.query else "document"
+        )
+
+        embeddings = self.model.encode(
+            sentences, task=task, prompt_name=prompt_name, **kwargs
```
Samoed: When I run with `sentence-transformers==5.2.2` I receive
```
ValueError: Prompt name 'query' not found in the configured prompts dictionary with keys ['Retrieval', 'Clustering', 'Classification', 'STS', 'PairClassification', 'BitextMining', 'MultilabelClassification', 'Reranking', 'Summarization', 'InstructionReranking'].
```

```bash
mteb run -m jinaai/jina-embeddings-v5-text-nano -t NFCorpus
```

mteb/models/model_implementations/jina_models.py
```diff
@@ -720,6 +721,170 @@ def get_programming_task_override(
...
```
Samoed: I read paper and seems you can simply change to
```suggestion
        prompt_name = (
            "Query: " if prompt_type and prompt_type == PromptType.query else "Document: "
        )

        embeddings = self.model.encode(
            sentences, task=task, prompt=prompt, **kwargs
```

mteb/models/model_implementations/jina_models.py
qherreros: Indeed, previous mapping strategy was overwritting the default prompt that should be {"query": "Query: ", "document": "Document: "}. With the last commit, I know keep model_prompts as it is, and I hard code the mapping (dataset => task) in the init function.

mteb/models/model_implementations/jina_models.py
qherreros: Ok I just saw you suggestion. Both options work, let me know the one you prefer.

Samoed: I think we can keep your implementation

Samoed:
@qherreros Seems that after you've moved definition of tasks lora is not applied correctly, because prompts are taken from `self.model_prompts` and no prompt name is selected. I think you can rollback you last change and just change last prompt. 


I tried to apply this and then my scores are matching with your

Samoed:
I've applied fix. If you agree then we can merge

qherreros:
Yes, this is what I spotted too! Thanks, we can merge

PR #3470, opened 2025-10-22
feat: Introduce MAEB

Samoed:
Fixes #2072

**What:** Introduces MAEB (Massive Audio Embedding Benchmark) - a comprehensive benchmark for evaluating audio embedding models

**Key additions:**
- **New audio task types:** Classification, multilabel classification, zero-shot classification, clustering, pair classification, reranking, and audio-text retrieval (A2T/T2A)
- **Audio models:** Wav2Vec2, WavLM, Whisper, CLAP, Qwen2Audio, MS-CLAP, and LCO-Embedding models
- **Datasets:** 95+ audio tasks including FSD50K, ESC-50, VoxCeleb, AudioCaps, Clotho, Common Voice, FLEURS, and many others
- **Benchmark variants:** MAEB (30 tasks), MAEB+ (97 tasks), MAEB(audio-only) (various subsets), and MAEB(audio-text) with lite/extended versions
- **Infrastructure:** Audio encoder interface, evaluation pipeline with resampling support, task selection using correlation analysis, and leaderboard integration

**Technical improvements:**
- Support for datasets v4+ with torchcodec for audio processing
- Efficient data collators for batch processing
- Descriptive statistics for all tasks
- Task selection methodology using correlation thresholds to reduce redundancy

PR #4103, opened 2026-02-17
 fix: Remove MAEB+ and MAEB(extended) from leaderboard and add "beta" to all MAEB

KennethEnevoldsen:
related to https://github.com/embeddings-benchmark/mteb/pull/3470

- Remove MAEB+ and MAEB(extended) from leaderboard
- Added the beta tag to denote that these might change
- Added whats new docs

Currently implemented it as keeping the two temporary benchmarks. We could consider removing them as well (I am unsure how much of a burden it is for us to maintain them), but I would probably not add them to the leaderboard.

All of these changes should be backward compatible

### DIFF
--- a/mteb/benchmarks/benchmarks/benchmarks.py
+++ b/mteb/benchmarks/benchmarks/benchmarks.py
@@ -3005,7 +3005,7 @@ MAEB_PLUS = Benchmark(
             "UrbanSound8KT2ARetrieval",
         ]
     ),
-    description="""MAEB+ is the full Massive Audio Embedding Benchmark (v1), containing 98 tasks with audio modality across 6 task types: classification, clustering, pair classification, reranking, zero-shot classification, and retrieval.""",
+    description="""MAEB+ is the full Massive Audio Embedding Benchmark (v1), containing 98 tasks with audio modality across 6 task types: classification, clustering, pair classification, reranking, zero-shot classification, and retrieval. The benchmark is currently in beta as the paper has been submitted for review and will be released in its final version after the review process.""",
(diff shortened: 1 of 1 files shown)


isaac-chung:
Is beta added because the paper is in review?

docs/whats_new.md
```diff
@@ -2,8 +2,38 @@
...
+### Added Audio Support
+Added audio support to MTEB 🎉. This includes support for loading and processing audio data in tasks. Overall this includes
+
+```python
+import mteb
+
+tasks= mteb.get_tasks()
+
+audio_task = [task for task in tasks if "audio" in task.metadata.modalities]
+len(audio_task) # 108 tasks
```
Samoed, line 18:
```suggestion
tasks= mteb.get_tasks(modalities=["audio"])

len(audio_task) # 108 tasks
```

mteb/leaderboard/benchmark_selector.py
```diff
@@ -54,10 +54,8 @@ class MenuEntry:
                 "Audio",
                 mteb.get_benchmarks(
                     [
-                        "MAEB(audio-only)",
```
isaac-chung, line 57: We do advertise Audio only in the paper, and agreed previously to keep it.

KennethEnevoldsen: not deleted just moved under MAEB

KennethEnevoldsen:
> Is beta added because the paper is in review?

yeah, essentially we could get reviews that point out flaws and we want to be able to accommodate those

Issue #4055, opened 2026-02-06
Look into integrating `mteb` with huggingfaces community evals


KennethEnevoldsen:
Currently it seems to be mostly relevant for generative models ([blog](https://huggingface.co/blog/community-evals)), but talked with @tomaarsen and it sounds like we could do something similar but where mteb replaces inspect AI as the runner.

<img width="1108" height="573" alt="Image" src="https://github.com/user-attachments/assets/454bac27-c5c5-44e5-8a31-67f470f3b070" />

---

I would say that we need the following:

- 1) Include the required metadata in the dataset repo (eq. to the [eval.yaml](https://huggingface.co/datasets/TIGER-Lab/MMLU-Pro/blob/main/eval.yaml)). I suppose we need the metadata as well as any additional settings such as `input_column_name`, `label_column_name` etc. - we could of course also allow a script itself as the config. We could even allow a "mteb version" tag for debugging.
- 2) We need to make this format loadable into MTEB
- 3) We need to a way for the community to push results

(the leaderboard in this case will only be a task-specific leaderboard, not for a full benchmark)

fixing 1 and 2 would also allow us to convert a lot of our code into metadata files that can simply be loaded in

KennethEnevoldsen:
@tomaarsen, @isaac-chung and @Samoed let me know if there is something I am missing here

Samoed:
1. We would need to create a script to export our format to eval.yaml. I think including script would be easier and more reproducible than just column names
2. Not sure about this, because we store more information in our resulta. I think we can update scores based on results from results repo to keep reproducibility. For metadata we can keep main in repo and just export it to eval.yaml
3. I think we can keep our current pipeline

KennethEnevoldsen:
1) I don't nec. think we need to use the eval.yaml format. Let us just start with the format that would work best for us - then we can work on compatibility (we can introduce it as a beta feature to allow us to make changes) - I suspect something like`mteb_task_config.toml`
1.1) re column names, it is more that we need more than just the metadata to run the task. I have added an example of how the data could look below.
2) Hmm I think there is a misunderstanding here - it is more about making the `mteb_task_config.toml` (or whatever we decide to use) loadable
3) def for the leaderboard - I was more thinking for what is displayed on the dataset page (see image above) for now though let us focus on 1 and 2


---
## Example implementation

### Writing the config (1):
```py
task = mteb.get_task("LccSentimentClassification")
task.to_config(name="mteb_task_config.toml") # default
[...]

Issue #4111, opened 2026-02-19
Consider aligning with evaluations standards


KennethEnevoldsen:
### Description of the feature

Similar to #4055 

There seems to be a standard here:
https://evalevalai.com/projects/every-eval-ever/

for standardizing LLM results. It is not quite for us, but we could consider to expand or formats to be more compatible. 

I def. don't think it is the primary focus, but e.g., adding a timestamp might be worthwhile. There is also the checksum that looks interesting.

Related issue on EuroEval to show that others also consider building around these: https://github.com/EuroEval/EuroEval/issues/1550

(it is probably more about making our results convertable to this format rather than reworking everything to this format)

Samoed:
I think our format is good enough and very comparable with their metadata. I don't think we need checksum of dataset, because we already have revision and we can reproduce with it. 

> adding a timestamp might be worthwhile

`retrieved_timestamp` or `evaluation_timestamp`?

KennethEnevoldsen:
> I think our format is good enough and very comparable with their metadata

Agree, it is mostly if we want to update anything or extend our format

> retrieved_timestamp or evaluation_timestamp?

evaluation_timestamp: what I was thinking (i.e., when the model was run).

PR #4114, opened 2026-02-19
feat: Integrate eval results

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/issues/4055

Pushing `eval.yaml`. PR https://huggingface.co/datasets/mteb/arguana/discussions/2
```python
import mteb
task = mteb.get_task("ArguAna")
task.push_eval_to_hub("mteb/ArguAna")
```

Push model results. PR https://huggingface.co/mteb/bm25s/discussions/3 (created dummy repo for bm25s)
```python
import mteb

cache = mteb.ResultCache()
results = cache.load_results(models=["bm25s"], include_remote=True)

model_result = results[0]
model_result.model_name = "mteb/bm25s"
model_result.push_model_results(True)
```

```python
import mteb

model_meta = mteb.get_model_meta("model_meta")
model_meta.push_eval_results()
```

Current not clear:
- What information to store. What we need to add more to support running from `eval.yaml`
- How we can customize `eval.yaml` (prompts, default metric)
- Should we save more metadata in task?
- How to save `filds config` for tasks with multiple splits?
- Should we add task results for split and subset (if task have all split results for subset)?


mteb/abstasks/_eval/eval_model.py
```diff
@@ -0,0 +1,67 @@
...
+from pydantic import BaseModel, ConfigDict, Field
+
+if TYPE_CHECKING:
+    from typing_extensions import Self
+
+    from mteb import TaskMetadata
+
+
+# limited version of eval.yaml model based on https://huggingface.co/datasets/Idavidrein/gpqa/blob/main/eval.yaml
+class EvalTaskConfig(BaseModel):
```
Samoed: Todo. Keep field ordering

mteb/abstasks/_eval/eval_model.py
KennethEnevoldsen, line 1: Would rename folder to _hf_eval_integration

mteb/abstasks/_eval/eval_model.py
```diff
@@ -0,0 +1,67 @@
...
+    field_spec: dict[str, str] | None = None
+
+
+class EvalMeta(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+    # TODO:
+    #  prompts? (what to do with default?)
+    #  default metric
+    #  more metadata about task? (domains etc.)
```
KennethEnevoldsen: hmm yeah, if we want to make it fully runnable, we probably need to at least this metadata for classification:

```toml
[task]
abstask="AbsTaskClassification" # required to specify runner
samples_per_label = 16 # could be more, but at least something like this

[task.metadata] 
name = "LccSentimentClassification" # does not nec. match repo name, should it?
type = "Classification" # probably not needed?
modalities = ["text"] # probably needed
eval_splits = ["test"] # already there
eval_langs = ["dan-Latn"] # already there
main_score = "accuracy" # required
prompt = "Classify texts based on sentiment" # required

dataset = { path = "DDSC/lcc", revision = "de7ba3406ee55ea2cc52a0a41408fa6aede6d3c6" } # what happens if I update the readme? 
```

mteb/abstasks/_eval/eval_model.py
```diff
@@ -0,0 +1,67 @@
...
+
+    from mteb import TaskMetadata
+
+
+# limited version of eval.yaml model based on https://huggingface.co/datasets/Idavidrein/gpqa/blob/main/eval.yaml
+class EvalTaskConfig(BaseModel):
+    id: str
+    config: str | None = Field(
+        None,
+        description="HF dataset config name (most time languages), e.g. `eng`, `fra`. None for mean all configs are used.",
```
KennethEnevoldsen:
```suggestion
        description="HF dataset config name (typically languages), e.g. `eng`, `fra`. None for mean all configs are used. Config is also sometimes referred to as subset or in mteb as `hf_subset`.",
```
Isn't it called something else in the UI (subset), or is it not the same? Also sounds like this is a list?

mteb/abstasks/abstask.py
```diff
@@ -603,6 +614,53 @@ def push_dataset_to_hub(self, repo_name: str, num_proc: int = 1) -> None:
...
+
+        eval_file_name = "eval.yaml"
+
+        existing_eval_path = _get_file_on_hub(
+            repo_id=repo_name,
+            file_name=eval_file_name,
+            repo_type="dataset",
+        )
+        # todo check how this works
+        # handle multiple tasks in one repo (e.g. MIRACLRetrievalHardNegatives, MIRACLRetrievalHardNegativesV2)
```
KennethEnevoldsen: Hmm yeah we might need two repos to handle these cases (which is a bit inefficient)

Alternatively, prioritize the latest version for superseeded datasets - but that would also require invalidation of results

mteb/models/get_model_meta.py
```diff
@@ -125,6 +125,8 @@ def get_model(
 
 _MODEL_RENAMES: dict[str, str] = {
     "bm25s": "baseline/bm25s",
+    # to store model's eval results to display on benchmark
+    "baseline/bm25s": "mteb/bm25s",
```
KennethEnevoldsen: Yeah saw this one, I think these are reasonable, we could consider renaming them to `mteb/baseline-bm25s`, I see that "baseline/" is not available

mteb/results/task_result.py
```diff
@@ -881,6 +888,46 @@ def get_hf_eval_results(self) -> list[EvalResult]:
                 )
         return results
 
+    def _get_hf_benchmark_result(self) -> list[BenchmarkEvalResult]:
```
KennethEnevoldsen: isn't to more "_to_hf_benchmark_result"

mteb/abstasks/_eval/eval_model.py
KennethEnevoldsen: Should we denote these specifically as `HFEvalTaskConfig`

KennethEnevoldsen (commented):
So currently it is not possible to run the repo with the `eval.yaml`. That seems to be somewhat expected, but we seem to mostly be missing the relevant data in the yaml.

mteb/abstasks/_eval/eval_model.py
Samoed: For now this is not clear if we allowed to create new field, or they will be prevented by hf validation

mteb/abstasks/_eval/eval_model.py
Samoed: No, this is just str https://github.com/huggingface/hub-docs/blob/f8b6bf740784ca40aa058620407f0e7b4a494322/docs/hub/eval-results.md?plain=1#L101

mteb/abstasks/abstask.py
Samoed: I think we need 2 repos only to show scores, I don't think we need to adapt to this

mteb/abstasks/_eval/eval_model.py
KennethEnevoldsen: Yeah let us ask about what the best approach is here

mteb/abstasks/_eval/eval_model.py
KennethEnevoldsen: Ahh, that is annoying that they use config there and subset in the UI.

mteb/abstasks/abstask.py
KennethEnevoldsen: So you want an empty repo with just the eval.yaml, but then I am not sure how this would be "runnable"

```
mteb run mteb/MIRACLRetrievalHardNegativesV2
```
Would download data from another repo?

mteb/abstasks/_eval/eval_model.py
Samoed: In some `hf_hub`/`datasets` functions it's name or config too

KennethEnevoldsen: yeah I actually recall that we naming variable in mteb whether I should go for the "config" or "subset" naming scheme.

KennethEnevoldsen: updated string to clarify

mteb/abstasks/abstask.py
Samoed: I think this would look like to be able to run any config from repo
```
mteb run repo_name repo_config_name (`id` in eval.yaml)
```

KennethEnevoldsen: Is that based on how it looks for ai-inspect? This would be very different from the current setup (would be doable, but I am not sure it is ideal)

Samoed: > Is that based on how it looks for ai-inspect?

I didn't look into it, but we define multiple `id` for tasks, so I think we should be able to run them separately

KennethEnevoldsen: Ahh as a way to have multiple in one

Samoed: Also to run subsets/splits independently

Samoed:
1. Moved utils functions around `huggingface_hub` to `hf_integration`
2. Added date for `TaskResult` as was suggested in https://github.com/embeddings-benchmark/mteb/issues/4111, because I needed it to test `date` field

mteb/results/task_result.py
```diff
@@ -851,6 +868,10 @@ def get_missing_evaluations(self, task: AbsTask) -> dict[str, list[str]]:
 
         return missing_splits
 
+    @deprecated(
+        "HF deprecated `EvalResults` in favor of `Benchmarks` and it's results. "
+        "To push new results use ModelMeta.push_eval_results()."
+    )
```
KennethEnevoldsen, line 875: Is it deprecated? I can't seem to find the docs on that (if we can it would be great to add a link

Samoed: > Old format of [EvalResult](https://github.com/huggingface/huggingface_hub/blob/7799c949f32745c6d0c6b2d48822a675c0d69057/src/huggingface_hub/repocard_data.py#L13) in model card is deprecated?
Correct, model repositories on the hub which include this have a "eval results (legacy)" tag with them.

https://github.com/embeddings-benchmark/mteb/issues/4055#issuecomment-3930062321 I can't find in docs previous format and not sure if it was existing

Samoed:
> How a way to run a model from the eval.yaml

I think we can add this later, when will be more info from HF team

> Any metadata in the yaml we want to add or is missing (see unfinished todos)

We can add, but I have a lot of problems with `eval.yaml` and I'm still not sure if we can add more (tried to add some fields, and it seems working, but who knows when it breaks)

mteb/results/task_result.py
KennethEnevoldsen: Yeah same thing I found - but good to see the comment. I think we can resolve this

KennethEnevoldsen:
> I think we can add this later, when will be more info from HF team

> We can add, but I have a lot of problems with eval.yaml and I'm still not sure if we can add more (tried to add some fields, and it seems working, but who knows when it breaks)

Hmm, I can't seem to figure out if the existing benchmarks are runnable using a CLI pointing at the hub, but if not, then I agree that we should just leave it for now.

Also considered two other things; 1) is it possible for mteb results to get the verified flag, 2) which tasks should ask to be made into benchmarks (and push the results to)

Both of these can be separate issues though.

Samoed:
> is it possible for mteb results to get the verified flag

The problem I see with our existing tasks that we can't verify them all 100%, e.g. implementation of qwen3 in our repo have different scores or scores for MAEB tasks is different after we've filtered out empty audios. Maybe we have more cases like. We can submit from MTEB user like currently for ArguAna and this will be enough
image

I think we can make scores verifiable that run thought HF job, but I can't imagine for now how this would work

Samoed:
Updated tests and added to `what's new`

KennethEnevoldsen:
It would be great to figure out what they require: https://github.com/embeddings-benchmark/mteb/issues/4138

I think there are multiple reasonable ways to have it verified - added an issue on it: #4138 

We might even be interested in adding a verification layer (currently, we can't display it on the leaderboard, though - it would probably require more than what gradio can do)

docs/whats_new.md
```diff
@@ -2,6 +2,35 @@
 
 This section is an overview of releases for more information check out the autogenerated [changelog](https://github.com/embeddings-benchmark/mteb/blob/main/pyproject.toml).
 
+## New in v2.9
+
+### Integration with HuggingFace Benchmarks
+
+Added integration with HuggingFace Benchmarks. This allows for users easily see scores of task directly on dataset card.
```
KennethEnevoldsen: Let us link to an example - probably also add an image

docs/whats_new.md
```diff
@@ -2,6 +2,35 @@
...
+model_meta = mteb.get_model_meta("model_name")
+model_meta.push_eval_results(
+    tasks=[
+        "Task1",
+        "Task2",
+    ],
+    cache=mteb.ResultCache(),
+    create_pr=True,
+)
+```
```
KennethEnevoldsen, line 35: Should we also add a statement that we will over the coming months convert tasks which are a part of mteb benchmarks to benchmark tasks and submit scores on existing models (and then make issues on those)

docs/whats_new.md
KennethEnevoldsen: I think people aren't aware what it is, so it might also be nice to add a link to the blog + potentially add what the different PRs add.

mteb/models/get_model_meta.py
KennethEnevoldsen: We need to resolve this one so that it is consistent across models

docs/whats_new.md
Samoed: We can create an issue, but I don't think that this took much time on our side

mteb/models/get_model_meta.py
Samoed: I can rename to `mteb/baseline-bm25s` if you want

docs/whats_new.md
KennethEnevoldsen: coming months might be overstating it - coming month or coming week is also possible ways to phrase it

docs/whats_new.md
Samoed: image

mteb/models/get_model_meta.py
```diff
@@ -124,9 +124,9 @@ def get_model(
 
 
 _MODEL_RENAMES: dict[str, str] = {
-    "bm25s": "mteb/bm25s",
+    "bm25s": "mteb/baseline-bm25s",
     # to store model's eval results to display on benchmark
-    "baseline/bm25s": "mteb/bm25s",
+    "baseline/bm25s": "mteb/baseline-bm25s",
```
KennethEnevoldsen, line 129: should we rename the other baselines for consistency

KennethEnevoldsen (approved):

mteb/models/get_model_meta.py
Samoed: Renamed to
```
    "baseline/bm25s": "mteb/baseline-bm25s",
    "baseline/random-cross-encoder-baseline": "mteb/baseline-random-cross-encoder",
    "baseline/random-encoder-baseline": "mteb/baseline-random-encoder",
    "baseline/bb25": "mteb/baseline-bb25",
```

KennethEnevoldsen:
I think is good to merge (though a test seem to be failing)

Samoed:
Renamed results in cache

Issue #2122, opened 2025-02-21
Release MMTEB datasets on Hugging Face


NielsRogge:
Hi @KennethEnevoldsen 🤗

I'm Niels and work as part of the open-source team at Hugging Face. I discovered your work through Hugging Face's daily papers as yours got featured: https://huggingface.co/papers/2502.13595.
The paper page lets people discuss about your paper and lets them find artifacts about it (your dataset for instance),
you can also claim the paper as yours which will show up on your public profile at HF.

Would you like to host the datasets you've released on https://huggingface.co/datasets?
Hosting on Hugging Face will give you more visibility/enable better discoverability, and will also allow people to do:

```python
from datasets import load_dataset

dataset = load_dataset("your-hf-org-or-username/your-dataset")
```

If you're down, leaving a guide here: https://huggingface.co/docs/datasets/loading.
We also support Webdataset, useful for image/video datasets: https://huggingface.co/docs/datasets/en/loading#webdataset.

Besides that, there's the [dataset viewer](https://huggingface.co/docs/hub/en/datasets-viewer) which allows people to quickly explore the first few rows of the data in the browser.

Let me know if you're interested/need any guidance.

Kind regards,

Niels

Samoed:
We have uploaded a lot of datasets in our organization https://huggingface.co/mteb and all datasets that presented in MTEB are from Hugging Face

isaac-chung:
Hi @NielsRogge !

Thanks for the message. Adding to what @Samoed pointed out, every task has a corresponding huggingface dataset name and revision. I wonder if a collection would also be beneficial to create here? Ultimately, the datasets are grouped by benchmarks under this repo's benchmarks.py file, so the collection might serve as a mirror for those.

We also have an ongoing task to move datasets under other users to the MTEB org on HF. (I'll have to find the link)

NielsRogge:
Yes that could be great! I see you have https://huggingface.co/collections/mteb/mmteb-67b74a586236bc839971e8cd but it does not include the datasets so far

KennethEnevoldsen:
Thanks for reaching out @NielsRogge 🤗 - love the work on datasets and I think I can speak for everyone when I say that it makes the dataset aspect of MTEB a lot easier. E.g. I don't think any of the datasets are on HF the ease of access is just a very nice thing to have. 

We could indeed start moving datasets into a collection - is there a way to do a bulk import for collections? Or automate it, then we could do it for all the benchmark in MTEB

[...]

PR #4115, opened 2026-02-19
add method for creating collection for benchmark

Samoed:
I added a method for collection creation for benchmark class. It will create a collection with benchmark name and description, and it's all datasets

Example: 
```python
import mteb
benchmark = mteb.get_benchmark("BEIR")
benchmark.push_collection_to_hub("Samoed")
```
Result https://huggingface.co/collections/Samoed/beir

Ref https://github.com/embeddings-benchmark/mteb/issues/2122

### DIFF
--- a/mteb/benchmarks/benchmark.py
+++ b/mteb/benchmarks/benchmark.py
@@ -114,25 +114,34 @@ class Benchmark:
             )
             return no_results_frame
 
-    def push_collection_to_hub(self, hf_username: str) -> None:
+    def push_collection_to_hub(
+        self,
+        hf_username: str,
+        collection_name: str | None = None,
+    ) -> None:
         """Push the benchmark collection to Hugging Face Hub.
 
         Args:
             hf_username: Hugging Face username or organization name
+            collection_name: Name for the collection on Hugging Face Hub. If not provided, the benchmark name will be used.
         """
         collections = huggingface_hub.list_collections(owner=hf_username)
+        collection_name = collection_name or self.name
         existing_collection = None
         for collection in collections:
-            if collection.title == self.name:
+            if collection.title == collection_name:
                 existing_collection = collection
                 break
 
         if existing_collection is None:
+            description = self.description
+            if description and len(description) > 150:
+                description = description[:147] + "..."
             collection = huggingface_hub.create_collection(
-                title=self.name,
+                title=collection_name,
                 namespace=hf_username,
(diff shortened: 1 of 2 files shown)


mteb/benchmarks/benchmark.py
```diff
@@ -112,6 +114,49 @@ def _create_per_language_table(
...
+
+        if existing_collection is None:
+            collection = huggingface_hub.create_collection(
+                title=self.name,
+                namespace=hf_username,
+                # hf collections have a 150 character limit for description, so we truncate it if it's too long
+                description=self.description[:150] if self.description else None,
+            )
+        else:
+            # list collections would output only 4 items
```
isaac-chung, line 147: Could we add a bit more context on how this relates to the outcome/logic?

Samoed: Collections from `list_collections` would contain only 4 items (e.g. 4 datasets) even if collection have more. https://github.com/huggingface/huggingface_hub/blob/7799c949f32745c6d0c6b2d48822a675c0d69057/src/huggingface_hub/hf_api.py#L8359-L8361

I fetch the full collection in order to retrieve all items from the collection, so as to avoid sending duplicate items to the hub. Otherwise, the API will return a "409 Conflict" error

mteb/benchmarks/benchmark.py
```diff
@@ -112,6 +114,49 @@ def _create_per_language_table(
...
+            if collection.title == self.name:
+                existing_collection = collection
+                break
+
+        if existing_collection is None:
+            collection = huggingface_hub.create_collection(
+                title=self.name,
+                namespace=hf_username,
+                # hf collections have a 150 character limit for description, so we truncate it if it's too long
+                description=self.description[:150] if self.description else None,
```
KennethEnevoldsen: maybe add some "..."

mteb/benchmarks/benchmark.py
KennethEnevoldsen, line 1: Looks good!

Nothing to add other than, let us make some collections! (though there might be a thing with the names)

Samoed: > though there might be a thing with the names

Do you mean allow to customize names? I thought about this too

mteb/benchmarks/benchmark.py
Samoed: Added

mteb/benchmarks/benchmark.py
Samoed: Added customization for collection name

PR #4120, opened 2026-02-20
fix: pylate model prompts

Samoed:
Seems that after `v2` pylate wrapper handled prompts incorrectly. Previously, prompts where passed explicitly to `get_prompt`, but in `v2` they're taken from `self.model_prompts` and this wasn't added during refactor

### DIFF
--- a/mteb/models/model_implementations/pylate_models.py
+++ b/mteb/models/model_implementations/pylate_models.py
@@ -309,7 +309,7 @@ class MultiVectorModel(AbsEncoder, PylateSearchEncoder):
         self.model = ColBERT(self.model_name, revision=revision, **kwargs)
         built_in_prompts = getattr(self.model, "prompts", None)
         if built_in_prompts and not model_prompts:
-            self.model_prompts = model_prompts
+            self.model_prompts = built_in_prompts
         elif model_prompts and built_in_prompts:
             logger.info(f"Model.prompts will be overwritten with {model_prompts}")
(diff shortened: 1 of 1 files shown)


Samoed:
Do you mean mteb v2?

KennethEnevoldsen:
Ahh thought you meant pylate - ahh so this is a bug introduced in mtebv2

Samoed:
@NohTow This fixes your problem, right?

NohTow:
Well this is very similar to the fix I did and we discussed the remaining
But I can launch some actual tests on Monday if we want to validate (I will try to replicate ColBERT-Zero with this branch, it has prompts)

mteb/models/model_implementations/pylate_models.py
```diff
@@ -308,11 +309,10 @@ def __init__(
         self.model = ColBERT(self.model_name, revision=revision, **kwargs)
         built_in_prompts = getattr(self.model, "prompts", None)
         if built_in_prompts and not model_prompts:
-            self.model.prompts = built_in_prompts
+            self.model_prompts = model_prompts
```
NohTow: Should be self.model_prompts = built_in_prompts

NohTow:
```suggestion
            self.model_prompts = built_in_prompts
```

NohTow:
I tested two datasets with ColBERT-Zero (prompt) and GTE-ModernColBERT (no prompt) and Zero models seems fixed while GTE model seems untouched, so LGTM

PR #4132, opened 2026-02-22
Add msmarco-minilm-v2 series

sarendis56:
Add `cross-encoder/ms-marco-MiniLM` v2 series. The main difference with `sentence-transformers/all-MiniLM` that is already included in the library is that it adopts cross-encoder architecture, which will output a single similarity score given a pair of text. The model is included in the Sentence-Transformers as well, therefore it's quite straightforward.

The reproduction results are as follows. Because the relative large corpus size of the datasets, it would be challenging to do forward pass of all queries and corpus items. Therefore, I use `sentence-transformers/all-MiniLM-L6-v2` to do the first-stage retrieval (coarse filtering). I do not know if the following results included in the original [model card](https://huggingface.co/cross-encoder/ms-marco-MiniLM-L12-v2) has this step; therefore the results are a bit different yet make sense.

| Model-Name                            | NDCG@10 (TREC DL 19) | Reproduced | MRR@10 (MS Marco Dev) | Reproduced |
| ------------------------------------- | -------------------- | ---------- | --------------------- | ---------- |
| cross-encoder/ms-marco-TinyBERT-L2-v2 | 69.84                | 66.39      | 32.56                 | 33.20      |
| cross-encoder/ms-marco-MiniLM-L2-v2   | 71.01                | 66.85      | 34.85                 | 35.60      |
| cross-encoder/ms-marco-MiniLM-L4-v2   | 73.04                | 70.88      | 37.70                 | 38.32      |
| cross-encoder/ms-marco-MiniLM-L6-v2   | 74.30                | 72.52      | 39.01                 | 39.89      |
| cross-encoder/ms-marco-MiniLM-L12-v2  | 74.31                | 72.93      | 39.02                 | 39.91      |


Samoed:
Overall metadata looks fine. Only thing can you fill `n_embedding_parameters`? You can get it from `ModelMeta.from_hub(moel_name)`

Issue #4035, opened 2026-02-02
Add saturation plot to the leaderboard


KennethEnevoldsen:
### Description of the feature

Add Performance over time as this can give a good indication of whether a benchmark is saturated or if we are still seeing improvements.

Required data:
- model release dates (we have)
- performance (we have)

Example:

<img width="431" height="293" alt="Image" src="https://github.com/user-attachments/assets/228d4cd2-2705-49e5-aca0-8cf8477a166c" />

Samoed:
Great idea! Like was in papers with code

KennethEnevoldsen:
exactly!

KennethEnevoldsen:
We could even split it on open vs. proprietary models

<img width="785" height="465" alt="Image" src="https://github.com/user-attachments/assets/42db357c-46d9-4c41-a166-6c953ba76fc5" />

PR #4134, opened 2026-02-22
fix: add performance over time

Samoed:
Add performance for leaderboard over time like `papers with code`
image
Close https://github.com/embeddings-benchmark/mteb/issues/4035

### DIFF
--- a/mteb/leaderboard/figures.py
+++ b/mteb/leaderboard/figures.py
@@ -234,12 +234,8 @@ def _performance_size_plot(df: pd.DataFrame) -> go.Figure:
 @_failsafe_plot
 def _performance_over_time_plot(df: pd.DataFrame) -> go.Figure:
     df = df.copy()
-    score_column = None
-    if "Mean (TaskType)" in df.columns:
-        score_column = "Mean (TaskType)"
-    elif "Mean (Task)" in df.columns:
-        score_column = "Mean (Task)"
-    if score_column is None or "Model" not in df.columns:
+    score_column = "Mean (Task)"
+    if score_column not in df.columns or "Model" not in df.columns:
         return _text_plot(
             "Couldn't produce timeline plot. Required columns are missing."
         )


KennethEnevoldsen:
I believe it should be a fix though so changes that one

mteb/leaderboard/figures.py
```diff
@@ -198,6 +231,72 @@ def _performance_size_plot(df: pd.DataFrame) -> go.Figure:
...
 
 
+@_failsafe_plot
+def _performance_over_time_plot(df: pd.DataFrame) -> go.Figure:
+    df = df.copy()
+    score_column = None
+    if "Mean (TaskType)" in df.columns:
+        score_column = "Mean (TaskType)"
+    elif "Mean (Task)" in df.columns:
+        score_column = "Mean (Task)"
```
KennethEnevoldsen: shouldn't we prioritize mean(Task) over mean tasktype?

KennethEnevoldsen: (since we do that on the other plots)

KennethEnevoldsen:
Hmm tested it out with model size filters and seems like changing the filter does not trigger an update of the plot

KennethEnevoldsen:
Also was thinking whether we should make a couple of lines potentially with [toggles](https://stackoverflow.com/questions/65941253/plotly-how-to-toggle-traces-with-a-button-similar-to-clicking-them-in-legend):

Something like "All models", "Open weight", "<1B parameters"

Would love to compare those three lines.

Samoed:
For me plot is updating after filter changing 
image
image

It just can work a bit slow, because firstly summary table should rerender, and then other plots would update

> Also was thinking whether we should make a couple of lines potentially with [toggles](https://stackoverflow.com/questions/65941253/plotly-how-to-toggle-traces-with-a-button-similar-to-clicking-them-in-legend):

I don't think we should, because we already have model filters or add this to all plots

mteb/leaderboard/figures.py
Samoed: Changed

KennethEnevoldsen:
> It just can work a bit slow, because firstly summary table should rerender, and then other plots would update

Hmm, but it doesn't seem to be loading - maybe loading is only tricked afterwards? Added a related issue on it (feel free to merge there)

PR #4128, opened 2026-02-21
fix: lotte wrong main metric

Samoed:
Previously, we had `Lotte` with `recall@5`, but in paper in mentioned as `Success@5 (sometimes referred to as Recall@5)`, which is hit_rate@5 in code https://github.com/stanford-futuredata/ColBERT/blob/cc4f3dc91c0b45d2d08c251d9d95178285c65f1c/utility/evaluate/evaluate_lotte_rankings.py#L39-L40, and we have `cv_recall` which is actually hit_rate (didn't figure out it during v2 for some reason)

Was found after this some discussion with @NohTow. Currently, I'm running some models to verify scores

Probably we can rename `cv_recall` to `hit_rate` and use it from `trec_eval` (it's called as `success` here)

PR #4142, opened 2026-02-23
fix: add `hit_rate` to retrieval metrics and remove `cv_recall`

Samoed:
Continue from https://github.com/embeddings-benchmark/mteb/pull/4128, Added `hit_rate` calculation to retrieval metrics

### DIFF
--- a/mteb/tasks/retrieval/eng/lotte_retrieval.py
+++ b/mteb/tasks/retrieval/eng/lotte_retrieval.py
@@ -27,7 +27,6 @@ class LoTTERetrieval(AbsTaskRetrieval):
         eval_langs={domain: ["eng-Latn"] for domain in HF_SUBSETS},
         # in paper mentioned as `Success@5 (sometimes referred to as Recall@5)`,
         # which is hit_rate@5 in code https://github.com/stanford-futuredata/ColBERT/blob/cc4f3dc91c0b45d2d08c251d9d95178285c65f1c/utility/evaluate/evaluate_lotte_rankings.py#L39-L40
-        # and mteb have hit_rate_at_5 which is actually hit_rate
         main_score="hit_rate_at_5",
         date=("2021-12-02", "2022-06-10"),
         domains=["Academic", "Web", "Social"],


mteb/_evaluators/retrieval_metrics.py
```diff
@@ -613,6 +637,9 @@ def evaluate_abstention(
     return naucs
 
 
+@deprecated(
+    "`calculate_cv_recall` is deprecated and will be replaced with `hit_rate` in a future version"
+)
```
KennethEnevoldsen: can we compare the scores on a few dataset just to ensure that they are compatible?

mteb/tasks/retrieval/eng/lotte_retrieval.py
```diff
@@ -28,7 +28,7 @@ class LoTTERetrieval(AbsTaskRetrieval):
         # in paper mentioned as `Success@5 (sometimes referred to as Recall@5)`,
         # which is hit_rate@5 in code https://github.com/stanford-futuredata/ColBERT/blob/cc4f3dc91c0b45d2d08c251d9d95178285c65f1c/utility/evaluate/evaluate_lotte_rankings.py#L39-L40
         # and mteb have cv_recall_at_5 which is actually hit_rate
-        main_score="cv_recall_at_5",
+        main_score="hit_rate@5",
```
KennethEnevoldsen: This will break backward compatibility when loading results (suddenly models that previously had results on the leaderboard will no longer have that)

KennethEnevoldsen: If they are 1-1 compatible (within our threshold), we could rename cv_recall_at_5 to hit_rate@5 on load.

mteb/_evaluators/retrieval_metrics.py
Samoed: Yes, I've checked and they're exactly the same

mteb/tasks/retrieval/eng/lotte_retrieval.py
Samoed: > This will break backward compatibility when loading results

Only results we have for LoTTE is https://github.com/embeddings-benchmark/results/pull/421 and we don't have it on leaderboard

Also, this just add new metric, but `main_score` would work still (like we've done with reranking tasks in v2)

KennethEnevoldsen: But if you load and validate it, it would give you a warning stating that it does not contain the main score "hit rate", no? Or do we just see if there is a main score and then just assume that is the same as `results.get_score(task.metadata.main_score)`

Samoed: > do we just see if there is a main score and then just assume that is the same as results.get_score(task.metadata.main_score)

It seems like this https://github.com/embeddings-benchmark/mteb/blob/d37253042e23d5047cdd187d84f2b39e77b9c1a6/mteb/results/task_result.py#L497-L505

We have warning for missing scores only when load pre 1.11

https://github.com/embeddings-benchmark/mteb/blob/d37253042e23d5047cdd187d84f2b39e77b9c1a6/mteb/results/task_result.py#L463-L470

KennethEnevoldsen:
This should work just fine then!

mteb/_evaluators/retrieval_metrics.py
```diff
@@ -613,6 +637,9 @@ def evaluate_abstention(
...
 def calculate_cv_recall(
```
KennethEnevoldsen: Can't we just remove this? (we can just assign cv_recall based on hit rate) - this is also in the private module so it should be fine to remove.

isaac-chung: Ooo don't forget all the previous results will not have hit rate. We'll be fine right? It'll just be under main score.

Samoed: Yes, we've done similarly for reranking scores in v2 and they're fine

PR #4143, opened 2026-02-23
docs: add cache/search backends to doc

Samoed:
Added cache and search backends to docs. Added links to classes in examples


KennethEnevoldsen:
Great changes!

PR #4145, opened 2026-02-23
fix: languages filtering scores

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3611

Make language filter update leaderboard scores

https://github.com/user-attachments/assets/fa98a556-e740-44db-8cf6-822ecf59fb20


KennethEnevoldsen:
Dockerfile seems to keep failing - I assume that the leaderboard is just fine though

Samoed:
Yes, probably timout problems

PR #4148, opened 2026-02-23
Start video

Samoed:
I've made `VideoInput` to receive [`VideoDecoder`](https://github.com/meta-pytorch/torchcodec/blob/40eae03bb46fd2bac7927d0a1b5158bb30ba5fd4/src/torchcodec/decoders/_video_decoder.py#L78) to make possible to override by models later to select required amount of frames (like for audio tasks we add resampling everywhere). 

I tried to run `baseline/random-encoder-baseline` on `MSR-VTT` 879 rows and it took me 50 s to encode videos (10 frames from video).

Also I've noticed that in our version MSR-VTT have only 879 rows, while source have 10k https://huggingface.co/datasets/friedrichor/MSR-VTT

To comparison MMEB selecting per task using ffpobe 
https://github.com/TIGER-AI-Lab/VLM2Vec/blob/6e1e1d42d4cba85ae4db4d3f7a25cba2465d1e16/src/data/dataset/msrvtt_dataset.py#L15-L17 
https://github.com/TIGER-AI-Lab/VLM2Vec/blob/6e1e1d42d4cba85ae4db4d3f7a25cba2465d1e16/src/utils/vision_utils/vision_utils.py#L317-L326


mteb/tasks/video/retrieval/eng/msr_vtt.py
```diff
@@ -0,0 +1,68 @@
...
+if __name__ == "__main__":
+    import logging
+
+    import mteb
+
+    logging.basicConfig(level=logging.INFO)
+
+    task = MSRVTT()
+    model = mteb.get_model("baseline/random-encoder-baseline")
+    mteb.evaluate(model, task)
```
Samoed: Remove before merge

AdnanElAssadi:
>Also I've noticed that in our version MSR-VTT have only 879 rows, while source have 10k

I just took the test set which is 1k. I also filtered out videos with missing/corrupted audio.

Samoed:
I think we need to upload full split just in case

mteb/types/_encoder_io.py
```diff
@@ -132,14 +133,32 @@ class AudioInput(TypedDict):
...
 
-class MultimodalInput(TextInput, CorpusInput, QueryInput, ImageInput, AudioInput):  # type: ignore[misc]
+class VideoInput(TypedDict):
+    """The input to the encoder for videos.
+
+    Attributes:
+        video: The video to encode. VideoDecoder object.
+    """
+
+    video: VideoDecoder
```
Samoed: Will change to include audio as well

AdnanElAssadi: I think having these separate like my PR might be better. I don't think there is a way to extract audio from VideoDecoder.

Samoed: There is no way, but we can change to `"video": [{"audio": AudioInputItem, "frames": VideoDecoder}]` as was discussed in slack because there can be tasks where can be audio and video together (not related) and we need to process them

mteb/tasks/video/retrieval/eng/msr_vtt.py
```diff
@@ -0,0 +1,95 @@
...
+        domains=None,
+        task_subtypes=None,
+        license=None,
+        annotations_creators=None,
+        dialect=None,
+        sample_creation=None,
+        bibtex_citation=None,
+    )
+
+    def load_data(self, num_proc: int | None = None, **kwargs) -> None:
```
KennethEnevoldsen, line 34: Assuming we want to change this in the future - would be good to see how the final format would look like when uploaded (but I don't see a reason why we can't use what we have now)

mteb/tasks/video/retrieval/eng/msr_vtt.py
```diff
@@ -0,0 +1,95 @@
...
+
+if __name__ == "__main__":
+    import logging
+
+    import mteb
+
+    logging.basicConfig(level=logging.INFO)
+
+    task = MSRVTTV2T()
+    model = mteb.get_model("baseline/random-encoder-baseline")
```
KennethEnevoldsen: This model was such a great addition - has really been great for testing

mteb/tasks/video/retrieval/eng/__init__.py
KennethEnevoldsen, line 1: looks good - I like that we add it directly without loading it

mteb/models/model_implementations/random_baseline.py
```diff
@@ -111,32 +145,58 @@ def _batch_to_embeddings(
...
+            ]
+
+        # Combine embeddings
+        max_len = max(
+            [
+                len(text_embeddings),
+                len(image_embeddings),
+                len(audio_embeddings),
+                len(video_embeddings),
+            ]
```
KennethEnevoldsen, line 179: why would the length be different?

KennethEnevoldsen (commented):
Look good - I like the structure a lot!

I am unsure if we should version bump based on these (as some of the changes might affect the core functionality, though ideally shouldn't) - leaning toward minor bumps

mteb/tasks/video/retrieval/eng/msr_vtt.py
Samoed: I think HF don't support preview of videos when they're inside dicts. Overall current format is limitation of torchcodec and datasets, because neither of them support audio in video

mteb/models/model_implementations/random_baseline.py
Samoed: Some of them can be with length 0, e.g. text only tasks don't have images

Samoed:
> as some of the changes might affect the core functionality, though ideally shouldn't

These changes don't affect core

mteb/tasks/video/retrieval/eng/msr_vtt.py
KennethEnevoldsen: Hmm I suppose we could have the columns be `video_frames` and `video_audio`, but we would still have the content as lists

KennethEnevoldsen: How much do we care about the preview?

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: and we won't have issues in cases where we have a text only sample interleaves with image only samples (I don't believe we have that dataset, but we could have)

mteb/tasks/video/retrieval/eng/msr_vtt.py
Samoed: I think preview is very helpful to understand a dataset. Overall, I think we can keep only one video overtime and then convert to list later, because even for images we don't have datasets with multiple images in one row

KennethEnevoldsen:
> These changes don't affect core

but It does affects the baseline encoder, which non-video users will use. 

(note meant patch bump above not minor)

mteb/models/model_implementations/random_baseline.py
Samoed: No, we don't have for now and such dataset would break a lot

Samoed:
> but It does affects the baseline encoder, which non-video users will use.

Yes a bit, but we just expand modalities, and this won't cause problems for them, and they wouldn't know until they look into code

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: Ohh! I would just imagine that we did something like:

```
{"texts":["text 1", "", "text 3"}, images: [[], [image1], []], ...}
```

KennethEnevoldsen: maybe we should just factor this out into an issue for later (I don't think it is a biggie for now)

KennethEnevoldsen:
Agree - so let us set it not to version bumps on these, unless we change core functionality

mteb/models/model_implementations/random_baseline.py
Samoed: Yes, better to create a new issue

mteb/types/_encoder_io.py
AdnanElAssadi: What if models have different sampling strategies, doesn't this restrict them?

Samoed: No. I've created similar collator as was for audio

mteb/tasks/video/retrieval/eng/msr_vtt.py
KennethEnevoldsen: Lets make an issue on it (this seems like an "omni" change)

mteb/models/model_implementations/random_baseline.py
KennethEnevoldsen: added

mteb/tasks/video/retrieval/eng/msr_vtt.py
Samoed: Should we support one video per item and add support for multiple videos later? I think all our current tasks would have only 1 video per row

AdnanElAssadi: Yes, we don't need multiple videos for now; we can stick to what you already did.

Samoed: Currently, it's implemented as list, but I don't think that this is good approach now. @KennethEnevoldsen WDYT?

AdnanElAssadi:
@KennethEnevoldsen Can you check this please? I think this is blocking integration of tasks and models.

mteb/tasks/video/retrieval/eng/msr_vtt.py
KennethEnevoldsen: We would ideally want multiple videos pr. item. I would probably favor going for the one that we want long-term.

Samoed: Should we add support for both `list[VideoItem]` and `VideoItem`? I think one item is required to support tasks that define video in one column

KennethEnevoldsen: Hmm, on a second thought - let us keep it at a single item for now - I think we have other fish to fry

mteb/tasks/video/retrieval/eng/msr_vtt.py
KennethEnevoldsen: remove this - otherwise good

PR #4149, opened 2026-02-24
Fix Error in Performance Per Model Size Plot in MIEB

ayush1298:
Before: 
image

After: 

image

### DIFF
--- a/mteb/benchmarks/benchmark.py
+++ b/mteb/benchmarks/benchmark.py
@@ -212,10 +212,9 @@ class MIEBBenchmark(Benchmark):
     ) -> pd.DataFrame:
         from mteb.benchmarks._create_table import _create_summary_table_mean_task_type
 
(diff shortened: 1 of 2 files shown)


mteb/benchmarks/benchmark.py
```diff
@@ -212,7 +212,10 @@ def _create_summary_table(
     ) -> pd.DataFrame:
         from mteb.benchmarks._create_table import _create_summary_table_mean_task_type
 
-        return _create_summary_table_mean_task_type(benchmark_results)
+        joint_table = _create_summary_table_mean_task_type(benchmark_results)
+        if "Mean (TaskType)" in joint_table.columns:
```
Samoed: I don't thunk that renaming column is good idea. I think it would be better to make customizable column name in `_create_summary_table_mean_task_type`

ayush1298: Added now

PR #4167, opened 2026-02-25
Added Embedding Parameter Value for Image and Audio Models

ayush1298:
Added Values for Image and Audio Models. Previously Audio branch was not merged, so this was missed and some Image Models were also missed.

For Image Models, I have used `vocab_size_text x hidden_dim_text `, not considered vocab_size_visual x hidden_dim_visual


Samoed:
Can you update tests?

ayush1298:
> Can you update tests?

Yes, updated now.

mteb/models/model_implementations/speecht5_models.py
Samoed, line 1: I see that in their config they have `vocab_size` 86, but I don't think that this is correct. Let's keep as None

mteb/models/model_implementations/sewd_models.py
Samoed, line 1: Here I'm not sure also

mteb/models/model_implementations/mme5_models.py
```diff
@@ -17,7 +17,7 @@
     release_date="2025-02-12",
     languages=["eng-Latn"],
     n_parameters=10_600_000_000,
-    n_embedding_parameters=None,  # 10.6B
+    n_embedding_parameters=525336576,  # 10.6B
```
Samoed:
```suggestion
    n_embedding_parameters=525336576,
```

mteb/models/model_implementations/wavlm_models.py
Samoed, line 1: Here I'm also unsure about embedding params

ayush1298:
> Can we update `n_embedding_parameters` only for models with text modality, because otherwise I'm not sure how it's accurate

Mostly added models are audio and Image only. I will remove above one, which you have pointed out.
What's your thought on this @KennethEnevoldsen ?

mteb/models/model_implementations/clap_models.py
```diff
@@ -141,7 +141,8 @@ def encode(
     revision="cca9e288ab447cee67d9ada1f85ddb46500f1401",
     release_date="2023-05-22",
     modalities=["audio", "text"],
-    n_parameters=153_507_530,  # Calculated using torch.numel(model.parameters())
+    n_parameters=153_507_530,
+    n_embedding_parameters=38603520,  # Calculated using torch.numel(model.parameters())
```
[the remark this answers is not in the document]
KennethEnevoldsen: @ayush1298

KennethEnevoldsen:
Hmm maybe it is worth making an assumption that there are no embedding parameters in image and audio models (without actually knowing too much about these models). E.g. WavLM doesn't have any non-active parameters as far as I can discern from the paper (the mask token vector would be the exception but that is a trivial amount)

mteb/models/model_implementations/clap_models.py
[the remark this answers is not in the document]
ayush1298: I got  `n_embedding_parameters=38603520` based on `vocab_size * hidden_size`. So, should I add that?

ayush1298:
> Hmm maybe it is worth making an assumption that there are no embedding parameters in image and audio models (without actually knowing too much about these models). E.g. WavLM doesn't have any non-active parameters as far as I can discern from the paper (the mask token vector would be the exception but that is a trivial amount)

So, then for which models we should add values? as mostly image and audio models only in this PR. Also, then avoid both these modality in tests, but then this having no values will cause issue in plots.

mteb/models/model_implementations/clap_models.py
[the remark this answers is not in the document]
KennethEnevoldsen: No - I would just remove the comment

mteb/models/model_implementations/clap_models.py
```diff
@@ -177,7 +178,8 @@ def encode(
     revision="8fa0f1c6d0433df6e97c127f64b2a1d6c0dcda8a",
     release_date="2023-05-22",
     modalities=["audio", "text"],
-    n_parameters=153_492_890,  # Calculated using torch.numel(model.parameters())
+    n_parameters=153_492_890,
+    n_embedding_parameters=38603520,  # Calculated using torch.numel(model.parameters())
```
KennethEnevoldsen:
```suggestion
    n_embedding_parameters=38603520,
```

mteb/models/model_implementations/data2vec_models.py
```diff
@@ -127,6 +127,7 @@ def encode(
     release_date="2022-02-07",  # Paper release date
     max_tokens=None,
     n_parameters=93_164_288,
+    n_embedding_parameters=24576,
```
KennethEnevoldsen: I would assume this is 0

KennethEnevoldsen: same below

mteb/models/model_implementations/hubert_models.py
```diff
@@ -126,6 +126,7 @@ def encode(
     release_date="2021-06-14",  # Paper release date
     max_tokens=float("inf"),
     n_parameters=95_000_000,
+    n_embedding_parameters=24576,
```
KennethEnevoldsen: also suspect this is 0

KennethEnevoldsen: and below

mteb/models/model_implementations/wavlm_models.py
KennethEnevoldsen: probably also 0

tests/test_models/test_model_meta.py
```diff
@@ -94,81 +81,39 @@
...
-    "VAGOsolutions/SauerkrautLM-ColMinistral3-3b-v0.1",
-    "VAGOsolutions/SauerkrautLM-ColQwen3-1.7b-Turbo-v0.1",
-    "VAGOsolutions/SauerkrautLM-ColQwen3-2b-v0.1",
-    "VAGOsolutions/SauerkrautLM-ColQwen3-4b-v0.1",
-    "VAGOsolutions/SauerkrautLM-ColQwen3-8b-v0.1",
     "MCINext/Hakim",
     "MCINext/Hakim-small",
     "MCINext/Hakim-unsup",
     # audio models
     "google/vggish",
```
KennethEnevoldsen: I would just set n_embeddings_parameters on these to zero (for any model that doesn't have text)

mteb/models/model_implementations/clap_models.py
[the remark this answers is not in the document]
ayush1298: Done

ayush1298:
Will do all these changes. Should I add all these models to Historic models or update tests to exclude models with modalities different from text? So, tests will consider models with only single modality `text`?

tests/test_models/test_model_meta.py
ayush1298: What to do for models with both `Image and Text` Modality?

Samoed: You can compute embedding parameters for them

Samoed:
> So, tests will consider models with only single modality text?

If model support text

ayush1298:
> If model support text

I think we should apply tests only when text is single modality, as there are some models where we have more some other modality along with text, but we don't have embedding parameter values. Or, we can keep those models in `HISTORIC_MODELS`?

ayush1298:
> Overall looks good, but I'm not sure if we want to keep `n_embedding_parameters=0`

> Hmm maybe it is worth making an assumption that there are no embedding parameters in image and audio models (without actually knowing too much about these models).

I think its better to keep it as 0 given the ambiguity that @KennethEnevoldsen mentioned.

ayush1298:
I have updated tests and merge main. I think it's ready to merge. Could you have a review it again? @KennethEnevoldsen @Samoed

Samoed:
Agree on tests. Probably we can more filters for `mteb.get_model_meta`

KennethEnevoldsen (approved):
Looks great - @Samoed added an issue on filter

PR #4170, opened 2026-02-25
feat: Add Matryoshka support when loading a model

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/2832

Add Matryoshka support. Extended `ModelMeta.embedding_dim` to support list for different embedding dimensions. Example:
```python
import mteb

model = mteb.get_model("jinaai/jina-embeddings-v5-text-nano", embed_dim=32)
```


mteb/models/get_model_meta.py
```diff
@@ -104,13 +106,14 @@ def get_model(
...
         device: Device used to load the model
+        embed_dim: Optional embedding dimension to load the model with. This is only used for models that support loading with a specified embedding dimension, and will be ignored for other models.
         **kwargs: Additional keyword arguments to pass to the model loader
 
     Returns:
         A model object
     """
     meta = get_model_meta(model_name, revision).model_copy(deep=True)
-    model = meta.load_model(device=device, **kwargs)
+    model = meta.load_model(device=device, embed_dim=embed_dim, **kwargs)
```
isaac-chung, line 116: What do you think about letting embed_dim be part of kwargs?

Samoed: I've added it to args of `get_model` for more visibility of this feature

mteb/models/model_implementations/random_baseline.py
```diff
@@ -155,11 +155,11 @@ def __init__(
         revision: str | None,
         array_framework: Literal["numpy", "torch"] = "numpy",
         dtype: torch.dtype | np.floating = np.float32,
-        embed_dim: int = _EMBEDDING_DIM,
+        embed_dim: int | None = None,
```
KennethEnevoldsen, line 214: Why the change here?

KennethEnevoldsen (commented):
Ah great addition! Can the following tests:

1) Ensure that it works with experiments
2) ensure that the metadata saves when running a model corresponds to the embedding dimension used. Currently I think it will save a list

mteb/models/model_implementations/random_baseline.py
Samoed: To match with `SentenceTransformerWrapper`

KennethEnevoldsen (approved):
great! We now only need to docs update, I would suggest the whats new section and probably also in the [models section](https://embeddings-benchmark.github.io/mteb/get_started/usage/defining_the_model/#using-a-pre-defined-model) as a subsection

Samoed:
Added docs

Issue #4176, opened 2026-02-27
Allow `ocr_text` in image benchmarks to allow for evaluation of text models


KennethEnevoldsen:
### Description of the feature

Currently it it not possible to evaluate text-only models on e..g vidore, even though the official datasets do have a markdown extraction. It would be possible to add this as an optional field using `ocr_text` or `image_ocr`.

---

An alternative way to add this is using wrappers:

```
vidore_task = ...

model = DoclingWrapper(bge_embedding_model)
results = mteb.evaluate(model, vidore_task)
```

Samoed:
I think we can create a new task like `OcrVidore` and then include this field as text. I don't think we should add new retrieval specific elements to input

KennethEnevoldsen:
But wouldn't we want to fairly compare if image only models can outcompete text only models. 

We also have the case that many image + text models obtain better performance using both:

<img width="566" height="216" alt="Image" src="https://github.com/user-attachments/assets/1f4a5e60-2b9f-40ae-a355-593aa86cbe6a" />

source: https://huggingface.co/llamaindex/vdr-2b-multi-v1

The counter argument against using the `text` field for OCR is that it makes it impossible to differentiate between text with images and an image with ocr'ed text (it also makes the it impossible to use both text with images + ocr).

The Docling wrapper approach is of course also an option, then we can treat it as an experiment: `model-name (ocr=True)`. A way to prevent this could be using some sort of cache:

```
vidore_task = ...

model = DoclingWrapper(bge_embedding_model, use_remote_cache="if-available") # downloads OCR data from remote if possible
results = mteb.evaluate(model, vidore_task)
```

This would require us to push the dataset to the hub.

KennethEnevoldsen:
@Samoed also added a example solution here #4184, which doesn't require any change to MTEB itself, but does include the OCR cost

KennethEnevoldsen:
I will close this in favor of #4193

Issue #4182, opened 2026-02-28
`"text"` is not included in `it2it` tasks


KennethEnevoldsen:
When running this dataset `EncyclopediaVQAIT2ITRetrieval`:

We can check the dataloader:
```py
dl = _create_queries_dataloader(
            dataset,
            task_metadata,
            batch_size=batch_size,
            input_column=input_column,
            num_proc=num_proc,
        )
next(iter(dl)).keys() # dict_keys(['image'])
```

even though the dataset:
```
Dataset({
    features: ['id', 'modality', 'text', 'image'],
    num_rows: 3743
})
```

@Samoed I assume that this is not desired behaviour would expect `# dict_keys(['image', 'text'])`

KennethEnevoldsen:
In general I had a look at the dataloader generation code and I would probably suggest a refactor where we do all the dataset specific transformation together, and then afterwards do the dataloader conversion e.g. using a function like:

```py
def _prepare_multimodal_dataset(
        dataset: Dataset,
    task_metadata: TaskMetadata,
    input_column: str | None = None, # TODO: why is this unused? 
    batch_size: int = 32, # TODO: I would generally not set the default at multiple layers - 
    # in case we don't pass it correctly, then this would not fails, but give an incorrect batch size
    num_proc: int | None = None,
):
    document_type = task_metadata.get_modalities(PromptType.document)
    dataset_cols = dataset.column_names
    transform = None # I don't see a case where this is not None am I missing something?
    collate_fn=_custom_collate_fn # TODO: same here - I would imagine this should be an input - might be overwritten later in which case this is fine

    if "text" in document_type: # if we want to fully generalize this, we should fork here for prompt type
        new_ds = dataset.map(
            _corpus_to_dict,
            desc="Standardizing text corpus format",
            num_proc=num_proc,
        )
        assert all(cn in new_ds.column_names for cn in dataset_cols), "ensure that no columns are removed" # TODO remove

    if "image" in document_type:
        new_ds = dataset.map(
            _convert_images_to_rgb,
            fn_kwargs={"image_col_name": "image", "transform": transform},
            desc="Converting images to RGB",
            num_proc=num_proc,
        )

    if "audio" in document_type:
        pass # seemingly no processing is needed or if so this is done in the collate_fn
    return new_ds # returned to a function that then handles the dataloader conversion
```

Samoed:
Yes, agree

PR #4183, opened 2026-02-28
fix: Enable OCR'ed text from images

KennethEnevoldsen:
Allow datasets like the vidore dataset to include OCR'ed text as a default.

I tested this using the following:

```py
from typing import Any

import mteb
from mteb.tasks import Vidore3HrRetrieval

metadata = Vidore3HrRetrieval.metadata
metadata.category = "t2it"  # add text modality to the dataset

class Vidore3HrRetrievalwithOCR(Vidore3HrRetrieval):
    metadata = metadata  # TODO: I would argue that the task is still "t2i", but it has support for text maybe "t2i(t)"?

    def dataset_transform(self, num_proc: int | None = None, **kwargs: Any) -> None:
        # add dummy OCR data to a dataset
        for lang, lang_subset in self.dataset.items():
            for split, ds in lang_subset.items():
                corpus = ds["corpus"]
                corpus = corpus.add_column("image_ocr_text", corpus["id"])
                self.dataset[lang][split]["corpus"] = corpus

task = Vidore3HrRetrievalwithOCR()

# choose an text only models
model = mteb.get_model_meta(
    "sentence-transformers/static-similarity-mrl-multilingual-v1"
)
res = mteb.evaluate(model, task, cache=None)
res[0].get_score() # np.float64(0.0052383333333333336)
```

Once we have agreed on this we should be able to redo the dataset to include the ocr markdowns 

related to #4176

### DIFF
--- a/mteb/types/_encoder_io.py
+++ b/mteb/types/_encoder_io.py
@@ -107,7 +107,9 @@ class ImageInput(TypedDict):
     """
 
     image: list[Image.Image]
-    image_ocr_text: NotRequired[list[str]] # TODO: we could rename to "image_text" to also allow for image captions?
+    image_ocr_text: NotRequired[
+        list[str]
+    ]  # TODO: we could rename to "image_text" to also allow for image captions?
 
 
 class AudioInputItem(TypedDict):


mteb/_create_dataloaders.py
```diff
@@ -341,12 +341,15 @@ def _create_image_dataloader(
     Returns:
         A DataLoader with the image dataset.
     """
+    columns_to_keep = ["image"]
+    if "image_ocr_text" in dataset.column_names:
```
Samoed, line 345: I think we should check for text and audio

Samoed (commented):
I don't think that this is good approach because this would be hard to find if task was run using text or not. We can just a new tasks with ocr text column included without refactoring core part

KennethEnevoldsen:
> I don't think that this is good approach because this would be hard to find if task was run using text or not. We can just a new tasks with ocr text column included without refactoring core part

What alternative do you suggest? (isn't this documented in the model implementation?)

We can of course update the version of the dataset to v2 `Vidore3HrRetrieval.v2`.

mteb/_create_dataloaders.py
KennethEnevoldsen: I don't think that is related to this fix but I did discover some issues on this and made an issue on #4182

KennethEnevoldsen: I don't think that it for this issue though probably related to #4182

Samoed:
> What alternative do you suggest?

Create new version of vidore tasks to make them `t2it`. This approach would help to evaluate all models on same data without additional processing

KennethEnevoldsen:
> Create new version of vidore tasks to make them t2it. This approach would help to evaluate all models on same data without additional processing

I agree with this on the short term, but I don't think it is a general problem as something like this would not be possible:

```
{"text: "as seen in figure 1 <image1>", "image" image1, "image_text": "{ocr from image}"}
```

However I think that is a future problem

Samoed:
Yes, but for now I think ViDoRe team want to use their markdown field, so I think this is fine for now

KennethEnevoldsen:
but this PR adds the `image_ocr_text` (eq. to `image_text` from the example), which would allow us to have both.

Samoed:
I don't think that manipulating with task metadata and changing `dataset_transform` is good solution, because this also would require to change model's code

KennethEnevoldsen:
Agree - I will close this for now due to #4191

PR #4184, opened 2026-02-28
Example solution for Vidore OCR as a wrapper

KennethEnevoldsen:
If we implement the OCR solution as a wrapper this is how it could look like. Note that I do not implement the actual OCR here but just fetches it from the dataset.

An alternative to #4183 and potential fix for #4176

### DIFF
--- a/vidore-with-docling-example.py
+++ b/vidore-with-docling-example.py
@@ -1,0 +1,96 @@
+import io
+
+from docling.datamodel.base_models import DocumentStream
+from docling.document_converter import DocumentConverter
+
+import mteb
+from mteb.models import EncoderProtocol
+from mteb.models.model_meta import ModelMeta
+from mteb.tasks import Vidore3HrRetrieval
+
+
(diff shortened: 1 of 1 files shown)


vidore-with-ocr-example.py
```diff
@@ -0,0 +1,73 @@
...
+
+import mteb
+from mteb.models import EncoderProtocol
+from mteb.models.model_meta import ModelMeta
+from mteb.tasks import Vidore3HrRetrieval
+
+meta = Vidore3HrRetrieval.metadata
+
+
+class VidoreOCRFetchWrapper(EncoderProtocol):
```
Samoed, line 11: For this case I think easier to make new task version. Downside that multimodal models would receive text and image together, but I think this would be much easier to maintain and validate

KennethEnevoldsen:
Closing this in favor of the other solution which I think is better, but might need some discussion (#4176)

PR #4185, opened 2026-03-01
fix: remove select column from dataloader

Samoed:
partly adresses https://github.com/embeddings-benchmark/mteb/issues/4182


KennethEnevoldsen (approved):
While this is a fix for some of the dataset (so we should merge) - I don't think it generally solves the issue.

E.g. corpus conversion does not happen. I have changed to just link the issue

Issue #4169, opened 2026-02-25
model: model nvidia/llama-nemotron-embed-vl-1b-v2


KennethEnevoldsen:
https://huggingface.co/nvidia/llama-nemotron-embed-vl-1b-v2

 _Originally posted by @KennethEnevoldsen in [#3127](https://github.com/embeddings-benchmark/mteb/issues/3127#issuecomment-3959849314)_

KennethEnevoldsen:
someone is already working on this

PR #4192, opened 2026-03-03
model: nvidia/llama-nemotron-embed-vl-1b-v2 for ViDoRe

gabrielspmoreira:

Close https://github.com/embeddings-benchmark/mteb/issues/4169

### DIFF
--- a/mteb/models/model_implementations/nvidia_nemotron_vl_models.py
+++ b/mteb/models/model_implementations/nvidia_nemotron_vl_models.py
@@ -367,14 +367,16 @@ class LlamaNemotronEmbedVL(AbsEncoder):
         device_map="cuda",
         torch_dtype=torch.bfloat16,
         attn_implementation="flash_attention_2",
-        use_image_modality: bool = False,
+        use_image_modality: bool = True,
         use_text_modality: bool = True,
         **kwargs,
     ):
         self.use_image_modality = use_image_modality
         self.use_text_modality = use_text_modality
         if not self.use_image_modality and not self.use_text_modality:
-            raise ValueError("At least one of use_image_modality or use_text_modality must be True")
+            raise ValueError(
+                "At least one of use_image_modality or use_text_modality must be True"
+            )
 
         install_hint = f"pip install 'mteb[{extra_name}]'"
 


mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -6,11 +6,12 @@
...
 from torch.utils.data import DataLoader
 from transformers import __version__ as transformers_version
-
 from mteb._requires_package import requires_package
 from mteb.models.abs_encoder import AbsEncoder
 from mteb.models.model_meta import ModelMeta
-
+from mteb.types import PromptType
+from torch.nn.functional import normalize
+from tqdm import tqdm
```
Samoed:
```suggestion
from tqdm.auto import tqdm
```

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -45,6 +54,7 @@
 _TRANSFORMERS_CONSTRAINTS: dict[str, str] = {
     "llama-nemotron-colembed-vl": "==4.49.0",  # llama-nemoretriever-colembed-*
     "nemotron-colembed-vl-v2": "==5.0.0",  # nemotron-colembed-vl-4b-v2, nemotron-colembed-vl-8b-v2
+    "llama-nemotron-embed-vl-1b-v2": ">=4.56.0", # llama-nemotron-embed-vl-1b-v2
```
Samoed: Is it working with v5?

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +353,133 @@ def encode(
...
+
+TRAINING_DATA_EMBED_VL_1B_V2 = {
+    "VidoreDocVQARetrieval",
+    "VidoreInfoVQARetrieval",
+    "VidoreTatdqaRetrieval",
+    "VidoreArxivQARetrieval",
+    "docmatix-ir",
+    "wiki-ss-nq",
+    "Cauldron (AI2D, OCRVQA, Websight)",
+    "vdr-multilingual-train dataset (vdr_en)",
```
Samoed:
```suggestion
    "VDRMultilingualRetrieval",
```

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +353,133 @@ def encode(
...
+    "HotpotQA",
+    "MIRACLRetrieval",
+    "NQ",
+    "StackExchangeClustering",
+    "SQuAD",
+    "MLDR",
```
Samoed:
```suggestion
    "MultiLongDocRetrieval",
```

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +353,133 @@ def encode(
...
+    "MLQA",
```
Samoed:
```suggestion
    "MLQARetrieval",
```

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +353,133 @@ def encode(
...
+        show_progress_bar: bool = True,
+        **kwargs: Any,
+    ) -> Array:       
+        with torch.inference_mode():    
+            embeddings_list = []
+            message = f"Extracting {prompt_type.name} embeddings..."
+            for batch in tqdm(inputs, desc=f"Extracting {message} embeddings...", disable=not show_progress_bar):
+                if prompt_type == PromptType.query:
+                    embeddings = self.model.encode_queries(batch["text"])
+                elif prompt_type == PromptType.document:
```
Samoed:
```suggestion
                else:
```
For non-retrieval/reranking tasks prompt_type will be None

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +353,133 @@ def encode(
...
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        show_progress_bar: bool = True,
+        **kwargs: Any,
+    ) -> Array:       
+        with torch.inference_mode():    
+            embeddings_list = []
+            message = f"Extracting {prompt_type.name} embeddings..."
+            for batch in tqdm(inputs, desc=f"Extracting {message} embeddings...", disable=not show_progress_bar):
```
Samoed:
```suggestion
            for batch in tqdm(inputs, desc=f"Extracting {prompt_type} embeddings...", disable=not show_progress_bar):
```

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
oliverholworthy: Yes it works correctly with 5.0.0 and above as well as 4.56.x and 4.57.x

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +355,141 @@ def encode(
...
+    release_date="2026-01-06",
+    modalities=["image", "text"],
+    n_parameters=1_678_252_480,
+    n_embedding_parameters=262_688_768,
+    memory_usage_mb=6402,
+    max_tokens=10240,
+    embed_dim=2048,
+    license="https://www.nvidia.com/en-us/agreements/enterprise-software/nvidia-open-model-license/",
+    open_weights=True,
+    public_training_code="Proprietary Code",
```
Samoed:
```suggestion
    public_training_code=None,
```

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +355,141 @@ def encode(
...
+        # Sets the number of tiles the image can be split into
+        self.model.processor.max_input_tiles = 4
+
+    def encode(
+        self,
+        inputs: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        gf_subset: str,
```
Samoed:
```suggestion
        hf_subset: str,
```

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +355,141 @@ def encode(
...
+        **kwargs: Any,
+    ) -> Array:
+        with torch.inference_mode():
+            embeddings_list = []
+            for batch in tqdm(
+                inputs,
+                desc=f"Extracting {prompt_type} embeddings...",
+                disable=not show_progress_bar,
+            ):
+                if prompt_type == PromptType.query:
```
Samoed: We have datasets where query can be image or audio or text+image etc
```suggestion
                if prompt_type == PromptType.query and "text" in batch:
```

KennethEnevoldsen:
@gabrielspmoreira can you take a look at @Samoed comment. Once that is considered then we can merge this

mteb/models/model_implementations/nvidia_nemotron_vl_models.py
```diff
@@ -343,3 +355,153 @@ def encode(
...
+        self,
+        model_name_or_path: str,
+        revision: str,
+        trust_remote_code: bool,
+        extra_name: str = "llama-nemotron-embed-vl-1b-v2",
+        device_map="cuda",
+        torch_dtype=torch.bfloat16,
+        attn_implementation="flash_attention_2",
+        use_image_modality: bool = False,
+        use_text_modality: bool = True,
```
Samoed: I don't think that disabling one of modalities by default is expected behavior for models

KennethEnevoldsen: I could see it as a routing thing. Like when there is text+images then use just the text, but this will def. punish the model on other tasks where the text doesn't provide enough information (I don't think that reflects the intended use).

KennethEnevoldsen: You can submit text-only as an experiment ([docs](https://embeddings-benchmark.github.io/mteb/whats_new/#experiments))

gabrielspmoreira: I see @Samoed . I did that so that users can reproduce the [results PR](https://github.com/embeddings-benchmark/results/pull/432) I submitted, which were run with these default values (use_text_modality = True, use_image_modality = False). The reason is that for Vidore V3, text-only modality provides slightly higher accuracy than image modality, but the embedding throughput is much higher. providing a better trade-off for this dataset.  Does that make sense?

Samoed: Yes, that make sense, but I think model should use all modalities by default

boliu61: Hi @Samoed , I think each model should use their best modality. 

E.g. we implemented and evaluated Qwen3-VL-Embedding in three modalities: image only, image + text, text only. The best one is image only (higher accuracy than image + text), so we reported image only results and made image only as the default (see https://github.com/embeddings-benchmark/mteb/pull/4198)

Samoed: I understand this, but this would create problems for users that would run text only tasks. Your approach is vidore focused, but I think we should use more general approach when we're adding models

KennethEnevoldsen: I feel like the naive application of this model to vidore is with the text+image and then the variation is with text-only. 

The models in mteb are intended to be a general implementation of the model. This implementation currently would not work for an image task, but its metadata states otherwise.

You are free to implement it so that, given any task with both image and text, the model encodes only the text, but that will then be the reference for that model that people will re-use again and again. 

I really don't believe that it is the desired general implementation. I agree with @Samoed that the experiment solution offers the best alternative, allowing the default to be reasonable while still allowing optimization for Vidore

gabrielspmoreira: @KennethEnevoldsen @Samoed we have followed your suggestions. Default modality for Nemotron Embed VL 1B is is image + text (when available). Results with text-only were submitted as an experiment

KennethEnevoldsen (approved):

Issue #3895, opened 2026-01-08
Add model: Qwen3-VL


Samoed:
### Model link on Hugging Face

https://huggingface.co/collections/Qwen/qwen3-vl-embedding

ayush1298:
@Samoed I saw your comment: https://github.com/embeddings-benchmark/mteb/issues/3939#issuecomment-3754235507
But didn't get that exctaly, are you saying that its not implemented in MTEB because we can load the model using: 

```
from transformers import AutoProcessor, AutoModelForImageTextToText

processor = AutoProcessor.from_pretrained("Qwen/Qwen3-VL-Embedding-2B")
model = AutoModelForImageTextToText.from_pretrained("Qwen/Qwen3-VL-Embedding-2B")
```

Samoed:
It will work, but this can produce different resutls from implementation that they have in scripts https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B/blob/main/scripts/qwen3_vl_embedding.py. Because if load with `AutoModel` `Qwen3VLPreTrainedModel` will be used, but in scripts they're inherited from it. Maybe @8igfive can tell us what's different and their plans about releasing their evaluation scripts

ayush1298:
Got it. thanks for explaination.

8igfive:
Using both the script https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B/blob/main/scripts/qwen3_vl_embedding.py and `from transformers import AutoProcessor, AutoModelForImageTextToText` will actually produce consistent results, as shown in the example below:

```
import torch
from src.models.qwen3_vl_embedding import Qwen3VLEmbedder # from the script
from qwen_vl_utils.vision_process import process_vision_info
from transformers import AutoProcessor, AutoModelForImageTextToText

model_path = "Qwen/Qwen3-VL-Embedding-2B"

inputs = [{
    "text": "A woman shares a joyful moment with her golden retriever on a sun-drenched beach at sunset, as the dog offers its paw in a heartwarming display of companionship and trust.",
    "image": "https://qianwen-res.oss-cn-beijing.aliyuncs.com/Qwen-VL/assets/demo.jpeg"
}]
pr_model = Qwen3VLEmbedder(model_path)
with torch.inference_mode():
    pr_embeddings = pr_model.process(inputs)


conversation = [{
    'role': 'system', 'content': [{'type': 'text', 'text': "Represent the user's input."}]
}, {
[...]

PR #4198, opened 2026-03-04
model: Qwen3-VL-Embedding

ybabakhin:
This PR adds support for:
* [Qwen/Qwen3-VL-Embedding-2B](https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B)
* [Qwen/Qwen3-VL-Embedding-8B](https://huggingface.co/Qwen/Qwen3-VL-Embedding-8B)

Close https://github.com/embeddings-benchmark/mteb/issues/3895

### DIFF
--- a/mteb/models/model_implementations/qwen3_vl_embedding_models.py
+++ b/mteb/models/model_implementations/qwen3_vl_embedding_models.py
@@ -157,11 +157,11 @@ class Qwen3VLEmbeddingWrapper(AbsEncoder):
     ):
         requires_image_dependencies()
         requires_package(
-            self, "transformers", model_name, "pip install transformers>=4.57.0"
+            self, "transformers", model_name, "pip install 'mteb[qwen-vl]'"
         )
 
         requires_package(
-            self, "qwen_vl_utils", model_name, "pip install qwen-vl-utils>=0.0.14"
+            self, "qwen_vl_utils", model_name, "pip install 'mteb[qwen-vl]'"
         )
 
         from transformers.models.qwen3_vl.processing_qwen3_vl import Qwen3VLProcessor
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,7 +101,7 @@ open_clip_torch = ["open_clip_torch==2.31.0"]
 nomic = ["einops>=0.8.1"]
 ark = ["volcengine-python-sdk[ark]==3.0.2", "tiktoken>=0.8.0"]
 colpali_engine = ["colpali_engine>=0.3.12; python_full_version < '3.14'"]
-colqwen3 = ["transformers>=4.57", "torchvision>=0.22.1"]
+colqwen3 = ["transformers>=4.57.0", "torchvision>=0.22.1"]
 sauerkrautlm-colpali = ["sauerkrautlm-colpali>=0.1.0; python_full_version < '3.14'"]
 xet = ["huggingface_hub>=0.32.0"]
 youtu = ["tencentcloud-sdk-python-common>=3.0.1454", "tencentcloud-sdk-python-lkeap>=3.0.1451"]
@@ -116,6 +116,7 @@ wav2clip = ["wav2clip==0.1.0"]
 torch-vggish-yamnet = ["torch-vggish-yamnet==0.2.1"]
 vllm = ["vllm>=0.11.1"]
 mctct = ["transformers<5"]  # mctct was removed in transformers 5
+qwen-vl = ["transformers>=4.57.0", "qwen-vl-utils>=0.0.14"] 
 
 [dependency-groups]
 lint = [
@@ -386,6 +387,7 @@ conflicts = [
         { extra = "msclap" },
         { extra = "muq" },
         { extra = "mctct" },
+        { extra = "qwen-vl" },
     ],
 ]
 


Samoed:
Great! Can you reproduce scores from paper?

ybabakhin:
> Great! Can you reproduce scores from paper?

Hi @Samoed! Yes, we reproduced Average ViDoRe-v3 number for 2B model from the paper here: https://github.com/embeddings-benchmark/results/pull/433

By default, it uses only image as an input (if available), but with an additional flag could embed image + text. According to our experiments, `image + text` mode works worse compared to `image-only` on ViDoRe-v3.
```
model = mteb.get_model("Qwen/Qwen3-VL-Embedding-2B", visual_document_use_text=False)
```

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -0,0 +1,429 @@
...
+            past_key_values: Cache | None = None,
+            inputs_embeds: torch.FloatTensor | None = None,
+            pixel_values: torch.Tensor | None = None,
+            pixel_values_videos: torch.FloatTensor | None = None,
+            image_grid_thw: torch.LongTensor | None = None,
+            video_grid_thw: torch.LongTensor | None = None,
+            cache_position: torch.LongTensor | None = None,
+            **kwargs,
+        ) -> tuple | Qwen3VLForEmbeddingOutput:
+            self.model.rope_deltas = None
```
Samoed, line 115: Can you add comment for this? I can't find this in their implementation https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B/blob/main/scripts/qwen3_vl_embedding.py#L172

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -0,0 +1,429 @@
...
+        return torch.cat(all_embeddings, dim=0)
+
+
+qwen3_vl_embedding_2b = ModelMeta(
+    loader=Qwen3VLEmbeddingWrapper,
+    name="Qwen/Qwen3-VL-Embedding-2B",
+    model_type=["dense"],
+    languages=[],
+    open_weights=True,
+    revision="main",
```
Samoed:
```suggestion
    revision="2a50926d213628c727f38025982a76f655673f54",
```

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -0,0 +1,429 @@
...
+    citation=QWEN3_VL_EMBEDDING_CITATION,
+)
+
+qwen3_vl_embedding_8b = ModelMeta(
+    loader=Qwen3VLEmbeddingWrapper,
+    name="Qwen/Qwen3-VL-Embedding-8B",
+    model_type=["dense"],
+    languages=[],
+    open_weights=True,
+    revision="main",
```
Samoed:
```suggestion
    revision="418d03899aeb9d954e776dfa762006616519a914",
```

Samoed:
```suggestion
    revision="418d03899aeb9d954e776dfa762006616519a914",
```

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -0,0 +1,429 @@
...
+                all_embeddings.append(embeddings.cpu())
+
+        return torch.cat(all_embeddings, dim=0)
+
+
+qwen3_vl_embedding_2b = ModelMeta(
+    loader=Qwen3VLEmbeddingWrapper,
+    name="Qwen/Qwen3-VL-Embedding-2B",
+    model_type=["dense"],
+    languages=[],
```
Samoed: Let's add at least 
```suggestion
    languages=["eng-Latn"],
```

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -0,0 +1,429 @@
...
+    ) -> list[dict]:
+        """Format a single input as a conversation for the model.
+
+        Mirrors ``Qwen3VLEmbedder.format_model_input`` from the official
+        script: system message carries the instruction, user message
+        carries image (if any) followed by text (if any).
+        """
+        instruction = instruction or DEFAULT_INSTRUCTION
+        instruction = instruction.strip()
+        if instruction and not unicodedata.category(instruction[-1]).startswith("P"):
```
KennethEnevoldsen, line 209: I am not sure I get this one - can you explain why you check that it starts with "P"?
(maybe add a comment on it as well)

mteb/models/model_implementations/qwen3_vl_embedding_models.py
ybabakhin: While evaluating ViDoRe-v3 with text + image mode, for some of the samples we've got the error below:
```
File "/usr/local/lib/python3.12/dist-packages/transformers/models/qwen3_vl/modeling_qwen3_vl.py", line 1267, in forward
    position_ids = self.compute_3d_position_ids(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/transformers/models/qwen3_vl/modeling_qwen3_vl.py", line 1181, in compute_3d_position_ids
    position_ids = position_ids + delta.to(device=position_ids.device)
                   ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
RuntimeError: The size of tensor a (32) must match the size of tensor b (30) at non-singleton dimension 1
```

Looks like there are text and non-text handling in [compute_3d_position_ids](https://github.com/huggingface/transformers/blob/7d9754a05193eb79b1d86aa744b622b8068008cd/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L1150), and it was caching `self.rope_deltas` with the dimensions from the previous batch.

So, we set it to None as it should only be required in KV-cache for generation, and shouldn't affect embeddings.

mteb/models/model_implementations/qwen3_vl_embedding_models.py
ybabakhin: This is taken from the [original implementation](https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B/blob/main/scripts/qwen3_vl_embedding.py#L211).

I think, it checks if the last character is not punctuation and appends `.` then

KennethEnevoldsen: right

it is fairly broad category of punctuations, e.g. "_" is included ([docs](https://www.unicode.org/reports/tr44/#General_Category_Values)). 

if it was in the original implementation I though I would be fine with leaving it

Samoed: Can you add this comment to line?

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -320,12 +323,21 @@ def encode(
...
         contains_text = "text" in inputs.dataset.features
         contains_image = "image" in inputs.dataset.features
 
         if not contains_text and not contains_image:
             raise ValueError("No text or image features found in inputs.")
 
+        if (
+            prompt_type == PromptType.document
+            and not self.visual_document_use_text
+            and contains_image
```
gabrielspmoreira, line 338: @ybabakhin  I suggest removing this condition `and contains_image` and moving the command below from before to after 
```bash
if not contains_text and not contains_image:
    raise ValueError("No text or image features found in inputs.")
```
That way using text or nor does not depend on whether there is image available. And if neither image or text are available, an exception is raised

ybabakhin: @gabrielspmoreira I moved that exception block below, but kept `contains_image` in the condition. The reason to have it there is to make the model work with the usual Text Retrieval datasets. In that case:
* `prompt_type == PromptType.document` is True
* `not self.visual_document_use_text` is True by default
* So, we need `contains_image` which is False, in order not to set `contains_text = False`

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -0,0 +1,433 @@
...
+  journal={arXiv preprint arXiv:2601.04720},
+  year={2026}
+}"""
+
+IMAGE_BASE_FACTOR = 16
+IMAGE_FACTOR = IMAGE_BASE_FACTOR * 2
+MIN_PIXELS = 4 * IMAGE_FACTOR * IMAGE_FACTOR
+MAX_PIXELS = 1800 * IMAGE_FACTOR * IMAGE_FACTOR
+MAX_LENGTH = 8192
+DEFAULT_INSTRUCTION = "Represent the user's input."
```
KennethEnevoldsen, line 38: Assuming this is from the original implementation

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -0,0 +1,433 @@
...
+    ):
+        requires_image_dependencies()
+        requires_package(
+            self, "transformers", model_name, "pip install transformers>=4.57.0"
+        )
+
+        requires_package(
+            self, "qwen_vl_utils", model_name, "pip install qwen-vl-utils>=0.0.14"
+        )
+
```
KennethEnevoldsen, line 166: add an optional dependency so that it becomes `pip install mteb[qwen-vl]`, which includes all the dependencies for the model.

mteb/models/model_implementations/qwen3_vl_embedding_models.py
Samoed: https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B/blob/main/scripts/qwen3_vl_embedding.py#L147https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B/blob/main/scripts/qwen3_vl_embedding.py#L147

Samoed (approved):

Issue #3797, opened 2025-12-25
Add model: `PE-AV`


Samoed:
### Model link on Hugging Face

https://huggingface.co/facebook/pe-av-large

### Arxiv link

https://arxiv.org/abs/2512.19687

PR #4199, opened 2026-03-05
[MVEB] PE-AV Model, Kinetics400 Dataset, RavdessAV Dataset

AdnanElAssadi56:
(From closed PR)

Adds the following:
[mteb/kinetics-400](https://huggingface.co/datasets/mteb/kinetics-400)
[mteb/RAVDESS_AV](https://huggingface.co/datasets/mteb/RAVDESS_AV)
[PE-AV (Facebook)](https://huggingface.co/facebook/pe-av-large) Close https://github.com/embeddings-benchmark/mteb/issues/3797

Also includes some remaining components from the parallel video integration work we accidently did.

### DIFF
--- a/mteb/_create_dataloaders.py
+++ b/mteb/_create_dataloaders.py
@@ -16,7 +16,7 @@ from mteb.types import (
 from mteb.types._encoder_io import AudioInputItem
 
 if TYPE_CHECKING:
-    from collections.abc import Callable
+    from collections.abc import Callable, Sequence
 
     from torchcodec.decoders import VideoDecoder  # type: ignore[import-untyped]
 
@@ -301,7 +301,7 @@ def create_dataloader(
     *,
     task_metadata: TaskMetadata,
     prompt_type: PromptType | None = None,
-    input_column: str | None = None,
+    input_column: str | Sequence[str] | None = None,
     batch_size: int = 32,
     num_proc: int | None = None,
     **kwargs: Any,
@@ -315,7 +315,8 @@ def create_dataloader(
         dataset: The dataset to create a dataloader from.
         task_metadata: The metadata of the task.
         prompt_type: The type of prompt to create a dataloader for. If None, it will be inferred from the task metadata.
-        input_column: The column to use as input. If None, it will use the first column that matches the modality.
+        input_column: The column(s) to use as input. If a string, used for column renaming.
+            If a Sequence, columns are assumed to already match modality names. If None, inferred from task metadata.
         batch_size: The batch size for the dataloader.
         num_proc: The number of processes to use for dataset processing.
         **kwargs: Additional arguments to pass to the dataloader creation functions.
@@ -323,13 +324,16 @@ def create_dataloader(
     Returns:
         A dataloader for the dataset.
     """
+    # Sequence means columns already match modality names, no renaming needed
+    _input_column = input_column if isinstance(input_column, str) else None
+
     if (
         prompt_type is None
         and task_metadata.modalities == ["text"]
-        and input_column is not None
+        and _input_column is not None
     ):
         return _create_dataloader_from_texts(
-            dataset[input_column],
+            dataset[_input_column],
             batch_size=batch_size,
         )
 
@@ -337,7 +341,7 @@ def create_dataloader(
         dataset,
         task_metadata,
         prompt_type=prompt_type,
-        input_column=input_column,
+        input_column=_input_column,
         num_proc=num_proc,
     )
 
--- a/mteb/_evaluators/clustering_evaluator.py
+++ b/mteb/_evaluators/clustering_evaluator.py
@@ -53,9 +53,7 @@ class ClusteringEvaluator(Evaluator):
         data_loader = create_dataloader(
             self.dataset,
             task_metadata=self.task_metadata,
-            input_column=self.input_column_name
-            if isinstance(self.input_column_name, str)
-            else None,
+            input_column=self.input_column_name,
             num_proc=num_proc,
             **encode_kwargs,
         )
--- a/mteb/_evaluators/sklearn_evaluator.py
+++ b/mteb/_evaluators/sklearn_evaluator.py
@@ -66,22 +66,17 @@ class SklearnEvaluator(Evaluator):
         encode_kwargs: EncodeKwargs,
         num_proc: int | None,
     ) -> tuple[DataLoader[BatchedInput], DataLoader[BatchedInput]]:
-        input_col = (
-            self.values_column_name
-            if isinstance(self.values_column_name, str)
-            else None
-        )
         dataloader_train = create_dataloader(
             self.train_dataset,
             task_metadata=self.task_metadata,
-            input_column=input_col,
+            input_column=self.values_column_name,
             num_proc=num_proc,
             **encode_kwargs,
         )
         dataloader_test = create_dataloader(
             self.eval_dataset,
             task_metadata=self.task_metadata,
-            input_column=input_col,
+            input_column=self.values_column_name,
             num_proc=num_proc,
             **encode_kwargs,
         )
--- a/mteb/_evaluators/zeroshot_classification_evaluator.py
+++ b/mteb/_evaluators/zeroshot_classification_evaluator.py
@@ -53,13 +53,9 @@ class ZeroShotClassificationEvaluator(Evaluator):
         encode_kwargs: EncodeKwargs,
         num_proc: int | None = None,
     ) -> Array:
-        input_col = (
-            self.input_column_name if isinstance(self.input_column_name, str) else None
-        )
-
         dataloader = create_dataloader(
             self.dataset,
-            input_column=input_col,
+            input_column=self.input_column_name,
             task_metadata=self.task_metadata,
             num_proc=num_proc,
             **encode_kwargs,
@@ -76,7 +72,7 @@ class ZeroShotClassificationEvaluator(Evaluator):
 
         dataloader = create_dataloader(
             self.dataset,
-            input_column=input_col,
+            input_column=self.input_column_name,
             task_metadata=self.task_metadata,
             **encode_kwargs,
         )
--- a/mteb/abstasks/classification.py
+++ b/mteb/abstasks/classification.py
@@ -320,9 +320,7 @@ class AbsTaskClassification(AbsTask):
         dataloader_train = create_dataloader(
             ds,
             task_metadata=self.metadata,
-            input_column=self.input_column_name
-            if isinstance(self.input_column_name, str)
-            else None,
+            input_column=self.input_column_name,
             num_proc=num_proc,
             **encode_kwargs,
         )
--- a/mteb/abstasks/clustering.py
+++ b/mteb/abstasks/clustering.py
@@ -236,9 +236,7 @@ class AbsTaskClustering(AbsTask):
             create_dataloader(
                 downsampled_dataset,
                 task_metadata=self.metadata,
-                input_column=self.input_column_name
-                if isinstance(self.input_column_name, str)
-                else None,
+                input_column=self.input_column_name,
                 num_proc=num_proc,
                 **encode_kwargs,
             ),


mteb/abstasks/classification.py
```diff
@@ -193,7 +193,23 @@ def evaluate(
...
 
             if isinstance(ds, Dataset | DatasetDict):
-                ds = ds.select_columns([self.label_column_name, self.input_column_name])
+                # Keep label and input columns, plus any columns required by
+                # the task's declared modalities (e.g., audio for va2c tasks)
+                modality_to_column = {
+                    "video": "video",
+                    "audio": "audio",
+                    "image": "image",
+                }
```
Samoed: Can we just extend `input_column_name` to list?

mteb/models/model_implementations/pe_av_models.py
Samoed, line 1: Can you check if scores are reproducible?

mteb/models/model_implementations/pe_av_models.py
```diff
@@ -0,0 +1,366 @@
...
+
+        if has_video:
+            inputs.collate_fn = VideoCollator(
+                max_frames=self.num_frames,
+                target_sampling_rate=self.sampling_rate,
+            )
+        elif has_audio:
+            from mteb._create_dataloaders import AudioCollator
+
+            inputs.collate_fn = AudioCollator(self.sampling_rate)
```
Samoed: I think it's better to create multimodal collator

mteb/models/model_implementations/pe_av_models.py
```diff
@@ -0,0 +1,366 @@
...
+    reference="https://huggingface.co/facebook/pe-av-base",
+    **_PE_AV_COMMON,
+)
+
+pe_av_large = ModelMeta(
+    loader=PEAudioVisualWrapper,
+    name="facebook/pe-av-large",
+    revision="0d24878d4107d64bef49e53602fc34ce6f94f6d8",
+    n_parameters=2_200_000_000,
+    memory_usage_mb=None,
```
Samoed: Can you fill `memory_usage_mb` and `n_embedding_parameters`. You can compute them by `ModelMeta.from_hub(model_name)` or use `meta._estimate_embedding_parameters_from_hub` and `meta._calculate_memory_usage_mb`

mteb/models/model_implementations/pe_av_models.py
```diff
@@ -0,0 +1,366 @@
...
+                video_embeds = self.model.get_video_embeds(
+                    pixel_values_videos=processed["pixel_values_videos"],
+                    padding_mask_videos=processed.get("padding_mask_videos"),
+                )
+                video_embeds = video_embeds / video_embeds.norm(dim=-1, keepdim=True)
+                all_embeddings.append(video_embeds.cpu().float().numpy())
+
+        return np.vstack(all_embeddings)
+
+    def get_audio_embeddings(
```
Samoed, line 124: Can you add to functions?
```suggestion

    @torch.inference_mode()
    def get_audio_embeddings(
```

AdnanElAssadi:
@Samoed I edited the collator to handle one video item.

Samoed:
Yeah, I forgot to update collator after changing to one video

AdnanElAssadi:
Results from `"facebook/pe-av-small"`:

[RAVDESS_AVClustering.json](https://github.com/user-attachments/files/25765545/RAVDESS_AVClustering.json)

Samoed:
This is hard to match results for ravdess. I think you can run one of these tasks (from https://arxiv.org/pdf/2512.19687)
image

mteb/_create_dataloaders.py
```diff
@@ -742,25 +751,20 @@ def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
...
-                        ),
-                    )
-                )
-            row["video"] = video_inputs
+            video = row.pop("video")
+            frames = self.resample_video(video["frames"], self.max_frames)
+            audio = self.audio_collator.resample_audio(
+                video,
+                target_sampling_rate=self.audio_collator.target_sampling_rate,
+                max_samples=self.audio_collator.max_samples,
```
Samoed: I think we can also add `fps` for resampling

AdnanElAssadi:
MSR-VTT:
Some discrepancy but maybe because we are using audio as well.
[MSRVTTV2T.json](https://github.com/user-attachments/files/25780166/MSRVTTV2T.json)

AdnanElAssadi:
@Samoed Anything else here?

Samoed:
Can you resolve my comments and make CI green?

AdnanElAssadi:
I think tests are failing because of `n_embedding_parameters`. It can't be calculated by the method in Model Meta:
`Could not calculate embedding parameters for facebook/pe-av-base-16-frame as config.json could not be loaded`

mteb/_create_dataloaders.py
AdnanElAssadi: Maybe in another PR

Samoed:
Strange. I get it without problems
```python
import mteb

meta = mteb.models.ModelMeta.from_hub("facebook/pe-av-base-16-frame")
meta.n_embedding_parameters
# 51576832
```

mteb/_create_dataloaders.py
```diff
@@ -590,6 +591,14 @@ def create_dataloader(
             batch_size=batch_size,
             num_proc=num_proc,
         )
+    if "video" in task_metadata.modalities:
+        return _create_video_dataloader(
```
KennethEnevoldsen: Should we consider doing a more principled refactor here as discussed in #4182

That issue also showed how the current approach can lead to some odd interactions between modalities.

Samoed: Yes, I'll do it a bit later

mteb/abstasks/classification.py
AdnanElAssadi: This will also cause changes in dataloader; we can do this separately.

AdnanElAssadi:
@Samoed   Tests resolved here

mteb/abstasks/classification.py
Samoed: What changs? I think easier to use list for processing rather than processing like this

mteb/abstasks/task_metadata.py
```diff
@@ -201,6 +201,15 @@
     "Any2AnyRetrieval",
 )
 
+MVEB_TASK_TYPE = (
+    "VideoClassification",
+    "VideoClustering",
+    "VideoPairClassification",
+    "VideoZeroshotClassification",
+    "VideoCentricQA",
+    "Any2AnyRetrieval",
```
isaac-chung, line 207: Duplicated?
```suggestion
```

mteb/tasks/video/classification/eng/kinetics400_classification.py
```diff
@@ -0,0 +1,85 @@
...
+}
+""",
+    )
+
+    input_column_name: str = "video"
+    label_column_name: str = "label"
+
+    is_cross_validation: bool = False
+
+    def load_data(self, **kwargs) -> None:
```
isaac-chung: This is a lot of code to be repeated for every task that needs this. Can we move the combine modalities code in the AbsTask instead? And in the task instance, can we set which modalities to be combined, perhaps via the task category, e.g. `va2t`, or an explicit param like `input_modalities_combine=["video","audio"]` ?

AdnanElAssadi:
@Samoed @isaac-chung Changed input_column to list.

AdnanElAssadi:
lint is giving error because list is mutable

isaac-chung:
It's looking for something like this I think:
```python
from typing import ClassVar

input_column_name: ClassVar[list[str]] = ["video", "audio"]
```

mteb/abstasks/classification.py
```diff
@@ -193,7 +193,20 @@ def evaluate(
...
+                columns_to_keep = {self.label_column_name}
+                if isinstance(self.input_column_name, list):
+                    columns_to_keep.update(self.input_column_name)
+                elif self.input_column_name is not None:
+                    columns_to_keep.add(self.input_column_name)
+
+                if isinstance(ds, DatasetDict):
+                    available = set(next(iter(ds.values())).column_names)
+                else:
+                    available = set(ds.column_names)
```
Samoed, line 210: If column not existing then I think better to raise an error
```suggestion
```

mteb/tasks/video/classification/eng/kinetics400_classification.py
```diff
@@ -0,0 +1,48 @@
+from typing import ClassVar
+
+from mteb.abstasks.classification import AbsTaskClassification
+from mteb.abstasks.task_metadata import TaskMetadata
+
+
+class Kinetics400Classification(AbsTaskClassification):
+    metadata = TaskMetadata(
+        name="Kinetics400",
```
Samoed, line 9: Does this task have as audio as separate modality originally or this is just part of video? Probably we need to add some annotation whether video has audio or not

mteb/abstasks/clustering.py
```diff
@@ -161,7 +161,7 @@ class AbsTaskClustering(AbsTask):
     k_mean_batch_size: int = 512
     max_depth = None
     abstask_prompt = "Identify categories in user passages."
-    input_column_name: str = "sentences"
+    input_column_name: str | list[str] = "sentences"
```
Samoed:
```suggestion
    input_column_name: str | Sequence[str] = "sentences"
```

mteb/tasks/video/clustering/eng/ravdess_av_clustering.py
```diff
@@ -0,0 +1,47 @@
+from typing import ClassVar
+
+from mteb.abstasks.clustering import AbsTaskClustering
+from mteb.abstasks.task_metadata import TaskMetadata
+
+
+class RAVDESSAVClustering(AbsTaskClustering):
```
Samoed, line 7: Does this task have as audio as separate modality originally or this is just part of video?

mteb/tasks/video/classification/eng/kinetics400_classification.py
AdnanElAssadi: For all our current tasks, the audio column is referring to the audio coming from the video itself. 

If we later add tasks that do otherwise, we can probably add some metadata differentiating the two.

Samoed: Can we keep `video` as one column input then?

AdnanElAssadi: Some videos don't have audio, so this wouldn't differentiate them.

AdnanElAssadi: We would also no longer need to combine them like you did in `msr-vtt`. It would be clear that video always refers to frames and audio refers to the audio from the video.

Samoed: Now video have incorrect format, because it should be a dict with frames

AdnanElAssadi: No, the `_create_dataloaders.py`'s `VideoCollator` still wraps the frames in a dictionary before passing it to the model.

Samoed: I can't find where this happen. When I tried to run `MSRVTTV2T` with `mteb/baseline-random-encoder` I got an error

```
TypeError: Unsupported key type: <class 'str'>. Supported types are int and slice
```
And I don't think that collator should change input format. Videos from new tasks will be passed in wrong format

mteb/_create_dataloaders.py
```diff
@@ -607,8 +619,17 @@ def create_dataloader(
...
     if "text" in task_metadata.modalities and input_column is not None:
+        if isinstance(input_column, str):
+            text_data = dataset[input_column]
+        elif "text" in input_column and "text" in dataset.column_names:
+            text_data = dataset["text"]
+        else:
+            raise ValueError(
+                "Cannot determine which column to use for text evaluation. "
+                "Please include 'text' in input_column_name or use a single string."
+            )
```
Samoed: Why this needed?

AdnanElAssadi: When the text-evaluator needs to pull text data, this ensures it selects the "text" column from that input column list rather than crashing or trying to embed an audio column as text.

mteb/tasks/video/retrieval/eng/msr_vtt.py
```diff
@@ -42,28 +47,6 @@ def load_data(self, num_proc: int | None = None, **kwargs) -> None:
...
-            features=Features(
-                {
-                    "id": query_features["id"],
-                    "video": {
-                        "frames": query_features["video"],
-                        "audio": query_features["audio"],
-                    },
-                }
-            ),
-        )
```
Samoed, line 66: Why do you cahnge input format? We're agreed to process videos as
`{"video": {"frames": ..., "audio": ...}}`. We need to pass data in correct format after `load_data`. We shouldn't fix it in collators

AdnanElAssadi: Do you want to do this on the task side for all?

AdnanElAssadi: The main issue I had with your approach was distinguishing between normal videos and silent videos. I think I’ll stick with `input_column` being just "video" rather than a list, and specify the modalities as video and audio, noting that the video contains audio. The only downside is that this makes VA2VA vs. V2A ambiguous.

Samoed: > distinguishing between normal videos and silent videos

We can make audio in video oprional

AdnanElAssadi:
@Samoed Can you give a look here when you have the time?

mteb/abstasks/classification.py
```diff
@@ -141,7 +142,7 @@ class AbsTaskClassification(AbsTask):
     n_experiments: int = 10
     train_split: str = "train"
     label_column_name: str = "label"
-    input_column_name: str = "text"
+    input_column_name: str | Sequence[str] = "text"
```
Samoed, line 147: Can we rollback this since we don't use multiple columns? I think you can rollback almost all changes in abstasks

mteb/_create_dataloaders.py
```diff
@@ -794,3 +817,41 @@ def resample_video(
...
+        self.audio_collator = AudioCollator(
+            target_sampling_rate=target_sampling_rate,
+            max_samples=max_samples,
+        )
+
+    def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
+        if "video" in inputs[0]:
+            return self.video_collator(inputs)
+        if "audio" in inputs[0]:
+            return self.audio_collator(inputs)
```
Samoed: What if video and audio in one task?

mteb/abstasks/classification.py
isaac-chung: Don't we still have multiple inputs used in these tasks? I find the current  `_combine_modalities(example)` in every task is way too bulky

Samoed: For now, we don't have such tasks. I think it's better to combine video in tasks code itself, to have ability to separate video's audio and just audio. We discussed this in https://github.com/embeddings-benchmark/mteb/pull/4148#discussion_r2848983644. Current implementation is just hack around torchcodec/datasets

isaac-chung: Video is a collection of frames, without audio. Both tasks in this PR use audio and video. That's multiple columns. The linked comment mentions preview and doesn't seem relevant to the point to the change: to clean up this hack of combining modalities in the task and handle this in AbsTask instead.

Samoed: > handle this in AbsTask instead

I don't think that this is good idea to have modality specific code in AbsTask. Also, this is not clear when task needs combination and when don't need

> The linked comment mentions preview

Yes, this is part of discussion. Main discussion of input format was in slack channel

isaac-chung: What's the relevance of the slack discussion? Can you be more specific on the conclusion? Is video not a collection of frames without audio?

Samoed: Video is collection of frames with optional audio channel

isaac-chung: That's confusing. How can we tell when it's included or not?

Samoed: Some tasks might simply not have audio in the video. For models side `audio` key will be missing or have `None` (probably second)

isaac-chung: Still don't get why video and audio are treated as one modality

Samoed: To distinguish audio and audio from video in cases when we can have task with inputs in audio + video

isaac-chung: Which tasks have audio & audio from video and video?

Samoed: For now, we don't have such tasks

isaac-chung: What's your goal here? I don't quite understand still. The two tasks have video and audio modalities, and it doesn't seem like you agree

isaac-chung: Input column names being a list has nothing to do with supporting a list of videos or a list of images. It just means inputs have multiple input columns, which can be video and text, or video and audio etc. Here, the two tasks have video and audio, and handling it in abstask is cleaner than repeating the code in each task.

isaac-chung:
How do you tell VA2C and V2C tasks apart? Is it that: only in VA2C tasks, we process the audio, regardless if it's from the video or in a separate column?

mteb/abstasks/classification.py
Samoed: > What's your goal here? I don't quite understand still.

We need to have flexible enough format to support any combination of modalities. If we treated video as just frames and audio independently, then would be hard to distinguish tasks that have both video and audio (independent from video) inputs. We can of course name like `video` and `audio_video` and process this in abstask automatically, but I don't think that dataset should be changed during evaluation  

> Input column names being a list has nothing to do with supporting a list of videos or a list of images. It just means inputs have multiple input columns, which can be video and text, or video and audio etc

I know. But if we don't have such tasks, for now we can remove this.

> Here, the two tasks have video and audio, and handling it in abstask is cleaner than repeating the code in each task

Tasks should have dataset in our expected format after `load_data`

AdnanElAssadi:
@Samoed @isaac-chung @KennethEnevoldsen
This is somewhat of a blocker right now. Can we discuss the approach here if you are available?

mteb/abstasks/classification.py
isaac-chung: > would be hard to distinguish tasks that have both video and audio (independent from video) inputs. We can of course name like video and audio_video

We don't have such tasks right now, so we don't need it, following your reasoning.

> Tasks should have dataset in our expected format after load_data

They do. They should either be re-uploaded or handled in AbsTask when the loading code repeats itself, as it does now.

isaac-chung:
One main thing we should clarify is how to handle video with and without audio + separate audio

AdnanElAssadi:
> One main thing we should clarify is how to handle video with and without audio + separate audio

My thought is to not include cases for separate audio for now. If they later arise, we can handle it in metadata or a col in the hf dataset or something.

isaac-chung:
Okay so how do you tell VA2C and V2C tasks apart?

AdnanElAssadi:
> Okay so how do you tell VA2C and V2C tasks apart?

Yes, I raised this earlier with @Samoed. I still prefer the modality list approach, but it seemed like @KennethEnevoldsen and @Samoed wanted to bundle audio into the video object. The distinction would basically have to be captured via something like [text, video] vs [text, video, audio] in task modalities.

Would it be okay to keep video and audio as separate inputs instead of combining them under video? Or, if we want to combine, is it problematic to do it at the collator level? I think it is cleaner than what we have.

P.S. When I said separate audio above, I meant audio from another video.

Samoed:
I dug into this a bit and think we can pass `audio` and `video` separately for now, since no libraries currently support extracting audio from video directly, and there’s no standard way to do this. Though we could try to define one, which will be a lot of work, but big step for community.

Pros:
1. `torchcodec` and `datasets` don’t extract audio from video directly, so we’re relying on hacks to work around this.
2. In general, `transformers` pass `video` as tensors ([processors signature](https://github.com/huggingface/transformers/blob/cecacd374f575ad7ffe37dcd69a98cf00b551011/src/transformers/processing_utils.py#L611-L618), [video type definition](https://github.com/huggingface/transformers/blob/cecacd374f575ad7ffe37dcd69a98cf00b551011/src/transformers/video_utils.py#L60-L75)), and some omni models support extracting audio from video, but only when processing videos from a file path (like we do when uploading datasets). I don’t think [this works when using `transformers` directly](https://github.com/huggingface/transformers/blob/cecacd374f575ad7ffe37dcd69a98cf00b551011/src/transformers/models/qwen2_5_omni/processing_qwen2_5_omni.py#L230-L263) (at least based on my understanding of the code)

Cons:
1. This might limit future implementations where input is a video (with audio) alongside additional audio input.
2. Some models may support this in the future, and we’ll need to determine whether the audio is related to the video input or not.

AdnanElAssadi:
Yes, let's just do separate video and audio columns for now, it will also allow more flexibility when defining tasks.

>Some models may support this in the future, and we’ll need to determine whether the audio is related to the video input or not.

Did you come across a model that does/expects that?

@isaac-chung @KennethEnevoldsen What do you guys think?

Samoed:
> Did you come across a model that does/expects that?

As I scan in transformers repository I didn't find such models, but format might change in future

isaac-chung:
In future, if we do have audio from the video and another source for the same sample, then we can consider accepting audio (or any modality) as a list, i.e. a list of audio. But maybe we can think about that for V3 or something, later.

@AdnanElAssadi56 do we need to update the hf datasets? Or are they fine as is, and we'd make changes in the code?

AdnanElAssadi:
> @AdnanElAssadi56 do we need to update the hf datasets? Or are they fine as is, and we'd make changes in the code?

Nope, on HF, we have video and audio separate. Changes will be in code.

mteb/tasks/video/clustering/eng/ravdess_av_clustering.py
```diff
@@ -0,0 +1,77 @@
...
+from datasets import Features
+
+from mteb.abstasks.clustering import AbsTaskClustering
+from mteb.abstasks.task_metadata import TaskMetadata
+from mteb.types._encoder_io import VideoInputItem
+
+
+class RAVDESSAVClustering(AbsTaskClustering):
+    metadata = TaskMetadata(
+        name="RAVDESS_AVClustering",
```
KennethEnevoldsen:
```suggestion
        name="RAVDESSAVClustering",
```
I don't think we in general use underscore

mteb/_create_dataloaders.py
KennethEnevoldsen, line 1: Just letting you know of #4232

Which refactors this part to be notably simpler (and correct) I would probably merge it in here and integrate video into those changes

Samoed:
@KennethEnevoldsen What do you think about input format?

KennethEnevoldsen:
Hmm I do think we should be able to specify multiple `input_column_name` but maybe it should be a mapping instead:

```
input_column_name = {"text": "sentences", ...}
```

This, we of course need to be backward compatible (which is not entirely trivial here).

As for audio vs video, I think they should be seperate (video should also contain audio)

isaac-chung:
Do you mean video should not contain audio? Otherwise it reads to me as a contradictory sentence

KennethEnevoldsen:
audio should contain audio and video should contain audio as well (but the audio of the image). An example where it could be relevant:

query:
*audio request*: can you find a transcripts of this video
*video*

Samoed:
You want to keep existing format with `"video": {"frames": ..., "audio": AudioInput}`?

KennethEnevoldsen:
yes or equivalent. E.g. {"video-frames": ..., "video-audio": AudioInput}

AdnanElAssadi:
> audio should contain audio and video should contain audio as well (but the audio of the image). An example where it could be relevant:
> 
> query: _audio request_: can you find a transcripts of this video _video_

I think this also applies for images, i.e audio requesting certain types of images. For our current mveb plan, maybe we should  focus on the tasks related to videos containing audio directly. Then, when we move the the "moeb" stage, we add all the omni combinations to all modalities not just video. For now, I say we keep them separate.

AdnanElAssadi:
I refactored video/audio handling: video and audio now stay as separate HF dataset columns end-to-end instead of being merged in dataset_transform; also changed input_column_name from `str | Sequence[str]` to `str | dict[str, str]` mapping where keys are modalities and values are column names. Deleted `dataset_transform` from all 3 video tasks, updated `_prepare_dataset` to handle the full mapping for column renaming, and updated `VideoInputItem` to only contain frames (audio removed). PE-AV model now reads audio from batch["audio"] instead of from inside the video object.

AdnanElAssadi:
@KennethEnevoldsen @isaac-chung @Samoed 

Do we agree on the approach?

isaac-chung:
Sounds good to me

Samoed:
Seems that your changes broke some tests

AdnanElAssadi:
@Samoed fixed

KennethEnevoldsen:
@AdnanElAssadi56 so how does a dataset with video (incl. audio) look on HF and how is it different if it also includes audio.

mteb/_evaluators/zeroshot_classification_evaluator.py
```diff
@@ -27,7 +27,7 @@ class ZeroShotClassificationEvaluator(Evaluator):
     def __init__(
         self,
         dataset: Dataset,
-        input_column_name: str,
+        input_column_name: str | dict[str, str],
```
KennethEnevoldsen: I believe it just needs to be a Mapping?

mteb/abstasks/retrieval_dataset_loaders.py
```diff
@@ -29,8 +29,8 @@ class RetrievalSplitData(TypedDict):
     """A dictionary containing the corpus, queries, relevant documents, instructions, and top-ranked documents for a retrieval task.
 
     Attributes:
-        corpus: The corpus dataset containing documents. Should have columns `id`, `title`, `text` or `image`.
-        queries: The queries dataset containing queries. Should have columns `id`, `text`, `instruction` (for instruction retrieval/reranking) or `image`.
+        corpus: The corpus dataset containing documents. Should have columns `id`, `title`, `text` or `image` or `audio` or `video`.
+        queries: The queries dataset containing queries. Should have columns `id`, `text`, `instruction` (for instruction retrieval/reranking) or `image` or `audio` or `video`.
```
KennethEnevoldsen: This makes no sense. It sounds like it should have "id", "text" and then one of the following "audio", "instruction" and "video". 

I would rewrite to:

```suggestion
        queries: The queries dataset containing queries. Should have columns `id`, and at least one modality column (e.g. `text`, `video`, `audio`). For instruction retrieval the column can also include `instruction`.
```
the above description should change accordingly

mteb/tasks/video/classification/eng/kinetics400_classification.py
```diff
@@ -0,0 +1,51 @@
...
+        description="Kinetics-400 is a large-scale action recognition dataset containing 400 human action classes from YouTube videos. Each clip is approximately 10 seconds long.",
+        reference="https://arxiv.org/abs/1705.06950",
+        dataset={
+            "path": "mteb/kinetics-400",
+            "revision": "e5b93b6eae80b8c9e9c88a381baae84d29b34fd2",
+        },
+        type="VideoClassification",
+        category="va2c",
```
KennethEnevoldsen, line 17: isn't it just video (with audio) classification and it that then va2c and not v2c?

mteb/tasks/video/clustering/eng/ravdess_av_clustering.py
KennethEnevoldsen: @AdnanElAssadi56 can you apply this change

mteb/tasks/video/classification/eng/kinetics400_classification.py
KennethEnevoldsen: (general question for multiple datasets)

mteb/_create_dataloaders.py
```diff
@@ -32,6 +32,13 @@
 logger = logging.getLogger(__name__)
 
 
+def _normalize_input_columns(input_column_name: str | dict[str, str]) -> dict[str, str]:
```
Samoed: I don't think we need this. If a dataset has columns that do not match our format, this can be changed in `dataset_transform`

mteb/_create_dataloaders.py
```diff
@@ -467,25 +498,23 @@ def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
...
-            videos = row.pop("video")
-            video_inputs = []
-            for video in videos:
-                frames = self.resample_video(video["frames"], self.max_frames)
+            video = row.pop("video")
+            # video is a raw VideoDecoder — pass directly to resample
+            frames = self.resample_video(video, self.max_frames)
+
+            # Process audio
+            if "audio" in row and row["audio"] is not None:
```
Samoed: Maybe we can move `AudioProcessing` to `MultimodalCollator` since they're now 2 different columns in current implementation

AdnanElAssadi:
> @AdnanElAssadi56 so how does a dataset with video (incl. audio) look on HF and how is it different if it also includes audio.

It basically looks like this, with audio col being present when the video has audio.

| video          | audio                          | label |
|----------------|--------------------------------|-------|
| VideoDecoder_0 | {"array": [...], "sr": 16000}  | 0     |
| VideoDecoder_1 | {"array": [...], "sr": 16000}  | 3     |

mteb/tasks/video/classification/eng/kinetics400_classification.py
AdnanElAssadi: We want to differentiate datasets that have video+audio versus just video(frames).

It's va2c because the HF dataset has both video and audio columns, and the PE-AV model for example produces joint audio-video embeddings from both.

mteb/_create_dataloaders.py
AdnanElAssadi: We use it in classification.py, clustering.py, and task_pipelines.py to avoid repeating the same isinstance check, but ill change to inline.

mteb/tasks/video/classification/eng/kinetics400_classification.py
KennethEnevoldsen: but how do we then differentiate video+audio and video+audio, audio?

AdnanElAssadi: right now the category describes what modalities the dataset actually has, va2c means both video and audio columns exist. If a dataset had video (no audio) it would be v2c. We don't have a case yet where there's video+audio AND a separate standalone audio as we discussed

isaac-chung: > but how do we then differentiate video+audio and video+audio, audio?

@KennenthEnevoldsen we discussed above: https://github.com/embeddings-benchmark/mteb/pull/4199#issuecomment-4088942681 we will accept audio as a list later. No other models or libraries discern this.

isaac-chung: I'm really not understanding why we seem to not get the video = list of frames thing, and that audio is an entire separate modality. Sentence Transformers keeps them separate too.

I hope we aren't stuck overthinking this, as it feels like that sometimes.

AdnanElAssadi56: @KennethEnevoldsen just pinging in case you missed this

AdnanElAssadi56: @Samoed Any points from your side?

Samoed: In the case of video format, I see both sides. If we keep video and audio as separate columns, then this will be easy to handle in the model and across the library, but this would limit us when we have `video(+audio)+audio` as input, though I don't think we would have such tasks. So, I'm not sure what to do here

KennethEnevoldsen: > No other models or libraries discern this.

Right, but we could easily imagine the query that contains both video(+audio) and audio. E.g.
```
[can you search for the product that appears in this video:](audio) 
[*video of product*](video)
```

Currently we do not support lists either:
```
find similar images/videos/sounds to the following:
[videos]/[audios snippets]/[images]
```
I am not saying we have to solve everything here and we could go for a simplified solution, but I don't see the argument for not going for the following columns

```
video_frames, video_audio, audio, images, text
```
This would allow for video(+audio) as well as standalone audio. How to best annotate that. I would probably do `"v(a)"` for video with audio and then `"va"` for video and audio. We could also just use`va` for both cases.

isaac-chung: Video audio is also just audio. I think down the line this case can be solved by supporting a list of inputs per modality.

AdnanElAssadi56:
Results from `pe-av-small`
[RAVDESSAVClustering.json](https://github.com/user-attachments/files/26636411/RAVDESSAVClustering.json)

Do i merge @KennethEnevoldsen @Samoed @isaac-chung ?

Samoed:
No, please address comments from our reviews

mteb/_create_dataloaders.py
Samoed: I don't see why we need this dict mapping

AdnanElAssadi56:
> No, please address comments from our reviews

Is there anything pending?

Samoed:
Comments that unresolved

mteb/_create_dataloaders.py
AdnanElAssadi56: @KennethEnevoldsen has suggested that

AdnanElAssadi56: We can either have sequence and dataset transform the columns that don't match the names we want or this.

KennethEnevoldsen: @Samoed we can have multiple column names right. Would you just use a list of strings instead? Then how to you know which one is which?

Samoed: All required transformation can happen in `dataset_transform`. I think handling mapping requires a lot of additional work. Currently, we except that in column "video" will be video, in "text" text, etc.

mteb/abstasks/task_metadata.py
```diff
@@ -198,6 +198,15 @@
     "Any2AnyRetrieval",
 )
 
+MVEB_TASK_TYPE = (
```
Samoed, line 201: Do we want to create new task types?

mteb/types/_encoder_io.py
```diff
@@ -140,11 +140,9 @@ class VideoInputItem(TypedDict):
 
     Attributes:
         frames: The video frames as Tensor.
-        audio: The audio array as AudioInputItem.
     """
 
     frames: torch.Tensor
```
Samoed: Do we need subdict in that case? We can pass `frames` in video column directly

mteb/_create_dataloaders.py
```diff
@@ -441,60 +466,22 @@ def resample_audio(
...
-            "BatchedInput",
-            {
-                key: [row[key] for row in collated_inputs]
-                for key in collated_inputs[0].keys()
-            },
-        )
+            video = row.pop("video")
+            frames = self.resample_video(video, self.max_frames)
+            row["video"] = VideoInputItem(frames=frames)
+        return inputs
```
Samoed: I think this should output 
```
{
                key: [row[key] for row in collated_inputs]
                for key in collated_inputs[0].keys()
            },
```
as audio collator

mteb/abstasks/task_metadata.py
AdnanElAssadi56: Thats how MIEB & MAEB do it

mteb/types/_encoder_io.py
AdnanElAssadi56: For now, i guess we dont. If torchcodec adds audio later, we can refactor.

mteb/_create_dataloaders.py
AdnanElAssadi56: ill change back to sequence

mteb/types/_encoder_io.py
Samoed: We should use one. 2 different columns or inside values

AdnanElAssadi56: I removed VideoInputItem, frames go directly in the video column and audio stays as a separate column now.

AdnanElAssadi56:
@Samoed Any more points?

mteb/_evaluators/clustering_evaluator.py
```diff
@@ -51,7 +53,9 @@ def __call__(
         data_loader = create_dataloader(
             self.dataset,
             task_metadata=self.task_metadata,
-            input_column=self.input_column_name,
+            input_column=self.input_column_name
+            if isinstance(self.input_column_name, str)
+            else None,
```
Samoed: Not sure that I understand why if `self.input_column_name` is Sequence then we just ignore them. Do we need `input_column` `Sequence` then?

mteb/abstasks/classification.py
```diff
@@ -470,31 +492,30 @@ def _undersample_data(
     def _calculate_descriptive_statistics_from_split(
         self, split: str, hf_subset: str | None = None, compute_overall: bool = False
     ) -> ClassificationDescriptiveStatistics:
+        col = (
+            self.input_column_name
+            if isinstance(self.input_column_name, str)
+            else self.input_column_name[0]
+        )
```
Samoed: We shouldn't ignore columns other than `input_column_name[0]`

mteb/abstasks/clustering.py
```diff
@@ -209,16 +210,32 @@ def _evaluate_subset(
...
         )
+        columns_to_keep = set(input_cols) | {self.label_column_name}
+
+        available = set(data_split.column_names)
+        missing_columns = columns_to_keep - available
+        if missing_columns:
+            raise ValueError(
+                f"Missing required columns in dataset: {missing_columns}. "
+                f"Available columns: {available}"
+            )
```
Samoed, line 230: Why we have this check only in clustering tasks?

AdnanElAssadi56: It's in classification as well

mteb/abstasks/classification.py
AdnanElAssadi56: I am not entirely sure how to handle multimodal statistics, can we do this in a follow up?

AdnanElAssadi56: Ill just leave it as a placeholder for now

mteb/_evaluators/clustering_evaluator.py
AdnanElAssadi56: input_column_name as Sequence is used for column selection in the abstask (deciding which columns to keep). The input_column param in create_dataloader is only for renaming a single column, which isn't needed when column names already match the modality names.

mteb/_create_dataloaders.py
KennethEnevoldsen: but if that is the case why even bother with making it a sequence?

AdnanElAssadi56: https://github.com/embeddings-benchmark/mteb/pull/4199#discussion_r3069819008

mteb/_evaluators/zeroshot_classification_evaluator.py
```diff
@@ -27,7 +29,7 @@ class ZeroShotClassificationEvaluator(Evaluator):
     def __init__(
         self,
         dataset: Dataset,
-        input_column_name: str,
+        input_column_name: str | Mapping[str, str],
```
KennethEnevoldsen: It seems like this is one of the main problems at the moment. The problem is that it is currently not consistently used.

for AbsTaskZeroShotClassification default is "image". E.g. for `PatchCamelyonZeroShot` it is "webp" meaning for the image column use "webp". However, for clustering it is `input_column_name="sentences"` and for classification it is text.

Currently, (as I see it) it is problematic in the sense that if the task contains any text, it is assumed that the column name is intended for text:

https://github.com/embeddings-benchmark/mteb/blob/e0c2f2db8589bad498561c97a9d2357a97d60bb2/mteb/_create_dataloaders.py#L319-L320

Which probably means that `PatchCamelyonZeroShot` is broken (@Samoed can I get your opinion here?). 

If this is the case, we could simply rename from input_column to `text` and then just use a predefined set of columns. Which could potentially be expanded into a rename mapping.

Samoed: > Which probably means that PatchCamelyonZeroShot is broken 

I checked and `PatchCamelyonZeroShot` work correctly, because we will this part of `_create_dataloader` when `input_column_name` is `None`, but `PatchCamelyonZeroShot` has `webp` here. 

> If this is the case, we could simply rename from input_column to text and then just use a predefined set of columns

I don't think that good idea, because we would have some datasets that would be broken due to this. I think most easier to keep sequence and rename columns in `dataset_transform` if needed.

KennethEnevoldsen: > I checked and PatchCamelyonZeroShot work correctly, because we will this part of _create_dataloader when input_column_name is None, but PatchCamelyonZeroShot has webp here.

I am not sure I understand the argument here. The linked segment is called when `input_column` is **not** `None`. 


> I don't think that good idea, because we would have some datasets that would be broken due to this. I think most easier to keep sequence and rename columns in dataset_transform if needed.

I think my argument is that it is not very clear right now what input_column does, which causes issues like: https://github.com/embeddings-benchmark/mteb/issues/4182

e.g. if I have input columns `["sentences", "webp"]` how do we map these currently (I don't think we can).

If the intention is to only support default names here, then we need to specify it somewhere potentially check it as well.

mteb/tasks/video/classification/eng/kinetics400_classification.py
```diff
@@ -0,0 +1,53 @@
...
+      year={2017},
+      eprint={1705.06950},
+      archivePrefix={arXiv},
+      primaryClass={cs.CV},
+      url={https://arxiv.org/abs/1705.06950}, 
+}
+""",
+    )
+
+    input_column_name: ClassVar[str | Sequence[str]] = ["video", "audio"]
```
KennethEnevoldsen:
```suggestion
    input_column_name = ["video", "audio"]
```

mteb/tasks/video/clustering/eng/ravdess_av_clustering.py
```diff
@@ -0,0 +1,52 @@
...
+  publisher = {Public Library of Science},
+  title = {The Ryerson Audio-Visual Database of Emotional Speech and Song (RAVDESS): A dynamic, multimodal set of facial and vocal expressions in North American English},
+  url = {https://doi.org/10.1371/journal.pone.0196391},
+  volume = {13},
+  year = {2018},
+}
+""",
+    )
+    max_fraction_of_documents_to_embed = None
+    input_column_name: ClassVar[str | Sequence[str]] = ["video", "audio"]
```
KennethEnevoldsen:
```suggestion
    input_column_name = ["video", "audio"]
```

mteb/tasks/video/retrieval/eng/msr_vtt.py
```diff
@@ -31,6 +37,8 @@ class MSRVTTV2T(AbsTaskRetrieval):
         bibtex_citation=None,
     )
 
+    input_column_name: ClassVar[str | Sequence[str]] = ["video", "audio"]
```
KennethEnevoldsen:
```suggestion
    input_column_name = ["video", "audio"]
```

mteb/_create_dataloaders.py
```diff
@@ -519,3 +495,49 @@ def resample_video(
             else list(range(video_frames))
         )
         return video.get_frames_at(selected_frames).data
+
+
+class MultimodalCollator:
```
KennethEnevoldsen: I would rename to VideoCollator and the other one to FramesCollator

mteb/types/_encoder_io.py
```diff
@@ -133,28 +133,14 @@ class AudioInput(TypedDict):
...
-        frames: The video frames as Tensor.
-        audio: The audio array as AudioInputItem.
-    """
-
-    frames: torch.Tensor
-    audio: AudioInputItem
-
-
 class VideoInput(TypedDict):
     """The input to the encoder for videos.
```
KennethEnevoldsen:
```suggestion
    """The input to the encoder for video frames. Audio is currently includes in the AudioInput.
```

mteb/tasks/video/retrieval/eng/msr_vtt.py
```diff
@@ -31,6 +37,8 @@ class MSRVTTV2T(AbsTaskRetrieval):
...
+
     def load_data(self, num_proc: int | None = None, **kwargs) -> None:
```
KennethEnevoldsen, line 37: should reupload in the correct format nwo that we have settled on it and the remove this load_data function

If the functionality is not there add a `NotImplementedError` and an issue on it and link it to the MVEB overview issue

mteb/models/model_implementations/pe_av_models.py
```diff
@@ -0,0 +1,364 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import torch
+from tqdm.auto import tqdm
+
+from mteb._create_dataloaders import MultimodalCollator
```
KennethEnevoldsen: Do we want a model in rely on a private collator. Seems like we also do this for AudioCollator. This makes us think we can change this without problems, but it will cause a problem. We should fix this. Though this is not for this PR.

mteb/abstasks/classification.py
```diff
@@ -140,7 +141,7 @@ class AbsTaskClassification(AbsTask):
     n_experiments: int = 10
     train_split: str = "train"
     label_column_name: str = "label"
-    input_column_name: str = "text"
+    input_column_name: str | Sequence[str] = "text"
```
KennethEnevoldsen, line 147: Please update description to match for this argument. If specified as a sequence we it has to be the default (add [link to where one can see the defaults](https://embeddings-benchmark.github.io/mteb/api/types/#mteb.types._encoder_io))

mteb/abstasks/_data_filter/task_pipelines.py
```diff
@@ -85,6 +85,14 @@ def process_classification(
     """Process classification task dataset(s) with cleaning pipeline."""
     if not task.data_loaded:
         task.load_data()
+    # Multi-column tasks (e.g. video+audio) don't support text cleaning
+    if not isinstance(task.input_column_name, str):
+        if isinstance(task.dataset, DatasetDict):
+            return task.dataset
+        if task.dataset is None:
+            raise ValueError("Task dataset is None.")
+        return dict(task.dataset)
```
KennethEnevoldsen: I am not sure what is happening here?

mteb/_evaluators/zeroshot_classification_evaluator.py
```diff
@@ -53,7 +55,9 @@ def __call__(
     ) -> Array:
         dataloader = create_dataloader(
             self.dataset,
-            input_column=self.input_column_name,
+            input_column=self.input_column_name
+            if isinstance(self.input_column_name, str)
+            else None,
```
KennethEnevoldsen: move to top of function and only do the transformation once.

mteb/_evaluators/zeroshot_classification_evaluator.py
Samoed: > I am not sure I understand the argument here. The linked segment is called when input_column is not None.

Ah, yes. But this code will be triggered only when task modality is `== ["text"]`, but `PatchCamelyonZeroShot` has image and text.

> e.g. if I have input columns ["sentences", "webp"] how do we map these currently (I don't think we can).

I think we can keep transformation based on `input_column` only for `str` values, if sequence then we would use only default names

KennethEnevoldsen: > Ah, yes. But this code will be triggered only when task modality is == ["text"], but PatchCamelyonZeroShot has image and text.

Yeah so it works, but I am not entirely sure it will work in all combinations. I will make an issue on creating tests to ensure that input_column works for all modalities.

mteb/tasks/video/retrieval/eng/msr_vtt.py
AdnanElAssadi56: The formatting in load data here is retrieval specific rather than related to video. We have similar handling for datasets in Audio and elsewhere. But reuploading is an option for all.

mteb/abstasks/_data_filter/task_pipelines.py
AdnanElAssadi56: will add longer comment

mteb/tasks/video/classification/eng/kinetics400_classification.py
AdnanElAssadi56: We faced issues with linter before when not including ClassVar

mteb/abstasks/_data_filter/task_pipelines.py
Samoed: I think simpler would be to raise `NotImplemented` here if we don't support for now

AdnanElAssadi56: Right, i just did in latest push

mteb/tasks/video/retrieval/eng/msr_vtt.py
AdnanElAssadi56: I added a TODO comment

KennethEnevoldsen: feel free to also do an issue on it and then link here and resolve

mteb/models/model_implementations/pe_av_models.py
KennethEnevoldsen: feel free to do an issue on it and then link here and resolve

mteb/tasks/video/classification/eng/kinetics400_classification.py
KennethEnevoldsen: Ahh the problem is that is it mutable, just specify it as a tuple and you will not have this issue

KennethEnevoldsen: ruff issue: https://docs.astral.sh/ruff/rules/mutable-class-default/

mteb/abstasks/_data_filter/task_pipelines.py
KennethEnevoldsen: I would also add issue on it as well

mteb/tasks/video/retrieval/eng/msr_vtt.py
AdnanElAssadi56: https://github.com/embeddings-benchmark/mteb/issues/4375

mteb/abstasks/_data_filter/task_pipelines.py
AdnanElAssadi56: https://github.com/embeddings-benchmark/mteb/issues/4377

mteb/models/model_implementations/pe_av_models.py
AdnanElAssadi56: https://github.com/embeddings-benchmark/mteb/issues/4376

mteb/abstasks/classification.py
AdnanElAssadi56: https://github.com/embeddings-benchmark/mteb/issues/4378

AdnanElAssadi56:
@KennethEnevoldsen @Samoed

Added relevant issues from discussions above and resolved convos (commented the issue links here as well).

mteb/_evaluators/sklearn_evaluator.py
```diff
@@ -64,17 +66,22 @@ def create_dataloaders(
         encode_kwargs: EncodeKwargs,
         num_proc: int | None,
     ) -> tuple[DataLoader[BatchedInput], DataLoader[BatchedInput]]:
+        input_col = (
+            self.values_column_name
+            if isinstance(self.values_column_name, str)
+            else None
+        )
```
Samoed: Maybe move this inside `create_dataloader`, because we just reusing this everywhere

AdnanElAssadi56: Done

mteb/tasks/video/classification/eng/kinetics400_classification.py
```diff
@@ -0,0 +1,48 @@
...
+      year={2017},
+      eprint={1705.06950},
+      archivePrefix={arXiv},
+      primaryClass={cs.CV},
+      url={https://arxiv.org/abs/1705.06950}, 
+}
+""",
+    )
+
+    input_column_name = ("video", "audio")
```
isaac-chung, line 45: Maybe including audio is not valid for this task. See [paper section 3.3, p.5](https://arxiv.org/pdf/1705.06950):
> Note, the AMT workers didn’t have access to the audio
to ensure that the video can be classified purely based on its
visual content.

isaac-chung: I'll create a video only instance in my PR.

PR #4202, opened 2026-03-05
Added model zeroentropy/zembed-1

rjwang6:
- [y] I have filled out the ModelMeta object to the extent possible
- [y] I have ensured that my model can be loaded using
  - [y] `mteb.get_model(model_name, revision)` and
  - [y] `mteb.get_model_meta(model_name, revision)`
- [y] I have tested the implementation works on a representative set of tasks.
- [y] The model is public, i.e., is available either as an API or the weights are publicly available to download


mteb/models/model_implementations/zeroentropy_models.py
Samoed, line 1: Can you fill params with `None`?

rjwang6: Just to clarify, do you mean there are more parameters I should set to None?

Samoed: No, you should fill with values `None` parameters

mteb/models/model_implementations/zeroentropy_models.py
```diff
@@ -0,0 +1,32 @@
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+from mteb.models.sentence_transformer_wrapper import sentence_transformers_loader
+
+zeroentropy_zembed_1 = ModelMeta(
+    loader=sentence_transformers_loader,
+    loader_kwargs={},
```
Samoed:
```suggestion
    loader_kwargs={"trust_remote_code": True},
```

mteb/models/model_implementations/zeroentropy_models.py
rjwang6: I filled most of the parameters, I wasn't able to find a complete list of languages so I just put eng-Latn

Issue #3707, opened 2025-12-10
Update documentation with insiders


KennethEnevoldsen:
### Description of the feature

mkdocs material and mkdocstring insiders is now free for everyone:

https://squidfunk.github.io/mkdocs-material/blog/2025/11/11/insiders-now-free-for-everyone/

We should be able to improve the docs with these being released.

This comes out due to the transition to [zensical](https://zensical.org/), a Rust-written backend to replace mkdocs. It looks great, and we could likely also move over, but it is currently in alpha. So want to make sure that everything works beforehand.

Samoed:
Sadly, mkdocs is no longer maintained. I went down a bit of a rabbit hole, and it seems that `requests` and `httpx` aren't being properly maintained either.

What features do you want to integrate? As I've seen we these can be helpful:
- [instant-prefetching](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#instant-prefetching)
- [custom-selectors](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#custom-selectors) (probably)
- [footnote-tooltips](https://squidfunk.github.io/mkdocs-material/reference/footnotes/#footnote-tooltips)

KennethEnevoldsen:
breadcrumbs might be decent as well

Samoed:
Seems that mkdocs have another drama https://github.com/mkdocs/mkdocs/discussions/4077. We probably need try to migrate to zeniscal

KennethEnevoldsen:
Yep - I have been using it for my personal website to test it - mostly works fine I think we could switch without too many issues.

Samoed:
Looks great! It even support docstrings, so should be doable https://zensical.org/docs/setup/extensions/mkdocstrings/#configuration-zensicaltoml

KennethEnevoldsen:
Yep I believe the mkdocs-docstring maintainer also moved to zensical

Samoed:
Issue no longer relevant. Migrated to zensical

PR #4203, opened 2026-03-05
docs: migrate to zensical

Samoed:
Overall everything is working, except for tasks/models overview. We need to list all files inside directory to correctly handled it. Maybe this is bug/missing feature

Ref https://github.com/embeddings-benchmark/mteb/issues/3707


Samoed:
> can't find that in the docs

Yes, they have issue for bibtex supporting https://github.com/zensical/backlog/issues/79

KennethEnevoldsen:
Seems like it is a low priority (from the linked kanban) - should we convert to links or wait?

Samoed:
I think it's better to use links (we can link to benchmark instance in overview)

KennethEnevoldsen:
I was thinking more link to the paper itself

Samoed:
Zensical have ["annotation" feature](https://zensical.org/docs/authoring/code-blocks/#add-annotations) and our comments with `(1)` rendered a bit incorrectly (`!` also works with mkdocs)

Samoed:
I think problem with tasks/models overview is still not resolved

KennethEnevoldsen:
Actually seems like available tasks give a 404 (thought it was only the logos)

Samoed:
Yes. If we list explicitly files in folder, then they will be correctly rendered. I'll try to test on demo project to reproduce an error

Samoed:
Created https://github.com/zensical/zensical/issues/416

Samoed:
If we really want to move to zenscal we can remove `nav` part and use titles in frontmatter to fix this. 
image
But then not clear how to add a link to the leaderboard and some md files from mmteb and mieb folders.

Samoed:
From https://github.com/zensical/zensical/issues/416#issuecomment-4011959930

> This is behavior inherited from MkDocs, which does not support mixing explicit and implicit navigation. We would consider this a feature request rather than a bug report for this reason.
Were you using the [awesome-nav](https://lukasgeiter.github.io/mkdocs-awesome-nav/) plugin by any chance when building with MkDocs?
We are currently re-thinking navigation to include degrees of freedom like this. See [ZAP 005](https://zensical.org/spark/proposals/zap-005-navigation-authoring-experience/) for an overview where this is going. We aim to support everything that awesome-nav can do and a whole lot more.

KennethEnevoldsen:
Hmm yeah. My suggestion would probably be to simplify the structure. So we don't have multiple types of classification that essentially all uses the same class.

I would here either:

1) use a modality similar to models

```
available tasks
|- Text
|     |- Classification
|     |     |- task 1
|     |     |- ...
|     |- Retrieval
|     |     |- task 1
|     |     |- ...
```

2) or use broader type categories 

```
available tasks
|- Classification
|     |- Text
|     |     |- task 1
|     |     |- ...
|     |- Audio
|     |     |- task 1
|     |     |- ...
```

Samoed:
What we should do with `mmeb` and ` mieb` folders? Can we remove them or move to different folder?

KennethEnevoldsen:
I would probably remove

Samoed:
I removed `nav` config from mkdocs and now header look like 
image
I don't think that we can change order without `nav` usage

KennethEnevoldsen:
Why would be remove it? (I suspect it is to add the tasks)

(I would instead refactor as suggested [here](https://github.com/embeddings-benchmark/mteb/pull/4203#issuecomment-4012491037))

KennethEnevoldsen:
I btw. added a copy paste button for model names (I found that I myself was missing that)

Screenshot 2026-03-07 at 13 00 56

Looks pretty good I would say

Samoed:
> Why would be remove it? (I suspect it is to add the tasks)
> 
> (I would instead refactor as suggested [here](https://github.com/embeddings-benchmark/mteb/pull/4203#issuecomment-4012491037))

We can't use `nav` and path to folder. If we use nav, then we need to list all files in nav then

KennethEnevoldsen:
Added the copy button for tasks and benchmarks as well and updated the tables with symbols so that it is more skimable:

Screenshot 2026-03-07 at 16 30 22

Samoed:
Looks great!

KennethEnevoldsen:
> We can't use nav and path to folder. If we use nav, then we need to list all files in nav then

But I think we kinda have to do that to have the control e.g. advanced usage is currently above usage.

For models it is not a big hassle (already normalized so that we never have more than 2 modalities), but for tasks we would need to make a seperate distinction. I would probably suggest:

```py
Bitext Mining
Classification
| - Classification # inserted in the .md
|      | - image
|      | - text
| - Multilabel Classification
| - Pair Classification
Clustering
Retrieval
| - Retrieval
| - Reranking
Semantic Similarity
Others
| - Summarization
| - Regression
```

Though it would probably be easier to do it by modality

Samoed:
Currently we have rask type per modality and we can adopt this structure only to `Any2Any` tasks which have image and audio. Other tasks have task type per modality (ImageClassification, AudioClassification etc). We can keep current approach and just explicitly add all paths to nav and update when we add some new type manually, because I don't think that change nav from script is good idea

KennethEnevoldsen:
> Other tasks have task type per modality (ImageClassification, AudioClassification etc)

but I think this task type categorization is poor. ImageClassification is just the modality and task specification in one. I believe we initially made this decision to make sure that the plots worked as intended on the leaderboard for MIEB.

KennethEnevoldsen:
I don't think I want to change them now (would break backward compat.), but would love to add a simpler type. We can add that in as a new value

Samoed:
> but I think this task type categorization is poor. ImageClassification is just the modality and task specification in one. I believe we initially made this decision to make sure that the plots worked as intended on the leaderboard for MIEB.

I agree. I also don't like them

KennethEnevoldsen:
Updated to look like this:

Screenshot 2026-03-10 at 11 15 00

Samoed:
Agree. I also think new doc is better

Samoed:
@KennethEnevoldsen 
```
deploy
Branch "main" is not allowed to[ deploy ](https://github.com/embeddings-benchmark/mteb/actions/runs/22898120498/job/66437880862)to github-pages due to environment protection rules.

The deployment was rejected or didn't satisfy other protection rules.
```
https://github.com/embeddings-benchmark/mteb/actions/runs/22898120498/attempts/1

PR #4209, opened 2026-03-07
refactor: move generate_model_card to ModelMeta method

### LINKED ISSUE #4129: Move `generate_model_card` from `cli` to `ModelMeta`

We have function `generate_model_card` that accessible only from cli. I think we can add this function as method to `ModelMeta` for more direct access

stbiadmin:
Closes #4129

Moves `generate_model_card` from the CLI module into `ModelMeta` as an instance method, so it can be called directly on a model metadata object.

Also fixes a mutable default argument bug where `ResultCache()` was shared across calls.


mteb/models/model_meta.py
```diff
@@ -1120,6 +1121,110 @@ def to_python(self) -> str:
         """Returns a string representation of the model."""
         return _pydantic_instance_to_code(self, exclude_fields=["experiment_kwargs"])
 
+    def generate_model_card(
```
Samoed: Can you also add here `ModelMeta.push_eval_results` function?

stbiadmin:
Added a push_eval_results wrapper too.

mteb/cli/generate_model_card.py
```diff
@@ -41,99 +35,48 @@ def generate_model_card(
...
-        existing_model_card_data.model_name = (
-            existing_model_card_data.model_name or model_name
-        )
-    else:
-        unique_eval_results = {
-            eval_result.unique_identifier: eval_result
-            for eval_result in existing_model_card_data.eval_results + eval_results
-        }
-        eval_results = list(unique_eval_results.values())
+def push_eval_results(
```
Samoed: I meant you should add `push_eval_results` to `generate_model_card` not moving it to cli

stbiadmin:
Sorry, not sure I follow. push_eval_results is already on ModelMeta (line 1228). Did you want a CLI wrapper for it in generate_model_card.py (which is what I added), or something else?

Samoed:
Can you add call of `ModelMeta.push_eval_results` to `ModelMeta.generate_model_card`?

stbiadmin:
Got it. Added push_eval_results as an optional call inside generate_model_card and dropped the standalone CLI wrapper.

KennethEnevoldsen:
@Samoed seems like this one is waiting on you - let me know if you want the second pair of eyes on it

Samoed:
I made some adjustments to arguments to reuse existing

mteb/models/model_meta.py
```diff
@@ -1163,6 +1164,105 @@ def to_python(self) -> str:
...
+        self,
+        tasks: Sequence[AbsTask] | None = None,
+        benchmarks: Sequence[Benchmark] | None = None,
+        existing_model_card_id_or_path: str | Path | None = None,
+        results_cache: ResultCache | None = None,
+        output_path: Path = Path("model_card.md"),
+        add_table_to_model_card: bool = False,
+        models_to_compare: Sequence[str] | None = None,
+        *,
```
KennethEnevoldsen: I make every argument only 2 args positional, I think here I might even argue for none

mteb/models/model_meta.py
```diff
@@ -1163,6 +1164,105 @@ def to_python(self) -> str:
...
+        token: str | None = None,
+        push_to_hub: bool = False,
+        push_eval_results: bool = False,
```
KennethEnevoldsen: Why does push_eval_results?

mteb/models/model_meta.py
```diff
@@ -1163,6 +1164,105 @@ def to_python(self) -> str:
         """Returns a string representation of the model."""
         return _pydantic_instance_to_code(self, exclude_fields=["experiment_kwargs"])
 
+    def generate_model_card(
```
KennethEnevoldsen: @Samoed is this the interface we want?

Seems like we are just moving the CLI, which leads us to having a lot of arguments. We could decompose it more by having

```
model_card = meta.generate_model_card()
meta.push_card_to_hub(model_card)
```

That way it is also possible to inspect steps during the process and it is also easier to write tests for the sub functionality

mteb/models/model_meta.py
```diff
@@ -1195,6 +1295,29 @@ def push_eval_results(
         )
 
 
+def _add_table_to_model_card(
```
KennethEnevoldsen, line 1443: I would love to see an example of how this would look like (does it just insert it in the buttom of the model card. 

What if I want it inserted in a specific place (e.g. ran some models and wanted to update the table). This would now require a manual edit.

mteb/models/model_meta.py
Samoed: To push eval results with new HF format with model card together. We can keep them separate

mteb/models/model_meta.py
Samoed: I thought making this compatible with cli, but agree this is not necessary

mteb/models/model_meta.py
Samoed: Probably it's better to rename `push_mode_card_to_hub`

mteb/models/model_meta.py
Samoed: Removed them

mteb/models/model_meta.py
Samoed: Created example https://huggingface.co/mteb/baseline-bm25s/discussions/7

```python
import mteb

meta = mteb.get_model_meta("mteb/baseline-bm25s")
benchmark = mteb.get_benchmark("RTEB(beta)")
meta.push_model_card_to_hub(
    tasks=benchmark,
    benchmarks=[benchmark],
    existing_model_card_id_or_path="mteb/baseline-bm25s",
    add_table_to_model_card=True,
    models_to_compare=["intfloat/multilingual-e5-small"],
    push_to_hub=True,
    create_pr=True,
)
```

KennethEnevoldsen: I would remove release date, but otherwise I think it is fine

KennethEnevoldsen (approved):

KennethEnevoldsen:
@Samoed merging this - assume that you are happy with it

PR #4221, opened 2026-03-10
fix: fix docs deploy

Samoed:
```
deploy
Branch "main" is not allowed to[ deploy ](https://github.com/embeddings-benchmark/mteb/actions/runs/22898120498/job/66437880862)to github-pages due to environment protection rules.

The deployment was rejected or didn't satisfy other protection rules.
```
https://github.com/embeddings-benchmark/mteb/actions/runs/22898120498/attempts/1

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -6,15 +6,13 @@ on:
   pull_request:
 
 permissions:
-  contents: write
+  contents: read
   pages: write
   id-token: write
 
 jobs:
   build:
     runs-on: ubuntu-latest
-    environment:
-      name: github-pages
     steps:
       - name: Free disk space
         run: |
@@ -48,7 +46,7 @@ jobs:
           python-version: "3.10"
 
       - name: Configure GitHub Pages
-        if: github.event_name == 'push'
+        # if: github.event_name == 'push'
         uses: actions/configure-pages@v5
 
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


Samoed:
Probably there will be solution https://github.com/orgs/community/discussions/39054#discussioncomment-4134107 . Tried to change to deploy from `action`, but this still didn't finish successfully

KennethEnevoldsen:
Currently, the rules are as follows. That seems like the main branch should be able to deploy (but this PR should not)

Screenshot 2026-03-11 at 08 52 12

.github/workflows/documentation.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           python-version: "3.10"
 
       - name: Configure GitHub Pages
-        if: github.event_name == 'push'
+        # if: github.event_name == 'push'
```
KennethEnevoldsen, line 49: We of course don't want this generally

KennethEnevoldsen: I have temporarily allowed all branches:

Image

Then we can make it work and then convert to only main again

Samoed: > We of course don't want this generally

Yes, this is for debugging

Samoed:
Still same error
image

KennethEnevoldsen:
Hmm, should we try testing on another repo to figure out the CI? An alternative approach is of course pushing the changes to the github-pages branch as before

Samoed:
Previously this was done by `mkdocs gh-deploy`, but now we don't have this option. I'll try to debug in my fork

Samoed:
Deploy was successful in https://github.com/embeddings-benchmark/mteb/commit/5530493a02c3d6c78cbaf930ee2d1526b34598a5 but in https://github.com/embeddings-benchmark/mteb/commit/c87573e61d8f519d2e0cd96792c0f68763d471e4 failed

Samoed:
Retried https://github.com/embeddings-benchmark/mteb/commit/c87573e61d8f519d2e0cd96792c0f68763d471e4 and seems everything is working. Now we deploy from github action, not from branch

PR #4231, opened 2026-03-12
fix: metadata getting computed for existing MTEB model

### LINKED ISSUE #4181: Metadata is computed even when loading models with existing metadata

for instance:

```
model = mteb.get_model_meta(
    "sentence-transformers/static-similarity-mrl-multilingual-v1", fetch_from_hf=False
)
model.load_model()
# Could not determine vocab size for model None and therefore cannot calculate number of embedding parameters. 
# Error: 
# 'tokenizers.Tokenizer' object has no attribute 'vocab'
# Model does not have a recognized architecture for calculating embedding parameters (model=None).

```

Which call:
https://github.com/embeddings-benchmark/mteb/blob/5b89317949d2686c526d8c7170efeef973695c00/mteb/models/sentence_transformer_wrapper.py#L80-L81

Which ill call `_from_sentence_transformer_model`, which calls `_get_n_embedding_parameters_from_sentence_transformers`. This can give errors like `Model does not have a recognized architecture for calculating embedding parameters`, even though the metadata is already specified.

Not sure what the best approach is to fix this, but ideally we should not try to compute the metadata if it is already there. One way would be to pass the `ModelMeta` to the initalization. This should be backward compatible following the `EncoderProtocol`, but might still cause issues for models that don't allow `**kwargs`.

ayush1298:
closes #4181 

Now, below code run without any error/warning: 

```
model = mteb.get_model_meta(
    "sentence-transformers/static-similarity-mrl-multilingual-v1", fetch_from_hf=False
)
model.load_model()
```

### DIFF
--- a/mteb/models/sentence_transformer_wrapper.py
+++ b/mteb/models/sentence_transformer_wrapper.py
@@ -297,10 +297,16 @@ class CrossEncoderWrapper:
 
         if isinstance(model, CrossEncoder):
             self.model = model
+            self.mteb_model_meta = ModelMeta.from_cross_encoder(self.model)
         elif isinstance(model, str):
             self.model = CrossEncoder(model, revision=revision, device=device, **kwargs)
-
-        self.mteb_model_meta = ModelMeta.from_cross_encoder(self.model)
+            self.mteb_model_meta = ModelMeta.create_empty(
+                overwrites=dict(
+                    name=model,
+                    revision=revision,
+                    loader=CrossEncoderWrapper,
+                )
+            )
         self.query_prefix = query_prefix
         self.passage_prefix = passage_prefix
 


Samoed:
> Now, below code run without any error/warning:

But this code not triggers your changes. You should use `mteb.get_model` instead

ayush1298:
> > Now, below code run without any error/warning:
> 
> But this code not triggers your changes. You should use `mteb.get_model` instead

The error was coming from `load_model`. So, that thing is resolved.
Also, using : 

```
import mteb
model = mteb.get_model(
    "sentence-transformers/static-similarity-mrl-multilingual-v1"
)
```
is running without any error.

Samoed:
> The error was coming from load_model

Yes, missed that part in description.

Wrapper for crossencoder should be also updated

ayush1298:
> Wrapper for crossencoder should be also updated

Updated it now.

PR #4233, opened 2026-03-12
Fix zeroentropy/zembed-1 metadata

blockingthesky:
This PR fixes three incorrect field values in the metadata for `zeroentropy/zembed-1`:

- `revision`: `fc4dc6ddd4a4f609978834de537b1f9f5cb29405` → `10378878bba40172305a1a979db64a413ab7da7b`
- `release_date`: `"2025-09-16"` → `"2026-03-02"`
- `max_tokens`: `40960` → `32768`


Samoed:
I don't think that max_tokens and revision should be changed, because they wasn't changed. Between revision only readme was changed, and max tokens is still 40960 https://huggingface.co/zeroentropy/zembed-1/blob/main/sentence_bert_config.json#L2

blockingthesky:
The correct value for `max_tokens` is `32768`; we [updated the README](https://huggingface.co/zeroentropy/zembed-1/blob/932370aeee09fef5ef7a0982fae3b37731a22d1e/README.md?code=true#L18) but thanks for flagging, looks like we also need to update that config.

PR #4245, opened 2026-03-17
Fix ColQwen3.5-4.5B - ColPaliEngineWrapper.encode() for multimodal datasets

athrael-soju:
### Changes

- Override `encode()` in `ColQwen3_5Wrapper` to route by `prompt_type`: queries use text, documents use images when available.
- Override `encode_input()` in `ColQwen3_5Wrapper` to clear stale `rope_deltas` cache before forward passes.

### DIFF
--- a/mteb/models/model_implementations/colqwen_models.py
+++ b/mteb/models/model_implementations/colqwen_models.py
@@ -189,7 +189,7 @@ class ColQwen3_5Wrapper(AbsEncoder):  # noqa: N801
                     imgs = [img.convert("RGB") for img in imgs]
                     inputs = self.processor.process_images(imgs)
                 else:
-                    inputs = self.processor.process_texts(texts)
+                    inputs = self.processor.process_queries(texts)
                 inputs = {k: v.to(self.device) for k, v in inputs.items()}
                 outs = self._encode_inputs(inputs)
                 all_embeds.extend(outs.cpu().to(torch.float32))


athrael-soju:
> I don't think that this is good idea to change base class of colpali models, because other than your model working correctly. With your implementation some tasks would work incorrectly (Vidorev3.1) where model would receive images and texts together

Thanks for the feedback.  I'll move the encode() and encode_input() overrides into ColQwen3_5Wrapper

Samoed:
Maybe with `self.mdl.rope_deltas = None` you won't need to change `encode` function? Your implementation would produce incorrect results a bit

athrael-soju:
> Maybe with `self.mdl.rope_deltas = None` you won't need to change `encode` function? Your implementation would produce incorrect results a bit

Tested with only `rope_deltas = None` and the original encode() fusion logic and it crashes with the previous error: `RuntimeError: The size of tensor a (3340) must match the size of tensor b (770) at non-singleton dimension 1.`

Samoed:
This is strange that images and text have different shapes

athrael-soju:
> This is strange that images and text have different shapes

ColQwen3.5's image processor has a much higher minimum resolution (shortest_edge=65536 vs 3136 for ColQwen2), producing ~3340 image tokens vs ~770 text tokens. ColQwen3 avoids this by processing both modalities in a single forward pass instead of separately fusing them.

At least that's what I think is the reason.

Samoed:
Yes, it is https://github.com/athrael-soju/mteb/blob/bb8ee1ffbac8f0dcc57484d751060931f39ff994/mteb/models/model_implementations/colqwen_models.py#L250-L277

Why you don't process this similarly?

athrael-soju:
> Yes, it is https://github.com/athrael-soju/mteb/blob/bb8ee1ffbac8f0dcc57484d751060931f39ff994/mteb/models/model_implementations/colqwen_models.py#L250-L277
> 
> Why you don't process this similarly?

Good question. Let me test that.

athrael-soju:
@Samoed Refactored ColQwen3_5Wrapper to follow the same pattern as ColQwen3Wrapper: inherits from AbsEncoder, processes text+image jointly so the shape mismatch never comes up, and clears rope_deltas before each forward pass as you suggested. Also fixed a bug where outputs[0] was silently slicing the batch dimension since ColQwen3_5.forward returns a plain tensor. 

Tested just now and seems ok.

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -85,36 +85,34 @@ def __init__(
         )
 
 
-class ColQwen3_5Wrapper(ColPaliEngineWrapper):  # noqa: N801
+class ColQwen3_5Wrapper(AbsEncoder):  # noqa: N801
```
Samoed, line 88: Maybe would be better to inherit from ColQwen3Wrapper?

athrael-soju: It would require overriding both _encode_inputs and get_fused_embeddings due to ColQwen3.5's different model output format and processor. It would work, but perhaps a little messier.

Samoed: I think we can keep as it is then. But you need to rerun your model

athrael-soju: Cool. Already running to fix linting.

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -85,36 +85,146 @@ def __init__(
...
+                        inputs["image_grid_thw"][:, 1] * inputs["image_grid_thw"][:, 2]
+                    )
+                    pixel_values = list(
+                        torch.split(inputs["pixel_values"], offsets.tolist())
+                    )
+                    inputs["pixel_values"] = torch.nn.utils.rnn.pad_sequence(
+                        pixel_values, batch_first=True
+                    )
+                else:
+                    inputs = self.processor.process_queries(texts)
```
Samoed:
```suggestion
                    inputs = self.processor.process_texts(texts)
```

athrael-soju:
@Samoed does everything look ok for a merge?

Samoed:
Looks good, but you need to rerun tasks firstly

athrael-soju:
> Looks good, but you need to rerun tasks firstly

I made a change to use process_images()/process_queries() separately instead of joint processing, matching the standard ColPali eval pipeline. The previous approach fused text+image embeddings via element-wise addition, which broke with ColQwen3.5's dynamic resolution. 

Re-running benchmarks now and will submit in https://github.com/embeddings-benchmark/results/pull/448

athrael-soju:
Added https://github.com/embeddings-benchmark/results/pull/448 with latest evals.

athrael-soju:
@Samoed can you please let me know what's holding up merging this? Model itself is already on HF and merged in vLLM & ColPali. The scores I generated initially are within 0.004 from the ones I commited after re-running with the mteb wrapper here.

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -85,36 +85,125 @@ def __init__(
...
 
-        super().__init__(
-            model_name=model_name,
-            model_class=ColQwen3_5,
-            processor_class=ColQwen3_5Processor,
-            revision=revision,
-            device=device,
+        self.model = ColQwen3_5.from_pretrained(
+            model_name,
+            torch_dtype=torch.bfloat16,
```
Samoed: I tried to run and still got exceptions. How did you evaluate your model?
```suggestion
```

athrael-soju: > I tried to run and still got exceptions. How did you evaluate your model?

I evaluated on a B200. What exceptions are you seeing?

Samoed: In this case `TypeError: colpali_engine.models.qwen3_5.colqwen3_5.modeling_colqwen3_5.ColQwen3_5.from_pretrained() got multiple values for keyword argument 'torch_dtype'`

athrael-soju: > In this case `TypeError: colpali_engine.models.qwen3_5.colqwen3_5.modeling_colqwen3_5.ColQwen3_5.from_pretrained() got multiple values for keyword argument 'torch_dtype'`

I've pushed your suggestion. See if that is fixed now

athrael-soju:
@Samoed have you tested after the latest push?

Samoed:
Added comment about results mismatch https://github.com/embeddings-benchmark/results/pull/448#discussion_r2954663550

Samoed (approved):

PR #29, opened 2022-07-28
Fix `hf_hub_name` of Sick-R STS task

cycycc:
Fixes #22.
Changes `hf_hub_name` of Sick-R STS task from `mteb/biosses-sts` to `mteb/sickr-sts`.

PR #4254, opened 2026-03-18
fix: skip single-label cluster sets in legacy clustering evaluation

### LINKED ISSUE #1866: Only one label in one of cluster set from ArxivClusteringP2P
```python
import mteb
t = mteb.get_task('ArxivClusteringP2P',languages=['eng'])
t.load_data()
print(len(set(t.dataset['test'][29]['labels'])))  # output: 1
```
Does the cluster set make any difference to main scores for different models?

fzoll:
## Summary
- Skip cluster sets with only one unique label in `AbsTaskClusteringLegacy._evaluate_subset`, logging a warning
- Cluster set #29 in ArxivClusteringP2P contains only the label `quant-ph`, always yielding `v_measure=1.0` regardless of model quality — inflating scores and wasting compute
- This fix benefits all legacy clustering tasks, not just ArxivClusteringP2P

## Test plan
- [x] All 155 existing abstask and evaluator tests pass
- [x] All 69 clustering-related tests pass

Fixes #1866

### DIFF
--- a/mteb/abstasks/clustering_legacy.py
+++ b/mteb/abstasks/clustering_legacy.py
@@ -116,6 +116,14 @@ class AbsTaskClusteringLegacy(AbsTask):
                 logger.info(
                     f"Running clustering on cluster ({i + 1}/{len(data_split)})"
                 )
+                labels = cluster_set[self.label_column_name]
+                if len(set(labels)) <= 1:
+                    logger.warning(
+                        f"Cluster set {i} has {len(set(labels))} unique label(s), "
+                        "skipping encode step — v_measure is always 1.0."
+                    )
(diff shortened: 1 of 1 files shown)


Samoed:
Does this affect results?

KennethEnevoldsen:
Yeah it does affect score. The class is already deprecated I would redo it such that the score remains the same (but we simply skip the encode step), though I am unsure if it is worth it (this abstask is already deprecated)

fzoll:
@KennethEnevoldsen I've reworked the approach — now instead of skipping single-label cluster sets entirely, we skip only the encode/clustering step and add `v_measure=1.0` directly to the results. This preserves backward-compatible scores while saving compute on trivial cluster sets.

PR #4262, opened 2026-03-20
fix: Add modality filtering to get_model_metas

dschechter27875:
Adds a modalities parameter to get_model_metas to filter models by supported modalities.

Also adds an `exclusive_modality_filter` flag to support exact modality matching.

Behavior:
- Default: returns models that include the requested modalities (subset match)
- Exclusive: returns models whose modalities exactly match the requested set

Examples:
get_model_metas(modalities=["text"])
get_model_metas(modalities=["text"], exclusive_modality_filter=True)

Close https://github.com/embeddings-benchmark/mteb/pull/4262

### DIFF
--- a/tests/test_models/test_model_meta.py
+++ b/tests/test_models/test_model_meta.py
@@ -368,4 +368,3 @@ def test_get_model_metas_without_modality_filter_returns_more_models():
     text_models = mteb.get_model_metas(modalities=["text"])
 
     assert len(all_models) > len(text_models)
-


mteb/models/get_model_meta.py
```diff
@@ -30,6 +30,8 @@ def get_model_metas(
     use_instructions: bool | None = None,
     zero_shot_on: list[AbsTask] | None = None,
     model_types: Iterable[str] | None = None,
+    modalities: Iterable[str] | None = None,
```
```suggestion
    modalities: Iterable[Modalities] | None = None,
```

dschechter27875: Thanks for the feedback! I’ll add tests for this and update the typing.

dschechter27875:
I just added tests and updated typing

KennethEnevoldsen:
@dschechter27875 I can't seem to find the tests - did you forget to add them to the PR?

dschechter27875:
> @dschechter27875 I can't seem to find the tests - did you forget to add them to the PR?

Good catch, sorry abt that — the tests are now included in the PR under tests/test_models/test_model_meta.py. They cover both subset and exact modality filtering.

tests/test_models/test_model_meta.py
```diff
@@ -345,3 +346,19 @@ def test_load_model_with_experiments():
         model2.mteb_model_meta.experiment_name
         == model_from_meta2.mteb_model_meta.experiment_name
     )
+
+
+def test_get_model_metas_modalities_subset():
+    models = mteb.get_model_metas(modalities=["text"])
```
KennethEnevoldsen, line 351: can you also test that I do get more when I don't add the filter

dschechter27875: Thx for the suggestion! I added a test verifying that calling get_model_metas without a modality filter returns more models than filtering for ["text"]. Also tried to clean up the unused import that was causing the lint failure.

PR #4271, opened 2026-03-22
Fix: align leaderboard summary tables to full benchmark tasks (#4263)

### LINKED ISSUE #4263: The results in leaderboard are not accurate because of skipna=True
### Describe the bug

I tried to evaluate `Alibaba-NLP/gte-modernbert-base` on 
```
retrieval_tasks = mteb.filter_tasks(benchmark, task_types=["Retrieval"])
```
and got a mean score of 57.03 across 10 tasks. But looking at the leaderboard result of the same benchmark and tasks it shows 54.02 !
After some investigation I think the cause of this is https://github.com/embeddings-benchmark/mteb/blob/main/mteb/benchmarks/_create_table.py#L65 when creating the table it skips tasks the model doesn't have cached results for. 
This model was probably evaluated on mteb v1 and is missing `ClimateFEVERHardNegatives` `FEVERHardNegatives` `HotpotQAHardNegatives` `Touche2020Retrieval.v3` so the average computed only on 6 available tasks.
Also I think there are differences in dataset versions between v1 and v2 so even if the task was evaluated in v1 its not the new version of the dataset. For example the leaderboard TRECCOVID score is 75.75 but evaluating v2 on this task result is 63.47

### To reproduce

```
import mteb

retrieval_tasks = mteb.filter_tasks(benchmark, task_types=["Retrieval"])
model = SentenceTransformer("Alibaba-NLP/gte-modernbert-base")
mteb.evaluate(model, tasks=retrieval_tasks)
```

### Additional information

_No response_

No

Rakshitha-Ireddi:
## Authors
- Ireddi Rakshitha
- Yashwanth Devavarapu

## Summary
Fixes incorrect benchmark means when a model is missing results for some tasks in the selected benchmark (e.g. older submissions without newer tasks). Previously, averages were computed only over tasks present in the results table, which overstated scores compared to a full-benchmark evaluation.

## Changes
- When `BenchmarkResults.benchmark` is set, reindex the per-task score matrix to include every leaf task in that benchmark; missing tasks are NaN.
- Use `skipna=False` for per-task-type means so incomplete coverage yields NaN instead of a mean over a subset.
- For public/private summary tables, derive public vs private task name lists from the benchmark when available; handle empty public/private lists.
- Use the benchmark’s full task list for zero-shot percentage when a benchmark is attached.

## Testing
- Added `tests/test_benchmarks/test_create_table.py` (incomplete STS12+STS13 benchmark with only STS12 scored → `Mean (Task)` is NaN).
- Ran `pytest tests/test_benchmarks/test_create_table.py tests/test_results/test_benchmark_results.py` (local env may need `pytest-rerunfailures` and optional deps for unrelated tests).

Closes #4263

### DIFF
--- a/mteb/benchmarks/_create_table.py
+++ b/mteb/benchmarks/_create_table.py
@@ -3,16 +3,46 @@ from __future__ import annotations
 import re
 from collections import defaultdict
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Literal, cast
 
 import numpy as np
 import pandas as pd
 
 import mteb
+from mteb.abstasks.abstask import AbsTask
+from mteb.abstasks.aggregated_task import AbsTaskAggregate
+from mteb.benchmarks.benchmark import Benchmark
 from mteb.get_tasks import get_task, get_tasks
 
 if TYPE_CHECKING:
     from mteb.results.benchmark_results import BenchmarkResults
+
+
+def _leaf_tasks_from_benchmark(benchmark: Benchmark) -> list[AbsTask]:
+    out: list[AbsTask] = []
+    for task in benchmark.tasks:
+        if task.is_aggregate:
+            out.extend(cast(AbsTaskAggregate, task).tasks)
+        else:
+            out.append(task)
+    return out
+
+
+def _all_benchmark_task_names(benchmark: Benchmark) -> list[str]:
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen:
Thanks for the proposed fix @Rakshitha-Ireddi 

I made a fix in #4283

which only target the skip-na

PR #4216, opened 2026-03-09
tests: Add test to ensure coverage of reference models

fzoll:
Issue: https://github.com/embeddings-benchmark/mteb/issues/3237

PR #4274, opened 2026-03-22
fix: update dataset card with descriptive category and paper title reference

### LINKED ISSUE #2807: Update dataset card template

https://huggingface.co/datasets/mteb/CodeFeedbackST/discussions/1#68484384790ce37286a91214

> [CoIR: A Comprehensive Benchmark for Code Information Retrieval Models](https://arxiv.org/abs/2407.02883)

> Would require us to parse the BibTeX to extract the title before auto-generating the sheet, but I think that is doable.
I also think the task category might be a bit misleading. Ideally, I would like it to be Retrieval (text-to-text) (t2t = text-to-text and the task.type is Retrieval)

fzoll:
## Summary
- Format task category as `{type} ({description})` (e.g. `Retrieval (text-to-text)`) instead of raw codes like `t2t`
- Extract paper title from BibTeX citation and format reference as a markdown link (e.g. `[CoIR: A Comprehensive Benchmark...](url)`) instead of a plain URL

**Before** (CodeFeedbackST example):
| | |
|---|---|
| Task category | t2t |
| Reference | https://arxiv.org/abs/2407.02883 |

**After**:
| | |
|---|---|
| Task category | Retrieval (text-to-text) |
| Reference | [CoIR: A Comprehensive Benchmark for Code Information Retrieval Models](https://arxiv.org/abs/2407.02883) |

## Test plan
- [x] Verified with CodeFeedbackST task — category renders as `Retrieval (text-to-text)`, reference renders as markdown link with paper title
- [x] Tested edge cases: no BibTeX, no reference URL, no category — all degrade gracefully
- [x] All existing tests pass (1232+ metadata validation tests, card tests, citation tests)

Closes #2807

### DIFF
--- a/mteb/abstasks/task_metadata.py
+++ b/mteb/abstasks/task_metadata.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import logging
+import re
 from collections.abc import Sequence
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, cast
@@ -272,6 +273,14 @@ TaskCategory = Literal[
 20. at2at: audio+text to audio+text
 21. v2t: video to text
 """
+
+_MODALITY_CODES: dict[str, str] = {
+    "t": "text",
+    "i": "image",
+    "a": "audio",
+    "v": "video",
+    "c": "category",
+}
 
 AnnotatorType = Literal[
     "expert-annotated",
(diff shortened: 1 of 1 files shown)


Samoed:
Seems that in this RP changes from https://github.com/embeddings-benchmark/mteb/pull/4216

KennethEnevoldsen:
Yeah let us get the https://github.com/embeddings-benchmark/mteb/pull/4216 to avoid reviewing code twice

fzoll:
@Samoed @KennethEnevoldsen Eh, my bad, mistakenly used the wrong base. I rebased; this should be clean now.

mteb/abstasks/task_metadata.py
```diff
@@ -273,6 +274,30 @@
 21. v2t: video to text
 """
 
+CATEGORY_DESCRIPTIONS: dict[str, str] = {
```
Samoed: We have mapping of valid categories in that will be applied https://github.com/embeddings-benchmark/mteb/blob/a29dc993d2aceb7e021b85e4fdb75455979bb7f9/mteb/abstasks/task_metadata.py#L827

mteb/abstasks/task_metadata.py
```diff
@@ -623,20 +648,50 @@ def _create_dataset_card_data(
...
             # parameters for readme generation
             dict(
                 citation=self.bibtex_citation,
                 dataset_description=self.description,
-                dataset_reference=self.reference,
+                dataset_reference=formatted_reference,
                 descriptive_stats=descriptive_stats,
                 dataset_task_name=self.name,
-                category=self.category,
+                category=formatted_category,
```
Samoed, line 658: This category only for table in 

Image

Probably we can remove it overall

mteb/abstasks/task_metadata.py
```diff
@@ -623,20 +648,50 @@ def _create_dataset_card_data(
...
                 domains=", ".join(self.domains) if self.domains else None,
             ),
         )
 
+    @staticmethod
+    def _extract_bibtex_title(bibtex_citation: str) -> str | None:
```
Samoed, line 675: I think it's better to move out from `metadata`

mteb/abstasks/task_metadata.py
KennethEnevoldsen: I think it is better the proposed way no?

Samoed: Agree, that's better, but this will dupulicate information from other dataset tags

mteb/abstasks/task_metadata.py
KennethEnevoldsen: I think it is quite decent here - though we could create a module for bibtex utilities (though that seems like a seperate issue)

Issue #4161, opened 2026-02-24
Figure out a better dataflow for the leaderboard


KennethEnevoldsen:
### Description of the feature

After having a lot of issues with a slow leaderboard (and maybe overly complex leaderboard here are a few suggestions - I would love to discuss a potential solution.

The data that we have right now is 

- model names along with their metadata
- results, which include model performance on tasks
  - notably this includes performance on subsets - e.g. while a model might have a score on MIRACL of 0.9 it might have a subscore on language 1 of 0.95 and on language 2 of 0.85.
- the task metadata
- the subset metadata (this is just a subset id to language mapping)

We use this data in various ways to for performance over time or by model size as well as the summary table diplaying rank and models scores.

My initial thought is to refactor the backend to use a SQL table. With a schema somewhat like this:

```
models(model_id, name, size, family, release_date, ...)
tasks(task_id, name, type, domain, ...)
subsets(subset_id, task_id, language, ...)
results(model_id, task_id, score, ...) # in principle not needed
subset_results(model_id, subset_id, score)
```

A query for performance over time could look something like this:

```sql
SELECT 
    m.name,
    m.release_date,
    AVG(r.score) as mean_score
FROM results r
JOIN models m ON r.model_id = m.model_id
GROUP BY m.model_id, m.name, m.release_date
ORDER BY m.release_date
```

---

A few things to be aware of:

The main table allows for custom summary tables per leaderboard and therefore also a custom aggregation metric.

PR #4276, opened 2026-03-22
fix: resolve API models in performance-over-time plot

### LINKED ISSUE #4141: Accidental filtering of APIs in the performance over time
seems like we filtered out APIs the performance over time plots (@Samoed)

Image

Image

fzoll:
## Summary
- API models (voyage, openai, cohere, etc.) were being filtered out of the "Performance over Time" plot because model metadata lookup only worked for HuggingFace URLs
- Added fallback resolution: first try HF URL extraction (existing), then match by (reference_url, display_name), then by display_name alone
- Also improved performance by building lookup dicts once per plot instead of per-row

## Test plan
- [x] Verified HuggingFace models still resolve correctly (e.g., all-MiniLM-L6-v2 → date=2021-08-30)
- [x] Verified API models now resolve (e.g., voyage-3-lite → date=2024-09-18, text-embedding-3-large → date=2024-01-25)
- [x] Verified unknown models gracefully return None
- [x] 54 non-HF API models now resolvable that were previously filtered out
- [x] Existing leaderboard tests pass

Fixes #4141

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -519,7 +519,7 @@ def get_leaderboard_app(
 
     logger.info("Step 5/7: Generating tables...")
     table_start = time.time()
-    summary_table = apply_summary_styling_from_benchmark(
+    summary_table, summary_raw = apply_summary_styling_from_benchmark(
         default_benchmark, filtered_benchmark_results
     )
     per_task_table = apply_per_task_styling_from_benchmark(
@@ -664,6 +664,7 @@ def get_leaderboard_app(
 
         scores = gr.State(default_scores)
         models = gr.State(filtered_models)
+        summary_data = gr.State(summary_raw)
         with gr.Row():
             with gr.Column(scale=1):
                 description = gr.Markdown(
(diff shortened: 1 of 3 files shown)


.github/workflows/reference_models.yml
KennethEnevoldsen, line 1: I don't think this should be included

KennethEnevoldsen (commented):
@fzoll can I ask you to run `mteb leaderboard` and see that these changes works as intended (adding a screenshot)

KennethEnevoldsen:
> I think it's better to change how we pass the summary table into plot functions, rather than hacking the parsing of this table

I think we can accept this for now (though I agree that it is not ideal), as the alternative requires more design (#4161)

fzoll:
@KennethEnevoldsen @Samoed I think this is now much simpler and works. @Samoed Is it what you mentioned?
Also, @KennethEnevoldsen here is the screenshot you asked:
leaderboard-performance-over-time-v2

KennethEnevoldsen:
@fzoll is this ready for review? Can you post a screenshot showing that is work as intended?

fzoll:
@KennethEnevoldsen Yes, i believe it is ready for review. Here are 2 screenshots: 
leaderboard_summary_v2
leaderboard_summary_final

Issue #3937, opened 2026-01-14
Examine and prevent circular imports


KennethEnevoldsen:
A friend suggest:
https://github.com/tach-org/tach

to prevent circular imports which we have had a few problems with. 

We don't have to use that tool specifically though.

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/pull/3800#pullrequestreview-3662621011_

PR #4277, opened 2026-03-22
fix: resolve circular imports and add CI enforcement

fzoll:
## Summary

- Eliminates **38 circular import chains** in the `mteb` package by moving top-level `import mteb` statements into function bodies or replacing them with direct submodule imports
- Adds `scripts/check_circular_imports.py` — an AST-based detector for runtime circular imports (no new dependencies)
- Integrates the check into `make lint` and `make lint-check` so CI prevents future regressions

## What changed

| File | Fix |
|------|-----|
| `mteb/abstasks/task_metadata.py` | Defer `import mteb` to method body |
| `mteb/cache.py` | Defer `import mteb` to function bodies |
| `mteb/benchmarks/_create_table.py` | Replace `import mteb` with direct `from mteb.models.get_model_meta import get_model_meta` |
| `mteb/deprecated_evaluator.py` | Defer `import mteb` to method body |
| `mteb/models/abs_encoder.py` | Defer `import mteb` to method bodies |
| `mteb/results/task_result.py` | Import `TaskMetadata` from submodule directly; defer `get_task` import |

The root cause in all cases was `import mteb` at the top level of modules that are themselves imported by `mteb/__init__.py`, creating re-entrant import chains. Python handles these by returning a partially-initialized module, which works but is fragile — any change to import ordering could cause `AttributeError` at import time.

## Context

Addresses https://github.com/embeddings-benchmark/mteb/issues/3937 (original suggestion by @KennethEnevoldsen in [PR #3800 review](https://github.com/embeddings-benchmark/mteb/pull/3800#pullrequestreview-3662621011)).

### DIFF
--- a/tests/test_result_cache_load_from_cache.py
+++ b/tests/test_result_cache_load_from_cache.py
@@ -98,7 +98,7 @@ class TestRebuildFromFullRepository:
         with (
             patch.object(cache, "download_from_remote") as mock_download,
             patch.object(cache, "load_results") as mock_load_results,
-            patch("mteb.models.get_model_meta.get_model_metas") as mock_get_model_metas,
+            patch("mteb.cache.get_model_metas") as mock_get_model_metas,
         ):
             # Mock model metas - None names should be filtered
             meta1 = MagicMock()
@@ -138,7 +138,7 @@ class TestRebuildFromFullRepository:
         with (
             patch.object(cache, "download_from_remote"),
             patch.object(cache, "load_results") as mock_load_results,
-            patch("mteb.models.get_model_meta.get_model_metas") as mock_get_model_metas,
+            patch("mteb.cache.get_model_metas") as mock_get_model_metas,
         ):
             meta = MagicMock()
             meta.name = "model1"


.github/workflows/reference_models.yml
KennethEnevoldsen, line 1: Hmm don't believe this one should be here

mteb/models/abs_encoder.py
```diff
@@ -166,7 +165,9 @@ def validate_task_to_prompt_name(
 
             if task_name not in task_types and task_name not in prompt_types:
                 try:
-                    mteb.get_task(task_name=task_name)
+                    from mteb.get_tasks import get_task
```
KennethEnevoldsen, line 168: why not set `get_task` as a top-level relative import

mteb/abstasks/task_metadata.py
```diff
@@ -588,6 +587,8 @@ def _create_dataset_card_data(
             multilinguality = "translated"
 
         if self.adapted_from is not None:
+            import mteb
```
KennethEnevoldsen: we probably should import mteb anywhere, and instead just import `get_tasks` using a relative import

KennethEnevoldsen:
I have also removed the "closes" for the issue as we still have a lot of functions that are moved inside the function, which is not ideal

fzoll:
@KennethEnevoldsen Kept as inline imports (circular dependency chains make top-level impossible):
  - mteb/abstasks/task_metadata.py: from mteb.get_tasks import get_tasks — cycle via get_tasks → filter_tasks → aggregated_task → task_result → task_metadata
  - mteb/results/task_result.py: from mteb.get_tasks import get_task — cycle via get_tasks → filter_tasks → aggregated_task → task_result

scripts/check_circular_imports.py
Samoed, line 1: Not sure if we need this script. Cycles will be checked on mteb import

KennethEnevoldsen: but this one is more restrictive - I could also see us not adding this one, but in the future add rules like:

https://github.com/astral-sh/ruff/issues/2914

Samoed: > but this one is more restrictive

Maybe, but this is hard to get from code and since that `lint` is passing I think it's not catches all cases

KennethEnevoldsen: Yeah - I would also lean toward removing it (useful here, but we don't want to keep it around)

Samoed:
Here you've also added https://github.com/embeddings-benchmark/mteb/pull/4216

fzoll:
@KennethEnevoldsen @Samoed Yeah, sorry again, the solution of #4216 is in my main (mistakenly), so i have to take extra care.

tests/test_result_cache_load_from_cache.py
```diff
@@ -98,7 +98,7 @@ def test_full_rebuild_process(self, tmp_path):
         with (
             patch.object(cache, "download_from_remote") as mock_download,
             patch.object(cache, "load_results") as mock_load_results,
-            patch("mteb.get_model_metas") as mock_get_model_metas,
+            patch("mteb.cache.get_model_metas") as mock_get_model_metas,
```
KennethEnevoldsen: this seems like the wrong path 


```suggestion
            patch("mteb.models.get_model_metas") as mock_get_model_metas,
```

tests/test_result_cache_load_from_cache.py
```diff
@@ -138,7 +138,7 @@ def test_rebuild_error_propagation(self, tmp_path):
         with (
             patch.object(cache, "download_from_remote"),
             patch.object(cache, "load_results") as mock_load_results,
-            patch("mteb.get_model_metas") as mock_get_model_metas,
+            patch("mteb.cache.get_model_metas") as mock_get_model_metas,
```
KennethEnevoldsen, line 141:
```suggestion
            patch("mteb.models.get_model_metas") as mock_get_model_metas,
```

Samoed:
Seems that test failing because we're import `get_model_metas` in different names somewhere inside

fzoll:
@Samoed Checking

PR #4283, opened 2026-03-22
fix tasktype aggregation

### LINKED ISSUE #4263: The results in leaderboard are not accurate because of skipna=True

KennethEnevoldsen:
Close #4263

Screenshot 2026-03-22 at 22 31 40

### DIFF
--- a/mteb/benchmarks/_create_table.py
+++ b/mteb/benchmarks/_create_table.py
@@ -81,7 +81,7 @@ def _get_means_per_types(per_task: pd.DataFrame):
                 dict(
                     model_name=model_name,
                     task_type=task_type,
-                    score=scores[tasks].mean(skipna=True),
+                    score=scores[tasks].mean(skipna=False),
                 )
             )
     return pd.DataFrame.from_records(records)


Samoed:
Tried to find missing score for solon on existing leaderboard, but can't find such one

KennethEnevoldsen:
It is just a local result cache

PR #4232, opened 2026-03-12
fix: Refactor queries and document dataloader to allow multiple modalities

ayush1298:
closes #4182 

Both  `_create_queries_dataloader` and `_create_document_dataloader` are refactored. 
Before refactor, these function just used 1 of the modality from `[ image, text, audio, video ] `, and uses that only to create dataloader. So, its giving issue for tasks like `EncyclopediaVQAIT2ITRetrieval` with both `['image', 'text']` modality in dataset.

In the refactor, a helper function `_prepare_multimodal_dataset` is used which does all dataset specific transformation together and then dataloader conversion is happened in respective function like `_create_queries_dataloader` and `_create_document_dataloader`. So, this now handles a dataset with more than 1 modality. 

Tested with `EncyclopediaVQAIT2ITRetrieval` using below dummy script: 

```
from datasets import Dataset
from datasets import Image as HFImage
from PIL import Image as PILImage

from mteb._create_dataloaders import _create_queries_dataloader
from mteb.tasks.retrieval.eng import EncyclopediaVQAIT2ITRetrieval
from mteb.types import PromptType

# Get the real task metadata
task = EncyclopediaVQAIT2ITRetrieval()
task_metadata = task.metadata

# Create a small synthetic dataset mimicking the real one:
# features: ['id', 'modality', 'text', 'image'], where image is HF Image type
dummy_images = [PILImage.new("RGB", (32, 32), color="red") for _ in range(4)]
dataset = Dataset.from_dict(
    {
        "id": ["q1", "q2", "q3", "q4"],
        "modality": ["image_text"] * 4,
        "text": ["What is this?", "Describe this.", "Name this.", "Identify this."],
        "image": dummy_images,
    }
).cast_column("image", HFImage())

# Verify: category is "it2it", so query modalities = ["image", "text"]
print("Query modalities:", task_metadata.get_modalities(PromptType.query))
# Expected: ['image', 'text']

# Now test the refactored function
from mteb.types import PromptType

dl = _create_queries_dataloader(
    dataset,
    task_metadata,
    batch_size=2,
    input_column=None,
    num_proc=None,
)

batch = next(iter(dl))
print("Batch keys:", batch.keys())
# Before refactor: dict_keys(['image'])
# After refactor:  dict_keys(['id', 'modality', 'text', 'query', 'image'])

# Verify text was processed (query key exists)
print("Has 'text':", "text" in batch)
print("Has 'query':", "query" in batch)
print("Has 'image':", "image" in batch)

```

[...]

PR #4285, opened 2026-03-23
fix: optionally pass embed_dim

Samoed:
Optionally pass `embed_dim` to the model, because some of them don't support them. Ref https://github.com/embeddings-benchmark/mteb/pull/4232#issuecomment-4109525281


Samoed:
Strangely automerge didn't merge

Issue #4273, opened 2026-03-22
Vidore (1-3) Doesn't load in the Leaderboard


athrael-soju:
### Describe the bug

When opening the mteb leaderboard Space, and clicking the Vidore V3 button, there is some loading, but the view doesn't switch.

### To reproduce

Open the mteb leaderboard Space and click Vidore V3, or go under other/ViDoRe (V1&V2)

### Are you interested to contribute a fix for this bug?

Yes

Samoed:
Yes, in logs
```
ValueError: An event handler (update_task_list) didn't receive enough input values (needed: 6, got: 0).
Check if the event handler calls a Javascript function, and make sure its return value is correct.
Wanted inputs:
    [<gradio.components.state.State object at 0x7f70ff4a2db0>, <gradio.components.checkboxgroup.CheckboxGroup object at 0x7f718ed9e360>, <gradio.components.checkboxgroup.CheckboxGroup object at 0x7f718df0d730>, <gradio.components.checkboxgroup.CheckboxGroup object at 0x7f718c8a34d0>, <gradio.components.checkboxgroup.CheckboxGroup object at 0x7f718df82e70>, <gradio.components.state.State object at 0x7f70ff504380>]
Received inputs:
    []
```

Samoed:
Hm, I can't reproduce locally

KennethEnevoldsen:
I also can't reproduce in on the leaderboard

<img width="1383" height="938" alt="Image" src="https://github.com/user-attachments/assets/c5670cef-94ce-4a73-b0f3-cebeb80a1896" />

Samoed:
@KennethEnevoldsen Kalm and qwen3-embedding does not support images and this table comes from `Mteb(multilingual, v2)`

KennethEnevoldsen:
Hmm seems to be a general issue that the table is re-used

Samoed:
This happens only for vidore, so I'm not sure why this can happen

Samoed:
This can be caused by https://github.com/embeddings-benchmark/mteb/issues/4280

KennethEnevoldsen:
> This happens only for vidore, so I'm not sure why this can happen

right, but if you go to vidore, then it you have to reload to fix it

KennethEnevoldsen:
Anyway I have added a fix and enables automerge assuming tests pass

Samoed:
[...]

PR #4287, opened 2026-03-23
fix: Revert gradio version bump

KennethEnevoldsen:
After a lot of local testing this seems to resolve the issue

intends to fix: #4273


KennethEnevoldsen:
apparently it is up to date - I suspect this is why I seemingly had the error fixes earlier (probably ran using `mteb leaderboard` instead of `make run-leaderboard`)

PR #4291, opened 2026-03-23
fix: handle empty sim_scores in confidence_scores

fzoll:
## Summary
- Handle empty `sim_scores` list in `confidence_scores()` by returning zeros instead of crashing with `IndexError: list index out of range`
- This occurs when BM25 returns no results for a query in reranking tasks

## Test plan
- Ran BM25 evaluation on all 8 reranking tasks that previously failed with this error — all succeed now


fzoll:
Moving this fix into PR #4216 instead

Issue #4295, opened 2026-03-24
`uv` `0.11.0` incompatability


Samoed:
After uv 0.11.0 release in ci we get this error
```
  × Failed to download and build `flash-attn==2.8.3`
  ╰─▶ Extra build requirement `torch` was declared with `match-runtime
      = true`, but `flash-attn` does not declare static metadata, making
      runtime-matching impossible
  help: `flash-attn` (v2.8.3) was included because `mteb[flash-attention]`
        (v2.11.5) depends on `flash-attn>=2.6.3`
make: *** [Makefile:8: install-for-tests] Error 1
```

Samoed:
Created an issue https://github.com/astral-sh/uv/issues/18690

PR #4296, opened 2026-03-24
Pin uv version

Samoed:
Pin uv version to make possible to run CI https://github.com/embeddings-benchmark/mteb/issues/4295


KennethEnevoldsen:
thanks!

PR #4304, opened 2026-03-25
fix: Showing loading screen while changing benchmark

### LINKED ISSUE #4137: leaderboard: reset to main table when switching benchmarks
Given that we currently reset the plots based on the main table, it can sometimes be confusing to not see that it is loading. A solution could be to reset the view to the main table every time you switch benchmark.

#4134

ayush1298:
closes #4137 
This will show a loading screen for the plot itself, not just for the radar plot, but also for performance over time and performance per task type plot as well. 

The Video is:

https://github.com/user-attachments/assets/a0f92933-14bd-486c-adbd-b5852c7fc208


mteb/leaderboard/app.py
```diff
@@ -866,6 +869,10 @@ def on_benchmark_select(benchmark_name, session_id=None):
                 gr.update(visible=show_zero_shot),
                 initial_models,
                 gr.update(visible=display_radar),
+                gr.update(value=summary_raw),
```
Samoed, line 872: Not sure that you need to return `gr.update(value=summary_raw)`, because `summary_raw` is just a variable

ayush1298: I am also not sure about these. Without this, it was not showing the loading screen. But, will check these again

KennethEnevoldsen:
Hmm I found this error here:

https://github.com/user-attachments/assets/9cf71a47-8006-4546-a5cd-80603dc89964

ayush1298:
I guess this is after, https://github.com/embeddings-benchmark/mteb/pull/4287 is merged. Will check

Samoed:
This gardio thing happens on main leaderboard too. It reqries to press multilingual part multiple times. In logs
```
gradio.route_utils.FnIndexInferError: Could not infer function index for API name: run_leaderboard
```
```
ValueError: An event handler (update_tables) didn't receive enough input values (needed: 5, got: 2).
Check if the event handler calls a Javascript function, and make sure its return value is correct.
Wanted inputs:
    [<gradio.components.state.State object at 0x7f5a3db166f0>, <gradio.components.checkboxgroup.CheckboxGroup object at 0x7f5a3f167740>, <gradio.components.state.State object at 0x7f5a3e3c1e80>, <gradio.components.state.State object at 0x7f5a3efe3020>, <gradio.components.checkboxgroup.CheckboxGroup object at 0x7f59ab3d3950>]
Received inputs:
    [None, None]
```
```
  File "/mteb/mteb/leaderboard/benchmark_selector.py", line 207, in _update_value
    return label_to_value[label]
           ~~~~~~~~~~~~~~^^^^^^^
KeyError: 'scandinavian'
ERROR:    Exception in ASGI application
```

KennethEnevoldsen:
Damn this gradio leaderboard is so fidgety

Samoed:
Yeah, maybe we should move leaderboard to different framework to speedup

KennethEnevoldsen:
> Yeah, maybe we should move leaderboard to different framework to speedup

Or at least to reduce bugs. 

I think it is worth looking at some overall design considerations -- I mention this a bit in #4161

Talked with a friend about it, and he suggested moving to a react frontend. Either loading the data in memory (similar to what the legacy mteb leaderboard does) or using a fastAPI backend. 

My only problem is that I don't know a lot of react. There is also the problem that it is somewhat misaligned with the producing a scientific output. Though I could imagine making a paper like "maintaining MTEB", but for visual design on the leaderboard (the workshop where we submitted that seems like a reasonable venue).

Samoed:
> Talked with a friend about it, and he suggested moving to a react frontend

I think this would be too much for us. Probably we can use starlette/Streamlit/nicegui to keep python, but this would require a lot of css probably

uv.lock
```diff
@@ -5263,7 +5264,7 @@ wheels = [
 
 [[package]]
 name = "mteb"
-version = "2.11.4"
+version = "2.11.10"
```
ayush1298: @KennethEnevoldsen This version is `2.12.0` in main, So what have to use here exactly. Other change is same as main

KennethEnevoldsen:
> I think this would be too much for us. Probably we can use starlette/Streamlit/nicegui to keep python, but this would require a lot of css probably

Let us move this to an issue to discuss it

uv.lock
ayush1298: What should we do here?

Samoed: Keep from main

ayush1298:
I have replace video in PR description after merging main. Seems working fine for me. @KennethEnevoldsen Could you check it again.

PR #4308, opened 2026-03-27
fix: Ensure that experiments_kwargs are passed for get_model_meta

### LINKED ISSUE #4307: Passed `experiment_kwargs` in `get_model_meta` is not used when loading model
When running:
```py
meta = mteb.get_model_meta(modelname, experiment_kwargs={"prompts": {taskname: prompt}})
mdl = meta.load_model()
```
I get:
```
taskname in mdl.model.prompts # False
```

Expected the same as:
```
mdl= mteb.get_model(modelname, prompts={taskname: prompt})
taskname in mdl.model.prompts # true
```

KennethEnevoldsen:
fixes #4307

(also accidentally pushed to main, but reverted it)

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -346,6 +346,7 @@ class ModelMeta(BaseModel):
             revision=_self.revision,
             **_kwargs,
         )
+        _self.loader_kwargs = _kwargs
         model.mteb_model_meta = _self  # type: ignore[misc]
         return model
 


mteb/models/model_meta.py
```diff
@@ -332,7 +332,7 @@ def load_model(
 
         # Allow overwrites
         _kwargs = self.loader_kwargs.copy()
-        _kwargs.update(kwargs)
+        _kwargs.update(self.experiment_kwargs)
```
```suggestion
        _kwargs.update(self.experiment_kwargs if self.experiment_kwargs is not None else {})
```

Samoed: Maybe also we can add after 
https://github.com/embeddings-benchmark/mteb/blob/9d742d54590cd50967c8f23d24064c1695bdd740/mteb/models/model_meta.py#L327-L331
```python
elif len(kwargs) == 0 and self.experiment_kwargs is not None:
    kwargs = self.experiment_kwargs
```

mteb/models/model_meta.py
```diff
@@ -295,53 +295,58 @@ def load_model(
         **kwargs: Any,
     ) -> MTEBModels:
         """Loads the model using the specified loader function."""
-        if self.loader is None:
+        # create a copy so that changing the model meta on the model does not influence the original meta
+        _self = self.model_copy(deep=True)
```
KennethEnevoldsen, line 299: Had to move the deep copy here to avoid overwriting it in get_model_meta (all tests worked individually fine but they had a side effect of adding exp. kwargs)

mteb/models/get_model_meta.py
```diff
@@ -125,18 +125,8 @@ def get_model(
...
-    meta = get_model_meta(model_name, revision, fetch_from_hf=True).model_copy(
-        deep=True
-    )
+    meta = get_model_meta(model_name, revision, fetch_from_hf=True)
     model = meta.load_model(device=device, embed_dim=embed_dim, **kwargs)
-
-    if kwargs:
-        logger.info(
-            f"Model '{model_name}' loaded with additional arguments: {list(kwargs.keys())}"
-        )
```
KennethEnevoldsen, line 136: moved this to `load_model`

KennethEnevoldsen (commented):
@Samoed had to make a slightly larger change here - can you re-review?

PR #4239, opened 2026-03-15
feat: Add submit_results to ResultsCache

ayush1298:
closes #4238
This is 3rd and final PR as part of adding `submit_results` feature into MTEB. Last 2 parts where: #4315, #4329 

This PR added `submit_results` to `ResultsCache`, which will be the main entry point for submitting results to MTEB. This aims to reduce the friction of submitting results to MTEB [result repo](https://github.com/embeddings-benchmark/results), and automating the process to the extent possible, from taking unsubmitted results in local to either giving steps to create a PR (`if create_pr=False`)/automatically creating a PR (`if create_pr=True`). 

Overall there are following key components: 

1. ReversibleWorkflow Pattern (added in PR #4239 under `mteb/_reversible_workflow/git_actions.py`)
       Purpose: Execute git operations with automatic rollback on failure
      - Sequential execution of actions
      -  If any action fails: undo all previous steps in reverse order
      -  Ensures repository always consistent (no half-completed operations)

2. Git Actions (added in PR #4239 under `mteb/_reversible_workflow/git_actions.py`)
     Concrete reversible actions for git operations:
    - `CreateBranchAction`       [Create branch → undo: delete branch + restore original]
    - `CommitAction`             [Create commit → undo: reset to previous commit]
    - `PushToForkAction`         [Push to fork → undo: force-push to revert]
    - `CreatePRAction`           [Create PR → undo: close PR]
    - `RestoreOriginalBranchAction` [Restore to original branch → undo: return to previous]

3. All git related utils functions like creating branch, restoring branch, deleting branch, checking uncommited changed, checking detached head state, getting github token,  creating PR, and post PR cleaup are added under `mteb/_reversible_workflow/git_utils.py`

4. Pre-flight Validation checks are used directly before workflow run which prevents data corruption before the workflow starts:
      - `_check_uncommitted_changes()`  [Detect staged changes that would be included]
      - `_check_detached_head()`        [Ensure on valid branch]

5. Actual `submit_results` and its helper functions to create PR content, pushing actual PR, and some other functions.

The flow of calls will be as follows: 

```
submit_results(models, create_pr=False)  [Main Entry]
│
├─ _normalize_models()              [Convert input to standard format]
│   └─ Handle: str, ModelMeta, list of both
│
├─ download_from_remote()           [Ensure remote repo exists locally]
│   └─ Clone or pull from GitHub
│
├─ _get_unsubmitted_results()       [Find what hasn't been submitted]
│   └─ Compare local vs remote files
│
[...]

PR #4315, opened 2026-03-28
fix: Make `ModelMeta` hashable

ayush1298:
Made ModelMeta hashable using `name, revision, experiment_kwargs and embed_dim.`
Also added test for this feature.
This is 1st PR as part of dividing #4239 into 3 parts.

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -141,7 +141,7 @@ class ModelMeta(BaseModel):
         output_dtypes: Output embedding data types (e.g. int8, binary, float) natively supported by the model. If None, it is assumed that the model only returns float embeddings.
     """
 
-    model_config = ConfigDict(extra="forbid", frozen=True)
+    model_config = ConfigDict(extra="forbid")
 
     # loaders
     loader: Callable[..., MTEBModels] | None
@@ -173,6 +173,16 @@ class ModelMeta(BaseModel):
     contacts: list[str] | None = None
     experiment_kwargs: Mapping[str, Any] | None = None
     output_dtypes: OutputDType | list[OutputDType] | None = None
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        """Deprecation warning for direct attribute mutation. Use model_copy(update={...}) instead."""
+        warnings.warn(
+            f"Mutating '{name}' is deprecated and will be removed in future versions. "
+            "Use .model_copy(update={...}) instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        super().__setattr__(name, value)
 
     @model_validator(mode="before")
     @classmethod


tests/test_models/test_model_meta.py
```diff
@@ -87,6 +87,86 @@
 ]
 
 
+def test_model_meta_hashable():
+    """Test that ModelMeta is hashable and considers experiment_kwargs in hash/equality."""
+    model1 = ModelMeta(
```
Samoed: Can you simplify a bit?
```suggestion
    model1 = ModelMeta.create_empty(
```

mteb/models/model_meta.py
```diff
@@ -287,6 +287,41 @@ def _check_name(cls, v: str | None) -> str | None:
...
+        """
+        # Serialize experiment_kwargs to a deterministic, hashable representation
+        exp_kwargs_repr = (
+            _serialize_experiment_kwargs_to_name(self.experiment_kwargs)
+            if self.experiment_kwargs
+            else None
+        )
+
+        if isinstance(self.embed_dim, Sequence) and not isinstance(
+            self.embed_dim, (str, bytes)
```
Samoed: Why embed_dim would be str or bytes?

mteb/models/model_meta.py
```diff
@@ -287,6 +287,41 @@ def _check_name(cls, v: str | None) -> str | None:
...
+        ):
+            embed_dim_hash = tuple(self.embed_dim)
```
```suggestion
            embed_dim = tuple(self.embed_dim)
```

mteb/models/model_meta.py
```diff
@@ -287,6 +287,41 @@ def _check_name(cls, v: str | None) -> str | None:
...
+        return hash((self.name, self.revision, exp_kwargs_repr, embed_dim_hash))
+
+    def __eq__(self, other: object) -> bool:
+        """Check equality based on name, revision, experiment_kwargs and embed_dim.
+
+        Two ModelMeta instances are equal if they have the same name, revision, experiment_kwargs and embed_dim.
+        """
+        if not isinstance(other, ModelMeta):
+            return NotImplemented
+        return (
```
Samoed: I think we can just `hash(self) == hash(other)`

mteb/models/model_meta.py
ayush1298: I didn't check this one, copilot suggest it above

ayush1298: Removing it

tests/test_models/test_model_meta.py
ayush1298: Made this change

mteb/models/model_meta.py
ayush1298: I guess we don't.
Because when `a == b`, then `hash(a) == hash(b)` is always true
But reverse may not hold always as collision can happen. `hash(a) == hash(b)` does NOT mean `a == b`

Samoed: Maybe we can also `model.to_dict() == other.to_dict()`?

ayush1298: I dont think we should use that also, like 1 of thing there is: 
```
if isinstance(meta.embed_dim, Sequence):
            meta.embed_dim = max(meta.embed_dim)
```
so this will cause issue

mteb/models/model_meta.py
```diff
@@ -287,6 +287,43 @@ def _check_name(cls, v: str | None) -> str | None:
...
+            )
+        )
+
+    def __eq__(self, other: object) -> bool:
+        """Check equality based on name, revision, experiment_kwargs and embed_dim.
+
+        Two ModelMeta instances are equal if they have the same name, revision, experiment_kwargs and embed_dim.
+        """
+        if not isinstance(other, ModelMeta):
+            return NotImplemented
```
KennethEnevoldsen, line 329:
```suggestion
            return False
```

mteb/models/model_meta.py
```diff
@@ -287,6 +287,34 @@ def _check_name(cls, v: str | None) -> str | None:
...
+        return (
+            self.name == other.name
+            and self.revision == other.revision
+            and self.experiment_kwargs == other.experiment_kwargs
+            and self.embed_dim == other.embed_dim
+        )
```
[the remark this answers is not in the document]
KennethEnevoldsen: Ahh yeah - it would be great to make it immutable. Probably not backward compatible though

mteb/models/model_meta.py
KennethEnevoldsen: Why not just compare the dumped json?

ayush1298: are you saying using `json.dumps,` not sure about these, but when I searched for this online, it showing its better to compare attributes by attributes

KennethEnevoldsen: Hmm - I am actually fine leaving it as it is right now. Easy to understand, and the only dimensions that we care about in terms of comparing models when loading. It could be a bit confusing for users that the metadata itself not a 1-1 match though

KennethEnevoldsen: An simple alternative is to use `meta.model_dump()`

ayush1298: Maybe we can. Not sure though, as I found nothing on this. The main thing is about whether it will bound or violate to Python's hash/equality contract

mteb/models/model_meta.py
ayush1298: I guess this is standard python protocol to keep it as NotImplemented

mteb/models/model_meta.py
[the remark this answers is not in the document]
ayush1298: Added this one

mteb/models/compression_wrappers/compression_wrapper.py
```diff
@@ -62,9 +61,14 @@ def __init__(
...
-                clipping_margin
-            )
+            exp_kwargs["clipping_margin"] = list(clipping_margin)
+
+        model.mteb_model_meta = meta.model_copy(
+            update={
+                "output_dtypes": [output_dtype],
+                "experiment_kwargs": exp_kwargs,
+            }
+        )  # type: ignore[misc]
```
ayush1298: Need this because, model is typed as `EncoderProtocol,` and in protocol, `mteb_model_meta` is a read-only property (no setter in protocol type). So typecheck is giving error

mteb/models/model_meta.py
KennethEnevoldsen: I am not sure what you mean by that?

mteb/models/model_meta.py
KennethEnevoldsen: Yeah I see that you are right

mteb/models/model_meta.py
ayush1298: >  when a == b, then hash(a) == hash(b) is always true
> But reverse may not hold always as collision can happen. hash(a) == hash(b) does NOT mean a == b

this thing

KennethEnevoldsen: if a==b then b==a is also true when using `meta.model_dump()` (you can add a test on that as well)

ayush1298: okay, will change these. I have already added test

ayush1298: Should I changed it to something like below? 

```
def __eq__(self, other: object) -> bool:
    self_dict = self.model_dump()
    other_dict = other.model_dump()
    return self_dict == other_dict
```

KennethEnevoldsen: assuming the other models is ModelMeta - so you need to add a type check

ayush1298: added

mteb/models/model_meta.py
```diff
@@ -141,7 +141,7 @@ class ModelMeta(BaseModel):
         output_dtypes: Output embedding data types (e.g. int8, binary, float) natively supported by the model. If None, it is assumed that the model only returns float embeddings.
     """
 
-    model_config = ConfigDict(extra="forbid")
+    model_config = ConfigDict(extra="forbid", frozen=True)
```
Samoed: I'm not sure if we want to activate `frozen` now

Samoed (commented):
Other than `frozen` looks good

mteb/models/model_meta.py
ayush1298: Any reason, why not?

KennethEnevoldsen: backward compatibility

ayush1298: Then, we can't ensure I guess `ModelMeta` to be `immutable`

KennethEnevoldsen: correct - which probably the issue with giving it a hash value 

We of course don't want to break backward compatibility, but I am not sure what the best direction forward is. One option is creating a FrozenModelMeta instead, but I am not sure that the the best one.

Another is to slowly move toward a frozen ModelMeta, by setting something like:

```py
    def __setattr__(self, name, value):
        warnings.warn(
            f"Mutating '{name}' is deprecated and will be removed in future versions. "
            "Use .model_copy(update={{...}}) instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        super().__setattr__(name, value)
```

That way we keep it unfrozen, but we incentivize people to not set attributes.

Let me know what you guys think about this

Samoed: I think we can use this `setattr`

ayush1298: same, I guess `setattr` is better

ayush1298: Added these

Issue #4127, opened 2026-02-21
Enable linting rule to prevent too many positional arguments


KennethEnevoldsen:
### Description of the feature

Having too many positional arguments can make a function harder to debug (misplaced positional arguments)  and refactor (switching the order of two arguments now becomes a breaking change).

ruff docs: https://docs.astral.sh/ruff/rules/too-many-positional-arguments/

related to #2416

ayush1298:
@KennethEnevoldsen I think we have to add below in `pyproject.toml` :

```
[tool.ruff.lint]
enable = ["PLR0917"] # Enable the rule

[tool.ruff.lint.pylint]
max-positional-args = 4 # What will be good value to set here?
```

Samoed:
I'll do this. I want to update ruff rules a bit

KennethEnevoldsen:
I am probably leaning towards 3, but the rule is still in preview (not sure we want to buy enable preview)

KennethEnevoldsen:
Though I would love to have it enabled

PR #4317, opened 2026-03-28
activate PLR rules

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4127

Activate PLR rule

### DIFF
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -100,7 +100,7 @@ def _get_loader_name(
     return loader.__name__
 
 
-class ModelMeta(BaseModel):
+class ModelMeta(BaseModel):  # noqa: PLR0904
     """The model metadata object.
 
     Attributes:


pyproject.toml
```diff
@@ -281,10 +301,22 @@ typing-modules = ["typing_extensions", "mteb.types", "collections.abc"]
...
+    # preview
+    "PLR0917",
 ]
 "mteb/tasks/*__init__.py" = ["F403"]  # undefined import `from .lang import *`
-"mteb/tasks/*" = ["D", "DOC"]
-"mteb/models/model_implementations/*" = ["D", "DOC"]
+"mteb/tasks/*" = [
+    "D",
+    "DOC",
+    "PL",
```
KennethEnevoldsen: what is PL?

pyproject.toml
```diff
@@ -265,9 +266,28 @@ ignore = [
...
+    "PLR6301",
+    "PLW0108",
+    "PLR0914",
+    "PLW1514",
+    "PLR0904",
+    "PLR6104",
 ]
 future-annotations = true  # For TC rules
 typing-modules = ["typing_extensions", "mteb.types", "collections.abc"]
+preview = true
```
KennethEnevoldsen, line 283: are we sure we want to enable preview rules?

pyproject.toml
```diff
@@ -265,9 +266,28 @@ ignore = [
     "D107",     # Missing docstring in __init__
     "D415",     # First line should end with a period
     "C408",     # don't use unecc. collection call, e.g. dict over {}
+    "PLR2004",   # magic value comparison
+    "PLC0415",   # import outside of top-level
+    "PLR0912",  # too many branches
+    "PLR0915",  # too many statements
+    "PLR0913",  # too many arguments
```
KennethEnevoldsen: Seems like a thing I would like to add (there is actually a few of these that would be nice to add)

Note that you can add `noqa` to all places where a rules apply to add the rule without having to fix everywhere (can be a good way to enable a rule for all future cases)

mteb/load_results.py
```diff
@@ -49,6 +49,7 @@ def _model_name_and_revision(
     + "Please use the `ResultCache`'s `.load_results` method instead."
 )
 def load_results(
+    *,
```
KennethEnevoldsen: public facing - I don't think we want to break backward compat

mteb/get_tasks.py
```diff
@@ -317,6 +317,7 @@ def get_tasks(
 
 def get_task(
     task_name: str,
+    *,
```
KennethEnevoldsen: I like it, but I am not sure we want to break backward compatibility for it

mteb/cache.py
```diff
@@ -146,6 +146,7 @@ def load_task_result(
         self,
         task_name: str,
         model_name: str | ModelMeta,
+        *,
```
KennethEnevoldsen: backward compat

pyproject.toml
KennethEnevoldsen: but these might be for a seperate PR

pyproject.toml
KennethEnevoldsen: I am unsure if we completely want to disable it for tests (there is at least some test with >4 args that are hard to get an overview on)

pyproject.toml
```diff
@@ -281,10 +301,22 @@ typing-modules = ["typing_extensions", "mteb.types", "collections.abc"]
...
-"mteb/models/model_implementations/*" = ["D", "DOC"]
+"mteb/tasks/*" = [
+    "D",
+    "DOC",
+    "PL",
+]
+"mteb/models/model_implementations/*" = [
+    "D",
+    "DOC",
+    "PL",
```
KennethEnevoldsen: would also enable it here - but just add noqa where needed (there is [a command for it](https://docs.astral.sh/ruff/linter/#inserting-necessary-suppression-comments))

pyproject.toml
Samoed: Yes, but this creates 100 linting errors, which we can't resolve easily

pyproject.toml
Samoed: Added noqa, but I think this is too much

KennethEnevoldsen: But the goal is to ensure future additions

Samoed: I agree, but even our models `__init__` or `encode` would trigger some of PL rules

KennethEnevoldsen: Right because the encode protocol does not have positional arguments - we could add that

Samoed: It triggers [`too-many-arguments`](https://docs.astral.sh/ruff/rules/too-many-arguments/#too-many-arguments-plr0913), but we can tweak it, of course

KennethEnevoldsen: Ahh, yeah default is 5, we could set it higher

pyproject.toml
```diff
@@ -322,6 +334,9 @@ runtime-evaluated-base-classes = [
     "mteb.abstasks.task_metadata.TaskMetadata",  # https://github.com/astral-sh/ruff/issues/7866
 ]
 
+[tool.ruff.lint.pylint]
+max-args = 10
```
KennethEnevoldsen: 7-9?

KennethEnevoldsen (approved):
typechecking seems to fail now - but otherwise I think this is good

pyproject.toml
KennethEnevoldsen: alright let us leave them out for now

KennethEnevoldsen:
@Samoed can you update the description (to avoid stating that the change it breaking) and then I think this is good to merge

PR #4045, opened 2026-02-03
model: add e5-omni (3B, 7B) omni-modal embedding models

sarendis56:
- Introduced E5OmniWrapper class for handling E5-Omni models.
- Updated pyproject.toml to include E5-Omni as an optional dependency.
- Modified uv.lock to reflect new dependencies.
- Added model metadata for E5-Omni 3B and 7B variants, including citation and training datasets.

If you add a model or a dataset, please add the corresponding checklist:

- [√] I have filled out the ModelMeta object to the extent possible
- [√] I have ensured that my model can be loaded using
  - [√] `mteb.get_model(model_name, revision)` and
  - [√] `mteb.get_model_meta(model_name, revision)`
- [√] I have tested the implementation works on a representative set of tasks.
- [√] The model is public, i.e., is available either as an API or the weights are publicly available to download

Close #4039

PR #4319, opened 2026-03-29
fix: activate flake8-bandit rules

Samoed:
Activate `S` rules. Ref https://github.com/embeddings-benchmark/mteb/issues/2416

### DIFF
--- a/mteb/cache.py
+++ b/mteb/cache.py
@@ -294,7 +294,7 @@ class ResultCache:
 
             if revision:
                 logger.info(f"Checking out revision '{revision}'")
-                subprocess.run(  # noqa: S603
+                subprocess.run(
                     ["git", "checkout", revision],
                     cwd=results_directory,
                     check=True,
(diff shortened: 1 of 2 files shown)


mteb/_evaluators/_download.py
Samoed, line 1: Wasn't used anywhere. Probably was used only in v1 during fetching results for 2 stage retrieval

mteb/cache.py
```diff
@@ -294,7 +294,7 @@ def download_from_remote(
 
             if revision:
                 logger.info(f"Checking out revision '{revision}'")
-                subprocess.run(
+                subprocess.run(  # noqa: S603
```
KennethEnevoldsen: I am not sure we want this -  rule from the docs:

> Prone to false positives as it is difficult to determine whether the passed arguments have been validated ([#4045](https://github.com/astral-sh/ruff/issues/4045)).

Samoed: Should we add it to global ignore?

KennethEnevoldsen: I would do that yes

KennethEnevoldsen:
@Samoed feel free to merge this whenever you are ready for it

PR #4325, opened 2026-03-30
fix: Fixed a bug where benchmarks could not be recorded.

q275343119:
* Fixed the issue where benchmark_change was not logged.
* The logic for generating session_id has been optimized.

### DIFF
--- a/tests/test_leaderboard/test_event_logger.py
+++ b/tests/test_leaderboard/test_event_logger.py
@@ -1,0 +1,132 @@
+"""Unit tests for leaderboard EventLogger (Mongo optional, silent degradation)."""
+
+from __future__ import annotations
+
+import atexit
+import logging
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+import mteb.leaderboard.event_logger.logger as logger_module
+from mteb.leaderboard.event_logger.logger import EventLogger
+
+LOGGER_NAME = "mteb.leaderboard.event_logger.logger"
+
+
+@pytest.fixture(autouse=True)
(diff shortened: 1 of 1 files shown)


mteb/leaderboard/app.py
```diff
@@ -588,57 +538,9 @@ def get_leaderboard_app(
...
-                } catch(e) {
-                    console.warn('MTEB: localStorage unavailable, using fingerprint mode');
-                    data._storage_available = false;
-                    window.__mteb_storage_available__ = false;
-                }
-
-                return [JSON.stringify(data)];
-            }
-            """,
-        )
```
q275343119, line 623: @Samoed During testing, I found that this callback does not always trigger when refreshing the page or reopening it, which seems to be due to caching in webstock.
If this does not trigger, session_id cannot be generated.
So, we ultimately used the built-in method based on the request to generate session_id.

KennethEnevoldsen:
@q275343119 any chance we can add a test on this functionality to make sure we don't accidentally break it with other functionality?

tests/test_leaderboard/test_event_logger.py
q275343119, line 1: Hi @KennethEnevoldsen, I've added unit tests

q275343119:
Hi @Samoed ,do you have time to take a look why this check can't pass?

Samoed:
Seems that network problem, that not related to your PR. I restarted test

q275343119:
> @q275343119 any chance we can add a test on this functionality to make sure we don't accidentally break it with other functionality?

Hi @KennethEnevoldsen ,could you please take a look when you have time?

Issue #4339, opened 2026-04-01
Test minial dependency version


Samoed:
### Description of the feature

We can create test with minimal dependencies version from pyproject and their subdependencies to test for errors like https://github.com/embeddings-benchmark/mteb/issues/4334

PR #4343, opened 2026-04-02
fix: Add support for the lower bound of dependencies

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4339

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,8 +125,8 @@ qwen-vl = ["transformers>=4.57.0", "qwen-vl-utils>=0.0.14"]
 lint = [
     "ruff==0.14.0", # locked so we don't get PRs which fail only due to a lint update
     "pre-commit>=4.1.0",
-    "bibtexparser>=1.4.3", # used for tests/test_citation_formatting.py
     "typos>1",
+    "bibtexparser>=1.4.3", # used for tests/test_citation_formatting.py
 ]
 test = [
     "pytest>=8.3.4,<8.4.0",
@@ -136,6 +136,7 @@ test = [
     "iso639>=0.1.4", # used for tests/scripts/test_generate_model_meta.py
     "GitPython>=3.0.0",
     "tabulate>=0.9.0",
+    "bibtexparser>=1.4.3", # used for tests/test_citation_formatting.py
 ]
 docs = [
     "zensical>=0.0.11",


pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
     "numpy>=1.0.0,<3.0.0",
     "requests>=2.26.0",
     "scikit-learn>=1.4.0",
-    "scipy>=0.0.0",
+    "scipy>=1.9.0",
```
Samoed: One of first versions that support python3.10 natively without additional wheels building

Samoed:
Yeah, I'm debbing them

pyproject.toml
```diff
@@ -32,18 +32,19 @@ classifiers = [
...
     "typing-extensions>=4.5.0",
-    "torch>1.0.0",
+    "torch>=2.0.0",
     "tqdm>1.0.0",
-    "rich>=0.0.0",
+    "rich>=13.0.0",
     "pytrec-eval-terrier>=0.5.6",
-    "pydantic>=2.0.0",
-    "polars>=0.20.22",
+    "pydantic>=2.11.0",
```
KennethEnevoldsen, line 46: this seems very recent - what is the failure mode here?

Samoed: On `2.10.0` on `BenchmarkResults` import
```
pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'mteb.abstasks.abstask.AbsTask'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.
If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.
For further information visit https://errors.pydantic.dev/2.10/u/schema-for-unknown-type
```

Samoed:
> I could see it as a one off that fixes the lack of good lower bounds

We need this to avoid introducing new incompatibilities. For example, older numpy versions do not support some methods, or as is currently the case with `CompressionWrapper`, it is not possible to even import mteb. Also, this test would ensure that mteb runs correctly without additional dependencies

PR #4344, opened 2026-04-02
fix: HF benchmark result

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/issues/4055

Add HF benchmark result. 
```python
import mteb

benchmark = mteb.get_benchmark("BRIGHT(v1.1)")
# benchmark.push_eval_to_hub(create_pr=True)  # push to eval.yaml
model_meta = mteb.get_model_meta("mteb/baseline-bm25s")
model_meta.push_eval_results(
    "mteb",
    tasks=benchmark,
    create_pr=True,
)
```

1. PR with benchmark config https://huggingface.co/datasets/mteb/BRIGHT/discussions/2
2. PR with models results https://huggingface.co/mteb/baseline-bm25s/discussions/6

Current implementation is a bit janky. Probably we can add methods for getting benchmark scores for model and aggregate it somehow. Maybe add `benchmark` property to model results maybe something else, I don't have ideas for now.


mteb/models/model_meta.py
```diff
@@ -1232,7 +1235,7 @@ def push_eval_results(
         self,
         user: str | None = None,
         *,
-        tasks: Sequence[AbsTask] | Sequence[str] | None = None,
+        tasks: Iterable[AbsTask] | Sequence[str] | Benchmark | None = None,
```
KennethEnevoldsen, line 1318: does iterable match with load_results?

mteb/results/model_result.py
```diff
@@ -449,14 +456,44 @@ def from_disk(cls, path: Path) -> ModelResult:
...
+        benchmark_result = None
+        if benchmark is not None:
+            benchmark_score = sum(
+                [task_res.get_score() for task_res in self.task_results]
+            )
+            benchmark_score = (
+                round(benchmark_score / len(self.task_results), 2)
+                if self.task_results
+                else None
+            )
```
KennethEnevoldsen: I don't think it is given that all benchmarks compute the mean score (they might, but then some use AggregateTask)

KennethEnevoldsen: Maybe we should put this as a function on the benchmark object instead?

KennethEnevoldsen (changes requested):
Only problem seems to be the mean assumption, which I think should be possible to change

mteb/results/model_result.py
Samoed: Do you want to `Benchmark` or `BenchmarkResults`?

KennethEnevoldsen: it should be on the Benchmark (it can change but the results can't)

Samoed: added

mteb/benchmarks/benchmark.py
```diff
@@ -205,6 +223,129 @@ def __repr__(self) -> str:
...
+            tasks=[
+                HFEvalTaskConfig(
+                    id=self.name,
+                    config=None,
+                    split=None,
+                )
+            ],
+        )
+
+    def get_scores(
```
KennethEnevoldsen: I don't think I would add it in this way I would rather add the aggregator as a function on the object.

```
agg: dict[str, Callable] = benchmark.get_score_aggregators()
```

If model result then already have the benchmark then, `get_score()` should just default to using the benchmark aggregators.

KennethEnevoldsen: that way it becomes:

```
res = cache.load_results(bench, model)
res = res.model_results[0]
res.get_score()
```

which would look like:

```
{"mean(Task)": 0.21, "mean(TaskType)": 0.23}
```

For the benchmark results we can then do:

```
res = cache.load_results(bench, model)
res.get_score()
```
which adds:
```
{(modelname, revision): {"mean(Task)": 0.21, "mean(TaskType)": 0.23}}
```

Samoed: I'm not sure with callable approach. Probably we can do like:
```python
class AggregationLevel(str, Enum):
    mean_per_task = "mean_per_task"
    men_per_task_type = "men_per_task_type"

def get_score(
    results: ModelResult,
    *,
    aggregation: Sequence[AggregationLevel] | Callable[[ModelResult], dict[str, float]]
) -> dict[AggregationLevel | str, float]:
    ...
```

KennethEnevoldsen: Hmm but what I want my own custom aggregation for a benchmark?

Samoed: You can still pass callable to `get_score`. Otherwise, you can overload `get_score`, for example for RTEB with private/public

KennethEnevoldsen: hmm sorry for being slow on this. I wanted to make sure I took the time to probably look through it.

I don't think there is a need for the `AggregationLevel` as this would essentially just be the defaults. I don't think people would change it. I would typically either use the default or use the custom I would probably never say just do this one of the defaults. 

The callable would also need to be able to handle stuff like borda rank so I would suspect we need it to take benchmarkresults and return a list of (modelname, rev) and their scores?

And I really don't think the aggregation itself should be done on the `Benchmark` object, for instance if you don't give it the right benchmark results it will produce the incorrect score, which should not happen. I would also say it is metadata object (similar to ModelMeta and TaskMetadata), while the `.get_score` belongs on the result objects

Samoed: > And I really don't think the aggregation itself should be done on the Benchmark object, for instance if you don't give it the right benchmark results it will produce the incorrect score, which should not happen

I agree on that and because of this I asked about this in https://github.com/embeddings-benchmark/mteb/pull/4344#discussion_r3033447168

> The callable would also need to be able to handle stuff like borda rank so I would suspect we need it to take benchmarkresults and return a list of (modelname, rev) and their scores?

I don't like approach with callable, because this would require for users to pass some functions all the time, but they can be specified by benchmark object for example. If we want customization, then we can pass callable for custom aggregation, otherwise we can select from one of existing functions

KennethEnevoldsen: > I agree on that and because of this I asked about this in https://github.com/embeddings-benchmark/mteb/pull/4344#discussion_r3033447168

Hah yeah seems like I am inconsistent.

Maybe the ideal way I would want it is to have the aggregation method on the benchmark, the application and `get_score` on the BenchmarkResults. 

Maybe something like:

```
BenchmarkResults.get_score() 
# produces mean(Task), mean(TaskType), Borda Rank, Borda Count, potentially also per task type (e.g. Classification)
# and also whatever the Benchmark itself defines as a custom function (e.g. mean(Private), mean(Public))
```

We can maybe start out with just implementing the default ones on the BenchmarkResult (but we should in some way make sure that we use the same implementation everywhere to the extent possible)

Samoed:
I updated methods for scores generation. Also created a method for benchmark card creation. Example result: https://huggingface.co/datasets/mteb/MMTEB-Multilingual-v2

KennethEnevoldsen (approved):
I think we can merge this - thanks for taking the time

KennethEnevoldsen:
debated a bit on the fix/feat. I think fix it appropriate, but I could go either way. If we do a feat we need to update docs.

PR #4346, opened 2026-04-03
dataset: Add INatSounds

isaac-chung:
## New Dataset: INatSounds

Adds the iNaturalist Sounds dataset for audio-based species classification.

### Dataset details
- **HuggingFace**: https://huggingface.co/datasets/mteb/inat_sounds
- **Paper**: https://arxiv.org/abs/2506.00343
- **Task type**: AudioClassification
- **Modality**: Audio
- **License**: CC-BY-4.0

The dataset contains recordings of 5,500+ species across birds, insects, amphibians, mammals, and reptiles, contributed by 27,000+ citizen scientists. The test split contains 49,527 recordings from 2023 observations.

### Checklist
- [x] Explain how this dataset fills a gap in MTEB
  - First large-scale bioacoustics species classification benchmark in MTEB
- [ ] Verify the dataset runs with the mteb package
- [ ] Test with `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
- [ ] Test with `intfloat/multilingual-e5-small`
- [ ] Confirm performance isn't trivial or random
- [x] Complete all `TaskMetadata` fields
- [x] Place task in `mteb/tasks` directory
- [x] Import in `__init__.py`

### DIFF
--- a/mteb/tasks/classification/zxx/inat_sounds.py
+++ b/mteb/tasks/classification/zxx/inat_sounds.py
@@ -1,0 +1,41 @@
+from mteb.abstasks.classification import AbsTaskClassification
+from mteb.abstasks.task_metadata import TaskMetadata
+
+
+class INatSoundsClassification(AbsTaskClassification):
+    metadata = TaskMetadata(
+        name="INatSounds",
+        description="iNaturalist Sounds dataset for species identification from audio. Contains recordings of 5,500+ species across birds, insects, amphibians, mammals, and reptiles, contributed by 27,000+ citizen scientists. Test split contains 49,527 recordings from 2023 observations.",
+        reference="https://github.com/visipedia/inat_sounds",
+        dataset={
+            "path": "mteb/inat_sounds",
+            "revision": "6db625bf80d71ded64690022f0df56ceae972b6e",
+        },
+        type="AudioClassification",
+        category="a2c",
+        eval_splits=["test"],
+        eval_langs=["zxx-Zxxx"],
+        main_score="accuracy",
+        date=("2024-01-01", "2025-06-01"),
+        domains=["Spoken", "Bioacoustics"],
+        task_subtypes=["Species Classification"],
+        license="cc-by-4.0",
+        annotations_creators="human-annotated",
+        dialect=[],
+        modalities=["audio"],
+        sample_creation="found",
+        bibtex_citation=r"""
+@article{chasmai2025inatsounds,
+  archiveprefix = {arXiv},
+  author = {Mustafa Chasmai and Alexander Shepard and Subhransu Maji and Grant Van Horn},
+  eprint = {2506.00343},
+  primaryclass = {cs.SD},
+  title = {The iNaturalist Sounds Dataset},
+  url = {https://arxiv.org/abs/2506.00343},
+  year = {2025},
+}
+""",
+    )
+
+    input_column_name: str = "audio"
+    label_column_name: str = "label"
--- a/mteb/tasks/classification/zxx/inat_sounds.py
+++ b/mteb/tasks/classification/zxx/inat_sounds.py
@@ -41,4 +41,5 @@ class INatSoundsClassification(AbsTaskClassification):
     label_column_name: str = "label"
     n_experiments = 1
     is_cross_validation: bool = True
+    train_split = "test"
     n_splits = 3
--- a/mteb/tasks/classification/zxx/inat_sounds.py
+++ b/mteb/tasks/classification/zxx/inat_sounds.py
@@ -39,3 +39,6 @@ class INatSoundsClassification(AbsTaskClassification):
 
     input_column_name: str = "audio"
     label_column_name: str = "label"
+    n_experiments = 1
+    is_cross_validation: bool = True
+    n_splits = 3
--- a/scripts/upload_inatsounds.py
+++ b/scripts/upload_inatsounds.py
@@ -1,0 +1,298 @@
+#!/usr/bin/env python3
+"""
+Script to download iNatSounds 2024 test set and upload to HuggingFace Hub.
+
+Usage:
+    python scripts/upload_inatsounds.py --token YOUR_HF_TOKEN
+
+    Or set HF_TOKEN environment variable:
+    export HF_TOKEN=YOUR_HF_TOKEN
+    python scripts/upload_inatsounds.py
+
+Requirements:
+    pip install datasets huggingface_hub tqdm requests
+"""
+
+import argparse
+import hashlib
+import json
+import os
+import shutil
+import tarfile
+import tempfile
+from pathlib import Path
+from typing import Dict, List
+
+import requests
+from datasets import Audio, Dataset, DatasetDict, Features, Value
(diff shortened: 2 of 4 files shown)


scripts/upload_inatsounds.py
KennethEnevoldsen, line 1: I don't believe we need this - I would rather put it in the repo with the dataset

mteb/tasks/classification/zxx/inat_sounds.py
KennethEnevoldsen, line 1: Metadata seems very reasonable

KennethEnevoldsen (commented):
would be great to run at least one model on it to make sure that everything works as intended, but otherwise it looks good

It also seems like we are missing the descriptive stats

isaac-chung:
I ran mini on Yamnet and for like 0.01 as main score - maybe species level is quite hard for it. Wonder how other models perform on this.

KennethEnevoldsen:
hmm, yeah that seems very low - we talked about also implementing a domain specific model - might be better to test with one of those

isaac-chung:
Yeah might look into one of these: OpenBEATs - a shikhar7ssu Collection https://share.google/nJLukTgjGY8mWEgZD

scripts/upload_inatsounds.py
isaac-chung: The challenge with this set is the train blob is 100+GB and I struggle to download, unzip, then sample it on my machine.

isaac-chung: We need some training data for classification

mteb/descriptive_stats/AudioClassification/INatSoundsMini.json
```diff
@@ -0,0 +1,3663 @@
...
+                    "count": 1
+                },
+                "Eopsaltria australis": {
+                    "count": 2
+                },
+                "Phacellodomus ruber": {
+                    "count": 1
+                },
+                "Aramus guarauna": {
+                    "count": 1
```
KennethEnevoldsen: We only have one label pr class? I think this is due to the downsample - I don't think that makes a lot of sense there

isaac-chung: Likely due to down sample yeah. Full set is 49k rows, and needs 30+GB to load

KennethEnevoldsen: We could still make it smaller, but probably good to keep it large enough to be meaningful. 

How about:
a stratified subsample of the test set from 49,527 to 2 of each unique label (so about 10k) and then ensure that we have 10 labels per unique label so around 50-60k?

I believe currently it would sample the entire dataset and reembed it every time during evaluation. This could be optimized (e.g. by sampling all the IDs, then encoding then encoding only the required documents once), but it is a problem in the task not the dataset. In this case to improve runtime we could reduce the number of experiments to 1. I suspect with the current setting all 10 experiment are essentially the same anyway.

isaac-chung: Also the current label is at the most granular species level. Maybe we want to explore something higher level? 

```
  ┌───────────────┬──────────┬────────────────┐
  │     Level     │ Distinct │     Range      │
  ├───────────────┼──────────┼────────────────┤
  │ supercategory │ 5        │ too few        │
  ├───────────────┼──────────┼────────────────┤
  │ order         │ 62       │ good candidate? │
  ├───────────────┼──────────┼────────────────┤
  │ family        │ 356      │ good candidate? │
  ├───────────────┼──────────┼────────────────┤
  │ genus         │ 2,075    │ over 2k        │
  ├───────────────┼──────────┼────────────────┤
  │ species       │ 5,569    │ over 2k        │
  └───────────────┴──────────┴────────────────┘
```

KennethEnevoldsen: > This could be optimized (e.g. by sampling all the IDs, then encoding then encoding only the required documents once)

I am actually not entirely sure on this part seems that there is a cache it place that re-uses embeddings. So I don't think there is an issue. On that point, but still believe that the due to the large number of unique labels we end up sampling the whole dataset (we can check this if you have the results file)

KennethEnevoldsen: > Also the current label is at the most granular species level. Maybe we want to explore something higher level?

Ahh yeah that could be a way out. Feels like hiearchical classification. I would probably suggest family, genus, species but let me ask a friend who is a biologist about what makes sense here.

I am probably leaning toward species still though

KennethEnevoldsen: talked with biology friend, she argued that it makes no sense to predict up the their - species or maybe genus.

She however suggest downsampling the number of species to e.g. 2k - I agree that this is probably the best approach. We could also split it into 5 datasets one for each subcategory (I don't know what a subcategory is though)

isaac-chung: Makes sense to down sample species. Thanks for asking!

KennethEnevoldsen: No problem - I also heard that a large majority of the dataset is birds (I think ~4k)

isaac-chung:
Going to stick with species and run the random encoder

isaac-chung:
random encoder:

```
{
  "dataset_revision": "6db625bf80d71ded64690022f0df56ceae972b6e",
  "task_name": "INatSounds",
  "mteb_version": "2.12.21",
  "scores": {
    "test": [
      {
        "scores_per_experiment": [
          {
            "accuracy": 0.000848,
            "f1": 0.000499,
            "f1_weighted": 0.000586,
            "precision": 0.000402,
            "precision_weighted": 0.00063,
            "recall": 0.001208,
            "recall_weighted": 0.000848,
            "ap": null,
            "ap_weighted": null
          },
          {
            "accuracy": 0.000969,
            "f1": 0.00077,
            "f1_weighted": 0.000858,
            "precision": 0.001076,
            "precision_weighted": 0.001555,
            "recall": 0.001029,
            "recall_weighted": 0.000969,
            "ap": null,
            "ap_weighted": null
          },
          {
            "accuracy": 0.000848,
            "f1": 0.000567,
            "f1_weighted": 0.001033,
            "precision": 0.001157,
            "precision_weighted": 0.002313,
            "recall": 0.000512,
            "recall_weighted": 0.000848,
            "ap": null,
            "ap_weighted": null
          }
        ],
        "accuracy": 0.000888,
        "f1": 0.000612,
        "f1_weighted": 0.000825,
        "precision": 0.000878,
        "precision_weighted": 0.001499,
        "recall": 0.000917,
        "recall_weighted": 0.000888,
        "ap": NaN,
        "ap_weighted": NaN,
        "main_score": 0.000888,
        "hf_subset": "default",
        "languages": [
          "zxx-Zxxx"
        ]
      }
    ]
  },
  "evaluation_time": 1415.2721869945526,
  "kg_co2_emissions": null,
  "date": 1777633656.245199
}
```

PR #4347, opened 2026-04-04
model: add Octen INT8 models and update HuggingFace org to Octen

bflhc:

### DIFF
--- a/mteb/models/model_implementations/octen_models.py
+++ b/mteb/models/model_implementations/octen_models.py
@@ -206,7 +206,7 @@ Octen_Embedding_0B6 = ModelMeta(
     public_training_data=None,
     training_datasets=training_data,
     citation=OCTEN_CITATION,
-    output_dtypes=[OutputDType.BF16],
+    output_dtypes=OutputDType.BF16,
     adapted_from="Qwen/Qwen3-Embedding-0.6B",
 )
 
@@ -238,7 +238,7 @@ Octen_Embedding_4B = ModelMeta(
     public_training_data=None,
     training_datasets=training_data,
     citation=OCTEN_CITATION,
-    output_dtypes=[OutputDType.BF16],
+    output_dtypes=OutputDType.BF16,
     adapted_from="Qwen/Qwen3-Embedding-4B",
 )
 
@@ -270,7 +270,7 @@ Octen_Embedding_8B = ModelMeta(
     public_training_data=None,
     training_datasets=training_data,
     citation=OCTEN_CITATION,
-    output_dtypes=[OutputDType.BF16],
+    output_dtypes=OutputDType.BF16,
     adapted_from="Qwen/Qwen3-Embedding-8B",
 )
 
@@ -302,7 +302,7 @@ Octen_Embedding_4B_INT8_dim_1024 = ModelMeta(
     public_training_data=None,
     training_datasets=training_data,
     citation=OCTEN_CITATION,
-    output_dtypes=[OutputDType.FLOAT16],
+    output_dtypes=OutputDType.FLOAT16,
     adapted_from="Qwen/Qwen3-Embedding-4B",
 )
 
@@ -334,6 +334,6 @@ Octen_Embedding_8B_INT8_dim_1024 = ModelMeta(
     public_training_data=None,
     training_datasets=training_data,
     citation=OCTEN_CITATION,
-    output_dtypes=[OutputDType.FLOAT16],
+    output_dtypes=OutputDType.FLOAT16,
     adapted_from="Qwen/Qwen3-Embedding-8B",
 )


mteb/models/model_implementations/octen_models.py
```diff
@@ -172,7 +172,7 @@ def instruction_template(
         max_seq_length=18480,
         model_kwargs={"torch_dtype": "bfloat16"},
     ),
-    name="bflhc/Octen-Embedding-0.6B",
+    name="Octen/Octen-Embedding-0.6B",
```
Samoed, line 190: Can you rename also in results repo?

mteb/models/model_implementations/octen_models.py
```diff
@@ -245,7 +245,71 @@ def instruction_template(
...
+
+Octen_Embedding_4B_INT8_dim_1024 = ModelMeta(
+    loader=InstructSentenceTransformerModel,
+    loader_kwargs=dict(
+        instruction_template=instruction_template,
+        apply_instruction_to_passages=True,
+        prompts_dict=_PREDEFINED_PROMPTS,
+        max_seq_length=18480,
+        embed_dim=1024,
+        model_kwargs={"torch_dtype": "bfloat16"},
```
Samoed: Is this correct dtype?

mteb/models/model_implementations/octen_models.py
```diff
@@ -245,7 +245,71 @@ def instruction_template(
...
+    license="apache-2.0",
+    reference="https://huggingface.co/Octen/Octen-Embedding-4B-INT8",
+    similarity_fn_name="cosine",
+    framework=["Sentence Transformers", "PyTorch", "safetensors"],
+    use_instructions=True,
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=training_data,
+    citation=OCTEN_CITATION,
+    adapted_from="Qwen/Qwen3-Embedding-4B",
```
Samoed, line 306: Can you add `output_dtypes`?

mteb/models/model_implementations/octen_models.py
bflhc: of course

mteb/models/model_implementations/octen_models.py
bflhc: Yes, the results I submitted are at 1024 dimensions — the goal is to showcase the performance of the quantized model compressed to 1024 dimensions as a reference for the community. However, I'm not sure how to include the 1024-dimension specification in the naming when submitting results.

mteb/models/model_implementations/octen_models.py
bflhc: No need to add output_dtypes — that only applies to the model's quantization parameters. The output embeddings are still standard float type.

mteb/models/model_implementations/octen_models.py
Samoed: I don't have questions to dimensions, but in name int8, but dtype is bfloat16

mteb/models/model_implementations/octen_models.py
Samoed: It's not for quantizations only, but generally for all models

mteb/models/model_implementations/octen_models.py
bflhc: The INT8 in the model name refers to the model parameters, not the output_dtypes. This is documented on the model's main page:
Quantization: This is an INT8 quantized version using bitsandbytes. INT8 quantization significantly reduces memory footprint (~50% smaller), making it suitable for deployment on resource-constrained environments. Note that while memory usage is reduced, inference speed may not necessarily improve and could be slightly slower than the BF16 version on some hardware.

Samoed: If this is a quantized model in int8, then why are you downloading it in bfloat16?

bflhc: Thanks for the feedback! After investigation, we agree that removing torch_dtype: bfloat16 from the INT8 models is the right approach. The original reason we kept it was that loading the 8B-INT8 model without model_kwargs caused incorrect document prefix behavior — the model page recommends using "- " as the document prefix due to an upstream Qwen3-Embedding issue, while the existing instruction_template uses " ". We've now addressed this properly by adding a dedicated instruction_template_8b_int8 for the 8B-INT8 model.                                                                                                                                                                                       

We've also verified on several datasets that the results are correct after these changes. The full evaluation is currently re-running, and we'll submit an updated results PR once it's complete.

If everything looks good on your end, feel free to merge this PR first.

mteb/models/model_implementations/octen_models.py
```diff
@@ -183,14 +198,15 @@ def instruction_template(
...
-    reference="https://huggingface.co/bflhc/Octen-Embedding-0.6B",
+    reference="https://huggingface.co/Octen/Octen-Embedding-0.6B",
     similarity_fn_name="cosine",
     framework=["Sentence Transformers", "PyTorch", "safetensors"],
     use_instructions=True,
     public_training_code=None,
     public_training_data=None,
     training_datasets=training_data,
     citation=OCTEN_CITATION,
+    output_dtypes=[OutputDType.BF16],
```
Samoed:
```suggestion
    output_dtypes=OutputDType.BF16,
```

bflhc: ok

PR #4351, opened 2026-04-05
fix: Change main score from accuracy to LRAP

KennethEnevoldsen:
For MAEB MultiLabelClassification tasks.

I suspect we also need to change the submitted results to match this change


AdnanElAssadi:
Can we merge this?

Samoed:
We need to change in results folder also

AdnanElAssadi:
>We need to change in results folder also

Is there an easy way to do it?

KennethEnevoldsen:
Hmm I think write a script is the easiest

KennethEnevoldsen:
added the fix here: https://github.com/embeddings-benchmark/results/pull/478

AdnanElAssadi:
I also already ran a similar script and updated
https://github.com/embeddings-benchmark/maeb-results

isaac-chung:
We use the main results repo now

PR #4352, opened 2026-04-05
Add model: BEATs pretrained audio models

isaac-chung:
## Summary
- Adds BEATs (Audio Pre-Training with Acoustic Tokenizers) pretrained model implementations
- Three pretrained checkpoints: BEATs_iter1, BEATs_iter2, BEATs_iter3
- Auto-downloads BEATs source files from the [unilm repo](https://github.com/microsoft/unilm/tree/master/beats) since there is no pip package
- Tested BEATs_iter2: **75% accuracy on ESC50**, also ran on AudioSetMini 0.38 mAP ([matches the paper](https://arxiv.org/pdf/2212.09058))

## Model checklist
- [x] I have ensured that my model can be loaded using:
  - [x] `mteb.get_model(model_name, revision)`
- [x] I have tested the implementation works on a representative set of tasks

## Notes
- Checkpoints are hosted on OneDrive (not HuggingFace), so `model_name` must point to a local `.pt` file
- BEATs source code (MIT license) is fetched at runtime to `~/.cache/torch/hub/beats_src/`

### DIFF
--- a/mteb/models/model_implementations/beats_models.py
+++ b/mteb/models/model_implementations/beats_models.py
@@ -193,18 +193,18 @@ _common_meta = dict(
 
 beats_iter1 = ModelMeta(
     loader=BEATsWrapper,
-    name="microsoft/beats-iter1",
+    name="microsoft/beats-iter1-pretrained",
     **_common_meta,
 )
 
 beats_iter2 = ModelMeta(
     loader=BEATsWrapper,
-    name="microsoft/beats-iter2",
+    name="microsoft/beats-iter2-pretrained",
     **_common_meta,
 )
 
 beats_iter3 = ModelMeta(
     loader=BEATsWrapper,
-    name="microsoft/beats-iter3",
+    name="microsoft/beats-iter3-pretrained",
     **_common_meta,
 )
--- a/mteb/models/model_implementations/beats_models.py
+++ b/mteb/models/model_implementations/beats_models.py
@@ -1,0 +1,210 @@
+"""BEATs: Audio Pre-Training with Acoustic Tokenizers.
+
+Reference: https://github.com/microsoft/unilm/tree/master/beats
+
+Setup:
+    BEATs is not available as a pip package. The source files (BEATs.py,
+    backbone.py, modules.py) are automatically downloaded from the unilm
+    repository to ``~/.cache/torch/hub/beats_src/`` on first use.
+
+    Pretrained checkpoint weights must be downloaded manually from OneDrive
+    and passed as ``model_name`` when constructing the wrapper:
+
+        - BEATs_iter1: https://1drv.ms/u/s!AqeByhGUtINrgcpmY7IHhgc9q0pT7Q
+        - BEATs_iter2: https://1drv.ms/u/s!AqeByhGUtINrgcpwwEGgUyiI-jQyQw
+        - BEATs_iter3: https://1drv.ms/u/s!AqeByhGUtINrgcpxJUNDxg4eU0r-vA
+
+    Example::
+
+        from mteb.models.model_implementations.beats_models import beats_iter2
+
+        beats_iter2.name = "/path/to/BEATs_iter2.pt"
+        results = mteb.evaluate(beats_iter2, tasks)
+"""
+
+from __future__ import annotations
+
+import importlib
+import logging
+import sys
+import urllib.request
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import torch
+from tqdm.auto import tqdm
+
+from mteb._create_dataloaders import AudioCollator
+from mteb._requires_package import requires_audio_dependencies
+from mteb.models import ModelMeta
+from mteb.models.abs_encoder import AbsEncoder
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+    from mteb.types._encoder_io import AudioInput
+
+logger = logging.getLogger(__name__)
+
+_BEATS_SRC_URL = "https://raw.githubusercontent.com/microsoft/unilm/master/beats"
+_BEATS_SRC_FILES = ["BEATs.py", "backbone.py", "modules.py"]
+
+_BEATS_CITATION = """
+@article{Chen2022beats,
+  title = {BEATs: Audio Pre-Training with Acoustic Tokenizers},
+  author = {Sanyuan Chen and Yu Wu and Chengyi Wang and Shujie Liu and Daniel Tompkins and Zhuo Chen and Furu Wei},
+  eprint = {2212.09058},
+  archiveprefix = {arXiv},
+  year = {2022},
+}
+"""
+
+
+def _load_beats_modules():
+    """Download BEATs source files if needed and import the module."""
+    try:
+        from BEATs import BEATs, BEATsConfig
+
+        return BEATs, BEATsConfig
+    except ImportError:
+        pass
+
+    cache_dir = Path(torch.hub.get_dir()) / "beats_src"
+    cache_dir.mkdir(parents=True, exist_ok=True)
+
+    for fname in _BEATS_SRC_FILES:
+        dest = cache_dir / fname
+        if not dest.exists():
+            logger.info("Downloading %s to %s", fname, dest)
+            url = f"{_BEATS_SRC_URL}/{fname}"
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/beats_models.py
```diff
@@ -0,0 +1,210 @@
...
+  title = {BEATs: Audio Pre-Training with Acoustic Tokenizers},
+  author = {Sanyuan Chen and Yu Wu and Chengyi Wang and Shujie Liu and Daniel Tompkins and Zhuo Chen and Furu Wei},
+  eprint = {2212.09058},
+  archiveprefix = {arXiv},
+  year = {2022},
+}
+"""
+
+
+def _load_beats_modules():
```
KennethEnevoldsen, line 65: Hmm don't we want to put it on huggingface instead?

mteb/models/model_implementations/beats_models.py
```diff
@@ -0,0 +1,210 @@
...
+        **kwargs: Any,
+    ):
+        requires_audio_dependencies()
+        beats_cls, beats_config_cls = _load_beats_modules()
+
+        self.model_name = model_name
+        self.device = device
+        self.sampling_rate = 16_000
+
+        checkpoint = torch.load(model_name, map_location="cpu", weights_only=False)
```
KennethEnevoldsen: why not load it directly onto device?

mteb/models/model_implementations/beats_models.py
KennethEnevoldsen: I think something like this could work:

```
from huggingface_hub import hf_hub_download
import torch

with open(hf_hub_download("username/repo", "model.pt"), "rb") as f:
    obj = torch.load(f, map_location="cpu", weights_only=True)
```

mteb/models/model_implementations/beats_models.py
```diff
@@ -0,0 +1,210 @@
...
+    from torch.utils.data import DataLoader
+
+    from mteb import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+    from mteb.types._encoder_io import AudioInput
+
+logger = logging.getLogger(__name__)
+
+_BEATS_SRC_URL = "https://raw.githubusercontent.com/microsoft/unilm/master/beats"
+_BEATS_SRC_FILES = ["BEATs.py", "backbone.py", "modules.py"]
```
KennethEnevoldsen, line 52: I would probably rather copy this in - (this could break very easily with change on their end)

Either than or put it together with the weight on hf

KennethEnevoldsen (commented):
I would really love to get this on HF, even if it is not supported that way everything we have is on HF.

isaac-chung:
So we didn't upload models like EvaCLIP to HF for the same reason - we don't maintain them. Could that be a caveat? That we can upload models but don't maintain them? Or we want to maintain them as well?

mteb/models/model_implementations/beats_models.py
```diff
@@ -0,0 +1,210 @@
...
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+    from mteb.types._encoder_io import AudioInput
+
+logger = logging.getLogger(__name__)
+
+_BEATS_SRC_URL = "https://raw.githubusercontent.com/microsoft/unilm/master/beats"
```
Samoed: If we won't reupload we can specify revision
```suggestion
_BEATS_SRC_URL = "https://raw.githubusercontent.com/microsoft/unilm/732d834db70ee0fc3886b4bcbcfb4ce7fb829be2/beats"
```

mteb/models/model_implementations/beats_models.py
```diff
@@ -0,0 +1,210 @@
...
+    public_training_code="https://github.com/microsoft/unilm/tree/master/beats",
+    public_training_data="https://research.google.com/audioset/dataset/index.html",
+    training_datasets={"AudioSetMini"},
+    modalities=["audio"],
+    citation=_BEATS_CITATION,
+)
+
+beats_iter1 = ModelMeta(
+    loader=BEATsWrapper,
+    name="microsoft/beats-iter1",
```
Samoed: Is this pretrained or finetuned model? They have a lot of models and this is not clear which it is

Image

isaac-chung: I added  the "Pre-trained Model" column

mteb/models/model_implementations/beats_models.py
isaac-chung: added to model meta as well

mteb/models/model_implementations/beats_models.py
isaac-chung: Will do

mteb/models/model_implementations/beats_models.py
KennethEnevoldsen: Are they then fine-utned on audioset?

Should we make sure they are eq. to the ones in the paper?

Should we rename to:
`microsoft/beats-iter1-pretrained`

mteb/models/model_implementations/beats_models.py
KennethEnevoldsen: I think this might be enough. We still can't version the fine on OneDrive though

mteb/models/model_implementations/beats_models.py
isaac-chung: > Are they then fine-utned on audioset?

Yeah, I've added it as a training dataset in ModelMeta.

> Should we make sure they are eq. to the ones in the paper?

we get the same scores as in the paper for AudioSetMini, but different for ESC50, maybe because we set it up differently? (cross validation)

isaac-chung: Renamed models.

KennethEnevoldsen: > Yeah, I've added it as a training dataset in ModelMeta.

But it sounds like it is not fine-tuned (yet) if we use the pre-trained

KennethEnevoldsen: > we get the same scores as in the paper for AudioSetMini, but different for ESC50, maybe because we set it up differently? (cross validation)

How big is the difference (can I ask you to add a table in the commit message so we have it for later)

isaac-chung: for ESC50, this PR gets 0.75, whereas the paper shows 0.951 in [table 1](https://arxiv.org/pdf/2212.09058)

KennethEnevoldsen: Hmm out classification tasks are different due sampling - do we have some of the other models in their table we can compare with to see if the ordering is approx. the same?

isaac-chung: There's SpeechCommandsv2 too - we have the mini version.

KennethEnevoldsen: Hmm looked to compare a few of the implementation I'm not sure exactly what to draw from these, but beats seems to be doing surprisingly poorly.

| mdl | ours | theirs |
| --- | --- | --- |
| beats | 0.75 | 0.951 |
| microsoft/msclap-2023 | 0.974 |    |
| microsoft/msclap-2022 | 0.9095 |   |
| CLAP (not sure which one) |  | 0.967  |
| MIT/ast-finetuned-audioset-10-10-0.4593 / AST | 0.962 | 0.887 | 
| lyrebird/wav2clip / Wav2CLIP | 0.721 | 0.860 | 

We do seem to be maintaining relative ordering if CLAP == microsoft/msclap-2023.

isaac-chung: Which dataset/task are these for?

KennethEnevoldsen: ESC50

KennethEnevoldsen:
@isaac-chung do we need to figure out what to do here?

isaac-chung:
Re: model hosting, I guess we could a) copy the needed files, or b) host it in the mteb org on HF.

Re: correctness/repro scores from the paper, we have the mini/sampled versions of the datasets they used in the paper. We can see if their repo has more scores. Otherwise I'm not sure what else we can try.

KennethEnevoldsen:
If we could move it all to a versioned repo, then I think that would be fine (e.g. hf)

KennethEnevoldsen:
@isaac-chung this seems to have been getting stale - do we want to finish it up?

isaac-chung:
Sure. But we might want to focus on the mveb effort first?

Issue #4350, opened 2026-04-05
Add `required_dependencies` to `ModelMeta`


KennethEnevoldsen:
### Description of the feature

Something like: `ModelMeta.required_dependencies` would allow one to see the required dependencies without having to run the model. It would also simplify required warnings

Samoed:
We can make like
```python
class ModelMeta:
    ...
    required_dependencies: Sequence[str] | Callable[None, None],
```
But we need to specify name somehow. Callable will be used, when some version checking is required

KennethEnevoldsen:
Hmm but couldn't we simply do it as a string and if we get an import error while loading the model then we raise raise the required dependencies?

So it in the model code it looks like:
```
if transformers.__version__ != "1.2.3" # dummy check as an exampel
  raise ImportError() # we could also make an IncorrectVersionError which might be more user friendly
```

And in the load_model code we have:

```
try:
  load_model() # again dummy to illustrate
except ImportError:
  print(f"Got an import error when loading the model please install required dependencies")  
```

PR #4356, opened 2026-04-06
fix: Add required_dependencies to model meta

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4350.

E.g. 
```python
llama_nemoretriever_colembed_1b_v1 = ModelMeta(
    ...
    extra_requirements_groups=["llama-nemotron-colembed-vl"],
)
```

### DIFF
--- a/tests/test_models/test_model_meta.py
+++ b/tests/test_models/test_model_meta.py
@@ -467,9 +467,11 @@ def test_model_meta_dependencies_not_installed_group():
     )
     with pytest.raises(
         ImportError,
-        match=re.escape(
-            'Model google/vggish is missing required dependencies: torch-vggish-yamnet==0.2.1; extra == "torch-vggish-yamnet".'
-            "\nYou can install it with `pip install mteb[torch-vggish-yamnet,audio]`."
+        match=(
+            'Model google/vggish is missing required dependencies:( torchaudio; extra == "audio",)? '
+            + re.escape(
+                'torch-vggish-yamnet==0.2.1; extra == "torch-vggish-yamnet".\nYou can install it with `pip install mteb[torch-vggish-yamnet,audio]`.'
+            )
         ),
     ):
         model_meta._check_requirements()


mteb/models/model_meta.py
```diff
@@ -139,6 +142,8 @@ class ModelMeta(BaseModel):  # noqa: PLR0904
         contacts: The people to contact in case of a problem in the model, preferably a GitHub handle.
         experiment_kwargs: A dictionary of parameters used in the experiment that are not covered by other fields. This is used to create experiment names for ablation studies and similar experiments.
         output_dtypes: Output embedding data types (e.g. int8, binary, float) natively supported by the model. If None, it is assumed that the model only returns float embeddings.
+        required_dependencies: Dependencies that required for model loading.
+        extra_requirements_group_name: Name of group of extra requirements.
```
KennethEnevoldsen: Hmm but now we have two places to manage the group name -> required dependencies. Could we fetch it from pyproject.toml? I am not sure if that is included in the package installation (I suspect it isn't).

(I suppose we remove it from the model so we get one less)

tests/test_models/test_model_meta.py
```diff
@@ -432,3 +434,26 @@ def test_get_model_metas_without_modality_filter_returns_more_models():
...
+def test_model_meta_dependencies_fail():
+    model_meta = mteb.get_model_meta("mteb/baseline-random-encoder").model_copy(
+        update={
+            "required_dependencies": ["test", "sentence-transformers<=1.0.0"],
+            "extra_requirements_group_name": "test_group",
+        }
+    )
+    with pytest.raises(
+        ImportError,
+        match="Missing required dependencies: test, sentence-transformers<=1.0.0\nYou can install it with `pip install test_group`.",
```
```suggestion
        match="Model 'mteb/baseline-random-encoder' is missing required dependencies: ['test', 'sentence-transformers<=1.0.0']\nYou can install it with `pip install 'mteb[test_group]'`.",
```

mteb/models/model_implementations/google_models.py
KennethEnevoldsen, line 1: We are still missing more models - we should be able to resolve any place that imports `requires_package` (by which point we can probably mark it as deprecated - I know it is hidden but people might still rely on it)

mteb/models/model_meta.py
KennethEnevoldsen: Looked into it and you can get it from the toml file using:

```py
from importlib.metadata import requires

package_name = "mteb"
extra = "nemotron-colembed-vl-v2"
specs = requires(package_name)

# Optional dependencies look like: "tiktoken>=0.8.0; extra == "ark""
optional_deps = [r.split(";")[0] for r in specs if f"extra == {extra}" in r]
```
Though we probably also need to handle:

```
colpali_engine>=0.3.12; python_full_version < "3.14" and extra == "colpali-engine"
```

Which actually improves the validation a bit.

mteb/models/model_implementations/tarka_models.py
```diff
@@ -340,6 +340,7 @@
     similarity_fn_name="cosine",
     memory_usage_mb=576,
     citation=Tarka_Embedding_150M_V1_CITATION,
+    required_dependencies=("transformers>=4.56.0",),
```
KennethEnevoldsen: We have a few more groups, which is just transformers:

```
mctct = ["transformers<5"]
```

Unsure if we should keep these specified in the pyproject.toml (removing it would make dependency management simpler, but we would no longer check for compatibility)

mteb/models/model_meta.py
Samoed: So we would use only `extra_requirements_group_name` without `required_dependencies`?

KennethEnevoldsen: Yeah, but that of course required an `extra_requirements_group_name` for everything (most models have it already)

Samoed: Updated to check based on mteb optional dependency group

Samoed:
I added check if modality specific dependencies are installed, but this creates cases like now `mteb/baseline-random-encoder` is not possible to run without image or audio. Maybe solution would be to keep `reqire_audio_dependencies`/`require_image_dependencies` functions

KennethEnevoldsen:
yes let us keep those two for now - at least for models which does not always require them

Samoed:
Because we use `requires_audio_dependencies`/`requires_image_dependencies` during init, I decided to make an exception for the modality check for our baseline models and keep the check for normal models as is during loading. I don't quite like this solution, because it adds an exception for models, but I think this is a better solution.

KennethEnevoldsen (approved):
I think this looks reasonable - though we still have to make sure that tests passes

Samoed:
Yes, I forgot to update test after this update

Issue #4204, opened 2026-03-06
Can't reproduce the scores of nomic-ai/nomic-embed-multimodal


KennethEnevoldsen:
This includes both:
- nomic-ai/nomic-embed-multimodal-7b
- nomic-ai/nomic-embed-multimodal-3b

This is further discussed in the PR #4186

Current implementation follows the doc as closely as possible, and we have performed an experiment to ensure the best performance.

QuentinJGMace:
@KennethEnevoldsen @Samoed I can give a few infos here as they are probably several technical details :

- On the `query_augmentation_token * 10` suffix. I believe it was used by the nomic team, as they derived the models from our implementation of colqwen2.5, and we use it in the code when you call `processor.process_queries` (c.f [here](https://github.com/illuin-tech/colpali/blob/3a562fc0d78acec847f067c832ad875fcdf51d32/colpali_engine/models/qwen2_5/colqwen2_5/processing_colqwen2_5.py#L25)). They could have modified this but I can't know for sure.
This is mostly an artefact from ColBert I believe.

- There are also some issues for all models since ViDoRe V2 is quite small so there is a lot of variance. For most models some hyperparameters can impact embeddings quite a lot when they shouldn't (e.g batch_size), this directly comes from their VLM backbone implementation on HF and as far as I know no one could figure the exact origin yet.

Samoed:
> On the query_augmentation_token * 10 suffix. I believe it was used by the nomic team, as they derived the models from our implementation of colqwen2.5, and we use it in the code when you call processor.process_queries

Does this required in other colpali models? I think this can drop performance of the models

QuentinJGMace:
Yes I think it should be used (I'm not the one who trained them but I'm fairly sure)

KennethEnevoldsen:
Hmm removing it def. improved performance, however I am unsure if there is an interaction with something else.

The 3b model we can get quite close to the reported performance (within 1-2 pp), the 7b is even worse than the 3b.

QuentinJGMace:
Just looked at [#2905](https://github.com/embeddings-benchmark/mteb/pull/2905) and the code you wrote for nomic in #4168 a bit more in detail.
I think they actually both do the exact same thing : `process_texts(text)` does not add query_augmentation_tokens whereas `process_queries(text)` does, so in the end, both codes I cited are equivalent.

[...]

PR #4361, opened 2026-04-07
fix: `colpali_engine` models query processing

### LINKED ISSUE #4360: Incorrect query processing for `colpali_engine` models
https://github.com/embeddings-benchmark/mteb/blob/426a445da812dcc6012b6cebc9fbaf01c975c228/mteb/models/model_implementations/colpali_models.py#L129-L135

Current status uses [`process_queries`](https://github.com/illuin-tech/colpali/blob/17b86cd01059b3765539789cb7232e5cf8b4f946/colpali_engine/utils/processing_utils.py#L89) which is expected to take the raw texts, and handles formatting with prefix and suffix. That creates duplicated prefix and suffix.

Additionally, [this function is deprecated, so we should replace it by `process_text`](https://github.com/illuin-tech/colpali/blob/17b86cd01059b3765539789cb7232e5cf8b4f946/colpali_engine/utils/processing_utils.py#L74).

Simple fix is to replace `process_queries` by `process_text`, which would lead to the expected behavior:
`<query> -> <prefix> + <query> + 10*<augmentation_token>`

paultltc:
Closes #4360 and related to #4204.

Running small test on `VidoreArxivQARetrieval`.

### DIFF
--- a/mteb/models/model_implementations/colpali_models.py
+++ b/mteb/models/model_implementations/colpali_models.py
@@ -132,7 +132,7 @@ class ColPaliEngineWrapper(AbsEncoder):
                     + self.processor.query_augmentation_token * 10
                     for t in batch["text"]
                 ]
-                inputs = self.processor.process_queries(batch)
+                inputs = self.processor.process_texts(batch)
                 inputs = {k: v.to(self.device) for k, v in inputs.items()}
                 outs = self.encode_input(inputs)
                 all_embeds.extend(outs.cpu().to(torch.float32))


paultltc:
Check performance.

Couldn't run larger models as I was constrained with one L4 GPU.

## Run Details

**Task:** `VidoreArxivQARetrieval`
**Models:** `["vidore/colSmol-256M", "vidore/colSmol-500M", "nomic-ai/nomic-embed-multimodal-3b"]`

| Parameter | Value |
|-----------|-------|
| Hardware | 1x L4 22GB |
| Precision | float32 (left) / bfloat16 (right) |
| Batch size | 1 |

## Results 

| | Colsmol 256 | Colsmol 512 | Colnomic 3B |
|---|---|---|---|
| **Current: `process_queries`** | 0.72243 / 0.71970 | 0.73621 / 0.73635 | – / 0.86797 |
| **Fix: `process_texts`** | 0.72579 / 0.72730 | 0.74793 / 0.74838 | – / 0.86797 |

- No loss in performance, even increasing for colsmol models as expected
- Exact same result for colnomic
- Consistent across torch dtype for colsmol (couldn't run float32 for colnomic but expect the same)

@KennethEnevoldsen @Samoed

Samoed:
I think drop in colsmol performance is big enough. Probably better to keep these tokens during encoding. CC @QuentinJGMace

KennethEnevoldsen:
@paultltc  I don't see this fixing #4204 though? Am I missing anything?

QuentinJGMace:
Hey @paultltc @KennethEnevoldsen the difference with #4204 is that colpali_models.py used `processor.process_queries` on top of already pre-processed text which added additional tokens (which this PR adresses).
The nomic team does not add those additional tokens before using `processor.process_queries` so that's perfectly equivalent to what's being done here, no need to fix anything on that side.

paultltc:
> @paultltc I don't see this fixing #4204 though? Am I missing anything?

Sorry should have said related, not fixing haha.

It is related to the discussion on the misuse of `processor.process_queries` as mentioned by @QuentinJGMace. I would say it's better to use `processor.process_texts` and let call site define the format, that's what we decided for `colpali_engine`. 

If the nomic team did not used the following format created by `processor.process_queries` that could explain the mismatch.

paultltc:
> I think drop in colsmol performance is big enough. Probably better to keep these tokens during encoding.

`process_texts` actually keeps the augmentation tokens as we add them in the line before, which is intended for colpali/colqwen models. The problem of using `process_queries` here is that it adds them twice, which confuses the embedder.

PR #4367, opened 2026-04-10
fix: Handle quantization in sentence transformers as an experiment

### LINKED ISSUE #4302: Handle `SentenceTransformers` quantization

After https://github.com/embeddings-benchmark/mteb/pull/3950 we added quantization methods to mteb. We should support sentence transformers quantizations too, but they'll be passed to `encode` during evaluation

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4302

Handle quantization passed in sentence-transformer style

### DIFF
--- a/mteb/evaluate.py
+++ b/mteb/evaluate.py
@@ -23,7 +23,7 @@ from mteb.models.sentence_transformer_wrapper import (
 )
 from mteb.results import ModelResult, TaskResult
 from mteb.results.task_result import TaskError
-from mteb.types import OutputDType, PromptType
+from mteb.types import PromptType
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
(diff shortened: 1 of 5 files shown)


mteb/types/_encoder_io.py
```diff
@@ -165,7 +167,7 @@ class MultimodalInput(  # type: ignore[misc]
     pass
 
 
-class OutputDType(str, Enum):
+class OutputDType(HelpfulStrEnum):
```
KennethEnevoldsen, line 169: might also be useful to change for PromptType

mteb/evaluate.py
```diff
@@ -348,6 +348,24 @@ def evaluate(  # noqa: PLR0913, PLR0914
     model, meta, model_name, model_revision = _sanitize_model(model)
     _check_model_modalities(meta, tasks)
 
+    if "precision" in encode_kwargs:
```
KennethEnevoldsen: hmm but shouldn't we only handle this for sentence transformers? (I am unsure if we should generally do this for all models). I would probably move it to the wrapper instead of here.

KennethEnevoldsen (approved):

PR #4370, opened 2026-04-12
model: add BidirLM/BidirLM-Omni-2.5B-Embedding

### LINKED ISSUE #4364: Add model: BidirLM

https://huggingface.co/BidirLM/BidirLM-Omni-2.5B-Embedding

### Arxiv link

https://arxiv.org/abs/2604.02045

### Description of the model

Congratulations with release @Nicolas-BZRD @Theodb!

Nicolas-BZRD:
## Model
BidirLM/BidirLM-Omni-2.5B-Embedding — omnimodal encoder (text/image/audio), 2.5B parameters.

Paper: https://arxiv.org/abs/2604.02045
Hub: https://huggingface.co/BidirLM/BidirLM-Omni-2.5B-Embedding

## Checklist
- [x] The model is public on HuggingFace Hub
- [x] `mteb.get_model_meta()` tested locally
- [x] `mteb.get_model()` tested locally
- [x] Tested on representative tasks locally
- [x] `n_embedding_parameters` set (315,098,112)
- [x] Training datasets listed (75 datasets)
- [x] Citation provided (arXiv:2604.02045)

Close https://github.com/embeddings-benchmark/mteb/issues/4364


mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,706 @@
...
+                    f"Cannot fuse modalities with different sample counts: "
+                    f"{[len(p) for p in present]}"
+                )
+            fused = fused + e
+        return fused.to(torch.float32).numpy()
+
+
+# ── ModelMeta ─────────────────────────────────────────────────────────────────
+bidirlm_omni_2_5b = ModelMeta(
+    name="BidirLM/BidirLM-Omni-2.5B-Embedding",
```
Samoed, line 187: Can you add other variants of your model?

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,706 @@
...
+        **kwargs: Any,
+    ) -> None:
+        from sentence_transformers import SentenceTransformer
+        from sentence_transformers import models as st_models
+
+        from huggingface_hub import snapshot_download
+        model_path_resolved = snapshot_download(repo_id=model_name, revision=revision)
+        if model_path_resolved not in sys.path:
+            sys.path.insert(0, model_path_resolved)
+        from input_module_bidirlm_omni import BidirLMOmniInputModule
```
Samoed: Can you make possible to include this with `trust_remote_code` or integrate with sentence transformers 5.4 format?

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,706 @@
...
+        # Asymmetric retrieval: documents get no instruction unless the prompt
+        # dict explicitly provides a "passage" key.
+        if (
+            ("Retrieval" in task_type or "Reranking" in task_type)
+            and prompt_type == PromptType.document
+            and not (isinstance(entry, dict) and "passage" in entry)
+        ):
+            return None
+
+        if not instruction and hasattr(task_metadata, "description") and task_metadata.description:
```
Samoed: Not sure if using description here is good idea

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: The other variants are text-only and don’t rely on the same encoding code. My co-author will push the rest this morning. For simplicity it might be best to split them as he knows his code better than I do and vice versa

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,706 @@
...
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput
+
+logger = logging.getLogger(__name__)
+
+# ── Languages ─────────────────────────────────────────────────────────────────
```
KennethEnevoldsen:
```suggestion
```

KennethEnevoldsen: can you remove these (it should be possible to have an overview of the code without it

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,706 @@
...
+        **kwargs: Any,
+    ) -> torch.Tensor:
+        input_module = self.model[0]
+        all_embeddings = []
+        for batch in tqdm(loader, disable=not show_progress_bar, desc="Audio"):
+            features = input_module.tokenize(batch["audio"])
+            all_embeddings.append(self._run_forward(features))
+        return torch.cat(all_embeddings, dim=0)
+
+    # ── AbsEncoder.encode ─────────────────────────────────────────────────────
```
KennethEnevoldsen: This seems more like a thing that should be in a docstring

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,706 @@
...
+
+        input_module = BidirLMOmniInputModule(
+            model_name=model_name,
+            trust_remote_code=True,
+            max_sequence_length=max_sequence_length,
+            attn_implementation=attn_implementation,
+        )
+        pooling = st_models.Pooling(POOLING_DIM, pooling_mode="mean")
+        self.model = SentenceTransformer(
+            modules=[input_module, pooling], trust_remote_code=True
```
KennethEnevoldsen: can you move trust remote code up to the init and add to loader_kwargs

mteb/models/model_implementations/bidirlm_omni_models.py
KennethEnevoldsen: We would really encourage integrating it with sentence transformers, both to make it more generally available and reduce maintenance costs on our end.

Nicolas-BZRD: I fixed the trust_remote_code issue; model loading now follows standard practices

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: Fix: Added high-scoring descriptions to the instruction dictionary and removed low-scoring ones. I updated results in the results PR in parallel

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: Regarding ST5.4, I worked with Tom and the model is now compatible. The results we are pushing were generated using this current modeling. For future benchmarks, we could transition to the new ST5.4 compatible version. However I suggest keeping the current version as is, since it now follows standard practices for loading and ensures reproducibility, especially as results on MIEB seem very sensitive to modeling changes

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,768 @@
...
+
+        # Fuse multiple modalities by element-wise addition
+        fused = present[0]
+        for e in present[1:]:
+            if len(e) != len(fused):
+                raise ValueError(
+                    f"Cannot fuse modalities with different sample counts: "
+                    f"{[len(p) for p in present]}"
+                )
+            fused = fused + e
```
Samoed: Seems you currently processing one modality at time. Maybe would be better to convert into message and process everything together?

mteb/models/model_implementations/bidirlm_omni_models.py
Samoed: > the model is now compatible

Hm. I don't see changes in model repo. UPD. Found that you put it in different repo https://huggingface.co/BidirLM/omni-st5.4

Samoed: Maybe you can add this model separately here then? Also, I think it's better to rename `omni-st5.4`, because it's a bit unclear what model it is

Nicolas-BZRD: Results were computed using this modeling, so I am keeping the new version separate for now but it will be pushed once this is finished. Since ST5.4 produces slightly different embeddings and we evaluated our model using the current version

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,768 @@
...
+
+    def encode(
+        self,
+        inputs: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
```
Samoed:
```suggestion
        **kwargs: Unpack[EncodeKwargs],
```

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: I pushed everything to ST5.4 and reran the full evaluation. We saw some minor variations on the fully zero-shot evaluations (i.e., retrieval and all other tasks that don't rely on linear probing), but on average, the results are the same. We can move forward with this version. I will push the updated results to the other PR; please let me know your feedback

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: This is now fix with the refactor to ST5.4 and encode function

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -679,75 +606,59 @@ def encode(
...
+                content = []
+                if has_image:
+                    content.append({"type": "image", "image": batch["image"][i]})
+                if has_audio:
+                    content.append({"type": "audio", "audio": batch["audio"][i]})
+                if has_text:
+                    text = batch["text"][i]
+                    content.append({
+                        "type": "text",
+                        "text": f"{instruction} {text}" if instruction else text,
```
tomaarsen: You can also incorporate this directly into the chat template of the model, if you wish. When calling `model.encode(...)` with `prompt="example prompt text"` and the model supports the `message` modality, the `prompt` will be passed as a `{"role": "system": "content": "example prompt text"}` or `{"role": "system": "content": {"type": "text", "text": "example prompt text"}}` (depending on whether the `message` `format` in the `sentence_bert_config.json` is "flat" or "structured") before the `{"role": "user", "content": ...}` with all of your information. 

The chat template can then e.g. assign the system prompt as a variable, skip the `"role": "system"` message (so it doesn't get included with all the `<|im_start|>`, `<|im_end|>`, etc.), and prepend the prompt only if there's text [here](https://huggingface.co/BidirLM/BidirLM-Omni-2.5B-Embedding/blob/main/chat_template.jinja#L56). It can just be ignored for all other modalities. 

Then you can construct a list like:
```
all_inputs = [{"text": ..., "image": ..., "audio": ...}, ...]
```
And it will be converted to this message format behind the scenes in ST. However, it's a bit of a personal preference. It might be nicer as otherwise it's not possible to use prompts with `model.encode(..., prompt="...")`.

This already looks a lot simpler than before though!

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,679 @@
...
+
+        # Limit text length if no image/audio is present, otherwise use the model's max context length (32768 tokens).
+        # This prevents truncating special tokens in multimodal which raised error while still enabling to limit text-only inputs.
+        if has_text and not has_image and not has_audio:
+            self.model.max_seq_length = self.max_text_length
+        else:
+            self.model.max_seq_length = 32768
+        return self.model.encode(
+            all_messages,
+            batch_size=inputs.batch_size or 32,
```
Samoed:
```suggestion
            batch_size=kwargs["batch_size"],
```

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,679 @@
...
+            show_progress_bar=True,
+            convert_to_numpy=True,
+        )
+
+
+bidirlm_omni_2_5b = ModelMeta(
+    name="BidirLM/BidirLM-Omni-2.5B-Embedding",
+    loader=BidirLMOmniEncoder,
+    loader_kwargs=dict(
+        trust_remote_code=True,
```
Samoed, line 190: Is it still required?

Nicolas-BZRD: trust_remote_code ? I think yes, the modeling will not be added on transformers lib

tomaarsen: Indeed, still required. Not to load custom Sentence Transformers modules, but only to load the custom Transformers code.

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: I think we are good, prompt is now handle in chat template

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,656 @@
...
+        for batch in inputs:
+            for i in range(len(batch[active_cols[0]])):
+                content = []
+                if has_image:
+                    content.append({"type": "image", "image": batch["image"][i]})
+                if has_audio:
+                    content.append({"type": "audio", "audio": batch["audio"][i]})
+                if has_text:
+                    content.append({"type": "text", "text": batch["text"][i]})
+                all_messages.append([{"role": "user", "content": content}])
```
tomaarsen:
```suggestion
        all_inputs = [
            {
                key: value for key, value in input_dict if key in ("text", "image", "audio")
            }
            for input_dict in inputs
        ]
```
I think something like this would also work, it'll be converted to the message format behind the scenes: we can just pass a list of dictionaries like `[{"text": ..., "image": ...}, ...]`. 
Great work on this! The implementation is getting very concise 👍 

Note: I didn't test this, I just wrote it in the GitHub UI.

Samoed: Doesn't this disable processing with the chat template with system/user?

Nicolas-BZRD: Ok I implemented it. We need to keep filtering column because unlike image=None which the image processor handles, audio=None crashes in transformers because it calls load_audio(value) on every audio content item without None guard. transformers/processing_utils.py -> lines 1808–1832

Nicolas-BZRD: @Samoed I will let @tomaarsen develop this, but now ST5.4 handles this in the background with chat templates, etc. It's managed more on the model modeling side than in the MTEB evaluation interface code

Nicolas-BZRD: *I tested it and everything is similar as before

Samoed: Saw that you pass `prompt=instruction`. Then yes, it would work same

Nicolas-BZRD:
Hey everyone, I’ve fixed the latest lint issues and all tests are passing, so I think we’re good to go. Let me know if you have any remaining concerns and thanks for all the reviews which massively improved the code simplicity

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,649 @@
...
+from mteb.types import PromptType
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+    from typing_extensions import Unpack
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, EncodeKwargs
+
+BIDIRLM_OMNI_LANGUAGES = [
```
Samoed: Can you reuse languages, datasets, citation, prompts values from text version? https://github.com/embeddings-benchmark/mteb/blob/e8a40693b00005cb0362bd6c5798d61287a196f3/mteb/models/model_implementations/bidirlm_models.py

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,649 @@
...
+            self.model.max_seq_length = self.max_text_length
+        else:
+            self.model.max_seq_length = 32768
+        return self.model.encode(
+            all_inputs,
+            prompt=instruction,
+            batch_size=kwargs["batch_size"],
+            show_progress_bar=kwargs["show_progress_bar"],
+            convert_to_numpy=True,
+        )
```
Samoed, line 183:
```suggestion
        return self.model.encode(
            all_inputs,
            prompt=instruction,
            **kwargs,
        )
```

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,649 @@
...
+        Builds conversation messages from whichever modalities are present and
+        delegates to SentenceTransformer.encode() via the native 'message' modality.
+        """
+        ds_features = inputs.dataset.features
+
+        active_cols = [
+            col
+            for col in ("image", "audio", "text")
+            if col in ds_features and inputs.dataset[0].get(col) is not None
+        ]
```
Samoed, line 162: We have integrated video tasks in main branch now https://github.com/embeddings-benchmark/mteb/pull/4383. I think it would be great if you could add video modality

mteb/models/model_implementations/bidirlm_omni_models.py
```diff
@@ -0,0 +1,649 @@
...
+
+        if task_type == "Summarization":
+            return None
+
+        entry = self._lookup_prompt(task_metadata.name)
+
+        # Asymmetric retrieval: documents get no instruction unless the prompt
+        # dict explicitly provides a "passage" key.
+        if (
+            ("Retrieval" in task_type or "Reranking" in task_type)
```
Samoed: Multimodal dataset could have different `task_type`, e.g. `Any2AnyRetrieval`. Because of that better to use `metadata.simplified_task_type`

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: The model was not specifically trained for video tasks, and I haven't yet verified the video processing pipeline. Given the rapid evolution of the multimodal field and my limited availability in the coming weeks, I prioritized shipping the model for its primary use case. Let's publish the initial results now, with plans to maybe explore integrating additional features in the future

Samoed: Yes, agree. I thought that the model could support videos as well

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: Done, and add the multimodal one

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: Done

mteb/models/model_implementations/bidirlm_omni_models.py
Nicolas-BZRD: good catch thanks, done

mteb/models/model_implementations/bidirlm_omni_models.py
Samoed: Can you import these values directly?

Nicolas-BZRD: What do you mean ? Importing them from text model and add the missing datasets ?

Samoed: Yes. Also, you can import languages and prompts

Nicolas-BZRD: done 👍

Samoed (approved):
Great work!

KennethEnevoldsen (approved):
Great to see this - I don't have anything to add here. I will just enable the automerge

PR #4383, opened 2026-04-15
Move Kinetics400 out of video and add zeroshot version

isaac-chung:
If you add a model or a dataset, please add the corresponding checklist:

* [dataset checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_dataset/#submit-a-pr)
* [model checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_model/#submitting-your-model-as-a-pr)

PR #4392, opened 2026-04-16
fix: Add `is_beta` to task metadata

KennethEnevoldsen:
- Added *is_beta to the task metadata
- Added a warning on initializing a dataset when `is_beta` is `False` (to avoid spamming people with warnings)
- Added exclude_beta to get_tasks and filter_tasks, for now I set it to False
- add tests

Note that the warning happens on `mteb.get_task("name of beta")`

Also note merge target is as some tests fail there, which will also fail here: #4383


mteb/filter_tasks.py
```diff
@@ -68,6 +69,8 @@ def filter_tasks(
     exclude_superseded: bool = False,
     exclude_aggregate: bool = False,
     exclude_private: bool = False,
+    exclude_beta: bool = False,
+    exlude_beta: bool = False,
```
isaac-chung:
```suggestion
    exclude_beta: bool = False,
```

mteb/get_tasks.py
```diff
@@ -261,6 +263,7 @@ def get_tasks(  # noqa: PLR0913, PLR0917
             If False, keep tasks if _any_ of the task's modalities match the filter modalities.
         exclude_aggregate: If True, exclude aggregate tasks. If False, both aggregate and non-aggregate tasks are returned.
         exclude_private: If True (default), exclude private/closed datasets (is_public=False). If False, include both public and private datasets.
+        exlude_beta: If True (default), exclude datasets that are in beta (is_beta=True). If False, include both beta and non-beta datasets.
```
isaac-chung:
```suggestion
        exclude_beta: If True (default), exclude datasets that are in beta (is_beta=True). If False, include both beta and non-beta datasets.
```

mteb/filter_tasks.py
```diff
@@ -103,6 +107,7 @@ def filter_tasks(  # noqa: PLR0913
             If False, keep tasks if _any_ of the task's modalities match the filter modalities.
         exclude_aggregate: If True, exclude aggregate tasks. If False, both aggregate and non-aggregate tasks are returned.
         exclude_private: If True (default), exclude private/closed datasets (is_public=False). If False, include both public and private datasets.
+        exclude_beta: If True (default), exclude datasets that are in beta (is_beta=True). If False, include both beta and non-beta datasets.
```
Samoed, line 109: Docstring not matches default value. I think we should exclude beta by default

KennethEnevoldsen:
You guys are fast! Fixed the remaining issues, added a test and converted default from False to True (see main comment as to why).

Note that the current formulation, when we create a benchmark, and load beta tasks it will raise a bunch on warnings when loading mteb. I am not sure how to best handle this. I suppose we normally wouldn't have a benchmark with beta tasks, but here we might have.

isaac-chung (approved):

PR #4395, opened 2026-04-16
fix: Auto add base model to `ModelMeta`

Samoed:
Automatically add source model from model's metadata to `ModelMeta`


KennethEnevoldsen (approved):

KennethEnevoldsen:
Lower deps seems to fail:

> FAILED tests/test_integrations/test_integration_with_sentencetransformers.py::test_sentence_transformer_integration_cross_encoder[cross-encoder/ms-marco-TinyBERT-L2-v2-task1] - AttributeError: 'CrossEncoder' object has no attribute 'model_card_data'

PR #4401, opened 2026-04-16
fix: move `mteb_version` to individual subsets in result files

### LINKED ISSUE #4354: Move mteb_version to individual subsets in result files

Currently there is a single mteb_version field in each task's result file. If new subsets are evaluated using a different MTEB version, the results of existing subsets will be erased.

Originally discussed in https://github.com/embeddings-benchmark/results/pull/465

fzoll:
Store mteb_version per subset in score dicts so that results evaluated with different MTEB versions can be merged without erasing existing subset scores. The top-level mteb_version is now computed as the latest version across all subsets.

- Add mteb_version to each subset's score dict in from_task_results()
- Remove mteb_version from default merge criteria (is_mergeable/merge)
- Compute top-level mteb_version from per-subset versions after merge
- Fix bug in aggregated_task.py where mteb_version was always overwritten regardless of version mismatch condition

Close #4354

### DIFF
--- a/mteb/results/task_result.py
+++ b/mteb/results/task_result.py
@@ -194,6 +194,7 @@ class TaskResult(BaseModel):  # noqa: PLR0904
         """
         task_meta = task.metadata
         subset2langscripts = task_meta.hf_subsets_to_langscripts
+        mteb_ver = version("mteb")
         flat_scores = defaultdict(list)
         for split, hf_subset_scores in scores.items():
             for hf_subset, hf_scores in hf_subset_scores.items():
@@ -202,13 +203,14 @@ class TaskResult(BaseModel):  # noqa: PLR0904
                     **hf_scores,
                     "hf_subset": hf_subset,
                     "languages": eval_langs,
+                    "mteb_version": mteb_ver,
                 }
                 flat_scores[split].append(_scores)
 
         return TaskResult(
             dataset_revision=task.metadata.revision,
             task_name=task.metadata.name,
-            mteb_version=version("mteb"),
+            mteb_version=mteb_ver,
             scores=flat_scores,
             evaluation_time=evaluation_time,
             kg_co2_emissions=kg_co2_emissions,
@@ -730,7 +732,6 @@ class TaskResult(BaseModel):  # noqa: PLR0904
         self,
         result: TaskResult | AbsTask,
         criteria: list[str] | list[Criteria] = [
-            "mteb_version",
             "dataset_revision",
         ],
         raise_error: bool = False,
@@ -739,7 +740,7 @@ class TaskResult(BaseModel):  # noqa: PLR0904
 
         Args:
             result: The TaskResult or Task object to check against.
-            criteria: Additional criteria to check for merging. Can be "mteb_version" or "dataset_revision".
+            criteria: Additional criteria to check for merging. Can be "dataset_revision" or "mteb_version" (opt-in).
                 It will always check that the task name match.
             raise_error: If True, raises an error if the objects cannot be merged. If False, returns False.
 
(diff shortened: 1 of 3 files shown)


mteb/results/task_result.py
```diff
@@ -730,7 +732,6 @@ def is_mergeable(
         self,
         result: TaskResult | AbsTask,
         criteria: list[str] | list[Criteria] = [
-            "mteb_version",
```
Samoed, line 733: If you removed it from here, then you should remove `if Criteria.MTEB_VERSION in criteria`

fzoll:
Removed the dead `Criteria.MTEB_VERSION` check and unused `mteb_version` variable assignments.

mteb/results/task_result.py
```diff
@@ -769,13 +768,6 @@ def is_mergeable(
             logger.debug(msg)
             return False
 
-        if Criteria.MTEB_VERSION in criteria and self.mteb_version != mteb_version:
```
KennethEnevoldsen, line 772: Don't we want to keep the option to use `mteb_version` but just not have it as the default? (i.e. if I want to be always use the same version)

tests/test_results/test_task_results.py
```diff
@@ -153,6 +156,109 @@ def test_task_results_validate_and_filter():
...
+    merged = existing.merge(new)
+
+    # Both subsets should be present
+    assert len(merged.scores["train"]) == 2
+    subsets = {s["hf_subset"]: s for s in merged.scores["train"]}
+    assert subsets["en-de"]["mteb_version"] == "2.12.4"
+    assert subsets["en-fr"]["mteb_version"] == "2.20.1"
+
+    # Top-level version should be the latest
+    assert merged.mteb_version == "2.20.1"
```
KennethEnevoldsen: hmm but shouldn't this be a range or list? 

It is also an option to remove it completely, but the add a property that returns a range. Then we also of course need to fix it when loading historic result files (which then just assigns the mteb version to each subset)

KennethEnevoldsen (commented):
Looks good, though we need to decide what to do with the top level version. I would suggest just removing (see comment below).

I would also keep the option to ensure the same version is required to merge, but not have it as the default.

With this change, we could also save the results per. subset (so if you stop a run you don't have to start from scratch). This is probably a seperate PR though.

fzoll:
Good point — restored `Criteria.MTEB_VERSION` as an opt-in check (not in default criteria). Regarding the top-level `mteb_version`: agree it could become a range or be removed in favor of per-subset versions. Happy to tackle that in a follow-up PR if you'd like.

KennethEnevoldsen:
> Regarding the top-level mteb_version: agree it could become a range or be removed in favor of per-subset versions. Happy to tackle that in a follow-up PR if you'd like.

Hmm but I don't want to merge it in with the top-level being a specific version when the sub-levels are not that version (I believe it would cause confusion)

fzoll:
Good point — changed the top-level `mteb_version` to a range (e.g. `"2.12.4-2.20.1"`) when subsets were evaluated with different versions. Single version if all match.

Samoed:
@fzoll Can you keep the commits separate? With force-pushed commits this is very hard to review

tests/test_results/test_task_results.py
```diff
@@ -153,6 +156,109 @@ def test_task_results_validate_and_filter():
...
+    )
+
+    merged = existing.merge(new)
+    # Old subset has no per-subset version, new one does
+    subsets = {s["hf_subset"]: s for s in merged.scores["train"]}
+    assert "mteb_version" not in subsets["en-de"]
+    assert subsets["en-fr"]["mteb_version"] == "2.20.1"
+
+    # Top-level should be the latest found across subsets
+    assert merged.mteb_version == "2.20.1"
```
KennethEnevoldsen, line 259: it should be "2.12.4-2.20.1" and subsets["en-de"]["mteb_version"]==2.12.4

KennethEnevoldsen (commented):
A minor change otherwise I think this is good

KennethEnevoldsen:
@Samoed added the missing fixes here, anything you want to add?

Samoed (approved):

Issue #3471, opened 2025-10-22
Add number of queries and documents to retrieval statistics


Samoed:
### Description of the feature

Add number of queries and documents to retrieval statistics, because for now we compute only unique items. We can do this without dataset loading, just by retrieving information about HF dataset.

I think we can add it in 2 ways.

1. Add to image/text statistics
```json
{
  "text_statistics": {
    "total": 100,
    ...
  },
  "image_statistics": {
    "total": 100,
    ...
  }
}
```
2. Or add it custom for retrieval task
```json
{
  "total_documents": 100,
  "document_statistics": {...}
}
```
or make it a not compatible with current implementation, but since this is just statistics I think we can do that
```json
{
  "document_statistics": {
    "total": 100,
    "image_statistics": {...},
    "text_statistics": {...},
  }
}
```
I think we might want to change our tasks to second format to not duplicate 3 times same information ("text_input", "image_input", "audio_input" etc.) and maybe simplify statistics more for all tasks.

Samoed:
Also, format of statistics can be changed for `PairClassification` and `STS`. E.g. 

```json
{
  "input1_statistics": {
    "text": {...},
    "image": {...},
  }
}
```

Samoed:
Overall, I think we can change format for all task types like

```python
class InputStatistics(TypedDict):
    text: TextStatistics | None
    image: ImageStatistics | None
    audio: AudioStatistics | None

class ClassificationStatistics(TypedDict):
    samples: int
    input_statistics: InputStatistics
    label_statistics: LabelStatistics

class PairClassificationStatistics(TypedDict):
    samples: int
    input1_statistics: InputStatistics
    input2_statistics: InputStatistics
    label_statistics: LabelStatistics

class RetrievalStatistics(TypedDict):
    samples: int
    # can be extended to include number of samples of queries/documents
    queries_statistics: InputStatistics
    documents_statistics: InputStatistics

    relevant_docs_statistics: RelevantDocsStatistics
    top_ranked_statistics: TopRankedStatistics | None
```

Samoed:
Also, modality statics can be changed like:

```python
class InputItemStatistics(TypedDict):
   min: int
   average: float
   max: int

class TextStatistics(TypedDict):
   total_characters: int
   text: InputItemStatistics  # maybe would be easier to inherit from InputItemStatistics, not sure
   unique: int

class ImageStatistics(TypedDict):
   total: int
   unique: int
   height: InputItemStatistics
[...]

PR #4403, opened 2026-04-16
feat: add num_queries and num_documents to retrieval statistics

fzoll:
Issue #3471

- Add InputStatistics TypedDict to group text/image/audio stats
- Add num_queries and num_documents fields to RetrievalDescriptiveStatistics
- Restructure retrieval stats: flat documents_text_statistics/queries_text_statistics keys replaced with nested documents_statistics/queries_statistics containing text/image/audio subkeys
- Migrate all existing retrieval descriptive stats JSON files to new format
- Update test_task_quality.py to handle both old and new stat key formats

### DIFF
--- a/mteb/abstasks/retrieval.py
+++ b/mteb/abstasks/retrieval.py
@@ -73,7 +73,9 @@ class RetrievalDescriptiveStatistics(SplitDescriptiveStatistics):
     """Descriptive statistics for Retrieval
 
     Attributes:
-        num_samples: Number of queries and documents
+        num_samples: Total number of queries and documents
+        num_queries: Number of queries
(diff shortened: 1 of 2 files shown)


Samoed (commented):
I think part with changing format of statistics is too soon. We haven't decided final format of statistics and if we want to break compatibility.

For now, I think you can keep `num_queries` and `num_documents` for now and revert other parts. How did you update statistics files?

fzoll:
Reverted the format restructuring — now only adds `num_queries` and `num_documents` fields, no other changes.

Samoed:
Can you update tests and fill old statistics?

Issue #4333, opened 2026-03-31
Remove tasks without a main score


KennethEnevoldsen:
Multiple tasks have no main score. E.g. SciDocsRR:

```
{
  "dataset_revision": "d3c5e1fc0b855ab6097bf1cda04dd73947d7caab",
  "mteb_dataset_name": "SciDocsRR",
  "mteb_version": "1.0.3.dev0",
  "test": {
    "evaluation_time": 2242.46,
    "map": 0.8458048271261831,
    "mrr": 0.9600679730581692
  }
}
```
I suspect these are probably just the main score (map_at_1000), so we could just rename them, we could also fix the loading script for historic data. **The ideal solution is of course simply to rerun these**

Discovered when loading results using:

```py
import mteb

cache = mteb.ResultCache()
results = cache.load_results()

main_score_is_none = []
for model_res in results.model_results:
    for task_result in model_res.task_results:
        for split, split_scores in task_result.scores.items():
            contains_none = False
            for subset_scores in split_scores:
                if subset_scores["main_score"] is None:
                    contains_none = True
            if contains_none:
                print(model_res.__repr__())
                print(f"\t- '{task_result.task_name}': main score is None")
                main_score_is_none.append([model_res, task_result])


# ModelResult(model_name='Alibaba-NLP/gme-Qwen2-VL-7B-Instruct', model_revision='external', task_results=[...](#101), ...)
#         - 'CMedQAv2-reranking': main score is None
# ModelResult(model_name='Alibaba-NLP/gme-Qwen2-VL-7B-Instruct', model_revision='external', task_results=[...](#101), ...)
#         - 'CMedQAv1-reranking': main score is None

# ModelResult(model_name='Alibaba-NLP/gme-Qwen2-VL-2B-Instruct', model_revision='external', task_results=[...](#101), ...)
#         - 'CMedQAv2-reranking': main score is None
# ModelResult(model_name='Alibaba-NLP/gme-Qwen2-VL-2B-Instruct', model_revision='external', task_results=[...](#101), ...)
#         - 'CMedQAv1-reranking': main score is None


# ModelResult(model_name='openai/text-embedding-3-small', model_revision='1', task_results=[...](#183), ...)
#         - 'MindSmallReranking': main score is None
# ModelResult(model_name='openai/text-embedding-3-small', model_revision='1', task_results=[...](#183), ...)
#         - 'SciDocsRR': main score is None
[...]

Issue #3921, opened 2026-01-12
Phase out model results without model revisions


KennethEnevoldsen:
### Description of the feature

Currently, the leaderboard displays model results without revisions. These are historic results, but they can be replaced with newer result or completely deleted (in cases which we deem it relevant to remove)

Discussed in: 
https://github.com/embeddings-benchmark/mteb/issues/3918#issuecomment-3738726190

isaac-chung:
We know [this PR](https://github.com/embeddings-benchmark/results/pull/399) essentially solves this. So what are we waiting for again? I'm re-reading these issues and am getting confused about the core issue.

KennethEnevoldsen:
We would need to rerun some results for popular models to not completely remove these from the leaderboard

isaac-chung:
I cross-checked models with `External`/`No Revision` in the results repo and the `model_implementations` folder, and got a list of models in ascending memory MB order. Good thing is that all these models have code implementations in mteb with proper revisions. 

<details>
  <summary>Models with External/No Revision - Ranked by Memory</summary>
  
```md
## Summary

- **Total models with external/no_revision in cache**: 227
- **Models with code implementation in mteb**: 227
- **Models with proper revision in code**: 227
- **Models with memory_usage_mb in code**: 177
- **API models (memory not expected)**: 16
- **Non-API models missing memory_usage_mb**: 34
- **Models missing memory info in cache**: 83

## Legend

- **Cache Memory**: Memory from model_meta.json in cache (external/no_revision dir)
- **Code Impl**: Whether model has ModelMeta in mteb/models/model_implementations/
- **Code Rev**: Revision specified in code (proper revision = commit hash)
- **Code Mem**: memory_usage_mb specified in code

## Models Ranked by Memory (Ascending)

| Model | Cache Rev | Cache Mem (MB) | Code Impl | Code Rev | Code Mem (MB) | File |
|-------|-----------|----------------|-----------|----------|---------------|------|
| minishlab/potion-base-2M | external | 7 | Yes | 86db093558fb... | 7 | model2vec_models.py |
| minishlab/potion-base-4M | external | 14 | Yes | 81b1802ada41... | 14 | model2vec_models.py |
| minishlab/potion-base-8M | external | 29 | Yes | dcbec7aa2d52... | 29 | model2vec_models.py |
| minishlab/M2V_base_output | external | 29 | Yes | 02460ae401a2... | 29 | model2vec_models.py |
[...]

PR #4404, opened 2026-04-16
fix: remove tasks without a main score when loading historic results

fzoll:
Historic result files that lack the expected main_score metric (e.g. SciDocsRR with 'map' instead of 'map_at_1000') previously set main_score=None, causing downstream errors. Now these subsets are filtered out during conversion, and task results with no valid scores are skipped during cache loading with a warning.

Issue #4333

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mteb"
-version = "2.12.18"
+version = "2.12.19"
 description = "Massive Text Embedding Benchmark"
 readme = "README.md"
 authors = [


Samoed (changes requested):
This change would disallow load results from our main repo. To resolve this issue we need to rerun models with old format scores. I started in https://github.com/embeddings-benchmark/results/pull/474, but there are a lot of model and results

fzoll:
Reworked — instead of removing old results, it now infers `main_score` from the base metric name (e.g. `map` → `map_at_1000`). Backward-compatible with existing results.

Samoed:
You close PR and I don't see changes in diff

fzoll:
@Samoed Yeah, sorry, my mistake. I started working on multiple issues in parallel, i mixed this one with another one. Anyway, let's keep this one closed, please let me know if i can help with your https://github.com/embeddings-benchmark/results/pull/474 PR.

Samoed:
You can run tasks with old results format. You can start by rerunning models with `no_revision_available` revision. Source issue https://github.com/embeddings-benchmark/mteb/issues/3921

PR #4405, opened 2026-04-16
model: add language support to BM25 and add the subword BM25

### LINKED ISSUE #4335: BM25 model for non-Latin languages

The current [bm25 model implementatation](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/model_implementations/bm25.py) works not well for symbolic languages as it relies on PyStemmer for the tokenization, e.g., for Chinese or Japanese texts the results are very bad. The problem is that symbolic languages can not get tokenized by spaces. A good tokenization is non-trivial. There are tokenizers for specific languages, e.g., [jieba](https://github.com/fxsjy/jieba) for Chinese that could be used. However, a more simple solution that works for a broad range of languages would be to use trained multilingual subword tokenizers of huggingface models. Those should be much more robust although not optimal for non-English languages.

```python
from tokenizers import Tokenizer

hf_tokenizer = Tokenizer.from_pretrained("xlm-roberta-base")

texts = [
    "Which tokenizer works best with BM25?",
    "哪种分词器与BM25配合效果最好？",
    "BM25と相性の良いトークナイザーはどれですか？"
]

for text in texts:
    raw_tokens = hf_tokenizer.encode(text, add_special_tokens=False).tokens
    
    # Strip the space marker (e.g. '_') and filter out empty tokens (e.g. standalone spaces)
    clean_tokens = [
        t.replace(' ', '').replace('▁', '') 
        for t in raw_tokens 
        if t.replace(' ', '').replace('▁', '') != ''
    ]
    
    print(clean_tokens)
```

```text
['Which', 'to', 'ken', 'izer', 'works', 'best', 'with', 'BM', '25', '?']
['哪', '种', '分', '词', '器', '与', 'BM', '25', '配合', '效果', '最好', '?']
['BM', '25', 'と', '相', '性', 'の良い', 'トーク', 'ナイ', 'ザー', 'は', 'どれ', 'ですか', '?']
```

```python
import bm25s

# Run BM25
retriever = bm25s.BM25()
retriever.index(clean_tokens)
```

I am not sure if it is better to implement it for all languages, non-Latin languages, or only symbolic (none-alphabetic languages).

fzoll:
Add a new BM25 model (mteb/baseline-bm25s-multilingual) that uses a trained multilingual subword tokeniser (xlm-roberta-base) instead of PyStemmer. This enables proper tokenisation for non-Latin scripts like Chinese and Japanese, where whitespace-based splitting fails.

Close #4335


mteb/models/model_implementations/bm25_multilingual.py
```diff
@@ -0,0 +1,197 @@
...
+        corpus_idx_to_id: dict[int, str]
+
+        def __init__(
+            self,
+            previous_results: str | None = None,
+            hf_tokenizer_name: str = "xlm-roberta-base",
+            **kwargs,
+        ):
+            self.model = None
+            self.hf_tokenizer = Tokenizer.from_pretrained(hf_tokenizer_name)
```
Samoed: I think it would be better to move `BM25MultilingualSearch` outside the function in the original `bm25` and inherit from it. The same way `bm25` can be changed too.

fzoll:
Refactored — moved `BM25Search` outside the loader function. `BM25MultilingualSearch` now inherits from it, both in `bm25.py`.

mteb/models/model_implementations/bm25.py
```diff
@@ -21,114 +21,178 @@
...
+        ]  # concatenate all document values (title, text, ...)
+        encoded_corpus = self._encode(corpus_texts)
+
+        logger.info(
+            f"Indexing Corpus... {len(encoded_corpus.ids):,} documents, {len(encoded_corpus.vocab):,} vocab"
+        )
 
-            logger.info(f"Retrieving Results... {len(queries):,} queries")
+        # Create the BM25 model and index the corpus
+        self.retriever = self.bm25s.BM25()
```
Samoed:
```suggestion
        import bm25s
        self.retriever = bm25s.BM25()
```

Samoed (approved):
Looks good. Can you fix the test?

Samoed (approved):

mteb/models/model_implementations/bm25.py
```diff
@@ -21,114 +21,176 @@
...
+        self,
+        previous_results: str | None = None,
+        tokenizer_name: str = "xlm-roberta-base",
+        **kwargs,
+    ):
+        requires_package(
+            self, "bm25s", "bm25-multilingual", "pip install 'mteb[bm25s]'"
+        )
+        requires_package(
+            self, "tokenizers", "bm25-multilingual", "pip install tokenizers"
```
KennethEnevoldsen: isn't this included in sentence transformers (I don't think we need to specify it - if not we should probably add it to the bm25s dependencies)

mteb/models/model_implementations/bm25.py
```diff
@@ -21,114 +21,176 @@
...
+        )
+        from tokenizers import Tokenizer
+
+        self.model = None
+        self.stopwords = None
```
KennethEnevoldsen, line 149: do we want to use across document frequent tokens as stopwords? (e.g. words that appear in 90% of docs)

KennethEnevoldsen (commented):
Looks good - We need to do a few sanity check, e.g. run it on an english and chinese split just to see the differences in performance. Ideally chinese performance should be notably better, but the English performance should be slightly worse

fzoll:
Removed the `tokenizers` require check — it's a transitive dependency of `transformers`.

Regarding stopwords: good idea. BM25 already handles high-IDF terms well, but explicit stopword removal via document frequency could help. Want me to add it here or keep it for a follow-up?

KennethEnevoldsen:
> Regarding stopwords: good idea. BM25 already handles high-IDF terms well, but explicit stopword removal via document frequency could help. Want me to add it here or keep it for a follow-up?

I think it depends on the performance. I would add whatever performs the best, but wouldn't add it in a follow up as the prevents reproducibility so would do the testing before the merge

KennethEnevoldsen:
Ran a comparison of BM25 variants on LeCaRDv2 (Chinese legal case retrieval, 3,795 docs, 159 queries, ndcg@10):

| Model | ndcg@10 |
|---|---|
| bm25s (English stemmer, this PR baseline) | 0.359 |
| bm25s-multilingual (XLM-R subword, this PR) | 0.585 |
| bm25s-multilingual + freq stopwords (tokens in ≥90% of docs) | 0.608 |
| bm25s-unicode (script-aware char split, no ML model) | 0.543 |
| bm25s-lang-aware + Jieba (auto-detected from task metadata) | **0.648** |

The subword tokenizer seems very reasonable, but of course the best result comes from adding Jieba word segmentation + Chinese stopwords, both derived automatically from the task's `eval_langs` at index time.
This way, we can extend BM25 without much hassle. 

I would probably go with 

bm25s-multilingual + freq stopwords (rename to bm25s-subword) and incorperate language aware tokenization to bm25. Is another alternative also to only keep BM25 with subword as the fallback? (or maybe white space for latin-scripts) 
Let me know what you think?

@guenthermi, let me know what you think as well - any tricks I missed?

(added a lot of AI noise in the code, will do a clean rewrite once we have a decision of what we want to do)

Samoed:
Just as an experiment, tried to run with Qwen tokenizer (as one of most popular Chinese model) and got `0.61652` on `LeCaRDv2`

```python
import mteb
task = mteb.get_task("LeCaRDv2")
model = mteb.get_model("mteb/baseline-bm25s-multilingual", tokenizer_name="Qwen/Qwen3-0.6B")
cache = mteb.ResultCache("bm25_test")
mteb.evaluate(model, task, cache=cache)
```

KennethEnevoldsen:
Gemma-2 got 0.592 without freq. stopword removal and 0.610 with. I I think I will go for qwen, but leave it as a setting to change it.

KennethEnevoldsen:
Alright did a large scale refactor here:

Overall here is the changes:
1) Changed BM25 to now select tokenizer based on language, default to English if unclear - performance on English task e.g. "SciFact" remains unchanged, but before it was  Added language support to  
2) Refactored a lot of the tokenization building out of BM25 everything related to tokenization is not in an sklearn-like tokenizer `BM25Search` with a `.fit`, `fit_transform` and `transform` method.
3) Renamed the multilingual bm25 to unigram to better reflect the implementation. It is also not a simply argument change on the original bM25 (@Samoed I see that you didn't like this earlier is that still the case?)

The idea is to make the core of the BM25 logic simple and easy to read and collect the tokenizer construction in a seperate abstraction. 

This will give breaking results for the BM25 for non-english results so I bumped the revision on it. For chinese it gives quite a notably increase in performance (see comment above).


task | language | bm25s-branch | bm25s-subword-branch | bm25s (main) | comment
-- | -- | -- | -- | -- | --
SciFact | eng | **0.686** | 0.6332 | **0.6863** | same
TwitterHjerneRetrieval | dan | 0.2824 | 0.2784 | 0.2596 | slightly better (due to better stopword list)
LeCaRDv2 | zho | 0.6478 | 0.6307 | 0.359 | notably better
MrTyDiJaRetrievalLite | jpn | 0.2232 | 0.2419 | 0.0501 | notably better (but still worse than subword)
GerDaLIRSmall | deu | 0.5925 | 0.4904 | 0.5944 | close to the same (different stopword list)


---

validation and some refactors done with AI, but reviewed by me.

pyproject.toml
```diff
@@ -89,7 +89,7 @@ openai = ["openai>=1.41.0", "tiktoken>=0.8.0"]
 model2vec = ["model2vec>=0.3.0"]
 pylate = ["pylate>=1.3.1; python_full_version < '3.13'", "transformers>=4.52.0; python_full_version < '3.13'"]  # Required for sentence-transformers 5.0.0 compatibility (otherwise 1.1.6), pylate requires voyager, which in turn requires <=3.12
 msclap = ["msclap>=1.3.4", "soundfile>=0.13.1"]
-bm25s = ["bm25s>=0.2.6", "PyStemmer>=2.2.0.3"]
+bm25s = ["bm25s>=0.2.6", "PyStemmer>=2.2.0.3", "jieba>=0.42.1"]
```
Samoed: I'm unsure if we want to include it as part of `bm25s` requirements, because this library wasn't updated for 6 years

mteb/models/model_implementations/bm25.py
```diff
@@ -19,122 +22,587 @@
...
+      tokens appearing in ≥ ``freq_threshold`` fraction of corpus documents are
+      removed automatically (set ``freq_threshold=0`` to disable).
+
+    Pass explicit ``stopwords`` or ``stemmer_language`` to pin those settings and
+    skip auto-detection for the corresponding parameter.
+    """
+
+    def __init__(
+        self,
+        previous_results: str | None = None,
```
Samoed, line 322: Seems unused
```suggestion
```

mteb/models/model_implementations/bm25.py
```diff
@@ -19,122 +22,587 @@
 
 logger = logging.getLogger(__name__)
 
+# Logographic/no-space scripts where each character is its own token.
+# Includes CJK, Japanese kana, Hangul, Thai, Khmer, Myanmar.
+_NO_SPACE_SCRIPT = (
```
Samoed: I'm not sure if we want to include this, because probably list will probably be huge for CJK

Samoed (commented):
Overall looks good. Maybe @xhluca can suggest how to handle CJK languages

xhluca:
Sorry, I don't have any experience with CJK languages :/

pyproject.toml
KennethEnevoldsen: I could change it out to: https://github.com/mxcoras/jieba-next?

mteb/models/model_implementations/bm25.py
KennethEnevoldsen: yeah, it was there before so I didn't want to break anything so left it in

mteb/models/model_implementations/bm25.py
KennethEnevoldsen: We could simply do a unicode character tokenize? 

```
def _unicode_tokenize(text: str) -> list[str]:
    return list(unicodedata.normalize("NFKC", text).lower().replace(" ", ""))
```

pyproject.toml
KennethEnevoldsen: we could also completely remove jieba

Samoed: This is very hard to understand what's happening in this lib, because it don't have non chineese documentation, so I think it's better to make it optional

mteb/models/model_implementations/bm25.py
Samoed: Yes, this would be better

pyproject.toml
KennethEnevoldsen: I will remove it for now for simplicity and instead add an example on how to use a custom tokenizer to the docs

KennethEnevoldsen:
I will add a seperate issue on adding bm25 subword tokenizer as a reference model

KennethEnevoldsen:
@Samoed I believe this is good to merge?

docs/get_started/usage/defining_the_model.md
```diff
@@ -111,3 +111,52 @@ model = mteb.evaluate(model, tasks=tasks)
 ```
 
 If you want to submit your implementation to be included in the leaderboard see the section on [submitting a model](../../contributing/adding_a_model.md).
+
+## Using BM25 Baselines
```
Samoed: Not sure if we need section about bm25 in `defining_the_model`. Maybe would be better to put in new section in advanced usage or create a new section `Examples`

mteb/models/model_implementations/bm25.py
```diff
@@ -20,121 +22,555 @@
...
+        logger.info(
+            f"Language settings — stopwords: {self.stopwords_key!r}, "
+            f"stemmer: {stemmer_lang!r}, tokenizer: {self._tok_arg!r}"
+        )
+
+    def fit(self, corpus_texts: list[str]) -> BM25Tokenizer:
+        self.fit_transform(corpus_texts)
+        return self
+
+    def fit_transform(self, corpus_texts: list[str]):
```
Samoed, line 180: Can you add return types?

Samoed (commented):
Code wise I think this is good, but CI is failing. I think we can also update `bb25` to use new implementation

mteb/models/model_implementations/bm25.py
KennethEnevoldsen: The type is `Tokenized`, which I can't have as a top-level import. I would have to wrap it in a function to add those I believe.

docs/get_started/usage/defining_the_model.md
KennethEnevoldsen: moved to advanced usage and change the section title of that to "Examples"

KennethEnevoldsen:
fixed linting and double checked bb25 works as intended

Samoed (approved):
Great!

PR #4406, opened 2026-04-16
feat: add required_dependencies field to ModelMeta

fzoll:
Add a `required_dependencies` field to `ModelMeta` that lists packages needed to load a model. When `load_model()` catches an ImportError, the error message now includes these dependencies so users know what to install.

Issue #4350


Samoed:
@fzoll We already have PR for this https://github.com/embeddings-benchmark/mteb/pull/4356

PR #4399, opened 2026-04-16
fix: double retrieval dataset loading

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4397

PR #4407, opened 2026-04-16
tests: add regression test for double loading 

Samoed:
Add regression test for double loading from https://github.com/embeddings-benchmark/mteb/pull/4399


KennethEnevoldsen (approved):

KennethEnevoldsen:
Ahh yes of course!

Issue #4130, opened 2026-02-22
MVEB Overview


AdnanElAssadi56:
Classification/ZeroshotClassifcation:

1. [mteb/kinetics-400](https://huggingface.co/datasets/mteb/kinetics-400)
2. [mteb/HMDB51](https://huggingface.co/datasets/mteb/HMDB51)
3. [mteb/Breakfast](https://huggingface.co/datasets/mteb/Breakfast)
4. [mteb/kinetics-700-2020](https://huggingface.co/datasets/mteb/kinetics-700-2020)
5. [mteb/UCF101-51VA](https://huggingface.co/datasets/mteb/UCF101-51VA)
6. [mteb/SomethingSomethingV2](https://huggingface.co/datasets/mteb/SomethingSomethingV2)
7. [mteb/kinetics-600](https://huggingface.co/datasets/mteb/kinetics-600)
8. [mteb/VGGSound](https://huggingface.co/datasets/mteb/VGGSound)
9. [mteb/AVE-Dataset](https://huggingface.co/datasets/mteb/AVE-Dataset)
10. [mteb/Human-Animal-Cartoon](https://huggingface.co/datasets/mteb/Human-Animal-Cartoon)
11. [mteb/RAVDESS_AV](https://huggingface.co/datasets/mteb/RAVDESS_AV)
12. [mteb/MUSIC-AVQA_cls-preprocessed](https://huggingface.co/datasets/mteb/MUSIC-AVQA_cls-preprocessed)
13. [mteb/MELD](https://huggingface.co/datasets/mteb/MELD)
14. [mteb/AVMeme-Exam](https://huggingface.co/datasets/mteb/AVMeme-Exam)
15. [mteb/WorldSense_1min](https://huggingface.co/datasets/mteb/WorldSense_1min)

Pair Classification:

1. [mteb/Human-Animal-Cartoon](https://huggingface.co/datasets/mteb/Human-Animal-Cartoon)
2. [mteb/AVE-Dataset](https://huggingface.co/datasets/mteb/AVE-Dataset)
3. [mteb/RAVDESS_AV](https://huggingface.co/datasets/mteb/RAVDESS_AV)
4. [mteb/MELD](https://huggingface.co/datasets/mteb/MELD)
5. [mteb/MUSIC-AVQA_cls-preprocessed](https://huggingface.co/datasets/mteb/MUSIC-AVQA_cls-preprocessed)

Clustering:

[...]

PR #4425, opened 2026-04-18
add mteb/DiDeMo dataset

manandey:

#4130
Ran the mteb/baseline-random-encoder model.
[DiDeMoV2TRetrieval_results.json](https://github.com/user-attachments/files/26848405/DiDeMoV2TRetrieval_results.json)

cc @Samoed


Samoed (commented):
I think would be better to move all tasks inside one file to reuse `load_data` and just change columns name when loading

manandey:
@Samoed updated!

Samoed (approved):
Great!

deep9539:
Hi, nit: should this be "new dataset" label?

Samoed:
Fixed

PR #4433, opened 2026-04-18
Add VATEX retrieval tasks 

manandey:

https://github.com/embeddings-benchmark/mteb/issues/4130
Ran the mteb/baseline-random-encoder model.

[Uploading VATEXV2TRetrieval_eval.json…]()

cc @Samoed


mteb/tasks/retrieval/eng/vatex_retrieval.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class VATEXV2TRetrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="VATEXV2TRetrieval",
+        description=(
+            "Retrieve the English caption that describes a given video clip from "
+            "VATEX, a large-scale multilingual video description dataset."
+        ),
+        reference="https://huggingface.co/datasets/mteb/VATEX_test_1k",
```
Samoed, line 61:
```suggestion
        reference="https://arxiv.org/abs/1904.03493",
```

Samoed (commented):
Can you fix citation and compute statistics?

AdnanElAssadi56:
Should we do VA2T and T2VA here as well?

Also, not sure if, A2T and T2A applicable if original dataset was intended for video.

Samoed:
I think this is fine to see which parts of video are important

AdnanElAssadi56:
> I think this is fine to see which parts of video are important

then we should isolate video as well (the frames); still thinking of the best naming though to avoid confusion.

mteb/tasks/retrieval/eng/vatex_retrieval.py
```diff
@@ -0,0 +1,175 @@
...
+        is_beta=True,
+    )
+
+    def load_data(self, num_proc: int | None = None, **kwargs) -> None:
+        _load_vatex(self, query_columns=["video", "audio"], corpus_columns=["caption"])
+
+
+class VATEXT2VRetrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="VATEXT2VRetrieval",
```
Samoed:
```suggestion
        name="VATEXT2VARetrieval",
```

AdnanElAssadi56: yes, so we can have both VA2T, V2T and their symmetric cases whenever video and audio are both available. 

For A2T, T2A, I dont see them as necessary, but including them won't hurt.

manandey:
@Samoed Kindly take a look and let me know if it's good to be merged. Thanks!

Samoed (approved):

AdnanElAssadi56 (approved):

Issue #4376, opened 2026-04-14
Expose collators (VideoCollator, AudioCollator) as public API


AdnanElAssadi56:
Models currently import collators from mteb._create_dataloaders, which is a private module. This creates a hidden dependency: changes to the module could break model implementations without warning. Collators should be exposed through the public API.

Related: #4199

KennethEnevoldsen:
Should we refactor the collators into a seperate module outside of the `_create_dataloader`, so that they are not private (since they are used in models).

PR #4441, opened 2026-04-19
[MVEB] Add fps implementation to Video Sampling

AdnanElAssadi56:
- I added the ability to use fps for models rather than specifying a fixed number of frames all the time.


mteb/_create_dataloaders.py
```diff
@@ -479,48 +512,65 @@ def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
...
     Uses FramesCollator and AudioCollator static methods to process each modality.
+    Supports both FPS-based and fixed-sample video frame selection.
     """
 
     def __init__(
         self,
         target_sampling_rate: int,
-        max_frames: int = 16,
+        fps: float | None = 2.0,
+        max_frames: int | None = 256,
```
Samoed: I think we can set defaults as `None` and use processors if possible

mteb/_create_dataloaders.py
```diff
@@ -479,48 +512,65 @@ def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
...
-        )
+        num_source_frames = video.metadata.num_frames
+
+        if num_frames is not None:
+            # Fixed-sample mode: always select exactly num_frames
+            target = num_frames
+        else:
+            # FPS-based mode: scale with duration
+            source_fps = video.metadata.average_fps
+            duration = num_source_frames / source_fps
```
Samoed: You can take `end_stream_seconds` for this https://github.com/meta-pytorch/torchcodec/blob/598aaf65e2d6125af34484e1f9b9f5e38ca9a4ca/src/torchcodec/_core/_metadata.py#L133

mteb/_create_dataloaders.py
```diff
@@ -479,48 +512,65 @@ def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
...
+            target = max(1, int(duration * fps))
```
Samoed, line 544: `Duration * fps` could not match total_fps in cases when video could have different fps. Generally, I don't think we would have such cases. For this you can use `num_frames` https://github.com/meta-pytorch/torchcodec/blob/598aaf65e2d6125af34484e1f9b9f5e38ca9a4ca/src/torchcodec/_core/_metadata.py#L139

mteb/_create_dataloaders.py
```diff
@@ -452,21 +452,54 @@ def resample_audio(
...
+                scales with video duration. Ignored when ``num_frames`` is set.
+            max_frames: Safety cap on the number of frames when using FPS-based
+                sampling. Ignored when ``num_frames`` is set.
+            num_frames: If set, use fixed-sample mode: always select this many
+                frames uniformly from the video. Overrides ``fps``/``max_frames``.
         """
+        if num_frames is not None and fps is not None:
+            fps = None
+        if num_frames is None and fps is None:
+            raise ValueError("Either fps or num_frames must be provided.")
```
Samoed: I think we can just skip processing if they're none

mteb/_create_dataloaders.py
AdnanElAssadi56: For processors, the tensors will already need to be ready, this just changes the fixes 16 frames we had to fps.

AdnanElAssadi56: Or do you mean give processor all frames, and then it should do sampling; not sure if all processors do that or if they expect it to be done beforehand like what we had with process_mm_info.

mteb/_create_dataloaders.py
AdnanElAssadi56: i am worrying about oom; we can do what you said and then tweak from model side as well

AdnanElAssadi56: Basically, if model decides to keep the none default, it gets all frames and can do its own sampling (though this will require loading all frames in collator which will be memory heavy).

mteb/_create_dataloaders.py
Samoed: Yes, I mean to give all frames to processors, and they would do sampling inside them, because qwen3-vl processor will do video resampling with 2 fps

mteb/_create_dataloaders.py
```diff
@@ -452,21 +452,54 @@ def resample_audio(
...
         Args:
-            max_frames: The maximum number of frames to keep for each video.
+            fps: Target frames per second for sampling. The number of frames
+                scales with video duration. Ignored when ``num_frames`` is set.
+            max_frames: Safety cap on the number of frames when using FPS-based
+                sampling. Ignored when ``num_frames`` is set.
+            num_frames: If set, use fixed-sample mode: always select this many
+                frames uniformly from the video. Overrides ``fps``/``max_frames``.
         """
+        if num_frames is not None:
```
Samoed: I think we can raise an error if they're both not none

mteb/_create_dataloaders.py
```diff
@@ -479,48 +512,70 @@ def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
...
+            target = num_frames
+        elif fps is not None:
+            # FPS-based mode: scale with duration
+            duration = video.metadata.end_stream_seconds
+            target = max(1, int(duration * fps))
+            if max_frames is not None:
+                target = min(target, max_frames)
+        else:
+            # No resampling: return all frames
+            return video.get_frames_at(list(range(num_source_frames))).data
```
Samoed: It's better to check if both num_frames and fps are none and return this before this if

mteb/_create_dataloaders.py
AdnanElAssadi56: Yes, now models have the option to do that, but to save memory, we also have the option now to do fps 2 directly before even passing to the processor and passing `"do_sample_frames": False` to processors. This was we can avoid holding all frames in memory if we want.

mteb/_create_dataloaders.py
```diff
@@ -452,21 +452,54 @@ def resample_audio(
...
+    to handle.
+
+    When using FPS-based sampling, ``max_frames`` acts as an optional safety
+    cap to prevent OOM on very long videos.
+    """
+
+    def __init__(
+        self,
+        fps: float | None = None,
+        max_frames: int | None = None,
```
Samoed: I think we can rename a bit to show that it related to fps mode
```suggestion
        max_fps_frames: int | None = None,
```

mteb/_create_dataloaders.py
AdnanElAssadi56: why, if we decide to specify them for a specific model later on, we should be able to do that.

mteb/_create_dataloaders.py
AdnanElAssadi56: we're in the else clause so theyre both none

mteb/_create_dataloaders.py
Samoed: Because they're used by 2 different methods of sampling and shouldn't be used together. This is better than silently ignore `fps`

AdnanElAssadi56: Oh I see what you mean

mteb/_create_dataloaders.py
Samoed: Or probably we can just leave `num_frames` and use it during `fps` method as currently `max_frames` is used

AdnanElAssadi56: yes max frames is only used in fps method

AdnanElAssadi56: but for clarity maybe we should keep max_fps_frames

AdnanElAssadi56:
@Samoed I think that's all right?

Samoed:
Generally I think yes, but I think it would be better to wait review from @KennethEnevoldsen

mteb/_create_dataloaders.py
```diff
@@ -479,48 +515,70 @@ def __call__(self, inputs: list[dict[str, Any]]) -> BatchedInput:
...
         """Resample a video input to a target number of frames.
 
+        When both ``fps`` and ``num_frames`` are None, all frames are returned
+        for the model's processor to handle.
+
         Args:
             video: A VideoDecoder object containing the video data.
-            max_video_frames: The maximum number of frames to keep for each video. If None, no truncation is applied.
+            fps: Target frames per second. Used when ``num_frames`` is None.
+            max_fps_frames: Safety cap when using FPS-based sampling.
```
KennethEnevoldsen: This arg doesn't make immediate sense from reading it.

Isn't it just `max_video_frames`? If so I would keep it as is - that also makes it backward compatible.

max_fps_frames could be read as, "any number of FPS is fine, but if it is higher than 60 fps downsample to 60 fps". I see that it is applied as for FPS, but I would just fix that in the docstring.

mteb/_create_dataloaders.py
```diff
@@ -452,21 +452,57 @@ def resample_audio(
...
+    performed and all frames are returned as-is for the model's processor
+    to handle.
+
+    When using FPS-based sampling, ``max_fps_frames`` acts as an optional safety
+    cap to prevent OOM on very long videos.
+    """
+
+    def __init__(
+        self,
+        fps: float | None = None,
```
KennethEnevoldsen, line 471: this is really max_fps (not the actual fps). I would rename)

KennethEnevoldsen (commented):
I also think this is good to merge.


I assume this influences some model implementations as well?

Should we refactor the collators into a seperate module outside of the `_create_dataloader`, so that they are not private (since they are used in models).

Samoed:
> Should we refactor the collators into a seperate module outside of the _create_dataloader, so that they are not private (since they are used in models).

We have issue for this https://github.com/embeddings-benchmark/mteb/issues/4376

mteb/_create_dataloaders.py
Samoed: > Isn't it just max_video_frames?

What do you mean by `max_video_frames`? `num_frames`?

mteb/_create_dataloaders.py
Samoed: Transformers using similar as `fps` https://github.com/huggingface/transformers/blob/ef97a7525dc6e4665db4ba5e1ec40d79e43cc74e/src/transformers/models/qwen2_5_omni/processing_qwen2_5_omni.py#L165 but I think we can change this

AdnanElAssadi56: max_fps sounds confusing to me. People will only see difference of actual max_fps when they set the other other max_fps_frames arg so it should be clear.

AdnanElAssadi56:
How about:

1. We keep fps.
2. We keep max_fps_frames but change docstring to make it more clear.

mteb/_create_dataloaders.py
KennethEnevoldsen: isn't it essentially the argument that was there before? (if so I would keep it backward compatible)

mteb/_create_dataloaders.py
KennethEnevoldsen: Let us keep it as fps, but the docstring need to make it clear that we don't resample to the FPS (only downsample)

KennethEnevoldsen:
> We have issue for this https://github.com/embeddings-benchmark/mteb/issues/4376

added a comment

> We keep fps.

Agree, but change the docstring to make it clear (see comment above)

> We keep max_fps_frames but change docstring to make it more clear.

I would still rename to simply max_frames and update the docstring, but if you prefer max_fps_frames then I am happy with a better docstring as well

---

Double checked model implementations see that everything works.

I am happy with a merge without a review after this

AdnanElAssadi56:
@Samoed Do you agree with max_frames?

AdnanElAssadi56:
@Samoed I'll be merging this as we have mostly agreed here. We can open a follow up if you have any concerns.

PR #4451, opened 2026-04-19
Make corpus_chunk_size tunable via the public evaluate() API

### LINKED ISSUE #4450: Make `corpus_chunk_size` tunable for retrieval tasks
### Problem

In MTEB v1, `corpus_chunk_size` could be tuned via the public evaluation API (e.g. `evaluation.run(..., corpus_chunk_size=500)`), which was useful to reduce memory usage on large retrieval corpora.

In the current v2 codebase, the `SearchEncoderWrapper` class accepts `corpus_chunk_size` as a constructor parameter with a default of **50,000**, but this parameter is never passed through the public evaluation API. As a result, there is no way to change this chunk size when running via new evaluation API `mteb.evaluate(...)`.

This is also the case for Bitext mining tasks (`BitextMiningEvaluator`).

### Feature request

Could `corpus_chunk_size` be made tunable through the public evaluation API again?

Rakshitha-Ireddi:
Fixes #4450
             
  ## Problem
  `corpus_chunk_size` was hardcoded inside `SearchEncoderWrapper` (default 50,000) and
  `BitextMiningEvaluator._similarity_search` (default 500,000) with no way to override them from the public
  `mteb.evaluate()` API.                                                                                                
                        
  ## Changes                                                                                                            
  - Added `corpus_chunk_size: int | None = None` to `evaluate()` and `_evaluate_task()` in `evaluate.py`
  - Threaded it through to `SearchEncoderWrapper` in `abstasks/retrieval.py`                            
  - Added `corpus_chunk_size` to `BitextMiningEvaluator.__call__()` and passed to `_similarity_search()`                
  - Threaded it through `AbsTaskBitextMining._evaluate_subset()`                                                        
                                                                                                                        
  ## Usage                                                                                                              
  ```python                                                                                                             
  # Reduce memory on large retrieval corpora                                                                            
  mteb.evaluate(model, task, corpus_chunk_size=10_000)

### DIFF
--- a/docs/get_started/usage/running_the_evaluation.md
+++ b/docs/get_started/usage/running_the_evaluation.md
@@ -169,6 +169,21 @@ sentence_trf_model.half() # half precision
 A last option is to make a [custom implementation](defining_the_model.md#using-a-custom-model) of the model. This way you have full flexibility of how the model handles the input.
 
 
+### Reduce Memory During Evaluation
+
+For large retrieval corpora, MTEB encodes the corpus in chunks to avoid running out of memory. The default chunk size is 50,000 documents. You can lower it to reduce peak memory usage:
+
+```python
+import mteb
+
+model = mteb.get_model("sentence-transformers/all-MiniLM-L6-v2")
+task = mteb.get_task("NFCorpus")
+
+results = mteb.evaluate(model, task, encode_kwargs={"corpus_chunk_size": 10_000})
+```
+
+Passing a smaller value increases the number of encoding passes over the corpus but keeps GPU/CPU memory usage lower. This works for both retrieval tasks (via `SearchEncoderWrapper`) and bitext mining tasks (via `BitextMiningEvaluator`) and has no effect on other task types.
+
 ### Speeding Download
 
 The simplest way to speed up downloads is by using Huggingface's [`xet`](https://huggingface.co/blog/xet-on-the-hub). You can use this simply using:
--- a/mteb/_evaluators/text/bitext_mining_evaluator.py
+++ b/mteb/_evaluators/text/bitext_mining_evaluator.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, cast
 
 import torch
 from datasets import Dataset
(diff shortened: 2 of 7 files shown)


Samoed (changes requested):
During creation of `v2` we intentionally decided to not include `corpus_chunk_size` in `evaluate` to not complicate usage. I still don't think that we want to update evaluate with task/model specifc arguments

Rakshitha-Ireddi:
> During creation of `v2` we intentionally decided to not include `corpus_chunk_size` in `evaluate` to not complicate usage. I still don't think that we want to update evaluate with task/model specifc arguments

Thanks for the feedback! Would it make more sense to expose this via encode_kwargs instead (e.g.                   encode_kwargs={"corpus_chunk_size": 10_000})? That way evaluate() stays clean and users can still tune it when      
needed. Alternatively, should it be a task-level attribute like skip_first_result? @Samoed

Samoed:
`corpus_chunk_size` is model level argument and this is possible to change it with 
```python
SearchEncoderWrapper(model, corpus_chunk_size=512)
```
Method with `encode_kwargs` is better, but might cause problems with sentence transformers.

KennethEnevoldsen:
So as @Samoed mentions the intended usage is to re-wrap the model. Though currently it could be documented better.

I would be fine with adding it as an `encode_kwargs` and then popping it before passing it to the encoder in the `SearchEncoderWrapper`. It is after all an argument of the model and the `SearchEncoderWrapper` is "the model".

Another consideration might also be to reduce the defaults here. I at least typically find myself adjusting them down. I think that could solve many cases. 

regardless documenting it in the docs is probably a good start. I would suggest to add it here:
https://embeddings-benchmark.github.io/mteb/get_started/usage/running_the_evaluation/#speeding-up-evaluations

Under a section called, reduce memory during evaluation.

KennethEnevoldsen:
Of course, using it as an `encode_kwarg` does not solve the problem that it will result in an error if passed to a `SearchEncoderWrapper` model, e.g. for non-retrieval tasks.

Samoed:
I think that defaults are fine. When I run smaller models I increase them. This is hard to find values that would work better

Rakshitha-Ireddi:
Hi @Samoed and @KennethEnevoldsen, thank you for the feedback! I've addressed the review comments in a follow-up commit. Instead of adding corpus_chunk_size to evaluate(), it's now supported via encode_kwargs:                                                                                                                  
                                                                  
mteb.evaluate(model, task, encode_kwargs={"corpus_chunk_size": 10_000})                                                                                            
                                                                                                                                                                     
SearchEncoderWrapper and BitextMiningEvaluator both intercept and pop corpus_chunk_size from encode_kwargs before forwarding to the underlying encoder, so non-retrieval tasks are completely unaffected. Default values (50,000 for retrieval, 500,000 for bitext mining) are unchanged. Also added a "Reduce Memory During Evaluation" section to the docs as suggested. Let me know if anything needs adjustment!

KennethEnevoldsen (commented):
I would be fine with this change - though I would probably want a test to ensure that it isn't passed on to the model. You can find define a simple wrapper model somewhat like this:

```py

@pytest.mark.parametrize([mock_bitext, mock_retrieval])
def test_ensure_corpus_chunk_size_is_not_passed(task):
  # init from the "mteb/baseline-random-encoder" 
  class TestEncoder(RandomEncoderBaseline):
       def __init__(**kwargs):
          assert "corpus_chunk_size" not in kwargs
          super().__init__(**kwargs)
```


@Samoed WDYT?

Samoed:
Yes, agree

PR #4453, opened 2026-04-20
Add VALOR-32K retrieval tasks

manandey:
- [x] I have outlined why this dataset is filling an existing gap in `mteb`
- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [ ] `sentence-transformers/paraphrase-multilingual-MiniLM-L12-v2`
  - [ ] `intfloat/multilingual-e5-small`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] I have considered the size of the dataset and reduced it if it is too big (2048 examples is typically large enough for most tasks)

https://github.com/embeddings-benchmark/mteb/issues/4130
Ran the mteb/baseline-random-encoder model.

cc @Samoed @AdnanElAssadi56

PR #4462, opened 2026-04-20
[MVEB] Add NExT-QA video-centric QA task

yaswanth169:
Addresses part of #4130 (MVEB — Video Question Answering).

Adds NExT-QA as `NExTQAVideoCentricQA`, a multiple-choice video QA task with 993 examples, framed as retrieval over 5 candidates per query. Task type is `VideoCentricQA` (confirmed in #4130).

## Summary

This PR introduces the first Video Question Answering task under MVEB by:
- Enabling the `VideoCentricQA` task type
- Adding the NExT-QA dataset as a retrieval-style benchmark

## Changes

- **`task_metadata.py`**
  - Uncomment `VideoCentricQA` in `MVEB_TASK_TYPE`
  - Add mapping in `_TASKTYPE2SIMPLIFIEDTASKTYPE`
  - Add HF dataset category mapping → `"visual-question-answering"`

- **New task: `NExTQAVideoCentricQA`**
  - Located at `mteb/tasks/multichoice/eng/nextqa.py`
  - Follows `AbsTaskRetrieval` pattern (similar to `cv_bench.py`)
  - Reformulates multiple-choice QA as retrieval over 5 answer candidates

- **Task registration**
  - Added import and entry in `__all__` in `multichoice/eng/__init__.py`

## Implementation Notes

- Uses path-based video loading (`{"path": ..., "bytes": None}`) cast to `Video(decode=True)` to avoid Arrow 2GB Binary column overflow (~3GB dataset size)
- Designed to be compatible with existing MVEB task patterns

## Known Issues (Draft)

1. **Arrow offset overflow**
   - `create_dataloaders.py` uses `.map()` which can overflow for large binary video columns
   - Handled locally in this task via path-based representation

2. **VideoCollator not wired**
   - `_video_to_vector` expects `VideoInputItem`, but current collate path passes raw decoder output
   - Leads to:
     ```
     TypeError: Unsupported key type: <class 'str'>
     ```
   - Can be addressed in a follow-up PR

## Checklist

- [x] Introduces first `VideoCentricQA` task in MVEB
- [x] Task loads and runs (smoke tested)
- [x] 993 evaluation examples (no truncation)
- [x] License: cc-by-4.0
- [ ] `results.json` (blocked on VideoCollator issue)


mteb/tasks/multichoice/eng/nextqa.py
```diff
@@ -0,0 +1,124 @@
...
+        ds = load_dataset(path, revision=revision, split=split)
+        ds_raw_video = ds.cast_column("video", Video(decode=False))
+
+        corpus_rows: list[dict] = []
+        query_rows: list[dict] = []
+        relevant_docs[split] = {}
+
+        for idx, row in enumerate(ds_raw_video):
+            qid = f"q{idx}"
+            video_path = _extract_video(row["video"], qid)
```
Samoed: I think problem is that you're recreating dataset object here. You can avoid this by selecting required columns for processing. This should resolve your problems. You can look in https://github.com/embeddings-benchmark/mteb/pull/4453 with similar processing

mteb/tasks/multichoice/eng/nextqa.py
```diff
@@ -0,0 +1,92 @@
...
+  title = {NExT-QA: Next Phase of Question-Answering to Explaining Temporal Actions},
+  year = {2021},
+}
+""",
+    )
+
+    def load_data(self, **kwargs) -> None:
+        if self.data_loaded:
+            return
+        self.corpus, self.queries, self.relevant_docs, self.top_ranked = _load_data(
```
Samoed: Update format to use `self.dataset["default"]["test"] = RetrievalSplitData`

Samoed:
Did you solve your problem?

> _video_to_vector expects VideoInputItem, but current collate path passes raw decoder output

We don't have anymore `VideoInputItem`

yaswanth169:
> Did you solve your problem?
> 
> > _video_to_vector expects VideoInputItem, but current collate path passes raw decoder output
> 
> We don't have anymore `VideoInputItem`

want me to draft the wording?

Samoed:
No, just FYI. Have you successfully run a task?

yaswanth169:
No, not yet . structural smoke passes, but I haven't run the full eval.

yaswanth169:
I attempted the end to end evaluation but got blocked by an upstream issue and wanted your direction.

The error is pyarrow.lib.ArrowInvalid: offset overflow while concatenating arrays in mteb/_create_dataloaders.py:275 during dataset.map(_combine_queries_with_instruction_text) on the queries.

The issue comes from our query dataset containing id, text, and video. With the default writer_batch_size=1000, the embedded video bytes are around 3 MB each, which exceeds Arrow’s 2³¹ binary offset limit. Even though the map only modifies text, Arrow rewrites all columns, which triggers the failure.

Other video tasks do not hit this because they use single modality queries, either video only or text only. This vt2t setup with both video and text in the query is the first case that exposes this limitation.

I see two possible paths forward. One is to fix this upstream by setting a smaller writer_batch_size such as 10 in that map call, which would unblock future vt2t tasks as well. I can open a PR for this. The other option is a local workaround where we pre extract video bytes to a cache directory and replace the video column with file paths wrapped in Video(decode=True). This keeps everything lazy and works with the current VideoCollator.

Samoed:
I think we can update `dataset.map(_combine_queries_with_instruction_text)` to remove map usage

yaswanth169:
Implemented the fix replaced dataset.map(_combine_queries_with_instruction_text) with direct add_column/remove_columns column ops in _create_dataloaders.py.  Full end-to-end eval now completes cleanly. Random baseline accuracy: 18.6% (expected ~20% for 5-choice MC). Results JSON attached.

yaswanth169:
[NExTQAVideoCentricQA.json](https://github.com/user-attachments/files/27007608/NExTQAVideoCentricQA.json)

AdnanElAssadi:
Hi!
Is there anything blocking here?

Samoed:
I'll need to fix code a bit. Though I've fixed everything

AdnanElAssadi:
@yaswanth169 do you think you have enough bandwidth to implement the rest of the datasets?

yaswanth169:
Hello @AdnanElAssadi56,
Yes I split some of them with @Rakshitha-Ireddi as well.

AdnanElAssadi:
> Hello @AdnanElAssadi56, Yes I split some of them with @Rakshitha-Ireddi as well.

Sounds good! Now that you have na example here, will you be able to open the PRs for the rest by EoD? Sorry for pressing this, but we are tight on time, and this is an important task that is blocking evaluation.

AdnanElAssadi:
@Samoed Does PR here need more work?

yaswanth169:
@AdnanElAssadi56,
We can work on it and try to complete majority of them. Before we need to get clarity from @Samoed that current PR is merging or not because remaining all depends on this itself.

Samoed:
No, I think this is good to merge

yaswanth169:
Then good now we will work on this and complete by the end of the day or by morning where remaining steps can be started as expected.

AdnanElAssadi56 (approved):

PR #4464, opened 2026-04-20
model: Adding prompts to voyage-4-large 2048d model

fzowl:

### DIFF
--- a/mteb/models/model_implementations/voyage_models.py
+++ b/mteb/models/model_implementations/voyage_models.py
@@ -196,6 +196,7 @@ class VoyageModel(AbsEncoder):
         max_tokens: int | None = None,
         model_prompts: dict[str, str] | None = None,
         output_dtype: str | None = None,
+        api_model_name: str | None = None,
         **kwargs,
     ) -> None:
         import voyageai
@@ -203,7 +204,10 @@ class VoyageModel(AbsEncoder):
         self._client = voyageai.Client(max_retries=max_retries)
         self._embed_func = rate_limit(max_rpm)(token_limit(max_tpm)(self._client.embed))
 
-        self._model_name = model_name.split("/")[-1].split()[0]  # noqa: PLC0207
+        if api_model_name:
+            self._model_name = api_model_name
+        else:
+            self._model_name = model_name.split("/")[-1].split()[0]
         self._max_tpm = max_tpm
         self._max_tokens = max_tokens
         self.model_prompts = self.validate_task_to_prompt_name(model_prompts)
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/voyage_models.py
```diff
@@ -294,11 +298,111 @@ def _batched_encode(
...
+    "HumanEvalRetrieval": "Python solution: {text}",
+    "LegalQuAD": "Relevant statute: {text}",
+    "LegalSummarization": "{text}",
+    "MBPPRetrieval": "Python function: {text}",
+    "MIRACLRetrievalHardNegatives": "Reference passage: {text}",
+    "WikiSQLRetrieval": "SQL statement execution: {text}",
+}
+
+
+class VoyageEvolvedPromptsModel(VoyageModel):
```
KennethEnevoldsen, line 344: Do you have any description anywhere of how the prompts where evolved/tuned?

mteb/models/model_implementations/voyage_models.py
```diff
@@ -294,11 +298,111 @@ def _batched_encode(
...
+        return self._batched_encode(sentences, batch_size, input_type)
+
+
 model_prompts = {
     PromptType.query.value: "query",
     PromptType.document.value: "document",
 }
 
+voyage_4_large_2048d_evolved = ModelMeta(
+    name="voyageai/voyage-4-large-2048d-evolved",
```
KennethEnevoldsen, line 380: Since this is basically an [experiment](https://embeddings-benchmark.github.io/mteb/whats_new/#experiments) I would name it as such:

```suggestion
    name="voyageai/voyage-4-large-2048d (evolved_prompts=True)",
```

KennethEnevoldsen (commented):
Hmm interesting. I have actually been looking into this (to what extent it is possible to tune the prompt to a benchmark) as a way to figure out how we best ensure fair comparison for across MTEB. 

I am not entirely sure what the best approach is here (@Samoed, @tomaarsen, @isaac-chung, @AdnanElAssadi56, would love your thoughts on this) 

@fzowl can I ask you to add a comparison table of the scores and thanks for making it explicit in the PR that it was evolved prompts.

mteb/models/model_implementations/voyage_models.py
Samoed: Can we create a pr with "presets" to use like this? I don't like that we create solutions like this for voyage

KennethEnevoldsen: I might agree with you but I think it kinda depends - let us take it once we have a bit more info e.g. of how the prompts were evolved

fzowl: @KennethEnevoldsen @Samoed The story behind this PR is that while I was working on some other PRs, I just realised that there are models that use prompts. This gave me the idea to try a few prompts for the RTEB datasets (except the large ones). So, i tried 2-3 prompts, then picked the best one, tried 2-3 similar prompts again. I did 3 iterations and used the best prompt to evaluate the datasets.

@Samoed, may I ask exactly what you don't like about the prompting? I mean, there are other models that use prompts. I believe -if we keep fair play/comparison in mind- there are 2 choices:
 - allow any model to apply prompts
 - do not allow prompts, remove the models+results with prompts

I don't like prompts myself. On the other hand, it seems like prompts help models to perform better, so in this regard, prompts are great! What's difficult is that the same prompt performs differently with different models, so we can't pre-define prompts per dataset at the application level - different models would benefit differently from the same prompt.

What do you think?

Please find attached the difference between the voyage-4-large prompted and non prompted version result. 
Screenshot 2026-04-23 at 14 12 15

KennethEnevoldsen: @fzowl I think you might be confusing @Samoed comment (I think he refers to how to best implement it, not the prompts themselves). 

On the policy, though, I would say the benchmark on a general level should reflect the kinda performance the user might see. 

Of course, the user could also tune the prompt; actually, the 3 attempts seem about the degree to which one might tune it; however, it will of course not reflect just using any prompt. I don't want to claim we currently have a policy on this, but so far, whenever I have asked on PRs, people have said they either wrote them from scratch or used a template, typically because the default prompt didn't match what they used for training. I would be surprised if no one has tuned their prompts, though.

One could also test the model's reliability and resilience by varying prompts.

tomaarsen: Personally, I think prompts are great, although often a bit confusing for users. However, the idea of iteratively updating a prompt until reaching the highest possible value on a benchmark just feels like overfitting. But at the same time, it's realistic for the user to also do this, and perhaps even useful if it results in prompts that work well with a task/model, then the user can adopt this and get better performance of their own.

Samoed: Let me clarify a bit. I'm not against prompts. I just don't like that we're creating another model with `(evolved_prompts=True)` suffix. I think we can add "presets" to make possible for user to get a model with predefined configuration and run it as an "experiment"

AdnanElAssadi: Problem here could be fairness. There are probably some submissions that have tuned their prompts and others not, so I guess every model should get its prompts optimized under a standardized protocol because we don't know which models already tuned theirs. This could be some standard prompt optimization at eval time (e.g. something like how DSPy on held-out data works, with a fixed budget per model). I'm not sure how cleanly this fits into the current eval harness, but it feels like the right direction.

KennethEnevoldsen: that was also what I expected @Samoed - so my suggestion is to implement it in the base model and then have an argument like `tuned_prompt=True`, which switched the prompt set

KennethEnevoldsen: I agree with your point @AdnanElAssadi56, but we probably need some testing of this before we implement something. For now I think disclosure in the model name is a good start, but we probably need a long term solution for it.

Samoed: I think such prompt optimization is a good research project

fzowl: As @KennethEnevoldsen noted, it would be surprising if no one else has tuned their prompts — the difference here is we're explicitly disclosing it via (evolved_prompts=True). Rejecting this while allowing undisclosed prompt tuning would penalise transparency. 
   
The tuning was modest: 3 manual iterations of 2-3 prompts per task, which is the kind of light exploration any practitioner does in production. In fact, prompt optimisation can be a good practice in real-world scenarios — when users know what type of data they're working with, tuning the prompt to their domain is exactly what they should do. As @tomaarsen put it, this is "perhaps even useful if it results in prompts that work well with a task/model, then the user can adopt this and get better performance of their own."

No one in the discussion objected to prompts in principle — @Samoed's concern was about implementation pattern (presets vs suffix), not policy. Happy to refactor toward the presets system or the tuned_prompt=True base model argument — either works as a follow-up without blocking the results.

KennethEnevoldsen:
Thanks for the ping @fzowl!

I would be fine with merging this PR - we have some experiments on prompt overfitting that I would love to share, once we have finalized them 

@Samoed what would resolve your concerns on the implementation pattern? Should we consider it an experiment instead? (evolved=True as a kwargs on the main class?)

Samoed:
> what would resolve your concerns on the implementation pattern? Should we consider it an experiment instead? (evolved=True as a kwargs on the main class?)

I think yes

KennethEnevoldsen:
@fzowl if you make a PR with the implementation, then we can finalize this PR and merge

KennethEnevoldsen (approved):

mteb/models/model_implementations/voyage_models.py
```diff
@@ -294,11 +309,82 @@ def _batched_encode(
...
+    similarity_fn_name="cosine",
+    framework=["API"],
+    use_instructions=True,
+    training_datasets=VOYAGE_TRAINING_DATA,
+    public_training_code=None,
+    public_training_data=None,
+    output_dtypes=OUTPUT_TYPES,
+    extra_requirements_groups=["voyageai"],
+    adapted_from="voyageai/voyage-4-large (embed_dim=2048)",
+    experiment_kwargs={"evolved_prompts": True},
```
KennethEnevoldsen: @Samoed I think this is good to merge, but maybe we should set this to None?
```suggestion
    experiment_kwargs=None,
```

KennethEnevoldsen (approved):

mteb/models/model_implementations/voyage_models.py
Samoed: We shoudn't create a new ModelMeta for this. Experiment will be activated with `mteb.get_model("voyageai/voyage-4-large (embed_dim=2048)", evolved_prompts=True)`. And evolved prompts should be added in `__init__`

KennethEnevoldsen: Right, but it wouldn't show up in the leaderboard then - I don't believe we currently show experiments)

Samoed: Yes, we don't show https://github.com/embeddings-benchmark/mteb/issues/1211

KennethEnevoldsen: I would probably for now then just remove `experiment_kwargs` and then merge (later once we have experiment set up we can refactor).

fzoll: @KennethEnevoldsen @Samoed Hm. But other models with prompts are already on the leaderboard. Those models do not advertise on the leaderboard whether they use prompts.

KennethEnevoldsen: I think you are misunderstanding @fzoll. Merging this would make it appear on the leaderboard

KennethEnevoldsen (approved):

Issue #4411, opened 2026-04-16
Improve the speed of the leaderboard 


akvachan:
### Description of the feature

The leaderboard is extremely slow.

This is just not acceptable. You guys seem to care abour performance, yet ship a product that takes 20 seconds to load a simple table. Like what's going on? And yeah, I am going to complain about it and be ass about it because the performance bar has been incresing lately exactly because of fuck ups like this. Good idea, nice project, but execution is extremely poor.

akvachan:
Kudos to frontend devs for an amazing job! You rock it guys! Here is a gem of code:

(sorting is done twice everywhere)

```sh
lang_select = gr.CheckboxGroup(
        sorted(default_results.languages),
        value=sorted(default_results.languages),
        show_label=True,
        show_select_all=True,
        label="Language",
        info="Select languages to include.",
    )
    type_select = gr.CheckboxGroup(
        sorted(default_results.task_types),
        value=sorted(default_results.task_types),
        show_label=True,
        show_select_all=True,
        label="Task Type",
        info="Select task types to include.",
    )
    domain_select = gr.CheckboxGroup(
        sorted(default_results.domains),
        value=sorted(default_results.domains),
        show_label=True,
        show_select_all=True,
        label="Domain",
        info="Select domains to include.",
    )
    task_select = gr.CheckboxGroup(
        sorted(default_results.task_names),
        value=sorted(default_results.task_names),
        show_label=True,
        show_select_all=True,
        label="Task",
        info="Select specific tasks to include",
    )
    modality_select = gr.CheckboxGroup(
        sorted(default_results.modalities),
        value=sorted(default_results.modalities),
        show_label=True,
        show_select_all=True,
        label="Modality",
        info="Select modalities to include.",
    )
```

tomaarsen:
Hello!

Please edit your messages to fix your tone, it is not welcome in open source. 
Beyond that, well spotted, would you like to open a pull request?

- Tom Aarsen

KennethEnevoldsen:
Hi @akvachan, 

Let me just first address the tone, as Tom says this is completely unacceptable tone especially for a free product that is developed by open source maintainers that contribute their time and knowledge free of charge. We are more than happy to take constructive criticism, but please keep a respectful and professional tone.

Now for the issue at hand. I agree that the leaderboard could be a lot faster and if you have a fix, we are more than happy to welcome a PR. 

[...]

PR #4467, opened 2026-04-21
[`leaderboard experiment`] cache and vectorize filter-driven table updates

tomaarsen:
Hello!

Edit: In hindsight, the leaderboard timings that I experience locally are very different from the ones on the Web:
1. The web its initial load is a lot slower for me
2. My local its filtering is a lot slower for me

So that might invalidate the experiments here somewhat. Back to the original PR text:

Earlier today, I was experimenting with letting Claude Code propose optimizations to the leaderboard update methods, based on how slow each of the various methods are. Initially, loading MTEB (Mult, v2) and then filtering by model size (e.g. <1B) took ~270 seconds locally, whereas it now takes ~40s on the first run and ~20s on every run afterwards. 

The changes are primarily caching, but also some vectorization improvements. I think you can also go a lot further, but you start to degrade the code quality more and more. For example, the current changes already hurt the simplicity of `BenchmarkResults.to_dataframe` considerably in favor of large gains. As a result, I pitch this PR more as an experiment in changes that can be made and a conversation surrounding the balance of efficiency and simplicity.

Personally, I think the question is mainly:
1. Do we care that the `BenchmarkResults` is complicated with caching for Leaderboard purposes: If no, then we can use extensive caching here.
3. If yes for the above, should we then instead move the caching outside of the core MTEB functionality, and into some Leaderboard-specific file(s)?

For context, here's the AI-generated PR description to explain what it changed:

## Pull Request overview
* Speed up leaderboard filter interactions from ~270s to ~20s on `MTEB(Multilingual, v2)`
* Cache `BenchmarkResults.to_dataframe` per instance, with parent-scope propagation for filtered children
* Vectorise borda + per-type means, memoise `get_tasks` at hot call sites

## Details
Filter interactions on the leaderboard were taking ~270s per click on `MTEB(Multilingual, v2)` (548 models × 131 tasks × ~1k languages), which made the "Customize this Benchmark" UI effectively unusable. I profiled each stage and the cost was spread across four hot paths rather than concentrated in one, so I went through them one by one.

On the leaderboard side, `mteb.get_tasks(tasks=..., languages=[...])` was being called on every filter change (42s for the multilingual case) and `select_tasks` then spent another 25s filtering hf_subsets. When the user hasn't narrowed languages, the filter is a no-op — `_filter_benchmark_results_for_tables` now computes `_all_langs_for_tasks(task_names)` directly from `_TASKS_REGISTRY[name].metadata` (no task instantiation) and short-circuits both calls when the selection already covers every task language. The remaining `get_tasks` / `get_task` call sites are wrapped with `functools.lru_cache`.

On the dataframe side, `BenchmarkResults.to_dataframe` was being called three times per filter change (summary, per-task, per-language), each re-running `join_revisions` + `_get_score_for_table` + DataFrame construction on the full filtered result. I added an instance-level `_df_cache` (pydantic `PrivateAttr`) with three layers: (1) per-key aggregated result, (2) pre-aggregation long df shared across `aggregation_level`s, (3) a parent pointer so filtered children isin()-filter the unfiltered parent's cached df instead of rebuilding from `model_results`. `_filter_benchmark_results_for_tables` wires the filtered result back to its parent and records the filter scope so the parent cache is reused transparently. The pre-agg df uses categorical dtypes on `model_name` / `task_name` / `split` / `subset` to cut memory ~4× (~3.5GB → ~900MB for the multilingual case); `_aggregate_and_pivot` now passes `observed=True` so filtered categoricals don't materialise empty groups. Startup precomputes the task-level long df per benchmark and evicts the larger pre-agg df, so only the per-language tab pays that cost and only on the first visit per benchmark.

While I was in there: `_get_means_per_types` was doing `per_task.iterrows()` inside a nested loop over task types; rewritten as one vectorised mean per type + `melt`. `_get_borda_rank` was using `DataFrame.apply(_borda_count, axis="index")` which loops columns in Python; `DataFrame.rank(axis=0)` is the same computation, ~6× faster on realistic sizes.

Net effect on `MTEB(Multilingual, v2)` steady-state:

| Stage            | Before   | After    |
|------------------|----------|----------|
| `filter_results` | 68.3s    | 0.4s     |
| summary          | 146.7s   | ~5s      |
| per_task         | 26.1s    | ~3s      |
| per_language     | 29.6s    | ~9s      |
| **total**        | **270s** | **~20s** |

The first per-language interaction per benchmark still pays ~20s to build the parent's pre-agg df lazily; every subsequent interaction reuses it. Startup is ~30–60s longer due to precomputing task-level dfs for all 55 leaderboard benchmarks, which felt like the right trade to make user interactions feel instant.

---

This is the end of the AI-generated PR description, and it's me again. I think we need to be careful to avoid the MTEB Leaderboard getting too slow, or it stops being a viable and convenient tool for the community, especially as there are more datasets, more models, etc. With this PR, I think we can see that there's a lot of savings possible.

- Tom Aarsen


KennethEnevoldsen (commented):
thanks for this PR. 

I think the answer to 1) is yes and 2) is also yes. I will try to do a refactor on this.

What did you use for benchmarking?

tomaarsen:
> What did you use for benchmarking?

Tracking the timing of `update_tables` when loading the leaderboard locally, and then selecting the `<1B`  filtering option. Locally, that would take between 200 and 350 seconds for me initially, and down to 20-40 seconds after the various caching changes.

It was mostly based on my experiences from a month or 2 ago, when filtering would take a long time. Nowadays, even loading the page or changing benchmarks takes longer than ideal.

- Tom Aarsen

Samoed:
Do you have code to get such values?

tomaarsen:
Just a `start_time = time.time()` and `time.time() - start_time` at the start and end of the function. Not very advanced, but works well.

- Tom Aarsen

KennethEnevoldsen:
I added a bit of a leaderboard_bench just to have a consistent script to compare with. 

I also added back the caching on the benchmark, but made a wrapper to contain it (also so that user of mteb are not required to use a ton of memory without knowing it). Added caching for a few more hot spots (zero-shot percentage).

Overall I don't think current setup is ideal (also discussed in #4161), but I have set up a meeting with frontend guy from nvidia at the start of next week where I will at least bring up the idea of doing more of a major overhaul.

So the gains are as follows (from main)
  - Warm benchmark switch (multilingual): 120 s → 30 s (−75%) — the parent-link backfire eliminated              
  - Cold benchmark switch: 62 s → 39 s (−38%) — columnar df construction                                         
  - English filter scenarios: −7% to −27% across the board                                                       
  - Default table render: 925 ms → 818 ms (−12%)         
  
after that I also added: I also added so that we don't sort twice as suggest by this isssue: #4411 (though it added almost nothing)

Lastly noticed that the summary table contain multiple models with no score in any of the columns (currently we keep models with a score for at least one of the tasks)

Tested the leaderboard as well. Not perfect, but def. waaay better.

@tomaarsen would love the review if you have the time

Samoed (approved):
Looks great! Hope this would work same on spaces as locally. I wanted to work on this also, but couldn't find time

Samoed:
> I have set up a meeting with frontend guy from nvidia at the start of next week where I will at least bring up the idea of doing more of a major overhaul.

Can I join? Would be great to discuss together

KennethEnevoldsen:
> Can I join? Would be great to discuss together

Sure we are still finalizing the time slot, will add you once I know

PR #4456, opened 2026-04-20
[MVEB] Add video statistics

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4378

Example format:
```python
class VideoStatistics(TypedDict):
    """Class for descriptive statistics for video.
    Attributes:
        total_duration_seconds: Total duration of all video clips in seconds
        total_frames: Total number of frames across all video clips
        min_width: Minimum width of video frames
        average_width: Average width of video frames
        max_width: Maximum width of video frames
        min_height: Minimum height of video frames
        average_height: Average height of video frames
        max_height: Maximum height of video frames
        min_duration_seconds: Minimum duration of a video clip in seconds
        average_duration_seconds: Average duration of a video clip in seconds
        max_duration_seconds: Maximum duration of a video clip in seconds
        unique_videos: Number of unique video clips
        average_fps: Average frames per second across all video clips
        fps: Dict of unique (rounded) fps values and their frequencies
    """

    total_duration_seconds: float
    total_frames: int

    min_width: int
    average_width: float
    max_width: int

    min_height: int
    average_height: float
    max_height: int

    min_duration_seconds: float
    average_duration_seconds: float
    max_duration_seconds: float

    unique_videos: int

    average_fps: float
    fps: dict[int, int]
```

PR #4470, opened 2026-04-21
fix: pass prompt_type to create_dataloader in pair classification evaluator

stef41:
The pair classification evaluator never passes `prompt_type` to `create_dataloader()`. For text-only tasks this is fine because both sides get the same treatment, but for cross-modal tasks (e.g. v2t where input1=video, input2=text) the modality routing in `get_modalities()` is skipped entirely — both sides are treated as the same modality, which crashes.

Two fixes here:

**`pair_classification_evaluator.py`**:
- Actually pass `input1_prompt_type` / `input2_prompt_type` through to `create_dataloader()`
- Fix `select_columns` to keep `id` and `title` columns when present (otherwise `_corpus_to_dict` blows up with `KeyError`)

**`_create_dataloaders.py`**:
- Drop the `prompt_type is None` guard on the text fast-path. Without this, text-only tasks that set a `PromptType` (like `TERRaV2`) fall through to `_corpus_to_dict` instead of `_create_dataloader_from_texts`, which is wrong.

Tested: 19/19 evaluator tests pass, TERRaV2 still works, cross-modal VideoPairClassification tasks run correctly.

Ref: #4130

### DIFF
--- a/mteb/_create_dataloaders.py
+++ b/mteb/_create_dataloaders.py
@@ -334,11 +334,7 @@ def create_dataloader(
     # Sequence means columns already match modality names, no renaming needed
     _input_column = input_column if isinstance(input_column, str) else None
 
-    if (
-        prompt_type is None
-        and task_metadata.modalities == ["text"]
-        and _input_column is not None
-    ):
+    if task_metadata.modalities == ["text"] and _input_column is not None:
         return _create_dataloader_from_texts(
             dataset[_input_column],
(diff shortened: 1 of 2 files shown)


Samoed:
Thanks for rising! I'll make some changes to this and problem would be solved in https://github.com/embeddings-benchmark/mteb/pull/4456 with multimodal dataset fixes

PR #4482, opened 2026-04-22
[MVEB] Move collators to separate file

### LINKED ISSUE #4376: Expose collators (VideoCollator, AudioCollator) as public API

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4376

FYI @AdnanElAssadi56


mteb/models/modality_collators.py
```diff
@@ -0,0 +1,277 @@
...
+    from mteb.types import BatchedInput
+
+logger = logging.getLogger(__name__)
+
+
+class AudioCollator:
+    """Collator for audio data that resamples audio to a target sampling rate and optionally truncates to a maximum number of samples."""
+
+    def __init__(
+        self,
```
KennethEnevoldsen, line 23: Do we want to make all or some arguments keyword args to ensure that it is easier to maintain going forward?

KennethEnevoldsen: I would say almot everything could be keyword args

KennethEnevoldsen:
```suggestion
        self,
        *,
```

KennethEnevoldsen (approved):
Only a minor comment - didn't review functionality (as the code was just moved)

mteb/models/modality_collators.py
Samoed: Yes, agree. I'll change that

Issue #2789, opened 2025-06-08
Floating-point non-determinism in GPU operations


isaac-chung:
Wherever we import torch. [Reference](https://discuss.pytorch.org/t/effect-of-torch-backends-cudnn-deterministic-true/113095).

```python

import torch
 
torch.manual_seed(seed)
torch.backends.cudnn.deterministic = True
torch.backends.cudnn.benchmark = False

```

Samoed:
I think [`transformers.set_seed`](https://github.com/huggingface/transformers/blob/f4fc42216cd56ab6b68270bf80d811614d8d59e4/src/transformers/trainer_utils.py#L93) cover this case

KennethEnevoldsen:
Hmm, how does this work if someone sets the random seed?

Could I e.g. seed hack the benchmark using:
```
import mteb

# set seed
```

Is there anywhere in the repo where we use random number generators from torch? (I only see a few places in tests)

There are ways to prevent this in numpy, some of them, as well as arguments for not using global seeds, are mentioned [here](https://builtin.com/data-science/numpy-random-seed)

Samoed:
> Is there anywhere in the repo where we use random number generators from torch?

I don't think so, but in `v2` I've added `transformers.set_seed` too https://github.com/embeddings-benchmark/mteb/blob/9c63639f76f45bd64ace7ccc34809ce40648e115/mteb/abstasks/AbsTask.py#L30

KennethEnevoldsen:
So we can close this?

isaac-chung:
Added to v2 issue: I'd like to use to remind us to actually test it out in v2, e.g. run 3-5 runs of the same small collection of representative datasets & models, if that's alright?

KennethEnevoldsen:
Perfectly fine with me (we need some validation of v2 anyway)

KennethEnevoldsen:
This essentially just needs to be tested by 
> run[ning] 3-5 runs of the same small collection of representative datasets & models

PR #4492, opened 2026-04-23
Fix floating-point non-determinism in GPU operations

Jah-yee:
Good day

This PR fixes floating-point non-determinism in GPU operations (issue #2789).

## Problem
GPU operations in PyTorch can produce non-deterministic results across runs due to non-deterministic cuDNN algorithms. This causes tests to be flaky and benchmarks to be unreproducible.

## Fix
At the torch import site (mteb/__init__.py), set:
- torch.backends.cudnn.deterministic = True
- torch.backends.cudnn.benchmark = False

This ensures that cuDNN uses deterministic algorithms, making GPU operations reproducible across runs.

## Changes
- Added torch import and cudnn configuration in mteb/__init__.py, right after version import and before any other mteb imports.

Thank you for your work on this project. I hope this small fix is helpful. Please let me know if there's anything to adjust。

Warmly, RoomWithRoof

### DIFF
--- a/mteb/__init__.py
+++ b/mteb/__init__.py
@@ -1,5 +1,11 @@
 from importlib.metadata import version
 
+import torch
+
+# Set GPU deterministic mode for reproducibility
+torch.backends.cudnn.deterministic = True
+torch.backends.cudnn.benchmark = False
+
 from mteb import types
 from mteb.abstasks import AbsTask
 from mteb.abstasks.task_metadata import TaskMetadata


Samoed (commented):
Thank you for addition! But I'm not sure if we want to add it to `__init__`

KennethEnevoldsen:
Hi @Jah-yee, and thanks for the PR.

I also don't believe we want it here. It is quite a large unintended side effect of importing MTEB. So far our results are mostly reproducible, so I am not sure this statements counts:

> This causes tests to be flaky and benchmarks to be unreproducible.

We accept a certain degree of noise (0.001), e.g. when changing a model implementation so we don't need perfect reproducibility, we simply need it to be good enough. 

I am unsure if we have a documented issue that warrant the performance drop. From the pytorch [docs](https://docs.pytorch.org/docs/stable/notes/randomness.html):

> Deterministic operations are often slower than nondeterministic operations, so single-run performance may decrease for your model. However, determinism may save time in development by facilitating experimentation, debugging, and regression testing.

@Jah-yee if you have experienced an issue, we would love to hear about it.

I would imagine if we want to implement this we want to add it in the `mteb.evaluate` and revert once we are done with the evaluation.

related: Seed is currently set here: https://github.com/embeddings-benchmark/mteb/blob/029457877491698717b31dd59bf15c0e700bea4c/mteb/_set_seed.py#L15-L16

KennethEnevoldsen (commented):
(see above)

Jah-yee:
@KennethEnevoldsen @Samoed  Thank you both for the review and maintenance.
I'll close this PR with this message.

PR #4516, opened 2026-04-26
ci: add .venv caching step to speed up dep install

isaac-chung:
Currently a simple test like linting takes 14 minutes to run, in which 13 minutes were used to install dependencies: e.g. https://github.com/embeddings-benchmark/mteb/actions/runs/24953209689/job/73066896371?pr=4515

Fix: use `hashFiles('uv.lock')` and other params to form a cache key to cache `.venv` for applicable workflows.

The first CI run will take the same amount of time as before. The second run (given no changes to uv.lock) should be much faster with cache read.

<details>

  Workflow: dataset_loading                    
  Cache key suffix: py3.11-install
  Notes: make install                                                                        
  ────────────────────────────────────────
  Workflow: documentation                                                                    
  Cache key suffix: py3.10-docs                                                            
  Notes: uv sync --group docs
  ────────────────────────────────────────
  Workflow: lint
  Cache key suffix: py3.10-install
  Notes: make install
  ────────────────────────────────────────
  Workflow: model_loading
  Cache key suffix: py3.10-model-loading
  Notes: uv sync --extra pylate --group dev
  ────────────────────────────────────────
  Workflow: test
  Cache key suffix: py${{ matrix.python-version }}-install-for-tests
  Notes: Matrix across 3.10-3.14 + Windows
  ────────────────────────────────────────
  Workflow: typechecking
  Cache key suffix: py3.10-install-for-tests
  Notes: Shares cache with test py3.10/ubuntu
  ────────────────────────────────────────
  Workflow: reference_models
  Cache key suffix: py3.11-install-for-tests
  Notes: Shares cache with test py3.11/ubuntu
  ────────────────────────────────────────
  Workflow: leaderboard_build
  Cache key suffix: py3.10-leaderboard
  Notes: uv run --group test --extra leaderboard
  ────────────────────────────────────────
  Workflow: update_leaderboard_models
  Cache key suffix: py3.10-sync
  Notes: Bare uv sync

</details>

  Skipped: test_lowest (uses --resolution lowest-direct which deliberately ignores the
  lockfile — caching would defeat its purpose), release, leaderboard_docker,
  leaderboard_refresh, leaderboard_healthcheck, hf_space_docker, stale_pr (no Python deps
  installed on runner).

### DIFF
--- a/.github/workflows/dataset_loading.yml
+++ b/.github/workflows/dataset_loading.yml
@@ -44,6 +44,12 @@ jobs:
       with:
         python-version: '3.11'
 
+    - name: Cache venv
+      uses: actions/cache@v5
+      with:
+        path: .venv
+        key: venv-${{ runner.os }}-${{ runner.arch }}-py3.11-install-${{ hashFiles('uv.lock') }}
+
     - name: Install dependencies
       run: |
         make install
--- a/.github/workflows/dataset_loading.yml
+++ b/.github/workflows/dataset_loading.yml
@@ -37,18 +37,13 @@ jobs:
       uses: astral-sh/setup-uv@v7
       with:
         enable-cache: true
+        prune-cache: false
         version: "0.10.12"
 
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
         python-version: '3.11'
-
-    - name: Cache venv
-      uses: actions/cache@v5
-      with:
-        path: .venv
-        key: venv-${{ runner.os }}-${{ runner.arch }}-py3.11-install-${{ hashFiles('uv.lock') }}
 
     - name: Install dependencies
       run: |
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -40,6 +40,7 @@ jobs:
         uses: astral-sh/setup-uv@v7
         with:
           enable-cache: true
+          prune-cache: false
           version: "0.10.12"
 
       - uses: actions/setup-python@v6
@@ -50,14 +51,8 @@ jobs:
         if: github.event_name == 'push'
         uses: actions/configure-pages@45bfe0192ca1faeb007ade9deae92b16b8254a0d
 
-      - name: Cache venv
-        uses: actions/cache@v5
-        with:
-          path: .venv
-          key: venv-${{ runner.os }}-${{ runner.arch }}-py3.10-docs-${{ hashFiles('uv.lock') }}
-
       - name: Install dependencies
-        run: uv sync --group docs
+        run: uv sync --frozen --group docs
 
       - name: Build
         run: make build-docs
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -50,6 +50,12 @@ jobs:
         if: github.event_name == 'push'
         uses: actions/configure-pages@45bfe0192ca1faeb007ade9deae92b16b8254a0d
 
+      - name: Cache venv
+        uses: actions/cache@v5
+        with:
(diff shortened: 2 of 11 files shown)


KennethEnevoldsen (commented):
We would probably need a test to check that uv.lock is up to date then (which I don't think it currently it)

KennethEnevoldsen:
I think it might be better to use the `uv` cache: https://docs.astral.sh/uv/guides/integration/github/#caching

It just caches the dependency and point to it during the setup. So you are never using the incorrect environment, but you can to "build" your environment every time (which is just setting pointer in uv so almost instant)

Samoed:
> We would probably need a test to check that uv.lock is up to date then (which I don't think it currently it)

We can run it after each new mteb version, but this solves problem only partly.

isaac-chung:
Each .venv cache is ~3.8 GB. GitHub Actions has a 10 GB cache limit per repository. With multiple jobs each creating separate venv cache keys (lint, tests, docs, model-loading, etc.), they evict each other almost immediately — only ~2-3 can exist at once.

Samoed:
Pip merged standart `py.lock` implementation. Might be interested in it https://github.com/pypa/pip/issues/13334

isaac-chung:
> .venv/lib/python3.13/site-packages/sentence_transformers/base/modality.py:1 
  4: in <module>                                                              
      from sentence_transformers.base.modality_types import (                 
  .venv/lib/python3.13/site-packages/sentence_transformers/base/modality_type 
  s.py:16: in <module>                                                        
      from torchcodec.decoders import AudioDecoder, VideoDecoder              
  .venv/lib/python3.13/site-packages/torchcodec/__init__.py:12: in <module>   
      from . import decoders, encoders, samplers, transforms  # noqa          
  .venv/lib/python3.13/site-packages/torchcodec/decoders/__init__.py:7: in    
  <module>                                     

Seems like sentence transformers need torchcodec now?

Samoed:
No, it has fallback. https://github.com/huggingface/sentence-transformers/blob/01cd5f5bf33ad7b6435cf08b1d7984ab59875c5c/sentence_transformers/base/modality_types.py#L15-L19

If torchcodec installed incorrectly it would fail on each import. In ci wrong version combination of torch and torchcodec is installed 

```python
 + torch==2.10.0
 + torchaudio==2.11.0
 + torchcodec==0.11.1
```
Torch `2.10` is compatible with torchcodec `0.10`, not `0.11`

Samoed:
Hm, after the changes here it seems all our caches were reset and now our dependencies are taking more than 20 minutes to install.

KennethEnevoldsen:
> Hm, after the changes here it seems all our caches were reset and now our dependencies are taking more than 20 minutes to install.

On resolution? (we could just force it to use the lock file and then add a test ensuring that the lock if up to date)

Samoed:
I don't know why. In this branch seems everything normal, but our test CIs started installing dependencies 20 min https://github.com/embeddings-benchmark/mteb/actions/runs/25146122299/job/73706301456 maybe this can be caused by changing cache here, idk

Samoed:
In 2a73572320475c812d88a8a565c2fa4f13318f2b ci took ~15 min https://github.com/embeddings-benchmark/mteb/actions/runs/24936417753/job/73022655878 but in f625341ce03afe795726f7bf944c644be119ee79 it took 30 min https://github.com/embeddings-benchmark/mteb/actions/runs/24954390462/job/73070030880

Between them CI or pyproject wasn't changed https://github.com/embeddings-benchmark/mteb/compare/2a73572...f625341

PR #4536, opened 2026-04-28
Add Zero shot classification for breakfast classification

deep9539:
  - [ ] `facebook/pe-av-small-16-frame` or another small model
- [x] I have checked that the performance is neither trivial (close to perfect scores) nor random.
- [x] I have considered the size of the dataset and reduced it if it is too big (e.g. 2048 examples for binary classification)

[mteb__baseline_random_encoder.json](https://github.com/user-attachments/files/27151243/mteb__baseline_random_encoder.json)


deep9539:
Running it on `facebook/pe-av-small-16-frame` is taking too long. Is there any way I can maybe run it on subsample of data.

Samoed (approved):

Samoed:
> Is there any way I can maybe run it on subsample of data.

No, we don't have such option

PR #4543, opened 2026-04-28
fix: ensure that rteb(english) only use the english subset of `CureV1`

### LINKED ISSUE #4535: RTEB is multilingual even when selecting the EN subset
### Describe the bug

When evaluating on `RTEB(eng, beta)`, the `CureV1` dataset also loads and is evaluated on the `fr` and `es` subsets.

As an aside, `CureV1` takes so long that running RTEB is impractical. Running all tasks on RTEB up until `CureV1` took about 3 hours, while `CureV1` itself takes about 8 hours for the first of many of its subsets. I think finishing it would take days. I don't really understand why, the dataset doesn't seem that big

### To reproduce

```python
from mteb.benchmarks import get_benchmark

task = get_benchmark("RTEB(eng, beta)").tasks[-7]
print(task)
# CUREv1Retrieval(name='CUREv1', languages=['eng', 'fra', 'spa'])
```

This is just not the string displaying these languages; it is genuinely evaluated on `fra` and `spa`, which takes a long time.

### Additional information

_No response_

No

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4535

Keep `CureV1` with only english language

### DIFF
--- a/.github/workflows/reference_models.yml
+++ b/.github/workflows/reference_models.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          make install-for-tests
+          uv sync --group test
 
       - name: Run reference model coverage test
         run: |
--- a/mteb/benchmarks/benchmarks/rteb_benchmarks.py
+++ b/mteb/benchmarks/benchmarks/rteb_benchmarks.py
@@ -87,7 +87,7 @@ RTEB_ENGLISH = RtebBenchmark(
     )
     + (
         get_task(
-            "CureV1",
+            "CUREv1",
             hf_subsets="en",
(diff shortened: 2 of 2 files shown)


KennethEnevoldsen (approved):
Could also have been done with the `exclusive_language_filter=True` - but this works as well

Samoed:
> Could also have been done with the exclusive_language_filter=True

No, I tried this, but got an error
```
ValueError: No subsets were found for AppsRetrieval with filters: language code ['eng'], script None, hf subsets None.
```

`AppsRetrieval` has in subsets `eval_langs=["eng-Latn", "python-Code"]`

PR #4547, opened 2026-04-28
fix: Only log missing subset once

KennethEnevoldsen:
Missing subset appears more than 14k times in the leaderboard. It is probably enough just to raise this warning once.

### DIFF
--- a/mteb/_log_once.py
+++ b/mteb/_log_once.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import ClassVar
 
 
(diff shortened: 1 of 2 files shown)


Samoed (commented):
I'm not sure if it's a good idea. Because results could have different missing scores and we should report them, but since this happens mostly when we're running leaderboard and loading old scores I think this is fine

KennethEnevoldsen:
On different errors it will print them all (it is only if it is exactly the same)

PR #4557, opened 2026-04-29
 model: Add LanguageBind video and audio model wrapper

### LINKED ISSUE #4951: Add model: LanguageBind

https://huggingface.co/LanguageBind/LanguageBind_Video

### Arxiv link

https://arxiv.org/pdf/2310.01852

### Description of the model

LanguageBind has been showing promising results on some of the papers I was reading. We discussed adding this for MVEB, but let's add it for Omni since this support all modalities.

myang333:
Hey! This adds model integration for [LanguageBind](https://github.com/PKU-YuanGroup/LanguageBind) (ICLR 2024) — a multimodal embedding model that aligns video, audio, and text into a shared embedding space.
Models added:

LanguageBind/LanguageBind_Video_FT (video + text, MIT license)
LanguageBind/LanguageBind_Audio_FT (audio + text, MIT license)

Implementation notes:

LanguageBind uses its own library (needs to be cloned from GitHub, built on OpenCLIP)
Video and audio models are loaded separately but share the same text encoder
Embedding dim: 768

Results:

Ran evaluation on VideoRetrieval task, results JSON attached

Refs: [Paper](https://arxiv.org/abs/2310.01852) | [HuggingFace](https://huggingface.co/LanguageBind/LanguageBind_Video_FT) | Related to #4130

Model checklist:
 The model is public, i.e., is available either as an API or the weights are publicly available to download
 I reproduced results from the original paper (if applicable) on at least one benchmark, and I am including the results in the PR description.(Note on the last item: LanguageBind's original paper doesn't include MTEB benchmarks, so there are no paper results to reproduce against.)


VideoRetrieval.json
Samoed, line 1: You shouldn't add results to main repo

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,334 @@
...
+        num_frames: int | None = 8,
+        max_samples: int | None = None,
+        **kwargs: Any,
+    ):
+        requires_package(
+            self,
+            "languagebind",
+            model_name,
+            install_instruction="pip install git+https://github.com/PKU-YuanGroup/LanguageBind.git",
+        )
```
Samoed: We don't use this anymore
```suggestion
```

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,334 @@
...
+    public_training_code="https://github.com/PKU-YuanGroup/LanguageBind",
+    public_training_data=None,
+    framework=["PyTorch"],
+    similarity_fn_name="cosine",
+    use_instructions=False,
+    training_datasets=set(),
+    citation=_LANGUAGE_BIND_CITATION,
+    license="mit",
+    reference="https://github.com/PKU-YuanGroup/LanguageBind",
+    extra_requirements_groups=[],
```
Samoed:
```suggestion
    extra_requirements_groups=["languagebind"],
```

Samoed: You also should add `languagebind` to `pyproject`

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,334 @@
...
+
+        if embeddings is not None:
+            return embeddings
+
+        raise ValueError(
+            f"No supported modality found in dataset features: {list(inputs.dataset.features.keys())}"
+        )
+
+
+# --- Model Metadata ---
```
Samoed:
```suggestion
```

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def _text_model(self) -> Any:
+        # Audio and video checkpoints share the same text-tower architecture;
+        # use the video-aligned tower to match the joint space we expose.
+        return self.video_model
+
+    def _tokenize(self, texts: list[str]) -> dict[str, torch.Tensor]:
+        tokens = self.video_tokenizer(
+            texts,
+            max_length=77,
```
Samoed, line 73: Why do you set max length to 77?

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,334 @@
...
+            install_instruction="pip install git+https://github.com/PKU-YuanGroup/LanguageBind.git",
+        )
+        from languagebind import (
+            LanguageBindAudio,
+            LanguageBindAudioProcessor,
+            LanguageBindAudioTokenizer,
+            LanguageBindVideo,
+            LanguageBindVideoProcessor,
+            LanguageBindVideoTokenizer,
+        )
```
Samoed: Can you also include image?

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,334 @@
...
+        if has_video and not has_audio and not has_text:
+            return self.get_video_embeddings(inputs, **kwargs)
+
+        # Audio-only
+        if has_audio and not has_video and not has_text:
+            return self.get_audio_embeddings(inputs, **kwargs)
+
+        # Text-only
+        if has_text and not has_video and not has_audio:
+            return self.get_text_embeddings(inputs, prompt_type=prompt_type, **kwargs)
```
Samoed: Can you use one `if` per modality? I don't think that `-only` is necessary
```suggestion
```

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+language_bind_video_ft = ModelMeta(
+    loader=LanguageBindWrapper,
+    name=_VIDEO_MODEL_NAME,
+    revision="main",
+    n_parameters=None,
+    memory_usage_mb=None,
+    max_tokens=77,
+    embed_dim=768,
```
Samoed: Can you fill n_parameters, n_embedding_parameters, memory_usage?

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,334 @@
...
+    license="mit",
+    reference="https://github.com/PKU-YuanGroup/LanguageBind",
+    extra_requirements_groups=[],
+)
+
+
+language_bind_video_ft = ModelMeta(
+    loader=LanguageBindWrapper,
+    name=_VIDEO_MODEL_NAME,
+    revision="main",
```
Samoed: Revision shouldn't be `main`

Samoed (changes requested):

mteb/models/model_implementations/language_bind_models.py
myang333: max_length=77 is the clip tokenizer default context length

pyproject.toml
```diff
@@ -45,6 +45,8 @@ dependencies = [
     "pytrec-eval-terrier>=0.5.6",
     "pydantic>=2.11.0",
     "polars>=1.0.0",
+    "torchcodec>=0.10.0",
+    "gradio>=6.0.1",
```
Samoed: You shouldn't update our core dependencies. You can install them via extras
```suggestion
```

pyproject.toml
```diff
@@ -127,6 +129,7 @@ omnivinci = ["transformers==4.46.0", "torch==2.8.*", "einops>=0.8.0", "openai-wh
 qwen_omni_utils = ["qwen_omni_utils"]
 embeddinggemma = ["transformers>=4.56.0"]
 ebind = ["ebind @ git+https://github.com/encord-team/ebind@7909701e9372353ca678b9515f9f61cf87c83c71", "soundfile>=0.13.1", "transformers>=4.57.1,!=5.0.0"]
+languagebind = ["languagebind @ git+https://github.com/PKU-YuanGroup/LanguageBind.git"]
```
Samoed: Can you pin version?

AdnanElAssadi56 (commented):
we can remove uv.lock I guess

Samoed:
@Michelleyyy333 How do you install languagebind? In logs
```
Updating https://github.com/PKU-YuanGroup/LanguageBind.git (HEAD)
    Updated https://github.com/PKU-YuanGroup/LanguageBind.git (7070c53375661cdb235801176b564b45f96f0648)
  × Failed to download and build `languagebind @
  │ git+[https://github.com/PKU-YuanGroup/LanguageBind.git`](https://github.com/PKU-YuanGroup/LanguageBind.git%60)
  ╰─▶ /home/runner/work/_temp/setup-uv-cache/git-v0/checkouts/a994bd661ba96114/7070c53
      does not appear to be a Python project, as neither `pyproject.toml` nor
      `setup.py` are present in the directory
```

> we can remove uv.lock I guess

No, we shouldn't delete it

myang333:
@Samoed 

Hey! The LanguageBind doesn't actually have a setup.py or pyproject.toml in their repo, so it can't be pip-installed. When I was testing it on GPU, I just cloned the repo and added it to PYTHONPATH:
git clone https://github.com/PKU-YuanGroup/LanguageBind.git
export PYTHONPATH="/path/to/LanguageBind:$PYTHONPATH"
Not sure what the best way to handle this in the pyproject.toml is — should I just remove the languagebind entry from optional deps and set extra_requirements_groups=[]?

Samoed:
> should I just remove the languagebind entry from optional deps and set extra_requirements_groups=[]?

Yes. Can you add an instruction with setup in class docstring?

uv.lock
Samoed, line 1: Can you merge main into your branch and sync lock?

Samoed (approved):

Samoed:
@myang333 Can you run this models on some tasks to see scores?

myang333:
CI tests are failing due to a timeout in the setup step (Launchpad PPA connection timeout) , doesn't seem related to my code changes. Should I retrigger the workflow or will it resolve on its own?

Samoed:
I think we need to wait a bit, and I'll restart CI

myang333:
gpu_result
<div class="overflow-x-auto w-full px-2 mb-6">

|Task | Type | Model | Split | Main Score|
|-|-|-|-|-|
|VideoRetrieval | Retrieval | LanguageBind_Video_FT | dev | 0.089|
|CREMA_D | AudioClassification | LanguageBind_Audio_FT | train | 0.334|
|ESC50_Zeroshot | AudioZeroshotClassification | LanguageBind_Audio_FT | train | 0.822|
|CIFAR10 | ImageClassification | LanguageBind_Image | test | 0.985|
|CIFAR100ZeroShot | ZeroShotClassification | LanguageBind_Image | test | 0.852|

Also found and fixed a bug in '_transform_audio', the audio processor expects a (waveform, sample_rate) tuple. Pushed the fix.

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,385 @@
...
+    citation=_LANGUAGE_BIND_CITATION,
+    license="mit",
+    reference="https://github.com/PKU-YuanGroup/LanguageBind",
+    extra_requirements_groups=[],
+)
+
+
+language_bind_video_ft = ModelMeta(
+    loader=LanguageBindWrapper,
+    name=_VIDEO_MODEL_NAME,
```
Samoed: Seems that breaks our parser
```suggestion
    name="LanguageBind/LanguageBind_Video_FT",
```

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,385 @@
...
+    max_tokens=77,
+    embed_dim=768,
+    modalities=["video", "text"],
+    loader_kwargs=dict(num_frames=8),
+    **_LANGUAGE_BIND_COMMON,
+)
+
+language_bind_audio_ft = ModelMeta(
+    loader=LanguageBindWrapper,
+    name=_AUDIO_MODEL_NAME,
```
Samoed:
```suggestion
    name="LanguageBind/LanguageBind_Audio_FT",
```

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,385 @@
...
+    revision="d8c2e37b439f4fc47c649dc8b90cdcd3a4e0c80e",
+    n_parameters=427_616_513,
+    n_embedding_parameters=37_945_344,
+    memory_usage_mb=1631,
+    max_tokens=77,
+    embed_dim=768,
+    modalities=["image", "text"],
+    loader_kwargs=dict(),
+    **_LANGUAGE_BIND_COMMON,
+)
```
Samoed, line 566: Hm. I looked again and I think we should split each model into different wrapper to make possible to run each modality independently and create an "omni" wrapper that uses all models together like was done for speech-t5 https://github.com/embeddings-benchmark/mteb/blob/d203140913ec4abbd207e359f3088d2331831927/mteb/models/model_implementations/speecht5_models.py#L364-L366

or we can just keep `omni` model

Samoed (changes requested):

Samoed (approved):

Samoed:
@myang333 Can you run these models on tasks from paper and try to reproduce their results?

myang333:
Sure! I'll check the paper for their reported benchmarks and try to reproduce.

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,550 @@
...
+    transform directly to the frame tensor and skip the file-loading step.
+    """
+
+    def __init__(
+        self,
+        model_name: str,
+        device: str = "cuda" if torch.cuda.is_available() else "cpu",
+        fps: float | None = None,
+        max_frames: int | None = None,
+        num_frames: int | None = 8,
```
AdnanElAssadi, line 126: Did you verify this is the number it expects (used in training or recommened)?

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,550 @@
...
+            embeddings = text_emb if embeddings is None else embeddings + text_emb
+        if has_image:
+            image_emb = self.image_model.get_image_embeddings(inputs, **kwargs)
+            embeddings = image_emb if embeddings is None else embeddings + image_emb
+        if has_audio:
+            audio_emb = self.audio_model.get_audio_embeddings(inputs, **kwargs)
+            embeddings = audio_emb if embeddings is None else embeddings + audio_emb
+        if has_video:
+            video_emb = self.video_model.get_video_embeddings(inputs, **kwargs)
+            embeddings = video_emb if embeddings is None else embeddings + video_emb
```
AdnanElAssadi, line 491: Is this how they fuse modalities in original implementation?

AdnanElAssadi56 (approved):
Looks good!

mteb/models/model_implementations/language_bind_models.py
Samoed: No, they're separate models. I implemented similarly with how we implemented speech t5

AdnanElAssadi:
Hi @myang333!

Any updates on running the models?

mteb/models/model_implementations/language_bind_models.py
myang333: Yes, this is the default used in LanguageBind's training. Model zoo link: https://github.com/PKU-YuanGroup/LanguageBind#-model-zoo

myang333:
Hey @Samoed I tried to reproduce the paper's results. Their main video retrieval benchmarks (MSR-VTT, MSVD, DiDeMo, ActivityNet) are in the repo now, but I couldn't get them running due to dependency conflicts 
between LanguageBind's older transformers and the newer torchcodec needed for video decoding.
gpu_result
The paper reports ~89.8% on ESC-50 — my 82.2% is a bit lower, probably because of different eval protocols. All three modalities seem to be working fine though.

Samoed:
@myang333 Seems you've merged incorrectly, and you have a lot of commits that not related to your PR


> I couldn't get them running due to dependency conflicts between LanguageBind's older transformers and the newer torchcodec needed for video decoding

In their requirements they have `transformers==4.30.2`. I think you can try to update it to `"transformers<5"` and this should resolve problems

myang333:
The LanguageBind has somecompatibility issues with newer dependencies:
1. '_expand_mask' was removed from 'transformers.models.clip.modeling_clip' in transformers 4.40+
2. 'clip_loss' was removed from the same module
3. 'torchvision.transforms.functional_tensor' was removed in torchvision 0.17+
4. 'torchaudio.set_audio_backend()' is deprecated
5. LanguageBind's tokenizer `__init__` uses positional args that don't work with newer CLIPTokenizer

1 and 2 are the blockers, LanguageBind crashes on import because of these 2. I'll add monkey-patches for them in the wrapper, and for 3, 4, and 5, I can patch them all in the wrapper too.....or just pin versions in the setup docstring?

Samoed:
I think these versions of transformers won't work with torchcodec

KennethEnevoldsen:
I don't think I have a lot to add here - anything specific you want my eye on @Samoed?

Samoed:
@KennethEnevoldsen Yes, a bit. I think we need to discuss whether we want to add models that are not possible to run directly. LanguageBind models require code from their repo (quite a lot to add directly in the model implementation, similarly to transformers modelling), but it is not possible to install it directly. Previously we had EvalCLIP and Vista that required cloning a repo, but we have quite a few models for video, so I'm not sure what we should do with them.

AdnanElAssadi:
The merged ones until now don't require cloning afair

Samoed:
> The merged ones until now don't require cloning afair

Yes, for text, audio and video models

myang333:
Cleaned up the commit history and added monkey patches in the wrapper for the 5 compatibility issues, so it works with MTEB's current dependency stack without needing to pin older versions.
Verified on GPU across 3 modalities:
截屏2026-05-09 上午2 40 10
The ESC-50 result of 0.9425 is really close to the paper's 94.0 (LanguageBind Large full tuning), so the audio path is working. The other 4 are basically unchanged from before, so the patches don't seem to break anything.

I tried to run the paper's video retrieval benchmarks too (MSR-VTT, MSVD, DiDeMo, ActivityNet) but kept hitting issues with MTEB's video pipeline since torch/torchcodec versions didn't really line up (PyTorch 2.7 isn't on the cu124 index, and the torchcodec wheels with AudioDecoder weren't ABI-compatible with 2.6), so I stopped there.

Installwise....LanguageBind still needs git clone + PYTHONPATH since there's no pip package..

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,659 @@
...
+# LanguageBind expects audio sampled at 16 kHz (its audio mel-spectrogram pipeline).
+_LANGUAGE_BIND_AUDIO_SR = 16000
+
+_LANGUAGE_BIND_SETUP_DOC = """
+    Setup::
+
+        git clone https://github.com/PKU-YuanGroup/LanguageBind.git
+        export PYTHONPATH="/path/to/LanguageBind:$PYTHONPATH"
+        pip install einops decord opencv-python-headless pytorchvideo peft
+"""
```
KennethEnevoldsen: @Samoed thanks for letting me know.

So generally we want the implementation to be reproducible, so it shouldn't require the user to download the files manually, but we have had a few that automatically downloads the relevant files and then runs it - which is also not ideal, but it doesn't require user setup and can be linked to a file.

Setups like this have several issues:
1) versions are not tracked so you can end up it a situation where it doesn't run. We don't want to merge this. We can fix this by adding it extra_requirement group in the model meta (and pyproject.toml)
2) The implementation is influenced by changes to the github, we can fix this by locking the git commit
3) It requires manual cloning and pathing, we can simply do this in code. You can see an example of how this is done [here](https://github.com/embeddings-benchmark/mteb/pull/4352) (however this PR has another problem that the weight are on an unversionable onedrive and would have to be moved over)

If we solve these, then I think we can merge it (though I think we all agree we would prefer an transformers integration or python package).

What do you guys think about this?

isaac-chung: I see the benefit of hosting models like these (language bind, and later evaCLIP) on our HF account, and maybe ping the authors to check (non-blocking), so that these models can easily be run with the generic mteb command `mteb run -m MODEL -t TASK`.

Samoed: > I see the benefit of hosting models like these

We can, but we need to host model files, do integrations etc. Just hosting won't solve problems.

KennethEnevoldsen: Yeah it is a lot of extra work we take on. There is also the option that for some models we don't add the implementation, but I don't think it is our style (reproducibility is king)

myang333: Hi @KennethEnevoldsen 
Just to follow up on the HF mirror idea, one quick question: I don't think I have permission to create repos under the mteb HF org. If we want to go that route, can I upload it to my own HF account first?

Samoed: Yes, you can add to your account for now

myang333:
Hello @Samoed @KennethEnevoldsen , I just pushed an update addressing all three reproducibility points.
Changes:
1. Auto-download — _ensure_languagebind_source() uses huggingface_hub.snapshot_download() to fetch source on first use (cached under ~/.cache/mteb/languagebind/). No manual git clone or PYTHONPATH needed.
2. Pinned revision — HF revision e857f973b672e783ca4100e444bec75363979324 of myang333/languagebind-source.
3. Extras group — added languagebind to ModelMeta.extra_requirements_groups and pyproject.toml. Install with pip install mteb[languagebind].

Also fixed a few small things I noticed in LanguageBindOmniWrapper  and let me know if you'd rather drop these: encode now raises ValueError on unmatched modality like sibling wrappers, fixed the docstring, removed an unused param.

And GPU verified:
截屏2026-05-22 下午11 51 11
ESC-50 matches the paper's 94.0 for LanguageBind Large.

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+    except Exception as e:
+        warnings.warn(f"LanguageBind compat patch `_expand_mask` failed: {e}")
+
+    try:
+        import transformers.models.clip.modeling_clip as _clip_mod
+
+        if not hasattr(_clip_mod, "clip_loss"):
+            import torch.nn as _nn
+
+            def clip_loss(similarity: torch.Tensor) -> torch.Tensor:
```
Samoed: Why do we need loss?

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+    """Compatibility shims so the unmaintained LanguageBind library imports
+    cleanly against current ``transformers`` / ``torchvision`` / ``torchaudio``.
+
+    Each shim is independent and fail-soft: a single broken patch only emits a
+    warning so the rest of the module still imports.
+    """
+    try:
+        import transformers.models.clip.modeling_clip as _clip_mod
+
+        if not hasattr(_clip_mod, "_expand_mask"):
```
Samoed: Why do we need this?

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+
+    try:
+        import torchaudio  # type: ignore[import-not-found]
+
+        if not hasattr(torchaudio, "set_audio_backend"):
+            torchaudio.set_audio_backend = lambda backend: None  # type: ignore[attr-defined]
+    except Exception as e:
+        warnings.warn(
+            f"LanguageBind compat patch `torchaudio.set_audio_backend` failed: {e}"
+        )
```
Samoed: Why are you modifying imports?

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+
+        tokenizer_cls.__init__ = _patched_init  # type: ignore[method-assign]
+        tokenizer_cls._mteb_patched = True  # type: ignore[attr-defined]
+    except Exception as e:
+        warnings.warn(
+            f"LanguageBind tokenizer patch for {tokenizer_cls.__name__} failed: {e}"
+        )
+
+
+if TYPE_CHECKING:
```
Samoed, line 13: TypeChecking should be after imports

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+
+        if not hasattr(torchaudio, "set_audio_backend"):
+            torchaudio.set_audio_backend = lambda backend: None  # type: ignore[attr-defined]
+    except Exception as e:
+        warnings.warn(
+            f"LanguageBind compat patch `torchaudio.set_audio_backend` failed: {e}"
+        )
+
+
+_patch_compatibility()
```
Samoed: This will be applied globally for all tasks and imports

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+
+
+_LANGUAGEBIND_HF_REPO = "myang333/languagebind-source"
+_LANGUAGEBIND_HF_REVISION = "e857f973b672e783ca4100e444bec75363979324"
+_LANGUAGEBIND_CACHE = Path.home() / ".cache" / "mteb" / "languagebind"
```
Samoed: You shouldn't use cache path like this
```suggestion
```

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+
+
+def _ensure_languagebind_source() -> Path:
+    """Download LanguageBind source from HF mirror at pinned revision.
+
+    Returns the local snapshot path (already added to sys.path).
+    """
+    try:
+        from huggingface_hub import snapshot_download
```
Samoed: HF hub always will be available

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+        ) from e
+    except Exception as e:
+        raise RuntimeError(
+            f"Failed to download LanguageBind source from HF "
+            f"({_LANGUAGEBIND_HF_REPO} at revision {_LANGUAGEBIND_HF_REVISION}). "
+            f"Check your internet connection. Original error: {e}"
+        ) from e
+
+    if str(path) not in sys.path:
+        sys.path.insert(0, str(path))
```
Samoed: Why do you need to modify `sys.path` with download model path?

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+
+
+_LANGUAGE_BIND_CITATION = r"""
+@inproceedings{zhu2024languagebind,
+      title={Language{B}ind: Extending Video-Language Pretraining to N-modality by Language-based Semantic Alignment},
+      author={Bin Zhu and Bin Lin and Munan Ning and Yang Yan and Jiaxi Cui and Wang HongFa and Yatian Pang and Wenhao Jiang and Junwu Zhang and Zongwei Li and Cai Wan Zhang and Zhifeng Li and Wei Liu and Li Yuan},
+      booktitle={The Twelfth International Conference on Learning Representations},
+      year={2024},
+      url={https://openreview.net/forum?id=QmZKc7UZCy},
+}
```
Samoed, line 510: I think it's better to cite arxiv https://arxiv.org/abs/2310.01852

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+        sys.path.insert(0, str(path))
+    return path
+
+
+def _patch_languagebind_tokenizer(tokenizer_cls: type) -> None:
+    """Force a LanguageBind tokenizer subclass to use ``CLIPTokenizer.__init__``.
+
+    LanguageBind's tokenizer subclasses inherit from ``CLIPTokenizer`` but
+    override ``__init__`` with a positional signature that breaks under
+    newer ``transformers`` releases. We re-route through
```
Samoed: You can just limit compatible transformers verisons

pyproject.toml
```diff
@@ -127,6 +127,16 @@ omnivinci = ["transformers==4.46.0", "torch==2.8.*", "einops>=0.8.0", "openai-wh
...
 ebind = ["ebind @ git+https://github.com/encord-team/ebind@7909701e9372353ca678b9515f9f61cf87c83c71", "soundfile>=0.13.1", "transformers>=4.57.1,!=5.0.0"]
+languagebind = [
+    "einops",
+    "decord",
+    "opencv-python-headless",
+    "pytorchvideo",
+    "peft",
+    "huggingface_hub",
+    "librosa>=0.10.0",
+    "soundfile>=0.13.1",
```
Samoed: Are they really needed?
```suggestion
```

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,710 @@
...
+            LanguageBindAudioTokenizer,
+        )
+
+        _patch_languagebind_tokenizer(LanguageBindAudioTokenizer)
+
+        self.model_name = model_name
+        self.device = device
+        self.sampling_rate = _LANGUAGE_BIND_AUDIO_SR
+
+        self.model = LanguageBindAudio.from_pretrained(model_name).to(self.device)
```
Samoed:
```suggestion
        self.model = LanguageBindAudio.from_pretrained(model_name, revision=revision).to(self.device)
```

mteb/models/model_implementations/language_bind_models.py
myang333: removed

mteb/models/model_implementations/language_bind_models.py
myang333: I think we need this one, LanguageBind's modeling_image.py imports _expand_mask straight from transformers' clip module, and newer transformers doesn't export it anymore, so the import just breaks without this. added a comment explaining it.

mteb/models/model_implementations/language_bind_models.py
myang333: tested both on GPU. the torchaudio one was dead code (set_audio_backend still exists, just a no-op now) so I removed it. the torchvision one I had to keep,pytorchvideo imports functional_tensor at load time and newer torchvision dropped it. added a comment there too.

mteb/models/model_implementations/language_bind_models.py
myang333: fixed

mteb/models/model_implementations/language_bind_models.py
myang333: removed

mteb/models/model_implementations/language_bind_models.py
myang333: dropped the ImportError handling around the huggingface_hub import

mteb/models/model_implementations/language_bind_models.py
myang333: LanguageBind isn't on PyPI (no pip install languagebind), so the source has to be vendored. I host a pinned snapshot on HF and download it, and once it's down the folder needs to be on sys.path for from languagebind import ... to work. I insert at the front with an if not in sys.path check so it's idempotent. happy to switch to a submodule or optional dep if that's cleaner.

mteb/models/model_implementations/language_bind_models.py
myang333: done

mteb/models/model_implementations/language_bind_models.py
myang333: moved it into the loader (_ensure_languagebind_source) so it only runs when someone actually loads a LanguageBind model, not at import. the two that are left just add a symbol if it's missing so they shouldn't affect anything else

mteb/models/model_implementations/language_bind_models.py
myang333: I tried thinking about that but core allows transformers>=4.40.0,<6.0.0 and LanguageBind needs things that got removed at different versions, so I couldn't find one version that works for both. and the torchvision one isn't a transformers thing anyway. so I just kept the 2 that are actually needed and dropped the rest

mteb/models/model_implementations/language_bind_models.py
Samoed: I doubt that we should patch external libraries for models. You can pin libraries versions in pyproject for the specific model

myang333: I reworked this so the MTEB wrapper no longer patches external libraries at runtime.

I moved the compatibility fixes into the pinned LanguageBind source snapshot instead (new revision d2d0f6f):

- `_expand_mask` / `clip_loss`: vendored directly in the LanguageBind modeling files. These were removed from transformers before the versions allowed by MTEB core, so pinning a transformers version in the extra would not solve it while still satisfying `transformers>=4.40`.
- tokenizer `__init__`: updated in the source to pass arguments to `CLIPTokenizer.__init__` by keyword.
- `functional_tensor`: handled inside the source before importing `pytorchvideo` (`pytorchvideo` needs the old module name; LanguageBind itself only uses a few transforms from it).
- `set_audio_backend`: guarded with `hasattr`.

The MTEB wrapper now only downloads and imports the pinned source revision. I removed the runtime monkey-patching from the wrapper and verified clean import plus a GPU forward pass with torch 2.11 / transformers 4.45.2.

pyproject.toml
```diff
@@ -130,8 +130,16 @@ qwen-vl = ["transformers>=4.57.0", "qwen-vl-utils>=0.0.14"]
...
 # ebind = ["ebind @ git+https://github.com/encord-team/ebind@7909701e9372353ca678b9515f9f61cf87c83c71", "soundfile>=0.13.1", "transformers>=4.57.1,!=5.0.0"]
-
-
+languagebind = [
+    "einops",
+    "decord",
+    "opencv-python-headless",
+    "pytorchvideo",
+    "peft",
+    "huggingface_hub",
```
```suggestion
```

pyproject.toml
```diff
@@ -130,8 +130,16 @@ qwen-vl = ["transformers>=4.57.0", "qwen-vl-utils>=0.0.14"]
 qwen_omni_utils = ["qwen_omni_utils"]
 embeddinggemma = ["transformers>=4.56.0"]
 # ebind = ["ebind @ git+https://github.com/encord-team/ebind@7909701e9372353ca678b9515f9f61cf87c83c71", "soundfile>=0.13.1", "transformers>=4.57.1,!=5.0.0"]
-
-
+languagebind = [
+    "einops",
+    "decord",
+    "opencv-python-headless",
+    "pytorchvideo",
```
Samoed: Is this needed?

Samoed (changes requested):
Can you remove `__pycache__` from models repo on hf?

mteb/models/model_implementations/language_bind_models.py
Samoed: Can you remove `_ensure_languagebind_source` function now?

pyproject.toml
myang333: einops, decord, opencv-python-headless, and pytorchvideo are all imported by the LanguageBind source so they're needed. I also dropped librosa and soundfile since LanguageBind itself doesn't import them (audio dataset decoding uses the existing audio extras instead). so the extra is down to einops / decord / opencv-python-headless / pytorchvideo / peft now.

Samoed: Can you pin dependencies?

mteb/models/model_implementations/language_bind_models.py
KennethEnevoldsen: @myang333 it is fairly easy to push it to pypi and would simplify this quite a bit.

I believe `uv build` followed by `uv publish` does the job for you (of course you need a pypi user). 

You might also need a pyproject.toml, but if you use an LLM it should be able to take you through the steps without issues.

pyproject.toml
```diff
@@ -130,8 +130,13 @@ qwen-vl = ["transformers>=4.57.0", "qwen-vl-utils>=0.0.14"]
 qwen_omni_utils = ["qwen_omni_utils"]
 embeddinggemma = ["transformers>=4.56.0"]
 # ebind = ["ebind @ git+https://github.com/encord-team/ebind@7909701e9372353ca678b9515f9f61cf87c83c71", "soundfile>=0.13.1", "transformers>=4.57.1,!=5.0.0"]
-
-
+languagebind = [
```
Samoed, line 169: Why you don't pin transformers?

KennethEnevoldsen: Any need to pin it more than nec.?

mteb/models/model_implementations/language_bind_models.py
KennethEnevoldsen: Did it here: https://pypi.org/project/languagebind/

mteb/models/model_implementations/language_bind_models.py
```diff
@@ -0,0 +1,596 @@
...
+    public_training_data=None,
+    framework=["PyTorch"],
+    similarity_fn_name="cosine",
+    use_instructions=False,
+    training_datasets=set(),
+    citation=_LANGUAGE_BIND_CITATION,
+    license="mit",
+    reference="https://github.com/PKU-YuanGroup/LanguageBind",
+    extra_requirements_groups=["languagebind"],
+    max_tokens=77,
```
KennethEnevoldsen, line 527: is max tokens really 77?

pyproject.toml
Samoed: Previosly in code was a lot of workarounds for different versions of transformers, so I think this is needed

KennethEnevoldsen:
@myang333 added refactor here for the pypi package - you will need to merge it into your branch (sadly I can't do it)

https://github.com/embeddings-benchmark/mteb/tree/myang333/main

It also included some tmp test that I just need to see pass on linux and then we can remove them

myang333:
Hi @KennethEnevoldsen , thanks so much for the refactor! I merged it into my branch and tried out the PyPI-package version on a Linux GPU box. The overall approach works, but I ran into two compatibility issues that languagebind==0.1.0 doesn't seem to handle by itself:
1. torchvision ≥0.17 breaks the import
Turns out pytorchvideo 0.1.5 (one of languagebind's deps) imports torchvision.transforms.functional_tensor, which got removed in torchvision 0.17:
File ".../pytorchvideo/transforms/augmentations.py", line 9, in <module>
    import torchvision.transforms.functional_tensor as F_t
ModuleNotFoundError: No module named 'torchvision.transforms.functional_tensor'
I added a small lazy shim in the wrapper (48c309ed) that just aliases functional_tensor → torchvision.transforms.functional right before the languagebind import. It only runs when a LanguageBind wrapper is actually created.

2. transformers needs an upper bound
The package pins transformers<5.0.0,>=4.40.0, so a clean install grabs 4.57.x — and LanguageBind's vendored CLIP code doesn't work with that version:
File ".../transformers/models/clip/modeling_clip.py", line 331, in forward
    attention_interface = ALL_ATTENTION_FUNCTIONS[self.config._attn_implementation]
KeyError: None
Pinning transformers==4.45.2 makes it work. I didn't want to just add an upper bound myself since transformers is shared across so many other model wrappers and I was worried about conflicting with what others need — is there a ceiling you'd prefer? 4.45.2 is the one I tested with.
Quick verification: with the torchvision shim + transformers 4.45.2, the real tasks reproduce fine — CIFAR10 came out to 0.98446, which matches what I got before.

On the temp tests: the modalities issue is fixed (I switched them to mteb.get_model(...) so the modalities get set up properly). The two mock tests still fail, but it looks like it's the mock data rather than the wrapper — MockImageClassificationTask makes RGBA images (real image tasks are RGB), and the mock audio's mel-spectrogram size doesn't line up with what LanguageBind expects. The real tasks (CIFAR10, ESC-50, CREMA-D) all run fine.

Samoed:
>  Pinning transformers==4.45.2 makes it work. I didn't want to just add an upper bound myself since transformers is shared across so many other model wrappers and I was worried about conflicting with what others need — is there a ceiling you'd prefer? 4.45.2 is the one I tested with.

You can just pin transformers version in dependencies as I requested in https://github.com/embeddings-benchmark/mteb/pull/4557#discussion_r3385827479

>  import torchvision.transforms.functional_tensor as F_t
ModuleNotFoundError: No module named 'torchvision.transforms.functional_tensor'

Why not pin torchvision version?

myang333:
I tried pinning torchvision , but torchvision<0.17 won't install anymore — on PyPI the versions jump from 0.17.0 straight down to the ancient 0.2.x range, and the old torch (2.1.x) that goes with it is gone too. So pinning it just makes the install fail.
The real culprit is pytorchvideo 0.1.5 importing torchvision.transforms.functional_tensor, which got removed in 0.17. Since I can't pin an old enough torchvision, I added a tiny lazy shim that just aliases functional_tensor → torchvision.transforms.functional right before the import .

AdnanElAssadi:
@myang333 do you plan to continue this pr?
#4951

myang333:
Hi @AdnanElAssadi56  yes. Last month I updated it by pinning transformers version, and as I explained above, torch version can not be pinned. Other than that, I think everything is ready, could you take a look where I need to change when you have a chance?

tests/test_models/test_language_bind.py
Samoed, line 1: I don't think we need test for that

Samoed (approved):

AdnanElAssadi:
@KennethEnevoldsen anything else needed here?

mteb/models/model_implementations/language_bind_models.py
isaac-chung: https://huggingface.co/LanguageBind/LanguageBind_Video_FT/blob/main/tokenizer_config.json

isaac-chung:
I wasn't able to run the models with the PR as-is (in the slurm cluster), let alone reproduce scores. I can push whatever changes I have. Mainly:
1. Audio: added deps in pyproject.toml
2. Video: added deps in pyproject.toml & updated `_transform_video_frames()`

And here is the script I used to run the model. Example here is for the audio task.

<details>

<summary>slurm script to run languagebind with this PR checked out</summary>

```bash
#!/bin/bash
#SBATCH --job-name=mteb-languagebind-audio
#SBATCH --nodes=1
#SBATCH --ntasks=1
#SBATCH --cpus-per-task=4
#SBATCH --mem=32G
#SBATCH --gres=gpu:1
#SBATCH --time=04:00:00
#SBATCH --output=%x-%j.out
#SBATCH --error=%x-%j.err

set -euo pipefail

REPO_ROOT="${SLURM_SUBMIT_DIR:-$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)}"
VENV_DIR="${REPO_ROOT}/.venv-languagebind"
RESULTS_DIR="/home/niklas/isaac/results"

# ── create venv and install deps ──────────────────────────────────────────────
if [[ ! -d "${VENV_DIR}" ]]; then
    uv venv "${VENV_DIR}"

    # languagebind==0.1.0 uses PEP 639 SPDX license strings in its
    # pyproject.toml, which requires setuptools>=77 + packaging>=24.2.
    BUILD_CONSTRAINTS="$(mktemp)"
    printf "setuptools>=77\npackaging>=24.2\n" > "${BUILD_CONSTRAINTS}"

    uv pip install \
        --python "${VENV_DIR}/bin/python" \
        --build-constraint "${BUILD_CONSTRAINTS}" \
        --extra-index-url https://download.pytorch.org/whl/cu128 \
        "${REPO_ROOT}[languagebind]"

    rm -f "${BUILD_CONSTRAINTS}"
fi

# ── run ───────────────────────────────────────────────────────────────────────
mkdir -p "${RESULTS_DIR}"

"${VENV_DIR}/bin/mteb" run \
    -m LanguageBind/LanguageBind_Image \
    -t CIFAR10 \
    --output_folder "${RESULTS_DIR}"

```

</details>

And here are the results: https://github.com/embeddings-benchmark/results/pull/683

isaac-chung:
The current problem is that the results are split into 3 separate models, and that is quite confusing. 
I think we should only have the omni model @Samoed @AdnanElAssadi @KennethEnevoldsen

Samoed:
I think thats fine. We have some similar models with tts, stt, tts + sst

isaac-chung:
Fine. I'll run a few more tasks on omni to nudge people to use it that way. Running MIEB/MVEB/MAEB as well but won't block this PR.

PR #4561, opened 2026-04-29
model: add IBM Granite multilingual r2 models

aashka-trivedi:

Paper will be released shortly.


mteb/models/model_implementations/ibm_granite_models.py
```diff
@@ -243,7 +277,7 @@
     public_training_data=None,
     use_instructions=False,
     training_datasets=granite_training_data,
-    citation=GRANITE_EMBEDDING_CITATION,
+    citation=GRANITE_EMBEDDING_R2_CITATION,
```
Samoed, line 333: Isn't it "r1" model?

aashka-trivedi: I believe the diff in this UI is misleading, the R2 citation change on line 333 corresponds to granite_english_r2. confirming that the citation is actually correct [here](https://github.com/aashka-trivedi/mteb/blob/main/mteb/models/model_implementations/ibm_granite_models.py#L210)

Samoed: Ah, yes. You're correct

mteb/models/model_implementations/ibm_granite_models.py
```diff
@@ -269,5 +356,82 @@
...
+    reference="https://huggingface.co/ibm-granite/granite-embedding-311m-multilingual-r2",
+    similarity_fn_name=ScoringFunction.COSINE,
+    use_instructions=False,
+    training_datasets=granite_multilingual_r2_training_data,
+    adapted_from=None,
+    superseded_by=None,
+    modalities=["text"],
+    model_type=["dense"],
+    citation=GRANITE_EMBEDDING_R2_CITATION,
+    contacts=None,
```
KennethEnevoldsen, line 396: If there is issues with the model do you want to be contacted (if so add github username here)

KennethEnevoldsen (approved):
Only a minor comment, otherwise not much to add - congratulations on the release!

mteb/models/model_implementations/ibm_granite_models.py
KennethEnevoldsen: (I will merge but do create a new PR if you want this added)

aashka-trivedi: I would like to be contacted, but since you have merged this PR, I can add my username when I update the citation with our new tech report (coming out soon)

PR #4569, opened 2026-04-30
Fix: Querit model revision and release_data

moshesbeta:
Add new model revision of Querit/Querit to match the results of new results on leaderboard


mteb/models/model_implementations/querit_models.py
```diff
@@ -226,8 +226,8 @@ def compute_mask_content_cls(block_types: list[int]) -> torch.Tensor:
     model_type=["cross-encoder"],
     languages=["eng-Latn"],
     open_weights=True,
-    revision="b611630803ed857ecb159ea9df75738bdd40d1c0",
-    release_date="2026-03-08",
+    revision="e50d4d8183ccec6923ac2b256c755ddd3ed24066",
```
Samoed, line 229: Why do you change revision?

moshesbeta: We release a new version of Querit, and we want to upload the new results.

KennethEnevoldsen (approved):

PR #4570, opened 2026-04-30
fix: use torch.long tensor in get_frames_at to handle empty frame lists

AdnanElAssadi56:
`torch.tensor([])` defaults to `dtype=float32`. When a video has zero frames (e.g. corrupted file or metadata reporting `num_frames=0`), `selected_frames` becomes an empty Python list `[]`. Passing that directly to torchcodec's
`get_frames_at` can result in a float tensor being used as frame indices internally causing type errors.


Samoed (commented):
> When a video has zero frames

I don't think that we should have such videos in datasets

AdnanElAssadi:
> > When a video has zero frames
> 
> I don't think that we should have such videos in datasets

yes, i am calculating stats for the datasets. but this shouldn't error out.

PR #4583, opened 2026-05-02
model: add prompts for ICT-TIME-and-Querit/BOOM_4B_v1 and add ICT-TIME-and-Querit/ICT-TIME-and-Querit-embedding-v1

hengran:
The prompts for both models are the same. Previously, with only some examples placed in implementation. This update simply moves all query prompts into implementation as well.


Samoed (approved):

mteb/models/model_implementations/ict_time_and_querit_models.py
```diff
@@ -112,4 +383,36 @@ def instruction_template(
...
+    citation=BOOM_CITATION,
+)
+
+ict_time_and_querit_embedding = ModelMeta(
+    loader=InstructSentenceTransformerModel,
+    loader_kwargs=dict(
+        instruction_template=instruction_template,
+        apply_instruction_to_passages=False,
+        prompts_dict=boom_4b_instructions,
+        trust_remote_code=True,
```
Samoed:
```suggestion
```

mteb/models/model_implementations/ict_time_and_querit_models.py
```diff
@@ -82,15 +144,224 @@ def instruction_template(
...
+  journal={arXiv preprint arXiv:2602.05787},
+  year={2026}
+}"""
 boom_4b_v1 = ModelMeta(
     loader=InstructSentenceTransformerModel,
     loader_kwargs=dict(
         instruction_template=instruction_template,
+        apply_instruction_to_passages=False,
+        prompts_dict=boom_4b_instructions,
+        trust_remote_code=True,
```
Samoed: Your model doesn't require trust remote code
```suggestion
```

mteb/models/model_implementations/ict_time_and_querit_models.py
```diff
@@ -82,15 +144,224 @@ def instruction_template(
...
+  author={Zhang, Hengran and Bi, Keping and Guo, Jiafeng and Zhang, Jiaming and Yang, Wenbo and Shi, Daiting and Cheng, Xueqi},
+  journal={arXiv preprint arXiv:2602.05787},
+  year={2026}
+}"""
 boom_4b_v1 = ModelMeta(
     loader=InstructSentenceTransformerModel,
     loader_kwargs=dict(
         instruction_template=instruction_template,
+        apply_instruction_to_passages=False,
+        prompts_dict=boom_4b_instructions,
```
Samoed, line 361: You should update old implementation, because results would be unreproducible

Samoed (changes requested):

mteb/models/model_implementations/ict_time_and_querit_models.py
hengran: ok, I have removed this hyperparameter

mteb/models/model_implementations/ict_time_and_querit_models.py
hengran: The prompts for both versions are the same. Previously, the full query prompt was stored at https://huggingface.co/ICT-TIME-and-Querit/BOOM_4B_v1, with only some examples placed in implementation. This update simply moves all query prompts into implementation as well. So it seems there’s no need to re-test the old version, right?

Samoed: If you're reffering to https://huggingface.co/ICT-TIME-and-Querit/BOOM_4B_v1/blob/main/task_prompts.json I don't think that there were used during evaluation

hengran: For the old version, the prompts used are those specified in `https://huggingface.co/ICT-TIME-and-Querit/BOOM_4B_v1/blob/main/task_prompts.json`. The evaluation code  I used was adapted from `https://github.com/QwenLM/Qwen3-Embedding/tree/main/evaluation` to fit the latest MTEB interface. This evaluation script reads from `task_prompts.json` instead of using MTEB's default instructions; if a task is missing from `task_prompts.json`, it falls back to the default MTEB instruction for the query.

KennethEnevoldsen:
@hengran, can I ask you to add the model checklist and add the description of why you changed the prompt for the previous to the description of the PR? 

(test failures seems unrelated to the PR, will rerun them)

hengran:
> @hengran, can I ask you to add the model checklist and add the description of why you changed the prompt for the previous to the description of the PR?
> 
> (test failures seems unrelated to the PR, will rerun them)

Sure. I have updated the description of the PR.

KennethEnevoldsen (approved):
Cleaned up the main comment - with that I believe the PR is good to merge - @Samoed anything you want to add?

Samoed (approved):

PR #4486, opened 2026-04-23
[MVEB] Refactor LCO Implementation

AdnanElAssadi56:
Add VideoCollator/AudioCollator for proper video/audio support, remove qwen_omni_utils dependency, add L2 normalization, and expose fps/max_frames/num_frames/max_audio_length params.

Related to:
https://github.com/embeddings-benchmark/mteb/pull/4384

If you add a model or a dataset, please add the corresponding checklist:

* [dataset checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_dataset/#submit-a-pull-request)
* [model checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_model/#submitting-your-model-as-a-pr)

PR #4591, opened 2026-05-03
Add URM R1 model supporting images, video and text representations

### LINKED ISSUE #4514: Add model: ume-r1

https://huggingface.co/collections/zhibinlan/ume-r1

### Arxiv link

https://arxiv.org/abs/2511.00405

### Description of the model

_No response_

deep9539:
  - [] Breakfast Classification dataset
  - [x] Breakfast Zero shot
  - [x] Breakfast ZeroShot performance is 38% while the paper reported the Hit@1 as 20.1%. (IIUC Hit@1 = accuracy)

[breakfast_zeroshot_X_ume_r1_2b.json](https://github.com/user-attachments/files/27325106/breakfast_zeroshot_X_ume_r1_2b.json)

Fixes #4514


deep9539:
@Samoed @AdnanElAssadi56 Please take a look at this PR adding URM-R1 model.

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+        revision: str | None = None,
+        device: str | None = None,
+        fps: float | None = 2.0,
+        max_frames: int | None = 64,
+        num_frames: int | None = None,
+        **kwargs,
+    ) -> None:
+        from mteb._requires_package import requires_package
+
+        requires_package(self, "qwen_vl_utils", "UME-R1", "pip install qwen_vl_utils")
```
Samoed: We don't use this anymore. You can add this to `ModelMeta.extra_requirements_groups`

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+            if torch.cuda.is_available()
+            else "mps"
+            if torch.backends.mps.is_available()
+            else "cpu"
+        )
+
+        attn_impl = kwargs.pop(
+            "attn_implementation",
+            "flash_attention_2" if is_flash_attn_2_available() else None,
+        )
```
Samoed:
```suggestion
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+
+            inputs.collate_fn = VideoCollator(
+                target_sampling_rate=16000,
+                fps=self.fps,
+                max_frames=self.max_frames,
+            )
+
+        all_embeddings: list[torch.Tensor] = []
+        disc_emb_id = self.processor.tokenizer.get_vocab().get("<disc_emb>")
+        gen_emb_id = self.processor.tokenizer.get_vocab().get("<gen_emb>")
```
Samoed: Better to move to `__init__`

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+
+            final_text = f"<disc_emb>\n{UME_R1_EMBED_PROMPT}"
+            content.append({"type": "text", "text": final_text})
+            messages.append([{"role": "user", "content": content}])
+        
+        return messages
+
+    def _process_to_model_inputs(self, messages: list[list[dict]]) -> dict:
+        """Applies chat templates and processes vision info into model tensors."""
+        from qwen_vl_utils import process_vision_info
```
Samoed: We don't use `qwen_vl_utils`, because it doesn't do anything mostly
```suggestion
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+    embed_dim=1536,
+    reference="https://huggingface.co/zhibinlan/UME-R1-2B",
+    **_UME_R1_BASE_KWARGS,
+)
+
+ume_r1_7b = ModelMeta(
+    name="zhibinlan/UME-R1-7B",
+    revision="b5c08e3273d979e0f22445306717c39ca8d45df0",
+    n_parameters=8_291_375_616,
+    n_embedding_parameters=None,
```
Samoed:
```suggestion
    n_embedding_parameters=544_997_376,
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+    use_instructions=True,
+    training_datasets=None,
+    citation=_UME_R1_CITATION,
+)
+
+ume_r1_2b = ModelMeta(
+    name="zhibinlan/UME-R1-2B",
+    revision="e7aaa253cd315be20bd15c8704ad281b4fdf82c9",
+    n_parameters=2_208_985_600,
+    n_embedding_parameters=None,
```
Samoed:
```suggestion
    n_embedding_parameters=233_373_696,
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+            if batch_images:
+                image_content = batch_images[i]
+                images = image_content if isinstance(image_content, list) else [image_content]
+                for img in images:
+                    content.append({"type": "image", "image": img})
+                    
+            if batch_videos:
+                video_content = batch_videos[i]
+                if isinstance(video_content, torch.Tensor):
+                    video_content = [F.to_pil_image(frame) for frame in video_content]
```
Samoed: Seems this is limited by qwen_vl_utils
```suggestion
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+                content.append({"type": "video", "video": video_content})
+
+            final_text = f"<disc_emb>\n{UME_R1_EMBED_PROMPT}"
```
Samoed, line 159: Should we use task-specific prompts? They could use similar like qwen, but this is not clear https://github.com/XMUDeepLIT/UME-R1/blob/ab4fc06667df63791e21efefc3fbfd2fac0faf0b/src/eval/VLM2Vec/adhoc/eval_mteb/mteb_utils.py#L123

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+
+logger = logging.getLogger(__name__)
+
+# Prompt used during model training
```
Samoed:
```suggestion
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+            )
+
+        all_embeddings: list[torch.Tensor] = []
+        disc_emb_id = self.processor.tokenizer.get_vocab().get("<disc_emb>")
+        gen_emb_id = self.processor.tokenizer.get_vocab().get("<gen_emb>")
+
+        if disc_emb_id is None or gen_emb_id is None:
+            logger.warning(
+                "<disc_emb> or <gen_emb> token not found in tokenizer. Extracting the last token instead."
+            )
```
Samoed:
```suggestion
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+                # Perform forward pass
+                output = self.model(
+                    **model_inputs, 
+                    max_new_tokens=8192, 
+                    output_hidden_states=True, 
+                    return_dict_in_generate=True
+                )
+
+                # Extract embeddings from the reasoning path
+                gen_embedding = UMER1Wrapper._extract_embeddings(
```
Samoed:
```suggestion
                gen_embedding = self._extract_embeddings(
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+                    model_inputs["input_ids"], 
+                    output.hidden_states[-1], 
+                    gen_emb_id
+                )
+                all_embeddings.append(gen_embedding)
+
+                # Explicitly clear large objects from GPU memory
+                del output, model_inputs, messages
+                if self.device.startswith("cuda"):
+                    torch.cuda.empty_cache()
```
Samoed:
```suggestion
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+            text=texts, 
+            images=image_inputs, 
+            videos=video_inputs, 
+            padding=True, 
+            return_tensors="pt"
+        ).to(self.device)
+
+    @staticmethod
+    def _extract_embeddings(
+        input_ids_batch: torch.Tensor,
```
Samoed: Shouldn't this be output ids? In their example they're passing only output
```python
generated_ids_trimmed = [
    out_ids[len(in_ids) :] for in_ids, out_ids in zip(inputs.input_ids, generated_ids)
]
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,257 @@
...
+        disc_emb_id = self.processor.tokenizer.get_vocab().get("<disc_emb>")
+        gen_emb_id = self.processor.tokenizer.get_vocab().get("<gen_emb>")
+
+        if disc_emb_id is None or gen_emb_id is None:
+            logger.warning(
+                "<disc_emb> or <gen_emb> token not found in tokenizer. Extracting the last token instead."
+            )
+
+        with torch.no_grad():
+            for batch in tqdm(inputs, desc="Encoding"):
```
Samoed:
```suggestion
            for batch in tqdm(inputs, desc="Encoding", disable=not show_progress_bar):
```

Samoed (commented):
Can you run some tasks to see if results are mathcing? Would be great to have review from @Nonoor

mteb/models/model_implementations/ume_r1.py
Samoed: You can check how we removed this in LCO https://github.com/embeddings-benchmark/mteb/pull/4486

mteb/models/model_implementations/ume_r1.py
deep9539: Ohh, good catch! Corrected, I verified that it is not reading the gen_emb token from the output tokens by adding adding some logging locally.

mteb/models/model_implementations/ume_r1.py
deep9539: Are we using instruction for all the omni models. I did thought about this earlier, but had few questions. Happy to hear what you all think:
1. Should we add this for all task, i.e. classification, retrieval, QA? 
2. How can we ensure that instructions are added for every newly added tasks?
3. Most importantly, are we adding instruction for all omni-models. By adding in only some and not all, I'm not sure whether we will be doing apple-to-apple comparison between models or not.

pyproject.toml
```diff
@@ -452,6 +452,8 @@ module = [
     "vllm",
     "torchaudio",
     "torchcodec",
+    "qwen_vl_utils",
+    "qwen_vl_utils.*",
```
```suggestion
```

Samoed (commented):
Please, don't use force pushing. Force pushed commits hard to review

pyproject.toml
deep9539: done

deep9539:
> Please, don't use force pushing. Force pushed commits hard to review

Apologies, will take care of it in future iterations.

deep9539:
Reran the analyses again on generative embedding for BreakfastZeroShot classification. Breakfast ZeroShot performance is 38% while the paper reported the Hit@1 as 20.1%. (IIUC Hit@1 = accuracy).

(Updated PR description too.)

[breakfast_zeroshot_X_ume_r1_2b.json](https://github.com/user-attachments/files/27325128/breakfast_zeroshot_X_ume_r1_2b.json)

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,272 @@
...
+
+        self.fps = fps
+        self.max_frames = max_frames
+        self.num_frames = num_frames
+
+        # Load model
+        self.model = Qwen2VLForConditionalGeneration.from_pretrained(
+            model,
+            torch_dtype=torch.bfloat16,
+            attn_implementation="flash_attention_2" if is_flash_attn_2_available() else None,
```
Samoed:
```suggestion
            attn_implementation=kwargs.get("attn_implementation", "flash_attention_2" if is_flash_attn_2_available() else None),
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,272 @@
...
+
+    @staticmethod
+    def _build_messages(batch: dict[str, Any]) -> list[list[dict]]:
+        import torchvision.transforms.functional as F
+
+        batch_texts = batch.get("text", [])
+        batch_images = batch.get("image", [])
+        batch_videos = batch.get("video", [])
+        batch_size = max(len(batch_texts), len(batch_images), len(batch_videos))
+        # print(batch)
```
Samoed:
```suggestion
```

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,272 @@
...
+                # Add 1 because step 0 in output.hidden_states is the prefill
+                embedding_idx.append(indices[-1].item() + 1)
+            else:
+                embedding_idx.append(-1)
+        return embedding_idx
+
+    @staticmethod
+    def _extract_generative_reasoning_embeddings(
+        output: Any, model_inputs: dict, emb_id: int | None
+    ) -> torch.Tensor:
```
Samoed: I think would be easier to use as non-static to not pass `emb_id` across 2 functions

Samoed (commented):
Can you address unresolved comments?

Samoed:
> Breakfast ZeroShot performance is 38% while the paper reported the Hit@1 as 20.1%. (IIUC Hit@1 = accuracy).

We use a bit different subset than mmeb, so results won't match. You can try to run vidore tasks to see if image/text part is matching. @AdnanElAssadi56 Can you suggest task from mmeb to match performance?

mteb/models/model_implementations/ume_r1.py
deep9539: done

mteb/models/model_implementations/ume_r1.py
deep9539: Done

mteb/models/model_implementations/ume_r1.py
deep9539: Done.

deep9539:
@Samoed I've address all the comments. I've left few question regarding task-specific instruction comment.

mteb/models/model_implementations/ume_r1.py
```diff
@@ -0,0 +1,275 @@
...
+            for m in messages
+        ]
+
+        return self.processor(
+            text=texts,
+            images=images if len(images) > 0 else None,
+            videos=videos if len(videos) > 0 else None,
+            padding=True,
+            return_tensors="pt",
+        ).to(self.device)
```
Samoed: Can we simply do?
```suggestion
    def _process_to_model_inputs(self, messages: list[list[dict]]) -> dict:
        return self.processor(
            messages,
            tokenize=True,
            padding=True,
            return_tensors="pt",
        ).to(self.device)
```

Samoed:
Yes, I saw your comment about instructions, but this is hard to say what we should do. This is not clear what they're doing for evaluation

mteb/models/model_implementations/ume_r1.py
deep9539: Done, please find the latest run with the updated code.

[breakfast_zeroshot_X_ume_r1_2b.json](https://github.com/user-attachments/files/27379907/breakfast_zeroshot_X_ume_r1_2b.json)

deep9539:
Addressed the comments.

@Samoed thank you very much for your thorough review. IIUC, there are two open questions:
1. Which dataset should we use to match the performance of this implementation with the reported numbers. ([Discussion](https://github.com/embeddings-benchmark/mteb/pull/4591#issuecomment-4368880819))
2. Whether to use task-specific instructions when calling the model? Discussion [here](https://github.com/embeddings-benchmark/mteb/pull/4591#discussion_r3178760214) I guess applies to all models and not just URM R1.

@AdnanElAssadi56 @isaac-chung Any thoughts?

Samoed:
@deep9539 Can you run some of vidore1&2 tasks to verify performance to understand if we need instructions or not

deep9539:
> @deep9539 Can you run some of vidore1&2 tasks to verify performance to understand if we need instructions or not

I ran for one of the task i.e. Vidore2EconomicsReportsRetrieval without instruction. Vidore task are retrieval only. I intuitively feel that instruction might help the most for classification, more so with zero shot classification.

Paper reported metrics is likely NDCG@5 is: 45.7
There are 4 languages, taking the average NDCG@5 is coming at 44.55

I'm also running Vidore2ESGReportsRetrieval, but let me know if you have any other thoughts.

[UME_R1_X_Vidore2EconomicsReportsRetrieval.json](https://github.com/user-attachments/files/27424917/UME_R1_X_Vidore2EconomicsReportsRetrieval.json)

deep9539:
@Samoed Please find the numbers for Vidore2ESGReportsRetrieval x UME_R1_2B:

Number reported [in paper](https://arxiv.org/pdf/2511.00405#page=20.10): 42.7
Average number from mteb run (NDCG@5) : 0.457615

[UME_R1_X_Vidore2ESGReportsRetrieval.json](https://github.com/user-attachments/files/27490121/UME_R1_X_Vidore2ESGReportsRetrieval.json)

Samoed:
> Please find the numbers for Vidore2ESGReportsRetrieval x UME_R1_2B

What do you mean?

Overall, I think this is fine that we have a bit different scores. This is very hard to reproduce fully. Since we have higher results I think this is fine

deep9539:
> > Please find the numbers for Vidore2ESGReportsRetrieval x UME_R1_2B
> 
> What do you mean?
> 
> Overall, I think this is fine that we have a bit different scores. This is very hard to reproduce fully. Since we have higher results I think this is fine

Yes, I agree that there would be variance in the numbers always. Does that mean we are not including instruction when generating the embedding?

Not related to the PR, but I'm still worried if we are using task-specific instructions in any of the video/omni models and will that give any of the model unintended edge than otherwise. 
Do you think it's worth bringing this up and discuss broadly on slack, or you think this is okay?

Samoed:
> Does that mean we are not including instruction when generating the embedding?

I think we can keep current implementation. 

> Not related to the PR, but I'm still worried if we are using task-specific instructions in any of the video/omni models and will that give any of the model unintended edge than otherwise.

Are you worried about that we evaluate some models with instructions and some without them?

deep9539:
> I think we can keep current implementation.

Thank you! 

> Are you worried about that we evaluate some models with instructions and some without them?

Yes, exactly!

Samoed:
> > Are you worried about that we evaluate some models with instructions and some without them?
> 
> Yes, exactly!

We're trying to implement how the authors created the model. If a model is using instructions then we're adding an implementation with instructions, as instructions don't always play a big role for models. For example, `pplx-embed` and `qwen3-embedding` are based on Qwen3 models, but `qwen3-embed` uses instructions while `pplx` doesn't, and `pplx` has higher scores than `qwen3-embed`. So, instructions matter for models individually and can drop or increase performance for different models.

deep9539:
Got it, thank you @Samoed ! 

I assume this PR is ready now, given all the questions and comments are resolved. Please let me know if you would like to see something more before merging.

AdnanElAssadi56 (approved):
LGTM

PR #4595, opened 2026-05-03
fix: maeb job fixes

isaac-chung:
1. fix input dtype in qwen3 omni 
2. VoxPopuliLanguageID input handling
3. fix fleurs query id type (cast to string)

### DIFF
--- a/mteb/models/modality_collators.py
+++ b/mteb/models/modality_collators.py
@@ -92,19 +92,29 @@ class AudioCollator:
         import torchaudio
 
         audio = audio["audio"]
-        if audio["sampling_rate"] != target_sampling_rate:
-            logger.debug(
-                f"Resampling audio from {audio['sampling_rate']} Hz to {target_sampling_rate} Hz."
+
+        try:
+            if audio["sampling_rate"] != target_sampling_rate:
+                logger.debug(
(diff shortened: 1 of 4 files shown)


mteb/models/modality_collators.py
```diff
@@ -92,19 +92,29 @@ def resample_audio(
...
+            # Create silent audio array with appropriate length
+            fallback_length = (
+                max_samples if max_samples is not None else target_sampling_rate
             )
-            audio_array = torch.from_numpy(audio["array"]).float()
-            audio_array = resampler(audio_array)
-            audio_array = audio_array.numpy()
-        else:
-            audio_array = audio["array"]
+            audio_array = np.zeros(fallback_length, dtype=np.float32)
```
Samoed: I don't think we should support such cases. I think better to reupload task without empty samples

isaac-chung: Okay. But I wonder how this was initially handled? I.e. How do we get results 3 months ago? And what broke between now and then?

Samoed: It was handled similarly. When maeb was created originally we had some tasks with empty samples, but then we started reuploading such tasks. Maybe some tasks weren't fixed. CC @AdnanElAssadi56

isaac-chung: That's ok. There's no need to bump versions then. Just reupload.

PR #4596, opened 2026-05-04
parallelize statistics computation

Samoed:
Parallelize hash computation in statistics computation. Mostly for videos, because their processing takes too much time

### DIFF
--- a/mteb/abstasks/_statistics_calculation.py
+++ b/mteb/abstasks/_statistics_calculation.py
@@ -2,7 +2,10 @@ from __future__ import annotations
 
 import hashlib
 from collections import Counter, defaultdict
+from concurrent.futures import ThreadPoolExecutor
 from typing import TYPE_CHECKING, Any, cast
+
+from tqdm.auto import tqdm
 
 from mteb.types.statistics import (
     AudioStatistics,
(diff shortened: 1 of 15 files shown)


mteb/abstasks/image/image_text_pair_classification.py
```diff
@@ -81,7 +81,11 @@ class AbsTaskImageTextPairClassification(AbsTask):
     texts_column_names: str | Sequence[str] = "caption"
 
     def _calculate_descriptive_statistics_from_split(
-        self, split: str, hf_subset: str | None = None, compute_overall: bool = False
+        self,
+        split: str,
+        hf_subset: str | None = None,
+        compute_overall: bool = False,
```
KennethEnevoldsen, line 88:
```suggestion
        compute_overall: bool = False,
        *,
```
Would probably put it before, but this is backward compatible

mteb/abstasks/_statistics_calculation.py
```diff
@@ -27,51 +30,71 @@
...
             frame_indices = list(range(0, num_frames, step))
         else:
             frame_indices = [0]
 
         frames = video.get_frames_at(frame_indices).data
-        hashes.append(
-            hashlib.md5(frames.numpy().tobytes(), usedforsecurity=False).hexdigest()
+        return hashlib.md5(frames.numpy().tobytes(), usedforsecurity=False).hexdigest()
+
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
```
KennethEnevoldsen, line 89: is it better is threads vs parallel process here?

mteb/abstasks/image/image_text_pair_classification.py
KennethEnevoldsen: (general comment)

KennethEnevoldsen (commented):
A few minor changes - otherwise looks good

mteb/abstasks/_statistics_calculation.py
Samoed: A bit AI answer

1. Processes would be slower due to serialization overhead — the inputs are large binary objects (PIL images, numpy audio arrays, VideoDecoder objects). Passing them to worker processes requires pickling, which means copying all that raw pixel/sample data.
2. Memory — threads share memory with the parent, so there's no data copying at all. With ProcessPoolExecutor, each worker gets a full copy of the data it processes.

KennethEnevoldsen: sorry could have done that myself as well ;)

mteb/abstasks/image/image_text_pair_classification.py
Samoed: Changed to
```python
    def _calculate_descriptive_statistics_from_split(
        self,
        split: str,
        *,
        hf_subset: str | None = None,
        compute_overall: bool = False,
        num_proc: int | None = None,
    ) -> SplitDescriptiveStatistics:
```
Since this is private function we can break compatibility a bit

PR #4600, opened 2026-05-04
fix: refactor classification to avoid maintaining a live cache of embeddings

### LINKED ISSUE #3598: Reduce the number of encoding steps in classification

currently classification runs 10 experiments for each subset x split, this means that you can often get a large degree of `model.encode` calls, this can lead to a large overhead in simply switching between tasks and doesn't allow for encode optimizations on the batch size. Doing all the experimentation at once could reduce the runtime without affecting the score.

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3598

Make classification and regression tasks to call `model.encode` once per subset. 

- Moved `undersample_data` call into the fold loop, so the fold's pre-computed embeddings can be sliced by `selected_idx` before passing to `_run_experiment`.
- Updated `_undersample_data` for Regression to return `selected_idx` to simplify main logic

I run `mteb run -m mteb/baseline-random-encoder -t RuSciBenchCitedCountRegression ImdbClassification` and got the same results on main and this branch


tests/test_evaluate.py
```diff
@@ -26,7 +26,7 @@
     MOCK_MVEB_TASK_GRID,
 )
 
-mock_classification = (MockSentenceTransformer(), MockClassificationTask(), 0.5)
+mock_classification = (MockSentenceTransformer(), MockClassificationTask(), 1)
```
Samoed, line 29: Score changed, because `MockSentenceTransformer` using 
```py
        rng_state = np.random.RandomState(42)
        return rng_state.randn(len(inputs), 10)

```
foe encode, and now we're doing only one `encode` instead of 10, so this will produce a bit different embedding

mteb/abstasks/classification.py
```diff
@@ -411,11 +427,11 @@ def _run_experiment(  # noqa: PLR0913
             model,
             encode_kwargs=encode_kwargs,
             test_cache=test_cache,
-            train_cache=sub_train_cache,
+            train_cache=train_cache,
```
KennethEnevoldsen: I think we could refactor to completely remove the cache (just passing in the relevant embeddings).

mteb/abstasks/classification.py
```diff
@@ -257,23 +257,49 @@ def _evaluate_subset(
...
+
+        union_idxs = sorted(set().union(*all_selected_idxs))
+        dataloader_train = create_dataloader(
+            train_split.select(union_idxs),
+            task_metadata=self.metadata,
+            input_column=self.input_column_name,
+            num_proc=num_proc,
+            **encode_kwargs,
+        )
+        logger.info(
```
KennethEnevoldsen, line 275: Hmm logging will look at bit odd now (maybe no need to log for the experiments?)

KennethEnevoldsen (commented):
Looks good, but I think we can simplify a bit more (to make it more maintainable - currently it can be a bit hard to understand why train_cache exists)

mteb/abstasks/classification.py
Samoed: I don't think so. We still run logreg multiple times and logging would help a bit

mteb/abstasks/classification.py
Samoed: Agree, updated

KennethEnevoldsen (approved):
love it!

PR #4601, opened 2026-05-04
fix: raise error on missing scores

### LINKED ISSUE #4599: `push_eval_results` pushed the mean score even if not all tasks are present
`push_eval_results` pushed the mean score for the benchmark even if not all tasks are present:

Example PR:
https://huggingface.co/mteb/baseline-bm25s/discussions/8/files#d2h-092221

Run using:
```py
import mteb

benchmark = mteb.get_benchmark("MTEB(Multilingual, v2)")
# benchmark.push_eval_to_hub(create_pr=True)  # push to eval.yaml
model_meta = mteb.get_model_meta("mteb/baseline-bm25s")
model_meta.push_eval_results(
    "mteb",
    tasks=benchmark,
    create_pr=True,
)
```

Ideally, this should raise an error, and we can potentially add the flag `raise_error=True`. 

A sidenote is that we might want to add support for pushing multiple benchmarks at once (to avoid creating many duplicate PRs)

cc @Samoed

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/4599

### DIFF
--- a/mteb/benchmarks/benchmark.py
+++ b/mteb/benchmarks/benchmark.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import tempfile
 from collections.abc import Iterator, Sequence
 from dataclasses import dataclass, field
@@ -25,6 +26,8 @@ from ._benchmark_metrics import (
 if TYPE_CHECKING:
     from mteb.abstasks.aggregated_task import AbsTaskAggregate
     from mteb.results import BenchmarkResults, ModelResult
+
+logger = logging.getLogger(__name__)
 
 
 @lru_cache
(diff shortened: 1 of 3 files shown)


KennethEnevoldsen (commented):
should we add the `raise_error=True` flag and to allow us to just push everything that is valid.

The issue also mentions that we could add support for multiple benchmarks at once (to avoid having multiple PRs). If we don't fix that here, let us make a new issue on that

Samoed:
Added flag for raise error. I think would be better to create a new issue/pr for multiple benchmarks, because it requires more changes in ResultsCache to do properly

KennethEnevoldsen (approved):

PR #4603, opened 2026-05-04
model: add GeeVec embedding models

lnxtree:
## Summary

Adds GeeVec embedding model definitions to the MTEB model registry:

- `geevec-ai/geevec-embeddings-1.0-lite`: public Hugging Face / Sentence Transformers compatible model with instruction prompts and Matryoshka embedding dimensions `[256, 512, 1024, 2048, 4096]`.
- `geevec-ai/geevec-embeddings-1.0`: API-backed GeeVec embedding model with query/document prompt handling and environment-based API configuration.

The implementation includes:

- GeeVec task prompt mappings
- a local lite model wrapper based on `InstructSentenceTransformerModel`
- an API model wrapper with batching, retries, endpoint resolution, empty-input handling, and long-input truncation fallback

## Model checklist


mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,699 @@
...
+GEEVEC_INSTRUCTION = "Instruct: {instruction}\nQuery: "
+GEEVEC_MAX_SEQ_LENGTH = 512
+GEEVEC_DEFAULT_BATCH_SIZE = 8
+GEEVEC_API_DEFAULT_BATCH_SIZE = 32
+GEEVEC_API_MODEL_BY_DOMAIN = {
+    "coding": "geevec-embeddings-coding-1.0",
+    "reasoning": "geevec-embeddings-reasoning-1.0",
+    "general": "geevec-embeddings-general-1.0",
+}
+GEEVEC_DEFAULT_PROMPT_TEMPLATE = "Instruct: {}\nQuery: {}"
```
Samoed: Why do you have `GEEVEC_INSTRUCTION` and `GEEVEC_DEFAULT_PROMPT_TEMPLATE`?

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,699 @@
...
+        return f"{normalized}/v1/embeddings"
+    return f"{normalized}/openapi/v1/embeddings"
+
+
+def _resolve_hf_endpoint() -> str:
+    return os.environ.get("HF_ENDPOINT", "https://hf-mirror.com")
+
+
+def _resolve_hf_hub_cache() -> str:
+    return os.environ.get("HF_HUB_CACHE", "/data/share/project/shared_models")
```
Samoed: These variables would be automatically read by HF. This is not nessary to set them here

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,699 @@
...
+        local_model_path = _download_geevec_snapshot(model_name, revision)
+        super().__init__(
+            local_model_path,
+            revision=None,
+            device=device,
+            instruction_template=instruction_template,
+            max_seq_length=max_seq_length,
+            prompts_dict=PROMPTS_DICT,
+            **kwargs,
+        )
```
Samoed: Seems that you model is SentenceTransforerms compativle and would be easier to use `snapshot_download`

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,699 @@
...
+
+    def encode(self, inputs, *, task_metadata, hf_split, hf_subset, prompt_type=None, **kwargs):
+        sentences = [text for batch in inputs for text in batch["text"]]
+        # domain = self._resolve_domain(task_metadata)
+        domain = kwargs.get("domain") or None
+
+        kwargs.setdefault("batch_size", GEEVEC_DEFAULT_BATCH_SIZE)
+        kwargs.setdefault("normalize_embeddings", True)
+        if domain is not None:
+            kwargs.setdefault("domain", domain)
```
Samoed: batch_size will always be passed in kwargs. What is domain?
```suggestion
```

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,699 @@
...
+        if (
+            not self.apply_instruction_to_passages
+            and prompt_type == PromptType.document
+        ):
+            return None
+
+        try:
+            instruction = self.get_instruction(task_metadata, prompt_type)
+        except KeyError:
+            # Some test/mocked tasks are not registered in mteb.get_task.
```
Samoed: When do you have this error?

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,699 @@
...
+    ) -> str:
+        try:
+            return self.get_instruction(task_metadata, prompt_type)
+        except KeyError:
+            prompt = getattr(task_metadata, "prompt", "")
+            if isinstance(prompt, dict) and prompt_type:
+                return str(prompt.get(prompt_type.value, ""))
+            return str(prompt or "")
+
+    def _format_query_with_instruction(self, instruction: str, text: str) -> str:
```
Samoed, line 637: I don't think that you need these checks

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,699 @@
...
+        input_type: str,
+        domain: str | None,
+        api_model_name: str,
+    ) -> list[list[float]]:
+        current_max_chars = max(self._min_input_chars, self._max_input_chars)
+        req_texts = self._truncate_texts_by_chars(texts, current_max_chars)
+
+        optional_fields = {}
+        if domain is not None:
+            optional_fields["domain"] = domain
```
Samoed, line 456: What is domain?

mteb/models/model_implementations/geevec_models.py
lnxtree: They are for two different backends:

- `GEEVEC_INSTRUCTION` is used by the local SentenceTransformers-compatible model through `InstructSentenceTransformerModel`, where MTEB passes the instruction separately as the `prompt`.
- `GEEVEC_DEFAULT_PROMPT_TEMPLATE` is used only by the API wrapper, because the API receives raw input text and needs the instruction and query text formatted into one string.

That said, I agree the naming makes this look duplicated. I can make this clearer.

mteb/models/model_implementations/geevec_models.py
lnxtree: Good point. These were added for my local evaluation environment, but they are not necessary in MTEB since Hugging Face already respects `HF_ENDPOINT` / cache-related environment variables. I will remove the custom `_resolve_hf_endpoint`, `_resolve_hf_hub_cache`, and explicit `endpoint` / `cache_dir` arguments.

mteb/models/model_implementations/geevec_models.py
lnxtree: Yes, the lite model is SentenceTransformers-compatible. I used `snapshot_download` mainly because of the custom HF endpoint/cache handling above. Since that is unnecessary, I can simplify this and load it in the standard MTEB/SentenceTransformers way with `model_name` and `revision` directly.

mteb/models/model_implementations/geevec_models.py
lnxtree: This was a defensive fallback for local tests/mocked task metadata where `get_instruction()` may try to resolve an unregistered task through `mteb.get_task(...)`. For normal MTEB tasks it should not be needed.

mteb/models/model_implementations/geevec_models.py
lnxtree: Makes sense. These checks were added to support a few possible template formats, but the default API template is fixed and does not need that flexibility. I will simplify this to a single explicit format path.

mteb/models/model_implementations/geevec_models.py
lnxtree: The `batch_size` default is there because the GeeVec API has a maximum request batch size. If MTEB does not pass an explicit value, the wrapper falls back to a safe default.

`domain` is a GeeVec-specific encode kwarg. It selects the domain-specific behavior for both implementations: the LoRA adapter in the lite model, and the corresponding domain backend in the API model.

mteb/models/model_implementations/geevec_models.py
lnxtree: `domain` is intentional for both GeeVec model variants.

For `geevec-embeddings-1.0-lite`, `domain` is passed to the underlying SentenceTransformers-compatible implementation to select the corresponding LoRA adapter, e.g. coding / reasoning / general.

For `geevec-embeddings-1.0`, `domain` is passed to the API so the service can switch to the corresponding domain-specific embedding backend.

lnxtree:
Thanks for the review @Samoed ! I pushed an update addressing the comments:

- removed the custom HF endpoint/cache handling and now rely on standard SentenceTransformers/HF loading
- clarified that `domain` is intentional: it selects the LoRA adapter for the lite model and the domain-specific backend for the API model
- removed the lite model batch size default and kept the API batch size default because the GeeVec API has a request batch limit
- simplified API instruction formatting
- removed the extra instruction fallback in the lite wrapper

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,653 @@
...
+    ) -> None:
+        super().__init__(
+            model_name,
+            revision=revision,
+            device=device,
+            instruction_template=instruction_template,
+            max_seq_length=max_seq_length,
+            prompts_dict=PROMPTS_DICT,
+            **kwargs,
+        )
```
Samoed:
```suggestion
```

Samoed (commented):
I see that domain is intentional and loads LoRA modules like in jina models, but this shouldn't be get from kwargs, because this is hard to reproduce. You can check if task contains code or not from `task_metadata.languages`, but I'm not sure how to choose `reasoning / general`

mteb/models/model_implementations/geevec_models.py
```diff
@@ -43,6 +57,36 @@ def _resolve_geevec_embeddings_endpoint(base_url: str) -> str:
...
+        return explicit_domain
+
+    eval_langs = task_metadata.eval_langs
+    if isinstance(eval_langs, dict):
+        languages = [lang for langs in eval_langs.values() for lang in langs]
+    else:
+        languages = eval_langs
+
+    if (
+        task_metadata.name in {"BrightRetrieval", "BrightLongRetrieval"}
```
Samoed, line 97: We also have updated version of bright with fixed datasets (previous version of bright have a bit wrong implementation) https://github.com/embeddings-benchmark/mteb/blob/bf25520040bd528d76aaa37c36c682123ef74201/mteb/benchmarks/benchmarks/benchmarks.py#L1379-L1405

lnxtree: Thanks for pointing that out. I updated the resolver to support both BRIGHT representations in MTEB

lnxtree:
Hi @Samoed , I pushed another update that fixes the failing CI by adding `n_embedding_parameters` for `geevec-ai/geevec-embeddings-1.0-lite`.

Samoed:
Can you fix linter?

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,706 @@
...
+        base_url: str | None = None,
+        api_key: str | None = None,
+        session: Any | None = None,
+        **kwargs,
+    ) -> None:
+        import requests
+
+        self._max_retries = max_retries
+        self._timeout = timeout
+        self._default_api_model_name = api_model_name
```
Samoed: You can use 
```suggestion
        self._default_api_model_name = model_name.rsplit("/", 1)[1]
```

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,706 @@
...
+
+        if prompt_type == PromptType.document:
+            return sentences
+
+        instruction = self._resolve_instruction(task_metadata, prompt_type)
+        if not instruction:
+            if self._assert_prompts_exist:
+                raise ValueError(
+                    f"Prompt for task '{task_metadata.name}' not found in prompts_dict or task metadata."
+                )
```
Samoed: You shouldn't raise error if no instruction found

lnxtree:
> Can you fix linter?

I fixed the lint failure by running `ruff format` on `geevec_models.py`.

Samoed (approved):

lnxtree:
Hi, CI is green now and the latest revision has been approved. Is there anything else needed from my side before this can be merged? @Samoed

Samoed:
I'm waiting for Kenneth review

mteb/models/model_implementations/geevec_models.py
```diff
@@ -0,0 +1,725 @@
...
+            )
+
+        embeddings = self.model.encode(
+            sentences,
+            prompt=instruction,
+            **kwargs,
+        )
+
+        if isinstance(embeddings, torch.Tensor):
+            embeddings = embeddings.cpu().detach().float().numpy()
```
KennethEnevoldsen:
```suggestion
            embeddings = embeddings.cpu()
```

KennethEnevoldsen (approved):
I think it is good, only a minor change!

PR #4604, opened 2026-05-04
model: jina-embeddings-v5-omni models

florian-hoenicke:
Hi,

This PR adds the jina-embeddings-v5-omni nano and small base models to MTEB. The text path is parity-verified against the corresponding v5 text models, so the same task routing is used here.

- [x] ModelMeta is filled for both models
- [x] `mteb.get_model_meta(...)` works for both models
- [x] Results PR: https://github.com/embeddings-benchmark/results/pull/513

Thanks!


mteb/models/model_implementations/jina_models.py
```diff
@@ -871,6 +871,118 @@ def encode(
...
+            "BitextMining": "text-matching",
+            "MultilabelClassification": "classification",
+            "Reranking": "retrieval",
+            "Summarization": "text-matching",
+            "InstructionReranking": "retrieval",
+        },
+    ),
+    name="jinaai/jina-embeddings-v5-omni-small",
+    model_type=["dense"],
+    modalities=["text", "image", "audio", "video"],
```
Samoed, line 970: Great addition! Current implementation supports only texts, but we have support of image, audio and video modalities. You can partly reuse implementation from from multimodal sentence transformer https://github.com/embeddings-benchmark/mteb/blob/bf25520040bd528d76aaa37c36c682123ef74201/mteb/models/sentence_transformer_wrapper.py#L237

florian-hoenicke: Thanks, addressed in the latest push. I added a dedicated `JinaV5OmniWrapper` that uses multimodal batching for text/image/audio/video while still passing the Jina `task=` adapter argument, and switched the two omni ModelMeta entries to it. Added focused tests for the loader and modality/task dispatch.

tests/test_models/test_model_meta.py
Samoed, line 1: I don't think you should add tests for your model. If you want to test if prompt is passed correctly is better to use [`MockTasks`](https://github.com/embeddings-benchmark/mteb/blob/bf25520040bd528d76aaa37c36c682123ef74201/tests/mock_tasks.py#L198)

florian-hoenicke:
Follow-up fix pushed in 320263b5: `JinaV5OmniWrapper` now defaults `jinaai/jina-embeddings-v5-omni-nano` to `torch.float32`, which restores text-path parity with `jinaai/jina-embeddings-v5-text-nano`. Verified on A2 with private HF auth via `mteb.get_model(...)`: omni-nano loads fp32 and matched text-nano exactly on retrieval/query and classification/document probes (`max_abs_diff=0.0`). Small remains unchanged.

florian-hoenicke:
Thanks, updated the prompt/task dispatch test in c42e7a97 to use `MockRetrievalTask().metadata` instead of a hand-rolled task metadata object.

florian-hoenicke:
Follow-up corner-case fix pushed in 9f405468: I found the omni HF remote code was stripping text before tokenization, which made trailing-space inputs differ from the text models. The private HF repos now preserve whitespace and the MTEB metadata pins those fixed revisions. Verified on A2 through `mteb.get_model(...)`: nano `36db6194...` and small `8b4f2c44...` match their text counterparts on trailing-space probes with `max_abs=0.0`.

Samoed:
> JinaV5OmniWrapper now defaults jinaai/jina-embeddings-v5-omni-nano to torch.float32, which restores text-path parity

You can just pass this in `loader_kwargs` without changes in `__init__`

florian-hoenicke:
Updated again after the latest private HF repo fix: MTEB now pins nano `6f88a89e...` and small `43affca6...`. Verified on A2 against the text counterparts on trailing-space probes and all unique `STSBenchmark` strings: `max_abs=0.0`, min cosine `1.0` / `0.99999988`.

tests/test_models/test_model_meta.py
florian-hoenicke: Thanks, addressed in 605cb143. I removed the Jina-specific tests from `tests/test_models/test_model_meta.py` instead of adding more model-specific coverage, and moved the omni-nano `torch.float32` default into the model's `loader_kwargs` as suggested. Local checks: `python3 -m pytest -o addopts='' tests/test_models/test_model_meta.py -q` (`1935 passed`) and `python3 -m ruff check mteb/models/model_implementations/jina_models.py tests/test_models/test_model_meta.py`.

florian-hoenicke:
Update: the referenced HF repos are now public and ungated:

- `jinaai/jina-embeddings-v5-omni-nano`: Hub API reports `private: false`, `gated: false`, `disabled: false`
- `jinaai/jina-embeddings-v5-omni-small`: Hub API reports `private: false`, `gated: false`, `disabled: false`

I also pushed 605cb143 to address the open review feedback: nano fp32 is now passed through `loader_kwargs`, and the Jina-specific tests were removed. Local checks passed (`test_model_meta.py`: 1935 passed; ruff on edited files passed).

Samoed (approved):

Samoed:
I think we would wait until public release

florian-hoenicke:
The public release is live now: both referenced model repos are publicly accessible and ungated on the Hub.

- https://huggingface.co/jinaai/jina-embeddings-v5-omni-nano
- https://huggingface.co/jinaai/jina-embeddings-v5-omni-small

Hub API currently reports `private: false`, `gated: false`, `disabled: false` for both. Please let me know if you need any other release artifact before merging.

PR #4620, opened 2026-05-07
model: adds ogma models

sam-at-axiotic:
Adds the Ogma model family


mteb/models/model_implementations/axiotic_models.py
```diff
@@ -0,0 +1,201 @@
...
+            revision=revision,
+            trust_remote_code=True,
+        )
+        self.model.to(self.device)
+        self.model.eval()
+        tokenizer_file = hf_hub_download(
+            model_name, filename="tokenizer.json", revision=revision
+        )
+        self.tokenizer = Tokenizer.from_file(tokenizer_file)
+        self.n_special_tokens = int(getattr(self.model.config, "n_special_tokens", 7))
```
Samoed: Why not to use `AutoTokenizer.from_pretrained`?

mteb/models/model_implementations/axiotic_models.py
```diff
@@ -0,0 +1,201 @@
...
+        self,
+        inputs: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **_: Any,
+    ) -> Array:
+        del task_metadata, hf_split, hf_subset
```
Samoed:
```suggestion
```

mteb/models/model_implementations/axiotic_models.py
```diff
@@ -0,0 +1,201 @@
...
+            tokens = enc.tokens
+            if tokens and tokens[0] in {"[CLS]", "<s>"}:
+                ids = ids[1:]
+                tokens = tokens[1:]
+            if tokens and tokens[-1] in {"[SEP]", "</s>"}:
+                ids = ids[:-1]
+            ogma_ids = (
+                [2] + [token_id + self.n_special_tokens for token_id in ids] + [3]
+            )
+            encoded.append(ogma_ids[: self.max_length])
```
Samoed: This approach would remove special tokens

mteb/models/model_implementations/axiotic_models.py
sam-at-axiotic: Thanks for the review @Samoed. Good catch, have switched

sam-at-axiotic:
@Samoed your comments lead to us finding a bug. Moving this back to draft whilst we investigate.

mteb/models/model_implementations/axiotic_models.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def to(self, device: torch.device) -> None:
+        self.device = device
+        self.model.to(device)
+
+    @staticmethod
+    def _resolve_task_token(task_metadata: TaskMetadata | None) -> str:
+        if task_metadata is not None and task_metadata.type in _QRY_TASK_TYPES:
+            return "QRY"
+        return "SYM"
```
Samoed, line 63: You're no longer using `[DOC]`? This token still mentioned in model card in `Task tokens`

sam-at-axiotic:
@Samoed We have added a tokenizer to ensure encoding is proper and reopened this PR

Samoed (approved):

mteb/models/model_implementations/axiotic_models.py
sam-at-axiotic: For now no. [DOC] is still a valid task token on the model itself (the model card is accurate), it
just isn't reachable through this wrapper. Our MTEB numbers are produced with (QRY, QRY) on
Retrieval/Reranking/InstructionRetrieval/InstructionReranking and SYM everywhere else. The wrapper hard-codes the routing that reproduces the leaderboard scores. We found we had better performance with symmetric retrieval. However, we are currently exploring further tasks where asymetric retrieval could work better. We will update the model cards to reflect this. Thanks again for the review @Samoed .

PR #4625, opened 2026-05-08
fix tolerate off-by-one frame metadata in compute_video_hashes

AdnanElAssadi56:
@Samoed 
Some videos' container metadata over-counts frames by one, it claims num_frames = N but only N-1 frames are actually decodable. compute_video_hashes samples at indices range(0, num_frames, round(avg_fps)), which lands on the imaginary last index when (N-1) is divisible by round(fps) and crashes.

Example (MSVD video 184): metadata says num_frames=226, avg_fps=25 → samples include index 225, but the file only has frames 0–224. Eval doesn't hit this because its stride (num_frames // target) doesn't align with N-1.

### DIFF
--- a/mteb/abstasks/_statistics_calculation.py
+++ b/mteb/abstasks/_statistics_calculation.py
@@ -57,7 +57,9 @@ def compute_video_hashes(videos: list[VideoDecoder]) -> list[str]:
     hashes = []
     for video in videos:
         meta = video.metadata
-        num_frames = meta.num_frames
+        # Drop the last frame index because some container metadata over-counts
+        # by one (the final claimed frame fails to decode).
+        num_frames = meta.num_frames - 1 if meta.num_frames else meta.num_frames
         avg_fps = meta.average_fps
 
         if num_frames is not None and avg_fps is not None and avg_fps > 0:


mteb/abstasks/_statistics_calculation.py
```diff
@@ -67,7 +67,14 @@ def compute_video_hashes(videos: list[VideoDecoder]) -> list[str]:
         else:
             frame_indices = [0]
 
-        frames = video.get_frames_at(frame_indices).data
+        try:
+            frames = video.get_frames_at(frame_indices).data
```
Samoed: Can we do then `num_frames = meta.num_frames - 1`? Do we have same problem in collator?

AdnanElAssadi: but this will remove one frame from everything

Samoed: Yes, but this is just hash computation and I don't think that this is important. Otherwise, we can apply approach similar to video collator.

mteb/abstasks/_statistics_calculation.py
```diff
@@ -57,7 +57,9 @@ def compute_video_hashes(videos: list[VideoDecoder]) -> list[str]:
     hashes = []
     for video in videos:
         meta = video.metadata
-        num_frames = meta.num_frames
+        # Drop the last frame index because some container metadata over-counts
+        # by one (the final claimed frame fails to decode).
+        num_frames = meta.num_frames - 1 if meta.num_frames else meta.num_frames
```
Samoed, line 62: Wouldn't this always be `num_frames - 1` then? Can you give an example of tasks where this is failing? Does the collator work correctly for these tasks?

AdnanElAssadi: yeah, but what if someone just puts `num_frames=0`

AdnanElAssadi: tasks that failed i remember were: MSVDV2TRetrieval and MSVDT2VRetrieva

AdnanElAssadi: You could try reproducing like this:

```
import mteb
task = mteb.get_task('MSVDV2TRetrieval')
task.load_data()
videos = task.dataset['default']['test']['queries']['video']
videos[184].get_frames_at(list(range(0, 226, 25))).data
# RuntimeError: Requested next frame while there are no more frames left to decode.

```

AdnanElAssadi: i think integer-division stride makes the collator skip the bad frame by accident or something. I never got this error in any of my evals until now. i guess it can be triggered with a malicious `N` or `target`

Samoed: I think we can leave as it is now. Would be great to submit an issue to torchcodec with this, but I don't see how we save this video in mp4 or other format to simplify reproduction

Samoed (approved):

mteb/abstasks/_statistics_calculation.py
Samoed: Created an issue https://github.com/meta-pytorch/torchcodec/issues/1363 It is possible to get full video from 
```python
import mteb

task = mteb.get_task('MSVDV2TRetrieval')
task.load_data()
videos = task.dataset['default']['test']['queries']['video']
v = videos[184]

# Get the filename extension from the path stored in the dataset
filename = v._hf_encoded['path']  # e.g. something like "vid184.avi"
print(f"Original filename: {filename}")

# Save the raw bytes to a file
with open(filename, 'wb') as f:
    f.write(v._hf_encoded['bytes'])

print(f"Saved {len(v._hf_encoded['bytes'])} bytes to {filename}")
```

PR #4636, opened 2026-05-08
[MVEB] Add RAVDESS-AV V/VA PairClassification

stef41:
Adds RAVDESS-AV to PC (one of the five from #4130).

Same-emotion / different-emotion pairs over the Ryerson Audio-Visual Database of Emotional Speech and Song (8 emotions, acted speech and song).

| Task | Dataset | Category | Pairs |
|---|---|---|---|
| RAVDESSAVVPairClassification | `zachz/RAVDESS-AV-PC-V` | v2v | 2,048 |
| RAVDESSAVVAPairClassification | `zachz/RAVDESS-AV-PC-VA` | va2va | 2,048 |

Pairs balanced 1:1 (`random.Random(42)`), pre-baked parquet, no `dataset_transform`.

Note on actor leakage: RAVDESS uses 24 actors and the current bake balances only on emotion, so some same-actor pairs do appear. If reviewers prefer, I can re-bake with same-actor pairs filtered out (the actor id is encoded in the source `video_id`, e.g. last two digits of `03-01-XX-XX-XX-XX-AA.mp4`). Happy to push that as a follow-up.

Bundles the `va2va` category, helpers and bake script (same as the HAC PR) so this branch can be merged independently.

### Random baseline

| Task | max_ap | max_accuracy |
|---|---|---|
| V | 0.5121 | 0.5249 |
| VA | 0.5127 | 0.5234 |

Both ~0.50. PE-AV pending the NaN issue from #4434.

### Notes

- `zachz/RAVDESS-AV-PC-{V,VA}` needs an `mteb/` org transfer before merge.
- Built with `python scripts/upload_video_pair_classification.py --dataset ravdess`.

### Checklist

- [x] Fills a gap in MVEB
- [x] Runs with `mteb`
- [x] Random baseline ≈ 0.50 on both
- [ ] PE-AV (NaN-blocked)
- [x] ≤ 2048 pairs per variant

Ref: #4130


mteb/tasks/pair_classification/eng/_video_pair_helpers.py
```diff
@@ -0,0 +1,115 @@
+"""Shared helpers for deriving VideoPairClassification datasets from
+classification-style sources.
+
+Each helper takes a HuggingFace ``Dataset`` whose rows contain (video[, audio],
+class_label) and produces a balanced same-class / different-class pair dataset
+suitable for :class:`mteb.abstasks.AbsTaskPairClassification`.
```
Samoed: We use `google`-style docstrings, not rst. These string not visible, but still should have correct structure
```suggestion
suitable for [`mteb.abstasks.AbsTaskPairClassification`][`mteb.abstasks.AbsTaskPairClassification`].
```

mteb/tasks/pair_classification/eng/_video_pair_helpers.py
Samoed, line 1: Move all "helpers" to task files

scripts/upload_video_pair_classification.py
```diff
@@ -0,0 +1,170 @@
...
+        "RAVDESS_AV",
+        "mteb/RAVDESS_AV",
+        "13af08387c3ce5e86c179a3718eb158669268d65",
+        "emotion",
+        "RAVDESS-AV-PC",
+    ),
+}
+
+
+def push_video_dataset(
```
Samoed: Why don't you simply use `datasets.push_to_hub`, or tell this your LLM?

Samoed: Also you can updaet `load_dataset` of tasks and then use `task.push_datset_to_hub`

Samoed:
Please, don't force push. This is hard to review

scripts/upload_video_pair_classification.py
stef41: the pairing is a one-shot offline step so it didn't really fit `load_data`. can move it task-side if you'd rather

stef41:
fixed and pushed on top, no force push. same commit on #4632 #4633 #4634 #4635

PR #4640, opened 2026-05-08
fix: make datasets v2 compatible

### LINKED ISSUE #3424: Make all `retrieval` datasets compatible with v2 format

In `v1` retrieval datsets stored data in `self.queries`, `self.documents`, `self.qrels`. In `v2` we've changed to 
```python
dataset = {
    "subset": {
        "queries": Dataset,
        "corpus": Dataset,
        "relevant_queries": dict[str, dict[str, float]]
    }
}
```
To do this you can use `task.push_dataset_to_hub()`

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/3424

I tried to refactor retrieval dataset loading with agents to match our new format. I looked on some of the implementations and run some tasks and didn't find difference in dataset construction

### DIFF
--- a/mteb/tasks/instruction_reranking/multilingual/m_follow_ir.py
+++ b/mteb/tasks/instruction_reranking/multilingual/m_follow_ir.py
@@ -1,6 +1,7 @@
 from logging import getLogger
 
 import datasets
+from datasets import Dataset
 
 from mteb._evaluators.retrieval_metrics import evaluate_p_mrr_change
 from mteb.abstasks import AbsTaskRetrieval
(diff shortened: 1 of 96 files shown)


KennethEnevoldsen:
is there a reason why we wouldn't want to re-upload them instead?

Samoed:
Yes, we can and this is probably better solution, but this would require a lot of resources, but I think I could try to do that.

KennethEnevoldsen:
If you have a script, I can also set it to run (we have an old server that is decent on CPUs + memory, but the GPUs are kinda outdated)

Samoed:
I don't have a script for uploading, because it will be mostly just `task.push_dataset_to_hub`, but I have a script for listing tasks with custom `load_data` in https://github.com/embeddings-benchmark/mteb/issues/3424 descripton

PR #4642, opened 2026-05-09
fix: Activate strict mypy

Samoed:
Activate strict mypy and update it to `v2.0.0`


mteb/models/modality_collators.py
```diff
@@ -226,7 +230,7 @@ def resample_video(
         else:
             frame_step = max(1, num_source_frames // target)
             selected_frames = list(range(0, num_source_frames, frame_step))[:target]
-        return video.get_frames_at(selected_frames).data
+        return cast("torch.Tensor", video.get_frames_at(selected_frames).data)
```
KennethEnevoldsen: I would prefer not to have casts in the return statement (it adds a layer of analysis when reading the output)

KennethEnevoldsen (commented):
only a minor comment otherwise I think it is fine

mteb/models/modality_collators.py
Samoed: Yes, but otherwise this would have ignore comment, because mypy suggest that torchcodec is untyped, because they don't have `py.typed` https://github.com/meta-pytorch/torchcodec/pull/1258 in case of sentence transformer wrappers in would narrow types

KennethEnevoldsen: but what about simply doing:
```suggestion
        frames = video.get_frames_at(selected_frames).data
        frames =  cast("torch.Tensor", frames)
        return frames
```

mteb/abstasks/task_metadata.py
```diff
@@ -826,7 +826,7 @@ def generate_dataset_card(
             template_path=str(path),
             **template_kwargs,
         )
-        return dataset_card
+        return cast("DatasetCard", dataset_card)
```
KennethEnevoldsen: the cast on return statement still seem to be there

Samoed: Removed most of `return cast`

KennethEnevoldsen:
The static type checks seems to be failing in a few places - but once they are fixes then I think it is good

KennethEnevoldsen (approved):

PR #4645, opened 2026-05-09
fix: PC multimodal support

Samoed:
Ref https://github.com/embeddings-benchmark/mteb/issues/4130#issuecomment-4413282731

Currently there is a bug when running multimoal PC task with `input_column` defined as `str`. When dataset have both `text` and `video` modalities it will try to get `video` collumn in dataloader creation, because `get_modality` will return both `video` and `text` instead just `text`.


mteb/_create_dataloaders.py
```diff
@@ -258,7 +258,7 @@ def _prepare_dataset(
     """
     modalities = task_metadata.get_modalities(prompt_type)
 
-    if "text" in modalities:
+    if "text" in modalities and task_metadata.simplified_task_type == "retrieval":
```
Samoed: If we pass `prompt_type` from PC task `_corpus_to_dict` will try to get `id` from each row, which is optional for PC tasks

mteb/_evaluators/pair_classification_evaluator.py
```diff
@@ -130,10 +134,13 @@ def __call__(
             create_dataloader(
                 self.dataset.select_columns(cols2).rename_columns(ds2_col_names),
                 task_metadata=self.task_metadata,
-                input_column=self.task_metadata.modalities[0]
+                input_column=self.task_metadata.modalities[
+                    0 if len(self.task_metadata.modalities) == 1 else 1
+                ]
```
AdnanElAssadi:
```suggestion
                input_column=self.task_metadata.get_modalities(self.input2_prompt_type)[0]
```

mteb/_evaluators/pair_classification_evaluator.py
```diff
@@ -130,10 +134,13 @@ def __call__(
...
                 if isinstance(self.input2_column_name, str)
                 else None,
                 num_proc=num_proc,
+                prompt_type=PromptType.document,  # explicitly pass prompt type to correctly get modality
```
AdnanElAssadi:
```suggestion
                prompt_type=self.input2_prompt_type
```

mteb/_evaluators/pair_classification_evaluator.py
```diff
@@ -120,7 +122,9 @@ def __call__(
         if isinstance(self.input2_column_name, str):
             cols2: str | list[str] = self.input2_column_name
             ds2_col_names: dict[str, str] = {
-                self.input2_column_name: self.task_metadata.modalities[0]
+                self.input2_column_name: self.task_metadata.modalities[
+                    0 if len(self.task_metadata.modalities) == 1 else 1
+                ]
```
AdnanElAssadi:
```suggestion
                self.input2_column_name: self.task_metadata.get_modalities(self.input2_prompt_type)[0]
```

mteb/_evaluators/pair_classification_evaluator.py
AdnanElAssadi: We can do the same for input1

mteb/tasks/pair_classification/eng/videocon_pc.py
```diff
@@ -44,8 +44,8 @@ class VideoConPairClassification(AbsTaskPairClassification):
 """,
     )
 
-    input1_column_name: str = "video"
-    input2_column_name: str = "text"
+    input1_column_name = (("video", "video"),)
+    input2_column_name = (("text", "text"),)
```
AdnanElAssadi:
```suggestion
    input1_column_name: Sequence[tuple[str, Modalities]] = (("video", "video"),)
    input2_column_name: Sequence[tuple[str, Modalities]] = (("text", "text"),)
```

tests/mock_tasks.py
AdnanElAssadi, line 1: Do we add a video-text mock as well?

mteb/_create_dataloaders.py
AdnanElAssadi: should we add a comment explaining?

mteb/_evaluators/pair_classification_evaluator.py
Samoed: This won't work for tasks with more than 2 modaities and `self.input2_prompt_type` is None

mteb/_evaluators/pair_classification_evaluator.py
Samoed: Prompt type here needed only to get correct modality. I think with `input2_prompt_type` this can break something

mteb/tasks/pair_classification/eng/videocon_pc.py
Samoed: This is not nessary for child classes

mteb/_evaluators/pair_classification_evaluator.py
AdnanElAssadi: i was just worried about ordering assumption

Samoed: What do you mean?

mteb/_evaluators/pair_classification_evaluator.py
```diff
@@ -120,7 +122,9 @@ def __call__(
         if isinstance(self.input2_column_name, str):
             cols2: str | list[str] = self.input2_column_name
             ds2_col_names: dict[str, str] = {
-                self.input2_column_name: self.task_metadata.modalities[0]
+                self.input2_column_name: self.task_metadata.modalities[
+                    0 if len(self.task_metadata.modalities) == 1 else 1
```
Samoed: Other solution would be to use format `input_col: str` for tasks with only one modality, and for multimodal tasks use `input_col: Sequence[tuple[str, str]]`

mteb/_evaluators/pair_classification_evaluator.py
AdnanElAssadi: If this happens:
```
modalities = ["text", "video"]  
category = "v2t"                 
input1_column_name = "video"
input2_column_name = "text"
```

How about getting modalilties using this then: `get_modalities(prompt_type)`

mteb/_evaluators/pair_classification_evaluator.py
AdnanElAssadi: why doesn't it break in `pair_classification.py`?

```
        if isinstance(self.input1_column_name, str):
            modality1 = self.metadata.get_modalities(self.input1_prompt_type)[0]
            col_modalities1: list[tuple[str, str]] = [
                (self.input1_column_name, modality1)
            ]
        else:
            col_modalities1 = list(self.input1_column_name)  # type: ignore[arg-type]

        if isinstance(self.input2_column_name, str):
            modality2 = self.metadata.get_modalities(self.input2_prompt_type)[0]
            col_modalities2: list[tuple[str, str]] = [
                (self.input2_column_name, modality2)
            ]
        else:
            col_modalities2 = list(self.input2_column_name)  # type: ignore[arg-type]

```

mteb/_evaluators/pair_classification_evaluator.py
Samoed: > How about getting modalilties using this then: get_modalities(prompt_type)

They're taken like this in collator. I think would be easier just to prohibit using multimodal datasets with `input2_column_name: str` https://github.com/embeddings-benchmark/mteb/pull/4645#discussion_r3213982626

mteb/_evaluators/pair_classification_evaluator.py
Samoed: Hm. I'll check how correct statistics computed then

mteb/_evaluators/pair_classification_evaluator.py
KennethEnevoldsen: I think I like this solution better

mteb/_create_dataloaders.py
KennethEnevoldsen: this change seems weird to me - shouldn't we be able to derive this from the data not the metadata

mteb/_evaluators/pair_classification_evaluator.py
Samoed: I checked and this methd woudn't work with cases when pairclassification task would have same prefix, e.g. `input1_type = PromptType.query, input2_type = PromptType.query`

KennethEnevoldsen (approved):
Should we also change the examples in the docs? (or add one)

Samoed:
Added doc example

docs/contributing/adding_a_dataset.md
```diff
@@ -281,6 +281,62 @@ Once we have decided on task, we can implement them as follows:
...
+    @article{bansal2023videocon,
+      author = {Bansal, Hritik and Bitton, Yonatan and Szpektor, Idan and Chang, Kai-Wei and Grover, Aditya},
+      journal = {arXiv preprint arXiv:2311.10111},
+      title = {VideoCon: Robust Video-Language Alignment via Contrast Captions},
+      year = {2023},
+    }
+    """,
+        )
+
+        input1_column_name = (("video", "video"),)
```
KennethEnevoldsen: Why not just:
("video", "video")

Samoed: PairClassification expect `str` for unimodal tasks and `Sequence[tuple[str, Modalities]]` for multimodal tasks

KennethEnevoldsen: but why `Sequence[tuple[str, Modalities]]` instead of `tuple[Modalities]`

Samoed: We need to match column names with actual modality, because there can be video1, video2, etc.

KennethEnevoldsen: Ahhh! And we don't think a mapping is a better expression of this?

Samoed: Changed to str | Mapping

Samoed:
@KennethEnevoldsen Can we merge this pr?

PR #4648, opened 2026-05-10
ci: Speedup CI

Samoed:
Changed to `uv sync --frozen`, because `uv lock` takes a lot of time: 30 minutes on CI and 20 minutes locally for me. Also moved pinning uv version from workflow files to pyproject


Samoed:
Speed up on 10 min, not very much

isaac-chung:
better than nothing

KennethEnevoldsen (approved):

PR #4649, opened 2026-05-10
Depreciate Model2VecModel and use SentenceTransformerEncoderWrapper instead

### LINKED ISSUE #2610: Change model2vec to SentenceTransformer
From their usage docs, it looks like `SentenceTransformers` supports them out of the box: https://github.com/MinishLab/model2vec/blob/main/docs/usage.md#inference-with-the-sentence-transformers-library. I think this would simplify usage quite a bit

ayush1298:
closes #2610

Depreciating `Model2VecModel` as all models using it are now supported by `Sentence-transformers,` so added a warning for depreciation and changed the loader at all places to `SentenceTransformerEncoderWrapper`


mteb/models/model_implementations/model2vec_models.py
```diff
@@ -515,7 +523,7 @@ def encode(
 )
 
 pubmed_bert_2m = ModelMeta(
-    loader=Model2VecModel,
+    loader=SentenceTransformerEncoderWrapper,
     name="NeuML/pubmedbert-base-embeddings-2M",
```
Samoed, line 517: I think this revision isn't supported and you should update it

Samoed (commented):
Can you verify that scores haven't changed? You also need to remove `extra_group`

mteb/models/model_implementations/model2vec_models.py
ayush1298: updated revision for all `NeuML` models.

Samoed (commented):
I meant to remove `extra_requirements_groups=["model2vec"],` without changed in pyproject

ayush1298:
> I meant to remove `extra_requirements_groups=["model2vec"],` without changed in pyproject

sorry, my bad. Corrected it now.

ayush1298:
@Samoed getting some very long error when running `NeuML/pubmedbert-base-embeddings-100K`. The sentence-transformers version itself is not working and giving same error.

Samoed:
@tomaarsen Do you have ideas why 100k model failed?

```python


model = SentenceTransformer("minishlab/potion-base-8M")  # worked
model = SentenceTransformer("minishlab/M2V_base_glove")  # worked
model = SentenceTransformer("NeuML/pubmedbert-base-embeddings-8M")  # worked
model = SentenceTransformer("NeuML/pubmedbert-base-embeddings-100K")

#   File "/Users/samoed/Desktop/mteb/.venv/lib/python3.14/site-packages/sentence_transformers/sentence_transformer/modules/static_embedding.py", line 96, in __init__
#     self.embedding = nn.EmbeddingBag.from_pretrained(embedding_weights, freeze=False)
#                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#   File "/Users/samoed/Desktop/mteb/.venv/lib/python3.14/site-packages/torch/nn/modules/sparse.py", line 536, in from_pretrained
#     embeddingbag = cls(
#         num_embeddings=rows,
#     ...<8 lines>...
#         padding_idx=padding_idx,
#     )
#   File "/Users/samoed/Desktop/mteb/.venv/lib/python3.14/site-packages/torch/nn/modules/sparse.py", line 416, in __init__
#     self.weight = Parameter(_weight)
#                   ~~~~~~~~~^^^^^^^^^
#   File "/Users/samoed/Desktop/mteb/.venv/lib/python3.14/site-packages/torch/nn/parameter.py", line 57, in __new__
#     return torch.Tensor._make_subclass(cls, data, requires_grad)
#            ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
# RuntimeError: Only Tensors of floating point and complex dtype can require gradients
```

It seems that `nn.EmbeddingBag` supports only float, but `pubmedbert-base-embeddings-100K` has weights in int

tomaarsen:
It does seem that way. The model card also describes that the weights are stored as int16 to try and make the model smaller. This apparently can't be loaded into EmbeddingBag, even when setting `freeze=True` (which would indicate that it doesn't need gradients). I'm not sure what the best way to fix that would be.

- Tom Aarsen

Samoed:
I think we can keep it with `model2vec` then

ayush1298:
> @tomaarsen Do you have ideas why 100k model failed?
> 
> ```python
> from sentence_transformers import SentenceTransformer
> 
> 
> model = SentenceTransformer("minishlab/potion-base-8M")  # worked
> model = SentenceTransformer("minishlab/M2V_base_glove")  # worked
> model = SentenceTransformer("NeuML/pubmedbert-base-embeddings-8M")  # worked
> model = SentenceTransformer("NeuML/pubmedbert-base-embeddings-100K")
> 
1M and 500k versions are also failing with same error.

Samoed:
Yeah, it seems that they're storing weights in ints

ayush1298:
change loader for 100k, 500k and 1M variant to `Model2VecModel`

ayush1298:
Results are as follows , here the metric reported is `main_score`  and task is `SciFact:` 


model | old results | new results |
| -------- | --------- | ----------- |
NeuML/pubmedbert-base-embeddings-2M | 0.54784 | 0.54843 
NeuML/pubmedbert-base-embeddings-8M | 0.60252 | 0.60458 

I am somehow not able to run `minishlab/potion-base-8M` . I am getting: 
```
ValueError: The checkpoint you are trying to load has model type `model2vec` but Transformers does not recognize this architecture. This could be because of an issue with the checkpoint, or because your version of Transformers is out of date.

You can update Transformers with the command `pip install --upgrade transformers`. If this does not work, and the checkpoint is very new, then there may not be a release version that supports this model yet. In this case, you can get the most up-to-date code by installing Transformers from source with the command `pip install git+[https://github.com/huggingface/transformers.git`](https://github.com/huggingface/transformers.git%60)
```
I tried upgrading both `transformers` as well as `sentence-transformers`

 @Samoed Can you check its score?

Samoed:
Now this potion models should be working. New revision required. I got same results as you, which is fine I guess

mteb/models/model_implementations/model2vec_models.py
```diff
@@ -362,7 +362,7 @@ def encode(
     model_type=["dense"],
     languages=["eng-Latn"],
     open_weights=True,
-    revision="fc914ed07732443f18799169a2571d427209c51d",
```
Samoed, line 365: `minishlab/potion-base-*` don't require updating revision

ayush1298: just updated if anything added in recent commits. Idt it will cause any issue.

Samoed: Yes, let's keep with new I think

ayush1298:
@KennethEnevoldsen Could you review this one?

Samoed:
@ayush1298 All models are working correctly?

ayush1298:
> @ayush1298 All models are working correctly?

yes, I have also checked `rasgaard` and `andersborges` models, and they are also working correctly.

Samoed (approved):

KennethEnevoldsen:
Yeah good to merge - should we let them know that their model is not sentence-transformers compatible?

ayush1298:
> Yeah good to merge - should we let them know that their model is not sentence-transformers compatible?

maybe we can, but I don't think these are that much popular that we should concern about them

PR #4655, opened 2026-05-11
fix: Save results for every `hf_subset` instead of task

### LINKED ISSUE #4612: Save results every `hf_subset` not just every task

Currently, we save the results once the task is completed, but that can often take a while, and especially for larger tasks, you can get 90% and have to start over. Saving every subset allows you to restart from as far in as possible.

ayush1298:
closes #4612 

Previously we are saving results after each task run completes, which becomes waste if results become interrupt/crash in between. To avoid these, now we can save results for every `hf_subset` which will avoid redundant recomputations in case of crash and also save things efficiently.  Because, every time we are just saving currently computed `hf_subset` results to already computed results in cache and also after crash when it will resume then these subset will not be run again.

### DIFF
--- a/mteb/abstasks/abstask.py
+++ b/mteb/abstasks/abstask.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 import json
 import logging
 import tempfile
@@ -26,6 +27,7 @@ from mteb.models import (
     EncoderProtocol,
     SearchProtocol,
 )
+from mteb.results import TaskResult
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -33,9 +35,11 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.cache import ResultCache
     from mteb.models import (
         MTEBModels,
     )
+    from mteb.models.model_meta import ModelMeta
     from mteb.types import EncodeKwargs, HFSubset, Modalities, ScoresDict
     from mteb.types.statistics import DescriptiveStatistics, SplitDescriptiveStatistics
 
@@ -147,6 +151,8 @@ class AbsTask(ABC):  # noqa: PLR0904
         encode_kwargs: EncodeKwargs,
         prediction_folder: Path | None = None,
         num_proc: int | None = None,
+        cache: ResultCache | None = None,
+        model_meta: ModelMeta | None = None,
         **kwargs: Any,
     ) -> Mapping[HFSubset, ScoresDict]:
         """Evaluates an MTEB compatible model on the task.
@@ -158,6 +164,8 @@ class AbsTask(ABC):  # noqa: PLR0904
             encode_kwargs: Additional keyword arguments that are passed to the model's `encode` method.
             prediction_folder: Folder to save model predictions
             num_proc: Number of processes to use for loading the dataset or processing.
+            cache: Cache to save results to after evaluating each subset.
+            model_meta: Model meta required to save subset results in the cache.
             kwargs: Additional keyword arguments that are passed to the _evaluate_subset method.
 
         Returns:
@@ -217,6 +225,37 @@ class AbsTask(ABC):  # noqa: PLR0904
                 **kwargs,
             )
             self._add_main_score(scores[hf_subset])
+
+            if cache is not None and model_meta is not None:
+                try:
+                    # Create a TaskResult with ONLY the subset that was just evaluated
+                    only_current_subset = {hf_subset: scores[hf_subset]}
+                    partial_task_results = {split: only_current_subset}
+                    new_result = TaskResult.from_task_results(
+                        self,
+                        partial_task_results,
+                        evaluation_time=0.0,
+                        kg_co2_emissions=None,
+                        date=datetime.datetime.now(tz=datetime.timezone.utc),
+                    )
+
+                    existing_cached = cache.load_task_result(
+                        self.metadata.name, model_meta
+                    )
+
+                    # Merge: existing subsets + new subset
+                    if existing_cached is not None:
+                        final_result = new_result.merge(existing_cached)
+                    else:
+                        final_result = new_result
+
+                    cache.save_to_cache(final_result, model_meta)
+                    logger.debug(
+                        f"Saved intermediate results after evaluating {self.metadata.name} on {split} subset {hf_subset}"
+                    )
+                except Exception as e:
+                    logger.warning(f"Failed to save intermediate results: {e}")
+
         return scores
 
     @abstractmethod
--- a/mteb/abstasks/abstask.py
+++ b/mteb/abstasks/abstask.py
@@ -33,11 +33,9 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from mteb.abstasks.task_metadata import TaskMetadata
-    from mteb.cache import ResultCache
     from mteb.models import (
         MTEBModels,
     )
-    from mteb.models.model_meta import ModelMeta
     from mteb.types import EncodeKwargs, HFSubset, Modalities, ScoresDict
     from mteb.types.statistics import DescriptiveStatistics, SplitDescriptiveStatistics
 
@@ -149,8 +147,6 @@ class AbsTask(ABC):  # noqa: PLR0904
         encode_kwargs: EncodeKwargs,
         prediction_folder: Path | None = None,
         num_proc: int | None = None,
-        cache: ResultCache | None = None,
-        model_meta: ModelMeta | None = None,
         **kwargs: Any,
     ) -> Mapping[HFSubset, ScoresDict]:
         """Evaluates an MTEB compatible model on the task.
@@ -162,8 +158,6 @@ class AbsTask(ABC):  # noqa: PLR0904
             encode_kwargs: Additional keyword arguments that are passed to the model's `encode` method.
             prediction_folder: Folder to save model predictions
             num_proc: Number of processes to use for loading the dataset or processing.
-            cache: Cache to save results to after evaluating each subset.
-            model_meta: Model meta required to save subset results in the cache.
             kwargs: Additional keyword arguments that are passed to the _evaluate_subset method.
 
         Returns:
@@ -223,7 +217,6 @@ class AbsTask(ABC):  # noqa: PLR0904
                 **kwargs,
             )
             self._add_main_score(scores[hf_subset])
-
         return scores
 
     @abstractmethod
--- a/mteb/abstasks/abstask.py
+++ b/mteb/abstasks/abstask.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import datetime
 import json
 import logging
 import tempfile
@@ -27,7 +26,6 @@ from mteb.models import (
     EncoderProtocol,
     SearchProtocol,
 )
-from mteb.results import TaskResult
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -225,36 +223,6 @@ class AbsTask(ABC):  # noqa: PLR0904
                 **kwargs,
             )
             self._add_main_score(scores[hf_subset])
-
-            if cache is not None and model_meta is not None:
-                try:
-                    # Create a TaskResult with ONLY the subset that was just evaluated
-                    current_subset = {hf_subset: scores[hf_subset]}
-                    partial_task_results = {split: current_subset}
-                    new_result = TaskResult.from_task_results(
-                        self,
-                        partial_task_results,
-                        evaluation_time=0.0,
-                        kg_co2_emissions=None,
-                        date=datetime.datetime.now(tz=datetime.timezone.utc),
-                    )
-
-                    existing_cached = cache.load_task_result(
-                        self.metadata.name, model_meta
-                    )
-
-                    # Merge: existing subsets + new subset
-                    if existing_cached is not None:
-                        final_result = new_result.merge(existing_cached)
-                    else:
-                        final_result = new_result
-
-                    cache.save_to_cache(final_result, model_meta)
-                    logger.debug(
-                        f"Saved intermediate results after evaluating {self.metadata.name} on {split} subset {hf_subset}"
-                    )
-                except Exception as e:
-                    logger.warning(f"Failed to save intermediate results: {e}")
 
         return scores
 
--- a/mteb/abstasks/abstask.py
+++ b/mteb/abstasks/abstask.py
@@ -229,8 +229,8 @@ class AbsTask(ABC):  # noqa: PLR0904
             if cache is not None and model_meta is not None:
                 try:
                     # Create a TaskResult with ONLY the subset that was just evaluated
-                    only_current_subset = {hf_subset: scores[hf_subset]}
-                    partial_task_results = {split: only_current_subset}
+                    current_subset = {hf_subset: scores[hf_subset]}
+                    partial_task_results = {split: current_subset}
                     new_result = TaskResult.from_task_results(
                         self,
                         partial_task_results,
--- a/mteb/evaluate.py
+++ b/mteb/evaluate.py
@@ -11,7 +11,7 @@ from datasets.exceptions import DatasetNotFoundError
 from tqdm.auto import tqdm
 
 from mteb._helpful_enum import HelpfulStrEnum
-from mteb.abstasks import AbsTaskRetrieval
+from mteb.abstasks import AbsTaskBitextMining, AbsTaskRetrieval
 from mteb.abstasks.abstask import AbsTask
 from mteb.abstasks.aggregated_task import AbsTaskAggregate
 from mteb.benchmarks.benchmark import Benchmark
@@ -26,14 +26,14 @@ from mteb.results.task_result import TaskError
 from mteb.types import PromptType
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable
+    from collections.abc import Iterable, Mapping
 
     from sentence_transformers import CrossEncoder, SentenceTransformer
 
     from mteb.models.models_protocols import (
         MTEBModels,
     )
-    from mteb.types import EncodeKwargs, HFSubset, SplitName
+    from mteb.types import EncodeKwargs, HFSubset, ScoresDict, SplitName
     from mteb.types._metadata import ModelName, Revision
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


mteb/abstasks/abstask.py
```diff
@@ -147,6 +151,8 @@ def evaluate(
         encode_kwargs: EncodeKwargs,
         prediction_folder: Path | None = None,
         num_proc: int | None = None,
+        cache: ResultCache | None = None,
+        model_meta: ModelMeta | None = None,
```
Samoed: I'm not sure if we want to use this approach, because this would require to update all `evaluate` methods

Samoed:
Seems your imports are broken

mteb/abstasks/abstask.py
ayush1298: Yes, I just saw these later. I though every tasks that override `evaluate` will be using `super().evaluate()`, so they will run these `evaluate` first. What would be a better solution then?
Or can we convert these to a helper function and then just call it in each of the `evaluate` methods?

Samoed: > I though every tasks that override evaluate will be using super().evaluate(), so they will run these evaluate first

I think this is not possible for classification and bitextmining tasks

> can we convert these to a helper function and then just call it in each of the evaluate methods?

We can, but I'm unsure if we want to pass `cache` to `evaluate`. Also, you can get `model_meta` from model object

ayush1298: agree for `model_meta.` Will there be any problem if we pass `cache` to `evaluate,` as it will be an optional parameter. 
Will wait for @KennethEnevoldsen opinion here

KennethEnevoldsen: I am not sure I like this either - I think we can do this entirely in the evaluate, by simply running one split hf_subset as a time.

I believe that for a few specific tasks (bitext notably), I think it might not be possible (as there is a dependence between hf_subset since they share pair - I believe flores is one such case), but otherwise I think it should be without any performance overhead

@Samoed anything I am missing here?

ayush1298: > I am not sure I like this either - I think we can do this entirely in the evaluate, by simply running one split hf_subset as a time.

We are doing it in evaluate only. But, the problem is then we have to add to evaluate of every task. And bitex is different case.

KennethEnevoldsen: Sorry meant in evaluate.py.

What I mean was that instead of:
```
subsets_to_run=[a, b, c]
task.evaluate(model, subsets_to_run=subsets_to_run)
```

You could do:
```
subsets_to_run=[a, b, c]
for ss in subsets_to_run:
  task.evaluate(model, subsets_to_run=[ss])
```

That way you don't have to pass anything to the task

Samoed: > You could do:
> 
> ```
> subsets_to_run=[a, b, c]
> for ss in subsets_to_run:
>   task.evaluate(model, subsets_to_run=[ss])
> ```
> 
> That way you don't have to pass anything to the task

Yes, but this would probably create issues with bitext mining, because it would require more encoding

ayush1298: Can't we handle Bitex separately and do it in this way for other tasks?

ayush1298: @Samoed

Samoed: Yes, probably we can add here `if` statement to route bitextmining tasks

ayush1298: okay, will add and change the implementation

tests/test_evaluate.py
```diff
@@ -405,3 +405,75 @@ def test_mock_mmeb_tasks(task: AbsTask):
...
+        )
+        self.data_loaded = True
+
+    def _evaluate_subset(self, model, data_split, hf_split, hf_subset, **kwargs):  # noqa: PLR6301
+        if hf_subset == "subset2":
+            raise RuntimeError("Crash on subset2")
+        return {"accuracy": 0.8}
+
+
+def test_evaluate_intermediate_cache_on_crash(tmp_path: Path):
```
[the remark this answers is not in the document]
ayush1298: added test for bitex

mteb/evaluate.py
```diff
@@ -159,14 +207,56 @@ def _evaluate_task(
...
-            model,
-            split,
-            subsets_to_run=hf_subsets,
-            encode_kwargs=encode_kwargs,
-            prediction_folder=prediction_folder,
-            num_proc=num_proc,
-        )
+        # BitextMining tasks evaluate all subsets together (e.g. for parallel subsets), so they
+        # are not run subset-by-subset or incrementally cached here. Their final results are cached at the end of evaluate.
+        if isinstance(task, AbsTaskBitextMining):
```
KennethEnevoldsen, line 183: hmm having the full evaluate here is a bit extensive you can simply simplify to create the `subsets_to_run` and then simply use one evaluate call.

```
        if isinstance(task, AbsTaskBitextMining):
           hf_subsets = [[hf_subsets]]
```

mteb/evaluate.py
```diff
@@ -159,14 +207,56 @@ def _evaluate_task(
...
+                    encode_kwargs=encode_kwargs,
+                    prediction_folder=prediction_folder,
+                    num_proc=num_proc,
+                    cache=cache,
+                    model_meta=model_meta,
+                )
+                tock_ss = time()
+                if res and ss in res:
+                    task_results[split].update(res)
+                    _save_intermediate_cache(
```
KennethEnevoldsen: I don't think we need this function, we can simply move the `TaskResult` construction up and overwrite it every step.

tests/test_evaluate.py
```diff
@@ -405,3 +405,100 @@ def test_mock_mmeb_tasks(task: AbsTask):
...
+class MockCrashTask(MockClassificationTask):
+    metadata = copy(MockClassificationTask.metadata)
+    metadata.name = "MockCrashTask"
+
+    def __init__(self):
+        super().__init__()
+        self.hf_subsets = ["subset1", "subset2"]
+
+    def load_data(self, num_proc: int | None = None, **kwargs) -> None:
+        from datasets import Dataset, DatasetDict
```
KennethEnevoldsen: top level...

tests/test_evaluate.py
```diff
@@ -405,3 +405,100 @@ def test_mock_mmeb_tasks(task: AbsTask):
     pytest.importorskip("torchcodec", reason="Audio dependencies are not installed")
     model = mteb.get_model_meta("mteb/baseline-random-encoder")
     mteb.evaluate(model, task, cache=None)
+
+
+class MockCrashTask(MockClassificationTask):
+    metadata = copy(MockClassificationTask.metadata)
+    metadata.name = "MockCrashTask"
```
KennethEnevoldsen:
```suggestion
```

tests/test_evaluate.py
```diff
@@ -405,3 +405,100 @@ def test_mock_mmeb_tasks(task: AbsTask):
...
+        mteb.evaluate(model, task, cache=cache, co2_tracker=False)
+
+    path = cache.get_task_result_path(
+        task.metadata.name,
+        model.mteb_model_meta.name,
+        model.mteb_model_meta.revision,
+    )
+    assert path.exists() and path.is_file()
+
+    from mteb.results import TaskResult
```
KennethEnevoldsen: top

tests/test_evaluate.py
```diff
@@ -405,3 +405,100 @@ def test_mock_mmeb_tasks(task: AbsTask):
...
+
+    cached_result = TaskResult.from_disk(path)
```
KennethEnevoldsen: I would love using cache object

tests/test_evaluate.py
```diff
@@ -405,3 +405,100 @@ def test_mock_mmeb_tasks(task: AbsTask):
...
+    assert "test" in cached_result.scores
+    scores = cached_result.scores["test"]
+    assert len(scores) == 1
+    assert scores[0]["hf_subset"] == "subset1"
+    assert scores[0]["main_score"] == 0.8
+
+
+def test_evaluate_bitext_mining_caching(tmp_path: Path):
+    from tests.mock_tasks import MockBitextMiningTask
```
KennethEnevoldsen: top

tests/test_evaluate.py
```diff
@@ -405,3 +405,100 @@ def test_mock_mmeb_tasks(task: AbsTask):
...
+
+    cached_result = TaskResult.from_disk(path)
+    assert "test" in cached_result.scores
+    scores = cached_result.scores["test"]
+    assert len(scores) == 1
+    assert scores[0]["hf_subset"] == "subset1"
+    assert scores[0]["main_score"] == 0.8
+
+
+def test_evaluate_bitext_mining_caching(tmp_path: Path):
```
KennethEnevoldsen: I am not sure what this test is testing? (isn't is just the test for running a bitextmining task)

ayush1298: copilot mentioned to add test for `bitex` task just to ensure it works as intended

tests/test_evaluate.py
ayush1298: using `cache.load_task_result` now

mteb/evaluate.py

mteb/evaluate.py
```diff
@@ -159,14 +172,55 @@ def _evaluate_task(
...
+                subsets_to_run=batch,
+                encode_kwargs=encode_kwargs,
+                prediction_folder=prediction_folder,
+                num_proc=num_proc,
+                cache=cache,
+                model_meta=model_meta,
+            )
+            tock = time()
+
+            if res:
```
Samoed: We always would have `res`

mteb/evaluate.py
```diff
@@ -92,7 +92,10 @@ def _evaluate_task(
     encode_kwargs: EncodeKwargs,
     prediction_folder: Path | None,
     public_only: bool | None,
+    cache: ResultCache | None = None,
+    model_meta: ModelMeta | None = None,
```
Samoed: You can take it from model directly
```suggestion
```

mteb/evaluate.py
```diff
@@ -159,14 +172,55 @@ def _evaluate_task(
...
+        for batch in subsets_batches:
+            res = task.evaluate(
+                model,
+                split,
+                subsets_to_run=batch,
+                encode_kwargs=encode_kwargs,
+                prediction_folder=prediction_folder,
+                num_proc=num_proc,
+                cache=cache,
+                model_meta=model_meta,
```
Samoed: These parameters not used by task.evaluate
```suggestion
```

mteb/evaluate.py
```diff
@@ -159,14 +172,55 @@ def _evaluate_task(
...
+                if cache and model_meta:
+                    try:
+                        new_result = TaskResult.from_task_results(
+                            task,
+                            task_results,
+                            evaluation_time=evaluation_time + (tock - tick),
+                            kg_co2_emissions=None,
+                            date=datetime.datetime.now(tz=datetime.timezone.utc),
+                        )
+                        cache.save_to_cache(new_result, model_meta)
```
Samoed: This way of saving can erase existing results

mteb/evaluate.py
```diff
@@ -510,9 +570,6 @@ def evaluate(  # noqa: PLR0913, PLR0914
             exceptions=[result],
         )
 
-    if existing_results:
-        result = result.merge(existing_results)
```
Samoed, line 514: I don't think that this should be deleted

tests/test_evaluate.py
Samoed: I don't think that this works as intended, because only `task.parallel_subsets = True` should fail like this

mteb/evaluate.py
```diff
@@ -126,13 +129,23 @@ def _evaluate_task(
...
 
-    task_results = {}
+    task_results: dict[SplitName, dict[HFSubset, ScoresDict]] = {}
+    evaluation_time = 0.0
+
+    if existing_results is not None:
+        for split, scores_list in existing_results.scores.items():
+            task_results[split] = {score["hf_subset"]: score for score in scores_list}
+        if existing_results.evaluation_time:
+            evaluation_time = existing_results.evaluation_time
```
Samoed, line 156: Why not merging before saving instead?

ayush1298: we are completely overwriting the existing results which kenneth suggested.

> I don't think we need this function, we can simply move the TaskResult construction up and overwrite it every step.

I guess these should work as intended

mteb/evaluate.py
ayush1298: wont it will overwrite instead of erase?

tests/test_evaluate.py
ayush1298: So, do we really need tests for `bitex` or should I remove it?

KennethEnevoldsen: I at least think that this test is irrelevant

ayush1298: okay, will remove it

mteb/evaluate.py
```diff
@@ -203,59 +168,43 @@ def _evaluate_task(  # noqa: PLR0913
...
+                subsets_to_run=batch,
+                encode_kwargs=encode_kwargs,
+                prediction_folder=prediction_folder,
+                num_proc=num_proc,
+            )
+            tock_ss = time()
+            task_results[split].update(res)
+            # Save intermediate cache
+            if cache and model_meta:
+                try:
```
KennethEnevoldsen: we def. don't want a try except here - we could spend all the runtime and not get a result out, we would much rather fail early

mteb/evaluate.py
```diff
@@ -203,59 +168,43 @@ def _evaluate_task(  # noqa: PLR0913
...
+                split,
+                subsets_to_run=batch,
+                encode_kwargs=encode_kwargs,
+                prediction_folder=prediction_folder,
+                num_proc=num_proc,
+            )
+            tock_ss = time()
+            task_results[split].update(res)
+            # Save intermediate cache
+            if cache and model_meta:
```
KennethEnevoldsen: don't we always have a model meta?

mteb/evaluate.py
```diff
@@ -604,9 +553,6 @@ def evaluate(  # noqa: PLR0913, PLR0914
             exceptions=[result],
         )
 
-    if existing_results:
```
KennethEnevoldsen, line 513: doesn't this remove merging?

mteb/evaluate.py
```diff
@@ -573,8 +522,8 @@ def evaluate(  # noqa: PLR0913, PLR0914
                 prediction_folder=prediction_folder,
                 public_only=public_only,
                 cache=cache,
-                model_meta=meta,
```
KennethEnevoldsen: I am not sure why this was removed?

KennethEnevoldsen: can we garauntee that is always accessible in the model?

mteb/evaluate.py
ayush1298: copilot suggested it, will remove it

mteb/evaluate.py
ayush1298: Where these is exactly, I guess you are referring to changes from old commits which were already changed

KennethEnevoldsen: indeed! will resolve

mteb/evaluate.py
ayush1298: No, it will not. So, the difference is previously, 
`_evaluate_task` was executed on only the `missing_eval` splits and subsets, returning a `TaskResult` with only those new splits. `evaluate()` would then merge this new result with the loaded `existing_results` using the above `result.merge(existing_results)`.
But now, we are passing `existing_results` directly into `_evaluate_task` and we are prepopulating `task_results` with these `existing_results`. And when new splits/subsets are evaluated they are directly added into the same `task_results`. So, no need of merging again.

Also, made some more changes to handle co2 emission and `mteb_version` thing

KennethEnevoldsen: Hmm might be worth adding a test on this for the future

ayush1298: we have `test_evaluate_w_missing_splits` and `test_evaluate_w_missing_subsets`. I guess they are checking these merging is happen correctly or not. Do, we need any other test?

KennethEnevoldsen: I was thinking for the co2 emissions and mteb_version

ayush1298: okay, will add it

ayush1298: added test

ayush1298:
@Samoed @KennethEnevoldsen Anything to add here?

mteb/results/task_result.py
```diff
@@ -215,14 +215,14 @@ def from_task_results(
...
-                    "mteb_version": mteb_ver,
+                    "mteb_version": hf_scores.get("mteb_version", mteb_ver),
                 }
                 flat_scores[split].append(_scores)
 
         return TaskResult(
             dataset_revision=task.metadata.revision,
             task_name=task.metadata.name,
-            mteb_version=mteb_ver,
+            mteb_version=cls._compute_top_level_mteb_version(flat_scores) or mteb_ver,
```
KennethEnevoldsen, line 225: I am not sure I understand these changes

mteb/results/task_result.py
```diff
@@ -837,18 +837,23 @@ def merge(
                 merged_scores[split] = scores
 
         existing_kg_co2_emissions = (
-            self.kg_co2_emissions if self.kg_co2_emissions else 0
+            self.kg_co2_emissions if self.kg_co2_emissions else 0.0
```
KennethEnevoldsen: Similarly here (seems like they are not required

KennethEnevoldsen (approved):
A few minor things

mteb/results/task_result.py
ayush1298: It's more of required when we are merging or resuming runs

KennethEnevoldsen: why is that the case?

ayush1298: So, the thing is we have `mteb_version` field at the top-level of file. So, if in case subsets were evaluated with different versions, then if we just use `mteb_version=mteb_ver,` we will lose that different version thing, so to avoid that, this is added.

mteb/evaluate.py
```diff
@@ -155,18 +176,38 @@ def _evaluate_task(
...
-            encode_kwargs=encode_kwargs,
-            prediction_folder=prediction_folder,
-            num_proc=num_proc,
-        )
+        # BitextMining tasks evaluate all subsets together (e.g. for parallel subsets), so they
+        # are not run subset-by-subset. Other tasks are run subset-by-subset for intermediate caching.
+        if isinstance(task, AbsTaskBitextMining):
+            subsets_batches = [hf_subsets]
+        else:
+            subsets_batches = [[ss] for ss in hf_subsets]
```
Samoed, line 186: I think `subset_batches` is a bit misleading, because we process them one by one

tests/test_evaluate.py
```diff
@@ -405,3 +408,108 @@ def test_mock_mmeb_tasks(task: AbsTask):
     pytest.importorskip("torchcodec", reason="Audio dependencies are not installed")
     model = mteb.get_model_meta("mteb/baseline-random-encoder")
     mteb.evaluate(model, task, cache=None)
+
+
+class MockCrashTask(MockClassificationTask):
```
Samoed: You can inherit from `MockMultilingualClassificationTask` to remove `load_data`

tests/test_evaluate.py
```diff
@@ -405,3 +408,108 @@ def test_mock_mmeb_tasks(task: AbsTask):
...
+    cached_result = cache.load_task_result(task.metadata.name, model.mteb_model_meta)
+    assert cached_result is not None
+    assert "test" in cached_result.scores
+    scores = cached_result.scores["test"]
+    assert len(scores) == 1
+    assert scores[0]["hf_subset"] == "subset1"
+    assert scores[0]["main_score"] == 0.8
+
+
+def test_task_result_from_task_results_merging():
```
Samoed, line 437: Don't we have such test in `test_results`?

ayush1298: I checked under `test_results`, we have such tests like: `test_merge_across_mteb_versions,` and `test_merge_without_per_subset_version` but they are not checking for co2_emission. So, should I update those tests only?

tests/test_evaluate.py
ayush1298: done

mteb/evaluate.py
ayush1298: I didn't have any better naming here. Do you have anything in mind to keep it more clear?

tests/test_evaluate.py
Samoed: Yes, I think better to update

mteb/evaluate.py
Samoed: Maybe just subsets?

ayush1298: wouldn't is be more confusing as we are getting it from `hf_subsets`

tests/test_evaluate.py
ayush1298: updated now

mteb/evaluate.py
Samoed: Let's keep as is for now

Samoed (approved):

KennethEnevoldsen (approved):

PR #4659, opened 2026-05-12
fix: add num_queries and num_documents to retrieval statistics

### LINKED ISSUE #3471: Add number of queries and documents to retrieval statistics

fzoll:
## Summary
- Add `num_queries` and `num_documents` fields to `RetrievalDescriptiveStatistics`
- These are computed alongside the existing `num_samples` (which is their sum)
- Updated all mock retrieval task expected stats

Closes #3471

## Notes
- Existing descriptive stats JSON files will be updated incrementally as tasks are re-run
- The values are already computed in `_calculate_descriptive_statistics_from_split()` but were not included in the output until now


Samoed:
Will you update older statistics like in your previous PR?

Samoed:
I created a script and filled most of the retrieval/reranking tasks with num_queries and num_documents

KennethEnevoldsen (approved):
thanks for finishing it up!

PR #4660, opened 2026-05-12
Support experiments for submit results

Samoed:
Support experiments for submit results. Example https://github.com/embeddings-benchmark/results/pull/536


ayush1298 (approved):
Looks good

ayush1298:
@Samoed Can you just change these thing also, the link was broken actually:
https://github.com/embeddings-benchmark/mteb/blob/main/mteb/cache/result_cache.py#L1392
It should be:  
`https://embeddings-benchmark.github.io/mteb/contributing/submitting_results/`

PR #4661, opened 2026-05-12
fix: restrict FSD2019Kaggle eval to curated subset only

AdnanElAssadi56:
The noisy subset has automatically-extracted labels with known label noise, it was intended as additional training data in the original FSDKaggle2019 Kaggle competition, not as a test set. Evaluating on noisy bounds lrap by the label-noise floor regardless of model quality, which confounds comparison across models.

### DIFF
--- a/mteb/tasks/multilabel_classification/eng/fsd2019_kaggle.py
+++ b/mteb/tasks/multilabel_classification/eng/fsd2019_kaggle.py
@@ -18,7 +18,7 @@ class FSD2019KaggleMultilingualClassification(AbsTaskMultilabelClassification):
         type="AudioMultilabelClassification",
         category="a2t",
         eval_splits=["test"],
-        eval_langs={"curated": ["eng-Latn"], "noisy": ["eng-Latn"]},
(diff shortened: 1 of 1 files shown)


Samoed:
I think we can keep this, just to not add to "benchmark"

AdnanElAssadi:
> I think we can keep this, just to not add to "benchmark"

how? leaderboard will aggeragate

Samoed:
> how? leaderboard will aggeragate

We can select different subsets/splits for each tasks and on leaderboard will be only selected splits and subsets https://github.com/embeddings-benchmark/mteb/blob/b8ade6e386166463965e9f788278f2db311618e0/mteb/benchmarks/benchmarks/benchmarks.py#L70-L78

AdnanElAssadi:
> > how? leaderboard will aggeragate
> 
> We can select different subsets/splits for each tasks and on leaderboard will be only selected splits and subsets
> 
> https://github.com/embeddings-benchmark/mteb/blob/b8ade6e386166463965e9f788278f2db311618e0/mteb/benchmarks/benchmarks/benchmarks.py#L70-L78

oh, got it

PR #4665, opened 2026-05-13
benchmark: Add CoREB benchmark

Geralt-Targaryen:
This PR adds the [CoREB](https://arxiv.org/abs/2605.04615) code search benchmark, which fills in the gap of missing code reranking evaluation in code search. The benchmark is built from LiveCodeBench in five programming languages and covers both retrieval and reranking in three tasks: code-to-text search, text-to-code search, code-to-code search.

The following models have been evaluated on the implementation:

| |C2TRetrieval | C2CRetrieval | T2CRetrieval | C2TReranking | C2CReranking | T2CReranking | Avg Retrieval | Avg Reranking | Avg|
|-|-|-|-|-|-|-|-|-|-|
|Qwen/Qwen3-Embedding-0.6B | 89.27 | 37.68 | 41.64 | 88.95 | 39.36 | 27.94 | 56.19 | 52.08 | 54.14|
|Qwen/Qwen3-Embedding-4B | 91.70 | 48.47 | 41.54 | 91.25 | 38.02 | 28.09 | 60.57 | 52.45 | 56.51|
|Qwen/Qwen3-Embedding-8B | 94.25 | 54.93 | 42.23 | 93.52 | 40.39 | 28.73 | 63.80 | 54.21 | 59.01|
|codefuse-ai/F2LLM-v2-80M | 83.37 | 52.79 | 40.82 | 88.47 | 41.10 | 28.43 | 58.99 | 52.66 | 55.83|
|codefuse-ai/F2LLM-v2-160M | 88.87 | 53.40 | 41.87 | 89.94 | 41.56 | 27.97 | 61.38 | 53.16 | 57.27|
|codefuse-ai/F2LLM-v2-330M | 96.69 | 54.42 | 43.13 | 97.24 | 41.43 | 27.94 | 64.75 | 55.54 | 60.14|
|codefuse-ai/F2LLM-v2-0.6B | 95.81 | 53.51 | 43.96 | 95.72 | 40.02 | 28.08 | 64.43 | 54.61 | 59.52|
|codefuse-ai/F2LLM-v2-1.7B | 97.37 | 52.60 | 44.09 | 97.18 | 38.39 | 28.38 | 64.69 | 54.65 | 59.67|
|codefuse-ai/F2LLM-v2-4B | 97.46 | 52.70 | 43.84 | 97.48 | 37.88 | 27.82 | 64.67 | 54.39 | 59.53|
[...]

PR #4671, opened 2026-05-14
fix: retrieval statistics calculation

Samoed:
Previously, average relevant docs were calculated using only relevant docs (non-zero score), while min and max used all relevant docs split (including 0). Found during reviewing https://github.com/embeddings-benchmark/mteb/pull/4665


KennethEnevoldsen (approved):
Ahh good catch!

PR #4674, opened 2026-05-15
fix: tolerate off-by-one frame metadata in resample_video

AdnanElAssadi56:
Mirror PR #4625 in the inference path: drop the trailing source-frame index so get_frames_at does not request a frame that fails to decode.

### DIFF
--- a/mteb/models/modality_collators.py
+++ b/mteb/models/modality_collators.py
@@ -202,7 +202,9 @@ class FramesCollator:
             num_frames: If set, select exactly this many frames uniformly
                 (fixed-sample mode).
         """
-        num_source_frames = video.metadata.num_frames
+        # Guard against torchcodec.num_frames over-counting by 1.
+        meta_num_frames = video.metadata.num_frames
+        num_source_frames = meta_num_frames - 1 if meta_num_frames else meta_num_frames
 
         if num_frames is None and fps is None:
             # No resampling: return all frames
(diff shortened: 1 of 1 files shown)


AdnanElAssadi:
@Samoed It is better to guard against this as it can happen if we have a high num_frames and the video is really short i.e trying to take all of it and the last frame is problematic.

mteb/models/modality_collators.py
```diff
@@ -202,7 +202,9 @@ def resample_video(
             num_frames: If set, select exactly this many frames uniformly
                 (fixed-sample mode).
         """
-        num_source_frames = video.metadata.num_frames
+        # Guard against torchcodec.num_frames over-counting by 1.
+        meta_num_frames = video.metadata.num_frames
+        num_source_frames = meta_num_frames - 1 if meta_num_frames else meta_num_frames
```
Samoed: wouldn't this be always -1?

AdnanElAssadi: yes, but i dont think removing one frame hurts; we are just sampling frames anyway. we can handle using try except or something if you dont like this

Samoed: We can leave only `meta_num_frames - 1` or use try except. Part with `if` looks a bit strange

AdnanElAssadi: ah, that one is just to guard against 0 - 1

Samoed: But we shouldn't have such videos

Samoed (approved):

PR #4677, opened 2026-05-15
feat: Add tracking of run settings

### LINKED ISSUE #4237: Add better tracking of run settings

Currently, we do not track specific settings when running a model, which can often influence the results.

My suggestion would be to add the following to the results:

```
# current results
|- model/revision
    |- taskname1.json
    |- ...
    |- model_meta.json
    |- run_settings.jsonl
```

with run_settings.jsonl being the run settings for each task:

```json
{"task": "taskname1", "version": {"mteb": 1.2.3, "torch": 2.8.2, "flash-attn": None, ...}, "encode_kwargs" {...}}
{"task": "taskname2", "version": {"mteb": 1.2.3, "torch": 2.8.2, "flash-attn": None, ...}, "encode_kwargs" {...}}
```

This would be great for debugging. It could be stored in other ways, this is just one suggestion. It could also be stored in the task results which might be easier, but expands the size of an object that we already use a lot.

Let me know what you think

ayush1298:
closes #4237

These PR added a functionality to create `run_settings.jsonl` file to track run settings. 
It will generate `run_settings.jsonl` files per `model/revision` and create one entry per `task/split/subset` combination.

This PR is created with help of Github Copilot.


mteb/evaluate.py
```diff
@@ -56,6 +57,24 @@ class OverwriteStrategy(HelpfulStrEnum):
     ONLY_CACHE = "only-cache"
 
 
+def _get_package_versions() -> dict[str, str | None]:
+    """Collect version information of key packages used during evaluation.
+
+    Returns:
+        A dictionary with package names as keys and versions (or None if not installed) as values.
+    """
+    packages = ["mteb", "torch", "sentence-transformers", "flash-attn", "transformers"]
```
ayush1298: I am not sure if we have to use these packages only, or add any other package. Also, should we include package from `extra_requirement_groups` for a specific model?

mteb/cache/result_cache.py
```diff
@@ -346,6 +347,24 @@ def save_to_cache(
...
+                        split=split,
+                        subset=hf_subset,
+                        version=task_result.version or {},
+                        encode_kwargs=task_result.encode_kwargs or {},
+                    )
+                    run_settings_list.append(run_settings)
+
+            if run_settings_list:
+                run_settings_path = result_path.parent / "run_settings.jsonl"
+                append_run_settings_to_file(run_settings_path, run_settings_list)
```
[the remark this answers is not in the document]
ayush1298: fix

mteb/evaluate.py
Samoed: I think we can start from these packages. I don't think that this is nessarty to include from `extra_requirement_groups`

mteb/results/task_result.py
```diff
@@ -1020,3 +1035,89 @@ class TaskError(BaseModel):
...
+        """Create RunSettings from a JSON line (JSONL format).
+
+        Args:
+            line: A JSON string from JSONL format.
+
+        Returns:
+            The created RunSettings object.
+        """
+        data = json.loads(line)
+        return cls.model_validate(data)
```
Samoed:
```suggestion
        return cls.model_validate_json(line)
```

mteb/results/task_result.py
```diff
@@ -1020,3 +1035,89 @@ class TaskError(BaseModel):
...
+    version: dict[str, str | None]
+    encode_kwargs: dict[str, Any]
+
+    def to_json_line(self) -> str:
+        """Convert RunSettings to a JSON line (JSONL format).
+
+        Returns:
+            A JSON string suitable for JSONL format (single line).
+        """
+        return json.dumps(self.model_dump(), separators=(",", ": "))
```
Samoed:
```suggestion
        return self.model_dump_json()
```

mteb/results/task_result.py
```diff
@@ -172,6 +175,8 @@ class TaskResult(BaseModel):  # noqa: PLR0904
     evaluation_time: float | None
     kg_co2_emissions: float | None = None
     date: datetime.datetime | None = None
+    version: dict[str, str | None] | None = None
+    encode_kwargs: dict[str, Any] | None = None
```
Samoed: I don't think that we should save info of libraries in `version` and that we should save `encode_kwargs` in task results

mteb/evaluate.py
ayush1298: cool !

mteb/results/task_result.py
ayush1298: How should we save them then?

Samoed: We can get them directly from env before dump jsonl

ayush1298: We can get libraries version from environment, how we can get `encode_kwargs?`
`evaluate` has `encode_kwargs`, can we pass it to `cache.save_to_cache` and from there we can pass to `RunSettings?`

Samoed: I think yes, we can both libraries and encode kwargs to there

ayush1298: I have fixed it. Can you check?

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,88 @@ class TaskError(BaseModel):
 
     task_name: str
     exception: str
+
+
+class RunSettings(BaseModel):
```
KennethEnevoldsen: I would move it to a seperate module

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,88 @@ class TaskError(BaseModel):
...
+        Args:
+            line: A JSON string from JSONL format.
+
+        Returns:
+            The created RunSettings object.
+        """
+        return cls.model_validate_json(line)
+
+
+def append_run_settings_to_file(
```
KennethEnevoldsen: what if I rerun a split?

mteb/evaluate.py
```diff
@@ -366,7 +366,11 @@ def evaluate(  # noqa: PLR0913, PLR0914
         )
         combined_results = aggregated_task.combine_task_results(results.task_results)
         if cache:
-            cache.save_to_cache(combined_results, meta)
+            cache.save_to_cache(
+                combined_results,
+                meta,
+                encode_kwargs=cast("dict[str, Any] | None", encode_kwargs),
```
KennethEnevoldsen: cast should not be in the call

KennethEnevoldsen: same below (typing should be on their own line)

mteb/cache/result_cache.py
```diff
@@ -346,6 +352,61 @@ def save_to_cache(
             with model_meta_path.open("w") as f:
                 json.dump(meta.to_dict(), f, default=str, indent=4)
 
+        version_dict = self._get_package_versions()
+
+        run_settings_list = []
+        for split, split_scores in task_result.scores.items():
+            for score_entry in split_scores:
+                hf_subset = score_entry.get("hf_subset", "default")
+                run_settings = RunSettings(
```
KennethEnevoldsen: I am not sure we need an object here. I think we just need a `json_serialize_kwargs` function and `write_to_keyed_json` function. If we run the same model twice we don't want duplicate entries - we need a test for this.

Note that encode_kwargs is not always json serializable (we probably need a test to ensure that we can handle these)

mteb/cache/result_cache.py
```diff
@@ -346,6 +352,61 @@ def save_to_cache(
...
+                for rs in existing_entries
+                if (rs.task, rs.split, rs.subset) not in new_keys
+            ]
+            all_entries = filtered_existing + run_settings_list
+            with run_settings_path.open("w", encoding="utf-8") as f:
+                for entry in all_entries:
+                    f.write(entry.to_json_line() + "\n")
+
+    @staticmethod
+    def _get_package_versions() -> dict[str, str | None]:
```
KennethEnevoldsen: this should not be on the object just place it above.

mteb/cache/result_cache.py
```diff
@@ -346,6 +352,61 @@ def save_to_cache(
...
+        """Get current package versions from the environment.
+
+        Returns:
+            A dictionary with package names as keys and versions (or None if not installed) as values.
+        """
+        from importlib.metadata import version
```
KennethEnevoldsen: top-level (I feel like I keep mentioning this - packages should always be top level if possible, especially for native modules)

KennethEnevoldsen (changes requested):

mteb/cache/result_cache.py
ayush1298: done

mteb/results/task_result.py
ayush1298: removing class

mteb/evaluate.py

mteb/results/task_result.py
ayush1298: this is already handled. If you see `write_to_keyed_json` function, it will check for new and existing entries and merged them correctly

mteb/cache/result_cache.py
ayush1298: agree. I have removed object and add helper functions.
Also added test for duplicate entries and serialization of `encode_kwargs`

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,52 @@ class TaskError(BaseModel):
 
     task_name: str
     exception: str
+
+
+def json_serialize_kwargs(kwargs: Any | None) -> Any:
+    """Convert keyword arguments into a JSON-serializable structure."""
+    return json.loads(json.dumps(kwargs, default=str)) if kwargs is not None else {}
```
Samoed: But wouldn't `json.dump` will do the same later? Also, there can be an error, so we should add try except on this

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,52 @@ class TaskError(BaseModel):
 
     task_name: str
     exception: str
+
+
+def json_serialize_kwargs(kwargs: Any | None) -> Any:
```
Samoed:
```suggestion
def json_serialize_kwargs(kwargs: dict[str, Any] | None) -> dict[str, Any]:
```

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,52 @@ class TaskError(BaseModel):
...
+            if isinstance(parsed, dict):
+                run_settings.append(parsed)
+    return run_settings
+
+
+def write_to_keyed_json(
+    path: Path,
+    entries: list[dict[str, Any]],
+    *,
+    key_fields: tuple[str, ...] = ("task", "split", "subset"),
```
Samoed: Is it expecting to be changed?

Samoed (commented):
Can you make your new helping function private?

mteb/results/task_result.py
ayush1298: no, I guess.

mteb/results/task_result.py
ayush1298: added try-except. 
I guess we have to serialize `encode_kwargs` before we can use it in `json.dump`

mteb/results/task_result.py
ayush1298: using `kwargs: Mapping[str, Any] | None` to match typing

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,59 @@ class TaskError(BaseModel):
 
     task_name: str
     exception: str
+
+
+def _json_serialize_kwargs(kwargs: Mapping[str, Any] | None) -> dict[str, Any]:
```
KennethEnevoldsen: @samoed don't we already have function for this in the experiments - seems like one that we should re-use?

tests/test_run_settings_jsonl.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_save_to_cache_replaces_existing_run_settings_entry(tmp_path, monkeypatch):
+    cache = ResultCache(cache_path=tmp_path)
+    task_result = DummyTaskResult(
+        task_name="STS12",
+        scores={"test": [{"hf_subset": "en"}]},
+    )
+
+    monkeypatch.setattr(
```
KennethEnevoldsen: why the monkeypatch here?

tests/test_run_settings_jsonl.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def _read_jsonl(path):
+    with path.open("r", encoding="utf-8") as handle:
+        return [json.loads(line) for line in handle if line.strip()]
+
+
+def test_save_to_cache_replaces_existing_run_settings_entry(tmp_path, monkeypatch):
+    cache = ResultCache(cache_path=tmp_path)
+    task_result = DummyTaskResult(
```
KennethEnevoldsen: Why not use a actual task result object?

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,59 @@ class TaskError(BaseModel):
...
+    """Convert keyword arguments into a JSON-serializable structure."""
+    if kwargs is None:
+        return {}
+
+    try:
+        return json.loads(json.dumps(kwargs, default=str))
+    except Exception as e:
+        logger.warning(f"Could not serialize encode_kwargs to JSON: {e}")
+        return {}
+
```
KennethEnevoldsen: No need for the try except - it should never fail
```
def make_json_serializable(kwargs: dict | None) -> dict:
    if kwargs is None:
        return {}
    return json.loads(json.dumps(kwargs, default=str))
```

tests/test_run_settings_jsonl.py
```diff
@@ -0,0 +1,88 @@
...
+        self.scores = scores
+
+    @staticmethod
+    def to_disk(path) -> None:
+        path.write_text("{}", encoding="utf-8")
+
+
+class UnserializableValue:
+    def __str__(self) -> str:
+        return "unserializable-value"
```
KennethEnevoldsen: you don't need the string conversion here - in the test simlpy check for that it is a string (it could seem like here that it is needed but itis not

mteb/results/task_result.py
KennethEnevoldsen: I don't think there can be an error here (it would be in the string serialization)

mteb/results/task_result.py
KennethEnevoldsen: (also see comment above)

KennethEnevoldsen (approved):
we are getting closer. 

Renamed from fix to feat as it adds a feature that user might rely on.

KennethEnevoldsen (commented):
(accidentally accepted

mteb/results/task_result.py
ayush1298: are you referring to this function? 
https://github.com/embeddings-benchmark/mteb/blob/892225eb2981060a8e68cca84c6e28e38631065e/mteb/models/model_meta.py#L1591

tests/test_run_settings_jsonl.py
ayush1298: removed

tests/test_run_settings_jsonl.py
ayush1298: done

ayush1298:
@KennethEnevoldsen I have updated test, and also using `task_result` object from `TaskResult.from_task_results`. Though I am not sure, are you referring to these thing only.

mteb/results/task_result.py
ayush1298: Its a different one and I think can't be used directly and apart from that I didn't find any other function

KennethEnevoldsen: Yeah no it doesn't seem like it - will close this

mteb/results/task_result.py
```diff
@@ -5,7 +5,8 @@
 import logging
 from collections import defaultdict
 from functools import cached_property
-from importlib.metadata import version
+from importlib.metadata import version as get_version
```
KennethEnevoldsen: why the rename here?

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,54 @@ class TaskError(BaseModel):
...
+                logger.warning(
+                    f"Could not parse run_settings line '{stripped_line}': {e}"
+                )
+                continue
+            if isinstance(parsed, dict):
+                run_settings.append(parsed)
+    return run_settings
+
+
+def _write_to_keyed_json(
```
KennethEnevoldsen:
```suggestion
def _write_and_merge_keyed_json(
```

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,54 @@ class TaskError(BaseModel):
...
+    path: Path,
+    entries: list[dict[str, Any]],
+    *,
+    key_fields: tuple[str, str, str] = ("task", "split", "subset"),
+) -> None:
+    """Write JSONL entries, replacing any existing entries with the same key."""
```
KennethEnevoldsen:
```suggestion
    """Write entries to `.jsonl`, if it already exist it will merge it, replacing any existing entries with the same key."""
```

KennethEnevoldsen (commented):
I few minor things - I think we need the what's new documentation, but otherwise I think we are good

mteb/cache/result_cache.py
```diff
@@ -50,6 +52,30 @@
 _EXPERIMENTS_FOLDER_NAME = "experiments"
 
 
+def get_package_versions() -> dict[str, str | None]:
```
Samoed:
```suggestion
def _get_package_versions() -> dict[str, str | None]:
```

mteb/cache/result_cache.py
```diff
@@ -346,6 +374,25 @@ def save_to_cache(
...
+                    "split": split,
+                    "subset": hf_subset,
+                    "version": version_dict,
+                    "encode_kwargs": _json_serialize_kwargs(encode_kwargs),
+                }
+                run_settings_list.append(run_settings)
+
+        if run_settings_list:
+            run_settings_path = result_path.parent / "run_settings.jsonl"
+            _write_to_keyed_json(run_settings_path, run_settings_list)
```
Samoed: Maybe better to move to other function?

mteb/results/task_result.py
ayush1298: was required in old implemetation. forgot to revert. Will revert it.

mteb/cache/result_cache.py
```diff
@@ -321,6 +347,7 @@ def save_to_cache(
         task_result: TaskResult,
         model_name: str | ModelMeta,
         model_revision: str | None = None,
+        encode_kwargs: Mapping[str, Any] | None = None,
```
Samoed, line 373:
```suggestion
        *,
        encode_kwargs: Mapping[str, Any] | None = None,
```

mteb/cache/result_cache.py
ayush1298: Its already a very small function, do we have to make separate function

Samoed:
I think this `run_expriments.json` wouldn't be handled by `results_cache.submit_results`

ayush1298:
> I think this `run_expriments.json` wouldn't be handled by `results_cache.submit_results`

Changed to handle in `submit_results`

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,54 @@ class TaskError(BaseModel):
 
     task_name: str
     exception: str
+
+
+def _json_serialize_kwargs(kwargs: Mapping[str, Any] | None) -> dict[str, Any]:
+    """Convert keyword arguments into a JSON-serializable structure."""
+    if kwargs is None:
+        return {}
+    return cast("dict[str, Any]", json.loads(json.dumps(kwargs, default=str)))
```
ayush1298: typecheck is giving error here, so need to add cast again

KennethEnevoldsen: I would generally recommend not having cast and functionality in the same line:


```suggestion
    json_kwargs = json.loads(json.dumps(kwargs, default=str))
    json_kwargs = cast("dict[str, Any]", json_kwargs)
    return json_kwargs
```

mteb/results/task_result.py
```diff
@@ -1020,3 +1021,54 @@ class TaskError(BaseModel):
 
     task_name: str
     exception: str
+
+
+def _json_serialize_kwargs(kwargs: Mapping[str, Any] | None) -> dict[str, Any]:
```
Samoed: Do we need this function? We'll use `json.dump` in `_write_and_merge_keyed_json`

ayush1298: removed function, but instead of adding to `_write_and_merge_keyed_json,` used it before only

Samoed: I meant if we doing `json.dump` later, why do we `json.dump` encode kwargs?

ayush1298: it may contain objects that can't be serialized, so I guess we need it

Samoed: How this would help? You'll use same json dump, but later

ayush1298: inside `_write_and_merge_keyed_json` function, we are not unfolding these `run_settings_list` which contain each entry as a dict inside which we have `encode_kwargs,` so why do all that unfolding stuff again? We can just do it when creating these `run_settings_list`

mteb/results/task_result.py
ayush1298: I have removed the function. (below comment)

KennethEnevoldsen: since we have talked about this a few times I made a rule request on ruff: 
https://github.com/astral-sh/ruff/issues/25215

mteb/results/task_result.py
ayush1298: @KennethEnevoldsen What's your opinion here?

Samoed: > it may contain objects that can't be serialized, so I guess we need it

This wouldn't help, because json.dump, will just raise an error

```py
import torch
import json
json.dumps({"dtype": torch.bfloat16})
# TypeError: Object of type dtype is not JSON serializable when serializing dict item 'dtype'
```

ayush1298:
```py
import torch
import json
encode_kwargs = {"dtype": torch.bfloat16}
print(json.dumps(encode_kwargs, default=str))
```

we are using `default=str,` and above code is running correctly

ayush1298:
Is there anything else left here?
I have added its support in `submit_results` and also updated `whats_new` docs

Samoed (approved):
Looks good to me

ayush1298:
@KennethEnevoldsen Do you want to review once again?

KennethEnevoldsen (approved):

KennethEnevoldsen:
Good catch on submit_results. I think the PR is looking good!

PR #4684, opened 2026-05-17
fix: route empty instructions through callable `instruction_template`

### LINKED ISSUE #4683: `get_task_instruction` skips `instruction_template` for empty instructions, breaking document-side prefix for ReasonIR / GritLM / Sarashina v2 / BMRetriever / Octen
### Summary

`mteb/models/abs_encoder.py:269` gates `format_instruction` behind `len(instruction) > 0`:

```python
if self.instruction_template and len(instruction) > 0:
    return self.format_instruction(instruction, prompt_type)
return instruction
```

When a task defines `prompt={"query": "..."}` without a `"document"` key (the convention in BRIGHT / BRIGHT(v1.1) and several other retrieval tasks), `get_instruction(task, document)` returns `""`. The gate then makes `get_task_instruction` return `""` directly — the model's `instruction_template` is **never invoked for documents**.

For models whose `instruction_template` is a callable that deliberately emits a non-empty prefix on empty input — ReasonIR (`<|embed|>\n`), GritLM (`<|embed|>\n`), Sarashina v2 (`text: `), BMRetriever, Octen — documents are encoded **without** the prefix the model was trained with. Queries are unaffected (their instruction body is non-empty for these tasks).

### Reproduction (ReasonIR-8B × `BrightProBiologyRetrieval`, full 60K corpus)

| pipeline | nDCG@10 |
|---|---:|
| BRIGHT-Pro paper saved | 0.33322 |
| BRIGHT-Pro `AutoModel + custom .encode` (in our env) | 0.33755 |
| MTEB pipeline (current upstream main) | 0.35323 |
| MTEB pipeline (with proposed fix) | 0.34269 |

Encoding the same 10,000 docs with `prompt=""` vs `prompt="<|embed|>\n"` gives **per-doc cosine median 0.94, min 0.25** — the docs genuinely diverge.

Full diagnostic scripts and JSON results: https://github.com/yilunzhao/bright-pro-mteb-validation (see `scripts/verify_reasonir_*.py`, `results/diagnostics/`).

### Affected models (callable template returns non-empty for empty/document)

- `ReasonIR/ReasonIR-8B` — `mteb/models/model_implementations/reasonir_model.py:14-22`
- `GritLM/GritLM-7B`, `GritLM/GritLM-8x7B` — `gritlm_models.py:21-26`
- `Sarashina v2` — `sarashina_embedding_models.py:97-109`
- `BMRetriever` × 5 ModelMeta — `bmretriever_models.py:17-24`
- `Octen` × 5 + `Octen-INT8` — `octen_models.py:6-35`

Unaffected: callable templates that already gate the empty case themselves (Qwen3-Embedding, gte-Qwen2, NV-Embed, etc.) and `str` templates of the form `f"Instruct: {instruction}\nQuery: "` which would produce malformed prefixes if formatted with an empty instruction.

### Likely also explains

- #3268 (CLOSED) "Results on `BRIGHT` not matching" — same symptom, never traced to this gate. The fix shipped there only split tasks per subset.
- #3506 (MERGED) "fix ReasonIR instruction" — modified the template body but left the gate, so the empty-instruction branch was still unreachable.

### Fix

PR coming next — invoke `format_instruction` for callable templates regardless of `len(instruction)`. `str` templates remain gated. Includes regression tests covering callable / str / None template cases.

yilunzhao:
Closes #4683.

## Summary

`AbsEncoder.get_task_instruction` previously skipped `instruction_template` whenever the resolved instruction was empty (the common case for documents when a task defines `prompt={"query": "..."}` only). Callable templates that deliberately emit a non-empty document prefix — ReasonIR / GritLM (`<|embed|>\n`), Sarashina v2 (`text: `), BMRetriever, Octen — never got the chance to emit that prefix, so docs were encoded with a different protocol than the model was trained with.

## Patch

Invoke `format_instruction` whenever `instruction_template` is a **callable**, regardless of `len(instruction)`. `str` templates of the form `f"Instruct: {instruction}\nQuery: "` would produce malformed prefixes like `"Instruct: \nQuery: "` on empty input, so they remain gated.

## Empirical impact

ReasonIR-8B × `BrightProBiologyRetrieval` (full 60K corpus):

| pipeline | nDCG@10 |
|---|---:|
| BRIGHT-Pro paper saved | 0.33322 |
| MTEB pipeline (before this PR) | 0.35323 |
| **MTEB pipeline (with this PR)** | **0.34269** |

Closes ~55% of the gap to the ReasonIR paper number. The residual ~0.009 is implementation-level noise (sentence-transformers' tokenizer/padding vs HF's custom `.encode`, bf16 numerical precision).

Per-doc cosine on a 10K subset: 9985 / 10000 docs had cosine < 0.99 between `prompt=""` (current upstream) and `prompt="<|embed|>\n"` (this PR) — the embeddings genuinely diverged.

Full diagnostic at https://github.com/yilunzhao/bright-pro-mteb-validation#root-cause-analysis-of-the-reasonir-8b-gap.

## Test plan

- [x] `tests/test_integrations/test_get_task_instruction.py` — 3 cases: callable template invoked on empty, `str` template gated, no template returns raw instruction. All pass locally.
- [x] `ruff check` + `ruff format --check` clean on changed files.
- [x] End-to-end re-evaluation of ReasonIR-8B on `BrightProBiologyRetrieval` with the patch applied reproduces 0.343.


mteb/models/abs_encoder.py
```diff
@@ -266,8 +266,23 @@ def get_task_instruction(
...
+            # empty-instruction gate for those.
+            #
+            # `Callable` templates are expected to handle the empty case
+            # themselves — and several models (ReasonIR-8B, GritLM-7B,
+            # Octen, Sarashina v2, BMRetriever) rely on this to emit a
+            # non-empty document prefix (e.g. ReasonIR/GritLM's
+            # `"<|embed|>\n"`). Always invoke callables so that prefix is
+            # not silently dropped, which previously caused those models'
+            # documents to be encoded without the prefix they were
+            # trained with.
```
Samoed:
```suggestion
            # Always invoke callables
```

Samoed (commented):
Great catch!

mteb/models/abs_encoder.py
KennethEnevoldsen: I would add a link to the issue for the expanded explanation, but agree that we can shorten it

tests/test_models/test_get_task_instruction.py
```diff
@@ -0,0 +1,59 @@
...
+            PromptType.query,
+            f"Instruct: {QUERY_INSTR}\nQuery: ",
+        ),
+        # No template: returns raw instruction (empty for document side)
+        (None, PromptType.document, ""),
+        # No template: returns raw instruction (query side)
+        (None, PromptType.query, QUERY_INSTR),
+    ],
+)
+def test_get_task_instruction(template, prompt_type, expected):
```
KennethEnevoldsen, line 55:
```suggestion
def test_get_task_instruction(template, prompt_type, expected):
   """regression test for issue #4683"""
```

tests/test_models/test_get_task_instruction.py
KennethEnevoldsen, line 1: would move into test/test_models

KennethEnevoldsen (approved):
Ahh amazing catch! Only a few minor comments

Samoed (approved):

PR #4692, opened 2026-05-19
fix revision for code fuse

Samoed:
I tried to run codefuse models on some tasks, but got an error `num_hidden_layers must be equeal to the number of layers`

CC @Geralt-Targaryen


Geralt-Targaryen:
It's an error in the config file that's fine with transformers v4 but problematic for transformers v5. It has been fixed in https://huggingface.co/codefuse-ai/F2LLM-v2-80M/discussions/1

PR #4696, opened 2026-05-19
fix: probe trailing frames in resample_video

AdnanElAssadi56:
Replace the unconditional -1 workaround with a backwards probe so healthy videos keep all frames and clips with multiple damaged trailing frames (e.g. HMDB51 AVIs) still decode.

### DIFF
--- a/mteb/models/modality_collators.py
+++ b/mteb/models/modality_collators.py
@@ -224,9 +224,15 @@ class FramesCollator:
         # Retry on the actual call: decrement source count when trailing
         # frames fail to decode.
         n = video.metadata.num_frames
+        original_n = n
         while n > 0:
             try:
                 frames: torch.Tensor = video.get_frames_at(_indices(n)).data
+                if n < original_n:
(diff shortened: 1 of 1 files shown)


mteb/models/modality_collators.py
```diff
@@ -205,37 +205,34 @@ def resample_video(
...
+        # frames fail to decode.
+        n = video.metadata.num_frames
+        while n > 0:
+            try:
+                frames: torch.Tensor = video.get_frames_at(_indices(n)).data
+                return frames
+            except RuntimeError as e:
+                if "no more frames" not in str(e):
+                    raise
+                n -= 1
```
Samoed, line 240: Should we add logging message in that case?

AdnanElAssadi: you meant before raise? or when dropping frames?

Samoed: I think this is good

Samoed (approved):

PR #4703, opened 2026-05-21
Tasks passing min frame and duration threshold for frame_scaling

deep9539:
Out of 27 selected tasks based on correlation analysis done in PR https://github.com/embeddings-benchmark/mteb/pull/4697, I've applied the following filter:
* average_duration > 15s
* average_fps > 300 (No effect, since they are already captured by average_duration > 15s)

Why?
And video tasks with average video below 15 seconds, should be captured by num_frames=16 and above, since each second will contribute a frame. 
The only value we get by adding tasks below 15 seconds is for num_frames 1 and 8, but the task passing the above threshold are also sufficient so prove the value num_frames 1/8 provides against 32/64.

Concerns?
Out of 27 tasks, only 9 tasks satisfy this criteria which I believe is okay, but please let me know if you think otherwise.
4 retrieval
2 Video(Audio)CentricQA
1 Classification
1 Clustering
1 PairClassification
0 ZeroShotClassification


Samoed:
Will merge since we finished paper for now

PR #4706, opened 2026-05-21
feat: Add Hybrid Models in MTEB

### LINKED ISSUE #4463: Figure out how best to add hybrid models

It would be great to be able to compare fusion models on MTEB (e.g. hybrid-dbsf as suggested [here](https://www.linkedin.com/feed/update/urn:li:activity:7451970469886459904?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A7451970469886459904%2C7452069079911153664%29&dashCommentUrn=urn%3Ali%3Afsd_comment%3A%287452069079911153664%2Curn%3Ali%3Aactivity%3A7451970469886459904%29) with refs [here](https://github.com/atlantis-nova/hybrid-dbsf)).

I would probably suggest selecting a reference embedding model and then use that across.

Implementation wise it should be possible to implement it as a wrapper:

```py
m1 = mteb.get_model("mteb/baseline-bm25s")
m2 = mteb.get_model("intfloat/multilingual-e5-small")
hybrid_model = DBSFHybridSearch(model1=m1, model2=m2)
```

Note that the models just need to implement the [SearchProtocol](https://embeddings-benchmark.github.io/mteb/api/model/?h=SearchProtocol#mteb.models.SearchProtocol)

ayush1298:
closes #4463

This PR adds hybrid search fusion wrappers in MTEB, which will allow users to compare hybrid models. 

Here, a new wrapper `HybridSearch` is added that accepts 2 or more models (which can be standard Encoder, CrossEncoder, or already wrapped SearchProtocol objects), index the corpus with all underlying models, and perform score/rank fusion during search.
We will support three primary hybrid search methods, each of which has a dedicated fusion function. (the original issue was related to adding DBSF only):

1. DBSF (Distribution-Based Score Fusion): Normalizes retriever scores query-by-query using the mean ($\mu$) and standard deviation ($\sigma$) within a $3\sigma$ bound, then combines them. Implemented through the `DBSFHybridSearch` class. 
2. RRF (Reciprocal Rank Fusion): Combines the reciprocal rank positions of the retrieved documents using a constant $k$ (typically 60). Implemented through the `RRFHybridSearch` class. 
3. Relative Score Fusion (RSF) / MinMax Fusion: Normalizes the scores to a $[0, 1]$ range query-by-query using standard MinMax scaling. Implemented through the `RelativeScoreFusionHybridSearch` class. 

Also, I have added a test under `tests/test_models/test_hybrid_search.py`.


mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,247 @@
...
+        model1: MTEBModels,
+        model2: MTEBModels,
+        *models: MTEBModels,
+        weights: list[float] | None = None,
+        sub_model_top_k: int | None = None,
+        rrf_k: int = 60,
+    ) -> None:
+        super().__init__(
+            model1, model2, *models, weights=weights, sub_model_top_k=sub_model_top_k
+        )
```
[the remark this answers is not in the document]
ayush1298: fix

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,247 @@
...
+
+        class_name = self.__class__.__name__
+        combined_name = f"hybrid-{class_name.lower()}/{'-'.join(names)}"
+
+        return ModelMeta.create_empty(
+            overwrites={
+                "name": combined_name,
+                "model_type": ["dense"],
+            }
+        )
```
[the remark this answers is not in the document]
ayush1298: fix

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,247 @@
...
+        for model in self.wrapped_models:
+            meta = getattr(model, "mteb_model_meta", None)
+            if meta and meta.name:
+                names.append(meta.name.split("/")[-1])
+            else:
+                names.append("unknown")
+
+        class_name = self.__class__.__name__
+        combined_name = f"hybrid-{class_name.lower()}/{'-'.join(names)}"
+
```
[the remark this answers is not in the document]
ayush1298: fix

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,247 @@
...
+        self.wrapped_models: list[SearchProtocol] = []
+        for model in self.models:
+            if isinstance(model, EncoderProtocol) and not isinstance(
+                model, SearchProtocol
+            ):
+                self.wrapped_models.append(SearchEncoderWrapper(model))
+            elif isinstance(model, CrossEncoderProtocol):
+                self.wrapped_models.append(SearchCrossEncoderWrapper(model))
+            elif isinstance(model, SearchProtocol):
+                self.wrapped_models.append(model)
```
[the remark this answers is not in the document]
ayush1298: I have implemented these one.

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,342 @@
...
+class BaseHybridSearch(SearchProtocol):
+    """Base class for hybrid search wrappers implementing the SearchProtocol."""
+
+    fusion_name: str | None = None
+
+    def __init__(
+        self,
+        model1: MTEBModels,
+        model2: MTEBModels,
+        *models: MTEBModels,
```
Samoed: Can we simplify to?
```suggestion
        models: Sequence[MTEBModels],
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,342 @@
...
+    from mteb.types import (
+        CorpusDatasetType,
+        EncodeKwargs,
+        QueryDatasetType,
+        RetrievalOutputType,
+        TopRankedDocumentsType,
+    )
+
+
+class BaseHybridSearch(SearchProtocol):
```
Samoed: You shouldn't inherit from protocols
```suggestion
class BaseHybridSearch:
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,342 @@
...
+        encode_kwargs: EncodeKwargs,
+        top_ranked: TopRankedDocumentsType | None = None,
+        num_proc: int | None = None,
+    ) -> RetrievalOutputType:
+        """Search the queries using all sub-models and fuse the results."""
+        if self.sub_model_top_k is None:
+            sub_top_k = max(top_k, 100)
+        else:
+            if self.sub_model_top_k <= 0:
+                raise ValueError("sub_model_top_k must be greater than 0")
```
Samoed: I think this should be in init

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,342 @@
...
+                retriever_results.append(res)
+
+            generated_top_ranked: dict[str, list[str]] = {}
+            for row in queries:
+                qid = row["id"]
+                candidates = set()
+                for res in retriever_results:
+                    if qid in res:
+                        candidates.update(res[qid].keys())
+                generated_top_ranked[qid] = list(candidates)
```
Samoed: List of candidates will be huge and you should trim it

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,342 @@
...
+            all_results = []
+            ret_idx = 0
+            ce_idx = 0
+            for is_ce in model_is_cross_encoder:
+                if is_ce:
+                    all_results.append(cross_encoder_results[ce_idx])
+                    ce_idx += 1
+                else:
+                    all_results.append(retriever_results[ret_idx])
+                    ret_idx += 1
```
Samoed: I'm unsure what's happening here

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,342 @@
...
+                    fusion_name = fusion_name[: -len(suffix)]
+            fusion_name = fusion_name.lower()
+
+        combined_name = f"hybrid-{fusion_name}/{'-'.join(names)}"
+        final_model_types = sorted(model_types) if model_types else ["dense"]
+
+        return ModelMeta.create_empty(
+            overwrites={
+                "name": combined_name,
+                "model_type": final_model_types,
```
Samoed:
```suggestion
                "model_type": "hybrid",
```

mteb/models/hybrid_wrappers.py
Samoed: I think this should be 
```suggestion
class BaseHybridSearch(ABC):
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,342 @@
...
+            fused_query_scores = self.fuse(query_scores_list)
+
+            sorted_fused = sorted(
+                fused_query_scores.items(), key=lambda x: x[1], reverse=True
+            )
+            fused_results[qid] = dict(sorted_fused[:top_k])
+
+        return fused_results
+
+    def fuse(self, query_scores_list: list[dict[str, float]]) -> dict[str, float]:
```
Samoed:
```suggestion
    @abstractmethod
    def fuse(self, query_scores_list: list[dict[str, float]]) -> dict[str, float]:
```

mteb/models/hybrid_wrappers.py
Samoed: Can you separately create an interface for this?

mteb/models/hybrid_wrappers.py
ayush1298: We are just realigning results into a single list. I have refactored it in a better way.

mteb/models/hybrid_wrappers.py
ayush1298: yes, done

mteb/models/hybrid_wrappers.py
ayush1298: but this is not a valid model_type. Currently it is just creating a list of `model_type` of all models

mteb/models/hybrid_wrappers.py
ayush1298: I have added like `HybridSearchProtocol` under `model_protocols.py`.

ayush1298:
@KennethEnevoldsen Could you review it?

KennethEnevoldsen (commented):
I would love to test this with a few models. This could very well be an example that goes in the documentation as well which we should also add. I think it can be under examples.

mteb/models/hybrid_wrappers.py
KennethEnevoldsen: We should just add hybrid

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,355 @@
...
+        TopRankedDocumentsType,
+    )
+
+
+class BaseHybridSearch(ABC):
+    """Base class for hybrid search wrappers implementing the SearchProtocol."""
+
+    fusion_name: str | None = None
+
+    def __init__(
```
KennethEnevoldsen, line 39: Unsure if we would rather just implement `HybridSearch` and then have a `fusion_strategy` instead.

ayush1298: maybe we could, I guess this is just a more structured way

KennethEnevoldsen: More structured how so?

From a user perspective, I get a new class for essentially doing the same thing.

ayush1298: okay, I though again and agreed with you. Having a single `HybridSearch` class will be better. Will refactor it and also add example. For adding model, we could take any 2 models, right?

KennethEnevoldsen: I would use bm25 as one of them and a dense encoder as the other

ayush1298:
I have updated the implementation with a single class and fusion_strategy as a parameter, so no need for multiple classes. (updated PR description as well). And also added some models under `hybrid_models.py` and in the docs. Also, I have tested them and it's working

docs/get_started/usage/defining_the_model.md
```diff
@@ -66,6 +66,43 @@ on `MTEB`:
 !!! note
     We do recommend using `mteb.get_model` which will by default load the model using the implementation in `mteb` if there is one, otherwise it will use `SentenceTransformers` or `CrossEncoder` from [sentence transformers](https://www.sbert.net/) if appropriate. The `mteb` implementations typically differ due to models requiring specific prompts or similar hyperparameters, and not specifying these may reduce performance (e.g. the [multilingual e5 models](https://huggingface.co/collections/intfloat/multilingual-e5-text-embeddings-67b2b8bb9bff40dec9fb3534) require specific prompts).
 
+## Using a Hybrid Model
+
+MTEB provides a unified [`mteb.HybridSearch`](../../api/model.md#mteb.HybridSearch) wrapper that allows you to combine multiple retrievers and cross-encoders using different fusion strategies (e.g. Reciprocal Rank Fusion, Distribution-Based Score Fusion, or custom fusion functions).
```
Samoed:
```suggestion
MTEB provides a unified [`mteb.HybridSearch`][mteb.HybridSearch] wrapper that allows you to combine multiple retrievers and cross-encoders using different fusion strategies (e.g. Reciprocal Rank Fusion, Distribution-Based Score Fusion, or custom fusion functions).
```

docs/get_started/usage/defining_the_model.md
```diff
@@ -66,6 +66,43 @@ on `MTEB`:
 !!! note
     We do recommend using `mteb.get_model` which will by default load the model using the implementation in `mteb` if there is one, otherwise it will use `SentenceTransformers` or `CrossEncoder` from [sentence transformers](https://www.sbert.net/) if appropriate. The `mteb` implementations typically differ due to models requiring specific prompts or similar hyperparameters, and not specifying these may reduce performance (e.g. the [multilingual e5 models](https://huggingface.co/collections/intfloat/multilingual-e5-text-embeddings-67b2b8bb9bff40dec9fb3534) require specific prompts).
 
+## Using a Hybrid Model
```
Samoed: I think this would be better in advanced usage in separate file

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,220 @@
...
+    retriever1 = MockRetriever()
+    retriever2 = MockRetriever()
+    cross_encoder = MockSearchCrossEncoderWrapper()
+
+    hybrid = HybridSearch(
+        [retriever1, retriever2, cross_encoder],
+        sub_model_top_k=3,
+        fusion_strategy="dbsf",
+    )
+    from tests.mock_tasks import MockRetrievalTask
```
Samoed: Move to top level

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,220 @@
...
+        def __init__(self):
+            pass
+
+        def search(self, queries, top_k, top_ranked, **kwargs):  # noqa: PLR6301
+            assert top_ranked is not None
+            assert len(top_ranked["q1"]) == top_k
+            return {"q1": dict.fromkeys(top_ranked["q1"], 1.0)}
+
+    retriever1 = MockRetriever()
+    retriever2 = MockRetriever()
```
Samoed: You can use `random-baseline` with different dims

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,220 @@
...
+    cross_encoder = MockSearchCrossEncoderWrapper()
```
Samoed: You can use mock cross-encoder

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,220 @@
...
+    res = hybrid.search(
+        queries=[{"id": "q1", "text": "query"}],
+        task_metadata=task.metadata,
+        hf_split="test",
+        hf_subset="default",
+        top_k=2,
+        encode_kwargs={},
+        top_ranked=None,
+    )
+    assert res is not None
```
Samoed: Better to set explicit value

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,220 @@
...
+    for strategy in [
+        "dbsf",
+        "rrf",
+        "relative-score-fusion",
+    ]:
+        hybrid = HybridSearch([m1, m2], fusion_strategy=strategy)
+        task = MockRetrievalTask()
+        results = mteb.evaluate(hybrid, task, cache=None)
+        assert len(results) > 0
+        assert results[0].scores is not None
```
Samoed: Better to set explicit value

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,220 @@
...
+from mteb import HybridSearch
+from tests.mock_tasks import MockRetrievalTask
+
+
+def test_hybrid_search_init_and_meta():
+    """Test the initialization, weight validation, and metadata generation of hybrid search wrapper."""
+    m1 = mteb.get_model("mteb/baseline-random-encoder", embed_dim=32)
+    m2 = mteb.get_model("mteb/baseline-random-encoder", embed_dim=10)
+
+    hybrid = HybridSearch([m1, m2], fusion_strategy="dbsf")
```
Samoed, line 17: Can you add check if HybridSearch is HybridProtocol

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+        self,
+        models: Sequence[MTEBModels],
+        weights: list[float] | None = None,
+        sub_model_top_k: int | None = None,
+        fusion_strategy: Literal["rrf", "dbsf", "relative-score-fusion"]
+        | Callable[[list[dict[str, float]], list[float]], dict[str, float]] = "rrf",
+        rrf_k: int = 60,
+    ) -> None:
+        self.models = list(models)
```
Samoed: Seems you don't need to save them
```suggestion
       models = list(models)
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+                normalized = (raw_scores - lower_limit) / denom
+                normalized = np.clip(normalized, 0.0, 1.0)
+
+            for doc_id, norm_score in zip(doc_ids, normalized):
+                fused[doc_id] = fused.get(doc_id, 0.0) + weight * float(norm_score)
+
+        return fused
+
+    def _fuse_rrf(
+        self, query_scores_list: list[dict[str, float]], weights: list[float]
```
Samoed:
```suggestion
        self, query_scores_list: RetrievalOutputType, weights: Sequence[float]
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+
+        return ModelMeta.create_empty(
+            overwrites={
+                "name": combined_name,
+                "model_type": ["hybrid"],
+            }
+        )
+
+    @mteb_model_meta.setter
+    def mteb_model_meta(self, value: ModelMeta) -> None:
```
Samoed: Why do we need setter? I don't think that we need it

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+            for doc_id, norm_score in zip(doc_ids, normalized):
+                fused[doc_id] = fused.get(doc_id, 0.0) + weight * float(norm_score)
+
+        return fused
+
+    @property
+    def mteb_model_meta(self) -> ModelMeta:
+        """Generate combined ModelMeta for the hybrid model."""
+        if hasattr(self, "_mteb_model_meta"):
+            return self._mteb_model_meta
```
Samoed: Why do we need this?

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+
+        names = []
+        for model in self.wrapped_models:
+            meta = getattr(model, "mteb_model_meta", None)
```
Samoed: Model would always have `mteb_model_meta`

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+            if meta and meta.name:
+                names.append(meta.name.split("/")[-1])
```
Samoed:
```suggestion
                names.append(meta.name.rsplit("/", 1)[-1])
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+                denom = upper_limit - lower_limit
+                normalized = (raw_scores - lower_limit) / denom
+                normalized = np.clip(normalized, 0.0, 1.0)
+
+            for doc_id, norm_score in zip(doc_ids, normalized):
+                fused[doc_id] = fused.get(doc_id, 0.0) + weight * float(norm_score)
+
+        return fused
+
+    def _fuse_rrf(
```
Samoed: I think would be better to move out from `fused` functions from class

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+        cross_encoder_models = []
+        retriever_models = []
+        model_is_cross_encoder = []
+        for model in self.wrapped_models:
+            if isinstance(model, SearchCrossEncoderWrapper):
+                cross_encoder_models.append(model)
+                model_is_cross_encoder.append(True)
+            else:
+                retriever_models.append(model)
+                model_is_cross_encoder.append(False)
```
Samoed: I think better to move to `init`

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,347 @@
...
+        if cross_encoder_models and effective_top_ranked is None:
+            if not retriever_models:
+                raise ValueError(
+                    "CrossEncoder sub-models require top_ranked documents for reranking, "
+                    "or at least one retriever sub-model in the hybrid wrapper to generate candidates."
+                )
+
+            retriever_results = []
+            for model in retriever_models:
+                res = model.search(
```
Samoed: Can we add logging here? Maybe with progress bar

docs/get_started/usage/defining_the_model.md
ayush1298: okay, will move

tests/test_models/test_hybrid_search.py
ayush1298: Is it same as `mteb/baseline-random-cross-encoder`?

Samoed: Yes, I meant this

mteb/models/hybrid_wrappers.py
ayush1298: removed

mteb/models/hybrid_wrappers.py
ayush1298: removed

mteb/models/hybrid_wrappers.py
ayush1298: Why we need these, any particular reason? Because before using we have to convert it again like: 
`scores_list: list[dict[str, float]] = query_scores_list`

mteb/models/hybrid_wrappers.py
ayush1298: done

mteb/models/hybrid_wrappers.py
Samoed: This is the same type, but has a naming. This is the same type that `SearchProtocol` output https://github.com/embeddings-benchmark/mteb/blob/e9a10bd087ffa788197323af27c5ee6ce111a294/mteb/models/models_protocols.py#L49-L63

docs/get_started/advanced_usage/hybrid_model.md
Samoed, line 1: You need update `mkdocs.yaml` to include this page

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+        corpus: CorpusDatasetType,
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        encode_kwargs: EncodeKwargs,
+        num_proc: int | None = None,
+    ) -> None:
+        """Index the corpus dataset using all sub-models."""
+        for model in self.wrapped_models:
```
Samoed: Can you add progress bar and logging here too?

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+            retriever_results = []
+            for model in tqdm(retriever_models, desc="Retriever sub-models"):
+                res = model.search(
+                    queries=queries,
+                    top_k=sub_top_k,
+                    task_metadata=task_metadata,
+                    hf_split=hf_split,
+                    hf_subset=hf_subset,
+                    encode_kwargs=encode_kwargs,
+                    top_ranked=None,
```
Samoed: I'm not sure if we want to ignore `top_ranked` here. This would produce incorrect results for reranking tasks

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+    CrossEncoderProtocol,
+    EncoderProtocol,
+    SearchProtocol,
+)
+from mteb.models.search_wrappers import (
+    SearchCrossEncoderWrapper,
+    SearchEncoderWrapper,
+)
+
+logger = logging.getLogger(__name__)
```
Samoed: Can you move after `TYPE_CHECKING`? Otherwise, linters won't sort type checking block

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Literal
+
+import numpy as np
+from tqdm import tqdm
```
Samoed:
```suggestion
from tqdm.auto import tqdm
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+    """Hybrid search wrapper that combines multiple models using a specified fusion strategy."""
+
+    def __init__(
+        self,
+        models: Sequence[MTEBModels],
+        weights: list[float] | None = None,
+        sub_model_top_k: int | None = None,
+        fusion_strategy: Literal["rrf", "dbsf", "relative-score-fusion"]
+        | Callable[[RetrievalOutputType, Sequence[float]], dict[str, float]] = "rrf",
+        rrf_k: int = 60,
```
Samoed, line 48: Can you add docs to params?

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+    ) -> None:
+        models = list(models)
+        if len(models) < 2:
+            raise ValueError("At least two models must be provided for hybrid search.")
+        self.models = models
```
Samoed: Not used
```suggestion
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+    def __init__(
+        self,
+        models: Sequence[MTEBModels],
+        weights: list[float] | None = None,
+        sub_model_top_k: int | None = None,
+        fusion_strategy: Literal["rrf", "dbsf", "relative-score-fusion"]
+        | Callable[[RetrievalOutputType, Sequence[float]], dict[str, float]] = "rrf",
+        rrf_k: int = 60,
+    ) -> None:
+        models = list(models)
```
Samoed: I don't think you need to convert this to list
```suggestion
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+                self.retriever_models.append(model)
+                self.model_is_cross_encoder.append(False)
+
+        names = []
+        for model in self.wrapped_models:
+            meta = model.mteb_model_meta
+            if meta and meta.name:
+                names.append(meta.name.rsplit("/", 1)[-1])
+            else:
+                names.append("unknown")
```
Samoed: Can you do this in loop with conversion to protocol? https://github.com/ayush1298/mteb/blob/9d02d6e2a0b547d0b26ef9a38a5036fb485e0d04/mteb/models/hybrid_wrappers.py#L54-L66

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+            elif fusion_strategy in {"relative-score-fusion", "relative_score_fusion"}:
+                self.fusion_name = "relative-score-fusion"
+                self._fuse_fn = fuse_relative_score_fusion
+            else:
+                raise ValueError(f"Unknown fusion strategy: {fusion_strategy}")
+        elif callable(fusion_strategy):
+            self.fusion_name = getattr(fusion_strategy, "__name__", "custom")
+            self._fuse_fn = fusion_strategy
+        else:
+            raise TypeError("fusion_strategy must be a string or a callable")
```
Samoed:
```suggestion
            raise TypeError("fusion_strategy must be a "rrf", "dbsf", "relative-score-fusion" or a callable")
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+
+        if weights is None:
+            self.weights = [1.0 / len(models)] * len(models)
+        else:
+            if len(weights) != len(models):
+                raise ValueError("Length of weights must match the number of models.")
+            self.weights = list(weights)
+
+        if sub_model_top_k is not None:
+            if sub_model_top_k <= 0:
```
Samoed:
```suggestion
        if sub_model_top_k is not None and sub_model_top_k <= 0:
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,357 @@
...
+                    if qid in res:
+                        scores = res[qid]
+                        sorted_docs = sorted(
+                            scores.keys(), key=lambda d: scores[d], reverse=True
+                        )
+                        for rank_idx, doc_id in enumerate(sorted_docs):
+                            rank = rank_idx + 1
+                            candidate_rrf[doc_id] = candidate_rrf.get(
+                                doc_id, 0.0
+                            ) + 1.0 / (60 + rank)
```
Samoed: What 60 means here?

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+        hybrid_only_ce.search(
+            queries=[{"id": "q1", "text": "query"}],
+            task_metadata=task.metadata,
+            hf_split="test",
+            hf_subset="default",
+            top_k=5,
+            encode_kwargs={},
+            top_ranked=None,
+        )
```
Samoed: Can we split this to different test?

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,226 @@
...
+
+def test_registered_hybrid_model_retrieval():
+    """Verify that the registered hybrid-bm25s-e5-small model can be loaded and its metadata is correct."""
+    meta = mteb.get_model_meta("mteb/hybrid-bm25s-e5-small")
+    assert meta is not None
+    assert meta.name == "mteb/hybrid-bm25s-e5-small"
+    assert meta.model_type == ["hybrid"]
+
+    with patch("mteb.models.model_meta.ModelMeta._check_requirements"):
+        model = mteb.get_model("mteb/hybrid-bm25s-e5-small")
```
Samoed: I think you can change to `mteb/hybrid-baseline_encoder-e5-small` to remove patching here

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,73 @@
...
+    license=None,
+    max_tokens=None,
+    reference=None,
+    similarity_fn_name=None,
+    framework=[],
+    use_instructions=False,
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
+    citation="",
```
Samoed:
```suggestion
    citation=None,
```

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,73 @@
...
+hybrid_bm25s_e5 = ModelMeta(
+    loader=hybrid_bm25s_e5_loader,
+    name="mteb/hybrid-bm25s-e5-small",
+    model_type=["hybrid"],
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="0_1_0",
+    release_date="2026-05-27",
+    n_parameters=0,
+    n_embedding_parameters=0,
```
Samoed: I think better to make this `None`, because `e5-small` have parameters and this would be incorrect

Samoed:
Would be good to add a section to `what's new`

mteb/models/hybrid_wrappers.py
ayush1298: didn't get it, what you mean here

mteb/models/hybrid_wrappers.py
ayush1298: cant iterate on `Sequence,` so need to convert to `list`

mteb/models/hybrid_wrappers.py
Samoed: Docstring to HybridModel `__init__`

mteb/models/hybrid_wrappers.py
ayush1298: constant used in rrf

mteb/models/hybrid_wrappers.py
Samoed: You can iterate over Sequence. All `Sequence` have `__iter__` 
```python
from collections.abc import Sequence
dir(Sequence)
[
...
'__iter__',
...
]
```

mteb/models/hybrid_wrappers.py
Samoed: Is it supposed to be `rrf_k`?

mteb/models/hybrid_wrappers.py
ayush1298: sorry, I confused with something else. Will remove it

mteb/models/hybrid_wrappers.py
ayush1298: I don't think so. If you see the entire loop, so here we are having cross-encoder models but top_ranked is None, so here in these 1st loop, we are using retriever models to generate these top_ranked and then loop on cross-encoder models with these generated top_ranked.

tests/test_models/test_hybrid_search.py
ayush1298: done

mteb/models/hybrid_wrappers.py
ayush1298: yes, I have changed it now to `self.rrf_k`

ayush1298:
> Would be good to add a section to `what's new`

Updated docs, and resolved other comments as well

mteb/models/hybrid_wrappers.py
Samoed: But if a task explicitly adds `top_ranked` docs to filter out some docs for queries, this method would add the filtered out docs back.

ayush1298: if task added top_ranked, then it will just go to else block, and do search with it as: 
`effective_top_ranked = top_ranked`

Samoed:
I was wrong a bit about typing, didn't notice that you processed output results

ayush1298:
> I was wrong a bit about typing, didn't notice that you processed output results

Didnt get it, why we have to change from `RetrievalOutputType` to `Sequence[dict[str, float]]`?

Samoed:
I didn't notice that you process results in
```python
            query_scores_list = []
            for res in all_results:
                query_scores_list.append(res.get(qid, {}))

            fused_query_scores = self.fuse(query_scores_list)
```

ayush1298:
@KennethEnevoldsen Can you review it once again? I have add some models as examples and updated docs as well

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,369 @@
...
+        """
+        if len(models) < 2:
+            raise ValueError("At least two models must be provided for hybrid search.")
+
+        if weights is None:
+            self.weights = [1.0 / len(models)] * len(models)
+        else:
+            if len(weights) != len(models):
+                raise ValueError("Length of weights must match the number of models.")
+            self.weights = list(weights)
```
Samoed:
```suggestion
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,369 @@
...
+logger = logging.getLogger(__name__)
+
+
+class HybridSearch:
+    """Hybrid search wrapper that combines multiple models using a specified fusion strategy."""
+
+    def __init__(
+        self,
+        models: Sequence[MTEBModels],
+        weights: list[float] | None = None,
```
Samoed:
```suggestion
        weights: Sequence[float] | None = None,
```

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,369 @@
...
+        sub_model_top_k: int | None = None,
+        fusion_strategy: Literal["rrf", "dbsf", "relative-score-fusion"]
+        | Callable[
+            [Sequence[dict[str, float]], Sequence[float]], dict[str, float]
```
Samoed:
```suggestion
            [Sequence[Mapping[str, float]], Sequence[float]], Mapping[str, float]
```

Samoed (approved):
Looks good!

docs/get_started/advanced_usage/hybrid_model.md
```diff
@@ -0,0 +1,36 @@
...
+from mteb import HybridSearch
+
+# Load individual sub-models (lexical/sparse and dense)
+bm25 = mteb.get_model("mteb/baseline-bm25s")
+dense = mteb.get_model("intfloat/multilingual-e5-small")
+
+# Create the hybrid model combining both sub-models
+hybrid_model = HybridSearch(
+    models=[bm25, dense],
+    fusion_strategy="rrf",  # Options: "rrf", "dbsf", "relative-score-fusion", or a custom Callable
```
KennethEnevoldsen:
```suggestion
    fusion_strategy="rrf",
```

KennethEnevoldsen: will change in the future and this will become outdated

docs/whats_new.md
```diff
@@ -7,6 +7,32 @@ icon: lucide/newspaper
...
+from mteb import HybridSearch
+
+# Load sub-models (e.g. BM25 and a dense retriever)
+bm25 = mteb.get_model("mteb/baseline-bm25s")
+dense = mteb.get_model("intfloat/multilingual-e5-small")
+
+# Wrap them using Reciprocal Rank Fusion (RRF)
+hybrid_model = HybridSearch(
+    models=[bm25, dense],
+    fusion_strategy="rrf",  # Options: "rrf", "dbsf", "relative-score-fusion", or a custom Callable
```
```suggestion
    fusion_strategy="rrf",
```

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,73 @@
...
+def hybrid_baseline_encoder_e5_loader(model_name: str, **kwargs) -> HybridSearch:
+    bm25 = mteb.get_model("mteb/baseline-random-encoder")
+    dense = mteb.get_model("intfloat/multilingual-e5-small")
+    return HybridSearch(
+        models=[bm25, dense],
+        fusion_strategy="rrf",
+    )
+
+
+hybrid_baseline_encoder_e5 = ModelMeta(
```
KennethEnevoldsen: Why do we want this model?

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,73 @@
...
+def hybrid_bm25s_e5_loader(model_name: str, **kwargs) -> HybridSearch:
+    bm25 = mteb.get_model("mteb/baseline-bm25s")
+    dense = mteb.get_model("intfloat/multilingual-e5-small")
+    return HybridSearch(
+        models=[bm25, dense],
+        fusion_strategy="rrf",
+    )
+
+
+hybrid_bm25s_e5 = ModelMeta(
```
KennethEnevoldsen: If this is a model for comparing hybrid models, should we include more than one fusion strategy?

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,73 @@
...
+    loader=hybrid_bm25s_e5_loader,
+    name="mteb/hybrid-bm25s-e5-small",
+    model_type=["hybrid"],
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="0.1.0",
+    release_date="2026-05-27",
+    n_parameters=0,
+    n_embedding_parameters=None,
+    memory_usage_mb=None,
```
KennethEnevoldsen: seems like the metadata is not correct

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+hybrid_bm25s_e5 = ModelMeta(
+    loader=hybrid_bm25s_e5_loader,
+    name="mteb/hybrid-bm25s-e5-small",
+    model_type=["hybrid"],
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="0.1.0",
+    release_date="2026-05-27",
```
KennethEnevoldsen: I would set the release date of e5

mteb/models/models_protocols.py
```diff
@@ -249,3 +251,12 @@ def mteb_model_meta(self) -> ModelMeta:
...
 """Type alias for all MTEB model types as many models implement multiple protocols and many tasks can be solved by multiple model types."""
+
+
+@runtime_checkable
+class HybridSearchProtocol(SearchProtocol, Protocol):
+    """Interface for hybrid search models that combine multiple sub-models."""
+
+    wrapped_models: Sequence[SearchProtocol]
+    weights: Sequence[float]
+    sub_model_top_k: int | None
```
KennethEnevoldsen: I don't think we need this

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,370 @@
...
+        model_is_cross_encoder = self.model_is_cross_encoder
+
+        effective_top_ranked = top_ranked
+
+        if cross_encoder_models and effective_top_ranked is None:
+            if not retriever_models:
+                raise ValueError(
+                    "CrossEncoder sub-models require top_ranked documents for reranking, "
+                    "or at least one retriever sub-model in the hybrid wrapper to generate candidates."
+                )
```
KennethEnevoldsen: shouldn't this error be handled downstream? The cross-encoder should raise

tests/test_models/test_hybrid_search.py
```diff
@@ -0,0 +1,227 @@
...
+    cross_encoder = mteb.get_model("mteb/baseline-random-cross-encoder")
+
+    hybrid = HybridSearch([retriever, cross_encoder], fusion_strategy="dbsf")
+    task = MockRetrievalTask()
+    results = mteb.evaluate(hybrid, task, cache=None)
+    assert len(results) == 1
+    assert results[0].get_score() == pytest.approx(0.81546)
+
+
+def test_hybrid_search_cross_encoder_requires_candidates():
```
KennethEnevoldsen: This don't think this should be a concern for the hybrid search - I would delete this test

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,370 @@
...
+                sorted_candidates = sorted(
+                    candidate_rrf.keys(), key=lambda d: candidate_rrf[d], reverse=True
+                )
+                generated_top_ranked[qid] = sorted_candidates[:sub_top_k]
+
+            effective_top_ranked = generated_top_ranked
+
+            logger.info("Running cross-encoder sub-models...")
+            cross_encoder_results = []
+            for model in tqdm(cross_encoder_models, desc="Cross-encoder sub-models"):
```
KennethEnevoldsen: I am not sure about the amount of custom code for cross-encoders. Any reason why we can't simply pass those arguments along to all models and let them use them if they use them otherwise ignore it.

docs/get_started/advanced_usage/hybrid_model.md
```diff
@@ -0,0 +1,36 @@
...
+### Registered Hybrid Models
+
+You can also register pre-configured hybrid models by adding a custom loader function and model metadata under `mteb/models/model_implementations/hybrid_models.py`. Once registered, the hybrid model can be loaded by name:
+
+```python
+import mteb
+
+# Load pre-registered hybrid model directly
+hybrid_model = mteb.get_model("mteb/hybrid-bm25s-e5-small")
+```
```
KennethEnevoldsen:
```suggestion
```

docs/get_started/advanced_usage/hybrid_model.md
```diff
@@ -0,0 +1,36 @@
...
+hybrid_model = HybridSearch(
+    models=[bm25, dense],
+    fusion_strategy="rrf",  # Options: "rrf", "dbsf", "relative-score-fusion", or a custom Callable
+    weights=[0.5, 0.5],     # Optional: weight assigned to each model
+)
+
+# Evaluate the hybrid model on your selected tasks
+tasks = mteb.get_tasks(tasks=["NFCorpus"])
+results = mteb.evaluate(hybrid_model, tasks=tasks)
+```
```
KennethEnevoldsen, line 25: I would love a table here comparing bm25, e5 small and e5 + bm25 using different fusion strategies. Maybe on 3 tasks.

mteb/models/hybrid_wrappers.py
ayush1298: > CrossEncoder sub-models are always wrapped with SearchCrossEncoderWrapper, but that wrapper requires top_ranked candidates and will raise at search-time when top_ranked is None (typical for retrieval tasks). Either disallow CrossEncoder models in this wrapper unless top_ranked is provided, or generate top_ranked candidates from the retriever sub-model(s) and pass them to CrossEncoder wrappers.

Above is a comment from Copilot, which seems correct, and because of which we need this custom code for cross-encoders:

mteb/models/model_implementations/hybrid_models.py
ayush1298: that was 1 of question, how exactly we should fill metadata of these hybrid models?

mteb/models/model_implementations/hybrid_models.py
ayush1298: you asked me to add some models to check implementations and test, so I have added it

mteb/models/model_implementations/hybrid_models.py
ayush1298: Didn't get it, are you saying more than 1 fusion strategy in a single model?

mteb/models/hybrid_wrappers.py
KennethEnevoldsen: Sure - but then just let `CrossEncoder` raise an error if it is not provided? I believe we can pass `top_ranked` to the other models as well.

mteb/models/model_implementations/hybrid_models.py
KennethEnevoldsen: I was suggesting whether we should register multiple hybrids - one for each search strategy

mteb/models/model_implementations/hybrid_models.py
KennethEnevoldsen: we should just consider it 1 one model.

I.e. sum parameters

mteb/models/model_implementations/hybrid_models.py
ayush1298: We can, these was just a to check 1 of them

mteb/models/hybrid_wrappers.py
ayush1298: Is there anything wrong in the current implementation where we are using retriever models to get these top_ranked, which is then used by cross-encoder models

ayush1298:
@KennethEnevoldsen could you check this one again? Mainly there are comments related to cross-encoder implementations. If, thats solved, I will share table comparing bm25, e5 small and e5 + bm25 using different fusion strategies for any 3 tasks, any suggestion on which tasks I should test?

mteb/models/hybrid_wrappers.py
KennethEnevoldsen: The current implementation splits the models in two piles and runs them seperately. I don't think this is needed and adds unec. complexity that we can avoid

KennethEnevoldsen:
checked

mteb/models/hybrid_wrappers.py
ayush1298: so, should we just avoid cross-encoders or what excatly?

KennethEnevoldsen: I just don't see a reason to treat them any differently from other models. If top-ranked is passed it should work with just passing it to both models (so no reason to split them up) and if it is not passed it will already raise an error (so again no reason to split them)

ayush1298: So, splitting is a case when the top-ranked is not passed, where we could make it work in this way.

ayush1298: @KennethEnevoldsen So, should I just removed the custom implementation and raise error if not pass? Or we want to make it work in these way by splitting?

KennethEnevoldsen: we can simply not split and just pass top-ranked always (regardless of whether it is none)

mteb/models/model_implementations/hybrid_models.py
ayush1298: @KennethEnevoldsen Should I do this?

mteb/models/hybrid_wrappers.py
ayush1298: removing these completely, as these will be handled by `SearchCrossEncoderWrapper` directly

mteb/models/hybrid_wrappers.py
ayush1298: I have refacored it

ayush1298:
@KennethEnevoldsen I have refactored and removed custom logic for cross-encoders. Could you check it once? It its correct, then I will run the models and add comparative results.
Also, 1 question from the above thread, do we have to add bm25 and e5 with other fusion strategies also under `hybrid_models.py?`

ayush1298:
I have registered `bm25+ e5 small` using `dbsf` and `relative-score-fusion` strategies, and below are the comparative score of running them on 3 different tasks: 

| Task | mteb/baseline-bm25s | intfloat/multilingual-e5-small | mteb/hybrid-bm25s-e5-small-rrf | mteb/hybrid-bm25s-e5-small-dbsf | mteb/hybrid-bm25s-e5-small-relative-score-fusion |
|---|---|---|---|---|---|
| NanoSciFactRetrieval | 0.70991 | 0.72458 | 0.75391 | 0.53757 | 0.76656 |
| NanoNFCorpusRetrieval | 0.32504 | 0.28820 | 0.32870 | 0.33824 | 0.35915 |
| NanoSCIDOCSRetrieval | 0.33512 | 0.34378 | 0.36856 | 0.34413 | 0.37242 |

As per these results, models with `rrf` and `relative-score-fusion` strategy are better across all tasks compared to bm25 and e5 small, while with `dbsf` for `NanoSciFactRetrieval` results became worse after fusion, while on rest 2 tasks, they are still better.

ayush1298:
@KennethEnevoldsen Can you check this one? I also have added comparative results of different fusion methods above

ayush1298:
@Samoed @KennethEnevoldsen Can you review these 1 more time? Its pending from long time

docs/get_started/advanced_usage/hybrid_model.md
KennethEnevoldsen: still missing (please add in in the file)

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,143 @@
...
+def hybrid_baseline_encoder_e5_loader(model_name: str, **kwargs) -> HybridSearch:
+    bm25 = mteb.get_model("mteb/baseline-random-encoder")
+    dense = mteb.get_model("intfloat/multilingual-e5-small")
+    return HybridSearch(
+        models=[bm25, dense],
+        fusion_strategy="rrf",
+    )
+
+
+hybrid_baseline_encoder_e5_rrf = ModelMeta(
```
KennethEnevoldsen: I don't see a reason for this model - why would we have a hybrid with random?

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,143 @@
...
+    dense = mteb.get_model("intfloat/multilingual-e5-small")
+    return HybridSearch(
+        models=[bm25, dense],
+        fusion_strategy="relative-score-fusion",
+    )
+
+
+hybrid_bm25s_e5_relative_score_fusion = ModelMeta(
+    loader=hybrid_bm25s_e5_relative_score_fusion_loader,
+    name="mteb/hybrid-relative-score-fusion-baseline-bm25s-multilingual-e5-small",
```
KennethEnevoldsen:
```suggestion
    name="mteb/baseline-hybrid-relative-score-fusion (me5-small+bm25)",
```
I think this is easier to read in the leaderboard. I would do similar renames for the other models

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,143 @@
...
+    model_type=["hybrid"],
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="0.1.0",
+    release_date="2024-02-08",
+    n_parameters=118_000_000,
+    n_embedding_parameters=96_014_208,
+    memory_usage_mb=None,
+    embed_dim=None,
+    license=None,
```
KennethEnevoldsen: whatever e5 has

mteb/models/model_implementations/hybrid_models.py
```diff
@@ -0,0 +1,143 @@
...
+    loader=hybrid_bm25s_e5_relative_score_fusion_loader,
+    name="mteb/hybrid-relative-score-fusion-baseline-bm25s-multilingual-e5-small",
+    model_type=["hybrid"],
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="0.1.0",
+    release_date="2024-02-08",
+    n_parameters=118_000_000,
+    n_embedding_parameters=96_014_208,
+    memory_usage_mb=None,
```
KennethEnevoldsen: whatever e5 has

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,288 @@
...
+            model.index(
+                corpus,
+                task_metadata=task_metadata,
+                hf_split=hf_split,
+                hf_subset=hf_subset,
+                encode_kwargs=encode_kwargs,
+                num_proc=num_proc,
+            )
+
+    def search(
```
KennethEnevoldsen, line 161: much cleaner <3

mteb/models/model_implementations/hybrid_models.py
KennethEnevoldsen: maybe shorten it to "rsf"?

mteb/models/model_implementations/hybrid_models.py
ayush1298: @Samoed Asked for it initially, so we can simply test it. I will remove it.

mteb/models/model_implementations/hybrid_models.py
ayush1298: agree for "rsf". Will change it

ayush1298: In naming, it generally gets created automatically based on the model added. 
https://github.com/ayush1298/mteb/blob/a6f41dacf3560716d1512f785868c8ddab97378a/mteb/models/hybrid_wrappers.py#L132
So adding short-form like these will not work there, though I can change it for already added models
also, why `baseline` word at front of model name, `baseline` is for `bm25s?` didn't get that thing

ayush1298:
I have removed the `random-encoder` model, added metadata of e5 to the hybrid models, renamed `relative-score-fusion` to `rsf,` and also added a performance table in the docs. 
The only thing remain is the naming convention. I agree in general model name will become too big, so need something better there

mteb/models/model_implementations/hybrid_models.py
KennethEnevoldsen: Generally, we keep baseline in from for all our baselines (this would be a baseline model)

We can't overwrite the auto-generated name? 

Generally think the autogenerated name is fine, alternative we should do it as experimental-kwargs?

@Samoed any thoughts or preferences?

docs/get_started/advanced_usage/hybrid_model.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Performance Comparison
+
+To demonstrate the effectiveness of combining different retrieval paradigms, the table below compares the performance (NDCG@10) of individual sub-models against their hybrid combinations using Reciprocal Rank Fusion (RRF), Distribution-Based Score Fusion (DBSF), and Relative Score Fusion (RSF):
+
+| Task | mteb/baseline-bm25s | intfloat/multilingual-e5-small | mteb/hybrid-rrf-baseline-bm25s-multilingual-e5-small | mteb/hybrid-dbsf-baseline-bm25s-multilingual-e5-small | mteb/hybrid-rsf-baseline-bm25s-multilingual-e5-small |
```
KennethEnevoldsen:
```suggestion
| Task | mteb/baseline-bm25s | intfloat/multilingual-e5-small | hybrid using `rrf` | hybrid using `dbsf` | hybrid using 'rsf` |
```

read nicer

docs/get_started/advanced_usage/hybrid_model.md
```diff
@@ -0,0 +1,37 @@
...
+|---|---|---|---|---|---|
+| NanoSciFactRetrieval | 0.70991 | 0.72458 | 0.75391 | 0.53757 | 0.76656 |
+| NanoNFCorpusRetrieval | 0.32504 | 0.28820 | 0.32870 | 0.33824 | 0.35915 |
+| NanoSCIDOCSRetrieval | 0.33512 | 0.34378 | 0.36856 | 0.34413 | 0.37242 |
+
```
KennethEnevoldsen:
```suggestion
| NanoSciFactRetrieval | 0.710 | 0.725 | *0.754* | 0.538 | **0.767** |
| NanoNFCorpusRetrieval | 0.325 | 0.288 | 0.329 | *0.338* | **0.359** |
| NanoSCIDOCSRetrieval | 0.335 | 0.344 | *0.369* | 0.344 | **0.372** |

```
a bit easier to read/skim

KennethEnevoldsen (approved):
other than the unreslved comment and a few minor things below then I think this is good to merge!

mteb/models/model_implementations/hybrid_models.py
ayush1298: > Generally, we keep baseline in from for all our baselines (this would be a baseline model)

but these are hybird models and no longer baseline, right?

> We can't overwrite the auto-generated name?

we can overwrite it manually. I also think auto-generated name is fine, but its just that it will become too long.

KennethEnevoldsen: Hmm but in our case I think they are mostly a display of what you can do with simple hybrid approaches (so yes it is a baseline)

ayush1298: Will wait if @Samoed as any better preference for naming. If not, then I guess, we can go with: 
`mteb/baseline-hybrid-rsf-baseline-bm25s-multilingual-e5-small` and similarly add `baseline` in name of others

Samoed: I think we're using `model/name (some text)` format only for experiments, but I'm not sure if it fit here

mteb/models/hybrid_wrappers.py
```diff
@@ -0,0 +1,292 @@
...
+            self.fusion_name = fusion_strategy
+            if fusion_strategy == "rrf":
+                self._fuse_fn = lambda q_scores, w: fuse_rrf(
+                    q_scores, w, rrf_k=self.rrf_k
+                )
+            elif fusion_strategy == "dbsf":
+                self._fuse_fn = fuse_dbsf
+            elif fusion_strategy in {
+                "relative-score-fusion",
+                "relative_score_fusion",
```
Samoed:
```suggestion
```

Samoed (approved):

mteb/models/model_implementations/hybrid_models.py
ayush1298: so, I guess we can keep current naming only

mteb/models/hybrid_wrappers.py
ayush1298: kept only "rsf" so sync with other places

ayush1298:
I have added "baseline" to model naming

KennethEnevoldsen (approved):
I would be happy to merge this as it is now

PR #4719, opened 2026-05-23
fix: Load aggregated results from task results

### LINKED ISSUE #3974: Load aggregated results from tasks reulsts
Shouldn't we instead fix the cache to load the result regardless of whether the combined results are there?

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/pull/3915#pullrequestreview-3662580302_

ayush1298:
closes #3974 

This PR adds `_check_cache()` to prevent crashes when no cached results exist. It fixes `NEVER` and `ONLY_CACHE` strategies to correctly retain cached results and only raise errors when missing evaluations are actually detected and ensure new evaluations merge into cached results without being overwritten.

### DIFF
--- a/mteb/evaluate.py
+++ b/mteb/evaluate.py
@@ -404,7 +404,7 @@ def evaluate(  # noqa: PLR0913, PLR0914
                 tasks, meta, cache, agg_strategy
             )
         except Exception:
-            if agg_strategy != OverwriteStrategy.ONLY_CACHE:
+            if agg_strategy == OverwriteStrategy.NEVER:
                 raise
             existing_results = None
             missing_eval = dict.fromkeys(tasks.eval_splits, tasks.hf_subsets)
--- a/mteb/evaluate.py
+++ b/mteb/evaluate.py
@@ -407,7 +407,7 @@ def evaluate(  # noqa: PLR0913, PLR0914
             if agg_strategy != OverwriteStrategy.ONLY_CACHE:
                 raise
             existing_results = None
-            missing_eval = True
+            missing_eval = dict.fromkeys(tasks.eval_splits, tasks.hf_subsets)
 
         if (
             existing_results
--- a/mteb/evaluate.py
+++ b/mteb/evaluate.py
@@ -287,11 +287,14 @@ def _check_cache(
 
     if (
         existing_results
-        and overwrite_strategy
-        not in (OverwriteStrategy.ALWAYS, OverwriteStrategy.NEVER)  # noqa: PLR6201
+        and overwrite_strategy != OverwriteStrategy.ALWAYS
         and (
-            not _requires_merge(task, existing_results)
-            or existing_results.is_mergeable(task)
+            overwrite_strategy
+            in [OverwriteStrategy.NEVER, OverwriteStrategy.ONLY_CACHE]  # noqa: PLR6201
+            or (
+                not _requires_merge(task, existing_results)
+                or existing_results.is_mergeable(task)
+            )
         )
     ):
         missing_eval = existing_results.get_missing_evaluations(task)
@@ -304,13 +307,15 @@ def _check_cache(
         OverwriteStrategy.ONLY_CACHE,
     ]:
         if existing_results is None:
+            if overwrite_strategy == OverwriteStrategy.ONLY_CACHE:
+                raise ValueError(
+                    f"overwrite_strategy is set to '{overwrite_strategy.value}' but no results found in cache for task {task.metadata.name}."
+                )
+        else:
             raise ValueError(
-                f"overwrite_strategy is set to '{overwrite_strategy.value}' but no results found in cache for task {task.metadata.name}."
+                f"overwrite_strategy is set to '{overwrite_strategy.value}' and the results file exists for task {task.metadata.name}. "
+                + f"However there are the following missing splits (and subsets): {missing_eval}. To rerun these set overwrite_strategy to 'only-missing'."
             )
-        raise ValueError(
-            f"overwrite_strategy is set to '{overwrite_strategy.value}' and the results file exists for task {task.metadata.name}. "
-            + f"However there are the following missing splits (and subsets): {missing_eval}. To rerun these set overwrite_strategy to 'only-missing'."
-        )
 
     return existing_results, missing_eval
 
@@ -394,7 +399,15 @@ def evaluate(  # noqa: PLR0913, PLR0914
     # AbsTaskAggregate is a special case where we have to run multiple tasks and combine the results
     if isinstance(tasks, AbsTaskAggregate):
         agg_strategy = OverwriteStrategy.from_str(overwrite_strategy)
-        existing_results, missing_eval = _check_cache(tasks, meta, cache, agg_strategy)
+        try:
+            existing_results, missing_eval = _check_cache(
+                tasks, meta, cache, agg_strategy
+            )
+        except Exception:
+            if agg_strategy != OverwriteStrategy.ONLY_CACHE:
+                raise
+            existing_results = None
+            missing_eval = True
 
         if (
             existing_results
@@ -426,7 +439,7 @@ def evaluate(  # noqa: PLR0913, PLR0914
         combined_results = tasks.combine_task_results(results.task_results)
 
         if existing_results:
-            combined_results = combined_results.merge(existing_results)
+            combined_results = existing_results.merge(combined_results)
 
         if cache:
             cache.save_to_cache(
--- a/mteb/evaluate.py
+++ b/mteb/evaluate.py
@@ -394,16 +394,16 @@ def evaluate(  # noqa: PLR0913, PLR0914
 
     model, meta, model_name, model_revision = _sanitize_model(model)
     _check_model_modalities(meta, tasks)
+    overwrite_strategy = OverwriteStrategy.from_str(overwrite_strategy)
 
     # AbsTaskAggregate is a special case where we have to run multiple tasks and combine the results
     if isinstance(tasks, AbsTaskAggregate):
-        agg_strategy = OverwriteStrategy.from_str(overwrite_strategy)
         try:
             existing_results, missing_eval = _check_cache(
-                tasks, meta, cache, agg_strategy
+                tasks, meta, cache, overwrite_strategy
             )
         except Exception:
-            if agg_strategy == OverwriteStrategy.NEVER:
+            if overwrite_strategy == OverwriteStrategy.NEVER:
                 raise
             existing_results = None
             missing_eval = dict.fromkeys(tasks.eval_splits, tasks.hf_subsets)
@@ -411,7 +411,7 @@ def evaluate(  # noqa: PLR0913, PLR0914
         if (
             existing_results
             and not missing_eval
-            and agg_strategy != OverwriteStrategy.ALWAYS
+            and overwrite_strategy != OverwriteStrategy.ALWAYS
         ):
             logger.info(
                 f"Results for {tasks.metadata.name} already exist in cache. Skipping evaluation and loading results."
@@ -489,7 +489,6 @@ def evaluate(  # noqa: PLR0913, PLR0914
             exceptions=exceptions,
         )
 
-    overwrite_strategy = OverwriteStrategy.from_str(overwrite_strategy)
     existing_results, missing_eval = _check_cache(task, meta, cache, overwrite_strategy)
 
     if (
--- a/mteb/evaluate.py
+++ b/mteb/evaluate.py
@@ -285,27 +285,26 @@ def _check_cache(
         if cache_results:
             existing_results = cache_results
 
+    dont_overwrite = overwrite_strategy in {
+        OverwriteStrategy.NEVER,
+        OverwriteStrategy.ONLY_CACHE,
+    }
+    is_mergable = existing_results is not None and (
+        not _requires_merge(task, existing_results)
+        or existing_results.is_mergeable(task)
+    )
+
     if (
         existing_results
         and overwrite_strategy != OverwriteStrategy.ALWAYS
-        and (
-            overwrite_strategy
-            in [OverwriteStrategy.NEVER, OverwriteStrategy.ONLY_CACHE]  # noqa: PLR6201
-            or (
-                not _requires_merge(task, existing_results)
-                or existing_results.is_mergeable(task)
-            )
-        )
+        and (dont_overwrite or is_mergable)
     ):
         missing_eval = existing_results.get_missing_evaluations(task)
     else:
         missing_eval = dict.fromkeys(task.eval_splits, task.hf_subsets)
         existing_results = None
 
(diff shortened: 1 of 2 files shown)


mteb/evaluate.py
```diff
@@ -350,6 +398,31 @@ def evaluate(  # noqa: PLR0913, PLR0914
 
     # AbsTaskAggregate is a special case where we have to run multiple tasks and combine the results
     if isinstance(tasks, AbsTaskAggregate):
+        agg_strategy = OverwriteStrategy.from_str(overwrite_strategy)
+        try:
+            existing_results, missing_eval = _check_cache(
+                tasks, meta, cache, agg_strategy
+            )
+        except Exception:
+            if agg_strategy != OverwriteStrategy.ONLY_CACHE:
```
Samoed: Shouldn't we raise error only on `ONLY_CACHE`?
```suggestion
            if agg_strategy == OverwriteStrategy.ONLY_CACHE:
```

tests/test_evaluate.py
```diff
@@ -226,6 +236,41 @@ def test_evaluate_aggregated_task():
...
+       to combining the cached subtask results under ONLY_CACHE strategy.
+    2. Once the aggregate file is generated and saved, it can be loaded directly
+       on subsequent runs without re-evaluating or calling fallback.
+    """
+    model = mteb.get_model("mteb/baseline-random-encoder")
+    task = MockAggregatedTask()
+    cache = ResultCache(tmp_path)
+
+    for subtask in task.metadata.tasks:
+        mteb.evaluate(model, subtask, cache=cache)
```
Samoed:
```suggestion
    mteb.evaluate(model, task.metadata.tasks, cache=cache)
```

mteb/evaluate.py
```diff
@@ -350,6 +393,23 @@ def evaluate(  # noqa: PLR0913, PLR0914
 
     # AbsTaskAggregate is a special case where we have to run multiple tasks and combine the results
     if isinstance(tasks, AbsTaskAggregate):
+        agg_strategy = OverwriteStrategy.from_str(overwrite_strategy)
+        existing_results, missing_eval = _check_cache(tasks, meta, cache, agg_strategy)
```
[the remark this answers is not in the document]
ayush1298: @Samoed these is reason why I set it `agg_strategy != OverwriteStrategy.ONLY_CACHE:`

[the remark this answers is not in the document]
Samoed: But then wouldn't be impossible to run any aggregated with `agg_strategy != OverwriteStrategy.ONLY_CACHE`?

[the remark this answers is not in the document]
ayush1298: No, because it will go in exception only in 2 cases, below is how it will work: 
1. For `overwrite_strategy`, `only-missing`(cache miss, cache hit complete, cache hit incomplete) and `default`, try block will succeed
2. For `overwrite_strategy = never`, cache miss and cache hit(complete) will work normally, while cache hit(incomplete), `_check_cache` will itself raise `ValueError,` and which will be catched and raised by `except` block
3. For `overwrite_strategy = only_cache`, cache hit(complete) will work normally, while for cache hit(incomplete), `_check_cache` will itself raise `ValueError,` which is catched by `except` block, and it will just override `existing_results`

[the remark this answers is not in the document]
Samoed: Can we then just `if agg_strategy == OverwriteStrategy.NEVER`?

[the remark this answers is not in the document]
ayush1298: yes, will change it

mteb/evaluate.py
```diff
@@ -272,6 +272,54 @@ def _requires_merge(task: AbsTask, existing_results: TaskResult) -> bool:
...
+        and overwrite_strategy != OverwriteStrategy.ALWAYS
+        and (
+            overwrite_strategy
+            in [OverwriteStrategy.NEVER, OverwriteStrategy.ONLY_CACHE]  # noqa: PLR6201
+            or (
+                not _requires_merge(task, existing_results)
+                or existing_results.is_mergeable(task)
+            )
+        )
+    ):
```
KennethEnevoldsen: This one is very hard to read can we assign it to a variable to document it:



```suggestion
    dont_overwrite = overwrite_strategy in [OverwriteStrategy.NEVER, OverwriteStrategy.ONLY_CACHE]
    is_mergable = not _requires_merge(task, existing_results) or existing_results.is_mergeable(task)
    
    if (
        existing_results
        and overwrite_strategy != OverwriteStrategy.ALWAYS
        and (dont_overwrite or is_mergable)
        )
    ):
```

mteb/evaluate.py
```diff
@@ -272,6 +272,54 @@ def _requires_merge(task: AbsTask, existing_results: TaskResult) -> bool:
...
+        missing_eval = existing_results.get_missing_evaluations(task)
+    else:
+        missing_eval = dict.fromkeys(task.eval_splits, task.hf_subsets)
+        existing_results = None
+
+    if missing_eval and overwrite_strategy in [  # noqa: PLR6201
+        OverwriteStrategy.NEVER,
+        OverwriteStrategy.ONLY_CACHE,
+    ]:
```
KennethEnevoldsen: re-use from above

tests/test_evaluate.py
```diff
@@ -226,6 +236,39 @@ def test_evaluate_aggregated_task():
...
+    Verifies both that:
+    1. If the aggregate file is missing but all sub-tasks are cached, it falls back
+       to combining the cached subtask results under ONLY_CACHE strategy.
+    2. Once the aggregate file is generated and saved, it can be loaded directly
+       on subsequent runs without re-evaluating or calling fallback.
+    """
+    model = mteb.get_model("mteb/baseline-random-encoder")
+    task = MockAggregatedTask()
+    cache = ResultCache(tmp_path)
+    mteb.evaluate(model, task.metadata.tasks, cache=cache)
```
KennethEnevoldsen, line 293:
```suggestion
    # evaluate tasks individually
    mteb.evaluate(model, task.metadata.tasks, cache=cache)
```

tests/test_evaluate.py
```diff
@@ -226,6 +236,39 @@ def test_evaluate_aggregated_task():
...
+
+    path = cache.get_task_result_path(
+        task.metadata.name,
+        model.mteb_model_meta.model_name_as_path(),
+        model.mteb_model_meta.revision,
+    )
+    assert not path.exists()
+
+    results = mteb.evaluate(model, task, cache=cache, overwrite_strategy="only-cache")
```
KennethEnevoldsen, line 302:
```suggestion
    # should be able to load results even if aggregated task results do not exist
    results = mteb.evaluate(model, task, cache=cache, overwrite_strategy="only-cache")
```

tests/test_evaluate.py
```diff
@@ -226,6 +236,39 @@ def test_evaluate_aggregated_task():
...
+    assert len(results.task_results) == 1
+    score1 = results.task_results[0].get_score()
+    assert path.exists()
+
+    cached_results = mteb.evaluate(
```
KennethEnevoldsen, line 317: I am not sure what this tests - it would work regardless of if it works from the task files or the aggregated task files

KennethEnevoldsen (commented):
@ayush1298 can you please add a minimal PR description and mark the pr with a "fix:" (this indicates that it should bump patch version)

tests/test_evaluate.py
ayush1298: It was slightly incorrect previously; I corrected it now. So, basically 1st time, when it calls evaluate, it will cache results of all tasks and convert it to an aggregate result and save it, and now we will delete those individual task files and run evaluate 2nd time, so if it try to take results from individual files, as they have deleted, it will give error, but if it run successfully then we can say it uses aggregated results file

ayush1298:
> @ayush1298 can you please add a minimal PR description and mark the pr with a "fix:" (this indicates that it should bump patch version)

updated PR description, added fix to name, and fixed other issues

mteb/evaluate.py
```diff
@@ -350,6 +397,31 @@ def evaluate(  # noqa: PLR0913, PLR0914
 
     # AbsTaskAggregate is a special case where we have to run multiple tasks and combine the results
     if isinstance(tasks, AbsTaskAggregate):
+        agg_strategy = OverwriteStrategy.from_str(overwrite_strategy)
+        try:
+            existing_results, missing_eval = _check_cache(
+                tasks, meta, cache, agg_strategy
+            )
+        except Exception:
+            if agg_strategy == OverwriteStrategy.NEVER:
```
KennethEnevoldsen: I don't think we need a try except here - we should be able to implement this without it

mteb/evaluate.py
```diff
@@ -350,6 +397,31 @@ def evaluate(  # noqa: PLR0913, PLR0914
 
     # AbsTaskAggregate is a special case where we have to run multiple tasks and combine the results
     if isinstance(tasks, AbsTaskAggregate):
+        agg_strategy = OverwriteStrategy.from_str(overwrite_strategy)
```
KennethEnevoldsen: why the conversion here? Don't we do it earlier?

mteb/evaluate.py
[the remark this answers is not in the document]
ayush1298: @KennethEnevoldsen This is reason why I added try-except, I think it seems reasonable to me to keep it

mteb/evaluate.py
ayush1298: It didn't. But we are doing it here and later again, so shifting this to start of function

mteb/evaluate.py
Samoed: @ayush1298 Can you adress this comment?

mteb/evaluate.py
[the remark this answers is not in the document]
ayush1298: @Samoed This one

ayush1298:
@KennethEnevoldsen Can you review this one? There is just one comment above related to try-except, which I have added because Copilot suggested. Other than that, I think this is good

mteb/evaluate.py
Samoed: I see your comment https://github.com/embeddings-benchmark/mteb/pull/4719#discussion_r3308928113, but maybe we can update `_check_cache` to not raise errors in such cases or handle it somehow differently? E.g. move scores check in `_check_cache`

ayush1298: I have changed it, though I am not completely sure, as it's going in a lot of conditions. Can you check it once?

mteb/evaluate.py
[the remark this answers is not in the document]
KennethEnevoldsen: I will take this in the comment below

tests/test_evaluate.py
```diff
@@ -101,6 +101,16 @@ def test_evaluate_with_cache(
         "main score should match the expected value"
     )
 
+    # test cache never strategy
+    never_results = mteb.evaluate(model, task, cache=cache, overwrite_strategy="never")
+    assert never_results[0].get_score() == expected_score
```
KennethEnevoldsen: what are you trying to test here? If you want to check "never" you need to both check that cache is used and that we don't write to cache.

mteb/evaluate.py
```diff
@@ -272,6 +272,60 @@ def _requires_merge(task: AbsTask, existing_results: TaskResult) -> bool:
     return False
 
 
+def _check_cache(
+    task: AbsTask,
+    meta: ModelMeta,
+    cache: ResultCache | None,
+    overwrite_strategy: OverwriteStrategy,
+) -> tuple[TaskResult | None, dict[str, list[str]]]:
+    """Load cached results, handle early skipping if up-to-date, or validate compatibility."""
```
KennethEnevoldsen: document return

tests/test_evaluate.py
KennethEnevoldsen: I would add a test to ensure that we don't write when results exists and I would maybe also factor it out to its own test (`test_evaluate_with_overwrite_strategy_never`)

KennethEnevoldsen (approved):
Only a few minor things

mteb/evaluate.py
KennethEnevoldsen: I think it is a lot cleaner - conditions looks solid to me - and yeah there are quite a few. Maybe we should comment them better.

tests/test_evaluate.py
ayush1298: added these test

ayush1298:
1 check is failing somehow because of the HF connection error. Tried re-running that, but still it failed

Samoed:
Restarted

ayush1298:
@KennethEnevoldsen @Samoed Can we merge this one?

KennethEnevoldsen:
I am happy @Samoed ?

mteb/evaluate.py
```diff
@@ -272,6 +272,88 @@ def _requires_merge(task: AbsTask, existing_results: TaskResult) -> bool:
...
+              Returns None if we need to re-run from scratch.
+            - A dictionary mapping evaluation split names to a list of missing subset names.
+              Returns an empty dict if no splits or subsets are missing.
+    """
+    # Load results from the cache if the overwrite strategy allows it
+    existing_results: TaskResult | None = None
+    if cache and overwrite_strategy != OverwriteStrategy.ALWAYS:
+        cache_results = cache.load_task_result(task.metadata.name, meta)
+        if cache_results:
+            existing_results = cache_results
```
Samoed:
```suggestion
        existing_results = cache.load_task_result(task.metadata.name, meta) or None
```

mteb/evaluate.py
```diff
@@ -272,6 +272,88 @@ def _requires_merge(task: AbsTask, existing_results: TaskResult) -> bool:
...
+
+    dont_overwrite = overwrite_strategy in {
+        OverwriteStrategy.NEVER,
+        OverwriteStrategy.ONLY_CACHE,
+    }
+
+    # Check if the cached results are compatible/mergeable with the current task definition
+    is_mergable = existing_results is not None and (
```
Samoed:
```suggestion
    is_mergeable = existing_results is not None and (
```

mteb/evaluate.py
```diff
@@ -272,6 +272,88 @@ def _requires_merge(task: AbsTask, existing_results: TaskResult) -> bool:
...
+        not _requires_merge(task, existing_results)
+        or existing_results.is_mergeable(task)
+    )
+
+    # Determine missing splits and subsets
+    if (
+        existing_results
+        and overwrite_strategy != OverwriteStrategy.ALWAYS
```
Samoed:
```suggestion
```

Samoed (approved):

mteb/evaluate.py
ayush1298: changed. `or None` part is not required as its intialized with default value `None`

ayush1298:
Applied all changes from review

PR #4730, opened 2026-05-25
Add more mveb plot scripts

Samoed:
I created scripts for 3 plots:
1. Comparison between mveb and other benchmark modalities. It's a bit usless because not a lot of models have coverage on other benchmarks
benchmark_cross_modality_correlation
2. Model performance vs model size
plot_size_vs_performance
3. Average task score + std
plot_task_difficulty

I'm ushure if we want to add something from this to paper from them, but might be useful


AdnanElAssadi:
Nice! I like what 2 and 3 are saying. I think we can certainly add these in appendix if we can't find space. We may be able to squeeze in 2 somewhere in main text. I think we can maybe have the second plot a bit nicer as well and try to not have overlaps.

Samoed:
I also created task correlation plot, might be interesting. I'll merge this pr into your branch
mveb_task_correlation_plot

PR #4733, opened 2026-05-25
Add script for analysis of video vs audio-video

shriyasudhakar:
Added `analyze_video_vs_audio_video.py` script for comparing video-only vs video+audio performance across all paired tasks.

The script runs three analyses:
1. **v vs va paired tasks**: computes per-model audio delta (va − v score) across classification, clustering, zeroshot, QA, and retrieval tasks
2. **AV-intentional vs incidental datasets**: checks whether audio gains are larger on datasets explicitly designed with audio in mind
3. **Cross-modal retrieval (v2a / a2v)**: high retrieval scores suggest audio and video embeddings are well-aligned, implying audio may be redundant

**Key findings:**
- Audio helps on AV-intentional datasets (mean delta +0.015) but slightly hurts on incidental ones (−0.017)
- Biggest gains on AudioCaps (+0.159) and AVMeme-Exam (+0.129)
- Audio hurts on action recognition datasets (Kinetics, UCF101)
- Cross-modal retrieval scores are moderate (0.13–0.43); suggests audio adds unique signal, not redundant

### DIFF
--- a/scripts/analyze_video_and_audio_video.py
+++ b/scripts/analyze_video_and_audio_video.py
@@ -1,0 +1,278 @@
+"""Compare video-only vs video+audio performance across all paired tasks.
+
+Three analyses:
+1. v vs va paired tasks (classification, clustering, zeroshot, QA, retrieval)
+   for each paired dataset, computes per-model audio delta (va - v score)
+2. AV-intentional vs incidental datasets -- are gains larger where audio matters?
+3. Cross-modal retrieval (v2a, a2v) -- high score = audio signal is redundant
+   with video (they share the same embedding space)
+
+Usage:
+    # Download remote results first (one-time, ~GB):
+    python scripts/analyze_video_vs_audio_video.py --download
+
+    # Then analyse (or point at a local results dir):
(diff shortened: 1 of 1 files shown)


scripts/analyze_video_and_audio_video.py
```diff
@@ -0,0 +1,288 @@
...
+
+import argparse
+from collections import defaultdict
+
+import pandas as pd
+
+import mteb
+
+# datasets annotated with audio as primary/co-primary signal
+AV_INTENTIONAL = {
```
Samoed, line 31: How decided if audio is intentional or not?

shriyasudhakar: They were chosen based on if audio was a main signal in the dataset's design, such as emotion recognition from speech + video (RAVDESS, MELD), audio-visual event detection (AVE), sound classification (VGGSound). 

After taking a second look I think Human-Animal-Cartoon doesn't fit this category well, since it is primarily visual even though audio is also provided.

Samoed:
Will merge since we finished paper for now

PR #4740, opened 2026-05-26
leaderboard: drop O(N) scores fingerprint from update_tables cache key

florian-hoenicke:
## Summary

`update_tables` (the callback that re-renders the summary / per-task / per-language tables when a filter or tab changes) is wrapped in `cachetools.cached`. Its cache key was built by iterating the entire `scores` list of dicts, constructing a `(model_name, task_name, score)` tuple for every entry, sorting them, and hashing the result.

On large benchmarks the `scores` list is long-form (one entry per `(model, task, subset)`) and can reach ~10^5 entries on `MTEB(Multilingual, v2)`. The cache key alone then takes ~300 ms per call, and `update_tables` fires from three triggers on every filter change (`models.change`, `task_select.change`, `lang_select.change`), so a single click pays ~1 s of pure key-hashing overhead — **even when the result is already cached** — before any actual work begins.

This PR replaces the score-content fingerprint with the `(benchmark_name, languages)` pair that already deterministically produces those scores:

- `update_scores_on_lang_change` (the producer when the user picks languages) is itself `cachetools.cached` on exactly `(benchmark_name, languages)` ([app.py:856-870](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/leaderboard/app.py#L856)).
- `_cache_on_benchmark_select` (the producer when the user picks a benchmark) is keyed on `benchmark_name` and returns the full-language scores; the lang-select widget is then set to the benchmark's full languages list, so `(benchmark_name, languages)` still uniquely identifies the scores ([app.py:219-260, 808-812](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/leaderboard/app.py#L219)).

`scores` is read-only inside both `update_tables` and `update_models` — never mutated — so hashing the producer key instead of fingerprinting the value is equivalent.

The new key cost is O(#languages + #tasks + #models), all bounded by the benchmark schema (typically <2k items combined).

## Measured improvement

Per-click cache-key overhead (3 triggers per filter change, measured with synthetic `scores` of N entries):

| N       | Before    | After       |
|---------|-----------|-------------|
| 5,000   | 9.4 ms    | 0.04 ms     |
| 20,000  | 52.3 ms   | 0.05 ms     |
| 80,000  | 312 ms    | 0.07 ms     |
| 200,000 | 981 ms    | 0.06 ms     |

Reproduction:

```python
import time

scores = [{'model_name': f'org/model-{i % 250}',
           'task_name': f'Task{(i // 250) % 200}',
           'score': 0.5 + (i % 100)/200} for i in range(200_000)]
tasks = sorted({s['task_name'] for s in scores})
models = sorted({s['model_name'] for s in scores})

def key_old():
    sig = [(e.get('model_name'), e.get('task_name'), e.get('score')) for e in scores]
    return hash((hash(tuple(sorted(sig))),
                 hash(tuple(sorted(tasks))),
                 hash(tuple(sorted(models))),
                 hash('MTEB(M)'),
                 hash(tuple(sorted(['eng-Latn'])))))

t = time.perf_counter()
for _ in range(3):
    key_old()
print(f'old key, 3 calls: {(time.perf_counter()-t)*1000:.1f} ms')
```

## Test plan

- [x] `app.py` still parses
- [x] Diff is localised to the inner `_cache_key_for_update_tables` function; signature and return type unchanged
- [ ] CI green
- [ ] Sanity-check on the deployed HF Space: filter/tab interactions on `MTEB(Multilingual, v2)` feel snappier

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -1146,20 +1146,14 @@ def get_leaderboard_app(  # noqa: PLR0914
         def _cache_key_for_update_tables(
             scores, tasks, models_to_keep, benchmark_name, languages
         ):
-            # Build a deterministic fingerprint from score content.
-            # This keeps cache hits for equivalent data while invalidating on language-driven score changes.
-            score_signature = []
-            for entry in scores:
-                score_signature.append(
-                    (
-                        entry.get("model_name"),
-                        entry.get("task_name"),
-                        entry.get("score"),
-                    )
-                )
-            scores_hash = hash(tuple(sorted(score_signature)))
+            # `scores` is uniquely determined by `(benchmark_name, languages)` via
+            # `update_scores_on_lang_change`'s cache (and by `_cache_on_benchmark_select`
+            # for the no-language-filter initial state, where languages == full benchmark
+            # languages). Hashing those is O(#languages) instead of iterating + sorting the
+            # `scores` list, which is O(N log N) over up to ~10^5 entries on large
+            # benchmarks like MTEB(Multilingual, v2) and dominates per-click latency since
+            # `update_tables` fires from three triggers per filter change.
             tasks_hash = hash(tuple(sorted(tasks))) if tasks is not None else None
-            # Sort models_to_keep to ensure consistent hash regardless of input order
             models_hash = (
                 hash(tuple(sorted(models_to_keep)))
                 if models_to_keep is not None
@@ -1169,9 +1163,7 @@ def get_leaderboard_app(  # noqa: PLR0914
             lang_hash = (
                 hash(tuple(sorted(languages))) if languages is not None else None
             )
-            key = hash((scores_hash, tasks_hash, models_hash, bench_hash, lang_hash))
-
-            return key
+            return hash((bench_hash, lang_hash, tasks_hash, models_hash))
 
         @cachetools.cached(
             cache={},


Samoed:
> Per-click cache-key overhead (3 triggers per filter change, measured with synthetic scores of N entries):

How did you test this?

Samoed (approved):
Yes, overall I agree with this

florian-hoenicke:
> How did you test this?

Just an estimation using synthetic numbers. Let me check on the real deployed app.

florian-hoenicke:
CLAUDE:
Honest answer: the table in the PR is a **synthetic micro-benchmark** of `_cache_key_for_update_tables` in isolation, not the deployed app. I picked N=5k..200k by extrapolating "548 models × 131 tasks × ~1k languages" from #4467; that overestimated — `_get_scores(format="long")` collapses further than I assumed.

Just re-measured against real data: built a venv with this repo, loaded the actual cached `BenchmarkResults` via `ResultCache()._load_from_cache(rebuild=False)`, called `all_results.select_tasks(b.tasks).join_revisions()._get_scores(format="long")` for the target benchmark, then timed the old key function (verbatim from upstream/main) vs the new one against that real `scores` list (5 iters, warm).

| Benchmark | len(scores) | Old key (1 call) | New key (1 call) | Per-click (3 triggers) saved |
|---|---|---|---|---|
| MTEB(Multilingual, v2) | 26,564 | 13.0 ms | 0.031 ms | ~39 ms |
| MTEB(eng, v2) | 9,980 | 4.3 ms | 0.011 ms | ~13 ms |

So the real win is **~40 ms per filter click** on the largest benchmark, not the ~1 s the synthetic table implied. Still a strict improvement on top of #4467 and zero correctness risk (the new key mirrors what `update_scores_on_lang_change`'s own cache is already keyed on; `scores` is read-only inside `update_tables`/`update_models` and deterministically produced by `(benchmark_name, languages)`).

I did **not** drive the full Gradio app in a browser end-to-end. Happy to do that, or to amend the PR body with these real numbers if you'd prefer.

KennethEnevoldsen:
~~Tried to make some tests for the leaderboard earlier: Can we run with:
https://github.com/embeddings-benchmark/mteb/blob/main/scripts/bench_leaderboard.py~~


I see that is doesn't test that part of the code

KennethEnevoldsen:
Tested this locally couldn't find any surprising bugs

PR #4741, opened 2026-05-26
ModelMeta: per-modality active params

florian-hoenicke:
Our omni model has a different number of active parameters depending on the use case (text vs image vs video vs audio). This adds an optional `n_active_parameters_by_modality` to `ModelMeta` and makes leaderboard tables pick the value matching the benchmark's modalities; existing single-value behavior is the fallback.


mteb/models/model_implementations/jina_models.py
```diff
@@ -1067,6 +1073,12 @@ def encode(
     release_date="2026-04-01",
     n_parameters=985984512,
     n_embedding_parameters=98503680,
+    n_active_parameters_by_modality={
+        "text": 113265408,
+        "image": 248489472,
+        "video": 248489472,
+        "audio": 799476480,
```
Samoed, line 1080: That's an interesting usage. Why do you have for image and video so few parameters? That's might be a case for low results on video https://github.com/embeddings-benchmark/results/pull/546. CC @AdnanElAssadi56

florian-hoenicke: agree. we wanted to have a small model but now i wish I would have taken a larger encoder.

Samoed: I mean why txt encoders have much bigger number of parameters than image/video/audio?

tests/test_models/test_model_meta.py
```diff
@@ -483,3 +484,21 @@ def test_model_meta_dependencies_not_installed_group():
...
+        update={
+            "n_active_parameters_by_modality": {
+                "text": 100,
+                "image": 500,
+                "audio": 900,
+            }
+        }
+    )
+    assert meta.n_active_parameters_for({"text"}) == 100
+    assert meta.n_active_parameters_for({"text", "image"}) == 500
```
KennethEnevoldsen, line 500: Wouldn't this be 600?

mteb/models/model_meta.py
```diff
@@ -249,6 +250,16 @@ def n_active_parameters(self) -> int | None:
             return self.n_parameters - self.n_embedding_parameters
         return None
 
+    def n_active_parameters_for(
```
KennethEnevoldsen, line 253: I don't think this is correct - it will just take the first match

mteb/models/model_implementations/jina_models.py
KennethEnevoldsen, line 1: If we are making this change, we would want to annotate this for all models

KennethEnevoldsen (commented):
Thanks for the PR @florian-hoenicke -- For PRs like these it is always good to open an issue first to discuss if is something that we want to add. it also allow us to discuss the implementation before starting anything.

I think there is a lot of work here on metadata before we can merge this in. There is also the problem that even for a single task (text-image) and a models with 100 and 500 parameters for each modality respectively, we have a model with 100-500 parameters depending on what part of the task it is running and for certain tasks it would have image text in one document (e.g. query: "find other images of this animal: [image]"). Here is would be 600 instead.

I have made some comments on the code, but let us start with discussion whether it is something we want to implement and how.

PR #4754, opened 2026-05-28
leaderboard: client IP retrieval in Hugging Face Spaces

q275343119:
Previously, the client IP was obtained using `request.client.host`.
However, when the service is deployed on Hugging Face Spaces, this value returns an internal network IP address (for example: `10.16.3.80`) instead of the real client IP.

This PR updates the IP extraction logic to use the `x-forwarded-for` header, which correctly preserves the original client IP behind the proxy/load balancer.

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -379,6 +379,7 @@ def _get_session_id(request: gr.Request) -> str:
 
 def _get_client_ip(request: gr.Request) -> str:
     """Common proxy/CDN headers. X-Forwarded-For may contain:
+
     client, proxy1, proxy2
     """
     headers = request.headers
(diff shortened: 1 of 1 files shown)


Samoed (commented):
Can you format your PR? I can't make changes to it

Samoed (approved):

KennethEnevoldsen (approved):

q275343119:
I tested locally with ·ruff check --fix .· and it returned “All checks passed!”. However, the online run-linting is still failing. This might be due to differences in the local Ruff version or configuration. I’ll take a closer look and fix it shortly.

Samoed:
> local Ruff version or configuration

You can get current version with `uv sync --group lint`, `pip install -e . --group lint` or `pip install ruff==0.14.0` https://github.com/embeddings-benchmark/mteb/blob/4869638002b6598635956a33cb66d4e26f078fcb/pyproject.toml#L146

PR #4756, opened 2026-05-28
fix: add support for loading local model

Samoed:
Add support for loading a local model with `mteb.get_model`. Current behavior:
```python
import mteb

model = mteb.get_model("/path/to/model")
# ...
# NotImplementedError: No model implementation is available for this model.
```

I think this could be helpful for testing trained models

### DIFF
--- a/mteb/models/get_model_meta.py
+++ b/mteb/models/get_model_meta.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import difflib
 import logging
 import warnings
+from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen (commented):
Hmm - not sure we want to take on this part. Sentence trf already support loading local models (though it doens't auto-detect the type, not sure we should either)

Wouldn't we rather just add the documentation on how to load a local model? (people can pass a sentence trf model to evaluate just fine).

Samoed:
> people can pass a sentence trf model to evaluate just fine

I don't think we need that. We have some examples in docs

PR #4760, opened 2026-06-02
feat: Add mteb/api FastAPI service for new leaderboard

Samoed:
A FastAPI service that powers the leaderboard frontend. New package under `mteb/api/`.

### Module layout (`mteb/api/`)

| Module               | Role                                                                                                                                                                                           |
|----------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `app.py`             | App factory: GZip + CORS + Prometheus middleware, mounts `router` under `/v1` and `infra_router` at root, lifespan-driven warmup, `/og` static mount with 1-day `Cache-Control`.               |
| `routes.py`          | All HTTP endpoints. Cached endpoints serve pre-built bytes via `_cached_json` (handles 304, gzip negotiation, `Cache-Control`); uncached ones return pydantic schemas.                         |
| `schemas.py`         | Pydantic response models. `snake_case` in Python, `camelCase` over the wire (matches `leaderboardv2/src/lib/types.ts`).                                                                        |
| `adapters.py`        | Memoised wrappers around `Schema.from_*` constructors so each benchmark/task/model pays construction cost once.                                                                                |
| `aggregators.py`     | Pure builders (`build_benchmark_summary`, `build_benchmark_per_language`, `build_benchmark_leaders`, `build_model_scores`, `build_task_scores`) — turn long polars frames into schema objects. |
| `frames.py`          | Process-wide long-results polars frames, loaded once. Sits below `cache` so `aggregators` can depend on it without dragging in the bytes cache.                                                |
| `cache.py`           | `CacheLayer` generic: single-flight per-key locks + LRU store + Prometheus labels. Holds the warm serialised bytes routes hand out.                                                            |
| `serialization.py`   | Off-thread schema → JSON bytes + gzip pair, used by the bytes cache.                                                                                                                           |
| `warmup.py`          | Lifespan-time warmup: builds frames, prewarms schema caches, optionally preloads every summary in the background (semaphore-capped).                                                           |
| `metrics.py`         | Prometheus middleware + counters/histograms/in-flight gauge; cache-outcome counter; `/metrics` renderer.                                                                                       |
| `otel.py`            | OTLP HTTP tracing setup — one span per request, W3C `traceparent` propagation. No-op unless `OTEL_EXPORTER_OTLP_ENDPOINT` is set.                                                              |
| `icons.py`           | Benchmark icon proxy + cache, so the browser gets immutable 1-year cache instead of upstream's 5-minute one.                                                                                   |
| `settings.py`        | `pydantic-settings` knobs: `CORS_ORIGINS`, `PRELOAD`, `CACHE_REPO`, `OG_DIR`, `PREWARM_MAX_WORKERS`, `PRELOAD_CONCURRENCY`, `HTTP_MAX_AGE`, `DISK_CACHE`, log level, OTEL vars.                |
| `static/favicon.png` | Shipped as package data.                                                                                                                                                                       |

### Endpoint map

Data routes under `/v1`, infra at root.

```
/health                                   GET   liveness
/metrics                                  GET   prometheus scrape
/robots.txt, /favicon.ico, /og/*          GET   static / proxied assets
/v1/icon/{name:path}                      GET   benchmark icon proxy

/v1/benchmarks                            GET   flat list
/v1/benchmarks/menu                       GET   nested menu tree
/v1/benchmarks/{name}                     GET   single benchmark metadata
/v1/benchmarks/{name}/scores              GET   full summary (legacy alias: /summary)
/v1/benchmarks/{name}/per-language        GET   per-language rows
/v1/benchmarks/{name}/leaders             GET   slim home-page leader tiles

/v1/tasks                                 GET   flat list
/v1/tasks/{name}                          GET   single task metadata
/v1/tasks/{name}/scores                   GET   per-task scores

/v1/models                                GET   flat list
/v1/models/{name}                         GET   single model metadata
/v1/models/{name}/scores                  GET   per-model scores
```

### Request flow

```
request → CORSMiddleware → GZipMiddleware → PrometheusMiddleware → router
            │
            ├── cached endpoints ──→ cache.CacheLayer (single-flight)
            │                          │ miss → aggregators.build_* → serialize_schema (off-thread)
            │                          └ hit  → cached (bytes, gzip, etag)
            │                                    → Response with 304 / Vary / Cache-Control
            │
            └── uncached endpoints → adapters.*_to_schema → pydantic-core JSON
```

Long-frame source of truth lives in `frames.py` (loaded once at startup or first request; persisted to `~/.cache/mteb/leaderboard/`, invalidated by HF dataset commit SHA).


mteb/api/static/favicon.png
KennethEnevoldsen, line 1: ? do we want use this as the favicon?

scripts/generate_og_images_mpl.py
Samoed, line 1: `generate_og_images_mpl.py` is for generating the same benchmark "preview" as `generate_og_images`, but using only Matplotlib. Currently in Docker, `generate_og_images` is used, but it renders from a browser with Playwright. I'm using it because it looks a bit nicer than Matplotlib

Dockerfile
KennethEnevoldsen, line 1: How big is this once built?

tests/test_benchmarks/test_get_benchmarks.py
```diff
@@ -34,7 +34,7 @@ def test_benchmark_aliases(alias, full_name):
 
 
 def test_benchmark_on_leaderboard():
-    on_leaderboard = "MTEB(Multilingual, v2)"
+    on_leaderboard = "RTEB(eng, beta)"
```
KennethEnevoldsen, line 37: why is this changed?

scripts/generate_og_images.py
KennethEnevoldsen, line 1: would be nice to combine all of these scripts into one folder

scripts/bench_leaderboard.py
KennethEnevoldsen, line 1: delete this file?

mteb/benchmarks/benchmark.py
```diff
@@ -92,6 +113,20 @@ class Benchmark:
     display_name: str | None = None
     language_view: list[str] | Literal["all"] = field(default_factory=list)
     benchmark_hf_repo: str | None = None
+    new_version: Sequence[str] | None = None
```
KennethEnevoldsen: superseded_by? (for consistency)

mteb/benchmarks/benchmark.py
```diff
@@ -92,6 +113,20 @@ class Benchmark:
...
+    # Api aggregation functions
+    aggregations: Sequence[BenchmarkAggregation] = (
```
KennethEnevoldsen, line 222: if we have this here - we might even be able to deprecate the RTEBBenchmark object

KennethEnevoldsen: maybe some of the other ones as well

mteb/languages/iso_mappings.py
```diff
@@ -50,6 +50,45 @@ def _get_iso3_to_default_script() -> dict[str, str]:
         return cast("dict[str, str]", json.load(f))
 
 
+# Short, reader-friendly overrides for ISO 15924 names whose canonical labels
+# would read awkwardly when appended in parentheses (e.g. the official name
+# for "Hant" is "Han (Traditional variant)").
+_SCRIPT_DISPLAY_ALIASES = {
```
KennethEnevoldsen, line 56: Does some of these belong in the API rather than in mteb core?

mteb/leaderboard/app.py
```diff
@@ -462,17 +458,30 @@ def get_leaderboard_app(  # noqa: PLR0914
 
     logger.info("Step 1/6: Fetching benchmarks...")
     bench_start = time.time()
-    benchmarks = sorted(
-        mteb.get_benchmarks(display_on_leaderboard=True), key=lambda x: x.name
-    )
+
+    seen: set[str] = set()
```
KennethEnevoldsen, line 462: why is this function changed in this PR?

KennethEnevoldsen (commented):
Let me start with some general comments before I go deeper.


**packaging**: Do we want the API in the mteb package

while I agree that we want the API in this repo (so we can test the two jointly), I would probably consider making it into a second package within the repo called `mteb-api`.

This made be think that we could also factor out models into a packages (`mteb-models`) to isolate our dependency hell outside of mteb. Not for this PR though.

**leaderboard cli**: 

For the current leaderboard CLI how do we want it to behave? should the leaderboard CLI run the current leaderboard? if so we probably want to clean up gradio.

mteb/languages/iso_mappings.py
Samoed: I think we can reuse this in future somewhere. I don't think that mapping from lang codes to langname is api specific

mteb/leaderboard/app.py
Samoed: I changed list of `visible` benchmarks. From GP_BENCHMARKS + RTEB_BENCHMARK to https://github.com/embeddings-benchmark/mteb/blob/30757af9aff61cb727d9a0bb735876b61c738db7/mteb/benchmarks/benchmark.py#L60 benchmark. I changed this to keep previous behavior

mteb/benchmarks/benchmark.py
Samoed: Yes, we can do this

scripts/bench_leaderboard.py
Samoed: I think this could help for future api modifications to not have performance degradation

tests/test_benchmarks/test_get_benchmarks.py
Samoed: I changed list of "visible" benchmarks with new menu. Mteb multilingual is "primary" benchmark, so it's not listed in new object. We can change this, but I'm not sure how to handle primary benchmarks

mteb/api/static/favicon.png
Samoed: I'm not sure. I used this to differentiate during development between api and leaderboard

Samoed:
> while I agree that we want the API in this repo (so we can test the two jointly), I would probably consider making it into a second package within the repo called mteb-api.

I think we can keep as is for now like previous leaderboard. We can refactor it in future

> For the current leaderboard CLI how do we want it to behave? should the leaderboard CLI run the current leaderboard? if so we probably want to clean up gradio.

I think we should keep old leaderboard. I think some people can fork mteb to create their propriete benchmarks and to see scores they could use gradio implmenetation. I don't think that this is possible to run new leaderboard from our package because it's js app

Dockerfile
Samoed: ~6 GB.
- 1 GB source image. 
- 2 GB pip dependencies
- Results 250 mb
- Open graph images 750mb

mteb/benchmarks/benchmark.py
Samoed: Updated to use aggregations directly

mteb/api/static/favicon.png
KennethEnevoldsen: Ahh - probably doesn't matter too much - I would use the current MTEB logo but if you think it is better to differentiate with this then we can keep it

Dockerfile
KennethEnevoldsen: Hmm seems quite big - would be nice to cut down on the pip dependencies if possible (we can put this into a future issue though)

tests/test_benchmarks/test_get_benchmarks.py
KennethEnevoldsen: Maybe it is something that lives UI and we can remove it from here?

scripts/bench_leaderboard.py
KennethEnevoldsen: Maybe we should integrate it more generally as a test then

KennethEnevoldsen:
> I think we should keep old leaderboard. I think some people can fork mteb to create their propriete benchmarks and to see scores they could use gradio implmenetation. I don't think that this is possible to run new leaderboard from our package because it's js app

I am just not sure this is something I want to maintain going forward. It is currently so slow. Maybe we should make a simplified version that (without gradio) just construct the leaderboard table and displays it. It could just write to an HTML file.

This is probably a seperate issue though (not for this PR)

mteb/api/static/favicon.png
Samoed: Yeah, this help to disctinct different pages during development a lot

Dockerfile
Samoed: > Hmm seems quite big - would be nice to cut down on the pip dependencies if possible

The biggest is pytorch, but this is not possible to install mteb without it. I make it reqire less space by installing cpu variant

tests/test_benchmarks/test_get_benchmarks.py
Samoed: Do you mean to remove test?

mteb/api/static/favicon.png
KennethEnevoldsen: sounds like it is a keep. Alternatively create a black and white version of the current MTEB logo and use that.

Dockerfile
KennethEnevoldsen: Hmm yeah this would be a reason for splitting up some of our dependencies. It shouldn't be required to use pytorch for just wrangling the data, but I am not sure this is something we solve in this PR.

mteb/leaderboard/app.py
KennethEnevoldsen: I am not sure I understand

Samoed: In main list of entries consist from https://github.com/embeddings-benchmark/mteb/blob/cfce4f2e38432afa088bd940c169990b2fd2750b/mteb/benchmarks/benchmark.py#L42

but I changed it to https://github.com/embeddings-benchmark/mteb/blob/30757af9aff61cb727d9a0bb735876b61c738db7/mteb/benchmarks/benchmark.py#L60

Which have a bit different layout of benchmarks. To keep previous benchmarks visible I made this change

mteb/benchmarks/benchmark.py
Samoed: Renamed

tests/test_benchmarks/test_get_benchmarks.py
KennethEnevoldsen: I suppose the "display_on_leaderboard" attribute is currently a bit misleading as everything is displayed. Maybe just remove test and deprecate `display_on_leaderboard`?

mteb/leaderboard/app.py
KennethEnevoldsen: I am still note sure I understand. Since all benchmarks are on the leaderboard

Samoed: Yes, but default grid was changed and it's no longer consist from `GP_BENCHMARK_ENTRIES + R_BENCHMARK_ENTRIES` and to keep previous leaderboard as it in `main` I updated it to keep these values

mteb/abstasks/classification.py
```diff
@@ -41,42 +39,10 @@
...
-        ImageStatistics,
-        LabelStatistics,
-        TextStatistics,
-        VideoStatistics,
-    )
 
 logger = logging.getLogger(__name__)
 
 
-class ClassificationDescriptiveStatistics(SplitDescriptiveStatistics):
```
KennethEnevoldsen, line 55: I am unsure why these are moved in this PR?

KennethEnevoldsen: it looks like just a refactor packaged in - so guess it is fine (but quite big)

KennethEnevoldsen (approved):
I think we can merge this in 

I think some create some issues. I suspect we are at a point where a refactor into mteb-core, mteb-models and mteb-api might be a thing we need to look into.

Do we want to document the API somewhere (find to not do it now, but I would do it now, but I suspect this will be a feature bump so would do it in an upcoming PR is we want to).

mteb/abstasks/classification.py
Samoed: Yes, I just moved them into one module to make better typing for API

Samoed:
> Do we want to document the API somewhere

What kind of docs do you want?

KennethEnevoldsen:
> What kind of docs do you want?

I could imagine some people might be interested in fetching data from the API. I am not sure we want to encourage this though

Samoed:
I have `README.md` inside API folder and docstrings will be visible in swagger doc

mteb/api/README.md
```diff
@@ -0,0 +1,61 @@
+# mteb.api
+
+FastAPI surface that powers the leaderboardv2 SvelteKit frontend.
```
KennethEnevoldsen:
```suggestion
FastAPI surface that powers the leaderboard frontend.
```

KennethEnevoldsen (approved):
A perfect missed that one

PR #4763, opened 2026-06-03
benchmarks: add MVEB benchmark suite + leaderboard Video menu

AdnanElAssadi56:
Adds the MVEB (Massive Video Embedding Benchmark) benchmark objects to main so the leaderboard and get_benchmark() can resolve them. The underlying tasks are already on main; this adds only the curated benchmark groupings and their registration.

- benchmarks.py: MVEB (23 tasks), MVEB(text, video) (19), MVEB(video) (9), MVEB(beta, extended) (184, alias MVEB(extended)).
- benchmarks/__init__.py: import + __all__ registration.
- _leaderboard_menu.py: new "Video" group under General Purpose.


mteb/benchmarks/_leaderboard_menu.py
```diff
@@ -59,6 +59,17 @@ class MenuEntry:
...
                 ),
             ),
+            MenuEntry(
+                "Video",
+                mteb.get_benchmarks(
+                    [
+                        "MVEB",
+                        "MVEB(video)",
+                        "MVEB(text, video)",
+                        "MVEB(beta, extended)",
```
KennethEnevoldsen:
```suggestion
```

mteb/benchmarks/_leaderboard_menu.py
```diff
@@ -59,6 +59,17 @@ class MenuEntry:
                     ]
                 ),
             ),
+            MenuEntry(
+                "Video",
+                mteb.get_benchmarks(
+                    [
+                        "MVEB",
+                        "MVEB(video)",
+                        "MVEB(text, video)",
```
KennethEnevoldsen:
```suggestion
                        "MVEB(beta)",
                        "MVEB(video, beta)",
                        "MVEB(text, video, beta)",
```

mteb/benchmarks/_leaderboard_menu.py
KennethEnevoldsen: should not need to be displayed

mteb/benchmarks/_leaderboard_menu.py
```diff
@@ -59,6 +59,17 @@ class MenuEntry:
                     ]
                 ),
             ),
+            MenuEntry(
+                "Video",
+                mteb.get_benchmarks(
+                    [
+                        "MVEB",
```
KennethEnevoldsen: Can we add video and video+text column to this benchmark (see e.g. RTEB on how to customize it)

mteb/benchmarks/_leaderboard_menu.py
AdnanElAssadi: oops, right

mteb/benchmarks/_leaderboard_menu.py
AdnanElAssadi: i think its better if we keep them separate (like we do for maeb). they have different models and tasks; they're not particularly subsets of each other

Samoed (approved):

mteb/benchmarks/_leaderboard_menu.py
KennethEnevoldsen: Hmm I understand that we can't change it now (at least during the review), but I would probably consider changing after the review as it make it much harder to maintain (suddenly we have to run 3 benchmarks instead of 1). 

I suspect this is a by-product of the task selection performed independently by the benchmark? (@isaac-chung)

KennethEnevoldsen:
I think this is good to merge. I will move the comment into the MVEB channel

isaac-chung (approved):

PR #4780, opened 2026-06-07
ci: Update actions

Samoed:
- Update versions of actions, e.g. documentation had warning `Node.js 20 actions are deprecated`
- Remove `free disk space`
- Add cache in some places


.github/workflows/dataset_loading.yml
```diff
@@ -6,42 +6,22 @@ on:
...
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   dataset-loading-check:
     runs-on: ubuntu-latest
     steps:
-    - name: Free disk space
```
KennethEnevoldsen, line 13: why is this no longer needed?

.github/workflows/dataset_loading.yml
```diff
@@ -6,42 +6,22 @@ on:
     paths:
       - "mteb/tasks/**.py"
 
+concurrency:
```
KennethEnevoldsen: do we want to cancel in progress?

KennethEnevoldsen: I typically find that it is good to know where the bug happens and that is hard to see when you cancel-in-progress

.github/workflows/dataset_loading.yml
Samoed: This suddenly appeared on github and was kind like a bug on their actions and seemed like a bug

KennethEnevoldsen: sounds good to simplify then

.github/workflows/dataset_loading.yml
Samoed: We're using `concurrency` in tests already. And I think we could speed up our actions a bit by adding concurrency everywhere

KennethEnevoldsen: Hmm, I would be happy adding this where tests are slow and it makes sense, but adding it everywhere as a default I don't think makes sense

Samoed: Removed this from PR

KennethEnevoldsen (approved):

PR #4785, opened 2026-06-08
Merge multimodal sentence transformers

Samoed:
Merge MultimodalSentenceTransformerEncoderWrapper with SentenceTransformerEncoderWrapper. Also update crossencoder to support multimodality


mteb/models/sentence_transformer_wrapper.py
```diff
@@ -100,7 +165,7 @@ def __init__(
                 overwrites=dict(
                     name=model,
                     revision=revision,
-                    loader=sentence_transformers_loader,
+                    loader=type(self),
```
KennethEnevoldsen, line 163: oh wow this looks so wrong - but works

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -241,137 +331,11 @@ def encode(
 
 
 class SentenceTransformerMultimodalEncoderWrapper(SentenceTransformerEncoderWrapper):
```
KennethEnevoldsen, line 325: always love some backward compat.

but maybe we should deprecate and search and replace where needed?

KennethEnevoldsen (approved):
Only a minor thing on deprecation

mteb/models/sentence_transformer_wrapper.py
Samoed: Added warning and updated usages

PR #4777, opened 2026-06-06
fix: Make project releasablle

Samoed:
Currently on PYPI latest version of mteb is `2.12.30`. In the next commits, we added git dependencies which are actually not valid for publishing on pypi. Noticed by @AdnanElAssadi56. Summary of investigation:

  The actual failure (not the no-op run you originally linked) is:

  - **Run**: https://github.com/embeddings-benchmark/mteb/actions/runs/27025138385
  - **Failing step**: `Publish package distributions to PyPI` (step 7 of the `release` job)
  - Full chain of identical failures going back to 2026-04-30 visible at: https://github.com/embeddings-benchmark/mteb/actions/workflows/release.yml?query=is%3Afailure

  The 400 response body in the log says explicitly:
  > Can't have direct dependency: s2wrapper @ git+https://github.com/bfshi/scaling_on_scales ; extra == "omnivinci". See https://packaging.python.org/specifications/core-metadata for more information.

  ## PyPI side — where the rule lives

  PyPI itself doesn't have a glossy policy page; the rule is enforced in Warehouse (the PyPI server) source code:

  - **Warehouse validation code that emits the error**: [`warehouse/forklift/metadata.py`](https://github.com/pypi/warehouse/blob/82b725b339dd7b2d06a764b44129425f7c497c76/warehouse/forklift/metadata.py#L240-L247) — iterates `requires_dist`, and if `req.url is not None` it appends `InvalidMetadata("Can't have direct dependency: …")`.
  - **Tracking issue / policy discussion** (the canonical place users get pointed to): [pypi/warehouse#7136 — "Allow PEP 508 URL requirements in requires_dist"](https://github.com/pypi/warehouse/issues/7136) and the duplicate [#9404](https://github.com/pypi/warehouse/issues/9404).
  - **Spec the error links to**: [Core metadata specifications — `Requires-Dist`](https://packaging.python.org/en/latest/specifications/core-metadata/#requires-dist-multiple-use) (PyPA packaging docs). Combined with [PEP 440 §Direct references](https://peps.python.org/pep-0440/#direct-references), which notes:
    > Public index servers SHOULD NOT allow the use of direct references in uploaded distributions.

  That "SHOULD NOT" is what Warehouse implements as a hard reject.

  Sources:
  - [mteb release run 27025138385 (failing)](https://github.com/embeddings-benchmark/mteb/actions/runs/27025138385)
[...]

PR #4787, opened 2026-06-09
fix: re-add missing dependencies

KennethEnevoldsen:
removed in #4777 due to missing pypi versions. Reuploaded them to match.

uv sync it back to taking quite a while though with this :/


Samoed:
Should we move these packages to embeddings-benchmark org? Can you update lock?

KennethEnevoldsen:
> Should we move these packages to embeddings-benchmark org? Can you update lock?

Yeah was unsure we we wanted them here (it kinda implied that we maintain them).

---

Will update lock

PR #4789, opened 2026-06-09
fix: Get modalities of models from config

Samoed:
Automatically get modalities from the sentence transformers model config & update model checking for sentence transformers 5.4


KennethEnevoldsen (approved):
Nothing to add here looks good!

PR #4783, opened 2026-06-08
fix: Update lock and remove python limit fo pylate and colbert_engine

Samoed:
Remove `python_full_version < '3.13'` for pylate and colbert_engine. They're supporting higher python verisions. I updated `uv.lock`

Issue #4292, opened 2026-03-23
test leaderboard dockerfile fails consistently


KennethEnevoldsen:
@isaac-chung, @Samoed any suggestion on what the best thing to do is?

<img width="640" height="163" alt="Image" src="https://github.com/user-attachments/assets/2a8cba71-e085-462e-a67b-c8aa9e8e49fa" />

isaac-chung:
The workflow was consistently passing 6 says ago: https://github.com/embeddings-benchmark/mteb/actions/workflows/leaderboard_docker.yml so something within the last 6 days caused this step to timeout at 12 minutes. This timeout was raised from 6min before. So some recent changes have probably made this step slower.

isaac-chung:
Maybe read through the LB start up logs and see where the time breakdown is like?

KennethEnevoldsen:
Looked through the logs but seems like it doesn't get logged consistently oO!
```
✅ Detected leaderboard app startup
Container still running after 30s... (startup: true, init_complete: false)
2026-03-16 14:33:44,486 - PID:12 - __main__ - INFO - === Starting leaderboard app initialization ===
2026-03-16 14:33:44,486 - PID:12 - __main__ - INFO - Step 1/7: Loading all benchmark results...
Container still running after 60s... (startup: true, init_complete: false)
2026-03-16 14:33:44,486 - PID:12 - __main__ - INFO - === Starting leaderboard app initialization ===
2026-03-16 14:33:44,486 - PID:12 - __main__ - INFO - Step 1/7: Loading all benchmark results...
Container still running after 480s... (startup: true, init_complete: false)
2026-03-16 14:41:11,695 - PID:12 - __main__ - INFO - Step 6/7 complete: Created Gradio components in 8.34s
2026-03-16 14:41:11,695 - PID:12 - __main__ - INFO - Step 7/7: Building Gradio interface and callbacks...
Container still running after 510s... (startup: true, init_complete: false)
2026-03-16 14:41:11,695 - PID:12 - __main__ - INFO - Step 7/7: Building Gradio interface and callbacks...
2026-03-16 14:41:16,620 - PID:12 - __main__ - INFO - Step 7/7 complete: Built Gradio interface in 4.93s
Container still running after 540s... (startup: true, init_complete: false)
2026-03-16 14:41:11,695 - PID:12 - __main__ - INFO - Step 7/7: Building Gradio interface and callbacks...
[...]

PR #4790, opened 2026-06-09
fix: Support scikit-learn 1.9 in ZeroShotClassification

### LINKED ISSUE #4784: Sklearn 1.9 is not compatible with ZeroShotClassification
### Describe the bug

Changes that break zeroshot classification that passed labels as strings mostly
> Fix [metrics.accuracy_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html#sklearn.metrics.accuracy_score), [metrics.hamming_loss](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.hamming_loss.html#sklearn.metrics.hamming_loss) [metrics.zero_one_loss](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.zero_one_loss.html#sklearn.metrics.zero_one_loss), [metrics.matthews_corrcoef](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.matthews_corrcoef.html#sklearn.metrics.matthews_corrcoef) and [metrics.confusion_matrix](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html#sklearn.metrics.confusion_matrix) (when labels is not None) now raise an error when y_true is string and y_pred is numeric, for all array-like inputs. Previously, lists and numpy arrays not of object dtype did not raise an error for this mixed input case. The above metrics will also raise an error for [label indicator matrix](https://scikit-learn.org/stable/glossary.html#term-label-indicator-matrix) inputs of inconsistent size, except for [metrics.confusion_matrix](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html#sklearn.metrics.confusion_matrix) which does not accept label indicator matrix inputs. By [Lucy Liu](https://github.com/lucyleeow). [#33086](https://github.com/scikit-learn/scikit-learn/pull/33086) https://scikit-learn.org/stable/whats_new/v1.9.html#sklearn-metrics

### To reproduce

Run any zeroshot 

### Additional information

_No response_

Yes

tejasnaladala:
Fixes #4784

scikit-learn 1.9 raises on mixed-type inputs to classification metrics ([scikit-learn#33086](https://github.com/scikit-learn/scikit-learn/pull/33086)). `AbsTaskZeroShotClassification` hits exactly that case: predictions are always integer indices into the candidate labels, while the dataset label column can contain strings. On `main` with scikit-learn 1.9.0, `tests/test_abstasks/test_predictions.py::test_predictions[task2-expected2]` fails with:

```
ValueError: Mix of label input types (string and number); Got ['label1' 'label2'] and [0 1]
```

The pre-1.9 behavior was arguably worse: `accuracy_score(["label1", "label2"], [0, 1])` compared strings to ints, never matched, and silently reported 0.0 accuracy.

## Changes

- `AbsTaskZeroShotClassification._normalize_labels`: integer labels pass through unchanged, string labels are mapped to their index in `get_candidate_labels()`, and strings that match no candidate raise a `ValueError` describing the label contract. For external task authors this is a behavior change: unmappable string labels now fail loudly on every scikit-learn version instead of silently scoring 0.0.
- Removed the `scikit-learn<1.9.0` stopgap pin from #4783 and regenerated the lock with `uv lock --upgrade-package scikit-learn`. Since scikit-learn 1.9 dropped Python 3.10, the lock forks to 1.9.0 on Python >= 3.11 and stays on 1.7.2 for 3.10, so the CI matrix exercises both sides.
- Mocks: the image zeroshot mock now uses integer labels like the audio and video mocks; the text zeroshot mock keeps string labels (matching its candidates) so the mapping path stays covered in CI.
- New `tests/test_abstasks/test_zeroshot_classification.py`: unit tests for the passthrough, mapping, and error contract, plus an end-to-end regression test that evaluates `mteb/baseline-random-encoder` on the string-label mock and asserts `accuracy == 1.0` (deterministic; the encoder seeds embeddings per input string).

## Leaderboard impact

None. Predictions are untouched and label normalization is an identity for integer labels. All registered zeroshot tasks store labels as integer indices: most derive candidates from `ClassLabel` features, SciMMIR maps strings to ints in `dataset_transform`, and the remaining label columns were verified as int64 via the HF datasets server (`mteb/esc50`, `mteb/SpeechCommandsZeroshotv0.01`, `mteb/urbansound8K`, `mteb/wds_imagenet1k`, and others). The string path only ever fired for the test mocks.

## Verification

- Reproduced the failure on `main` with scikit-learn 1.9.0, then confirmed it passes with this branch (red/green on the same test).
- `tests/test_abstasks` (154 passed), `tests/test_integrations` + prompt validation (314 passed), `tests/test_evaluators`, `tests/test_evaluate.py`, `tests/test_result_cache.py` (87 passed) all green locally on scikit-learn 1.9.0.
- `ruff format --check`, `ruff check`, `typos`, and `mypy mteb` (mypy 2.1.0) clean; `uv lock --check` passes.


Samoed:
Can you run some tasks to verify that scores are matching?

tejasnaladala:
Ran `openai/clip-vit-base-patch32` on two real zeroshot tasks (fresh runs, `cache=None`, same machine):

| Task | `main` (5cfdb8ec) + scikit-learn 1.7.2 | this PR + scikit-learn 1.9.0 |
|---|---|---|
| RenderedSST2 | 0.5848434925864909 | 0.5848434925864909 |
| DTDZeroShot | 0.4186170212765957 | 0.4186170212765957 |

Identical to full float precision, as expected: `_normalize_labels` is an identity passthrough for integer labels, all registered zeroshot tasks have integer label columns, and predictions are untouched.

The RenderedSST2 value also matches the published result for this model in embeddings-benchmark/results ([`results/openai__clip-vit-base-patch32/3d74acf9.../RenderedSST2.json`](https://github.com/embeddings-benchmark/results/blob/main/results/openai__clip-vit-base-patch32/3d74acf9a28c67741b2f4f2ea7635f0aaf6f0268/RenderedSST2.json), accuracy 0.5848434925864909 from mteb 1.14.15) digit for digit. The published DTDZeroShot value (0.41914893617021276) was produced on dataset revision `d2afa97d`, while the task now pins `96726183`, so that one is not the same data; `main` and this branch agree exactly on the current revision.

Repro:

```python
import mteb

model = mteb.get_model_meta("openai/clip-vit-base-patch32")
tasks = mteb.get_tasks(tasks=["RenderedSST2", "DTDZeroShot"])
results = mteb.evaluate(model, tasks, cache=None, co2_tracker=False)
print({r.task_name: r.scores["test"][0]["accuracy"] for r in results})
```

Samoed (approved):

tejasnaladala:
Note on the one red check: test-dockerfile also fails on main itself (pushes at 16:06 and 16:48 UTC today) and on the other open PRs, so it is the pre-existing #4292 flake, not this diff. Everything else is green, including the py3.11-3.14 jobs that run scikit-learn 1.9.0.

KennethEnevoldsen (approved):

PR #4804, opened 2026-06-12
ci: Setup hf trusted publishing

Samoed:
Setup trusted publishing from https://huggingface.co/changelog/trusted-publishers

### DIFF
--- a/.github/workflows/leaderboard_refresh.yaml
+++ b/.github/workflows/leaderboard_refresh.yaml
@@ -8,6 +8,10 @@ on:
 jobs:
   rebuild:
     runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen (approved):
This seems fine to me - should I delete the secrets.hf_token?

Samoed:
I think we're using it for overcome rate limiting in dataset loading https://github.com/embeddings-benchmark/mteb/blob/cfce4f2e38432afa088bd940c169990b2fd2750b/.github/workflows/dataset_loading.yml#L27-L30

HF has feature for enterprise organizations to have service acconts to solve this problem, but we're not enterprize org https://huggingface.co/changelog/service-accounts

PR #4808, opened 2026-06-14
Update Querit model implementation: 4B version of Querit-Reranker newly open-sourced

moshesbeta:
Add new model revision of Querit/Querit-4B to match the results of new results on leaderboard


mteb/models/model_implementations/querit_models.py
```diff
@@ -222,18 +222,18 @@ def compute_mask_content_cls(block_types: list[int]) -> torch.Tensor:
     loader_kwargs={
         "fp_options": "bfloat16",
     },
-    name="Querit/Querit",
+    name="Querit/Querit-4B",
```
Samoed, line 253: This seems as new model rather than updated implementation. Can you use add ne `ModelMeta`?

moshesbeta: We trained Querit-4B on a larger backbone model, while the wrapper logic remains unchanged. We will update `model_meta.json` in the subsequent results PR.
Is this acceptable?

Samoed: I think would be better to add new `ModelMeta` for "Querit/Querit-4B"

mteb/models/model_implementations/querit_models.py
```diff
@@ -217,7 +217,7 @@ def compute_mask_content_cls(block_types: list[int]) -> torch.Tensor:
     "T2Reranking",  # https://huggingface.co/datasets/THUIR/T2Ranking & The corpus and queries that overlap with mteb/T2Reranking have been removed.
 }
 
-model_meta = ModelMeta(
+Querit_Reranker_A0.6B = ModelMeta(
```
Samoed:
```suggestion
Querit_Reranker_A0_6B = ModelMeta(
```

Samoed: Can you fill then `n_active_parameters_override`?

mteb/models/model_implementations/querit_models.py
moshesbeta: We have modified. Please review~

mteb/models/model_implementations/querit_models.py
moshesbeta: Sure, it's added. Please review~

mteb/models/model_implementations/querit_models.py
```diff
@@ -230,6 +230,7 @@ def compute_mask_content_cls(block_types: list[int]) -> torch.Tensor:
     release_date="2026-05-14",
     n_parameters=4919636992,
     n_embedding_parameters=131907584,
+    n_active_parameters_override=433833858
```
Samoed:
```suggestion
    n_active_parameters_override=433_833_858,
```

Samoed (approved):

PR #4809, opened 2026-06-14
Rename `number_texts_intersect_with_train` to `samples_in_train`

Samoed:
Previosly `number_texts_intersect_with_train` was renamed to `samples_in_train` , but metadata of tasks wasn't updated


Samoed:
I'll add tese to ignore for now, since they're already merged 

```
AssertionError: AVEDatasetClassification (test) has an overlap between train and test (samples_in_train=6)
AVEDatasetVideoClassification (test) has an overlap between train and test (samples_in_train=6)
HMDB51Classification (test) has an overlap between train and test (samples_in_train=1)
AudioSet (test) has an overlap between train and test (samples_in_train=9)
```
CC @KennethEnevoldsen

PR #4815, opened 2026-06-15
tests: Fix flackyness of test_timing_stack test

### LINKED ISSUE #4810: Tests `test_timing_stack` is flacky
Sometimes test `test_timing_stack` is failing with:
```
FAILED tests/test_timing.py::test_timing_stack - assert 0.0 > 0.0
```
CC @ayush1298

ayush1298:
closes #4810

I guess these solution should work, though locally test was always passing

### DIFF
--- a/tests/test_timing.py
+++ b/tests/test_timing.py
@@ -1,27 +1,23 @@
 from __future__ import annotations
 
 import time
-from unittest.mock import patch
 
 import pytest
 
 from mteb.timing import TimingStack
 
 
-@patch("mteb.timing.time.monotonic")
-def test_timing_stack(mock_monotonic):
-    mock_monotonic.side_effect = [10.0, 10.0, 10.01, 10.01, 10.02]
-
+def test_timing_stack():
     timer1 = TimingStack()
     assert timer1.phases == []
     assert timer1._start_time is None
     assert timer1.plot() == "No timing phases recorded."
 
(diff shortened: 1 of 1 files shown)


Samoed (approved):

ayush1298:
@Samoed Can we merge these?

tests/test_timing.py
KennethEnevoldsen, line 35:
```suggestion
    assert timer1.phases[0]["end"] >= 0.0
```

KennethEnevoldsen (commented):
I think we can solve it without the patching by simply using ">="

tests/test_timing.py
ayush1298: It should not be equal to 0, so if we add `>=`, then it will always pass, even though it should fail if its equal to 0. These patches will make it just more deterministic.

KennethEnevoldsen: I am not a big fan of this fix. It assumes something about the implementation's internals that has nothing to do with the test.

I would either use a dependency injection, but the simpler alternative seems to just bump the time.sleep

ayush1298: okay, just bumping `time.sleep` to `0.1` from `0.01`

ayush1298:
I have simplified it by changing the bump. I guess it's good now. 
@KennethEnevoldsen @Samoed

Samoed (approved):

Issue #2873, opened 2025-07-02
Integrate `SparseEncoder` model from SentenceTransformers `v5`


arthurbr11:
Hello!

## Context
We released Sentence Transformers v5 yesterday, introducing a new `SparseEncoder` model category. These are a subclass of `SentenceTransformer` that output sparse vectors instead of dense embeddings.

## Feature Description
Add support for `SparseEncoder` models in MTEB. I tested the current version and found one main blocking issue in the `Sentence TransformerWrapper` to make it compatible : https://github.com/embeddings-benchmark/mteb/blob/f27648baec2086389b162b2b94026d2ed6325416/mteb/models/sentence_transformer_wrapper.py#L114

with the .numpy (applied on a sparse tensor it's breaking and not wanted anyway)

## Questions
1. Is there a specific reason for the `.numpy()` conversion? Can we make it conditional for sparse models?
2. For similarity computation, sparse models have an optimized `similarity` function that should be use this instead of MTEB's standard similarity functions, is it gonna be always the case ? Because if dense way of computing the similarity is used it will broke or be really long.
3. Since no sparse indexes exist in MTEB, encoding by chunk would be the best option to make sure the similarity compute is reasonable but is it plan to handle other indexes ? 

This would probably be part of a bigger refactor of the Sentence Transformers handling with https://github.com/embeddings-benchmark/mteb/issues/2871 and other possible feature.

Would love to hear your thoughts thist. 

cc @tomaarsen  

Arthur Bresnu

Samoed:
1. What embedding type `SparseEncoder` expecting?
2. We've changed this in `v2` branch to use models similarity function if available, so I think this is resolved in `v2` https://github.com/embeddings-benchmark/mteb/blob/ed69e60c17177ab1c66358e99022e13ad4f7c184/mteb/models/abs_encoder.py#L35-L53
3. Yes, we have plans for new indexes https://github.com/embeddings-benchmark/mteb/issues/1593 https://github.com/embeddings-benchmark/mteb/issues/2728

arthurbr11:
1. `SparseEncoder` models output a `torch.sparse_coo` by default.

2. That's great!

3. Nice! That means having one optimized for sparse vectors would be beneficial, it would be faster than the current model similarity computation.

KennethEnevoldsen:
[...]

PR #4828, opened 2026-06-18
model: Add SPLADE-Code sparse retrieval models (naver/splade-code-06B, splade-code-8B)

### LINKED ISSUE #4551: Add model: splade-code

https://huggingface.co/collections/naver/splade-code

### Arxiv link

_No response_

### Description of the model

_No response_

fzowl:
Adds the naver/splade-code sparse retrieval models (splade-code-06B and splade-code-8B) to the model registry. These are SPLADE models adapted from Qwen3-0.6B and Qwen3-8B that produce vocabulary-sized sparse embeddings, loaded via the existing SparseEncoderWrapper. Verified with a smoke test on AILAStatutes. Fixes #4551.

### DIFF
--- a/mteb/models/model_implementations/splade_code_models.py
+++ b/mteb/models/model_implementations/splade_code_models.py
@@ -1,0 +1,67 @@
+from __future__ import annotations
+
+from mteb.models.model_implementations.opensearch_neural_sparse_models import (
+    SparseEncoderWrapper,
+)
+from mteb.models.model_meta import ModelMeta
+
+splade_code_languages = [
+    "python-Code",
+    "javascript-Code",
+    "go-Code",
+    "ruby-Code",
+    "java-Code",
+    "php-Code",
+]
+
+splade_code_06b = ModelMeta(
+    name="naver/splade-code-06B",
+    model_type=["sparse"],
+    languages=splade_code_languages,
+    open_weights=True,
+    revision="e53a0b8bd312d83955598a392dc826b3fc4028f7",
+    release_date="2026-02-24",
+    n_parameters=596049920,
+    n_embedding_parameters=155582464,
+    memory_usage_mb=1137,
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/splade_code_models.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import annotations
+
+from mteb.models.model_implementations.opensearch_neural_sparse_models import (
+    SparseEncoderWrapper,
```
Samoed, line 4: I don't think that reusing this wrapper a good idea. We need to implement a proper sparsewrapper firstly, but previsoly we had some problems with it https://github.com/embeddings-benchmark/mteb/issues/2873. I'll take a look if it possible to run it now

fzowl:
Looked into this — `SparseEncoderWrapper` actually instantiates the native `sentence_transformers.SparseEncoder` internally, so it *is* the native path already. The auto-detect route in `_detect_loader` routes SparseEncoder models to the dense `SentenceTransformerEncoderWrapper` (commented as "Not support for now"), which hits the `.numpy()`-on-sparse breakage from #2873. So the explicit `loader=SparseEncoderWrapper` is the correct and only working approach until the v2 native integration from #2873 lands.

Verified end-to-end on GPU: both models load, encode, and retrieve successfully (AILAStatutes ndcg@10 ≈ 0.32).

@Samoed How to proceed with this one?

Samoed:
I'll make updates in this pr with proper wrapper implementation

fzowl:
Sounds good, happy to rebase on top of whatever you push.

fzowl:
@Samoed Can you please take a look?

Samoed:
Will add in https://github.com/embeddings-benchmark/mteb/pull/5087

PR #4846, opened 2026-06-22
fix: drop duplicate train documents in SyntheticText2SQL corpus

### LINKED ISSUE #4626: Potential Bug: SyntheticText2SQL evaluation penalizes exact text duplicates as false positives
### Describe the bug

In mteb/SyntheticText2SQL, the test retrieval corpus pools both train and test partitions. However, over 94% of the test queries and documents have exact text duplicates in the train partition under different IDs.
Because the qrels only map the query to the official test document ID, retrieving the identical train document is incorrectly scored as a false positive. This artificially tanks Top-1 / NDCG metrics.

If a model maps both to the exact same vector, standard tie-breaking (e.g., argsort) often favors the lower-index train ID (c91481). Because c91481 isn't in the qrels, it registers as a false positive despite being the correct SQL answer.

Suggest Fix: To directly delete the test data from the training set to avoid this issue and data leakage.
Might worth running a full check for other datasets as well, especially those whose corpus includes the training data.

### To reproduce

Concrete Example
```
Train ID: c91481 | Test ID: c100010
Doc Text: "What is the average sales quantity and sales amount for each drug..." (Identical for both)

MTEB qrel: q100010 -> c100010 ONLY.
```
Check example on:
https://huggingface.co/datasets/mteb/synthetic-text2sql/viewer/queries?row=100009
https://huggingface.co/datasets/mteb/synthetic-text2sql/viewer/corpus?row=100009
https://huggingface.co/datasets/mteb/synthetic-text2sql/viewer/queries?row=91480
https://huggingface.co/datasets/mteb/synthetic-text2sql/viewer/corpus?row=91480

### Additional information

_No response_

Yes

fzowl:
Fixes #4626.

- fix: drop duplicate train documents in SyntheticText2SQL corpus

### DIFF
--- a/mteb/tasks/retrieval/code/synthetic_text2_sql_retrieval.py
+++ b/mteb/tasks/retrieval/code/synthetic_text2_sql_retrieval.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from mteb.abstasks.retrieval import AbsTaskRetrieval
 from mteb.abstasks.task_metadata import TaskMetadata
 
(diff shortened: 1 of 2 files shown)


mteb/tasks/retrieval/code/synthetic_text2_sql_retrieval.py
```diff
@@ -36,3 +38,43 @@ class SyntheticText2SQLRetrieval(AbsTaskRetrieval):
 }
 """,
     )
+
+    def dataset_transform(self, num_proc: int | None = None, **kwargs) -> None:
```
Samoed, line 54: You need to create a v2 version of dataset rather than modifing existing

fzowl (commented):
Fixed in c274de18. Moved the fix to a new `SyntheticText2SQL.v2` task class; original kept intact with `superseded_by`.

PR #4852, opened 2026-06-24
Add overview benchmark

### LINKED ISSUE #2038: Add an overview benchmark
It would be great to have a leaderboard that is an overview of current benchmarks. This should be reasonably easy to add using the `AbstaskAggregate`.

The overview could include:

```
MTEB(eng)
MTEB(Multilingual)
MTEB(Europe)
MTEB(Indic)
MTEB(Code)
MTEB(Law)
MTEB(Medical)
FollowIR
LongEmbed
MTEB(Image) / MIEB
```

And simply be implemented as a new benchmark called `Overview`.

Debated here:
https://github.com/embeddings-benchmark/mteb/issues/2026#issuecomment-2651599525

ayush1298:
closes #2038

Add overview benchmark to LB

### DIFF
--- a/mteb/abstasks/aggregate_task_metadata.py
+++ b/mteb/abstasks/aggregate_task_metadata.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 
 from pydantic import ConfigDict, Field, model_validator
 
+from mteb.benchmarks.get_benchmark import get_benchmark
 from mteb.types import (
     Languages,
 )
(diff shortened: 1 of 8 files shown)


Samoed:
I'm unsure if we need this with new leaderboard. WDYT @KennethEnevoldsen?

KennethEnevoldsen:
I would def. still consider it, but I am not sure this is the way to add it. I think it requires a bit more development.

I have currently worked on the compare view a bit more (when adding openness) and here I think we e.g. could add a default view that covers multilingual, English, multimodality if relevant, length generalization using LongEmbed etc.

I think that is the more appropriate place to start but it requires more experimentation

ayush1298:
@KennethEnevoldsen Do you mean it should no longer be a separate overview benchmark that the original issue demands?

KennethEnevoldsen:
Yeah I am actually not sure what is the best way to display it is (probably require some experimentation)

ayush1298:
> Yeah I am actually not sure what is the best way to display it is (probably require some experimentation)

cool, thanks for looking into it

PR #4855, opened 2026-06-24
perf(leaderboard): lazy load benchmark plots

SmileLikeYe:
## Context

When switching benchmarks in the MTEB leaderboard, the app was eagerly generating all three plot figures: performance by model size, performance over time, and radar chart. These plots already have tab-select callbacks, so users only need them when they actually open the corresponding plot tabs.

## Why This PR

Benchmark switching should feel fast and should only update the data needed for the currently visible leaderboard view. Eagerly computing hidden plots adds unnecessary work to a common interaction, especially for users who only want to compare summary tables across benchmarks.

This PR reduces benchmark-switch latency by deferring plot generation until the user selects a plot tab.

## Changes

- Removed eager plot generation from the benchmark selection callback.
- Kept the existing initial plot rendering behavior, so the default page still shows plots as before.
- Kept plot tab callbacks wired to the latest `summary_data`, so plots are generated on demand with current benchmark data.
- Updated benchmark-change timing logs to remove misleading plot timing fields.
- Added focused tests proving benchmark selection does not compute plot callables, while tab-style plot generation still works.

### DIFF
--- a/mteb/leaderboard/app.py
+++ b/mteb/leaderboard/app.py
@@ -35,6 +35,10 @@ from mteb.leaderboard.figures import (
     _performance_size_plot,
     _radar_chart,
 )
+from mteb.leaderboard.lazy_plots import (
+    BenchmarkSelectOutputValues,
+    build_benchmark_select_outputs,
+)
 from mteb.leaderboard.table import (
     apply_per_language_styling_from_benchmark,
     apply_per_task_styling_from_benchmark,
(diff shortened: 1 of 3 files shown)


Samoed (commented):
You can just move out plot loading from this function, but it made like this to sepedup loading of benchmark pages. Ovearall, I'm not sure if we want to make updates to the old leaderboard since we have rewrote it

SmileLikeYe:
Thanks for the context. That makes sense.

My understanding is that the current eager plot generation makes plot tabs faster once opened, but it also adds extra work to benchmark switching even when users only look at the summary table.

If the old leaderboard is no longer actively maintained because of the rewrite, I’m happy to close this PR. Otherwise, I think this is a small targeted change that improves the common benchmark-switching path while keeping the existing tab-level plot loading behavior.

KennethEnevoldsen:
@SmileLikeYe do you utilize the old leaderboard locally? We don't plan to maintain the gradio leaderboard going forward, but if you wish we could factor it out into a seperate package that the community could maintain if there is interest for it?

related PRs:
#4854
#4853

KennethEnevoldsen:
will assume it is not used and then I will close this one

PR #4860, opened 2026-06-28
fix: add retrieval character n-gram overlap descriptive stats

### LINKED ISSUE #4466: Extend descriptive statistics for retrieval to include keyword overlap

Many retrieval datasets are very easy due to a high degree of keyword overlap between the query and the document. This could be estimated either by comparing it with bm25 scores, but maybe simpler in the desc stats using something like:

```py
def query_token_recall(query: str, doc: str) -> float:
    q = set(query.lower().split())
    d = set(doc.lower().split())
    return len(q & d) / len(q) if q else 0.0
```

wgu9:
## Summary

Adds text corpus overlap statistics for retrieval and reranking tasks.

- Normalizes text with Unicode NFKC and case folding, removes punctuation, symbols, and whitespace, then measures query 4-gram coverage across the full text corpus.
- Computes the statistic only when both query and corpus contain text.
- Prefixes aggregate query and document IDs by split and subset so duplicate local IDs do not overwrite each other.
- Checks the generated values in the task quality tests.

Closes #4466

## Validation

- Ruff format and lint
- mypy on the changed source and test files
- statistics, task metadata, and task quality tests: 68 passed, 25 skipped
- git diff --check


wgu9:
Update after the CI rerun on 74809dff:

- The deterministic metadata failure from the prior run is fixed; the torchcodec-enabled mock MMEB path now has the expected `text_relevance_overlap_statistics`.
- Current remaining failures are limited to `test (ubuntu-latest, 3.10)` and `test (ubuntu-latest, 3.13)`, both in the existing `test_benchmark_datasets` coverage for `Core17InstructionRetrieval`. The 3.10 job hit a Hugging Face 504 while resolving `mteb/Core17InstructionRetrieval`, and the 3.13 job fell back to a cached dataset state where the requested builder config was unavailable.
- The same commit passes the other test matrix jobs (3.11, 3.12, 3.14, Windows 3.10), lowest-deps test, lint, typecheck, docs build, and leaderboard build.

I tried rerunning the failed jobs from the CLI, but GitHub requires repository admin rights for that rerun.

mteb/types/statistics.py
```diff
@@ -228,6 +228,25 @@ class RelevantDocsStatistics(TypedDict):
...
+        average_query_token_recall: Average fraction of unique query tokens found
+            in the relevant document text.
+        max_query_token_recall: Maximum fraction of unique query tokens found in
+            the relevant document text.
+    """
+
+    num_pairs: int
+    min_query_token_recall: float
+    average_query_token_recall: float
+    max_query_token_recall: float
```
Samoed: Shoudn't this be `ovelap` instead of `recall`?

mteb/abstasks/_statistics_calculation.py
```diff
@@ -416,6 +417,44 @@ def calculate_relevant_docs_statistics(
...
+    """Calculate query-token recall against positive relevant documents."""
+
+    def _resolve_text(text_by_id: Mapping[str, str], item_id: str) -> str:
+        if item_id in text_by_id:
+            return text_by_id[item_id]
+        # Multilingual aggregate retrieval stats prefix ids as
+        # ``{split}_{hf_subset}_{id}`` to keep qrels unique across subsets.
+        _, _, unprefixed_id = item_id.partition("_")
+        _, _, unprefixed_id = unprefixed_id.partition("_")
+        return text_by_id[unprefixed_id]
```
Samoed: I don't think you need to split ids

Samoed (changes requested):
Ovearall I like the idea, but I'm not sure if we want to have a modality specific statistics in metadata

wgu9:
Thanks, addressed the two inline points in `8f15f97f`:

- renamed the fields from `*_query_token_recall` to `*_query_token_overlap`
- removed the split/subset id fallback and now use the qrels query/doc ids directly

Validation:
- `python -m ruff check mteb/types/statistics.py mteb/abstasks/_statistics_calculation.py tests/mock_tasks.py`
- `python -m compileall -q mteb/abstasks mteb/types tests/mock_tasks.py`
- `git diff --check`

I also understand the broader concern about whether retrieval-specific descriptive stats belong in metadata at all. If you prefer not to carry this field in task metadata, I can narrow or drop this part rather than pushing the design point.

wgu9:
Follow-up in `11f526f7`: the aggregate `test` job exposed that multilingual overall qrels use prefixed ids such as `test_eng_q1` while the aggregate query/corpus datasets still expose local ids like `q1`/`d1`.

I kept the overlap helper as direct id lookup, as requested, and moved the prefixed-id text map construction into the multilingual aggregate caller in `retrieval.py`. Local validation available in this environment:

- `python -m ruff check mteb/types/statistics.py mteb/abstasks/_statistics_calculation.py mteb/abstasks/retrieval.py tests/mock_tasks.py`
- `python -m compileall -q mteb/abstasks mteb/types tests/mock_tasks.py`
- `git diff --check`

I could not rerun the exact `uv run --group test pytest tests/test_abstasks/test_task_metadata.py -q` locally because this machine has `uv 0.9.18` and the project requires `uv >=0.10.0`; GitHub CI should now verify the targeted path.

wgu9:
One more follow-up in `0909295d`: the new CI run showed `ruff format --check` wanted the multi-line condition in `retrieval.py`. I ran `python -m ruff format mteb/abstasks/retrieval.py`, then revalidated locally with:

- `python -m ruff check mteb/types/statistics.py mteb/abstasks/_statistics_calculation.py mteb/abstasks/retrieval.py tests/mock_tasks.py`
- `python -m ruff format --check mteb/abstasks/retrieval.py`
- `python -m compileall -q mteb/abstasks mteb/types tests/mock_tasks.py`
- `git diff --check`

wgu9:
Follow-up pushed in d379c853: the version-matrix jobs cover non-text retrieval tasks, and those hit the previous direct-map construction before the text/text modality guard. That meant query ids were zipped with an empty text list under strict=True.

I moved query/corpus text map construction inside the text/text guard, so non-text retrieval tasks now leave text_relevance_overlap_statistics unset as intended.

Local validation:
- python -m ruff check mteb/types/statistics.py mteb/abstasks/_statistics_calculation.py mteb/abstasks/retrieval.py tests/mock_tasks.py
- python -m ruff format --check mteb/abstasks/retrieval.py
- python -m compileall -q mteb/abstasks mteb/types tests/mock_tasks.py
- git diff --check

wgu9:
The follow-up CI is now green across build, lint, typecheck, leaderboard, aggregate test, and the OS/Python version matrix. Ready for re-review when you have a chance.

wgu9:
Friendly re-review ping when you have a chance. I addressed the two inline points, kept the retrieval-specific statistic gated to text/text tasks, fixed the multilingual aggregate ID mapping exposed by CI, and the follow-up CI is green across lint, typecheck, aggregate tests, and the OS/Python matrix. I am happy to revise further if the broader metadata-design concern remains.

mteb/abstasks/_statistics_calculation.py
```diff
@@ -416,6 +417,34 @@ def calculate_relevant_docs_statistics(
...
 
+def calculate_text_relevance_overlap_statistics(
+    relevant_docs: Mapping[str, Mapping[str, int]],
+    queries: Mapping[str, str],
+    corpus: Mapping[str, str],
+) -> TextRelevanceOverlapStatistics | None:
+    """Calculate query-token overlap against positive relevant documents."""
+    overlaps: list[float] = []
+    for query_id, docs in relevant_docs.items():
+        query_tokens = set(queries[query_id].lower().split())
```
KennethEnevoldsen: this will work only for certain languages (clarifying it above I think would be enough)

mteb/types/statistics.py
```diff
@@ -228,6 +228,25 @@ class RelevantDocsStatistics(TypedDict):
...
+    """Token-overlap statistics for query and relevant document text pairs.
+
+    Attributes:
+        num_pairs: Number of positive query-document relevance pairs considered.
+        min_query_token_overlap: Minimum fraction of unique query tokens found in
+            the relevant document text.
+        average_query_token_overlap: Average fraction of unique query tokens found
+            in the relevant document text.
+        max_query_token_overlap: Maximum fraction of unique query tokens found in
+            the relevant document text.
```
KennethEnevoldsen: I would denote that we define tokens are lowercasing + white space split. This doesn't really tell us anything about non-latin languages.

mteb/abstasks/_statistics_calculation.py
```diff
@@ -416,6 +417,34 @@ def calculate_relevant_docs_statistics(
...
+        if not query_tokens:
+            continue
+        for doc_id, score in docs.items():
+            if score == 0:
+                continue
+            doc_tokens = set(corpus[doc_id].lower().split())
```
KennethEnevoldsen: I would extract the into `_tokenize()` just to make it easier for us to refactor in the future.

I would imagine stripping punctuation would be reasonable here as well.

KennethEnevoldsen (commented):
Sorry for the wait @wgu9, let me jump in here. 

I only have a few minor concerns here. Mostly around language support. In the [BM25](https://docs.mteb.org/get_started/advanced_usage/bm25/), we implemented a subword tokenizer from qwen and show that it performs well for both english and chinese (though not ideal for either).

Would require us to download a tokenizer, but it would make this much more generalizable.

@Samoed do you have any other concerns here?

Samoed:
I don't want to create modality specific statics and not sure that splitting by spaces is good for all cases

KennethEnevoldsen:
> I don't want to create modality specific statics 

But why not? Isn't this a specific stat that is important only for text? 

> splitting by spaces is good for all cases

agree (also noted above, let discuss that one there)

Samoed:
Maybe we can calculate ngrams isntead, e.g. 3-5.

mteb/types/statistics.py
wgu9: Agreed. The updated schema now names these fields min/average/max_query_character_ngram_overlap; there is no recall naming left.

mteb/abstasks/_statistics_calculation.py
wgu9: Agreed. The statistics helper uses the qrels query/document IDs directly. The only prefix handling is now in the multilingual aggregate caller, where the aggregate qrels themselves use split/subset-prefixed IDs.

mteb/abstasks/_statistics_calculation.py
wgu9: Addressed by replacing whitespace tokens with normalized character n-grams of lengths 3-5. I added coverage for a no-space Chinese query as well as Unicode normalization.

mteb/types/statistics.py
wgu9: Updated the metric and its documentation: text is NFKC-normalized and case-folded, punctuation/symbols/whitespace are removed, and overlap is computed over unique query character 3-5-grams. This no longer depends on Latin-style word boundaries.

mteb/abstasks/_statistics_calculation.py
wgu9: Done. Normalization and 3-5-gram extraction now live in _character_ngrams(), and punctuation, symbols, and whitespace are stripped before extraction.

wgu9:
Implemented the latest 3-5 character n-gram direction in 53673591.

Root cause: the previous lower().split() definition assumed whitespace word boundaries, so it was not comparable across no-space scripts and handled punctuation inconsistently. The updated implementation:

- Unicode NFKC-normalizes and case-folds text
- removes punctuation, symbols, and whitespace
- extracts unique character n-grams of lengths 3-5
- reports the fraction of query n-grams present in each positive relevant document
- remains gated to query/document pairs that both contain text
- keeps direct qrels ID lookup; multilingual prefix mapping stays in the aggregate caller

I also merged the latest main branch and migrated the expected mock statistics from the removed tests/mock_tasks.py into the new mteb/mocks retrieval/reranking modules. All five inline review threads have replies and are resolved.

Validation:
- ruff format/check: passed
- mypy on changed source files: passed
- targeted statistics + metadata tests: 42 passed, 45 skipped
- full non-network test suite: 6556 passed, 240 skipped

@Samoed @KennethEnevoldsen, could you please re-review when CI finishes?

wgu9:
CI is now fully green on 53673591: docs, leaderboard, lint, typecheck, lowest dependencies, Ubuntu Python 3.10-3.14, and Windows Python 3.10 all passed. The PR is mergeable and all review threads are resolved; re-review is requested.

KennethEnevoldsen (commented):
This is good with me - though I am quite unsure if 3-5 is a meaningful signal for overlap. @Samoed maybe you know better?

Samoed:
I think this is fine. Most common words will be 3-5 letters. If we select more, then this will be leass meanengfull for CJK languages

KennethEnevoldsen:
check - are we then happy to merge here?

Samoed:
Can you also add this to test from https://github.com/embeddings-benchmark/mteb/pull/5090?

wgu9:
Added this in ef2eccad after merging the latest main. The quality test now checks that the overlap values are finite, stay within 0 to 1, follow min <= average <= max, and that num_pairs is positive and does not exceed the number of positive qrels.

I kept this to data integrity checks since the project does not currently define a threshold for what counts as high overlap.

Validation:
- tests/test_tasks/test_task_quality.py: 2 passed
- tests/test_abstasks/test_statistics_calculation.py: 4 passed
- ruff format and check: passed
- mypy on the updated quality test: passed
- git diff check: passed

The fresh CI run is also green across docs, leaderboard, lint, typecheck, lowest dependencies, Ubuntu 3.10 through 3.14, and Windows 3.10.

Thanks for pointing me to the new quality test.

mteb/abstasks/_statistics_calculation.py
```diff
@@ -416,6 +418,49 @@ def calculate_relevant_docs_statistics(
...
 
 
+def calculate_text_relevance_overlap_statistics(
+    relevant_docs: Mapping[str, Mapping[str, int]],
+    queries: Mapping[str, str],
+    corpus: Mapping[str, str],
+) -> TextRelevanceOverlapStatistics | None:
+    """Calculate query character n-gram overlap against relevant documents."""
+    overlaps: list[float] = []
+    for query_id, docs in relevant_docs.items():
```
Samoed: I think would be better to check across full collection of documents and quries, not only for relevant documents

mteb/mocks/mock_tasks/retrieval.py
```diff
@@ -469,6 +524,9 @@ class MockMultilingualRetrievalTask(AbsTaskRetrieval):
                         "max_relevant_docs_per_query": 1,
                         "unique_relevant_docs": 2,
                     },
+                    "text_relevance_overlap_statistics": (
```
Samoed: Better to not reuse from other statistics, but write as plain dict

wgu9:
Audit follow-up in da6afd3a: overlap calculation now skips qrels pairs whose query or corpus IDs are absent from the loaded text maps. Previously the optional statistic raised `KeyError` and prevented descriptive stats generation, which also hid the existing qrels integrity checks.

Added regression coverage for missing query and document IDs.

Validation:
- statistics and task-quality tests: 8 passed
- Ruff 0.16.0 format and check: passed
- `git diff --check`: passed

wgu9:
One more audit hardening is in 593891d4. The overlap aggregate now keeps only count, sum, minimum, and maximum instead of one float per relevant pair, so memory stays constant as qrels grow.

Validation:

- 8 focused tests passed
- Ruff 0.16 format and lint passed
- git diff check passed

Samoed (commented):
I forgot to submit review

mteb/types/statistics.py
```diff
@@ -528,6 +552,7 @@ class RetrievalDescriptiveStatistics(SplitDescriptiveStatistics):
     queries_video_statistics: VideoStatistics | None
 
     relevant_docs_statistics: RelevantDocsStatistics
+    text_corpus_overlap_statistics: NotRequired[TextCorpusOverlapStatistics]
```
Samoed:
```suggestion
    text_corpus_overlap_statistics:TextCorpusOverlapStatistics | None
```

mteb/abstasks/retrieval.py
```diff
@@ -573,13 +574,64 @@ def _calculate_descriptive_statistics_from_split(  # noqa: PLR0914
         )
 
         relevant_docs_statistics = calculate_relevant_docs_statistics(relevant_docs)
+        text_corpus_overlap_statistics = None
+        if "text" in queries_col_inputs and "text" in corpus_col_inputs:
+            if compute_overall:
+                query_text_by_id: dict[str, str] = {}
+                corpus_text_by_id: dict[str, str] = {}
+                for subset_name in self.metadata.eval_langs:
+                    split_data = self.dataset[subset_name][split]
```
Samoed: In `compute_overall` we already combine all queries and corpus subset and splits, but we don't update ids of queries and docuemnts to also include split and subset names, but update like this relevant docs

https://github.com/wgu9/mteb/blob/497611374e19fe0763ea152c0edae9645f70a403/mteb/abstasks/retrieval.py#L537-L550

mteb/abstasks/retrieval.py
```diff
@@ -595,6 +647,9 @@ def _calculate_descriptive_statistics_from_split(  # noqa: PLR0914
             relevant_docs_statistics=relevant_docs_statistics,
             top_ranked_statistics=top_ranked_statistics,
         )
+        if text_corpus_overlap_statistics is not None:
+            stats["text_corpus_overlap_statistics"] = text_corpus_overlap_statistics
+        return stats
```
Samoed: Better to set this valaue in `if "text" in queries_col_inputs and "text" in corpus_col_inputs:` otherwise `None`

mteb/abstasks/_statistics_calculation.py
```diff
@@ -416,6 +418,70 @@ def calculate_relevant_docs_statistics(
...
+
+    return TextCorpusOverlapStatistics(
+        num_queries=num_queries,
+        min_query_character_ngram_overlap=min_overlap,
+        average_query_character_ngram_overlap=total_overlap / num_queries,
+        max_query_character_ngram_overlap=max_overlap,
+    )
+
+
+def _character_ngrams(text: str, min_n: int = 3, max_n: int = 5) -> set[str]:
```
Samoed: I think ranged approach for ngrams will reqire a lot of memory for big datasets and better to calculate only one ngram set, e.g. n=4

mteb/abstasks/_statistics_calculation.py
```diff
@@ -416,6 +418,70 @@ def calculate_relevant_docs_statistics(
...
+    all_query_ngrams: set[str] = set()
+    for query_text in queries.values():
+        all_query_ngrams.update(_character_ngrams(query_text))
+
+    if not all_query_ngrams:
+        return None
+
+    covered_query_ngrams: set[str] = set()
+    has_corpus_ngrams = False
+    for document_text in corpus.values():
```
Samoed: For better visibility
```suggestion
    for document_text in tqdm(corpus.values()):
```

mteb/abstasks/_statistics_calculation.py
wgu9: Thanks, updated. The statistic now checks each query against the full text corpus, not only relevant documents.

mteb/mocks/mock_tasks/retrieval.py
wgu9: Thanks, updated. The mock expectation is now a plain dict.

mteb/types/statistics.py
wgu9: Updated in a844642c. The field is now `TextCorpusOverlapStatistics | None`.

mteb/abstasks/retrieval.py
wgu9: Updated in a844642c. Query and corpus IDs now get the same split and subset prefix as qrels before aggregation. The overlap code uses those combined datasets directly.

mteb/abstasks/retrieval.py
wgu9: Updated. Text/text tasks compute the value; other retrieval tasks set it to `None`.

mteb/abstasks/_statistics_calculation.py
wgu9: Updated to one 4-gram set. I refreshed the tests and mock statistics.

mteb/abstasks/_statistics_calculation.py
wgu9: Added `tqdm` around the corpus scan with a short description.

wgu9:
@Samoed I merged the latest main and resolved the two conflicts in a51132c4. The resolution keeps the new qrel ID integrity checks from main and the full-corpus overlap statistic from this PR.

Checks run:
- 74 relevant tests passed, 25 optional media tests skipped
- repository lint check passed
- mypy passed for all 5 changed source files
- diff check passed

Could you take another look when you have a chance?

Samoed:
Can you submit statistics for some tasks? I'm not sure that we have good setup now

wgu9:
I ran the current branch (`a51132c4`) on several retrieval tasks using `_calculate_descriptive_statistics_from_split`, including three scripts from the same multilingual task. These are the full-corpus character 4-gram overlap statistics:

| Task / subset | Queries used | Documents | Min | Mean | Max |
|---|---:|---:|---:|---:|---:|
| SciFact | 300 | 5,183 | 0.8718 | 0.9888 | 1.0000 |
| NFCorpus | 321 / 323 | 3,633 | 0.0000 | 0.9760 | 1.0000 |
| ArguAna | 1,406 | 8,674 | 1.0000 | 1.0000 | 1.0000 |
| QuoraRetrieval | 9,999 / 10,000 | 522,931 | 0.7333 | 0.9990 | 1.0000 |
| MultilingualNanoSciFact / English | 50 | 2,919 | 0.9600 | 0.9912 | 1.0000 |
| MultilingualNanoSciFact / Japanese | 50 | 2,919 | 0.2222 | 0.7111 | 0.9565 |
| MultilingualNanoSciFact / Arabic | 50 | 2,919 | 0.8276 | 0.9604 | 1.0000 |

The reduced query counts for NFCorpus and Quora are queries that have no character 4-grams after normalization.

My read is that the present full-corpus setup saturates for the English datasets, especially as corpus size grows: ArguAna is exactly 1.0 for every query and Quora averages 0.9990. It does retain a noticeably different signal for Japanese in this sample, with Arabic between Japanese and English. So the calculation is behaving as implemented, but the English results support the concern that full-corpus coverage may be too close to 1.0 to be very discriminative on typical retrieval corpora.

Samoed:
Yes. Maybe we need to increase gap then, and handle CJK seperatly

PR #4867, opened 2026-06-30
Add structure-aware corpus-adaptive hybrid retriever

thu-nmrc:
## Summary

Adds `mteb/structure-aware-corpus-adaptive-hybrid`, a corpus-adaptive hybrid retrieval model for academic retrieval tasks.

Following review feedback, the implementation has been reworked to build on the existing `HybridSearch` wrapper instead of using a standalone dense search loop. The current structure is:

- `CorpusAdaptiveHybridSearch(HybridSearch)` for indexing/search orchestration and adaptive fusion;
- `sentence-transformers/all-MiniLM-L6-v2` loaded through `mteb.get_model(...)` for the dense component;
- a small `FieldWeightedBM25Search` component for the title/text BM25F-style lexical signal that is not currently covered by `mteb/baseline-bm25s`;
- corpus-level routing over existing fusion patterns, including RRF and residual structure fusion.

The paired results submission is here: https://github.com/embeddings-benchmark/results/pull/580

## Local Validation

Local checks run before updating this PR:

- `python -m py_compile mteb/models/model_implementations/structure_aware_hybrid.py`
- `uvx ruff check mteb/models/model_implementations/structure_aware_hybrid.py`
- `git diff --check`

## Result Summary From Local HF-MTEB Runs

Revision: `1`

Macro nDCG@10 over four full retrieval tasks: `0.42089022568272455`.

Task-level nDCG@10:

- SciFact: `0.7085019697253156`
- NFCorpus: `0.3482959056916576`
- FiQA2018: `0.4063591243963125`
- SCIDOCS: `0.22040390291761284`

### DIFF
--- a/mteb/models/model_implementations/structure_aware_hybrid.py
+++ b/mteb/models/model_implementations/structure_aware_hybrid.py
@@ -1,0 +1,482 @@
+from __future__ import annotations
+
+import math
+import re
+from collections import Counter, defaultdict
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+from mteb._create_dataloaders import _combine_queries_with_instruction_text
+from mteb.models.hybrid_wrappers import HybridSearch
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping, Sequence
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.models.models_protocols import MTEBModels
+    from mteb.types import (
+        CorpusDatasetType,
+        EncodeKwargs,
+        QueryDatasetType,
+        RetrievalOutputType,
+        TopRankedDocumentsType,
+    )
+
+
+TOKEN_RE = re.compile(r"[A-Za-z0-9]+")
+
+
+def _tokenize(text: str) -> list[str]:
+    return TOKEN_RE.findall((text or "").lower())
+
+
+def _mean_len(texts) -> float:
+    lengths = [len(_tokenize(text)) for text in texts]
+    return float(np.mean(lengths)) if lengths else 0.0
+
+
+def _minmax(scores: Mapping[str, float]) -> dict[str, float]:
+    if not scores:
+        return {}
+    lo = min(scores.values())
+    hi = max(scores.values())
+    if hi <= lo:
+        return dict.fromkeys(scores, 1.0)
+    return {key: (val - lo) / (hi - lo) for key, val in scores.items()}
+
+
+def _rank_map(scores: Mapping[str, float]) -> dict[str, int]:
+    return {
+        did: rank
+        for rank, (did, _) in enumerate(
+            sorted(scores.items(), key=lambda kv: (-kv[1], kv[0])), 1
+        )
+    }
+
+
+def _rrf(
+    results_list: Sequence[Mapping[str, float]],
+    weights: Sequence[float],
+    *,
+    top_k: int,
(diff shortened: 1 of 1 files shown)


Samoed (changes requested):
Can you explain a bit more why you need this? We already have hybrid wrappers that can combine models together https://github.com/embeddings-benchmark/mteb/pull/4706. Will this solve your problem?

thu-nmrc:
Thanks for the pointer. The existing `HybridSearch` wrapper solves the static-combination part, but the result submission this PR is paired with depends on two behaviors that are not directly expressible with the current wrapper API:

1. corpus-statistics-gated strategy selection: the retriever switches between RRF, dense-title RRF, lexical-consensus residual, and depth-regularized structure fusion based on average document length, title length, and query length;
2. field-aware lexical structure: the submitted runs use both text-only BM25 and a BM25F-style title/text index, then fuse the structure residual differently per corpus.

So this is not intended to replace `HybridSearch`; it is a corpus-adaptive hybrid retriever that uses the same underlying idea but needs task/corpus metadata before choosing the fusion policy.

If the preferred direction is to avoid a standalone implementation file, I can rework this into a smaller extension around `HybridSearch`, e.g. adding metadata-aware fusion/strategy selection hooks and then registering this model as a thin loader using those hooks. I opened it this way first to keep the submitted results reproducible while the wrapper API discussion is resolved.

thu-nmrc:
I reworked the implementation to follow the `HybridSearch` direction:

- removed the standalone dense retrieval loop;
- load the dense component via `mteb.get_model("sentence-transformers/all-MiniLM-L6-v2")`;
- made the main retriever subclass `HybridSearch` and use its indexing/search orchestration;
- kept only a small field-weighted BM25 component because the submitted results rely on a title/text BM25F-style signal in addition to text-only BM25.

Local checks pass: `py_compile`, `uvx ruff check`, and `git diff --check`.

thu-nmrc:
I also configured a local Python 3.10 MTEB test environment with `uv sync --group test` and caught/fixed a loader bug: the outer model revision `1` was being forwarded to the MiniLM child model. The loader now strips `revision` before calling `mteb.get_model("sentence-transformers/all-MiniLM-L6-v2")`.

Local checks now completed in the full environment:

- `py_compile` + `ruff`: passed
- `mteb.get_model_meta("mteb/structure-aware-corpus-adaptive-hybrid")`: passed, registry contains the model
- `pytest tests/test_models/test_model_meta.py::test_check_model_name_and_revision -q`: 747 passed
- `mteb.get_model("mteb/structure-aware-corpus-adaptive-hybrid", device="cpu")`: passed
- micro index/search smoke with 3 docs and 2 queries: passed

mteb/models/model_implementations/structure_aware_hybrid.py
```diff
@@ -0,0 +1,482 @@
...
+            if 11.0 <= avg_query_len < 40.0:
+                return self._structure_residual_hybrid(
+                    bm25, bm25f, dense, alpha=0.40, residual_weight=0.01, top_k=top_k
+                )
+            return self._structure_residual_hybrid(
+                bm25, bm25f, dense, alpha=0.15, residual_weight=0.02, top_k=top_k
+            )
+        return self._structure_residual_hybrid(
+            bm25, bm25f, dense, alpha=0.40, residual_weight=0.01, top_k=top_k
+        )
```
Samoed, line 349: I don't think we want to choose models based on document statistics, because this prevents users from choosing parameters themselves, and it is not clear why these parameters were selected. Also this is very language specific and most pobably also dataset specific

Issue #4868, opened 2026-06-30
Retrieval Systems [Agentic MTEB]


AdnanElAssadi:
# Answer-mode retrieval

## The idea

mteb scores a retriever by the quality of its ranking. But most retrieval today happens inside a larger system: the documents are retrieved so that an LLM can answer a question. Two systems with similar rankings can produce very different answers, and some systems (agents, closed-book models) do not produce a ranking at all.

This work lets mteb evaluate that full path: given a question and a fixed corpus, produce an answer, and grade the answer. Ranking evals and answer evals run on the same datasets, so retrievers, RAG variants, and agents stay directly comparable.

## How it works

- A task provides a corpus, questions, and reference answers.
- A system is anything with two methods: `index(corpus)` once, then `answer(query_id, question)` per question.
- A judge grades each answer against the reference (exact match by default, LLM judge where the benchmark officially grades that way).
- `mteb.evaluate(system, task)` runs it like any other task. Accuracy is the main score, LLM cost is reported next to it.

Three built-in systems bracket every result: closed-book (the LLM alone, the floor), RAG (retrieve then answer), and oracle (answer from the gold documents, the ceiling). The gap between RAG and oracle shows how much accuracy the retriever is costing you.

## Status

1. **#5130** (in review): LLM retrieval strategies as `SearchProtocol` wrappers (query rewrite, HyDE, rerank, multi-query, multi-hop), plus the LLM client interface with cost accounting. These improve or wrap any existing retriever and are scored with normal ranking metrics.
2. **#5131** (in review, stacked on #5130): the answer evaluation above, with `BrowseCompPlusQA` as the first task.

## Next steps

3. More QA datasets: HotpotQA, MuSiQue, MultiHopRAG, and a variant for tasks where each question has its own context (LongBench-v2, OOLONG).
4. Agents as answer systems: an in-process search agent (tool loop over search and get_document), then containerized coding agents (claude-code, codex). The sandbox runner for those is an open choice: Harbor, or a thinner docker wrapper.
5. Re-upload the BCP mirror with the reference answers included, so the task loads from one source.

A working prototype of every paradigm (search agents, iterative RAG, full-context, RLM, containerized agents, 6 tasks) is on the [`agentic` branch](https://github.com/embeddings-benchmark/mteb/tree/agentic) for reference.

## References

[...]

PR #4869, opened 2026-06-30
agentic: add answer-mode retrieval core (contract, evaluator, baselines)

AdnanElAssadi56:
First PR of the `mteb.agentic` subsystem (see #4868 ).

- Contract: `AnswerSystem`, `CorpusHandle`, `ChatModel`, `AnswerResult`, `Usage`
- `AnswerEvaluator`: runs a system per query, scores accuracy / cost / latency
- Judges: `ExactMatchJudge`, `LLMJudge`
- Baselines: `ClosedBookSystem` (floor), `OracleContextSystem` (ceiling)
- `from_mteb_retrieval` data adapter
- Tests (`tests/test_agentic.py`)

### DIFF
--- a/mteb/agentic/corpus.py
+++ b/mteb/agentic/corpus.py
@@ -1,0 +1,14 @@
+"""Reference CorpusHandle implementations."""
+
+from __future__ import annotations
+
+
+class InMemoryCorpus:
+    """Read-only corpus backed by a dict of documents."""
+
+    def __init__(self, documents: dict[str, dict[str, str]]) -> None:
+        self._docs = documents
+
+    def get(self, doc_id: str) -> dict[str, str]:
+        """Return one document with its id."""
+        return {"id": doc_id, **self._docs[doc_id]}
--- a/mteb/agentic/evaluator.py
+++ b/mteb/agentic/evaluator.py
@@ -1,0 +1,70 @@
+"""Runs one AnswerSystem over a question set and scores it on three axes."""
+
+from __future__ import annotations
+
+import time
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+from mteb.agentic.metrics import aggregate
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+    from mteb.agentic.interface import AnswerResult, AnswerSystem, CorpusHandle
+    from mteb.agentic.metrics import AggregateScores, Judge
+
+
+@dataclass
+class AnswerEvaluationResult:
+    """Aggregate scores plus a per-question record."""
+
+    scores: AggregateScores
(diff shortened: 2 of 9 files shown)


mteb/agentic/evaluator.py
```diff
@@ -0,0 +1,70 @@
...
+        questions: Mapping[str, str],
+        references: Mapping[str, str],
+        corpus: CorpusHandle,
+        judge: Judge,
+    ) -> None:
+        # questions and references are keyed by query id.
+        self.questions = questions
+        self.references = references
+        self.corpus = corpus
+        self.judge = judge
```
Samoed, line 40: I don't think we need an llm judge, because we can just compare ids. With llm answer and cheking it I think this is too much LLM checking

mteb/agentic/evaluator.py
```diff
@@ -0,0 +1,70 @@
...
+
+@dataclass
+class AnswerEvaluationResult:
+    """Aggregate scores plus a per-question record."""
+
+    scores: AggregateScores
+    per_question: list[dict[str, Any]]
+
+
+class AnswerEvaluator:
```
Samoed, line 26: I think we could use some tools, but I don't have, but I don't think we will have much

mteb/agentic/corpus.py
```diff
@@ -0,0 +1,14 @@
+"""Reference CorpusHandle implementations."""
+
+from __future__ import annotations
+
+
+class InMemoryCorpus:
```
Samoed, line 6: I think we can reuse our existing indexes for that

mteb/agentic/interface.py
```diff
@@ -0,0 +1,73 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+# Provider-agnostic chat message.
+Message = dict[str, str]
```
Samoed, line 12: I think we would need to add multimodal support in future

Samoed (commented):
Currently a bit hard to judge without full pipeline

KennethEnevoldsen:
@AdnanElAssadi56 seems like this has gotten stale - I suspect people are working on other things atm (MOEB). Do you plan to work on this in parallel?

AdnanElAssadi:
> @AdnanElAssadi56 seems like this has gotten stale - I suspect people are working on other things atm (MOEB). Do you plan to work on this in parallel?

Yup, but a PR like this would miss much of the context. I think working on a separate branch might be better considering most of this will not collide with original mteb. Closing this for now.

PR #4873, opened 2026-07-01
Cap Seq Length for Some Models

AdnanElAssadi56:
I got oom for some models while evaluating MTEB(eng, v2) on an h100. I think these new caps match the authors' evals.


mteb/models/model_implementations/e5_omni_models.py
```diff
@@ -43,8 +43,10 @@ def __init__(
                     "max_pixels": 64 * 28 * 28,
                     "do_sample_frames": False,
                 },
+                "text": {"truncation": True, "max_length": 512},
             }
         )
+        self.model.max_seq_length = 512
```
Samoed: I don't think this is needed for new sentence transformers
```suggestion
```

Samoed (approved):

PR #4875, opened 2026-07-02
feat: auto-install optional model dependencies via MTEB_AUTO_INSTALL_EXTRAS

### LINKED ISSUE #4349: Look into auto-installing optional dependencies

Currently when running a model handling the dependencies is often a bit of a hassle when running a lot of models. It would be great if it was possible to just say "install whatever the requires dependencies are". This could probably be done using `uv`. 

It would probably require us to do something like this:
```
model_meta.optional_dependency
```

which would then enable us to run something like:
```
uv pip install "mteb[model_meta.optional_dependency]"
```

It might even allow us to simplify the dependencies greatly by simply removing then from mteb pyproject.toml and just have tests that ensures that things work. This is just to outline the idea, but I think we should take it in steps. Starting with the simplest first steps.

fzowl:
Fixes #4349.

- feat: auto-install optional model dependencies via MTEB_AUTO_INSTALL_EXTRAS


Samoed (approved):
Great feature!

KennethEnevoldsen (approved):

KennethEnevoldsen:
Added what's new but otherwise good

Issue #4366, opened 2026-04-10
Add an openness score to the documentation/leaderboard


KennethEnevoldsen:
### Description of the feature

We collect a lot of data on openness but do not display it in the table. I would love to display it more prominently. However, displaying each score independently might be hard. However we could display an score like 7 out of 9.

E.g. taking inspiration from the osai index:
- https://osai-index.eu/
- https://codeberg.org/AI-technology-assessment/main-database/src/branch/preview/readme.md

They include:

- Base model data: not collected
- End model data: collected
- base model weight: not collected
- End model weight: collected
- Training code: Currently collected
- Training code documented: Not collected
- Hardware architecture documented in paper: Not collected
- Preprint: somewhat collected
- review paper: somewhat collected
- model card: currently required
- datasheet: not collected
- osi compatible license: collected

---

A good place to start is probably adding it to the documentation. Currently, the documention look like:

<img width="777" height="366" alt="Image" src="https://github.com/user-attachments/assets/35b40ebd-10ce-4d21-a02b-c18999bec640" />

We could add a checklist like:

| OS Score | open weight | open license | open code | open data | Paper |
| -- | -- | -- | -- | -- | -- |
| 4/10  | Yes | Yes | Yes | Yes | Yes |

but could probably spent some more time to add something like:

<img width="278" height="73" alt="Image" src="https://github.com/user-attachments/assets/68fae49e-5313-4b01-9b8f-476a0d2bfb0b" />

which on hover would give:

<img width="227" height="166" alt="Image" src="https://github.com/user-attachments/assets/77b2878e-1b8c-4524-9afc-d2f481db1115" />

That would take up less space in the docs while still displaying the info prominently

PR #4877, opened 2026-07-02
feat: Add openness score to model metadata and documentation

fzowl:
addresses #4366, but not on the leaderboard site (yet)

- Add openness score to model metadata and documentation


KennethEnevoldsen:
Refactored to display as:

Screenshot 2026-07-05 at 19 09 44 

@Samoed, from my understanding of the fastapi imp., this should also be available there so adding it the leaderboard shouldn't be too much of a hassle.

KennethEnevoldsen (approved):

PR #4881, opened 2026-07-02
fix(JinaV5OmniWrapper): text-task prefix + channels-last video frames

florian-hoenicke:
## Summary

Two independent bugs in `JinaV5OmniWrapper` that explain the non-reproducible / near-zero scores reported in embeddings-benchmark/results#513 and embeddings-benchmark/results#546.

### 1. Missing Query/Document prefix on non-retrieval text tasks

`JinaV5OmniWrapper` was omitting `"Query: "` / `"Document: "` prefix for all non-retrieval tasks (STS, Classification, Clustering, PairClassification, Summarization), while `JinaV5TextWrapper` always sends the prefix. An incorrect code comment said _"Non-retrieval adapters were trained without the Query/Document prefix"_ — empirical evaluation proves the opposite.

**Measured score delta (text-nano adapters, A8000 GPU, mteb 2.10.8):**

| Task | with prefix | no prefix (old bug) | Δ |
|------|-------------|---------------------|---|
| STS12 | 0.8523 | 0.3845 | **−0.47** |
| STSBenchmark | 0.9443 | 0.4785 | **−0.47** |
| BIOSSES | 0.8670 | 0.6152 | **−0.26** |
| Banking77Classification | 0.9018 | 0.2952 | **−0.61** |
| ArXivHierarchicalClusteringP2P | 0.6454 | 0.6281 | −0.02 |
| SprintDuplicateQuestions | 0.9516 | 0.1806 | **−0.77** |
| SummEvalSummarization.v2 | 0.2986 | 0.1885 | −0.13 |

New logic: audio/video non-retrieval tasks keep the empty prefix; text and image inputs always use the prefix.

### 2. Video frames fed channels-first, silently embedded as text

torchcodec's `VideoDecoder` yields **(T, C, H, W)** uint8 frame batches (default `dimension_order="NCHW"`), and mteb's `FramesCollator.resample_video` passes them through as-is. The model's HF remote code detects video only for channels-last `(T, H, W, 3|4)` arrays; the mismatched tensor fell through to the text fallback and was embedded as `str(array)` — every video collapsed to near-identical garbage embeddings. This is why `Shot2Story20KAT2VRetrieval` scored 0.0009 in results#546 while audio-carrying tasks (AVEDataset*) still scored reasonably: the audio side was processed correctly and carried the signal.

**GPU verification:** the same video encoded channels-first vs channels-last gives cosine **0.14** between embeddings; the channels-first embedding has cosine 0.90 with the embedding of the array's literal string repr.

Fix: convert `(T, C, H, W)` tensors to channels-last numpy before passing to the model.

Unit tests added/updated for both behaviors.

### DIFF
--- a/mteb/models/model_implementations/jina_models.py
+++ b/mteb/models/model_implementations/jina_models.py
@@ -841,15 +841,16 @@ class JinaV5OmniWrapper(SentenceTransformerEncoderWrapper):
                 task_metadata.simplified_task_type
             )
         task = jina_task_name or "retrieval"
-        # Non-retrieval adapters were trained without the Query/Document prefix.
-        if task == "retrieval":
+        # All text and image adapters use Query/Document prefix.
+        # Audio/video non-retrieval adapters do not use the prefix.
+        if (has_video or has_audio) and task != "retrieval":
+            prompt = ""
+        else:
             prompt = (
                 "Query: "
                 if prompt_type and prompt_type == PromptType.query
                 else "Document: "
             )
-        else:
-            prompt = ""
 
         logger.info(
             f"Using prompt=`{prompt}` and task={task} for task={task_metadata.name} prompt_type={prompt_type}"
--- a/mteb/models/model_implementations/jina_models.py
+++ b/mteb/models/model_implementations/jina_models.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar, Unpack
 
 import numpy as np
 import torch
@@ -21,7 +21,7 @@ if TYPE_CHECKING:
     from torch.utils.data import DataLoader
 
     from mteb.abstasks.task_metadata import TaskMetadata
-    from mteb.types import Array, BatchedInput
+    from mteb.types import Array, BatchedInput, EncodeKwargs
 
 logger = logging.getLogger(__name__)
 
@@ -805,8 +805,8 @@ def _video_frames_to_channels_last(video: Any) -> Any:
     if (
         isinstance(video, torch.Tensor)
         and video.ndim == 4
-        and video.shape[1] in (3, 4)
-        and video.shape[-1] not in (3, 4)
+        and video.shape[1] in {3, 4}
+        and video.shape[-1] not in {3, 4}
     ):
         return video.permute(0, 2, 3, 1).contiguous().cpu().numpy()
     return video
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/jina_models.py
```diff
@@ -1066,6 +1074,12 @@ def encode(
     release_date="2026-04-01",
     n_parameters=985984512,
     n_embedding_parameters=98503680,
+    n_active_parameters_by_modality={
```
Samoed: Can you remove `n_active_parameters_by_modality`? We didn't add this field

Samoed:
Can you also try to investigate video part https://github.com/embeddings-benchmark/results/pull/546#issuecomment-4526652386? Might be there can be prompt difference too?

mteb/models/model_implementations/jina_models.py
florian-hoenicke: Done — removed. That field was work-in-progress from my local branch and slipped into the commit by accident. The PR now contains only the prefix fix and the test updates.

florian-hoenicke:
@Samoed Investigated the video part — you were right that something was off, but it's not a prompt difference. It's a tensor-layout bug, and it's now fixed in this PR.

**Root cause of the near-zero video retrieval scores in results#546:**

1. torchcodec's `VideoDecoder` returns frame batches as **(T, C, H, W)** uint8 (default `dimension_order="NCHW"`), and mteb's `FramesCollator.resample_video` passes that tensor through unchanged.
2. The jina omni model's HF remote code (`custom_st.py`) detects video only for channels-last `(T, H, W, 3|4)` arrays — the check `x.shape[-1] in (3, 4)` fails when `shape[-1]` is the frame width.
3. The unrecognized tensor falls through to the final fallback `("text", str(x))` — every video document was embedded as the text `"[[[[255 254 ..."` (the truncated numpy repr). All videos collapse to near-identical garbage embeddings, hence `Shot2Story20KAT2VRetrieval` nDCG = 0.0009.

**GPU verification** (omni-nano, A8000): encoding the same video as a channels-first tensor vs the documented channels-last layout gives cosine **0.14** between the two embeddings, and the channels-first embedding has cosine 0.90 with the embedding of the array's literal string repr — confirming the stringify path.

This also explains the pattern you noticed: `AVEDatasetVAPairClassification` and `AVEDatasetAudioVideoClustering` scored fine because those inputs also carry audio, which was processed correctly and carried the signal, while the video part was garbage on both sides.

Fix (second commit in this PR): the wrapper now converts `(T, C, H, W)` frame tensors to channels-last numpy before passing them to the model, with unit tests. Prompts were not the issue for video — retrieval tasks get `"Query: "`/`"Document: "` in both the old and new code.

After this PR merges, the video results in results#546 should be re-run.

Samoed:
Can you resolve your merge conflict? CC @AdnanElAssadi56 for video part

florian-hoenicke:
@Samoed Rebased onto main and resolved the conflict (the `JinaV5OmniWrapper` base class had changed to `SentenceTransformerEncoderWrapper` upstream). Both fixes are intact, wrapper unit tests pass.

Samoed (approved):

AdnanElAssadi:
@florian-hoenicke I'll be rerunning the 23 task MVEB subset, and i'll see if i can do the full set after. do you have bandwidth to do some runs because I've reran jina several times now?

PR #4897, opened 2026-07-05
fix: correctly isolate splits when aggregating scores in aggregated tasks

### LINKED ISSUE #4896: Bug: Aggregated tasks incorrectly average scores across all splits simultaneously
### Describe the bug

Currently, when evaluating an `AbsTaskAggregate` over multiple splits (e.g. `test`, `dev`), the scoring logic inside `mteb/abstasks/aggregated_task.py` loops over `self.metadata.eval_splits` but passes the entire list of splits (`splits=self.metadata.eval_splits`) down to `task_res._get_score_fast`. 

As a result, `_get_score_fast` calculates the combined average across **all** splits simultaneously, and assigns that same merged value to both the `test` score and the `dev` score, instead of calculating their true, independent values for that specific split.

### To reproduce

You can verify this bug by mocking a simple `AbsTaskAggregate` composed of a single task that has different scores for `test` and `dev` splits, and then running it through `task_results_to_scores`. 

Instead of isolating the scores per split, it averages them together and assigns the identical merged value to both.

```python
import numpy as np
from mteb.abstasks.aggregated_task import AbsTaskAggregate
from mteb.results.task_result import TaskResult
from mteb.abstasks.aggregate_task_metadata import AggregateTaskMetadata
from mteb.abstasks.task_metadata import TaskMetadata
from mteb.abstasks.abstask import AbsTask

# Mock a simple task metadata
mock_task_meta = TaskMetadata(
    name="MockTask1",
    description="Mock",
    dataset={"path": "mock", "revision": "1.0"},
    reference=None,
    type="Classification",
    category="t2t",
    modalities=["text"],
    eval_splits=["test", "dev"],
    eval_langs=["eng-Latn"],
    main_score="accuracy",
    date=("2024-01-01", "2024-12-31"),
    domains=["News"],
    task_subtypes=["Topic classification"],
    license="mit",
    annotations_creators="expert-annotated",
    dialect=[],
    sample_creation="found",
    bibtex_citation=""
)

class MockTask(AbsTask):
    metadata = mock_task_meta
    def _evaluate_subset(self, *args, **kwargs): pass
    def _calculate_descriptive_statistics_from_split(self, *args, **kwargs): pass

mock_task = MockTask()

# Create a mock TaskResult with DIFFERENT scores for test and dev
mock_task_result = TaskResult(
    dataset_revision="1.0",
    task_name="MockTask1",
    mteb_version="1.0",
    scores={
        "test": [
            {"hf_subset": "default", "languages": ["eng-Latn"], "main_score": 0.8, "accuracy": 0.8, "mteb_version": "1.0"}
        ],
        "dev": [
            {"hf_subset": "default", "languages": ["eng-Latn"], "main_score": 0.5, "accuracy": 0.5, "mteb_version": "1.0"}
        ]
    },
    evaluation_time=1.0,
)

# Mock aggregate task
agg_meta = AggregateTaskMetadata(
    name="AggTask",
    description="Mock agg",
    reference=None,
    dataset={"path": "mock", "revision": "1.0"},
    type="Classification",
    category="t2t",
    modalities=["text"],
    eval_splits=["test", "dev"],
    eval_langs={"default": ["eng-Latn"]},
    main_score="accuracy",
    date=("2024-01-01", "2024-12-31"),
    domains=["News"],
    task_subtypes=["Topic classification"],
    license="mit",
    annotations_creators="expert-annotated",
    dialect=[],
    sample_creation="found",
    bibtex_citation="",
    tasks=[mock_task]
)

class MockAggTask(AbsTaskAggregate):
    metadata = agg_meta

agg_task = MockAggTask()
scores = agg_task.task_results_to_scores([mock_task_result])

for split, subset_scores in scores.items():
    print(f"Split '{split}': {subset_scores['default']['main_score']}")
```

```
Output (Buggy):
Split 'test': 0.65
Split 'dev': 0.65
```

### Additional information

This happens because `AbsTaskAggregate.task_results_to_scores()` incorrectly passes `splits=self.metadata.eval_splits` into the `task_res._get_score_fast()` call on line 72, which computes the mean across *all* of those splits simultaneously. 

The fix is to simply pass the current isolated split variable from the loop: `splits=[split]`.

Yes

pranaysb:
Related Issue: Closes #4896 

**Description:**
This PR fixes a critical logic bug in `AbsTaskAggregate.task_results_to_scores`. 

Previously, when iterating through `self.metadata.eval_splits`, the code incorrectly passed the entire list of splits (`splits=self.metadata.eval_splits`) to `task_res._get_score_fast`. This caused the scores to be averaged across **all** splits simultaneously, resulting in mathematically identical merged values being assigned to every split (e.g., the `test` and `dev` scores would both be overwritten with the merged average of the two).

This fix simply passes the current isolated split variable from the loop: `splits=[split]`. This ensures that the aggregated scores remain properly isolated per split.

---

### Test Results / Proof of Fix

**1. Regression test added and passing:**
Added `test_task_results_to_scores_isolates_splits` to `tests/test_abstasks/test_aggregate_task.py`, which mocks an `AbsTaskAggregate` with distinct `test` (0.8) and `dev` (0.5) scores and asserts they remain isolated.

```bash
$ uv run pytest tests/test_abstasks/test_aggregate_task.py
======================== 2 passed, 4 warnings in 1.03s =========================
```

Before fix — Split 'test': 0.65, Split 'dev': 0.65
After fix  — Split 'test': 0.8,  Split 'dev': 0.5


tests/test_abstasks/test_aggregate_task.py
```diff
@@ -13,3 +13,100 @@ def test_is_aggregate_property_correct():
...
+        task_subtypes=["Topic classification"],
+        license="mit",
+        annotations_creators="expert-annotated",
+        dialect=[],
+        sample_creation="found",
+        bibtex_citation="",
+        tasks=[mock_task],
+    )
+
+    class MockAggTask(AbsTaskAggregate):
```
Samoed: We have mock aggregated tasks https://github.com/embeddings-benchmark/mteb/blob/786b31c6a0a8737cc99394d1a9d9de03857cb8aa/tests/mock_tasks.py#L2926 can you use it instead?

tests/test_abstasks/test_aggregate_task.py
```diff
@@ -13,3 +13,51 @@ def test_is_aggregate_property_correct():
...
+                }
+            ],
+        },
+        evaluation_time=1.0,
+    )
+
+    scores = agg_task.task_results_to_scores([mock_task_result])
+
+    assert scores["test"]["default"]["main_score"] == 0.8
+    assert scores["dev"]["default"]["main_score"] == 0.5
```
Samoed: Shoudn't be there `None` because results only for one tasks existis?

Samoed: I meant currently aggregation works a bit incorrectly. Can you add check & test when results provided in `task_results_to_scores` should match requested from metadata and use only them

pranaysb: Ah, I see — thanks for clarifying. That's a different issue from what I addressed in the last commit (which just gave the mock more complete data). You're right that task_results_to_scores currently has no check that the task_results passed in actually correspond to the tasks in self.metadata.tasks — it just averages whatever it's given.
I just pushed a validation step that checks each result's task name against the expected tasks from metadata, uses only the matching ones, and I added a test that covers the mismatch case (e.g. extra/unrelated or missing task results).
(One quick note: if the provided results are missing data for any of the required sub-tasks, the code now raises a hard ValueError rather than silently computing an artificially skewed partial average. I assumed this is the safest choice for a benchmark tool. If you'd prefer it to just log a warning and return np.nan instead, let me know and I can easily soften it!)

Samoed: I think would be better to to return None instead

pranaysb: Tried implementing this literally as None, but it actually breaks at runtime, not just in typechecking: combine_task_results passes whatever task_results_to_scores returns straight into TaskResult.from_task_results, which immediately does scores.items() on it. A literal None there raises AttributeError the moment any aggregate task hits this path.
To get the same practical effect you're after (don't crash, don't record a misleading score) without breaking that, I'd propose returning an empty dict {} instead, plus a log message noting the task was skipped due to missing results. Downstream code handles an empty dict safely — the task just ends up with no scores recorded, same as you'd get with None, but without the crash.
Let me know if that works for you, or if you would suggest something else  — happy to adjust before I push.

Samoed: I meant do `"main_score": None`

Samoed (approved):
Thanks for the fix!

pranaysb:
> Thanks for the fix!

Thanks for the quick review and approval, @Samoed!

I'll stand by for the final review from the other maintainer. Let me know if anything else is needed!

KennethEnevoldsen (approved):
did a minor fix of imports otherwise good

PR #4779, opened 2026-06-07
WIP leaderboard integation with experiments

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/1211
Frontend https://github.com/embeddings-benchmark/leaderboardv3/pull/4

PR #4900, opened 2026-07-06
Leaderboard integation with experiments

### LINKED ISSUE #1211: Figure out an approach for adding experiments to the leaderboard

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/1211
Frontend https://github.com/embeddings-benchmark/leaderboard-frontend/pull/4
Reapplying changes from https://github.com/embeddings-benchmark/mteb/pull/4779


mteb/results/benchmark_results.py
```diff
@@ -458,10 +473,19 @@ def _build_pre_agg_df(self, include_model_revision: bool) -> pd.DataFrame | None
...
         for model_result in bench_results:
             mn = model_result.model_name
             mr = model_result.model_revision
+            mm = model_result.model_meta
+            exp_kwargs = (
+                dict(mm.experiment_kwargs)
+                if mm is not None and mm.experiment_kwargs
+                else None
+            )
+            meta_dict = mm.to_dict() if mm is not None and exp_kwargs else None
```
Samoed, line 488: I added model meta to the results to load it for experiments for setups like jina v4 when based on experiment paramters model meta changes a bit (e.g. from dense to late interaction). But with this approach model meta that stored in results can be old and not match with what we have in code, so I'm not sure if it good approach 

Image

mteb/api/adapters.py
```diff
@@ -84,6 +88,39 @@ def model_meta_to_schema(
     return cached.model_copy(update={"zero_shot_pct": int(zero_shot_pct)})
 
 
+def run_model_meta_to_schema(
+    run_meta: dict[str, Any],
+    *,
+    zero_shot_pct: int | None = None,
+) -> ModelMetaSchema | None:
+    """`ModelMetaSchema` built straight from one *run's own* `model_meta.json`.
```
KennethEnevoldsen, line 96: Ahh but we I am not at all sure we trust the model_meta.json except for the experiment kwargs (the rest can change)

KennethEnevoldsen: I suspect that we need to handle a merge here

KennethEnevoldsen: (e.g. training data)

mteb/api/aggregators.py
```diff
@@ -244,6 +244,49 @@ async def build_benchmark_summary(  # noqa: PLR0914
...
+    # ``experiments`` column so variant rows can surface the exact ablation
+    # dict that produced them. Empty when the parquet pre-dates the
+    # experiments work.
+    #
+    # Also carries each variant's own `model_meta.json` dict (the
+    # ``model_meta`` column) so its row can build a `ModelMetaSchema` from
+    # that run's *actual* metadata instead of the static MODEL_REGISTRY
+    # entry — an ablation can change more than the kwarg it's named after
+    # (e.g. jinaai/jina-embeddings-v4's `vector_type=multi_vector`
+    # experiment runs `late-interaction`, not the base model's `dense`).
```
KennethEnevoldsen, line 258: this is very long and I am not quite sure what it is trying to say -- I suspect it is ai generated an probably can be rewritten to be more precise

mteb/benchmarks/_create_table.py
```diff
@@ -252,24 +252,63 @@ def _get_embedding_size(embed_dim: int | Sequence[int] | None) -> int | None:
     return None
 
 
+_VARIANT_ID_COL = "_variant_id"
+
+
+def _ensure_variant_id(pl_df: pl.DataFrame) -> pl.DataFrame:
+    """Add a stable ``_variant_id`` Utf8 column derived from ``experiments``.
```
KennethEnevoldsen, line 259: this column seems a bit unclear to me? Is it just the model id (including experiments)

variant_id seems very vague to me

KennethEnevoldsen: or is it just the id for the experiment? in which case why not use `experiment_id`?

mteb/results/benchmark_results.py
KennethEnevoldsen: Hmm maybe we need to merge. Essentially state what we trust from the code and what we allow to change.

e.g. license does not change, training data should probably just include the joint set (I could imagine experiment fitting on data).

We probably need some tests in the results repo to validate the model_meta.json if we plan to use them like this (e.g. checking that license match)

KennethEnevoldsen: Or we should be clear what an experiment can an can't change (and then merge on the rest)

mteb/results/benchmark_results.py
```diff
@@ -256,16 +256,21 @@ def join_revisions(self) -> BenchmarkResults:
         Returns:
             A new BenchmarkResults object with the revisions joined.
         """
+        _experiment_sentinel = ""
```
KennethEnevoldsen, line 259: I am not sure why this is needed (also sting seems like an incorrect replacement for experiment name, I would argue that at non-experiment should have None)

mteb/results/benchmark_results.py
KennethEnevoldsen, line 1: would love a test on these changes

mteb/cache/result_cache.py
KennethEnevoldsen, line 1: would love a test on these changes

KennethEnevoldsen (commented):
initial review, I think I am mostly missing info - can you add a PR comment describing the PR?

Looks like we also need to update API docs?

mteb/results/benchmark_results.py
Samoed: Yes, but overall this complicates pipeline and I'm not sure if we really want to add, because this is really niche use case. I think only embed dim and dtype can be changes, but experiment "tag" would give more visibility. I think we have only one model with such case like jina which probably just rename arg instead and not to save `model_meta` in results

mteb/api/aggregators.py
Samoed: Yes, I'll update a bit later

mteb/results/benchmark_results.py
KennethEnevoldsen: So what are you suggesting?

Samoed: I'm not sure if we want to cover cases like this. Ovearall, I think we can scope that `model_type` and `embed_dim` can change in experiments to be loaded, but I think would be just easier to ignore this, because embed dim would be more visilble as "eximent" tag and `model_type` change is very rare case which we might want to ignore it (overall I know only jina-v4 and in theory bge-m3 with this case) and update experiment kwarg a bit to be more clear on ledaerboard. What do you think?

KennethEnevoldsen: Hmm I could just imagine it to be a more common case in the future (one model to do more things) -- but agree I would probably consider reducing experiments to only include embed-dim and dtype for now (then we can expand it in the future). Does that work?

Samoed: Yes, I think that would work

KennethEnevoldsen:
@Samoed I assume this is still ongoing - are you waiting for anything on my end?

Samoed:
Not yet, I'll update this soon

KennethEnevoldsen:
@Samoed will remove stale here, I assume we don't want this closed

Samoed:
Yeah, I'll try to get back to my pr tomorrow. Hope, there won't be much a lot of pr to review

KennethEnevoldsen:
I am trying to get back to some reviewing now that I see that the ICLR paper will make it (the gsm one). I was a bit worried it wouldn't so had to work a bit overtime on that

Samoed:
I haven't look into gym at all. Maybe I need to look at it a bit too

KennethEnevoldsen:
No the math one (I haven't looked at gym either)

mteb/cache/result_cache.py
```diff
@@ -718,13 +722,16 @@ def _load_from_cache(
                 {cache_path}/leaderboard/{cache_filename}.
             rebuild: If True, force a full rebuild from the results repository, bypassing any
                      pre-computed JSON cache.
+            save_rebuilt_to_disk: Whether a full rebuild (strategy 3, or
+                forced via ``rebuild=True``) persists its result to
+                ``cache_path``. See ``_rebuild_from_full_repository``.
```
KennethEnevoldsen: I am not sure what this means

mteb/cache/result_cache.py
```diff
@@ -941,43 +964,49 @@ def get_task_names(
...
-                f"model_meta.json not found in {revision_path}, extracting model_name and revision from the path"
-            )
+        model_meta_path = revision_path / "model_meta.json"
+        if not model_meta_path.exists():
             if _EXPERIMENTS_FOLDER_NAME in revision_path.parts:
-                logger.debug(
-                    f"Path {revision_path} contains an experiment folder, extracting model_name and revision accordingly"
+                # Warning (not debug): drops every task file under this
+                # folder, and only logs once per folder since this method
+                # is lru_cache'd.
```
KennethEnevoldsen: I get the last part but not the first part here

mteb/cache/result_cache.py
```diff
@@ -1402,14 +1431,18 @@ def load_results(
...
         for path in paths:
             task_result = TaskResult.from_disk(path)
 
             if only_main_score:
                 task_result = task_result.only_main_score()
-            model_name, revision, experiment_name = (
-                self._get_model_name_and_revision_from_path(path.parent)
-            )
+            # `_get_model_name_and_revision_from_path` already warns (once
+            # per folder, lru_cache'd) when it returns None.
```
KennethEnevoldsen:
```suggestion
```

mteb/models/model_meta.py
```diff
@@ -1850,6 +1853,19 @@ def _collect_similar_tasks(dataset: str, visited: set[str]) -> set[str]:
     return similar
 
 
+def _has_meaningful_value(value: Any) -> bool:  # noqa: ANN401 -- checks arbitrary kwarg values
```
```suggestion
def _has_meaningful_value(value: Any) -> bool:  # noqa: ANN401
```

KennethEnevoldsen (commented):
mostly minor stuff

mteb/cache/result_cache.py
Samoed: `save_rebuilt_to_disk=False` won't save `__cached_results.json` in cache. I had some validation problems with it, and I don't think it was possible to load it before pr. Overall, I don't think it was really used at any time

mteb/cache/result_cache.py
Samoed: No results in the `experiments` folder will be loaded if the `model_meta` folder does not exist

PR #4904, opened 2026-07-08
fix: ModelMeta Validation

### LINKED ISSUE #4902: Currently not possible to submit results from cache due to ModeMeta validation
### Describe the bug

Source of error https://github.com/embeddings-benchmark/mteb/blob/702c9fa4165f085272759b3d3eb43cac3796d4ee/mteb/cache/result_cache.py#L1081-L1087

### To reproduce

```
cache = mteb.ResultCache("/cache/")
cache.submit_results(create_pr=True)
Failed to load ModelMeta from /Users/samoed/Downloads/vidore_res/results/google__siglip-base-patch16-224/7fd15f0689c79d79e38b1c2e2e2370a7bf2761ed/model_meta.json: 1 validation error for ModelMeta
loader
  Input should be callable [type=callable_type, input_value='SiglipModelWrapper', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/callable_type
Failed to load ModelMeta from /Users/samoed/Downloads/vidore_res/results/google__siglip-base-patch16-256-multilingual/8952a4eafcde3cb7ab46b1dd629b33f8784ca9c6/model_meta.json: 1 validation error for ModelMeta
loader
  Input should be callable [type=callable_type, input_value='SiglipModelWrapper', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/callable_type
```

### Additional information

_No response_

No

ayush1298:
closes #4902.
Updated the signature of the `loader` to accept `str` along with `Callable[..., MTEBModels]` and added a field validator that will check if a string is provided. If a string is provided, it tries to match it against any registered loaders in `MODEL_REGISTRY` and dynamically resolve it back to the original Callable.
If it cannot find a registered loader or if `MODEL_REGISTRY` is not populated yet, it still allows the string value to pass validation safely.
It gave an error previously because, when saving model metadata in `model_meta.json`, it serializes `loader` to `string`. When calling `submit_results`, the code reads `model_meta.json` from the cache and attempts to reconstruct the ModelMeta object using `ModelMeta.model_validate_json()`, and during this, Pydantic expects loader to be `Callable` or `None`, but it is a `string` so it is giving an error.

### DIFF
--- a/mteb/cache/result_cache.py
+++ b/mteb/cache/result_cache.py
@@ -37,7 +37,11 @@ from mteb.benchmarks.benchmark import Benchmark
 from mteb.benchmarks.get_benchmark import get_benchmark
 from mteb.models import ModelMeta
 from mteb.models.get_model_meta import get_model_metas
-from mteb.models.model_meta import _serialize_experiment_kwargs_to_name
+from mteb.models.model_implementations import MODEL_REGISTRY
+from mteb.models.model_meta import (
+    _get_loader_name,
+    _serialize_experiment_kwargs_to_name,
+)
 from mteb.results import BenchmarkResults, ModelResult, TaskResult
 from mteb.results.task_result import (
     _read_run_settings_from_file,
@@ -1080,8 +1084,21 @@ class ResultCache:
 
         try:
             with meta_file.open("r") as f:
-                meta_dict = f.read()
-            return ModelMeta.model_validate_json(meta_dict)
+                meta_dict = json.load(f)
+
+            # Resolve the string loader name back to the Callable registry object
+            loader_name = meta_dict.get("loader")
+            resolved_loader = None
+            if isinstance(loader_name, str):
+                for model_meta in MODEL_REGISTRY.values():
+                    if model_meta.loader is not None:
+                        name = _get_loader_name(model_meta.loader)  # type: ignore[arg-type]
+                        if name == loader_name:
+                            resolved_loader = model_meta.loader
+                            break
+
+            meta_dict["loader"] = resolved_loader
+            return ModelMeta.model_validate(meta_dict)
         except Exception as e:
             logger.warning(f"Failed to load ModelMeta from {meta_file}: {e}")
             return None
--- a/mteb/cache/result_cache.py
+++ b/mteb/cache/result_cache.py
@@ -37,11 +37,7 @@ from mteb.benchmarks.benchmark import Benchmark
 from mteb.benchmarks.get_benchmark import get_benchmark
 from mteb.models import ModelMeta
 from mteb.models.get_model_meta import get_model_metas
-from mteb.models.model_implementations import MODEL_REGISTRY
-from mteb.models.model_meta import (
-    _get_loader_name,
-    _serialize_experiment_kwargs_to_name,
-)
+from mteb.models.model_meta import _serialize_experiment_kwargs_to_name
 from mteb.results import BenchmarkResults, ModelResult, TaskResult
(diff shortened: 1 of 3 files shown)


mteb/models/model_meta.py
```diff
@@ -229,7 +229,7 @@ class ModelMeta(BaseModel):  # noqa: PLR0904
     model_config = ConfigDict(extra="forbid")
 
     # loaders
-    loader: Callable[..., MTEBModels] | None
+    loader: Callable[..., MTEBModels] | str | None
     loader_kwargs: dict[str, Any] = field(default_factory=dict)
```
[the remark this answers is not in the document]
ayush1298: fix

mteb/models/model_meta.py
```diff
@@ -229,7 +230,7 @@ class ModelMeta(BaseModel):  # noqa: PLR0904
     model_config = ConfigDict(extra="forbid")
 
     # loaders
-    loader: Callable[..., MTEBModels] | None
+    loader: Callable[..., MTEBModels] | str | None
```
Samoed: I don't think that adding str to loader is good idea, but we need to load model meta from file somewhow. I think would be easier just to load file firstly and then remove loader string, but I'm not sure how if it will be better

ayush1298: > I don't think that adding str to loader is good idea,

Is losing type safety is the concern here?

> I think would be easier just to load file firstly and then remove loader string, but I'm not sure how if it will be better

I guess we can do that, after loading file as a dictionary `(meta_dict),` we can look up and resolve the loader string name to the actual `Callable` (or set it to `None` if not resolvable), and then call `ModelMeta.model_validate(meta_dict).` By these way, we dont have to make changes in `loader` signature

Samoed: > Is losing type safety is the concern here?

Some users could use str as loader in metadata and we don't handle it

ayush1298: so, should we do the 2nd option of loading from file and then handling?

Samoed: I think so

ayush1298: made changes

mteb/cache/result_cache.py
```diff
@@ -1080,8 +1084,21 @@ def _load_model_meta_from_cache(
...
+            if isinstance(loader_name, str):
+                for model_meta in MODEL_REGISTRY.values():
+                    if model_meta.loader is not None:
+                        name = _get_loader_name(model_meta.loader)  # type: ignore[arg-type]
+                        if name == loader_name:
+                            resolved_loader = model_meta.loader
+                            break
+
+            meta_dict["loader"] = resolved_loader
+            return ModelMeta.model_validate(meta_dict)
```
Samoed: I think would be better to convert to method in ModelMeta

ayush1298: done

mteb/models/model_meta.py
```diff
@@ -414,6 +414,24 @@ def _check_name(cls, v: str | None) -> str | None:
             )
         return v
 
+    @classmethod
+    def model_validate_json_resolved(cls, json_data: str | bytes) -> ModelMeta:
```
Samoed, line 418: Can you create test for this function?

ayush1298: added

Samoed (approved):

KennethEnevoldsen (approved):

PR #4905, opened 2026-07-08
Remove AfriMTEB task dataset_transform()

TianyouXie828:
[edit] Remove `dataset_transform()` for some tasks as the dataset was uploaded in the format MTEB expects already. 

This PR fixes task schema issues for several AfriMTEB tasks.


Samoed (commented):
Can you give some description about your changes? Seems that your changes will changes existing scores and in that case you need to create v2 versions of tasks and benchmarks

TianyouXie828:
Sorry for the delayed response. Over the past two weeks, I have been working with a collaborator on several other AfriMTEB task fixes and reconsidering how to organize the changes into separate PRs.

Regarding the scores: with the currently configured dataset revisions, the existing implementations of these three tasks fail during dataset transformation and therefore do not produce evaluation scores.

The failures are caused by outdated schema assumptions:

- AfriHateClassification expects `tweet` and string labels, while the current dataset provides `text` and integer labels.
- KinNewsClassification expects separate `title` and `content` columns, while the current dataset already provides a combined `text` column.
- AfriXNLI expects `premise`, `hypothesis`, and `label`, while the current dataset already provides `sentence1`, `sentence2`, and binary `labels`.

Therefore, this PR is a bug fix that restores these tasks to a runnable state. It does not intentionally change the benchmark definitions or replace scores produced by the current implementations, since the current implementations fail before producing any results.

mteb/benchmarks/benchmarks/benchmarks.py
```diff
@@ -1344,11 +1344,11 @@
             "MasakhaNEWSClassification",
             "KinNewsClassification",
             "AfriXNLI",
-            "EmotionAnalysisPlus",
+            #"EmotionAnalysisPlus",
```
KennethEnevoldsen: why are these commented out?

mteb/tasks/classification/multilingual/afri_hate_classification.py
KennethEnevoldsen: Isn't it better just to reupload the dataset in the correct format and remove the dataset transform? Dataset transforms can often lead to bugs in certain version of the datasets, which is why we tend to avoid it.

for how to reupload the dataset in the correct format you can find the guides [here](https://docs.mteb.org/contributing/adding_a_dataset/#pushing-the-dataset-to-the-hub)

if you need access to the MTEB org for these datasets let me know

mteb/tasks/pair_classification/multilingual/afri_xnli.py
```diff
@@ -54,19 +54,7 @@ class AfriXNLI(AbsTaskPairClassification):
...
-                        "labels": 0 if example["label"] == 2 else 1,
-                    }
-
-                self.dataset[lang][split] = ds.map(
-                    map_labels, remove_columns=ds.column_names
-                )
+        # The current HF dataset already has the correct schema:
+        # sentence1, sentence2, labels with binary 0/1 labels.
+        # Do not filter or remap labels.
+        return
```
KennethEnevoldsen: just delete the function

KennethEnevoldsen (commented):
Glad to see the fix, but I think we need to polish this up a bit more to merge it. Outlined a few suggestions below

mteb/benchmarks/benchmarks/benchmarks.py
TianyouXie828: These tasks were commented out only for local testing and were accidentally included in the commit. I have restored the benchmark definition and removed this unrelated change from the PR. Thanks for catching it.

mteb/tasks/pair_classification/multilingual/afri_xnli.py
TianyouXie828: Ok, this function has been deleted

mteb/benchmarks/benchmarks/benchmarks.py
Samoed, line 1: Seems changes were applied incorrectly

TianyouXie828: The previous restoration accidentally brought newer unrelated changes from `main` into this branch. I have now restored the file to the branch's original base, so `benchmarks.py` is no longer part of this PR.

isaac-chung (commented):
In addition to @KennethEnevoldsen 's unresolved comment, it would be great to remove the commented out code too.

mteb/tasks/classification/multilingual/afri_hate_classification.py
isaac-chung: I don't think a reupload is needed for this task. It is already in the correct format. We can remove the dataset_transform(). https://huggingface.co/datasets/mteb/AfriHateClassification/viewer/amh/test

PR #4909, opened 2026-07-10
model: Add fusion-embedding-1 (EximiusLabs/fusion-embedding-1-2b-preview)

abtonmoy:
Adds `EximiusLabs/fusion-embedding-1-2b-preview`, a unified multimodal embedding model
covering text, images, video, and audio in a single vector space. It extends the frozen
Qwen3-VL-Embedding-2B (text/image/video paths byte-identical to the original release)
with an audio pathway: a trained 16.4M-parameter perceiver-resampler maps frozen
Qwen2.5-Omni audio-tower features into the base model's input space (both towers
frozen; audio–text contrastive training).

The model loads through `transformers.AutoModel` with `trust_remote_code=True`; the
remote code downloads the frozen base and audio tower from their original repositories
and only the trained connector weights from the model repository. The wrapper exposes
all four encoding paths (audio, image, text, video); inputs that combine modalities
are embedded as the element-wise sum of the per-modality embeddings. Extra dependencies are declared in the `fusion-embedding` optional-dependency
group.

## MAEB(beta) results

Evaluated on 10 MAEB(beta) tasks across all seven task types with this implementation
(main_score; results PR: https://github.com/embeddings-benchmark/results/pull/601):

| Task | Score | Task | Score |
|---|---:|---|---:|
| BeijingOpera | 92.79 | RavdessZeroshot | 32.22 |
| ClothoT2ARetrieval | 27.84 | SpeechCommandsZeroshotv0.02 | 18.95 |
| GTZANAudioReranking | 70.66 | UrbanSound8KT2ARetrieval | 0.94 |
| GTZANGenre | 63.90 | VehicleSoundClustering | 3.49 |
| MACST2ARetrieval | 13.23 | FSD2019Kaggle | 77.82* |

\* evaluated but withheld from the results submission: 608 of FSDKaggle2019's 4,481
test clips (13.6%) appear in the FSD50K dev split, which is in this model's training
corpus (verified by Freesound id via the Zenodo post-competition metadata).
`FSD2019Kaggle` is listed in `training_datasets` accordingly.

**Video (MVEB beta):** MSVDT2VRetrieval nDCG@10 **82.6** (recall@1 69.2), run end-to-end
through this wrapper (torchcodec-decoded frames, standard test split). The video path is
the frozen base model's own; no video-specific training.

Validation performed:

- Ran `MSVDT2VRetrieval` end-to-end through the wrapper (video path); the reworked
  frame preprocessing was verified bitwise-identical to the base model's reference
  pipeline.

- Confirmed `mteb.get_model(...)` and `mteb.get_model_meta(...)` resolve at the pinned
  revision and the meta registers in `MODEL_REGISTRY`.
- Verified on GPU that the `AutoModel` remote-code path produces outputs identical to
  the model repositories' reference implementation for text and audio.
- The model card's benchmark numbers were produced with the same encoding paths this
  wrapper implements (no separate paper).
- `training_datasets` lists every mteb task built on a training corpus (AudioCaps,
  FSD50K, AudioSet); Clotho/ESC-50/UrbanSound8K/VGGSound evaluation clips were excluded
  from training by id blacklist at ingestion (see the model card).


mteb/models/model_implementations/fusion_embedding_models.py
```diff
@@ -0,0 +1,190 @@
...
+        try:
+            import dataclasses
+
+            from fusion_embedding.config import FusionConfig
+            from fusion_embedding.hf_components import load_audio_tower, load_base
+            from fusion_embedding.model import FusionEmbeddingModel
+        except ImportError as e:
+            raise ImportError(
+                "fusion-embedding-1 requires its package: "
+                "pip install git+https://github.com/Eximius-Labs/fusion-embedding-1"
```
Samoed: Can you add your library to pyproject as extra depedency? https://docs.mteb.org/contributing/adding_a_model/#adding-model-dependencies

mteb/models/model_implementations/fusion_embedding_models.py
```diff
@@ -0,0 +1,190 @@
...
+            ) from e
+
+        self.device = device or ("cuda" if torch.cuda.is_available() else "cpu")
+        ckpt_path = hf_hub_download(model_name, CKPT_FILE, revision=revision)
```
Samoed: Can you make your model compatible with transformers `AutoModel` with `trust_remote_code`?

mteb/models/model_implementations/fusion_embedding_models.py
abtonmoy: Added — a `fusion-embedding` extras group with the audio/vision runtime libraries
(librosa, soundfile, torchvision, pillow, accelerate), referenced from the ModelMeta
as `extra_requirements_groups=["audio", "fusion-embedding"]`. Since the model now
loads through `AutoModel` with `trust_remote_code` (other thread), the git package
dependency is gone entirely.

mteb/models/model_implementations/fusion_embedding_models.py
abtonmoy: Done. Both model repos now ship remote-code files
(`configuration_fusion_embedding.py` / `modeling_fusion_embedding.py` with `auto_map`
in config.json), so
`AutoModel.from_pretrained("EximiusLabs/fusion-embedding-1-2b-preview",
trust_remote_code=True)` works with plain transformers, and the wrapper here reduces
to that call plus batching. The remote code loads only the trained connector weights
from the model repo; the frozen Qwen3-VL-Embedding base and Qwen2.5-Omni audio tower
download from their original repositories. Verified on GPU that the remote-code
outputs match our reference implementation exactly. Revision is re-pinned to the
commit that carries the remote code (checkpoint bytes unchanged vs the scored tag).

mteb/models/model_implementations/fusion_embedding_models.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+fusion_embedding_1_2b_preview = ModelMeta(
+    loader=FusionEmbeddingWrapper,
+    name="EximiusLabs/fusion-embedding-1-2b-preview",
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="e6d91bc06920e74553b5ea52244ebdf7d1a82402",
+    release_date="2026-07-06",
+    modalities=["audio", "text"],
```
Samoed: In the description and examples this model states that it supports images, but you didn't add support for it

abtonmoy: Added — `get_image_embeddings` routes through the frozen base model's own image path (exposed by the repo's remote code; the audio-side components never touch it), and `modalities` now lists `image`. Images are embedded per item, matching other merged wrappers. Fused image+text inputs raise `NotImplementedError` for now. Video encoding is likewise available through the base's API; happy to add it if there are video-bearing tasks you would want covered.

Samoed: > happy to add it if there are video-bearing tasks you would want covered.

We have video tasks if you want to test your model https://github.com/embeddings-benchmark/mteb/blob/aa68e9cacc912e83d3ef637e388b450b14ce5003/mteb/benchmarks/benchmarks/benchmarks.py#L3357-L3467

> Fused image+text inputs raise NotImplementedError for now

Would be better to add it's support, because we have some tasks that combine them. You can implemented it like combination of text and image embedings (same for other modalities)

abtonmoy: Done on both counts, plus the CI failure:

- **Video**: `get_video_embeddings` added — `VideoCollator` frames go through the model's video path, which is the frozen base's own (the repo's remote code now exposes `embed_video`; we verified its outputs bitwise against running the identical frames through the raw base model). `modalities` lists all four.
- **Fused**: implemented as you suggested — element-wise sum of the per-modality embeddings, for any combination of present modalities (the `clip_models.py` convention).
- **CI**: `n_embedding_parameters` is now set (311,164,928 — the base's token-embedding layer); `tests/test_models/test_model_meta.py` passes locally (2171 passed). `qwen-vl-utils` (the base's own video preprocessing package) is added to the `fusion-embedding` extras group, and the revision is re-pinned to the commit carrying the video support.

mteb/models/model_implementations/fusion_embedding_models.py
```diff
@@ -0,0 +1,266 @@
...
+    citation="""
+@software{fusion_embedding_2026,
+  title  = {Fusion Embedding 1: A Unified Embedding Space for Text,
+            Image, Video, and Audio},
+  author = {Tonmoy, Abdul Basit},
+  year   = {2026},
+  url    = {https://github.com/Eximius-Labs/fusion-embedding}
+}
+""",
+    extra_requirements_groups=["audio", "fusion-embedding"],
```
Samoed:
```suggestion
    extra_requirements_groups=["fusion-embedding"],
```

mteb/models/model_implementations/fusion_embedding_models.py
```diff
@@ -0,0 +1,266 @@
...
+                embeddings += modality_embeddings
+        return embeddings
+
+
+fusion_embedding_1_2b_preview = ModelMeta(
+    loader=FusionEmbeddingWrapper,
+    name="EximiusLabs/fusion-embedding-1-2b-preview",
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="e63b6c0b6a9b58c68e1b17fb55698fe5342b8028",
```
Samoed: I think you should update revision to support video

mteb/models/model_implementations/fusion_embedding_models.py
```diff
@@ -0,0 +1,266 @@
...
+    release_date="2026-07-06",
+    modalities=["audio", "image", "text", "video"],
+    n_parameters=2_800_000_000,
+    n_embedding_parameters=311_164_928,  # base token-embedding layer: 151,936 x 2,048
+    memory_usage_mb=10681,  # Calculated using model.calculate_memory_usage_mb()
+    max_tokens=254,
+    embed_dim=[2048, 1536, 1024, 512, 256, 128, 64],  # Matryoshka ladder
```
Samoed:
```suggestion
    n_embedding_parameters=311_164_928,
    memory_usage_mb=10681,
    embed_dim=[2048, 1536, 1024, 512, 256, 128, 64],
```

mteb/models/model_implementations/fusion_embedding_models.py
```diff
@@ -0,0 +1,266 @@
...
+                        self.model.embed_video(
+                            [
+                                Image.fromarray(frame.permute(1, 2, 0).cpu().numpy())
+                                for frame in video
+                            ],
+                            max_frames=self.video_num_frames,
+                        )
+                        for video in batch["video"]
+                    ]
+                )
```
Samoed, line 105: I see that you're using qwen3-vl-utils underthehood, but on our pracrice it don't do anything complex and you can process video with using torchcodec `VideoDecoder` without splitting video to images

mteb/models/model_implementations/fusion_embedding_models.py
abtonmoy: Applied.

mteb/models/model_implementations/fusion_embedding_models.py
abtonmoy: Done — revision is now `b551ea8033be…`, the commit carrying `embed_video` (including the torchcodec rework from the other thread).

mteb/models/model_implementations/fusion_embedding_models.py
abtonmoy: Applied, with one adjustment: kept `max_tokens=254` (the suggestion range removed it, but it is a required `ModelMeta` field — validation fails without it). Comments stripped as suggested.

mteb/models/model_implementations/fusion_embedding_models.py
abtonmoy: Done. The wrapper now passes the `VideoCollator`'s decoded frame tensors directly to the model (no image splitting), and `qwen-vl-utils` is removed everywhere, including the extras group — the remote code reproduces the base's reference preprocessing natively and decodes file paths with torchcodec's `VideoDecoder`. We verified the reworked path produces bitwise-identical outputs to both the previous implementation and the raw base model on identical inputs before pushing. To exercise it on a real task, we also ran `MSVDT2VRetrieval` end-to-end through this wrapper (nDCG@10 = 0.826, recall@1 = 0.692); happy to add more video results to the results PR if useful.

Samoed (approved):

PR #4911, opened 2026-07-11
fix to EmotionAnalysisPlus : align task with pre-processed mirror export

O-Nicolinho:
Issue:

EmotionAnalysisPlus fails to load out of the box. Here's why: The "validation" split is declared but empty for every config, so the loader raises "Instruction "validation" corresponds to no data!" when materializing splits. The mirror already ships data in the expected {text, label} form, but the task's dataset_transform was written for the raw per-emotion-column format (anger, disgust, etc as separate columns) and errors when those columns aren't found.

Fix:

- We pin split="test" in the dataset metadata so the loader only builds the populated test split and never touches the empty validation one. 
-We drop "validation" from eval_splits
- In dataset_transform we re-wrap the resulting bare Dataset into a DatasetDict (single-split load returns a bare Dataset), and pass through splits already in the expected form. The original raw per-emotion-column path is kept as a fallback.

This fix allows the task to run as shipped.

### DIFF
--- a/mteb/tasks/multilabel_classification/multilingual/emotion_analysis_plus.py
+++ b/mteb/tasks/multilabel_classification/multilingual/emotion_analysis_plus.py
@@ -25,11 +25,12 @@ class EmotionAnalysisPlus(AbsTaskMultilabelClassification):
         dataset={
             "path": "mteb/EmotionAnalysis",
             "revision": "554dbe305cad4f86705c8b6389c76f7f33fc6fd8",
+            "split": "test",
         },
         type="MultilabelClassification",
         category="t2c",
         modalities=["text"],
-        eval_splits=["validation", "test"],
+        eval_splits=[ "test"],
         eval_langs={
             # (ISO-639-3 code : BCP-47 tag)
             "afr": ["afr-Latn"],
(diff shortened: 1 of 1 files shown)


mteb/tasks/multilabel_classification/multilingual/emotion_analysis_plus.py
```diff
@@ -91,47 +92,62 @@ class EmotionAnalysisPlus(AbsTaskMultilabelClassification):
...
             for split in self.dataset[lang]:
                 ds = self.dataset[lang][split]
-
-                # ── 1️⃣  locate the text column ───────────────────────────────
                 cols = ds.column_names
-                text_col = next(c for c in self._LOOKUP_TEXT if c in cols)
 
-                # ── 2️⃣  locate all emotion columns that are present ──────────
+                # -- 0.5️⃣  already canonical {text, label}: pass through.
+                if "text" in cols and "label" in cols:
```
Samoed: Why this is needed? Does dataset have different columns per subset?

O-Nicolinho: I just double checked and all 30 configs are already in the (Text,label) format. I think the original file was written to handle the case where it had to deal with the raw emotion columns instead of the organized label one that aggregates everything. My fix just skipped that extra (needless) step, since given that the dataset is already in the correct format, that preprocessing would simply never happen. 

I just updated it and removed the dead code to simplify everything and make it less convoluted.

PR #4878, opened 2026-07-02
Add CO₂ estimate per million tokens from active parameters

fzowl:
Fixes #4368.

- Add CO₂ estimate per million tokens from active parameters

PR #4918, opened 2026-07-13
Fix #4368: Add CO₂ inference-cost estimate (gCO₂/Mtoken) to model metadata

### LINKED ISSUE #4368: Display CO₂ estimate per million tokens

We have active parameters for most models in MTEB, which is sufficient to estimate inference cost as gCO₂ / Mtoken. This is a useful signal for model selection that no other embedding leaderboard currently surfaces.
Proposed display

A compact badge (e.g. ~42 gCO₂ / Mtoken) in the model documentation page. On hover, a popover expands showing the assumptions behind the estimate:
- Active parameters
- Benchmark hardware (e.g. A100 80GB)
- Carbon intensity (400 gCO₂/kWh, ADEME world 
- PUE (1.2)

Image

All estimates are prefixed with ~ — we do not have visibility into deployment hardware or grid carbon intensity, so confidence should not be overstated for any model. The footnote in the popover makes clear that the figure reflects benchmark conditions, not real-world deployment.

For closed models where active parameters are not disclosed, we have two options: (1) fall back to EcoLogits estimates, or (2) estimate active parameters from observed token throughput and known hardware specs, since inference latency is roughly linear in active parameters. Either way, the assumption should be noted explicitly in the popover.

**Methodology**

To be defined, but we could follow the EcoLogits linear regression structure (E_per_token = a * P_active + b) fitted on embedding-specific benchmark data, so that our estimates are on the same scale as EcoLogits figures for closed generative models. 

**References**

- EcoLogits methodology: https://ecologits.ai/0.2/methodology/llm_inference/
- Luccioni et al. (2024) — “Power Hungry Processing: Watts Driving the Cost of AI Deployment?” — the most directly relevant prior work; measures deployment cost as energy and carbon required to perform 1,000 inferences across task-specific and general-purpose models, finding that generative architectures are orders of magnitude more expensive than task-specific systems : https://arxiv.org/abs/2311.16863
- Faiz et al. (2024) — “LLMCarbon: Modeling the end-to-end Carbon Footprint of Large Language Models” — covers both dense and MoE LLMs including embodied carbon: https://arxiv.org/abs/2309.14393
- Luccioni et al. (2022) — “Estimating the Carbon Footprint of BLOOM” — foundational methodology for inference carbon accounting including PUE: https://jmlr.org/papers/volume24/23-0069/23-0069.pdf

Related to #4366

fzoll:
Closes #4368

## What

Adds an estimate of a model's inference carbon cost, expressed as grams of CO₂-equivalent per million tokens (gCO₂ / Mtoken), derived from its **active parameters**. This surfaces a model-selection signal (relative inference cost) that no other embedding leaderboard currently shows.

## Methodology

Follows the EcoLogits linear energy model referenced in the issue:

```
energy_per_token (Wh) = a · active_params_billions + b
```

The GPU energy is scaled by the datacenter PUE and multiplied by the grid carbon intensity. Using the EcoLogits coefficients keeps the figures on roughly the same scale as EcoLogits estimates for closed generative models, as suggested in the issue.

All assumptions are documented module constants in `mteb/models/model_meta.py` so they can be **refit on embedding-specific benchmark data** later:

- Energy coefficients `a`, `b`
- Carbon intensity — 400 gCO₂/kWh (ADEME world average)
- PUE — 1.2
- Benchmark hardware — A100 80GB

All figures are approximate (to be displayed with a `~` prefix) and reflect **benchmark conditions**, not any specific real-world deployment.

## API surface

- `ModelMeta.co2_cost_per_million_tokens` → the numeric estimate (`None` when active parameters are unknown).
- `ModelMeta.co2_cost_estimate` → the estimate plus the assumptions behind it, intended to back a compact badge with a popover (per the mockup in the issue).
- `Co2EstimateSchema` added to `ModelMetaSchema` so the leaderboard API exposes the estimate + assumptions to the frontend.

## Tests

Added unit tests in `tests/test_models/test_model_meta.py` covering the linear model, monotonicity in active parameters, use of active (not total) parameters, the reported assumptions, and the `None` case. All pass locally.

## Notes / follow-ups

- The rendered badge + popover UI lives in the SvelteKit frontend; this PR provides the backend estimate and API field it consumes.
- For closed models where active parameters are undisclosed, the estimate returns `None`; the EcoLogits-fallback / throughput-based estimation discussed in the issue can be layered on top later.

### DIFF
--- a/mteb/api/schemas.py
+++ b/mteb/api/schemas.py
@@ -183,6 +183,19 @@ class TaskMetaSchema(_CamelModel):
         )
 
 
+class Co2EstimateSchema(_CamelModel):
+    """Estimated inference carbon cost with the assumptions behind it (for a badge + popover).
+
+    All figures are approximate and reflect benchmark conditions, not real-world deployment.
+    """
+
+    g_co2_per_million_tokens: float
+    active_parameters: int | None
+    benchmark_hardware: str
+    carbon_intensity_g_per_kwh: float
+    pue: float
+
+
 class ModelMetaSchema(_CamelModel):
     """Static model metadata; ``name`` is the canonical ``org/name`` HF identifier."""
 
@@ -204,6 +217,7 @@ class ModelMetaSchema(_CamelModel):
     languages: list[str] = Field(default_factory=list)
     citation: str | None = None
     memory_usage_mb: float | None = None
+    co2_estimate: Co2EstimateSchema | None = None
     license: str | None = None
     public_training_code: str | None = None
     public_training_data: str | None = None
@@ -253,6 +267,11 @@ class ModelMetaSchema(_CamelModel):
             memory_usage_mb=float(meta.memory_usage_mb)
             if meta.memory_usage_mb is not None
             else None,
+            co2_estimate=(
+                Co2EstimateSchema(**meta.co2_cost_estimate)
+                if meta.co2_cost_estimate is not None
+                else None
+            ),
             license=str(meta.license) if meta.license else None,
             public_training_code=str(meta.public_training_code)
             if meta.public_training_code
--- a/mteb/models/model_meta.py
+++ b/mteb/models/model_meta.py
@@ -161,6 +161,30 @@ OPEN_LICENSES: frozenset[str] = frozenset(
 )
 
 
+# --- CO2 inference-cost estimation ---------------------------------------------------
+# We estimate the inference carbon cost of a model as grams of CO2-equivalent per one
+# million tokens (gCO2eq / Mtoken). The estimate follows the linear energy model used by
+# EcoLogits (https://ecologits.ai/0.2/methodology/llm_inference/) and Luccioni et al.
+# (2024, https://arxiv.org/abs/2311.16863), so that figures are on roughly the same scale
+# as EcoLogits estimates for closed generative models:
+#
+#     energy_per_token (Wh) = CO2_ENERGY_ALPHA * active_params_billions + CO2_ENERGY_BETA
+#
+# The GPU energy is scaled by the datacenter PUE and multiplied by the grid carbon
+# intensity. All figures are approximate (prefixed with "~" when displayed) and reflect
+# *benchmark* conditions (a single reference GPU on an average grid), not any particular
+# real-world deployment. The coefficients are kept as module constants so they can be
+# refit on embedding-specific benchmark data in the future.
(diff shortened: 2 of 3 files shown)


Samoed:
I think you've already opened PR for that https://github.com/embeddings-benchmark/mteb/pull/4878

KennethEnevoldsen:
So had quite a long discussion on this here:

https://github.com/betagouv/ComparIA/issues/505

the short form: CO2 is very context dependent, what we really care about (and can measure) is energy efficiency, we can do this by splitting the models into energy categories (in the US I believe this use stars instead of letters). Here is the example for **generative** models:

| Class | Wh / 1M tokens |
| ----- | -------------- |
| A     | `< 100`        |
| B     | `100 - 300`    |
| C     | `300 - 1 000`  |
| D     | `1 000 - 3 000`|
| E     | `3 000 - 10 000`|
| F     | `>= 10 000`    |

Which gives something like this:

| Model shape           | Energy     | Class |
| --------------------- | ---------- | ----- |
| 8B dense              | ~90 Wh     | A     |
| 24-35B dense          | ~110-250   | B     |
| 70B dense             | ~660 Wh    | C     |
| 120B MoE (5B active)  | ~340 Wh    | C     |
| 400B MoE (17B active) | ~1 600 Wh  | D     |
| 700B dense            | ~30 000 Wh | F     |

For more discussion see the issue, but generally I think we should focus on energy usage and then create our energy class bands. One thing we might need to consider which they do not is embedding size (which for us has downstream effect)

Let me know what you think about this alternative. For me it is easier to justify than co2.

Samoed:
If energy effiency will be based on model size, I don't think that it will much new information and all models will be in `A`, because they're under 24B size

KennethEnevoldsen:
> If energy effiency will be based on model size, I don't think that it will much new information and all models will be in A, because they're under 24B size

Yeah we would have to recreate the scale for embedding models

Samoed:
I'm not sure that this would give new information in that case. Also, this is not architecture dependent

KennethEnevoldsen:
I think this is a case of a thing that needs more discussion before we can propose a solution. Sorry for the inconvenience @fzoll. Will close this for now

> Also, this is not architecture dependent

Energy usage is architecture dependent (but non-trivial: see https://github.com/betagouv/ComparIA/issues/505)

PR #4922, opened 2026-07-14
model: Nemotron 3 Embed models

ybabakhin:
This PR adds a new family of Nemotron 3 Embed text embedding models


Samoed (approved):
Сongratulations with release! Everything looks good

PR #4925, opened 2026-07-14
Update revision of minishlab/potion-code-16M-v2

ayush1298:
updated revision of `minishlab/potion-code-16M-v2` model


Samoed:
Why this is needed?

ayush1298:
> Why this is needed?

Sorry, not needed. I thought they had updated the model and thereby updated the weight, but it's just readme change. Will close these PR

PR #4926, opened 2026-07-14
Add Hanno-Labs/dinghy-law-0.6b-v1 (legal embedding model)

trashhalo:
Adds the `ModelMeta` for `Hanno-Labs/dinghy-law-0.6b-v1`, a compact (0.6B) legal text-embedding model — a contrastive fine-tune of `Qwen/Qwen3-Embedding-0.6B`. It keeps the base architecture/interface exactly (last-token pooling, cosine, query-side `Instruct: {instruction}\nQuery:` prefix with unprefixed documents), so it reuses the shared Qwen3 instruct loader. Public weights (apache-2.0), permissively/publicly-licensed training data.

### Results — MTEB(Law, v1), reproduced with `mteb.evaluate` 2.18.1 (this PR's ModelMeta config)

| Task | nDCG@10 |
|---|---|
| AILACasedocs | 41.32 |
| AILAStatutes | 77.58 |
| LegalSummarization | 61.74 |
| GerDaLIRSmall | 39.33 |
| LeCaRDv2 | 73.62 |
| LegalBenchConsumerContractsQA | 79.58 |
| LegalBenchCorporateLobbying | 94.95 |
| LegalQuAD | 58.52 |
| **Mean(Task)** | **65.83** |

For reference, voyage-law-2 = 65.39 on the same benchmark.

Note: declared non-zero-shot on MTEB(Law) — the model trained on GerDaLIR, which overlaps the GerDaLIRSmall task (declared in `training_datasets`).

<sup>†</sup> **Hardware note:** these numbers were produced on an NVIDIA **H200** in bf16. The two smallest tasks (AILACasedocs and AILAStatutes, ~50 queries each) are sensitive to GPU/bf16 matmul differences at the ±~0.5 nDCG@10 level; the identical eval on an **A100** yields Mean(Task) = **65.74** (AILACasedocs 40.69 / AILAStatutes 77.29, all other tasks within ±0.1). Model weights, instructions, and eval config are byte-identical across runs — only the GPU differs. Both settings rank #2 on MTEB(Law, v1).

Results PR: embeddings-benchmark/results#613

### Checklist

- [x] I reproduced results (included above).

### DIFF
--- a/mteb/models/model_implementations/hanno_labs_models.py
+++ b/mteb/models/model_implementations/hanno_labs_models.py
@@ -1,0 +1,75 @@
+from __future__ import annotations
+
+from mteb.models.model_implementations.qwen3_models import q3e_instruct_loader
+from mteb.models.model_meta import ModelMeta
+
+# dinghy-law-0.6b-v1 is a legal-domain contrastive fine-tune of Qwen/Qwen3-Embedding-0.6B. It keeps the
+# base architecture and interface exactly (last-token pooling, cosine similarity, query-side
+# "Instruct: {instruction}\nQuery:" prefix with unprefixed documents), so it reuses the shared Qwen3
+# instruct loader. Trained on public/permissively-licensed legal corpora; the GerDaLIR and BillSum splits
+# below overlap MTEB tasks and are declared for zero-shot accounting.
+
+DINGHY_LAW_CITATION = """@misc{dinghy-law-0.6b,
+  title  = {dinghy-law-0.6b: a compact legal text-embedding model},
+  author = {Solka, Stephen},
+  year   = {2026},
+  note   = {Hanno Labs / Clause Logic Inc.},
+  howpublished = {\\url{https://huggingface.co/Hanno-Labs/dinghy-law-0.6b-v1}}
+}"""
+
+# Sources used to train dinghy-law-0.6b-v1 that correspond to MTEB tasks (for train/test leakage accounting).
+# GerDaLIR training data overlaps the GerDaLIR / GerDaLIRSmall corpora; BillSum (US) was used as a
+# general-legal replay set. Other training sources (ECtHR, SCOTUS, LEDGAR, CAIL, GerLayQA, Indian statutes)
+# do not correspond to MTEB tasks. Note: the Indian-statute training text shares public-domain statute
+# documents with the AILAStatutes corpus, though none of AILA's queries or relevance labels were used.
+# The Indian-statute training signal is publicly released as the CC-BY-4.0 dataset
+# https://huggingface.co/datasets/Hanno-Labs/indian-ipc-statute-identification (Indian court judgments are public
+# records under CC-BY-4.0; IPC statute text is a public-domain enactment; section citations masked as [STATUTE]).
+# The training mix is permissively/publicly licensed -- no CC-BY-NC (non-commercial) sources are used.
+training_data = {
+    "GerDaLIR",
+    "GerDaLIRSmall",
+    "BillSumUS",
+}
+
+dinghy_law_0_6b = ModelMeta(
+    loader=q3e_instruct_loader,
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/hanno_labs_models.py
```diff
@@ -0,0 +1,75 @@
...
+
+# Sources used to train dinghy-law-0.6b-v1 that correspond to MTEB tasks (for train/test leakage accounting).
+# GerDaLIR training data overlaps the GerDaLIR / GerDaLIRSmall corpora; BillSum (US) was used as a
+# general-legal replay set. Other training sources (ECtHR, SCOTUS, LEDGAR, CAIL, GerLayQA, Indian statutes)
+# do not correspond to MTEB tasks. Note: the Indian-statute training text shares public-domain statute
+# documents with the AILAStatutes corpus, though none of AILA's queries or relevance labels were used.
+# The Indian-statute training signal is publicly released as the CC-BY-4.0 dataset
+# https://huggingface.co/datasets/Hanno-Labs/indian-ipc-statute-identification (Indian court judgments are public
+# records under CC-BY-4.0; IPC statute text is a public-domain enactment; section citations masked as [STATUTE]).
+# The training mix is permissively/publicly licensed -- no CC-BY-NC (non-commercial) sources are used.
```
Samoed: Can you remove logging messages, they're too verbose
```suggestion
```

mteb/models/model_implementations/hanno_labs_models.py
```diff
@@ -0,0 +1,75 @@
...
+training_data = {
+    "GerDaLIR",
+    "GerDaLIRSmall",
+    "BillSumUS",
+}
+
+dinghy_law_0_6b = ModelMeta(
+    loader=q3e_instruct_loader,
```
Samoed:
```suggestion
    loader=SentenceTransformerEncoderWrapper,
```

trashhalo:
Thanks for the review. Both addressed in the latest push (8a344d3):

1. Verbose comments: trimmed to match the repo convention; the licensing and leakage detail lives on the model card instead.
2. Loader: switched to `SentenceTransformerEncoderWrapper`, carrying the per-task instructions via `model_prompts` (query `"Instruct: {instruction}\nQuery:"`, unprefixed documents), the same pattern as the e5/bge/voyage entries. I verified on one GPU that this reproduces the previous `q3e_instruct_loader` numbers exactly: 0.00 delta on all 8 MTEB(Law, v1) tasks (Mean 65.85 to 65.85), so the loader swap is behavior preserving.

mteb/models/model_implementations/hanno_labs_models.py
```diff
@@ -0,0 +1,69 @@
...
+    "LegalBenchConsumerContractsQA": "Retrieve the contract text most relevant to answering the given question.",
+    "LegalBenchCorporateLobbying": "Given a bill title, retrieve the corresponding bill summary.",
+    "GerDaLIRSmall": "Retrieve relevant legal case documents.",
+    "LegalQuAD": "Retrieve relevant legal case documents.",
+    "LeCaRDv2": "Retrieve the case document that most closely matches or is most relevant to the scenario described in the provided query.",
+}
+dinghy_law_prompts = {
+    **{f"{t}-query": f"Instruct: {i}\nQuery:" for t, i in _instructions.items()},
+    **{f"{t}-document": "" for t in _instructions},
+}
```
Samoed: if you're using instructions then you can use https://github.com/embeddings-benchmark/mteb/blob/651d53f380e9d2c7dc5a2bed889a1355386785bb/mteb/models/instruct_wrapper.py#L139 where you can set custom template for inctructions

trashhalo:
Good call, thanks. Switched to `InstructSentenceTransformerModel` with `instruction_template="Instruct: {instruction}\nQuery:"` and the bare per-task instructions in `prompts_dict` (9d2207d). That is cleaner than the hand-built `model_prompts`, and since it is the same wrapper the previous loader was built on, the numbers are unchanged (still 65.85 mean on MTEB(Law, v1)).

Samoed (approved):

PR #4928, opened 2026-07-14
model: Add BGE Small Structural Separator model implementation

thu-nmrc:
## Add BGE Small Structural Separator

This adds the reference `EncoderProtocol` implementation and `ModelMeta` for
`thu-nmrc/bge-small-structural-separator`.

The retriever inserts one learned shared vocabulary token before a document
title and each punctuation-delimited sentence, produces one normalized CLS
vector per document, and produces one unmodified no-instruction query vector.
The standard MTEB search wrapper performs cosine retrieval, while the optional
standard cache wrapper manages reusable embeddings. The model does not fuse,
route, consume qrels, or use task-name-dependent inference.

The learned 384-value row, immutable training manifest, deterministic training
code, algorithm definition, and complete positive and negative evaluation
disclosure are public at
https://github.com/thu-nmrc/bge-small-structural-separator.

There is no archival paper for this release. A self-contained publication-style
method note covering motivation, novelty, formal retrieval definition, training
objective, controlled evaluation, falsification result, limitations, and
related work is available at
https://github.com/thu-nmrc/bge-small-structural-separator/blob/main/METHOD_CARD.md.

### Plain BGE comparison

The control uses the same pinned BGE-small base encoder, no query instruction,
one title-plus-body document vector, and the same exact cosine search. All
values are official MTEB 2.18.1 NDCG@10; no benchmark labels were used for
training or selection.

| Task | Structural separator | Plain BGE | Delta |
| --- | ---: | ---: | ---: |
| SciFact | 0.72081 | 0.72000 | +0.00081 |
| NFCorpus | 0.35474 | 0.33708 | +0.01766 |
| FiQA2018 | 0.38367 | 0.38483 | -0.00116 |
| SCIDOCS | 0.21257 | 0.19726 | +0.01531 |
| ArguAna | 0.61278 | 0.59499 | +0.01779 |
| TRECCOVID | 0.72880 | 0.66596 | +0.06284 |
| QuoraRetrieval | 0.88742 | 0.88786 | -0.00044 |
| DBPedia | 0.37854 | 0.38971 | -0.01117 |
| Eight-task macro | 0.534916 | 0.522211 | +0.012705 |

The full comparison, reproducibility details, negative extension result, and
limitations are documented in the public method repository linked above.

### Verification

- [x] I have filled out the ModelMeta object to the extent possible.
- [x] I have ensured the model can be loaded with `mteb.get_model(model_name, revision)`.
- [x] I have ensured the metadata can be loaded with `mteb.get_model_meta(model_name, revision)`.
- [x] I have tested the standard `EncoderProtocol` on real base weights and the released artifact.
- [x] The encoder produces document tokens, document vectors, query vectors,
  cosine scores, and rankings exactly equal to the previous search adapter on
  the equivalence fixture (maximum absolute difference `0.0`).
- [x] The standard MTEB `CachedEmbeddingWrapper` completes a second
  NanoSciFact run with the backbone disabled, reproducing all score fields from
  cache.
- [x] Official NanoSciFact and NanoArguAna results reproduce all finite score
  fields exactly: NDCG@10 is `0.77211` and `0.64307`, respectively.
- [x] Full SciFact reproduces every score field exactly. Full NFCorpus
  reproduces the submission metrics exactly (NDCG@10 `0.35474`, MAP@100
  `0.16858`, Recall@100 `0.31149`, Recall@1000 `0.64636`). Its auxiliary MRR
  fields differ because the standard search wrapper, rather than the former
  adapter's document-ID rule, now determines the order of exactly tied scores.
- [x] The implementation compiles on Python 3.10 and passes the repository's Ruff 0.14.0 configuration.
- [x] The complete standard Transformers checkpoint is public on Hugging Face at
  https://huggingface.co/thu-nmrc/bge-small-structural-separator, pinned to
  revision `9a0a8aa92400202dd1ef6950ed9cd4a116dfb03d`. All eight remote files pass
  SHA256 verification, and the pinned checkpoint passes
  `AutoModel.from_pretrained()` plus exact separator-row validation.
- [x] The learned row, training code, and training manifest are public.
- [x] Full eight-task BEIR evidence passes the frozen gate with a +0.012705
  NDCG@10 macro delta over protocol-matched plain BGE.
- [x] A preregistered six-task English-v2 Retrieval extension is fully disclosed:
  its macro NDCG@10 delta is -0.011038, so broad unseen-task superiority is not
  claimed. The official ten-task macro delta is +0.002855.
- [x] Fourteen identity-preserving official result files are prepared under the
  final model name and immutable revision.

The results PR intentionally waits for this model implementation to merge,
because results CI installs MTEB main and verifies the registered model.


Samoed (commented):
Can you fully upload your model to huggingface and make it compatible with AutoModel?

thu-nmrc:
Thanks for the review. I have prepared a complete standard Transformers checkpoint (config, tokenizer, and safetensors weights) and verified that `AutoModel.from_pretrained()` loads it locally. An exact state-dict check confirms that it differs from the pinned BGE base only in the learned 384-value `[unused2]` embedding row. I will upload it to `thu-nmrc/bge-small-structural-separator`, pin the immutable Hub revision, and simplify this loader to use that checkpoint directly.

thu-nmrc:
Implemented the requested Hugging Face AutoModel path. The complete public checkpoint is now available at https://huggingface.co/thu-nmrc/bge-small-structural-separator and is pinned in the loader to immutable revision `9a0a8aa92400202dd1ef6950ed9cd4a116dfb03d`.

PR commit `36c15549` removes the GitHub artifact download and runtime row injection; the loader now calls `AutoTokenizer.from_pretrained()` and `AutoModel.from_pretrained()` directly on the pinned Hub checkpoint. I verified all eight remote file SHA256 hashes, exact equality of the learned separator row, Ruff/compile, six focused tests, and an official NanoSciFactRetrieval run loaded from the Hub revision with zero exceptions and score-field-exact equality to the previously verified local checkpoint.

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+    """One-vector exact retrieval with a learned shared structural separator."""
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None,
+        *,
+        device: str | None = None,
+        model_source: str | Path | None = None,
+        batch_size: int = 64,
```
Samoed: Batch size will be passed to in `encode_kwargs`
```suggestion
```

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+class StructuralSeparatorSearch:
+    """One-vector exact retrieval with a learned shared structural separator."""
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None,
+        *,
+        device: str | None = None,
+        model_source: str | Path | None = None,
```
Samoed:
```suggestion
```

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+        self.document_vectors = np.memmap(
+            self._index_path,
+            mode="r",
+            dtype=np.float32,
+            shape=(len(self.docids), dimension),
+        )
+
+    def search(
+        self,
+        queries: Iterable[dict[str, Any]],
```
Samoed:
```suggestion
        queries: Iterable[BatchedInput],
```

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+        top_k: int,
+        encode_kwargs: dict[str, Any],
+        top_ranked: dict[str, list[str]] | None = None,
+        num_proc: int | None,
+    ) -> dict[str, dict[str, float]]:
+        del task_metadata, hf_split, hf_subset, encode_kwargs, num_proc
+        if top_ranked is not None:
+            raise ValueError(
+                "This is a first-stage retriever; reranking inputs are not accepted"
+            )
```
Samoed: So, this pipeline doesn't support prexsiting reranking tasks?

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+            document_ids.clear()
+            documents.clear()
+
+        with self._index_path.open("wb") as stream:
+            for row in corpus:
+                document_ids.append(_row_id(row))
+                documents.append(
+                    {
+                        "title": str(row.get("title", "") or ""),
+                        "text": str(row.get("text", "") or ""),
```
Samoed: Text is consist from `title + body`

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+        if model_name != MODEL_NAME or revision not in {None, MODEL_REVISION}:
+            raise ValueError(f"Unsupported model identity: {model_name}@{revision}")
+
+        from transformers import AutoModel, AutoTokenizer
+
+        source = str(model_source or MODEL_NAME)
+        source_revision = None if model_source is not None else MODEL_REVISION
+        self.device = torch.device(
+            device or ("mps" if torch.backends.mps.is_available() else "cpu")
+        )
```
Samoed: Can you add cuda?

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+                padding=True,
+                truncation=True,
+                max_length=self.max_length,
+                return_tensors="pt",
+            )
+            vectors.append(self._encode(tokens["input_ids"], tokens["attention_mask"]))
+        return np.concatenate(vectors, axis=0)
+
+    def close(self) -> None:
+        numpy = globals().get("np")
```
Samoed: You can import numpy directly
```suggestion
```

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+        num_proc: int | None,
+    ) -> None:
+        del task_metadata, hf_split, hf_subset, encode_kwargs, num_proc
+        self.close()
+        self.docids = []
+        handle = tempfile.NamedTemporaryFile(
+            prefix="structural-separator-index-", suffix=".f32", delete=False
+        )
+        handle.close()
+        self._index_path = Path(handle.name)
```
Samoed: Why are you creating tmp file and close it immediately

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,337 @@
...
+        self.close()
+
+    def index(
+        self,
+        corpus: Iterable[dict[str, Any]],
+        *,
+        task_metadata: Any,
+        hf_split: str,
+        hf_subset: str,
+        encode_kwargs: dict[str, Any],
```
Samoed:
```suggestion
        task_metadata: TaskMetadata,
        hf_split: str,
        hf_subset: str,
        encode_kwargs: EncodeKwargs,
```

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `e52a4e75`. I removed the constructor-level `batch_size`; both indexing and query encoding now read and validate `encode_kwargs["batch_size"]`, with the previous value of 64 retained only as the default. I also added focused coverage for the passed and default values.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `e52a4e75`. The explicit `model_source` parameter has been removed, and the loader now always uses the pinned immutable Hugging Face model revision.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Updated in `e52a4e75`. I used `QueryDatasetType`, which is the query type in the current `SearchProtocol`, rather than the more general `Iterable[BatchedInput]`. The corpus side now likewise uses `CorpusDatasetType`; mypy passes with these protocol-aligned types.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: It does now. In `e52a4e75`, a non-`None` `top_ranked` restricts the same normalized-vector cosine scoring to the supplied candidate IDs and returns the best `top_k` candidates. The existing full-corpus retrieval path is unchanged, and focused tests verify that results cannot escape the supplied candidate set.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Updated in `e52a4e75`. The document formatter now consumes the structured `title` and `body` fields when `body` is available, and falls back to legacy `text` otherwise. This avoids duplicating the title when MTEB has already formed `text = title + body`, while preserving the separate title/body boundaries required by this model. A focused boundary test covers this case.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `e52a4e75`. Automatic device selection is now CUDA first, then MPS, then CPU; an explicitly supplied device still takes precedence. I also ran the public checkpoint smoke successfully on the automatic MPS path.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `e52a4e75`. Cleanup now uses the imported `np` module directly. I retained a small `getattr(np, "empty", None)` guard only for interpreter shutdown, where NumPy module attributes may already have been cleared; there is a regression test for that path.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Changed in `e52a4e75`. The `NamedTemporaryFile` now remains open while encoded vectors are written, and it is closed only before opening the completed file as a read-only memmap. Error cleanup is handled around the same context, with a focused lifecycle test.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `e52a4e75`. `index()` now uses `TaskMetadata`, `EncodeKwargs`, and `CorpusDatasetType`; `search()` uses the corresponding protocol types, including `QueryDatasetType`, `TopRankedDocumentsType`, and `RetrievalOutputType`. Ruff and mypy both pass.

thu-nmrc:
Addressed all current review feedback in `e52a4e75` and replied on each inline thread. The update aligns the search adapter with the current protocol types and `encode_kwargs`, adds CUDA selection and candidate-set reranking compatibility, preserves structured `title`/`body` boundaries without duplicating the concatenated `text`, and simplifies the temporary memmap lifecycle.

Validation completed locally:
- Ruff format/check: pass
- mypy on the implementation: pass
- focused structural-separator tests: 7 passed
- model metadata regression suite: 2,252 passed
- public pinned-checkpoint full-corpus and candidate-set smoke: pass
- official `NanoSciFactRetrieval` with the now-effective `encode_kwargs` batch size: `nDCG@10 0.77211`, `MAP@100 0.74678`, `Recall@100 0.96`, zero exceptions; exactly matches the frozen reference result.

The full-corpus retrieval representation and scoring path remain unchanged; `top_ranked` support is an interface-compatibility path for existing reranking tasks.

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+        TopRankedDocumentsType,
+    )
+
+MODEL_NAME = "thu-nmrc/bge-small-structural-separator"
+MODEL_REVISION = "9a0a8aa92400202dd1ef6950ed9cd4a116dfb03d"
+BASE_MODEL_NAME = "BAAI/bge-small-en-v1.5"
+BASE_MODEL_REVISION = "5c38ec7c405ec4b44b94cc5a9bb96e735b38267a"
+REPOSITORY_URL = "https://huggingface.co/thu-nmrc/bge-small-structural-separator"
+TRAINING_REPOSITORY_URL = "https://github.com/thu-nmrc/bge-small-structural-separator"
+TRAINING_REPOSITORY_REVISION = "e8b0a9325409d791981b7410679ae8c152fd6e00"
```
Samoed: Can you move directly to meta?

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+            return str(row[key])
+    raise KeyError(f"MTEB row has no supported id field: {sorted(row)}")
+
+
+def _split_sentences(text: str) -> list[str]:
+    chunks = re.split(r"(?<=[.!?])\s+|\n+", text.strip())
+    return [chunk.strip() for chunk in chunks if chunk.strip()]
+
+
+def _document_ids(
```
Samoed: I'd rename something like
```suggestion
def _get_document_tokens(
```

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+                device = "cpu"
+        self.device = torch.device(device)
+        self.tokenizer = AutoTokenizer.from_pretrained(
+            MODEL_NAME, revision=MODEL_REVISION
+        )
+        self.backbone = (
+            AutoModel.from_pretrained(MODEL_NAME, revision=MODEL_REVISION)
+            .to(self.device)
+            .eval()
+        )
```
Samoed: You can just do 
```suggestion
        self.tokenizer = AutoTokenizer.from_pretrained(
            model_name, revision=revision
        )
        self.backbone = (
            AutoModel.from_pretrained(model_name, revision=revision)
            .to(self.device)
            .eval()
        )
```

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+        self._index_path: Path | None = None
+        self._model_meta: ModelMeta | None = None
+
+    @property
+    def mteb_model_meta(self) -> ModelMeta:
+        return self._model_meta or bge_small_structural_separator
+
+    @mteb_model_meta.setter
+    def mteb_model_meta(self, value: ModelMeta) -> None:
+        self._model_meta = value
```
Samoed: This will be set during `mteb.get_model`

KennethEnevoldsen:
Hi @Samoed, is there anything specific you want my eyes on here?

Samoed:
Just overall model implementation, because this is search wrapper and is a bit not standart

tests/test_models/test_bge_structural_separator.py
KennethEnevoldsen, line 1: I am not sure we want to maintain these tests. Would probably consider deleting them

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+    open_weights=True,
+    revision=MODEL_REVISION,
+    release_date="2026-07-14",
+    n_parameters=33_360_000,
+    n_embedding_parameters=11_720_448,
+    memory_usage_mb=None,
+    embed_dim=384,
+    license="mit",
+    max_tokens=512,
+    reference=REPOSITORY_URL,
```
KennethEnevoldsen: In general I would say that the documentation on this approach is somewhat minimal. Will be there be a paper, blogpost or similar? 

It is def. good enough to merge, but I assume the goal of the merge is to highlight the method?

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+    input_ids = torch.full((len(sequences), maximum), pad_token_id, dtype=torch.long)
+    attention_mask = torch.zeros((len(sequences), maximum), dtype=torch.long)
+    for row, sequence in enumerate(sequences):
+        input_ids[row, : len(sequence)] = torch.tensor(sequence, dtype=torch.long)
+        attention_mask[row, : len(sequence)] = 1
+    return input_ids, attention_mask
+
+
+class StructuralSeparatorSearch:
+    """One-vector exact retrieval with a learned shared structural separator."""
```
KennethEnevoldsen: Can you extend this to either minimally introduce the method or link to relevant material on the implementation.

KennethEnevoldsen: e.g. the one you have in your PR description.

mteb/models/model_implementations/bge_structural_separator.py
KennethEnevoldsen: Yes in general, model_name and revision should be passed not assumed

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+        self.backbone = (
+            AutoModel.from_pretrained(MODEL_NAME, revision=MODEL_REVISION)
+            .to(self.device)
+            .eval()
+        )
+        self.separator_token_id = SEPARATOR_TOKEN_ID
+        if self.tokenizer.convert_ids_to_tokens(SEPARATOR_TOKEN_ID) != SEPARATOR_SYMBOL:
+            raise RuntimeError("Separator token does not match the pinned tokenizer")
+
+        self.max_length = MAX_LENGTH
```
KennethEnevoldsen: I would move many of these out of the init function, e.g.:
```
class Model:
  max_length = 512

  def init... 
```

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None,
+        *,
+        device: str | None = None,
+        index_encoding_chunk_size: int = 4_096,
+        query_search_batch_size: int = 64,
+        document_search_block_size: int = 65_536,
+        **_: Any,
```
KennethEnevoldsen:
```suggestion
        **kwargs: Any,
```

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+        self, input_ids: torch.Tensor, attention_mask: torch.Tensor
+    ) -> np.ndarray:
+        with torch.inference_mode():
+            output = self.backbone(
+                input_ids=input_ids.to(self.device),
+                attention_mask=attention_mask.to(self.device),
+                return_dict=True,
+            ).last_hidden_state[:, 0]
+            return (
+                F.normalize(output, dim=1).cpu().numpy().astype(np.float32, copy=False)
```
KennethEnevoldsen: numpy or type conversion should not be required

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,409 @@
...
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        top_k: int,
+        encode_kwargs: EncodeKwargs,
+        top_ranked: TopRankedDocumentsType | None = None,
+        num_proc: int | None,
+    ) -> RetrievalOutputType:
+        del task_metadata, hf_split, hf_subset, num_proc
```
KennethEnevoldsen: I don't think this is needed

KennethEnevoldsen (commented):
I would say my primary concern is documentation. This is generally less of a concern with models like encoders that have a standardized interface, but because the Search interface is so broad I would make the reference link a bit more friendly to those that are unfamiliar.

Can you add a result comparison with BGE just to confirm that it runs as intended

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `7e67418f`. Model identity and publication fields now live directly in `ModelMeta`; fixed runtime representation values are class attributes on `StructuralSeparatorSearch`.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `7e67418f`: renamed the helper to `_get_document_tokens`.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `7e67418f`. `AutoTokenizer` and `AutoModel` now use the `model_name` and `revision` passed by the loader, with the hardcoded identity check removed.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `7e67418f`. I removed the custom property/setter and backing field. The class now declares `mteb_model_meta: ModelMeta | None = None`, and `ModelMeta.load_model()` supplies the concrete metadata during `mteb.get_model()`.

tests/test_models/test_bge_structural_separator.py
thu-nmrc: Done in `7e67418f`. I deleted the dedicated model test file as suggested and retained validation through the repository protocol/meta tests plus real-checkpoint smoke evaluation.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Addressed in `d795bff` and `7e67418f`. There is no archival paper for this release, so I added a self-contained publication-style method note covering motivation, novelty, formal retrieval definition, training, the protocol-matched BGE control, the falsification result, limitations, reproducibility, and related work: https://github.com/thu-nmrc/bge-small-structural-separator/blob/main/METHOD_CARD.md. The `ModelMeta.reference` now points directly to that note.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `7e67418f`. The class docstring now minimally defines the document representation, unchanged query path, and exact cosine matching, and links directly to the public method note for the full implementation material.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `7e67418f`. `max_length`, separator identity, default batch size, and embedding dimension are now class attributes rather than values assigned in `__init__`.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `7e67418f`: renamed the catch-all loader argument to `**kwargs: Any`.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Updated in `7e67418f`. I removed the redundant `.astype(np.float32, copy=False)`; the checkpoint output is already float32. I retained `.cpu().numpy()` because this custom `SearchProtocol` writes encoded blocks to an `np.memmap` and performs exact blockwise NumPy search, so a NumPy array is required at that storage/search boundary (the repository FAISS backend similarly converts Torch tensors before indexing). The current implementation reproduces NanoSciFact exactly: NDCG@10 0.77211, MAP@100 0.74678, Recall@100 0.96.

mteb/models/model_implementations/bge_structural_separator.py
thu-nmrc: Done in `7e67418f`. The explicit `del task_metadata, hf_split, hf_subset, num_proc` statements were removed from both protocol methods.

thu-nmrc:
Implemented the remaining review feedback.

- Model documentation commit `d795bff` publishes a self-contained method note: https://github.com/thu-nmrc/bge-small-structural-separator/blob/main/METHOD_CARD.md
- MTEB commit `7e67418f` simplifies the loader, uses the passed model identity, relies on MTEB metadata injection, moves fixed values to class attributes, removes the dedicated tests, and links `ModelMeta.reference` directly to the method note.
- The PR description now includes the requested protocol-matched plain-BGE comparison. BEIR8 macro NDCG@10 is 0.534916 versus 0.522211 (+0.012705). The preregistered six-task extension is also disclosed as negative (-0.011038), so the claim remains bounded.

Validation on the current commit: Ruff format/check and mypy pass; 2,261 relevant repository tests pass (7 optional-dependency skips); the public checkpoint loads through `mteb.get_model()`; full-corpus and supplied-candidate searches pass; and official NanoSciFact reproduces NDCG@10 0.77211, MAP@100 0.74678, Recall@100 0.96 with no evaluation exceptions.

I replied to each inline thread with its corresponding change and left the threads open for maintainer confirmation.

mteb/models/model_implementations/bge_structural_separator.py
```diff
@@ -0,0 +1,399 @@
...
+                                else row.get("text", "")
+                            ),
+                        }
+                    )
+                    if len(documents) == self.index_encoding_chunk_size:
+                        flush(stream)
+                flush(stream)
+        except Exception:
+            self.close()
+            raise
```
Samoed: I don't think that you need to store embeddings like this. We have encoder wrapper for such cases, but it won't work with search https://docs.mteb.org/get_started/advanced_usage/cache_embeddings/

thu-nmrc: Agreed. I refactored this in 59c6afc4 from the custom `SearchProtocol` implementation to the standard `EncoderProtocol`. The model now owns only document/query encoding and cosine similarity; MTEB's `SearchEncoderWrapper` owns retrieval/reranking, and `CachedEmbeddingWrapper` can own persistent embedding reuse. The temporary memmap, custom index, search, and reranking code were removed.

With the same public checkpoint, document tokens/vectors, query vectors, cosine scores, and fixture rankings are byte-identical (maximum absolute difference `0.0`). NanoSciFact and NanoArguAna reproduce every finite score field exactly, and a second NanoSciFact run completed entirely from the standard cache with the backbone disabled. Full SciFact reproduces every score field; full NFCorpus reproduces the reported NDCG@10, MAP@100, and Recall metrics exactly. I also updated the public method docs and PR description to distinguish the encoder contract from the historical evaluation blocking and tie policy.

Samoed (approved):

Issue #4932, opened 2026-07-15
Relevance scores create spurious ties in low-precision evaluation


KisuYang:
### Describe the bug

When embedding/reranking models run in **float16/bfloat16**, MTEB computes the final relevance score in that same low precision in a couple of paths. Reduced-mantissa formats coarsely bucket values in the narrow score range, so **distinct query–document scores collapse into identical values ("spurious ties")**. Ranking then falls back to arbitrary tie-breaking (index order), which inflates the variance and bias of retrieval metrics and can even flip model rankings.

The standard `SentenceTransformerEncoderWrapper` already upcasts embeddings to float32 on return, so most bi-encoder leaderboard runs are fine. Two gaps remain on current `main`:

1. **Similarity primitives** — `mteb/similarity_functions.py::_convert_to_tensor` converts *non-tensor* inputs to float32 but leaves existing tensors at their incoming dtype. An encoder that returns bf16/fp16 tensors is therefore scored in low precision by `cos_sim` / `dot_score` / `max_sim`.
2. **Cross-encoder rerankers** — `qwen3_reranker.py` and `rerankers_monot5_based.py` apply `log_softmax` directly to logits from a bf16/fp16 model. (`kalm_reranker.py` already calls `.float()` first, which is exactly the fix.)

This is analyzed in "Reliable Evaluation Protocol for Low-Precision Retrieval" (Yang et al., 2026), https://aclanthology.org/2026.acl-short.33. The remedy — **High-Precision Scoring (HPS)** — upcasts *only* the final scoring op to float32; it is value-preserving (every fp16/bf16 value is exactly representable in fp32), zero-parameter, and a no-op for already-float32 embeddings, so full-precision leaderboard numbers are unchanged.

### To reproduce

```python
import torch
qn = torch.nn.functional.normalize
q, d = torch.randn(1, 32), torch.randn(200, 32)

# bi-encoder cosine score computed in bf16 (no HPS) vs upcast-then-score (HPS)
naive = qn(q.bfloat16(), dim=1) @ qn(d.bfloat16(), dim=1).T
hps   = qn(q.bfloat16().float(), dim=1) @ qn(d.bfloat16().float(), dim=1).T
print(torch.unique(naive).numel(), torch.unique(hps).numel())  # e.g. 184 vs 200

# reranker softmax over 200 two-logit rows, bf16 vs HPS
logits = torch.randn(200, 2).bfloat16()
[...]

PR #4933, opened 2026-07-15
fix: compute relevance scores in float32 to avoid low-precision ties (HPS)

KisuYang:
## What
`mteb/similarity_functions.py::_convert_to_tensor` upcasts **float16/bfloat16** embeddings to **float32** before scoring. It already converted non-tensor inputs to float32; this extends the same guarantee to tensors that arrive in low precision.

## Why
When an encoder returns bf16/fp16 tensors, `cos_sim` / `dot_score` / `max_sim` (and the pairwise variants) computed the score in that low precision. Reduced-mantissa formats coarsely bucket values in the narrow score range, collapsing distinct relevance scores into **spurious ties**; ranking then depends on arbitrary tie-breaking, inflating retrieval-metric variance and bias. This is the **High-Precision Scoring (HPS)** remedy from "Reliable Evaluation Protocol for Low-Precision Retrieval" (Yang et al., 2026), https://aclanthology.org/2026.acl-short.33.

The upcast is **value-preserving** (every fp16/bf16 value is exactly representable in fp32) and a **no-op for embeddings that are already float32**, so full-precision leaderboard results are unchanged. It only affects models that return low-precision embeddings, and there it *reduces* noise.

## Changes
- `mteb/similarity_functions.py`: upcast fp16/bf16 tensors to fp32 in `_convert_to_tensor` (used by every similarity primitive).
- `tests/test_similarity_functions.py`: new tests — scores are fp32 for low-precision inputs; HPS collapses spurious ties (recovers ~all unique fp32 scores); fp32 inputs unchanged; numpy inputs still supported.

## Checks
- `ruff check` / `ruff format --check`: pass on changed files
- `mypy`: no new errors in the changed file
- New tests pass; `tests/test_evaluators/test_RetrievalEvaluator.py` unaffected

## Notes
Disclosure: I'm the first author of the referenced paper. Companion PR handles cross-encoder rerankers. Related issue: #4932.


KennethEnevoldsen (approved):
Thanks for the fix - tests looks reasonable good to see in included

Samoed:
@KisuYang I think you can suggest your improvements to sentense-transformers directly

KennethEnevoldsen:
@tomaarsen you might be interested in this fix as well.

PR #4958, opened 2026-07-18
fix: count symmetric STS pairs once

### LINKED ISSUE #4901: Add a normalized STS pair quality check
## Problem

#4876 cleaned STSBenchmark.v2 by using order-insensitive normalized sentence-pair keys and split-aware overlap filtering. Kenneth suggested tracking a general descriptive-statistics check separately.

The current STS descriptive stats expose `unique_pairs`, but that count is directional and per split. It can flag exact same-direction duplicate rows, but it does not by itself cover swapped `(sentence1, sentence2)` pairs or normalized pair overlaps across validation/test/train.

## Suggested scope

- Add a quality check for STS tasks that can detect duplicated normalized pairs within eval splits.
- Include split-aware checks for validation/test pairs against train and against each other where train/validation/test splits are available.
- Avoid hard-failing existing historical tasks initially; either use a known-issue list or only enforce the check for new or v2 STS tasks.
- Keep the check close to `tests/test_tasks/test_task_quality.py` or the descriptive-statistics generation path so contributors get actionable failures.

## Benefit

This would make future STS additions less likely to reintroduce the leakage pattern discussed in #4798 and #4876 without requiring maintainers to manually inspect pair keys each time.

I can help with the implementation once the preferred enforcement boundary is clear.

Ref https://github.com/embeddings-benchmark/mteb/issues/5021

GioiaZheng:
## Summary

- canonicalize text STS sentence pairs before writing the existing `unique_pairs` statistic, so `(a, b)` and `(b, a)` count as one pair;
- apply the same calculation to per-subset statistics and multilingual overall statistics, where duplicates across subsets are visible in the aggregate count;
- add focused coverage for both within-split swapped pairs and swapped pairs across multilingual subsets.

## Scope

This revision follows the review feedback to reuse `unique_pairs`. It removes the proposed `pair_overlap` field and the separate cross-split quality-check path. Train/validation/test overlap is not reported by this change because descriptive statistics remain split-based.

## Validation

- `pytest tests/test_abstasks/test_sts_statistics.py tests/test_abstasks/test_task_metadata.py::test_descriptive_statistics_mock_tasks -q --disable-warnings` — 31 passed
- `ruff check mteb/abstasks/sts.py tests/test_abstasks/test_sts_statistics.py`
- `ruff format --check mteb/abstasks/sts.py tests/test_abstasks/test_sts_statistics.py`
- `mypy mteb/abstasks/sts.py`
- `git diff --check`

Close #4901

### DIFF
--- a/mteb/abstasks/sts.py
+++ b/mteb/abstasks/sts.py
@@ -29,6 +29,19 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _canonical_text_pair_keys(
+    text1: list[Any], text2: list[Any]
+) -> set[tuple[str, str]]:
+    """Return exact text-pair keys with sentence order removed."""
+    if len(text1) != len(text2):
+        raise ValueError("STS pair columns must contain the same number of rows")
+
+    return {
+        (left, right) if left <= right else (right, left)
+        for left, right in zip(text1, text2)
+    }
+
+
 class STSMetrics(TypedDict):
     """Metrics for STS tasks
 
@@ -195,12 +208,23 @@ class AbsTaskSTS(AbsTask):
                 else self.dataset[split]
             )
         )
+
+        def _load_split_col(target_split: str, col: str) -> list[Any]:
+            if hf_subset:
+                return list(self.dataset[hf_subset][target_split][col])
+            if compute_overall:
+                result: list[Any] = []
+                for subset in self.metadata.eval_langs:
+                    result.extend(self.dataset[subset][target_split][col])
+                return result
+            return list(self.dataset[target_split][col])
+
         if hf_subset:
             score = self.dataset[hf_subset][split]["score"]
             n = len(score)
 
             def _load_col(col: str) -> list[Any]:
-                return list(self.dataset[hf_subset][split][col])
+                return _load_split_col(split, col)
 
         elif compute_overall:
             score = []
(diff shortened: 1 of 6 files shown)


mteb/abstasks/sts.py
```diff
@@ -238,7 +259,33 @@ def _load_col(col: str) -> list[Any]:
         )
         labels_statistics = calculate_score_statistics(score)
 
-        return AnySTSDescriptiveStatistics(
+        pair_overlap: dict[str, int] | None = None
```
Samoed: We already have `unique_pairs`. I think would be better to just a bit update it instead

GioiaZheng: Good point. I can make the existing STS `unique_pairs` count order-insensitive and drop the new `pair_overlap` field. That would catch swapped duplicates within a split, but not overlap across train/validation/test from #4901. Would you prefer narrowing this PR to the within-split check?

Samoed: Overall our statistics are split based. I think overlap across subsets can be computed in overall

mteb/abstasks/sts.py
```diff
@@ -238,6 +252,20 @@ def _load_col(col: str) -> list[Any]:
...
+            and isinstance(column2, str)
+            and len(self.metadata.modalities) == 1
+            and self.metadata.modalities[0] == "text"
+        ):
+            pair_stats["unique_pairs"] = len(
+                _canonical_text_pair_keys(
+                    _load_col(column1),
+                    _load_col(column2),
+                )
+            )
```
Samoed: I think better to modify here https://github.com/embeddings-benchmark/mteb/blob/be0f6231866159afc7e6ebc30313ca3dcd8287cb/mteb/abstasks/_statistics_calculation.py#L517

GioiaZheng: Moved this into `calculate_pair_modality_statistics` with an opt-in `symmetric=True` for STS, so directional pair-classification tasks keep their existing behavior. I also added a regression test for the default ordered case. All checks are green now.

isaac-chung (approved):
Seems like all of @Samoed 's comments have been addressed.

Samoed (approved):

KennethEnevoldsen (approved):

PR #4961, opened 2026-07-19
model: Add LightOn-rerank models (3 pointwise + 3 listwise)

### LINKED ISSUE #4939: Add model: Light on rerank

https://huggingface.co/collections/lightonai/lighton-rerank

### Arxiv link

_No response_

### Description of the model

Congratulations with release @coreprinciple6! Would be great to integrate your model and run on private part!

coreprinciple6:
Closes #4939

This PR adds the six LightOn-rerank models ([collection](https://huggingface.co/collections/lightonai/lighton-rerank)) — merged Qwen3.5 LoRAs that rerank both text passages and document page images against text queries:

| Model | Type | Params |
|---|---|---|
| `lightonai/LightOn-rerank-PW-0.8B` / `-2B` / `-4B` | pointwise cross-encoder | 0.85B / 2.2B / 4.5B |
| `lightonai/LightOn-rerank-LW-0.8B` / `-2B` / `-4B` | generative listwise reranker | 0.85B / 2.2B / 4.5B |

## Implementation notes (two firsts — maintainer input welcome)

1. **First listwise reranker in mteb.** The LW models generate a permutation over a window of 4 candidates; larger pools use a sliding window (window 4, stride 2, bottom→top, RankGPT-style — each window sees prior windows' reordering). Since `CrossEncoderProtocol.predict` must return one scalar per (query, doc) pair, the wrapper groups the flattened pairs by query (they arrive grouped from `SearchCrossEncoderWrapper`), reranks each group, and emits `score = n_docs − final_rank`, which reproduces the listwise ordering when sorted. Malformed generations fall back to the input order and are counted (`parse_stats`) so silent regressions are visible.
2. **First `modalities=["image","text"]` cross-encoder.** The PW wrapper subclasses `CrossEncoderWrapper` and unwraps multimodal samples into the raw payloads the LightOn sentence-transformers integration accepts (PIL image or string). The LW wrapper reads `batch["image"]` from the documents dataloader and renders image windows via the processor's chat template.

### Upstream fix included: `SearchCrossEncoderWrapper` queries dataloader prompt type

`SearchCrossEncoderWrapper.search` built the **queries** dataloader with `prompt_type=PromptType.document` (`mteb/models/search_wrappers.py`). On mixed-modality tasks (e.g. t2i `DocumentUnderstanding`), the document prompt type prepares rows with the **corpus** modality, so text queries went down the image-preparation path → `KeyError: 'image'`, and the trailing modality-column cleanup would drop the queries' `text` column. This PR changes it to `PromptType.query`, matching `SearchEncoderWrapper`. Latent until now because no cross-encoder declared image modality.

**Behavior-change note for maintainers:** the t2i `KeyError` previously crashed, so no existing numbers change on image tasks. On *text* reranking tasks, however, the query prompt type changes what existing cross-encoders receive on instruction-bearing tasks: queries get the instruction merged into `text` (same as the dense path via `_combine_queries_with_instruction_text`), and custom rerankers that read `batch["instruction"]` separately (e.g. `qwen3_reranker.py`) would see it twice — this could shift existing leaderboard numbers on those tasks. Tasks without instructions are unaffected (verified: AskUbuntuDupQuestions identical pre/post fix). **Happy to split both `search_wrappers.py` fixes into their own PR** if you prefer reviewing the blast radius separately — this PR would then depend on it.

### Second upstream fix: memory blowup on image reranking

`SearchCrossEncoderWrapper.search` materialized one copied corpus row per (query, candidate) pair via `Dataset.from_list`, decoding every image to raw pixels up front — for ViDoRe-scale tasks that's ~16× duplication × ~75× decompression ≈ 330 GB RSS and an hour-long silent stall before evaluation starts. Replaced with `dataset.select(indices)` (an indices-mapped view, duplicates allowed, order preserved): memory stays flat and images decode lazily per batch. Text behavior verified unchanged.

The PW models load through their native sentence-transformers `CrossEncoder` integration (ST ≥ 5.4.0; the bundled `reranker.jinja` template pins non-thinking generation). The LW-4B prompt prepends an empty `<think>` block per its model card (`use_think_block=True` in `loader_kwargs`); 0.8B/2B do not.

## Reproduction against published results

Blog appendix: https://huggingface.co/blog/lightonai/lighton-rerank#appendix

To compare against the published numbers exactly, evaluation reused the same first-stage candidates as the blog runs (BEIR: BM25 top-100; converted to mteb prediction format and loaded via `task.convert_to_reranking(..., top_k=100)`).

| Task | Model | Published nDCG@10 | This PR | Δ |
|---|---|---|---|---|
| SciFact (BM25 top-100) | PW-0.8B | 75.23 | 74.92 | −0.31 |
| SciFact (BM25 top-100) | LW-0.8B | 74.22 | 74.44 | +0.22 |
| ViDoRe v3 CS-EN (ColQwen2.5 top-100) | PW-0.8B | 73.96 | 73.98 | +0.02 |
| ViDoRe v3 CS-EN (ColQwen2.5 top-100) | LW-0.8B | 79.14 | 78.80 | −0.34 |

Listwise permutation parse fallbacks: SciFact 5.5% of windows, matching the reference implementation's 5.6% on the same dataset. Run-to-run variance across different GPUs/kernel selection in bf16 is ≈ ±0.1–0.2 nDCG on these tasks.

## Training data annotation

`training_datasets` = {MSMARCO, MSMARCOHardNegatives, NQ, NQHardNegatives, VidoreDocVQARetrieval, VidoreInfoVQARetrieval, VidoreTatdqaRetrieval, VidoreArxivQARetrieval} — text mix is MS MARCO + NQ; the visual-document training mix shares lineage with the colpali train set, so the ViDoRe v1 train-split overlap is annotated following `colpali_models.py`.

## Checklist
- [X ] I have filled out the ModelMeta object to the extent possible
- [ X] I have ensured that my model can be loaded using
  - [ X] `mteb.get_model(model_name, revision)` and
  - [ X] `mteb.get_model_meta(model_name, revision)`
- [X ] I have tested the implementation works on a representative set of tasks.
- [ X] The model is public, i.e., is available either as an API or the weights are publicly available to download
- [ X] I reproduced results from the original paper (if applicable) on at least one benchmark, and I am including the results in the PR description.

### DIFF
--- a/mteb/models/model_implementations/lighton_rerank_models.py
+++ b/mteb/models/model_implementations/lighton_rerank_models.py
@@ -1,0 +1,490 @@
+"""LightOn-rerank models: pointwise (PW) cross-encoders and generative listwise (LW) rerankers.
+
+- PW models ship with a sentence-transformers CrossEncoder integration and score
+  each (query, document) pair independently via logit("Yes") - logit("No").
+- LW models rank exactly ``WINDOW_SIZE`` documents per forward pass by generating a
+  permutation string (e.g. ``[2] > [1] > [4] > [3]``). Larger candidate pools are
+  reranked with a sliding window (window 4, stride 2) moving from the bottom of the
+  list to the top, so the best candidates bubble up to the front.
+Reference: https://huggingface.co/collections/lightonai/lighton-rerank
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import torch
+from tqdm.auto import tqdm
+
+from mteb.models.model_meta import ModelMeta
+from mteb.models.sentence_transformer_wrapper import CrossEncoderWrapper
+
+if TYPE_CHECKING:
+    from PIL import Image
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+
+logger = logging.getLogger(__name__)
+
+
+class LightOnPointwiseRerankerWrapper(CrossEncoderWrapper):
+    """CrossEncoderWrapper that feeds the LightOn ST integration raw payloads.
+
+    The LightOn CrossEncoder integration accepts a text string or a PIL image as
+    the document side of each pair, so multimodal samples collected by
+    ``CrossEncoderWrapper._collect_inputs`` (dicts keyed by modality) are unwrapped
+    into the raw payload: the image if present, the text otherwise.
+
+    ``predict_batch_size`` bounds the sentence-transformers prediction batch size
+    (the search wrapper does not forward ``encode_kwargs`` to ``predict``).
+    """
+
+    def __init__(
+        self,
+        model: Any,
+        revision: str | None = None,
+        device: str | None = None,
+        predict_batch_size: int = 32,
+        **kwargs: Any,
+    ) -> None:
+        self.predict_batch_size = predict_batch_size
+        super().__init__(model, revision=revision, device=device, **kwargs)
+
+    @staticmethod
+    def _unwrap(items: list[Any]) -> list[Any]:
+        unwrapped = []
+        for item in items:
+            if isinstance(item, dict):
+                if item.get("image") is not None:
+                    unwrapped.append(item["image"])
+                else:
+                    unwrapped.append(item.get("text", ""))
+            else:
+                unwrapped.append(item)
+        return unwrapped
+
+    def predict(
+        self,
+        inputs1: DataLoader[BatchedInput],
+        inputs2: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> Array:
+        queries = self._unwrap(self._collect_inputs(inputs1, self.query_prefix))
+        documents = self._unwrap(self._collect_inputs(inputs2, self.passage_prefix))
+        kwargs.setdefault("batch_size", self.predict_batch_size)
+        return self.model.predict(list(zip(queries, documents)), **kwargs)
+
+
+class LightOnListwiseRerankerWrapper:
+    """Generative listwise reranker with sliding-window inference.
+
+    ``predict`` receives the flattened (query, document) pairs produced by
+    ``SearchCrossEncoderWrapper`` (grouped per query, in candidate order), reranks
+    each query's candidate list with the sliding window, and returns one score per
+    pair (``n_docs - final_rank``) so that sorting by score reproduces the
+    listwise ranking.
+
+    Windows for all queries that sit at the same window position are batched into a
+    single ``generate()`` call (queries are independent, so this is exact), following
+    the LightOn reference implementation.
+    """
+
+    prompt_template = "<|im_start|>user\n{user}<|im_end|>\n<|im_start|>assistant\n"
+    # The 4B backbone enables thinking by default; an empty think block keeps the
+    # generation budget for the ranking permutation. The 0.8B/2B prompts omit it.
+    think_block = "<think>\n\n</think>\n\n"
+
+    def __init__(  # noqa: PLR0913
+        self,
+        model_name: str,
+        revision: str | None = None,
+        device: str | None = None,
+        *,
+        use_think_block: bool = False,
+        window_size: int = 4,
+        stride: int = 2,
+        max_new_tokens: int = 30,
+        max_length: int = 2048,
+        text_window_batch_size: int = 16,
+        image_window_batch_size: int = 8,
+        **kwargs: Any,
+    ):
+        from transformers import AutoModelForImageTextToText, AutoProcessor
+
+        self.model_name = model_name
+        self.device = device or (
+            "cuda"
+            if torch.cuda.is_available()
+            else "mps"
+            if torch.backends.mps.is_available()
+            else "cpu"
+        )
+        kwargs.setdefault("dtype", torch.bfloat16)
+        self.model = (
+            AutoModelForImageTextToText.from_pretrained(
+                model_name, revision=revision, **kwargs
+            )
+            .to(self.device)
+            .eval()
+        )
+        self.processor = AutoProcessor.from_pretrained(model_name, revision=revision)
+        self.processor.tokenizer.padding_side = "left"
+
+        self.use_think_block = use_think_block
+        self.window_size = window_size
+        self.stride = stride
+        self.max_new_tokens = max_new_tokens
+        self.max_length = max_length
+        self.text_window_batch_size = text_window_batch_size
+        self.image_window_batch_size = image_window_batch_size
+        # A malformed generation silently falls back to the input order, which drags
+        # scores toward the first-stage ranking with no error raised. Track fallbacks
+        # so silent regressions stay visible.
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,490 @@
...
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> Array:
+        queries = self._unwrap(self._collect_inputs(inputs1, self.query_prefix))
+        documents = self._unwrap(self._collect_inputs(inputs2, self.passage_prefix))
+        kwargs.setdefault("batch_size", self.predict_batch_size)
```
Samoed: batch size will be always passed to predict
```suggestion
```

mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,490 @@
...
+        stride: int = 2,
+        max_new_tokens: int = 30,
+        max_length: int = 2048,
+        text_window_batch_size: int = 16,
+        image_window_batch_size: int = 8,
+        **kwargs: Any,
+    ):
+        from transformers import AutoModelForImageTextToText, AutoProcessor
+
+        self.model_name = model_name
```
Samoed: Not used
```suggestion
```

mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,490 @@
...
+    from PIL import Image
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+
+logger = logging.getLogger(__name__)
+
+
+class LightOnPointwiseRerankerWrapper(CrossEncoderWrapper):
```
Samoed: Is this wrapper needed?

mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,490 @@
...
+        kwargs.setdefault("dtype", torch.bfloat16)
+        self.model = (
+            AutoModelForImageTextToText.from_pretrained(
+                model_name, revision=revision, **kwargs
+            )
+            .to(self.device)
+            .eval()
+        )
+        self.processor = AutoProcessor.from_pretrained(model_name, revision=revision)
+        self.processor.tokenizer.padding_side = "left"
```
Samoed: Better to set in `tokenizer_config`

mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,490 @@
...
+        revision: str | None = None,
+        device: str | None = None,
+        *,
+        use_think_block: bool = False,
+        window_size: int = 4,
+        stride: int = 2,
+        max_new_tokens: int = 30,
+        max_length: int = 2048,
+        text_window_batch_size: int = 16,
+        image_window_batch_size: int = 8,
```
Samoed:
```suggestion
```

mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,490 @@
...
+        **kwargs: Any,
+    ) -> Array:
+        queries: list[str] = [text for batch in inputs1 for text in batch["text"]]
+        documents: list[str | Image.Image] = []
+        for batch in inputs2:
+            images = batch.get("image")
+            if images is not None:
+                documents.extend(images)
+            else:
+                documents.extend(batch["text"])
```
Samoed: Images can also be passed in `input1`

mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,490 @@
...
+        group_queries: list[str] = []
+        group_docs: list[list[str | Image.Image]] = []
+        for query, document in zip(queries, documents):
+            if not group_queries or query != group_queries[-1]:
+                group_queries.append(query)
+                group_docs.append([])
+            group_docs[-1].append(document)
+
+        show_progress_bar = kwargs.get("show_progress_bar", True)
```
Samoed: Better to put as arg of function

mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,490 @@
...
+    ) -> tuple[str, list[Image.Image]]:
+        """Render a window containing at least one image via the processor's chat template."""
+        content: list[dict[str, Any]] = [
+            {
+                "type": "text",
+                "text": f"Query: {query}\n\nRank these documents from most to least relevant:\n",
+            }
+        ]
+        images = []
+        for i, doc in enumerate(docs):
```
Samoed, line 132: Doc can be a pair of text and image. Better to add check on this if you don't support that case

mteb/models/model_implementations/lighton_rerank_models.py
Samoed: Also in batch can be pairs of image and documents

mteb/models/model_implementations/lighton_rerank_models.py
coreprinciple6: You're right, not needed. I verified

mteb/models/model_implementations/lighton_rerank_models.py
coreprinciple6: Done. "padding_side": "left" is now set in tokenizer_config.json on all six HF repos, and I bumped the pinned revisions accordingly.

mteb/models/model_implementations/lighton_rerank_models.py
coreprinciple6: The models support text queries only, so I added an explicit check that raises a clear ValueError instead of the bare KeyError this would have produced.

mteb/models/model_implementations/lighton_rerank_models.py
coreprinciple6: Added the check (in predict, where documents are collected, so it covers both the prompt builder and the batched path). The models score a candidate as either a text passage or a page image.

mteb/models/model_implementations/lighton_rerank_models.py
```diff
@@ -0,0 +1,448 @@
...
+        self, windows: list[tuple[str, list[str | Image.Image]]]
+    ) -> list[list[int]]:
+        """Rank a batch of (query, window_docs) windows, halving the batch on OOM."""
+        results: list[list[int] | None] = [None] * len(windows)
+        has_images = any(
+            not isinstance(doc, str) for _, docs in windows for doc in docs
+        )
+        batch_size = (
+            self.image_window_batch_size if has_images else self.text_window_batch_size
+        )
```
Samoed: Not sure about this, because such approach overrides provided batch size

Samoed (commented):
Oerall looks good

KennethEnevoldsen (approved):
I don't think I have a lot to add here besides what is already there. 

Very happy to see the PR though.

mteb/models/model_implementations/lighton_rerank_models.py
coreprinciple6: The provided batch size actually never reaches predict because SearchCrossEncoderWrapper.search only applies encode_kwargs to the dataloaders but agreed the semantics were wrong. Changed slightly so predict now accepts batch_size and only falls back to the modality-specific defaults when it isn't provided.

Samoed (approved):

PR #4962, opened 2026-07-19
task: add MRMR reasoning-intensive multimodal retrieval (5 subsets)

### LINKED ISSUE #4940: Add dataset: MRMR

https://huggingface.co/MRMRbenchmark

### Arxiv link

https://arxiv.org/abs/2510.09510

Expert-level reasoning-intensive multimodal retrieval benchmark from ICLR 2026. 1,502 queries across 23 domains where both queries and corpus documents are interleaved image-text, including a novel contradiction-retrieval subset. A dataset added for #4842: it adds a reasoning axis to multimodal retrieval. Claimed by me in #4842.

dukesun99:
Part of the MOEB effort (#4842). Closes #4940.

Adds the MRMR benchmark ([arXiv 2510.09510](https://arxiv.org/abs/2510.09510), ICLR 2026): expert-level reasoning-intensive retrieval where both queries and corpus documents are interleaved image-text, including a contradiction-retrieval subset. Five tasks: `MRMRNegationRetrieval`, `MRMRDesignRetrieval`, `MRMRTrafficRetrieval`, `MRMRKnowledgeRetrieval`, `MRMRTheoremRetrieval`.

Data is hosted by the authors in corpus/query/qrels layout under the [MRMRbenchmark](https://huggingface.co/MRMRbenchmark) HF organization (MIT license), pinned by revision. A shared loader normalizes two quirks of the source data: the corpus mixes text-only and image documents with None in the missing modality column, so the always-present unified `vision` rendering is used as the image side and None texts become empty strings; and one subset names its qrels columns with underscores instead of hyphens. The knowledge subset contains 2 text-only queries (of 555) which are excluded to keep a uniform query schema (documented in the loader).

Dataset checklist:
- [x] Fills an existing gap in mteb: first reasoning-intensive interleaved image-text retrieval tasks
- [x] Tested that the dataset runs with the mteb package
- [x] `mteb/baseline-random-encoder` run — negation 0.0125, theorem 0.0008, design 0.0, traffic 0.0288, knowledge 0.0 (nDCG@10)
- [x] Small real model run — Qwen/Qwen3-VL-Embedding-2B: negation 0.0929, design 0.5063, traffic 0.3912, theorem 0.3811, knowledge 0.7020 (nDCG@10)
- [x] Performance is neither trivial nor random
- [x] Sizes are already evaluation-scale (hundreds of queries per subset); no downsampling needed
- [x] Paper-score reference: the MRMR paper reports best full-benchmark averages around 31 nDCG@10 with larger retrievers; our per-subset Qwen3-VL-2B numbers (9-70) are consistent with the paper's per-domain spread. Not an exact reproduction since we evaluate subsets separately


Samoed:
Can you reproduce scores from paper? You can aggregate to get original scores format

Samoed:
Can you reproduce scores from paper? They didn't evaluate qwen3-vl-embeddings

dukesun99:
> Can you reproduce scores from paper? They didn't evaluate qwen3-vl-embeddings

Work in progress. A bit complicated on this issue, as their experiment settings is not quite the same as what I adapted for. I will pin you here again once I have an update.

dukesun99:
@Samoed Right that the paper doesn't include Qwen3-VL, so I reproduced with `TIGER-Lab/VLM2Vec-Full`, which it does evaluate (Table 3). Two parts: a faithful reproduction of the paper, and the numbers this PR produces under standard mteb.

**1. Faithful reproduction.** I ran the authors' own evaluation code (`github.com/rebeccaz4/MRMR`, a fork of mteb) with `VLM2Vec-Full` on the exact data these tasks port. It reproduces Table 3 across all five tracks:

| Track | Faithful repro | Paper |
|---|---|---|
| Knowledge | 35.0 | 34.2 |
| Theorem | 2.7 | 2.65 |
| Design | 5.6 | 5.6 |
| Traffic | 18.4 | 18.3 |
| Negation | 12.0 | 11.5 |

Every track lands within half a point of the paper, so the data itself is faithful. (Theorem per-domain: Math 2.1, Physics 2.8, Engineering 2.8, Business 2.8 vs the paper's 2.1 / 2.8 / 2.8 / 2.9.)

**2. Standard mteb (this PR's tasks), VLM2Vec-Full + flash-attn + the paper's Table-6 instructions, nDCG@10 ×100:**

| Track | This PR | Paper |
|---|---|---|
| Knowledge | 35.9 | 34.2 |
| Theorem | 13.5 | 2.65 |
| Design | 8.8 | 5.6 |
| Traffic | 10.2 | 18.3 |
| Negation | 6.75 | 11.5 (Hit@1 over 4 candidates) |

Knowledge reproduces closely. The other tracks differ, and the cause is the model-side pipeline, not the data or the retrieval scoping. The authors' fork runs VLM2Vec through a custom path: query/document images resized to ≤2000px, multi-image queries fused into a single composite, and task-specific prompts, whereas mteb's standard VLM2Vec wrapper encodes differently. The retrieval matches: their reasoning tasks retrieve against the full corpus with per-domain result splitting (not per-domain corpora), so mteb's full-corpus retrieval is the right analogue. Negation is the one genuinely different setup that the paper scores as Hit@1 over each query's 4 candidates, while this PR retrieves against the full 800-doc corpus.

Samoed:
> Negation is the one genuinely different setup that the paper scores as Hit@1 over each query's 4 candidates, while this PR retrieves against the full 800-doc corpus.

Then I think you can transform this task to the reranking by providing `top_ranked` split. 

> multi-image queries fused into a single composite, and task-specific prompts, whereas mteb's standard VLM2Vec wrapper encodes differently

We also don't support multi-image queries

mteb/tasks/retrieval/eng/mrmr_retrieval.py
```diff
@@ -0,0 +1,289 @@
...
+        dataset={
+            "path": "MRMRbenchmark/negation",
+            "revision": "ee5e669c8b257790a9c66c4d9464f6d77f33636b",
+        },
+        type="Any2AnyRetrieval",
+        category="it2it",
+        modalities=["image", "text"],
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="recall_at_1",
```
Samoed:
```suggestion
        main_score="hit_rate_at_1",
```

Samoed:
How scores where changed for negation? Did you investigate performance different for other splits?

dukesun99:
> How scores where changed for negation? Did you investigate performance different for other splits?

Only Negation changed. I added a `top_ranked` split so each query reranks its four candidates and is scored by Hit@1 (`hit_rate_at_1`), matching the paper's 4-way multiple-choice protocol, instead of retrieving against the full 800-document corpus. That moves VLM2Vec-Full's Negation from 6.75 (full-corpus nDCG@10) to ~12.0 Hit@1 — in line with the paper's 11.5 — and the random-encoder baseline to ~27 (≈1/4, as expected for one gold among four candidates).

The other four subsets (Design, Traffic, Knowledge, Theorem) are unchanged. The paper scores those as full-corpus retrieval (per-domain result splitting over a shared corpus, not multiple-choice), which is exactly what those tasks already do so the reranking change doesn't touch them.

Samoed:
> The other four subsets (Design, Traffic, Knowledge, Theorem) are unchanged.

I know, but we need try to reproduce scores from paper. These subsets have pretty big margin

dukesun99:
> > The other four subsets (Design, Traffic, Knowledge, Theorem) are unchanged.
> 
> I know, but we need try to reproduce scores from paper. These subsets have pretty big margin

I think the main difference here is the model wrapper:

Authors — [`mteb/models/vlm2vec_models.py#L138-L139`](https://github.com/rebeccaz4/MRMR/blob/3b026f98f42fee39cb092bafb2e260f6d630f60d/mteb/models/vlm2vec_models.py#L138-L139):
```python
instruction = Wrapper.get_instruction(task_name, prompt_type)   # the paper's Table-6 prompt
text = f"{instruction} <|image_1|>"
```

mteb — [`vlm2vec_models.py#L124`](https://github.com/embeddings-benchmark/mteb/blob/5f517f5d/mteb/models/model_implementations/vlm2vec_models.py#L124) and [#L228](https://github.com/embeddings-benchmark/mteb/blob/5f517f5d/mteb/models/model_implementations/vlm2vec_models.py#L228):
```python
text = "<|image_1|> Represent the given image."
# query + text:
f"<|image_1|> Represent the given image with the following question: {item_text}"
```
The mteb wrapper hard-codes this generic prompt and ignores the task's instruction.
Running the authors' code on the same data can reproduce Table 3 across all five tracks. Should we modify the MTEB wrapper?

Samoed:
Hm. Can you evaluate this task on siglip or clip models then?

> Should we modify the MTEB wrapper?

I'm not sure what is expected format for this model

PR #4973, opened 2026-07-20
model: add SigLIP2 family (15 checkpoints)

### LINKED ISSUE #2301: Integrate siglip2
HF blog: https://huggingface.co/blog/siglip2
Model list: https://huggingface.co/collections/google/siglip2-67b5dcef38c175486e240107
Paper: https://arxiv.org/abs/2502.14786

shriyasudhakar:
Adds ModelMeta entries for the google/siglip2-* fixed-resolution checkpoints (base/large/so400m/giant-opt). They reuse the existing SiglipModelWrapper since these checkpoints load as SiglipModel. NaFlex variants are excluded as they need different processor handling.

Closes #2301

### Reference results: `google/siglip2-base-patch16-224` (run on Apple M4 Pro, MPS)

| Task | Main score (ndcg@10 / accuracy) | recall@1 |
|---|---|---|
| CIFAR10ZeroShot | 0.9377 (accuracy) | — |
| MSCOCOT2IRetrieval | 0.6401 | 0.4848 |
| MSCOCOI2TRetrieval | 0.5712 | 0.1310* |

Paper comparison: SigLIP2 reports COCO R@1 of 52.1 (t→i) / 68.9 (i→t) for B/16-224
on the Karpathy 5k split. mteb's MSCOCO tasks use the M-BEIR corpus construction, so
exact parity isn't expected; t→i recall@1 of 48.5 vs 52.1 is consistent with the paper
given the split difference.

*i→t recall@1 is capped at ~0.2 in this variant because each image query has ~5
relevant captions; ndcg@10 is the meaningful number there.


Samoed:
I think you can try to run it on `XM3600`, it should have same data

mteb/models/model_implementations/siglip_models.py
```diff
@@ -419,3 +419,447 @@ def encode(
...
+      eprint={2502.14786},
+      archivePrefix={arXiv},
+      primaryClass={cs.CV}
+}"""
+
+siglip2_base_patch32_256 = ModelMeta(
+    loader=SiglipModelWrapper,
+    name="google/siglip2-base-patch32-256",
+    model_type=["dense"],
+    languages=None,  # multilingual WebLI, no official language list
```
Samoed: I think you can put at least eng

mteb/models/model_implementations/siglip_models.py
```diff
@@ -419,3 +419,447 @@ def encode(
...
+    framework=["PyTorch", "Transformers", "safetensors"],
+    reference="https://huggingface.co/google/siglip2-base-patch32-256",
+    similarity_fn_name=ScoringFunction.COSINE,
+    use_instructions=False,
+    training_datasets=siglip_training_datasets,
+    citation=SIGLIP2_CITATION,
+    extra_requirements_groups=[
+        "siglip",
+        "image",
+    ],
```
Samoed: These are not needed. Siglip2 can run without sentencepiece and protobuf, `image` will be added automatically
```suggestion
```

Samoed (approved):

PR #4975, opened 2026-07-20
model: Add gve models

### LINKED ISSUE #3770: Add model: GVE-7B

https://huggingface.co/Alibaba-NLP/GVE-7B

### Arxiv link

https://arxiv.org/abs/2510.27571

### Description of the model

_No response_

shriyasudhakar:
Adds `Alibaba-NLP/GVE-{3B,7B}` (arXiv:2510.27571), general video embedders built on
Qwen2.5-VL supporting text, image, video, and composed queries. The HF repos ship a
custom `Qwen25VLForEmbedding` class whose remote code is incompatible with
transformers >= 4.56, so the wrapper loads the native
`Qwen2_5_VLForConditionalGeneration` instead (verified numerically identical).
Embeddings are the L2-normalized hidden state of an appended `<|endoftext|>` token.
The released checkpoints pool this token rather than the position shown in the model
card demo (see the reproduction comment). Video frames are sampled via
`FramesCollator` (fps=2, up to 32 frames) with the model card's 200x28x28 per-frame
pixel budget.

Closes #3770

### Results: `Alibaba-NLP/GVE-3B` (bf16)

| Evaluation data | Protocol | R@1 (this wrapper) | R@1 (paper) |
|---|---|---|---|
| UVRB MSRVTT (author's split, 1,000 JSFusion pairs) | paper: 8 uniform frames, 200 tokens/frame | **0.436** | **0.431** |
| mteb `MSRVTTT2V` (879-pair `msrvtt_ret_test1k` variant) | PR defaults: fps=2, up to 32 frames | 0.502 (ndcg@10 0.660) | — |

Row 1 is the direct paper reproduction: same data and procedure. Row 2 is the
wrapper on mteb's task with default settings. It is not paper comparable since smaller
corpus, denser frames. GVE-7B: load and text/video encoding verified (3584-dim embeddings), couldn't run task-level scores on my hardware given the ~16 GB size.


mteb/models/model_implementations/gve_models.py
```diff
@@ -0,0 +1,241 @@
...
+            revision=revision,
+            dtype=torch.bfloat16,
+            low_cpu_mem_usage=True,
+            **kwargs,
+        )
+        self.model.eval()
+        self.model.to(self.device)
+
+        self.processor = AutoProcessor.from_pretrained(
+            model_name, revision=revision, use_fast=True
```
Samoed:
```suggestion
            model_name, revision=revision
```

mteb/models/model_implementations/gve_models.py
```diff
@@ -0,0 +1,241 @@
...
+            max_video_pixels = 200 * 28 * 28
+            if isinstance(getattr(video_processor, "size", None), dict):
+                video_processor.size = {
+                    **video_processor.size,
+                    "longest_edge": max_video_pixels,
+                }
+            if hasattr(video_processor, "max_pixels"):
+                video_processor.max_pixels = max_video_pixels
+            if hasattr(video_processor, "do_sample_frames"):
+                video_processor.do_sample_frames = False
```
Samoed: This should be passed to processor directly

mteb/models/model_implementations/gve_models.py
```diff
@@ -0,0 +1,241 @@
...
+            prompts = []
+            for text, image, video in zip(texts, images, videos):
+                content = ""
+                if video is not None:
+                    content += _VIDEO_PLACEHOLDER
+                if image is not None:
+                    content += _IMAGE_PLACEHOLDER
+                if text is not None:
+                    content += text
+                prompts.append(self._build_prompt(content, instruction))
```
Samoed: I think would be better to use `apply_chat_template`
```suggestion
```

mteb/models/model_implementations/gve_models.py
```diff
@@ -0,0 +1,241 @@
...
+        self.max_length = max_length
+        self.fps = fps
+        self.max_frames = max_frames
+        self.num_frames = num_frames
+
+        self.model = Qwen2_5_VLForConditionalGeneration.from_pretrained(
+            model_name,
+            revision=revision,
+            dtype=torch.bfloat16,
+            low_cpu_mem_usage=True,
```
Samoed:
```suggestion
```

mteb/models/model_implementations/gve_models.py
```diff
@@ -0,0 +1,241 @@
...
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        show_progress_bar: bool = True,
+        **kwargs: Any,
+    ) -> Array:
+        instruction = None
+        if prompt_type != PromptType.document:
```
Samoed, line 108: Where are you found this?

shriyasudhakar:
Sorry, I've gotten busy in the past few days and haven't had a chance to resolve the comments. I should be able to by tomorrow night at the latest.

mteb/models/model_implementations/gve_models.py
shriyasudhakar: Done.

mteb/models/model_implementations/gve_models.py
shriyasudhakar: Done.

mteb/models/model_implementations/gve_models.py
shriyasudhakar: Done.

mteb/models/model_implementations/gve_models.py
shriyasudhakar: Done.

mteb/models/model_implementations/gve_models.py
shriyasudhakar: I found it in the existing gem wrapper (gme_v_models.py), which is the closest in format. The gve model card itself always embeds candidates with the plain default system prompt, so documents get no task instruction. Happy to change if there's a preferred alternative.

mteb/models/model_implementations/gve_models.py
```diff
@@ -0,0 +1,230 @@
...
+            else "mps"
+            if torch.backends.mps.is_available()
+            else "cpu"
+        )
+        self.max_length = max_length
+        self.fps = fps
+        self.max_frames = max_frames
+        self.num_frames = num_frames
+
+        self.model = Qwen2_5_VLForConditionalGeneration.from_pretrained(
```
Samoed, line 76: Shoudn't this model be used with `trust_remote_code`?

Samoed (commented):
Can you reproduce results from paper?

mteb/models/model_implementations/gve_models.py
shriyasudhakar: I tried this out, but `modeling_gve.py` imports SlidingWindowCache and it seems it crashes on import with the transformers version mteb uses. Since the custom class is a plain subclass of Qwen2_5_VLForConditionalGeneration with no extra weights, the wrapper loads the native class instead; I also tried verifying hidden_states[-1] is bit identical to what the remote code returns, and checks out for text, image, and video inputs.

shriyasudhakar:
> Can you reproduce results from paper?

Done.

Setup: I evaluated GVE-3B on the paper's UVRB MSRVTT split (1,000 JSFusion pairs from Alibaba-NLP/UVRB), using the paper's setup as well: 8 uniformly sampled frames, 200 visual tokens per frame, last-token pooling + L2 norm and query prompt in Table 5.

I'm rerunning the mteb (MSRVTTT2V) reference table, but video encoding takes a few hours with my current local setup); I'll update the description with the new numbers by tomorrow morning.
Edit: Results for this are in PR description.

Samoed:
Please, keep your results and results from paper in one table, otherwise they're hard to read. What is difference between tables? Do we have other tasks that would be fully comparable from paper?

shriyasudhakar:
> Please, keep your results and results from paper in one table, otherwise they're hard to read. What is difference between tables? Do we have other tasks that would be fully comparable from paper?

Done, merged into single table with the differences. Row 1 is the paper's UVRB split using the paper's procedure (directly comparable, 0.436 vs 0.431). Row 2 is mteb's MSRVTTT2V variant under the wrapper defaults, not comparable to paper since the corpus is smaller and frame sampling denser.

On other fully comparable tasks, we don't have any currently. MomentSeeker exists in mteb but with a different corpus construction (30s chunked windows with graded qrels vs the paper's 10 candidate scored with P@1), and mteb's DiDeMo is caption level while the paper uses concatenated paragraph queries. I noticed that LoVR and MSRVTT-I2V aren't merged yet, so maybe we can consider those. We can run GVE on those once they are merged.

Samoed (approved):

PR #4982, opened 2026-07-20
[MOEB] Add VELOCITI video pair classification task (#4980)

### LINKED ISSUE #4980: Add dataset: VELOCITI

https://huggingface.co/datasets/katha-ai-iiith/VELOCITI

### Arxiv link

https://arxiv.org/abs/2406.10889

Compositional video-language entailment benchmark (CVPR 2025) testing whether models correctly bind agents to actions across multiple events in a video. 8,792 samples, video segments paired with a positive and negative caption (pos/neg), built via two negative-generation strategies: in-video negation (swap agents/actions occurring in the same video) and text-inspired negation (LLM-generated contradictions). License: CC-BY-NC-SA-4.0 (explicitly a test-only benchmark — fits our eval-only use case).

Filling the "video pair classification" gap under MOEB Track 1 (unclaimed, thin direction listed in #4842). Plan: AbsTaskPairClassification with pos/neg captions as the two label classes per video.

Gated dataset — access already granted to my account.

Claimed in #4842 (comment) — cc @AdnanElAssadi56 @Samoed

yaswanth169:
Add VELOCITIPairClassification (v2t) for the VELOCITI benchmark (CVPR 2025) — filling the "video pair classification" gap under MOEB Track 1 (#4842).

VELOCITI tests fine-grained compositional binding of agents to actions across events in a video, via two negative-caption strategies: in-video negation (swap agents/actions occurring in the same video) and text-inspired negation (LLM-generated contradictions).

The raw dataset (katha-ai-iiith/VELOCITI, gated) is not in a directly usable shape for AbsTaskPairClassification — it stores (video_id, pos, neg) per row rather than (video, text, label). I reshaped it into 17,584 rows (video, text, label), one row per pos/neg caption per original example, and repackaged it to yaswanth169/VELOCITI-PC.

Baseline (mteb/baseline-random-encoder) results attached — max_ap 0.504, max_accuracy 0.512, both at chance level as expected for a random encoder on a balanced binary task.

Closes #4980

### DIFF
--- a/mteb/tasks/pair_classification/eng/velociti.py
+++ b/mteb/tasks/pair_classification/eng/velociti.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+from mteb.abstasks import AbsTaskPairClassification
+from mteb.abstasks.task_metadata import TaskMetadata
+
+
+class VELOCITIPairClassification(AbsTaskPairClassification):
+    metadata = TaskMetadata(
+        name="VELOCITIPairClassification",
+        description=(
+            "Pair classification on VELOCITI: determining whether a text caption "
+            "correctly entails the events in a video or is a compositional negative "
+            "constructed via in-video negation (swapping agents/actions occurring in "
+            "the same video) or text-inspired negation (LLM-generated contradictions). "
+            "Tests fine-grained video-language compositional binding of agents to actions "
+            "across events."
+        ),
+        reference="https://arxiv.org/abs/2406.10889",
+        dataset={
+            "path": "yaswanth169/VELOCITI-PC",
+            "revision": "509144fa4727a3f9b92602d491725154c32a958f",
+        },
+        type="VideoPairClassification",
+        category="v2t",
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="max_ap",
+        date=("2024-06-16", "2024-06-16"),
+        domains=["Activity", "Web"],
+        task_subtypes=["Caption Pairing"],
+        license="cc-by-nc-sa-4.0",
+        annotations_creators="LM-generated and reviewed",
+        dialect=[],
+        modalities=["video", "text"],
+        sample_creation="LM-generated and verified",
+        is_beta=True,
+        bibtex_citation=r"""
+@inproceedings{saravanan2025velociti,
+  author = {Saravanan, Darshana and Gupta, Varun and Singh, Darshan and Khan, Zeeshan and Gandhi, Vineet and Tapaswi, Makarand},
+  booktitle = {Proceedings of the IEEE/CVF Conference on Computer Vision and Pattern Recognition},
+  title = {VELOCITI: Benchmarking Video-Language Compositional Reasoning with Strict Entailment},
+  year = {2025},
+}
+""",
+    )
+
+    input1_column_name = {"video": "video"}
+    input2_column_name = {"text": "text"}
+    label_column_name: str = "label"
--- a/mteb/tasks/pair_classification/eng/velociti.py
+++ b/mteb/tasks/pair_classification/eng/velociti.py
@@ -1,7 +1,12 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING, ClassVar
+
 from mteb.abstasks import AbsTaskPairClassification
 from mteb.abstasks.task_metadata import TaskMetadata
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
 
 
 class VELOCITIPairClassification(AbsTaskPairClassification):
@@ -44,6 +49,6 @@ class VELOCITIPairClassification(AbsTaskPairClassification):
 """,
     )
 
-    input1_column_name = {"video": "video"}
-    input2_column_name = {"text": "text"}
+    input1_column_name: ClassVar[Mapping[str, str]] = {"video": "video"}
+    input2_column_name: ClassVar[Mapping[str, str]] = {"text": "text"}
     label_column_name: str = "label"
--- a/mteb/tasks/pair_classification/eng/velociti.py
+++ b/mteb/tasks/pair_classification/eng/velociti.py
@@ -1,6 +1,12 @@
 from __future__ import annotations
 
+import tempfile
+import zipfile
+from pathlib import Path
 from typing import TYPE_CHECKING, ClassVar
+
+from datasets import Dataset, load_dataset
+from huggingface_hub import hf_hub_download
 
 from mteb.abstasks import AbsTaskPairClassification
 from mteb.abstasks.task_metadata import TaskMetadata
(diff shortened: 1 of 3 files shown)


yaswanth169:
@AdnanElAssadi56  and @Samoed can you look into this one

AdnanElAssadi:
can you also give us the original source/ how you built this?

yaswanth169:
@AdnanElAssadi56 sure, here's the exact process:

**Source**: katha-ai-iiith/VELOCITI (gated dataset) : https://huggingface.co/datasets/katha-ai-iiith/VELOCITI
- The parquet data (8,792 rows) has columns: video_id, test_name, event, pos, neg, in_subset : no native video column, video_id is just a filename string
- The actual video files are in a separate velociti_videos.zip in the same repo (906 mp4s), where video_id maps directly to the filename inside (e.g. video_id "v_x_seg_125_135.mp4" → "velociti_videos/v_x_seg_125_135.mp4")

**Build process**:
1. Downloaded and extracted velociti_videos.zip
2. Verified all 864 unique video_ids referenced by the 8,792 rows exist in the extracted files (confirmed, no missing files)
3. Reshaped each row into two: (video, pos, label=1) and (video, neg, label=0) → 17,584 rows total
4. Built as a datasets.Dataset with a proper Video() feature type, video values as {"path": local_path, "bytes": None}
5. Pushed to yaswanth169/VELOCITI-PC as a DatasetDict with a "test" split

Repackaged dataset: https://huggingface.co/datasets/yaswanth169/VELOCITI-PC

**One thing to flag proactively**: the pushed dataset ended up larger than it should be (~32GB) : when saving/pushing, the Video() feature materialized full video bytes into every row rather than referencing the 864 unique files by pointer, so videos got duplicated ~20x across rows that share the same video. It works correctly for eval (confirmed via baseline run), but it's inefficient. Happy to rebuild it referencing shared files instead if that's a blocker for merging : should bring it down to closer to 1-2GB.

yaswanth169:
Hi @AdnanElAssadi56 @Samoed , checking in on this one. All checks are passing and there are no merge conflicts. I flagged above that the pushed dataset is larger than it needs to be (~32GB due to video duplication across rows) and offered to rebuild it referencing the 864 unique files by pointer instead, which should bring it down to ~1-2GB — just let me know if that's needed before merging, or if the current version is fine as-is. Happy to make whatever changes would help this move forward.

mteb/tasks/pair_classification/eng/velociti.py
```diff
@@ -0,0 +1,54 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+
+class VELOCITIPairClassification(AbsTaskPairClassification):
+    metadata = TaskMetadata(
+        name="VELOCITIPairClassification",
+        description=(
+            "Pair classification on VELOCITI: determining whether a text caption "
```
KennethEnevoldsen: > Pair classification on VELOCITI

The reader doesn't know what velociti is.

In general, descriptions should describe what the task measure, how it measures it and then additional details. (most important to least important)

mteb/tasks/pair_classification/eng/velociti.py
```diff
@@ -0,0 +1,54 @@
...
+        main_score="max_ap",
+        date=("2024-06-16", "2024-06-16"),
+        domains=["Activity", "Web"],
+        task_subtypes=["Caption Pairing"],
+        license="cc-by-nc-sa-4.0",
+        annotations_creators="LM-generated and reviewed",
+        dialect=[],
+        modalities=["video", "text"],
+        sample_creation="LM-generated and verified",
+        is_beta=True,
```
KennethEnevoldsen, line 44: @Samoed how do we handle is_beta datasets in the leaderboard (I assume right now they are showed equal to everything else?)

We should probably either warn or not include them

KennethEnevoldsen (approved):
A few concerns on dataset quality. 

Did you run any models on this -- does it differentiate meaningfully different models? I suspect that it might not

32 gb download for an eval def. seem like a blocker to me

mteb/tasks/pair_classification/eng/velociti.py
```diff
@@ -0,0 +1,54 @@
...
+            "Pair classification on VELOCITI: determining whether a text caption "
+            "correctly entails the events in a video or is a compositional negative "
+            "constructed via in-video negation (swapping agents/actions occurring in "
+            "the same video) or text-inspired negation (LLM-generated contradictions). "
+            "Tests fine-grained video-language compositional binding of agents to actions "
+            "across events."
+        ),
+        reference="https://arxiv.org/abs/2406.10889",
+        dataset={
+            "path": "yaswanth169/VELOCITI-PC",
```
KennethEnevoldsen, line 28: needs to be moved to MTEB

mteb/tasks/pair_classification/eng/velociti.py
Samoed: > (I assume right now they are showed equal to everything else?)

Yes, they're all showning on LB

yaswanth169:
@KennethEnevoldsen @Samoed thanks for the feedback, here's where I landed on both points.

Size — fixed. Turns out the original push was materializing full video bytes into every one of the 17,584 rows instead of referencing the 864 unique videos, so they were getting duplicated ~20x. Rebuilt the dataset so it stores just a video_id string plus a single videos.zip of the unique files, and load_data() downloads/extracts that once. Repo's down to 1.6GB from ~32GB now. Also rewrote the description so it doesn't assume the reader already knows what VELOCITI is.

Model differentiation — this one's less clean. I ran X-CLIP-base and X-CLIP-large on 60 pos/neg pairs each to check paired accuracy (does the correct caption score higher than the wrong one for the same video):

- xclip-base-patch32: 56.7%
- xclip-large-patch14: 50.0% — exactly chance

Bigger didn't do better than smaller, and neither is far from chance, so honestly I can't say this proves the task differentiates models well. Two ways to read it: either it lines up with what VELOCITI's own paper found (most existing video-language models score near chance on its hardest compositional negatives, that's kind of the point of the benchmark), or X-CLIP's coarse frame sampling just isn't a fair test of this kind of temporal binding regardless of the task being fine. I don't have a strong opinion on which it is. Happy to try a model that's actually built for retrieval-style similarity (ImageBind or similar) if that would help settle it, or if you'd rather just make the call on whether X-CLIP is representative enough as-is, that works too.

yaswanth169:
One more thing  I haven't moved the dataset to the mteb org yet since my account doesn't have write access there. Could one of you either move/duplicate yaswanth169/VELOCITI-PC to mteb/VELOCITI-PC on your end, or add me to the mteb org so I can push it myself? Whichever's easier on your side works for me.

mteb/tasks/pair_classification/eng/velociti.py
```diff
@@ -0,0 +1,102 @@
...
+            revision=self.metadata.dataset["revision"],
+            split="test",
+        )
+
+        zip_path = hf_hub_download(
+            repo_id=self.metadata.dataset["path"],
+            filename="videos.zip",
+            repo_type="dataset",
+            revision=self.metadata.dataset["revision"],
+        )
```
Samoed: I don't think that storing data in zip is good idea. You can create parquet file with videos instead if you want to store them seperatly. This will simplify usage of this dataset

yaswanth169:
Hello @Samoed , can you check now

mteb/tasks/pair_classification/eng/velociti.py
```diff
@@ -0,0 +1,80 @@
...
+            return
+        path = self.metadata.dataset["path"]
+        revision = self.metadata.dataset["revision"]
+
+        # (video_id, text, label), 17,584 rows -- one per pos/neg caption.
+        # force_redownload: this small table is occasionally re-pushed under the
+        # same file path with different content (e.g. dedup fixes), and datasets'
+        # local cache can otherwise silently keep serving an older snapshot.
+        raw = load_dataset(
+            path, revision=revision, split="test", download_mode="force_redownload"
```
Samoed:
```suggestion
            path, revision=revision, split="test",
```

Samoed (approved):

mteb/tasks/pair_classification/eng/velociti.py
```diff
@@ -0,0 +1,90 @@
...
+        # one HF cache dir -- a concurrent load from another worker can race
+        # the rebuild. hf_hub_download is a single atomic, revision-pinned
+        # file fetch with no cache-build step to race.
+        parquet_path = hf_hub_download(
+            repo_id=path,
+            filename="data/test-00000-of-00001.parquet",
+            repo_type="dataset",
+            revision=revision,
+        )
+        raw = Dataset.from_pandas(pd.read_parquet(parquet_path), preserve_index=False)
```
Samoed: Why not just `load_datset` here

yaswanth169:
Hello @Samoed , Done

PR #5004, opened 2026-07-23
task: add uCO3D object-instance retrieval (i2v, v2v, t2v)

dukesun99:
Part of the MOEB effort (#4842), filling the image ↔ video direction. Self-contained: adds the `i2v` TaskCategory code.

Adds three object-instance retrieval tasks built from uCO3D ([UnCommon Objects in 3D](https://arxiv.org/abs/2501.07574), CVPR 2025, CC-BY-4.0), a large collection of 360° turntable videos of individual objects. Each orbit is split into a viewpoint-disjoint first half and second half; the corpus is the second-half clip and the query comes from the first half, so matching requires viewpoint-invariant instance recognition rather than exact-view matching. 500 object instances sampled with a fixed seed (same-category instances are hard negatives); videos re-encoded to 360p. Data at [dukesun99/uCO3D-{I2V,V2V,T2V}](https://huggingface.co/datasets/dukesun99/uCO3D-I2V).

- `UCO3DI2VRetrieval` (**i2v**): query = a single frame → retrieve the same instance's video clip
- `UCO3DV2VRetrieval` (**v2v**): query = first-half clip → second-half clip
- `UCO3DT2VRetrieval` (**t2v**): query = short caption → the described instance's clip

Dataset checklist:
- [x] Fills a gap: first image↔video (i2v) instance-retrieval task; adds `i2v` category code
- [x] Tested that the dataset runs with the mteb package
- [x] `mteb/baseline-random-encoder`: i2v 1.03 / v2v 0.96 / t2v 1.03 nDCG@10
- [x] Real model — Qwen/Qwen3-VL-Embedding-2B: **i2v 72.6 / v2v 89.8 / t2v 44.5** nDCG@10. 
- [x] Downsampled to evaluation scale (500 instances)
- [x] Paper-score reproduction: N/A — uCO3D is a 3D-reconstruction dataset with no retrieval benchmark; the instance-retrieval formulation is an adaptation

### DIFF
--- a/mteb/abstasks/task_metadata.py
+++ b/mteb/abstasks/task_metadata.py
@@ -278,6 +278,7 @@ TaskCategory = Literal[
     "va2va",
     "v2a",
     "a2v",
+    "i2v",
     "vt2a",
     "at2v",
     "a2i",
--- a/mteb/descriptive_stats/Image/Any2AnyRetrieval/UCO3DI2VRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyRetrieval/UCO3DI2VRetrieval.json
@@ -1,0 +1,83 @@
+{
+    "test": {
+        "num_samples": 1000,
+        "num_queries": 500,
+        "num_documents": 500,
+        "number_of_characters": 0,
+        "documents_text_statistics": null,
+        "documents_image_statistics": null,
+        "documents_audio_statistics": null,
+        "documents_video_statistics": {
+            "total_duration_seconds": 6913.861789698899,
+            "total_frames": 222109,
+            "min_width": 162,
+            "average_width": 403.072,
+            "max_width": 800,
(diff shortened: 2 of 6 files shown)


Samoed:
> UCO3DI2VRetrieval (i2v): query = a single frame → retrieve the same instance's video clip

I don't think we want that. What is format of the original dataset?

isaac-chung:
Thanks for the PR. It seems like this is a dataset for training. Also from the [paper](https://arxiv.org/pdf/2501.07574), the authors eval their models on something else. Perhaps we could add those instead? (Sec 4.1)

> We evaluate each trained model in a novel-view synthesis setting on two small-scale highquality object-centric datasets: OmniObject3D [68] and Stanford-ORB [34].

Closing this now. But feel free to work on either OmniObject3D or Stanford-ORB.

PR #5007, opened 2026-07-23
Fix SIB200Classification.v2 dataset source and language configs

TianyouXie828:
## Summary

This PR fixes the dataset configuration for `SIB200Classification.v2`.

## Changes

- Change the dataset source from `mteb/sib200.v2` to `Davlan/sib200_14classes`.
- Add the missing language-script configurations:
  - `ace_Arab`
  - `arb_Arab`
  - `bjn_Arab`
  - `kas_Arab`
  - `knc_Arab`
  - `min_Arab`
  - `taq_Latn`
  - `zho_Hans`


Samoed:
Will existing scores change? Can you recompute descriptive statistics with udpated splits `task.calculate_descriptive_statistics(overwrite_results=True)`?

TianyouXie828:
Yes, the aggregate task score can change because this fix expands the task from 197 to 205 language-script configurations. The eight newly included configurations are:

- ace_Arab
- arb_Arab
- bjn_Arab
- kas_Arab
- knc_Arab
- min_Arab
- taq_Latn
- zho_Hans

I have recomputed the descriptive statistics using task.calculate_descriptive_statistics(overwrite_results=True) and pushed the updated descriptive statistics for all 205 configurations.

The change is intended to make SIB200Classification.v2 use the complete 205-language dataset. If this requires a new task version under the MTEB versioning policy, I can introduce SIB200Classification.v3 instead.

Samoed:
I think would be better to create v3 version. CC @KennethEnevoldsen

KennethEnevoldsen:
Hmm, yeah I see. Here is the pros and cons:

Pro:
Updating v2 would allow us to keep all the existing results without requiring us to update any results. This can also partly be achieved by duplicating the results files in the results repo, but future run will not cover this.

Cons:
For someone who has plain `SIB200Classification.v2` in a benchmark (can be internal in a company) they will suddenly see their results change without having done anything.

My solution:
Update to v3, duplicate the files on the results repository to have coverage, update existing benchmarks to use SIB200Classification.v3, but limit the languages to only the old subset to ensure reproducibility. This ensures that we don't need to constantly run both.

slightly more complex but I think it saved us evaluation time - WDYT?

TianyouXie828:
That sounds good to me. Should I also handle duplicating the existing result files, or will that be done separately?

Samoed:
Would be great if you open pr for that

PR #5010, opened 2026-07-24
model: add webAI ColVec1.1 4B and 8B models

zhanlunchang-webai:
## Summary

- Add `ColVec11Wrapper` for the retrieval-only ColVec1.1 processor API.
- Register `webAI-Official/webAI-ColVec1.1-4b`.
- Register `webAI-Official/webAI-ColVec1.1-8b`.
- Use SDPA as the default attention implementation.
- Require Transformers 5.14.1 or newer through the `colvec1_1` dependency group.
- Use FP32 MaxSim scoring through `processor.score_retrieval()`.
- Prefer page images over auxiliary OCR Markdown for mixed corpus rows.

## Model revisions

- 4B: `b4f97a769bb1ac1de3a35308783ddd0d006d0394`
- 8B: `5a6eef07351b1c1653aa82e2e4c2f84f45a0d13d`

## Validation

- Tested `mteb.get_model_meta()` for both models.
- Tested `mteb.get_model()` for both models.
- Tested query and image encoding.
- Tested `ColVec11Wrapper.similarity()`.
- Tested image precedence for mixed image-plus-OCR corpus rows.
- Confirmed that the models run with SDPA without requiring `flash_attn`.

Benchmark results generated through the finalized SDPA registry configuration will be submitted separately after this model registration is merged.

## Model checklist

- [x] I have filled out the ModelMeta object to the extent possible.
- [x] I have ensured that my model can be loaded using:
  - [x] `mteb.get_model(model_name, revision)`
- [x] I have tested the implementation on a representative set of tasks.
- [x] The model weights are publicly available.
- [x] I reproduced the reported results on at least one benchmark. The result files will be submitted separately after this model registration is merged.

### DIFF
--- a/mteb/models/model_implementations/webai_models.py
+++ b/mteb/models/model_implementations/webai_models.py
@@ -376,8 +376,7 @@ colvec1_1_4b = ModelMeta(
     max_tokens=262144,
     embed_dim=640,
     license=(
-        "https://huggingface.co/webAI-Official/"
-        "webAI-ColVec1.1-4b/blob/main/LICENSE.md"
+        "https://huggingface.co/webAI-Official/webAI-ColVec1.1-4b/blob/main/LICENSE.md"
     ),
     open_weights=True,
     public_training_code=None,
@@ -392,7 +391,7 @@ colvec1_1_4b = ModelMeta(
     citation=COLVEC1_1_CITATION,
     contacts=["zhanlunchang-webai"],
     output_dtypes=None,
-    extra_requirements_groups=["flash_attention"],
+    extra_requirements_groups=["flash_attention", "colvec1_1"],
 )
 
 
@@ -425,8 +424,7 @@ colvec1_1_8b = ModelMeta(
     max_tokens=262144,
     embed_dim=640,
     license=(
-        "https://huggingface.co/webAI-Official/"
-        "webAI-ColVec1.1-8b/blob/main/LICENSE.md"
+        "https://huggingface.co/webAI-Official/webAI-ColVec1.1-8b/blob/main/LICENSE.md"
     ),
     open_weights=True,
     public_training_code=None,
@@ -441,5 +439,5 @@ colvec1_1_8b = ModelMeta(
     citation=COLVEC1_1_CITATION,
     contacts=["zhanlunchang-webai"],
     output_dtypes=None,
-    extra_requirements_groups=["flash_attention"],
+    extra_requirements_groups=["flash_attention", "colvec1_1"],
 )
--- a/mteb/models/model_implementations/webai_models.py
+++ b/mteb/models/model_implementations/webai_models.py
@@ -139,6 +139,97 @@ class ColVec1Wrapper(AbsEncoder):
         return self.processor.score_multi_vector(a, b, device=self.device)
 
 
+class ColVec11Wrapper(ColVec1Wrapper):
+    """
+    MTEB wrapper for ColVec1.1 (ColQwen3.5-based) retrieval models.
+
+    ColVec1.1 uses the retrieval-only processor API and allows model and
+    processor keyword arguments to be configured independently.
+    """
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None = None,
+        device: str | None = None,
+        trust_remote_code: bool = True,
+        torch_dtype: torch.dtype | None = torch.bfloat16,
+        processor_kwargs: dict[str, Any] | None = None,
+        **model_kwargs,
+    ):
+        requires_image_dependencies()
+        self.device = device or ("cuda" if torch.cuda.is_available() else "cpu")
+        processor_kwargs = dict(processor_kwargs or {})
+
+        self.model = AutoModel.from_pretrained(
+            model_name,
+            revision=revision,
+            dtype=torch_dtype,
+            trust_remote_code=trust_remote_code,
+            **model_kwargs,
+        ).to(self.device)
+        self.model.eval()
+
+        self.processor = AutoProcessor.from_pretrained(
+            model_name,
+            revision=revision,
+            trust_remote_code=trust_remote_code,
+            **processor_kwargs,
+        )
+
+    def get_image_embeddings(
+        self, images, batch_size=32, show_progress_bar=True, **kwargs
+    ):
+        import torchvision.transforms.functional as F
+        from PIL import Image
+
+        all_embeds = []
+        with torch.no_grad():
+            for batch in tqdm(
+                images, disable=not show_progress_bar, desc="Encoding images"
+            ):
+                imgs = [
+                    F.to_pil_image(b) if not isinstance(b, Image.Image) else b
+                    for b in batch["image"]
+                ]
+                imgs = [img.convert("RGB") for img in imgs]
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/webai_models.py
```diff
@@ -139,6 +139,97 @@ def similarity(self, a, b):
...
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> Array:
+        # ViDoRe v3.1 corpora expose both the page image and OCR markdown.
+        # ColVec1.1 encodes documents from images, so images must take
+        # precedence and corpus OCR text must not be encoded as a query.
+        if "image" in inputs.dataset.features:
+            return self.get_image_embeddings(inputs, **kwargs)
+        if "text" in inputs.dataset.features:
+            return self.get_text_embeddings(inputs, **kwargs)
```
Samoed: Why do you not use both image and text for vidore3.1?

mteb/models/model_implementations/webai_models.py
```diff
@@ -226,3 +345,99 @@ def similarity(self, a, b):
     contacts=["psam-ai"],
     output_dtypes=OutputDType.FLOAT16,
 )
+
+
+colvec1_1_4b = ModelMeta(
+    loader=ColVec11Wrapper,
+    loader_kwargs={
+        "torch_dtype": torch.bfloat16,
+        "attn_implementation": "flash_attention_2",
```
Samoed: Is it possible to run without flash_attn?

mteb/models/model_implementations/webai_models.py
```diff
@@ -226,3 +345,99 @@ def similarity(self, a, b):
...
+        "fra-Latn",
+        "deu-Latn",
+        "spa-Latn",
+        "ita-Latn",
+        "por-Latn",
+    ],
+    modalities=["image", "text"],
+    n_parameters=8395317104,
+    n_embedding_parameters=2622080,
+    n_active_parameters_override=None,
```
Samoed:
```suggestion
```

mteb/models/model_implementations/webai_models.py
zhanlunchang-webai: ColVec1.1 was trained with natural-language text as queries and rendered page images as documents. ViDoRe v3.1 corpus rows also contain OCR Markdown, but the model does not have a trained path for combining OCR text with image embeddings. Passing the Markdown through process_queries() would incorrectly apply the query encoding path to document text, while combining both embeddings would introduce an untrained representation. Therefore, when both fields are present, the wrapper intentionally uses the page image. I can clarify this in the code comment.

mteb/models/model_implementations/webai_models.py
zhanlunchang-webai: Thanks, agreed. I’ll remove it from both new ColVec1.1 ModelMeta entries for consistency.

mteb/models/model_implementations/webai_models.py
zhanlunchang-webai: Yes, the underlying model supports SDPA through attn_implementation="sdpa". For this MTEB registration, we prefer to keep FlashAttention 2 as the default because it was used for all validation and for the archived ViDoRe results that will be submitted after this PR merges. Keeping FlashAttention 2 ensures that the registered configuration matches the evaluated configuration. Users loading the model directly can select SDPA when FlashAttention is unavailable. However, we are open to adjusting the registration if an optional backend or SDPA default is preferred for MTEB.

mteb/models/model_implementations/webai_models.py
zhanlunchang-webai: Clarified the text-query/page-image modality behavior in 2546ed9c.

mteb/models/model_implementations/webai_models.py
zhanlunchang-webai: Removed the redundant n_active_parameters_override=None field from both new ColVec1.1 entries in 2546ed9c.

mteb/models/model_implementations/webai_models.py
Samoed: How different scores for flashattention vs sdpa? Personally I always have different problems when I try to run with flash attention

zhanlunchang-webai: Thanks for raising this. I have not evaluated either model with SDPA, so I cannot provide a reliable score comparison. The archived results were produced with FlashAttention 2.8.3 in a pinned Python 3.12, PyTorch 2.9.0, and CUDA 12.8 environment.

FlashAttention worked reliably in that environment. I can provide exact installation instructions, including the compatible PyTorch CUDA wheel and the matching flash_attn, causal_conv1d, flash_linear_attention, and TileLang packages.

What would you prefer for this registration? Should we keep the tested FlashAttention 2 configuration and document the installation clearly, or would you prefer another configuration that does not require flash_attn?

Samoed: I don't think we want to document flash attn installation, but I'd prefer to make possible to run model without flash attention

zhanlunchang-webai: Thanks. I reran the 8B model with SDPA using the same pinned revision and evaluation configuration. The results are very close to the FlashAttention 2 run:

| Domain | FA2 | SDPA | Difference |
| :--- | ---: | ---: | ---: |
| Computer Science | 80.152 | 80.081 | -0.071 |
| Energy | 70.070 | 70.118 | +0.048 |
| FinanceEn | 71.895 | 71.900 | +0.005 |
| FinanceFr | 55.094 | 54.865 | -0.229 |
| HR | 68.526 | 68.555 | +0.029 |
| Industrial | 57.465 | 57.651 | +0.186 |
| Pharmaceuticals | 67.881 | 67.882 | +0.001 |
| Physics | 51.276 | 51.501 | +0.225 |
| **Average** | **65.295** | **65.319** | **+0.024** |

The average changes by only 0.024 points, and the largest task difference is 0.229 points. Based on this result, I agree that the registration should not require FlashAttention.

I can remove the `flash_attention` requirement and use SDPA explicitly for the registered models. Would you prefer an explicit `attn_implementation="sdpa"` setting for reproducibility, or should I omit the setting and rely on the Transformers default?

One detail is that this SDPA run still used `causal_conv1d` and `flash_linear_attention` for Qwen3.5's GatedDeltaNet layers. Without those optional packages, Transformers can use its fallback implementation, but I have not measured its score difference and expect it primarily to affect speed and memory usage.

For the later results PR, would you prefer results generated with SDPA so they match the new registry default? If so, I will also run the 4B model with SDPA before submitting both result sets.

Samoed: I think we could have with any attention support, but preferrebly with SDPA

zhanlunchang-webai: Thanks. I updated both ColVec1.1 registrations to use explicit `attn_implementation="sdpa"` and removed the `flash_attention` requirement in commit d56a414.

I am now evaluating both the 4B and 8B models through the updated MTEB registry on 8×H200. The later results submission will use these registry-generated SDPA results so that the submitted results match the registered configuration.

Samoed (approved):

PR #5011, opened 2026-07-24
model: Add models multi-modal-embed [MOEB]

wissam-sib:
Adding multi modal embedding models
https://huggingface.co/llm-semantic-router/multi-modal-embed-small and large

Was able to successfully evaluated on both macos and linux, gpu or cpu.

Results on text, image and audio tasks for `multi-modal-embed-small`
```
Task         | Modality | Split | Main score | Metric 
-------------+----------+-------+------------+----------
STSBenchmark | text     | test  | 0.7521     | spearman 
CIFAR10      | image    | test  | 0.6620     | accuracy  
FSDD         | audio    | test  | 0.6197     | accuracy 
```

Results on text, image and audio tasks for `multi-modal-embed-large`
```
Task         | Modality | Split | Main score | Metric 
-------------+----------+-------+------------+----------
STSBenchmark | text     | test  | 0.7408     | spearman
CIFAR10      | image    | test  | 0.8320     | accuracy
FSDD         | audio    | test  | 0.5967     | accuracy 
```


wissam-sib:
@Samoed should the uv lock file be updated? Not 100% sure mine was updating properly locally in macos.

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,571 @@
...
+        self,
+        inputs: DataLoader[TextInput],
+        show_progress_bar: bool = True,
+        **kwargs: Any,
+    ) -> Array:
+        embeddings = []
+        for batch in tqdm(inputs, disable=not show_progress_bar, desc="Encoding text"):
+            with torch.inference_mode():
+                emb = self.model.encode_text(batch["text"])
+            embeddings.append(emb.detach().cpu().float().numpy())
```
KennethEnevoldsen:
```suggestion
            embeddings.append(emb.detach().cpu().float())
```
Numpy shouldn't be needed, please change throughout

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,571 @@
...
+            inputs.collate_fn = AudioCollator(
+                target_sampling_rate=self.sampling_rate,
+                max_samples=self.max_audio_samples,
+            )
+
+        show_progress_bar = kwargs.get("show_progress_bar", True)
+        all_embeddings: list[np.ndarray] = []
+        for batch in tqdm(inputs, disable=not show_progress_bar, desc="Encoding"):
+            parts: list[torch.Tensor] = []
+            with torch.inference_mode():
```
KennethEnevoldsen: I would move to decorator:

```
@torch.inference_mode()
def encode()
```

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,571 @@
...
+                    parts.append(self.model.encode_audio(waveforms))
+            if not parts:
+                raise ValueError(
+                    f"No supported modality found in batch: {batch.keys()}"
+                )
+            fused = parts[0]
+            for part in parts[1:]:
+                fused += part
+            fused = F.normalize(fused, p=2, dim=-1)
+            all_embeddings.append(fused.detach().cpu().float().numpy())
```
KennethEnevoldsen: should already be on the cpu, detached and float (worth considering if it is better to move everything only once)

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,571 @@
...
+        return np.vstack(all_embeddings)
```
KennethEnevoldsen: can just be done with torch

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,571 @@
...
+
+
+class SemanticRouterMultiModalEmbedLargeWrapper(AbsEncoder):
```
KennethEnevoldsen: what is the diff between the two? Seems like a case where we could just have one?

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,571 @@
...
+                )
+            fused = parts[0]
+            for part in parts[1:]:
+                fused += part
+            fused = F.normalize(fused, p=2, dim=-1)
+            all_embeddings.append(fused.detach().cpu().float().numpy())
+        return np.vstack(all_embeddings)
+
+
+_COMMON = dict(
```
KennethEnevoldsen: just write it out in both

KennethEnevoldsen (commented):
please also update the lock.

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: They're both a mixture of 3 models (3 small, and 3 larger) with slightly different implementation. I updated the code to factorize what is possible into a same base class.
On the three tested tasks, the performance of the two models is different.

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: done

wissam-sib:
@KennethEnevoldsen thank you for the review
I updated the lock and addressed your comments:
* No numpy for embeddings
* Collect tensors, torch.cat(...).float().cpu() once at the end
* decorator @torch.inference_mode() on get_*_embeddings / encode
* _COMMON expanded into both ModelMetas

And added a shared base — SemanticRouterMultiModalEmbedWrapper with small/large subclasses (different backends; encode path shared)

Will re-run both on some tasks to check if nothing broke.

EDIT:
Re-ran both model on the same three tasks and results are not affected by these changes. Ready for re-review.

pyproject.toml
```diff
@@ -139,6 +139,7 @@ speechbrain = ["speechbrain>=0.5.12"]
 muq = ["muq==0.1.0"]
 wav2clip = ["wav2clip==0.1.0"]
 fusion-embedding = ["librosa>=0.10.0", "soundfile>=0.13.1", "torchvision>=0.22.1", "pillow>=10.0.1", "accelerate>=0.26.0"]
+multi-modal-embed = ["librosa>=0.10.0"]
```
Samoed: Which model use librosa?

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,484 @@
...
+        return self.model.encode_image(images)
+
+    def _encode_audio_batch(self, waveforms: list[np.ndarray]) -> torch.Tensor:
+        return self.model.encode_audio(waveforms)
+
+
+class SemanticRouterMultiModalEmbedLargeWrapper(SemanticRouterMultiModalEmbedWrapper):
+    """Wrapper for llm-semantic-router/multi-modal-embed-large (packaged hf_st_mm)."""
+
+    model_label = "multi-modal-embed-large"
```
Samoed:
```suggestion
```

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,484 @@
...
+    ) -> None:
+        from huggingface_hub import snapshot_download
+
+        super().__init__(device=device)
+        self.model_name = model_name
+
+        local_dir = Path(snapshot_download(repo_id=model_name, revision=revision))
+        src_dir = str(local_dir / "src")
+        if src_dir not in sys.path:
+            sys.path.insert(0, src_dir)
```
Samoed: I don't think that loading code from model like this is good idea

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,484 @@
...
+            fn.cache_clear()
+        if hasattr(module, "is_flash_attn_2_available"):
+            module.is_flash_attn_2_available = _unavailable  # type: ignore[misc]
+
+    for name in list(sys.modules):
+        if name.endswith("modeling_modernbert"):
+            del sys.modules[name]
+
+
+def _patch_sentence_transformer_sdpa(hf_st_mm_model: Any) -> None:
```
Samoed: Not sure that I get why you're patching ST

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,484 @@
...
+
+_CITATION = """@misc{multi-modal-embed-2026,
+  title={multi-modal-embed: Compact Multimodal Embeddings with 2DMSE},
+  author={Semantic Router Team},
+  year={2026},
+  url={https://huggingface.co/llm-semantic-router/multi-modal-embed-small}
+}"""
+
+
+def _disable_broken_flash_attn() -> None:
```
Samoed: Can you just pin versions or use SDPA?

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,484 @@
...
+        from transformers.utils import import_utils
+    except Exception:
+        return
+
+    for module in (import_utils, utils):
+        fn = getattr(module, "is_flash_attn_2_available", None)
+        if fn is not None and hasattr(fn, "cache_clear"):
+            fn.cache_clear()
+        if hasattr(module, "is_flash_attn_2_available"):
+            module.is_flash_attn_2_available = _unavailable  # type: ignore[misc]
```
Samoed: I think you can pin transformers

pyproject.toml
wissam-sib: `multi-modal-embed-large`, see here:

https://huggingface.co/llm-semantic-router/multi-modal-embed-large/blob/main/src/hf_st_mm/model.py

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: this comes from the authors python usage guide: https://huggingface.co/llm-semantic-router/multi-modal-embed-large

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: fair point, I was trying to minimize pin but you're right, probably best to do this. Will try, and see if I'm able to simplify implem by removing patches

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: will try to cherry pick from the custom code instead, may allow me to simplify the monkey patches about sdpa etc

wissam-sib:
Good news @Samoed, by cherry picking code from "https://huggingface.co/llm-semantic-router/multi-modal-embed-large/tree/main/src/hf_st_mm", I was able to get the model to work, without having to import librosa (not sure why the authors were importing it). Also, since I can now modify the code, I was able to remove patching.
Im running some test to check it still works the same.
See changes here: https://github.com/embeddings-benchmark/mteb/pull/5011/changes/6eae5828c6ee616852670cc842be958e1256fbe3

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: removed it everywhere and used `mdel_name` instead

pyproject.toml
wissam-sib: by cherry picking code, was able to remove dependence

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: Was able to remove this with code cherry picking

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: no need anymore

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: no need anymore

mteb/models/model_implementations/semantic_router_multimodal_models.py
wissam-sib: no need to pin version, all works fine now without altering uv files

wissam-sib:
> Good news @Samoed, by cherry picking code from "https://huggingface.co/llm-semantic-router/multi-modal-embed-large/tree/main/src/hf_st_mm", I was able to get the model to work, without having to import librosa (not sure why the authors were importing it). Also, since I can now modify the code, I was able to remove patching. Im running some test to check it still works the same. See changes here: [6eae582](https://github.com/embeddings-benchmark/mteb/commit/6eae5828c6ee616852670cc842be958e1256fbe3)

Results are the same, all fixes applied, ready for re-review

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,538 @@
...
+        self.max_audio_samples = int(self.max_audio_seconds * self.sampling_rate)
+
+    @abstractmethod
+    def _encode_text_batch(self, texts: list[str]) -> torch.Tensor: ...
+
+    @abstractmethod
+    def _encode_image_batch(self, images: list[Any]) -> torch.Tensor: ...
+
+    @abstractmethod
+    def _encode_audio_batch(self, waveforms: list[np.ndarray]) -> torch.Tensor: ...
```
Samoed:
```suggestion
```

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,538 @@
...
+    remapped: dict[str, Any] = {}
+    for key, value in state_dict.items():
+        if key.startswith(old):
+            remapped[new + key[len(old) :]] = value
+        else:
+            remapped[key] = value
+    return remapped
+
+
+def _waveforms_from_batch(batch: BatchedInput) -> list[np.ndarray]:
```
Samoed: Why this is needed?

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,538 @@
...
+        self.image_processor = SiglipImageProcessor.from_pretrained(self.IMAGE_ENCODER)
+        self.image_encoder = SiglipModel.from_pretrained(
+            self.IMAGE_ENCODER
+        ).vision_model
+        self.image_proj = nn.Linear(768, self.EMBED_DIM)
+
+        self.audio_processor = WhisperFeatureExtractor.from_pretrained(
+            self.AUDIO_ENCODER
+        )
+        self.audio_encoder = WhisperModel.from_pretrained(self.AUDIO_ENCODER).encoder
```
Samoed: Can you also set reivisions?

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,538 @@
...
+        )
+        self.image_processor = AutoProcessor.from_pretrained(
+            image_encoder_name, trust_remote_code=True
+        )
+
+        whisper = WhisperModel.from_pretrained(audio_encoder_name)
+        self.audio_model = whisper.encoder
+        self.audio_processor = WhisperFeatureExtractor.from_pretrained(
+            audio_encoder_name
+        )
```
Samoed, line 210: Here also?

mteb/models/model_implementations/semantic_router_multimodal_models.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+multi_modal_embed_small = ModelMeta(
+    loader=SemanticRouterMultiModalEmbedSmallWrapper,
+    name="llm-semantic-router/multi-modal-embed-small",
+    revision="fdf8e01b7b0f3a69ac1ac8e2a64dcb1ede177ba4",
+    release_date="2026-02-05",
+    languages=["eng-Latn"],
+    n_parameters=120_000_000,
+    n_embedding_parameters=11_720_448,
```
Samoed: I'm not sure what we should report here, since this model is combination of 3 models

wissam-sib: for n_parameter, it is the sum between mini LM, siglip and whisper (rounded)
for n_embedding, it is only the text part (which seems to be the choice elsewhere as well). So it is only computed from miniLM (siglip is only used for image here).

Samoed: > for n_parameter, it is the sum between mini LM, siglip and whisper (rounded)

siglip is 200M, minilm 20M, whishper 40M, router itself 300m, so they have more than 120M parameters

wissam-sib: Oh... in fact this is what is reported in the model card (see here: https://huggingface.co/llm-semantic-router/multi-modal-embed-small#model-description)

They count it as 90m for siglip, around 8m for whisper, and around 20m for mini LM.

The reason it is lower for siglip is because it only counts the image encoder part (the remaining 110m are the text encoder). And for whisper, the only count the encoder params (the 32m remaining are for the decoder).

I now realize that this opens an interesting question for the whisper family, and probably other models in the repo. We do report the number of parameters for the full models. But if you are only interested in producing embeddings, one could argue you only need to load the whisper encoders so the number you'd care about would be the number of parameters of what you load. I will open an issue to discuss this matter more broadly. In the meantime, I can update the number w.r.t full models if you think it is best.

EDIT: actually, is this what n_embedding_parameters is meant for?

Samoed: By default we're loading all models with all parameters, so I think we should count them all. Also you can try to count number of parameters your self on real downloaded model. 

> But if you are only interested in producing embeddings

For producing embedding you'd need all parameters, so this is `n_parameters`,

> But if you are only interested in producing embeddings, one could argue you only need to load the whisper encoders so the number you'd care about would be the number of parameters of what you load

I don't get that. Do you mean to load only audio part?

> actually, is this what n_embedding_parameters is meant for?

Size of embedding layer for text models

wissam-sib: Great, I updated the numbers to be the sum of all parameters of all models.

What I mean for whisper is that you can implement it in a way that only loads the encoder weights to infer the embeddings. I will share details in the issue.

Thank you for the details about n_embedding_parameters etc.

Samoed: > What I mean for whisper is that you can implement it in a way that only loads the encoder weights to infer the embeddings

Yes, but we're using `WhisperModel` that loads full model

Samoed (commented):
Can you run `MSCOCOT2IRetrieval` to try to repdouce results?

mteb/models/model_implementations/semantic_router_multimodal_models.py
KennethEnevoldsen: `n_embedding_parameters` is mostly intended to calculate the active number of parameters. It is a proxy though so it it doesn't fit (e.g. for MoE) you need to add it.

Of course here it varies by modality, but I would assume that all modalities are present so (as I understand it) it would be number of parameters - embedding params (I don't see any non-active otherwise)

KennethEnevoldsen (commented):
Looked through this and it seems good. @Samoed I suspect you wanted me to look at the embedding layer stuff (I will respond above)

Would be good to have the reproduction on the dataset as suggested by @Samoed  @wissam-sib

Samoed:
> I suspect you wanted me to look at the embedding layer stuff (I will respond above)

I think I wanted your look for overall implementation, because this is a bit strange model, because it reqires to load 3 extra models

KennethEnevoldsen:
Ahh yes indeed. 

I would ideally want it to be a sentence transformer compatible router, which as I understand it is supported in sentence transformers v6. It would however, be quite the refactor and would probably require a reupload, but it would make the model waay more usable by more people.

It would be quite a refactor though and would require a reupload.

isaac-chung:
@KennethEnevoldsen @Samoed would you like to get this merged without the v6 support for now, so that we can run this model for moeb? We can create an issue for the v6 refactor for tracking.

Samoed (approved):
I think we can merge this model

PR #5022, opened 2026-07-25
Dump vLLM to v0.26.0

noooop:
vLLM primarily focuses on generative models. Progress for embedding, rerank, and other models that MTEB focuses on (which vLLM refers to as pooling models) has been slower, and most of the related PR improvements are hardly noticeable. 

1. However, there is one PR that absolutely cannot be missed.

- https://github.com/vllm-project/vllm/pull/47699

For these small models, preprocessing bottlenecks are often encountered.
- Use multithreading to accelerate preprocessing. You can specify the number of threads using renderer_num_workers. The total time scales down almost linearly as the number of renderer workers increases, if you encounter preprocessing bottlenecks.
- Tiling to overlap preprocessing and computation for pooling models offline inference. When preprocessing takes less time than computation, the preprocessing overhead can be almost entirely overlapped.

Note:
Sadly, preprocessing time currently grows with input length, even though the data will be truncated to a small prompt size afterward. This results in a typical embedding working flow—such as preprocessing a wiki page containing hundreds of thousands of tokens and then truncating it to 512 tokens—taking an incredibly long time.

We currently do not have a better truncation method that would make preprocessing take a reasonable amount of time—specifically, a method where the preprocessing time depends on the truncation length rather than the input length. However, we can speed things up as much as possible using multithreading, as well as tiling to overlap preprocessing and computation.

2. Document content alignment

- https://github.com/vllm-project/vllm/pull/35592

The documentation for the vLLM pooling model has been reorganized.

3. Add some benchmark scripts to show how "Why is vLLM Fast" actually works.


Samoed:
Great to hear! I think you can also update this [part of documentation](https://docs.vllm.ai/en/v0.26.0/configuration/engine_args/#-renderer-num-workers) 
> The offline LLM entrypoint uses the synchronous renderer path and processes prompts (including multimodal preprocessing) serially, so this setting has no effect there.

pyproject.toml
```diff
@@ -139,7 +139,7 @@ muq = ["muq==0.1.0"]
 wav2clip = ["wav2clip==0.1.0"]
 fusion-embedding = ["librosa>=0.10.0", "soundfile>=0.13.1", "torchvision>=0.22.1", "pillow>=10.0.1", "accelerate>=0.26.0"]
 torch-vggish-yamnet = ["torch-vggish-yamnet==0.2.1"]
-vllm = ["vllm>=0.11.1"]
+vllm = ["vllm>=0.26.0"]
```
Samoed: I think we can keep previous version range of vllm to keep it more compatible

Samoed (approved):

Issue #591, opened 2024-04-28
Add SciRepEval


Muennighoff:
Would be a cool addition together with MMTEB although a bit orthogonal to the multilingual aspect; cc @amanpreet692

lazizbekravshanov:
I'd like to take this on. SciRepEval is a multi-task benchmark, so to keep things reviewable I'll add it one task at a time rather than in a single large PR.

Starting with the **DRSM** (Disease Research State Model) subtask (`allenai/scirepeval`, config `drsm`) as an `AbsTaskClassification` — a 5-way single-label text classification of biomedical papers (title + abstract) into research-state categories. Since the HF config ships only a single `evaluation` split, the task builds a stratified train/test split for the classification probe, and I'll include descriptive statistics plus random-encoder and `multilingual-e5-small` reference runs.

I'll follow up with additional SciRepEval subtasks (e.g. biomimicry, the SciDocs-derived classification sets) in separate PRs. Happy to adjust the starting task or scope if you'd prefer a different subset first.

PR #5026, opened 2026-07-25
dataset: Add SciRepEval classification tasks (DRSM, biomimicry, FoS, MeSH)

lazizbekravshanov:
## Why

[SciRepEval](https://aclanthology.org/2023.emnlp-main.338/) (Singh et al., EMNLP 2023) is the Semantic Scholar team's recommended successor to SciDocs for scientific-document embedding evaluation, and adding it was requested in #591. It is a multi-format benchmark; this PR adds all of its **classification-type** subtasks. Retrieval- and proximity-style subtasks will follow in a separate `tasks/retrieval/` PR (see below).

## What

Adds four SciRepEval classification tasks (all `category="t2c"`, `eng-Latn`, source `allenai/scirepeval` pinned to `781d35d1bf87253b3dcd0fadcb82bfbee9c244f1`):

| Task | Class | Labels |
|---|---|---|
| `SciRepEvalDRSMClassification` | `AbsTaskClassification` | 5-way Disease Research State Model |
| `SciRepEvalBiomimicryClassification` | `AbsTaskClassification` | binary biomimicry relevance |
| `SciRepEvalFoSClassification` | `AbsTaskMultilabelClassification` | multi-label MAG Fields of Study |
| `SciRepEvalMeSHDescriptorsClassification` | `AbsTaskClassification` | 30-way MeSH descriptor |

Implementation notes:
- Input text is `title` + `abstract`.
- `drsm` and `biomimicry` expose only a single `evaluation` split, so `dataset_transform()` builds a stratified train/test split (test subsampled to 2048).
- `fos` and `mesh_descriptors` have huge upstream `train` splits (0.5M / 2M rows), so only the `evaluation` split is loaded (via `split` in `metadata.dataset`) and split locally.
- `mesh_descriptors` rows are `(paper, descriptor)` pairs, so papers are **deduplicated** before splitting to prevent one abstract leaking across train/test with different labels.
- Each task registers in the relevant `eng/__init__.py` and ships precomputed descriptive statistics.

## Reference runs (accuracy, CPU)

`mteb run -m {model} -t {task}`:

| Task | `baseline-random-encoder` | `multilingual-e5-small` |
|---|---|---|
| DRSM | 0.199 | 0.598 |
| Biomimicry | 0.510 | 0.728 |
| FoS (multi-label) | 0.001 | 0.178 |
| MeSH | 0.033 | 0.582 |

Every task is well above its random baseline and none is trivial.

## Out of scope (separate follow-up PR)

These SciRepEval configs are retrieval/proximity/regression tasks and belong under `tasks/retrieval/` (or pair-classification / regression), not `tasks/classification/`, so they are intentionally excluded here: `scidocs_mag_mesh`, `scidocs_view_cite_read`, `cite_prediction*`, `high_influence_cite`, `same_author`, `paper_reviewer_matching`, `nfcorpus`, `search`, `trec_covid`, `relish`, `tweet_mentions`, `cite_count`, `pub_year`, `peer_review_score_hIndex`.

## Checklist

- [x] Metadata filled for every task
- [x] Descriptive statistics computed
- [x] Tasks picked up by `tests/test_tasks/test_metadata.py` (4 passed) and `test_dataset_on_hf.py` (revision check passes)
- [x] `ruff check` and `ruff format --check` clean
- [x] Reference runs beat the random baseline

Part of #591.

### DIFF
--- a/mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
+++ b/mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+from mteb.abstasks.classification import AbsTaskClassification
+from mteb.abstasks.task_metadata import TaskMetadata
+
+
+class SciRepEvalBiomimicryClassification(AbsTaskClassification):
+    metadata = TaskMetadata(
+        name="SciRepEvalBiomimicryClassification",
+        description=(
+            "Biomimicry relevance classification task from the SciRepEval "
+            "benchmark. Given the title and abstract of a scientific paper, "
+            "predict whether the paper is relevant to biomimicry research "
+            "(binary classification)."
+        ),
+        reference="https://aclanthology.org/2023.emnlp-main.338/",
+        dataset={
+            "path": "allenai/scirepeval",
+            "name": "biomimicry",
+            "revision": "781d35d1bf87253b3dcd0fadcb82bfbee9c244f1",
+        },
+        type="Classification",
+        category="t2c",
+        modalities=["text"],
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="accuracy",
+        date=("2022-01-01", "2023-12-06"),
+        domains=["Academic", "Engineering", "Written"],
+        task_subtypes=["Topic classification"],
+        license="not specified",
+        annotations_creators="derived",
+        dialect=[],
+        sample_creation="found",
+        bibtex_citation=r"""
+@inproceedings{singh-etal-2023-scirepeval,
+  address = {Singapore},
+  author = {Singh, Amanpreet  and
+D{'}Arcy, Mike  and
+Cohan, Arman  and
+Downey, Doug  and
+Feldman, Sergey},
+  booktitle = {Proceedings of the 2023 Conference on Empirical Methods in Natural Language Processing},
+  doi = {10.18653/v1/2023.emnlp-main.338},
+  month = dec,
+  pages = {5548--5566},
+  publisher = {Association for Computational Linguistics},
+  title = {{SciRepEval: A Multi-Format Benchmark for Scientific Document Representations}},
+  url = {https://aclanthology.org/2023.emnlp-main.338/},
+  year = {2023},
+}
+""",
+    )
+
+    def dataset_transform(self, num_proc: int | None = None) -> None:
+        # SciRepEval exposes only a single "evaluation" split for `biomimicry`,
+        # so we build a stratified train/test split for the classification probe.
+        ds = self.dataset["evaluation"]
+        ds = ds.map(
+            lambda x: {"text": f"{x['title'] or ''}\n\n{x['abstract'] or ''}".strip()},
+            num_proc=num_proc,
+        )
+        keep = {"text", "label"}
+        ds = ds.remove_columns([c for c in ds.column_names if c not in keep])
+        ds = ds.class_encode_column("label")
+        ds = ds.train_test_split(
+            test_size=0.3, seed=self.seed, stratify_by_column="label"
+        )
+        ds = self.stratified_subsampling(
+            dataset_dict=ds, seed=self.seed, splits=["test"]
+        )
+        self.dataset = ds
--- a/mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
+++ b/mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
@@ -8,10 +8,10 @@ class SciRepEvalBiomimicryClassification(AbsTaskClassification):
     metadata = TaskMetadata(
         name="SciRepEvalBiomimicryClassification",
         description=(
-            "Biomimicry relevance classification task from the SciRepEval "
-            "benchmark. Given the title and abstract of a scientific paper, "
-            "predict whether the paper is relevant to biomimicry research "
-            "(binary classification)."
+            "Biomimicry relevance classification. Given the title and abstract "
+            "of a scientific paper, predict whether the paper is relevant to "
+            "biomimicry research (binary classification). Published as a part of "
+            "SciRepEval."
         ),
         reference="https://aclanthology.org/2023.emnlp-main.338/",
         dataset={
@@ -28,8 +28,8 @@ class SciRepEvalBiomimicryClassification(AbsTaskClassification):
         date=("2022-01-01", "2023-12-06"),
         domains=["Academic", "Engineering", "Written"],
         task_subtypes=["Topic classification"],
-        license="not specified",
-        annotations_creators="derived",
+        license="odc-by",
+        annotations_creators="human-annotated",
         dialect=[],
         sample_creation="found",
         bibtex_citation=r"""
--- a/mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
+++ b/mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
@@ -22,7 +22,7 @@ class SciRepEvalBiomimicryClassification(AbsTaskClassification):
         type="Classification",
         category="t2c",
         modalities=["text"],
-        eval_splits=["train"],
+        eval_splits=["evaluation"],
         eval_langs=["eng-Latn"],
         main_score="accuracy",
         date=("2022-01-01", "2023-12-06"),
@@ -53,10 +53,11 @@ Feldman, Sergey},
     )
 
     is_cross_validation: bool = True
+    train_split: str = "evaluation"
 
     def dataset_transform(self, num_proc: int | None = None) -> None:
         # SciRepEval exposes only a single "evaluation" split for `biomimicry`.
-        # We keep it as one split named "train" and let MTEB's cross-validation
+        # We keep it named "evaluation" and let MTEB's cross-validation
         # path (is_cross_validation) run KFold over the whole split.
         from datasets import DatasetDict
 
@@ -68,4 +69,4 @@ Feldman, Sergey},
         keep = {"text", "label"}
         ds = ds.remove_columns([c for c in ds.column_names if c not in keep])
         ds = ds.class_encode_column("label")
-        self.dataset = DatasetDict({"train": ds})
+        self.dataset = DatasetDict({"evaluation": ds})
--- a/mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
+++ b/mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
@@ -22,7 +22,7 @@ class SciRepEvalBiomimicryClassification(AbsTaskClassification):
         type="Classification",
         category="t2c",
         modalities=["text"],
-        eval_splits=["test"],
+        eval_splits=["train"],
         eval_langs=["eng-Latn"],
         main_score="accuracy",
         date=("2022-01-01", "2023-12-06"),
@@ -52,9 +52,14 @@ Feldman, Sergey},
 """,
     )
 
+    is_cross_validation: bool = True
+
     def dataset_transform(self, num_proc: int | None = None) -> None:
-        # SciRepEval exposes only a single "evaluation" split for `biomimicry`,
-        # so we build a stratified train/test split for the classification probe.
+        # SciRepEval exposes only a single "evaluation" split for `biomimicry`.
+        # We keep it as one split named "train" and let MTEB's cross-validation
+        # path (is_cross_validation) run KFold over the whole split.
+        from datasets import DatasetDict
+
         ds = self.dataset["evaluation"]
         ds = ds.map(
             lambda x: {"text": f"{x['title'] or ''}\n\n{x['abstract'] or ''}".strip()},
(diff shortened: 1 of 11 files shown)


mteb/tasks/classification/eng/scirepeval_drsm_classification.py
```diff
@@ -0,0 +1,74 @@
...
+            "Disease Research State Model (DRSM) classification task from the "
+            "SciRepEval benchmark. Given the title and abstract of a biomedical "
+            "paper, classify it into one of five research-state categories: "
+            "'clinical characteristics or disease pathology', 'disease mechanism', "
+            "'therapeutics in the clinic', 'patient-based therapeutics', or 'other'."
+        ),
+        reference="https://aclanthology.org/2023.emnlp-main.338/",
+        dataset={
+            "path": "allenai/scirepeval",
+            "name": "drsm",
```
Samoed, line 21: Why do you choose only this config?

lazizbekravshanov: Good question! I scoped this first PR to just the `drsm` config to keep the initial review small and to establish the SciRepEval task pattern in MTEB — `drsm` is a clean single-label 5-class classification that maps directly onto `AbsTaskClassification`. As noted on #591, my plan was to add the other SciRepEval subtasks in follow-up PRs: `biomimicry` (binary classification), the multi-label sets `fos` / `mesh_descriptors` (which need `AbsTaskMultilabelClassification`), and the `scidocs_*` retrieval/classification tasks.

That said, I’m very happy to fold more configs into this PR if you’d prefer a single, fuller SciRepEval contribution rather than incremental ones — just let me know which grouping you’d like and I’ll add them here. Thanks for reviewing!

Samoed: Since `biomimicry` is also classification, I think you can integrate it in same task, or in same PR

lazizbekravshanov: Done — I've expanded this PR to include all the classification-type SciRepEval configs:

- **SciRepEvalBiomimicryClassification** — binary, `AbsTaskClassification`
- **SciRepEvalFoSClassification** — multi-label, `AbsTaskMultilabelClassification` (MAG fields of study)
- **SciRepEvalMeSHDescriptorsClassification** — 30-class, `AbsTaskClassification` (papers are deduplicated so an abstract can't leak across the train/test split)

…alongside the original `drsm`. Each ships pinned metadata + descriptive stats and passes `test_metadata` / `test_dataset_on_hf`. Reference runs (accuracy, random-encoder → e5-small): DRSM 0.20→0.60, biomimicry 0.51→0.73, FoS 0.001→0.18, MeSH 0.03→0.58.

I kept the retrieval/proximity configs (`scidocs_*`, `cite_*`, `nfcorpus`, `search`, `trec_covid`, `relish`, `paper_reviewer_matching`, `tweet_mentions`, `same_author`, `high_influence_cite`) and the regression ones (`cite_count`, `pub_year`, `peer_review_score_hIndex`) out of this PR, since they belong under `tasks/retrieval/` (or pair-classification / regression) rather than `tasks/classification/`. I'll add the retrieval group as a separate follow-up. Does that scope work for you?

lazizbekravshanov:
Pushed `c22bd5be` to fix the two CI failures on the expanded PR:

- **`test_dataset_quality`** flagged one duplicate title+abstract in the `SciRepEvalDRSMClassification` train split (6169 samples / 6168 unique). I now drop duplicate texts before the train/test split so the same document can't leak across it, and regenerated the descriptive stats.
- **`typos`** flagged the `FoS` (Fields of Study) acronym in the new multilabel task. Added `SciRepEvalFoSClassification` and `FoS` to the `[tool.typos...extend-identifiers]` allowlist, matching how the repo already whitelists intentional task-name tokens.

`make lint-check` is green locally and `_task_quality` reports 0 errors for all four SciRepEval tasks. CI re-running now.

Samoed:
You can also run tasks in cross_validation config

mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class SciRepEvalBiomimicryClassification(AbsTaskClassification):
+    metadata = TaskMetadata(
+        name="SciRepEvalBiomimicryClassification",
+        description=(
+            "Biomimicry relevance classification task from the SciRepEval "
+            "benchmark. Given the title and abstract of a scientific paper, "
+            "predict whether the paper is relevant to biomimicry research "
+            "(binary classification)."
```
KennethEnevoldsen: In general I would focus write it like so: 


```suggestion
            "Biomimicry relevance classification. Given the title and abstract of a scientific paper, predict whether the paper is relevant to biomimicry research (binary classification). Published as a part of SciRepEval."
```

Focusing on what it tests, how it tests it and then other relevant attributes

mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
```diff
@@ -0,0 +1,72 @@
...
+  pages = {5548--5566},
+  publisher = {Association for Computational Linguistics},
+  title = {{SciRepEval: A Multi-Format Benchmark for Scientific Document Representations}},
+  url = {https://aclanthology.org/2023.emnlp-main.338/},
+  year = {2023},
+}
+""",
+    )
+
+    def dataset_transform(self, num_proc: int | None = None) -> None:
```
KennethEnevoldsen, line 58: We prefer to avoid these (harder to maintain) and instead upload it as a formatted dataset.

You can see how to do so here:
https://docs.mteb.org/contributing/adding_a_dataset/#pushing-the-dataset-to-the-hub

@Samoed also sometimes does it at the end. @Samoed would you rather do it yourself?

mteb/tasks/multilabel_classification/eng/scirepeval_fos_classification.py
```diff
@@ -0,0 +1,72 @@
...
+        type="MultilabelClassification",
+        category="t2c",
+        modalities=["text"],
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="accuracy",
+        date=("2022-01-01", "2023-12-06"),
+        domains=["Academic", "Written"],
+        task_subtypes=["Topic classification"],
+        license="not specified",
```
KennethEnevoldsen: License seems to generally be unspecified. I suspect this is just from looking at the huggingface? I typically limits the downstream use quite a bit not having it (some companies don't want the risk)

mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
```diff
@@ -0,0 +1,72 @@
...
+        category="t2c",
+        modalities=["text"],
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="accuracy",
+        date=("2022-01-01", "2023-12-06"),
+        domains=["Academic", "Engineering", "Written"],
+        task_subtypes=["Topic classification"],
+        license="not specified",
+        annotations_creators="derived",
```
KennethEnevoldsen: how are these derived? Seems like a label that is annotated

KennethEnevoldsen (commented):
A few minor comments, but otherwise I think it looks good

lazizbekravshanov:
Thanks — the `allenai/scirepeval` configs do expose cross-validation splits, so that's a good option. For now these four are framed as single train/test-split classification probes, consistent with how the other MTEB classification tasks run. Would you like me to add a separate cross-validation-based variant for any of them, or is the current single-split framing fine here? Happy to go either way.

mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
lazizbekravshanov: Applied, thanks — much cleaner framing. I've also rewritten the DRSM, FoS, and MeSH descriptions in the same "what it tests → how → other attributes" style so the four read consistently. Pushed in `4bd90d9b`.

mteb/tasks/multilabel_classification/eng/scirepeval_fos_classification.py
lazizbekravshanov: Good catch — the HF card is empty, which is where "not specified" came from. The license is actually specified in the SciRepEval source repo: the README's License section states *"The aggregate benchmark is released under ODC-BY license … For constituent datasets, also go through the individual licensing requirements, as applicable"* (https://github.com/allenai/scirepeval#license). I've set `license="odc-by"` on all four SciRepEval tasks (`4bd90d9b`). Note the repo's `LICENSE` file is Apache-2.0, but that covers the code — the README explicitly assigns ODC-BY to the benchmark data. The FoS fields-of-study labels ultimately trace to MAG, but the redistributed SciRepEval benchmark is governed by ODC-BY.

mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
lazizbekravshanov: You're right that these are annotated rather than structurally derived — I've changed `annotations_creators` to `human-annotated` (`4bd90d9b`). Per the SciRepEval paper, the biomimicry labels are relevance tags from NASA's PeTaL database ("We sample tags for a set of papers in the PeTaL database … indicating whether each paper is about biomimicry"), and the paper notes the Biomimicry/DRSM/FoS evaluation subsets have "manually annotated gold labels" (https://aclanthology.org/2023.emnlp-main.338/). I kept it at `human-annotated` rather than `expert-annotated` since the paper describes manual gold labeling but doesn't specify domain-expert annotators (PeTaL used a project/crowdsourced labeling effort). Happy to switch to `expert-annotated` if you'd prefer.

mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
lazizbekravshanov: Makes sense — I read through https://docs.mteb.org/contributing/adding_a_dataset/#pushing-the-dataset-to-the-hub. I'm happy to apply the current `dataset_transform` once for each of the four tasks, push the pre-split `train`/`test` data with `push_dataset_to_hub`, then point each task's `dataset` at the new repo and drop the `dataset_transform` methods so they just load the formatted splits. Just let me know the target org/naming you'd prefer (e.g. `mteb/{task.metadata.name}`).

@Samoed — Kenneth mentioned you sometimes handle this step at the end; would you rather push these yourself, or should I prepare the formatted uploads for you to review?

lazizbekravshanov:
Thanks, that makes sense. To confirm I understood: there is no cross_validation config in the allenai/scirepeval dataset itself (the HF configs are just the subsets: biomimicry, drsm, fos, mesh_descriptors, and biomimicry/drsm ship only a single evaluation split). So you mean MTEB's own is_cross_validation mode on AbsTaskClassification, where the task points its eval split at the single split and MTEB runs KFold (n_splits) over the whole set instead of a fixed train/test split.

Right now these four tasks take the single evaluation split and build a fixed stratified 70/30 split, then the standard path runs n_experiments subsamples against that fixed test set. The CV path instead rotates the held-out fold over all the data, which for drsm and biomimicry (single split, roughly 9k and 11k rows) is a cleaner fit and closer to how the SciRepEval paper scores these (average over the full labeled set rather than one held-out cut).

This is a small change: keep the whole evaluation split as one split, set eval_splits to it, set is_cross_validation = True, and drop the manual train_test_split. That is the same setup that around 40 existing single-split tasks use (for example SpokeNEnglish).

One caveat: the CV code path uses the single-label _undersample_data and _run_experiment helpers, and AbsTaskMultilabelClassification (our FoS task) overrides different methods, so CV would not run correctly there without extra framework changes. So I would apply CV to the single-label tasks (drsm, biomimicry, and mesh if you want) and leave FoS on the current path.

Do you want these as separate CV tasks alongside the current ones, or should I just switch the existing single-label tasks over to is_cross_validation? Happy to do either, just let me know which you prefer.

KennethEnevoldsen:
My comments have been resolved - @Samoed not sure what you meant with cross-val?

Samoed:
I meant that this is possible to run tasks in cross validation config https://github.com/embeddings-benchmark/mteb/blob/f5337a721cb185d415dd29c50b5027ea5f6654e3/mteb/abstasks/classification.py#L120 without splittings tasks into train/test

KennethEnevoldsen:
So you are suggesting we change the single label data to CV? (I would be all for that)

lazizbekravshanov:
Yes, exactly. I switched the single-label tasks (drsm, biomimicry, and mesh) to is_cross_validation, so they now run 5-fold cross validation over the whole evaluation split instead of a fixed train/test split, and dropped the manual train_test_split. I confirmed the CV path runs (drsm scores 0.594 with e5-small vs 0.194 for the random baseline). FoS is multilabel and the CV path uses the single-label helpers, so I left FoS on the current path. Pushed in da7340a6. One small note: the mesh task keeps a size cap that samples up to 8192 rows before CV, happy to remove that cap and run CV over the full mesh set if you would prefer.

KennethEnevoldsen (approved):
I am good with this - @Samoed anything to add?

mteb/tasks/classification/eng/scirepeval_biomimicry_classification.py
```diff
@@ -0,0 +1,71 @@
...
+
+        ds = self.dataset["evaluation"]
+        ds = ds.map(
+            lambda x: {"text": f"{x['title'] or ''}\n\n{x['abstract'] or ''}".strip()},
+            num_proc=num_proc,
+        )
+        keep = {"text", "label"}
+        ds = ds.remove_columns([c for c in ds.column_names if c not in keep])
+        ds = ds.class_encode_column("label")
+        self.dataset = DatasetDict({"train": ds})
```
Samoed: I don't think you need to rename to `train`. You can keep `evaluation` split as a name

Samoed (approved):

isaac-chung:
Since both reviewers have approved, I'm merging this now. Thanks stale bot lol

PR #5029, opened 2026-07-26
Add CoREB task-type router

keon:
## Summary

Adds a transparent task-type router that uses F2LLM-v2-330M for Retrieval and
C2LLM-7B for Reranking. Both child revisions are pinned, and the router reads
only the coarse MTEB task type.

## Reproduction

- Benchmark: `CoREB(v1)` (all six tasks)
- MTEB: `2.18.6`
- Device: NVIDIA L4
- Child settings: each pinned model's official MTEB loader configuration
- Precision: bfloat16
- Batch size: 4 for Retrieval; 2 for Reranking
- Routing: Retrieval -> F2LLM-v2-330M; Reranking -> C2LLM-7B
- Measured CoREB(v1) macro mean: `0.6017383`

## Disclosure

This composition was selected using public CoREB results. CoREB exposes only a
test split, so this is a leaderboard-adapted routing policy. It does not route
on task name, examples, labels, or scores.

## Checklist

- [x] I have tested the implementation works on representative Retrieval and
  Reranking task metadata
- [x] The model card/config repository is public
- [x] There is no original-paper result to reproduce; both child revisions and
  the complete CoREB reproduction environment are documented

## Before merging

- Results PR: https://github.com/embeddings-benchmark/results/pull/642
- Confirm maintainers accept task-dependent embedding dimensions for a router.


mteb/models/model_implementations/keonkim_coreb_router.py
```diff
@@ -0,0 +1,187 @@
...
+ROUTES = {
+    "Retrieval": Route(
+        model_name=F2LLM_v2_330M.name,
+        revision=F2LLM_v2_330M.revision,
+    ),
+    "Reranking": Route(
+        model_name=C2LLM_7B.name,
+        revision=C2LLM_7B.revision,
+    ),
+}
```
Samoed, line 45: Not sure I get this idea. Why do you want to add this setup? This seems like just combining scores for 2 models.

PR #5030, opened 2026-07-26
ci: test reference models on all benchmarks

Samoed:


KennethEnevoldsen (approved):
<3

tests/test_benchmarks/test_reference_models.py
```diff
@@ -31,6 +31,9 @@ def _get_expected_task_names(benchmark, model_name):
     task_names = []
     benchmark_tasks = []
     for task in benchmark.tasks:
+        if task.metadata.name == "CodeRAGStackoverflowPosts":
+            # CodeRAGStackoverflowPosts is too big
```
Samoed, line 38: This dataset is too big. It took for me 4h to encode queries on e5-small. I don't think we can run it

KennethEnevoldsen: Hmm maybe just ignore the benchmark instead? (So that new benchmark don't include it)

Or the benchmark / task combo

KennethEnevoldsen (approved):

PR #5039, opened 2026-07-27
model: add RTriever-4B and DIVER-Retriever-4B

yilunzhao:
Adds two reasoning-intensive retrievers.

Both are Qwen3-Embedding-4B fine-tunes, so both reuse `q3e_instruct_loader`: last-token pooling, cosine similarity, and an `Instruct: ...\nQuery:` prefix on queries only. I checked that against each checkpoint's `config_sentence_transformers.json` — the template the loader builds matches the card verbatim.

| model | params | license | citation |
| --- | --- | --- | --- |
| `yale-nlp/RTriever-4B` | 4.02B | MIT | ACL 2026 |
| `AQ-MedAI/Diver-Retriever-4B-1020` | 4.02B | Apache-2.0 | arXiv:2508.07995 |

### Training-data annotations

`AQ-MedAI/Diver-Retriever-4B-1020` declares the BRIGHT corpora in `training_datasets`. Its model card lists [`reasonir/reasonir-data`](https://huggingface.co/datasets/reasonir/reasonir-data) among its training sets, and the hard-query half of that dataset stores positive documents by id and reconstructs them by loading `xlangai/BRIGHT` "documents" — so those corpora are part of DIVER's training data transitively. Its other listed sets (MATH_qCoT..., truehealth/medqa, AQ-MedAI/PRGB-ZH) are not MTEB tasks.

`yale-nlp/RTriever-4B` is trained on synthetic data with no MTEB task overlap.

### Checks

- `pytest tests/test_models/test_model_meta.py` — 1964 passed
- `ruff check` / `ruff format --check` — clean
- Both loaded and evaluated end-to-end across a 7-subset retrieval benchmark

---

### Unrelated follow-up on `ReasonIR/ReasonIR-8B`

Two things I ran into while working with the existing entry in `reasonir_model.py`, happy to open a separate issue or PR if useful:

**1. The registered name does not match the HF repo id.** It is `ReasonIR/ReasonIR-8B`, but the repo is [`reasonir/ReasonIR-8B`](https://huggingface.co/reasonir/ReasonIR-8B) (lowercase org), so `get_model_meta("reasonir/ReasonIR-8B")` raises `KeyError`.

**2. `training_datasets` omits BRIGHT.** It currently lists the nine public sets from the paper's section D. But the hard-query half of `reasonir-data` stores its positive documents by id only, and those ids resolve against the BRIGHT corpus. Checking directly rather than inferring from the dataset card — rows from config `hq` give positives like `['', 'camel_37833']`, and each id is present in `xlangai/BRIGHT` config `documents`:

| id in `reasonir-data` hq | found in BRIGHT `documents` | content |
| --- | --- | --- |
| `camel_37833` | aops, theoremqa_questions | "A math student needs to find the next three terms…" |
| `aqua_rat_14014` | aops, theoremqa_questions | "When the positive integer x is divided by…" |
| `camel_44852` | aops, theoremqa_questions | "A sound signal is given by f(t) = sin(2πt)…" |

So the training positives are BRIGHT passages verbatim (the queries are synthetic). Adding the BRIGHT tasks to `training_datasets` would let the leaderboard flag that overlap, which currently goes unreported on BRIGHT.

### DIFF
--- a/mteb/models/model_implementations/aq_medai_models.py
+++ b/mteb/models/model_implementations/aq_medai_models.py
@@ -1,0 +1,62 @@
+from __future__ import annotations
+
+from mteb.models.model_implementations.qwen3_models import q3e_instruct_loader
+from mteb.models.model_meta import ModelMeta
+
+# Fetched verbatim from https://arxiv.org/bibtex/2508.07995 (v5); the copy in
+# the model card predates the revision and lists an older author set.
+DIVER_CITATION = """@misc{sun2026divermultistageapproachreasoningintensive,
+      title={DIVER: A Multi-Stage Approach for Reasoning-intensive Information Retrieval},
+      author={Duolin Sun and Meixiu Long and Dan Yang and Junjie Wang and Yecheng Luo and Yue Shen and Jian Wang and Hualei Zhou and Chunxiao Guo and Peng Wei and Jiahai Wang and Jinjie Gu},
+      year={2026},
+      eprint={2508.07995},
+      archivePrefix={arXiv},
+      primaryClass={cs.IR},
+      url={https://arxiv.org/abs/2508.07995},
+}"""
+
+DIVER_RETRIEVER_4B = ModelMeta(
+    # A Qwen3-Embedding-4B fine-tune, so it keeps that interface: last-token
+    # pooling, cosine similarity, and an "Instruct: ...\nQuery:" prefix on
+    # queries only (verified against its config_sentence_transformers.json).
+    loader=q3e_instruct_loader,
+    name="AQ-MedAI/Diver-Retriever-4B-1020",
+    model_type=["dense"],
+    languages=["eng-Latn", "zho-Hans"],
+    open_weights=True,
+    revision="3984377d297f4c3bcf5d625ea7b48feea7f58c57",
+    release_date="2025-10-20",
+    n_parameters=4021774336,
+    n_embedding_parameters=388262400,
+    memory_usage_mb=7670,
+    embed_dim=2560,
+    max_tokens=40960,
+    license="apache-2.0",
+    reference="https://huggingface.co/AQ-MedAI/Diver-Retriever-4B-1020",
+    similarity_fn_name="cosine",
+    framework=["Sentence Transformers", "PyTorch", "safetensors", "Transformers"],
+    use_instructions=True,
+    public_training_code="https://github.com/AQ-MedAI/Diver",
+    public_training_data=None,
+    # The model card lists reasonir/reasonir-data among its training sets, and
+    # the hard-query half of that dataset reconstructs its positive documents
+    # from xlangai/BRIGHT — so the BRIGHT corpora are part of this model's
+    # training data too. The other listed sets (MATH_qCoT..., truehealth/medqa,
+    # AQ-MedAI/PRGB-ZH) are not MTEB tasks.
+    training_datasets={
+        "BrightAopsRetrieval",
+        "BrightBiologyRetrieval",
+        "BrightEarthScienceRetrieval",
+        "BrightEconomicsRetrieval",
+        "BrightLeetcodeRetrieval",
+        "BrightPonyRetrieval",
+        "BrightPsychologyRetrieval",
+        "BrightRoboticsRetrieval",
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/aq_medai_models.py
```diff
@@ -0,0 +1,62 @@
...
+    similarity_fn_name="cosine",
+    framework=["Sentence Transformers", "PyTorch", "safetensors", "Transformers"],
+    use_instructions=True,
+    public_training_code="https://github.com/AQ-MedAI/Diver",
+    public_training_data=None,
+    # The model card lists reasonir/reasonir-data among its training sets, and
+    # the hard-query half of that dataset reconstructs its positive documents
+    # from xlangai/BRIGHT — so the BRIGHT corpora are part of this model's
+    # training data too. The other listed sets (MATH_qCoT..., truehealth/medqa,
+    # AQ-MedAI/PRGB-ZH) are not MTEB tasks.
```
ayush1298: you can add them in `training_datasets` and comment them out.

ayush1298:
@yilunzhao There are three other variants of Diver-Retriever; can you integrate them as well?
DIVER-Retriever-4B, DIVER-Retriever-1.7B, DIVER-Retriever-0.6B

mteb/models/model_implementations/yale_nlp_models.py
```diff
@@ -0,0 +1,57 @@
...
+    doi = "10.18653/v1/2026.acl-long.1705",
+    pages = "36776--36806",
+    ISBN = "979-8-89176-390-6",
+}"""
+
+RTRIEVER_4B = ModelMeta(
+    # RTriever-4B keeps the Qwen3-Embedding interface it was fine-tuned from:
+    # last-token pooling, cosine similarity, and an "Instruct: ...\nQuery:"
+    # prefix on queries only.
+    loader=q3e_instruct_loader,
```
Samoed:
```suggestion
    loader=InstructSentenceTransformer,
```

mteb/models/model_implementations/aq_medai_models.py
```diff
@@ -0,0 +1,62 @@
...
+      archivePrefix={arXiv},
+      primaryClass={cs.IR},
+      url={https://arxiv.org/abs/2508.07995},
+}"""
+
+DIVER_RETRIEVER_4B = ModelMeta(
+    # A Qwen3-Embedding-4B fine-tune, so it keeps that interface: last-token
+    # pooling, cosine similarity, and an "Instruct: ...\nQuery:" prefix on
+    # queries only (verified against its config_sentence_transformers.json).
+    loader=q3e_instruct_loader,
```
Samoed:
```suggestion
    loader=InstructSentenceTransformer,
```

yilunzhao:
Good call — added all three. The PR now covers four DIVER-Retriever checkpoints:

| model | params | base |
| --- | --- | --- |
| `AQ-MedAI/Diver-Retriever-4B-1020` | 4.02B | Qwen3-Embedding-4B |
| `AQ-MedAI/Diver-Retriever-4B` | 4.02B | Qwen3-Embedding-4B |
| `AQ-MedAI/Diver-Retriever-1.7B` | 1.72B | Qwen3-1.7B base LM |
| `AQ-MedAI/Diver-Retriever-0.6B` | 0.60B | Qwen3-Embedding-0.6B |

All four expose the same interface, so they all reuse `q3e_instruct_loader`. I marked the older 4B as `superseded_by` the `-1020` release, and factored the shared training-data annotation into `DIVER_TRAINING_DATA`.

One thing worth flagging upstream: `Diver-Retriever-1.7B`'s `1_Pooling/config.json` carries `word_embedding_dimension: 2560`, which is the 4B model's hidden size — the 1.7B checkpoint's is 2048. Looks like a copied config. `ModelMeta` records the real value and I left a note in the code.

Evaluations for the three new variants are running; results will go to `embeddings-benchmark/results`.

Samoed:
> All four expose the same interface, so they all reuse q3e_instruct_loader

You shoudn't use `q3e_instruct_loader`. Better to set class with parameters explicitly

ayush1298 (approved):
LGTM!

Samoed (approved):

PR #5045, opened 2026-07-29
[MOEB] Add dataset SHS100K

### LINKED ISSUE #5042: Add dataset: SHS100k

https://huggingface.co/datasets/Wissam42/SHS100K-A2A

### Arxiv link

_No response_

Packaging classic NovaFrost SHS100K-TEST into MTEB a2a retrieval format.

Source metadata (YouTube URLs only): https://github.com/NovaFrost/SHS100K

Joins SHS100K-TEST (set_id, ver_id) with `list`, downloads audio-only via yt-dlp, then builds corpus / queries / qrels.

wissam-sib:
- [x] I have outlined why this dataset is filling an existing gap in the MOEB benchmark
  - [x] `LCO-Embedding-Omni-3B`

Results before adding skip first result:
```
metric                 random LCO-Embedding-Omni-3B
ndcg_at_1             0.00098        0.00098
ndcg_at_5             0.03085        0.55617
ndcg_at_10            0.03646        0.63544
ndcg_at_20            0.03999        0.65189
ndcg_at_100           0.06952        0.67249
map_at_10             0.00260        0.21507
recall_at_1           0.00005        0.00005
recall_at_5           0.00487        0.13658
recall_at_10          0.01014        0.28543
recall_at_20          0.02067        0.45478
mrr_at_10             0.07594        0.44866
```

The queries are also in the corpus set (under a different id), so first results are always relevant but not in ground truth (which causes low ndcg at 1 and recall at 1). Fixed after generating the results. Applied same fix to Covers80 which has the same issue.

### DIFF
--- a/mteb/tasks/retrieval/eng/covers80_retrieval.py
+++ b/mteb/tasks/retrieval/eng/covers80_retrieval.py
@@ -5,6 +5,9 @@ from mteb.abstasks.task_metadata import TaskMetadata
 
 
 class Covers80A2ARetrieval(AbsTaskRetrieval):
+    # Query audio is also in the corpus under a different id; drop that self-hit.
+    skip_first_result = True
+
     metadata = TaskMetadata(
         name="Covers80A2ARetrieval",
         description=(
--- a/mteb/descriptive_stats/Image/Any2AnyRetrieval/SHS100KA2ARetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyRetrieval/SHS100KA2ARetrieval.json
@@ -1,0 +1,46 @@
+{
+    "test": {
+        "num_samples": 2048,
+        "num_queries": 1024,
+        "num_documents": 1024,
+        "number_of_characters": 0,
+        "documents_text_statistics": null,
+        "documents_image_statistics": null,
+        "documents_audio_statistics": {
+            "total_duration_seconds": 257476.19076643992,
+            "min_duration_seconds": 56.85333333333333,
+            "average_duration_seconds": 251.44159254535148,
+            "max_duration_seconds": 1136.768,
+            "unique_audios": 1023,
+            "average_sampling_rate": 47291.6015625,
+            "sampling_rates": {
+                "48000": 838,
+                "44100": 186
+            }
+        },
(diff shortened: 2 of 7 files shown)


mteb/tasks/retrieval/eng/covers80_retrieval.py
```diff
@@ -5,6 +5,9 @@
 
 
 class Covers80A2ARetrieval(AbsTaskRetrieval):
+    # Query audio is also in the corpus under a different id; drop that self-hit.
+    skip_first_result = True
```
Samoed, line 9: Why do you change this task?

Samoed (commented):
I made review, but didn't submit it

mteb/tasks/retrieval/eng/covers80_retrieval.py
wissam-sib: Thanks for all the reviews btw 🙏 really appreciate!!

As you can see in the PR comment above, I noticed the results for k=1 were strange (figured it was because the top retrieved element in the corpus is the query itself - but with different id) so I applied a fix for SHS. Then I realized that the issue was also in covers80 (see here https://github.com/embeddings-benchmark/mteb/pull/4986#issue-4939106280 - but I had missed it for that one) so I also applied the fix. I can move this change to another PR if preferred.

btw, I'm running the experiment with SHS100k again right now to make sure the fix works. Will post update here in an hour.

wissam-sib: The fix did the trick! Looks good now

```
metric                 random LCO-Embedding-Omni-3B
ndcg_at_1             0.00098        0.86133
ndcg_at_5             0.03085        0.83748
ndcg_at_10            0.03646        0.81389
ndcg_at_20            0.03999        0.76428
ndcg_at_100           0.06952        0.74829
map_at_10             0.00260        0.30095
recall_at_1           0.00005        0.03525
recall_at_5           0.00487        0.16872
recall_at_10          0.01014        0.31058
recall_at_20          0.02067        0.46567
mrr_at_10             0.07594        0.88561
```

Samoed (approved):

PR #4545, opened 2026-04-28
model: Add DenseOn/LateOn definitions

NohTow:
If you add a model or a dataset, please add the corresponding checklist:

This is a PR to add the definition of LateOn/DenseOn models.
Collection: https://huggingface.co/collections/lightonai/denseon-and-lateon
Blogpost: https://huggingface.co/blog/lightonai/denseon-lateon
Corresponding results PR: https://github.com/embeddings-benchmark/results/pull/498

As detailed in the BP, the training data is a reproduction of the data we release that is Apache 2.0 compatible.
I did put as overlap for the zero-shot computation, the list of what I find overlapping with the public data for fairness.

Feel free to tell me if I missed some, but I discussed with @Samoed about it!

PR #5048, opened 2026-07-30
model: Add mDenseOn/mLateOn definitions

NohTow:
This is a PR to add the definition of the mLateOn/mDenseOn models, the multilingual extension of LateOn/DenseOn (#4545) trained with translate-train on eight target languages plus English.

Collection: https://huggingface.co/collections/lightonai/mdenseon-and-mlateon
Blogpost: https://huggingface.co/blog/lightonai/mDenseOn-mLateOn
Paper: https://arxiv.org/abs/2607.27178
Corresponding results PR: https://github.com/embeddings-benchmark/results/pull/653

A few notes:
- As with LateOn/DenseOn, the training data is a reproduction of the data we publicly release that is Apache 2.0 compatible. For the zero-shot computation, I listed as overlap what I find overlapping with the public data for fairness (same list as #4545 for the English seed, since the multilingual pre-training corpus is a translate-train extension of it).
- On top of the translated fine-tuning data, the models are fine-tuned on the training splits of MIRACL, MLDR, and the LateOn-Code data (which covers the MTEB code tasks), so those are listed in `training_datasets`. For CodeEditSearch, we built a training set from CommitPackFT and decontaminated it against the evaluation set (shared commit SHAs, exact normalized-text matches, and ≥50% 13-gram containment), but I still list it as overlap to be conservative.
- `languages` lists English plus the eight translate-train target languages (French, German, Italian, Spanish, Portuguese, Swedish, Norwegian, Arabic). The models transfer well beyond these (see the blog post), but I only listed the languages actually present in retrieval training.
- The backbone is [mmBERT-base](https://huggingface.co/jhu-clsp/mmBERT-base) (307M parameters, 8192 context).

cc @Samoed

### DIFF
--- a/mteb/models/model_implementations/pylate_models.py
+++ b/mteb/models/model_implementations/pylate_models.py
@@ -548,6 +548,21 @@ topk_io__iso_moderncolbert = ModelMeta(
     extra_requirements_groups=["pylate"],
 )
 
+# Code fine-tuning data from LateOn-Code (training splits)
+late_on_code_fine_tuning_data = {
+    "AppsRetrieval",
+    "SyntheticText2SQL",
+    "CosQA",
+    "CodeFeedbackMT",
+    "CodeFeedbackST",
+    "StackOverflowQA",
+    "CodeTransOceanContest",
+    "CodeTransOceanDL",
+    "CodeSearchNetRetrieval",
+    "CodeSearchNetCCRetrieval",
+    "COIRCodeSearchNetRetrieval",
+}
+
 late_on_code_citation = """@misc{LateOn-Code,
   title  = {LateOn-Code: a Family of State-Of-The-Art Late Interaction Code Retrieval Models},
   author = {Chaffin, Antoine},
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/pylate_models.py
```diff
@@ -1020,3 +1004,108 @@ def _encode(
...
+    "ruby-Code",
+    "php-Code",
+]
+
+mdenseon_mlateon_citation = r"""@misc{sourty2026mdenseonmlateon,
+  title={{mDenseOn with the mLateOn}: Open Multilingual, Long-Context, and Code Retrieval Models},
+  author={Sourty, Raphael and Chaffin, Antoine and Moura Junior, Paulo Roberto and Chatelain, Amelie},
+  year={2026},
+  howpublished={\url{https://huggingface.co/blog/lightonai/mDenseOn-mLateOn}},
+}"""
```
Samoed: Maybe change to https://arxiv.org/abs/2607.27178?

Samoed (approved):

mteb/models/model_implementations/pylate_models.py
NohTow: Good catch!
I swapped it everywhere but here :')
Updated!

PR #5037, opened 2026-07-27
Declare `training_datasets` for `minetta/nemotron-3-embed-8b-legal`

banyaneth:
The entry for `minetta/nemotron-3-embed-8b-legal` has `training_datasets=None`, so the leaderboard shows N/A in the zero-shot column: `get_training_datasets()` returns `None` and `_zero_shot_pct_cached` short-circuits before computing.

Fine-tuning sources are ECtHR case law, GDPR enforcement decisions, CaseHOLD, SEC EDGAR agreements, public terms-of-service text, Australian tax guidance and UK legislation. None of those is an MTEB dataset, so the declared set is empty and the base model's training data is inherited via `adapted_from="nvidia/Nemotron-3-Embed-8B-BF16"`.

With this change the entry resolves to the 59 datasets inherited from the base, with no overlap against MTEB(Law, v1), so the column computes rather than reading N/A.

A per-document containment audit of every training input against the evaluated corpora is linked from the comment and published in [CONTAMINATION.md](https://huggingface.co/minetta/nemotron-3-embed-8b-legal/blob/main/CONTAMINATION.md).

PR #5051, opened 2026-07-30
Add ModelMeta: minetta/nemotron-3-embed-8b-medical

banyaneth:
Adds a ModelMeta entry for [minetta/nemotron-3-embed-8b-medical](https://huggingface.co/minetta/nemotron-3-embed-8b-medical), a medical adaptation of nvidia/Nemotron-3-Embed-8B-BF16.

Weights are unchanged from the base. The loader applies a fixed orthonormal projection when
clustering and passes every other task through, keyed on task name the same way
`model_prompts` is. Output stays 4096-dimensional, so `embed_dim` holds for every task.

`processor_kwargs` sets `model_max_length` to 4096, the same as the base entry in
`nvidia_models.py`. The checkpoint advertises a 32k window and `max_tokens` reflects that, but
TRECCOVID and TRECCOVID-PL hold 122k-character documents and running those uncapped does not
fit in 80GB.

`training_datasets` is an empty set rather than None. No MTEB dataset or training split was
used at any stage. The projection is fitted on PubMed abstracts that were checked against all
twelve MTEB(Medical, v1) corpora first; the gate, the counts and what it removed are written up
in [CONTAMINATION.md](https://huggingface.co/minetta/nemotron-3-embed-8b-medical/blob/main/CONTAMINATION.md).

All twelve tasks, run through the loader and `loader_kwargs` in this diff, single H100,
batch 16. `mteb.get_model` and `mteb.get_model_meta` were checked separately against this
branch, along with the projection firing on the two clustering tasks and output staying
4096-dimensional:

| task | score |
|---|---|
| PublicHealthQA | 90.88 |
| MedicalQARetrieval | 87.16 |
| TRECCOVID | 86.78 |
| TRECCOVID-PL | 85.17 |
| SciFact | 83.33 |
| SciFact-PL | 80.71 |
| CMedQAv2-reranking | 79.91 |
| CUREv1 | 66.92 |
| MedrxivClusteringP2P.v2 | 42.77 |
| NFCorpus | 42.32 |
| MedrxivClusteringS2S.v2 | 41.96 |
| CmedqaRetrieval | 37.64 |
| **Mean(Task)** | **68.80** |

Results go to the results repo once this merges. That repo installs `mteb` from main, and its
`test_all_models_are_implemented_in_mteb` check needs this entry to exist first.

- [x] I have tested the implementation works on a representative set of tasks. All twelve
  MTEB(Medical, v1) tasks were run through it, table above.
- [ ] I reproduced results from the original paper (if applicable) on at least one benchmark, and I am including the results in the PR description. No paper.

### DIFF
--- a/mteb/models/model_implementations/minetta_models.py
+++ b/mteb/models/model_implementations/minetta_models.py
@@ -1,4 +1,12 @@
-from mteb.models import ModelMeta, sentence_transformers_loader
+from __future__ import annotations
+
+import numpy as np
+
+from mteb.models import (
+    ModelMeta,
+    SentenceTransformerEncoderWrapper,
+    sentence_transformers_loader,
+)
 from mteb.models.model_meta import ScoringFunction
 
 nemotron_3_embed_8b_legal = ModelMeta(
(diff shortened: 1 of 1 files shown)


banyaneth:
The test_lowest failure here isn't from this PR. Same test fails on main:

- main 136b3995, run 30568455289: `test_hps_collapses_spurious_ties[dtype0]`, `clamp_min_scalar_cpu not implemented for 'Half'`
- same test, same error on this branch
- main was green on ce4fb38a about 15 min earlier

Looks like it needs an xfail like the other Half/CPU cases in that file. Ping me and I'll rebase.

banyaneth:
Windows 3.10 now fails the same test, dtype1 instead of dtype0.

It came in with #4933 (4b4f5de), which added `test_hps_collapses_spurious_ties`. It fails
anywhere float16 lands on CPU, which is what the other 64 xfails in that file cover. The other
open PRs look green because their runs predate that commit.

Samoed:
I fixed this test in https://github.com/embeddings-benchmark/mteb/pull/5053

banyaneth:
Thank you! Will rebase once it lands

mteb/models/model_implementations/minetta_models.py
```diff
@@ -21,7 +29,89 @@
...
+    modalities=["text"],
+    model_type=["dense"],
+    citation=None,
+    contacts=None,
+)
+
+_HEADS = {
+    "MedrxivClusteringS2S.v2": "clustering_head/head_s2s.npz",
+    "MedrxivClusteringP2P.v2": "clustering_head/head_p2p.npz",
+}
```
Samoed, line 44: You shoudn't load anything during evaluation. We have wrappers to cache model embeddings https://docs.mteb.org/get_started/advanced_usage/cache_embeddings/

banyaneth: Moved it into the loader, so both projections load once when the model is built and nothing is fetched during encode. Scores are unchanged.

banyaneth:
Split the legal `training_datasets` change out of this PR so it can be sorted separately in #5037. This one is just the medical entry now.

mteb/models/model_implementations/minetta_models.py
Samoed: I'm not sure that adding adding task specific embeddings is good idea

isaac-chung: Just wondering why not? I don't think we have limitations on that in our contribution guidelines.

isaac-chung: Could we @KennethEnevoldsen and @Samoed get an agreement here? Happy to go either way.

KennethEnevoldsen: While domain specific head routing would be totally fine (it is essentially a fancy prompt), I would argue that this should be a part of the model itself (included in parameters). So no need to have this added complexity I would just wrap it into the model itself. This would allow us to just use sentence transformers and it would also simplify the model usage for others.

(@Samoed I think you might have misunderstood this as a cached embedding, it is model parameters)

isaac-chung:
> While domain specific head routing would be totally fine (it is essentially a fancy prompt), I would argue that this should be a part of the model itself (included in parameters). So no need to have this added complexity I would just wrap it into the model itself. This would allow us to just use sentence transformers and it would also simplify the model usage for others.
> 
> (@Samoed I think you might have misunderstood this as a cached embedding, it is model parameters)

Agreed @KennethEnevoldsen , that we do not want this complexity to be maintained here. Closing this now. Feel free to re-open or open a new PR once the integrated model is available.

PR #5057, opened 2026-07-31
model: add Qwen3-VL-Reranker (2B, 8B)

### LINKED ISSUE #3951: Add model: Qwen/Qwen3-VL-Reranker-8B

https://huggingface.co/Qwen/Qwen3-VL-Reranker-8B

### Arxiv link

https://arxiv.org/abs/2601.04720

### Description of the model

Ref https://github.com/embeddings-benchmark/mteb/issues/3895

hubielu:
Adds `Qwen/Qwen3-VL-Reranker-2B` and `-8B` as cross-encoders. Closes #3951.

Ships as a standard `sentence_transformers.CrossEncoder`, so `CrossEncoderWrapper` is reused directly — its multimodal input collection already produces the `{text, image}` dicts these models expect. The subclass supplies the task-level instruction, gated behind `use_instructions` (default `True`). Metadata via `ModelMeta.from_hub()`; languages from the model card (33).

### Validation

Two-stage retrieval → rerank on `VidoreShiftProjectRetrieval` (100 queries / 999 image docs, A6000, bf16):

| | NDCG@10 |
| --- | --- |
| `Qwen3-VL-Embedding-2B` | 0.8403 |
| + reranker (top_k=100) | **0.8949** (+0.055) |

AskUbuntuDupQuestions 0.64856 · HUMECore17InstructionReranking 0.99028

The model card example can't serve as a parity check: its scores are multiples of 1/16, the bf16 step at these logits is 0.0625, and the text-only document — no image in it — alone spans 1.6875 to 1.875 across attention kernels and batch sizes. fp32 is stable across machines (1.82623 / 0.72302 / 1.46066, confirmed by @tomaarsen) but 0.16 above published on image-only. Details in the thread.

### Paper reproduction

Authors' MMEB-v2 harness on an A6000: **Qwen3-VL-Embedding-2B IMG_RET = 74.9 vs 74.8 in Table 2** across all 12 image retrieval datasets.

The reranker stage runs ~52 s/query at top-100 on MMEB images (batch 16, bf16), ≈173 GPU-hours for the Image aggregate. Table 5 reports only that aggregate (74.0), so a partial run has nothing to compare against. Vidore document scans are cheaper at ~23 s/query.

Two upstream issues, neither affecting mteb: the checkpoint's root-level `reranker.jinja` breaks `AutoProcessor` remote resolution (workaround: local snapshot path), and transformers v5 breaks `get_rope_index` in the official Qwen eval repo's code path — the `CrossEncoder` path is unaffected, re-confirmed on 5.15.0.

### DIFF
--- a/mteb/models/model_implementations/qwen3_vl_reranker.py
+++ b/mteb/models/model_implementations/qwen3_vl_reranker.py
@@ -1,0 +1,170 @@
+from __future__ import annotations
+
+import unicodedata
+from typing import TYPE_CHECKING, Any
+
+from mteb.models.model_implementations.qwen3_vl_embedding_models import (
+    QWEN3_VL_EMBEDDING_CITATION,
+)
+from mteb.models.model_meta import ModelMeta
+from mteb.models.sentence_transformer_wrapper import CrossEncoderWrapper
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+
+DEFAULT_INSTRUCTION = "Retrieve images or text relevant to the user's query."
+
+
+class Qwen3VLRerankerWrapper(CrossEncoderWrapper):
+    """Wrapper for the Qwen3-VL-Reranker series.
+
+    The reranker scores a (query, document) pair jointly, where either side may
+    be text, an image, a video, or a mixture of these. It is exposed as a
+    standard `sentence_transformers.CrossEncoder`, so the multimodal input
+    collection in `CrossEncoderWrapper` is reused as-is; this subclass only
+    supplies the task-level instruction.
+
+    Reference implementation: https://github.com/QwenLM/Qwen3-VL-Embedding
+    """
+
+    def __init__(
+        self,
+        model: str,
+        revision: str | None = None,
+        device: str | None = None,
+        min_pixels: int | None = None,
+        max_pixels: int | None = None,
+        fps: float | None = 2.0,
+        max_frames: int | None = 64,
+        num_frames: int | None = None,
+        **kwargs: Any,
+    ) -> None:
+        # Default to the checkpoint's own preprocessor_config.json. Forcing a
+        # pixel budget changes how images are resized and measurably shifts
+        # scores away from the reference implementation, so it is opt-in only
+        # (useful if you need to bound memory on a small GPU).
+        processor_kwargs = dict(kwargs.pop("processor_kwargs", {}) or {})
+        if min_pixels is not None:
+            processor_kwargs.setdefault("min_pixels", min_pixels)
+        if max_pixels is not None:
+            processor_kwargs.setdefault("max_pixels", max_pixels)
+        if processor_kwargs:
+            kwargs["processor_kwargs"] = processor_kwargs
+
+        super().__init__(
+            model,
+            revision=revision,
+            device=device,
+            fps=fps,
+            max_frames=max_frames,
+            num_frames=num_frames,
+            **kwargs,
+        )
+
+    @staticmethod
+    def _normalize_instruction(instruction: str) -> str:
+        instruction = instruction.strip()
+        # Mirrors Qwen3VLEmbeddingWrapper: append a period if the instruction
+        # does not already end in punctuation.
+        if instruction and not unicodedata.category(instruction[-1]).startswith("P"):
+            instruction += "."
+        return instruction
+
+    def get_task_instruction(
+        self,
+        task_metadata: TaskMetadata,
+        prompt_type: PromptType | None = None,
+    ) -> str:
+        """Resolve the single task-level instruction passed to the reranker.
+
+        Unlike a bi-encoder, the reranker consumes the query and the document in
+        one forward pass, so it takes one instruction describing the relevance
+        criterion rather than a per-side prefix. When a task defines separate
+        query/document prompts, the query-side prompt is used.
+        """
+        prompt = task_metadata.prompt
+        if isinstance(prompt, dict):
+            prompt = prompt.get("query") or next(iter(prompt.values()), None)
+        if not prompt:
+            return DEFAULT_INSTRUCTION
+        return self._normalize_instruction(prompt)
+
+    def predict(
+        self,
+        inputs1: DataLoader[BatchedInput],
+        inputs2: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> Array:
+        kwargs.setdefault(
+            "prompt", self.get_task_instruction(task_metadata, prompt_type)
+        )
+        return super().predict(
+            inputs1,
+            inputs2,
+            task_metadata=task_metadata,
+            hf_split=hf_split,
+            hf_subset=hf_subset,
+            prompt_type=prompt_type,
+            **kwargs,
+        )
+
+
+qwen3_vl_reranker_2b = ModelMeta(
+    loader=Qwen3VLRerankerWrapper,
+    name="Qwen/Qwen3-VL-Reranker-2B",
+    model_type=["cross-encoder"],
+    languages=None,  # supports 30+ languages
+    open_weights=True,
+    revision="4bd860ac4f15ad1897a214615cccc700f8f71818",
+    release_date="2026-01-07",
+    modalities=["image", "text", "video"],
+    n_parameters=2_127_532_032,
+    n_embedding_parameters=311_164_928,
+    memory_usage_mb=4058,
+    embed_dim=None,  # not applicable: outputs a scalar relevance score
+    license="apache-2.0",
+    max_tokens=32768,
+    reference="https://huggingface.co/Qwen/Qwen3-VL-Reranker-2B",
+    similarity_fn_name=None,
+    framework=["Sentence Transformers", "PyTorch", "Transformers", "safetensors"],
+    use_instructions=True,
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
+    citation=QWEN3_VL_EMBEDDING_CITATION,
+    extra_requirements_groups=["multimodal-sbert"],
+)
+
+qwen3_vl_reranker_8b = ModelMeta(
+    loader=Qwen3VLRerankerWrapper,
+    name="Qwen/Qwen3-VL-Reranker-8B",
+    model_type=["cross-encoder"],
+    languages=None,  # supports 30+ languages
+    open_weights=True,
+    revision="b212dc8c91a8164aef1ea2de9c1a867611e75c04",
+    release_date="2026-01-07",
+    modalities=["image", "text", "video"],
+    n_parameters=8_767_123_696,
+    n_embedding_parameters=622_329_856,
+    memory_usage_mb=16722,
+    embed_dim=None,  # not applicable: outputs a scalar relevance score
+    license="apache-2.0",
+    max_tokens=32768,
+    reference="https://huggingface.co/Qwen/Qwen3-VL-Reranker-8B",
+    similarity_fn_name=None,
+    framework=["Sentence Transformers", "PyTorch", "Transformers", "safetensors"],
+    use_instructions=True,
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
+    citation=QWEN3_VL_EMBEDDING_CITATION,
+    extra_requirements_groups=["multimodal-sbert"],
+)
--- a/mteb/models/model_implementations/qwen3_vl_reranker.py
+++ b/mteb/models/model_implementations/qwen3_vl_reranker.py
@@ -62,7 +62,8 @@ class Qwen3VLRerankerWrapper(CrossEncoderWrapper):
     be text, an image, a video, or a mixture of these. It is exposed as a
     standard `sentence_transformers.CrossEncoder`, so the multimodal input
     collection in `CrossEncoderWrapper` is reused as-is; this subclass only
-    supplies the task-level instruction.
+    supplies the task-level instruction. Pass `use_instructions=False` to
+    score pairs without one.
 
     Reference implementation: https://github.com/QwenLM/Qwen3-VL-Embedding
     """
@@ -77,6 +78,7 @@ class Qwen3VLRerankerWrapper(CrossEncoderWrapper):
         fps: float | None = 2.0,
         max_frames: int | None = 64,
         num_frames: int | None = None,
+        use_instructions: bool = True,
         **kwargs: Any,
     ) -> None:
         # Default to the checkpoint's own preprocessor_config.json. Forcing a
@@ -100,6 +102,7 @@ class Qwen3VLRerankerWrapper(CrossEncoderWrapper):
             num_frames=num_frames,
             **kwargs,
         )
+        self.use_instructions = use_instructions
 
     @staticmethod
     def _normalize_instruction(instruction: str) -> str:
@@ -140,9 +143,10 @@ class Qwen3VLRerankerWrapper(CrossEncoderWrapper):
         prompt_type: PromptType | None = None,
         **kwargs: Any,
     ) -> Array:
-        kwargs.setdefault(
-            "prompt", self.get_task_instruction(task_metadata, prompt_type)
-        )
+        if self.use_instructions:
+            kwargs.setdefault(
+                "prompt", self.get_task_instruction(task_metadata, prompt_type)
+            )
         return super().predict(
             inputs1,
             inputs2,
--- a/mteb/models/model_implementations/qwen3_vl_reranker.py
+++ b/mteb/models/model_implementations/qwen3_vl_reranker.py
@@ -18,6 +18,43 @@ if TYPE_CHECKING:
 DEFAULT_INSTRUCTION = "Retrieve images or text relevant to the user's query."
 
 
+QWEN3_VL_LANGUAGES = [
+    "eng-Latn",
+    "cmn-Hans",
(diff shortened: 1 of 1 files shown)


Samoed:
> Reproducing the published example

Which task are these scores from?

Samoed:
> Happy to add a first-stage-retrieval → rerank path as a follow-up if that's useful infra beyond this model

We already support to stage retrieval https://docs.mteb.org/get_started/advanced_usage/two_stage_reranking/

hubielu:
> > Reproducing the published example
> 
> Which task are these scores from?

@Samoed Not from an mteb task — sorry, that section wasn't clear. Those are the raw relevance scores from the worked example on the model card itself (one query against three documents: text-only, an image URL, and a `{text, image}` dict), published there as `[1.8125, 0.5625, 1.3125]`.

I ran that exact example through `mteb.get_model(...)` and called `.predict()` on it, so it's checking that loading the model via mteb reproduces the reference implementation's own numbers, and that the image and text+image paths both work. Not a benchmark result.

The actual task result is AskUbuntuDupQuestions at 0.64856 (main_score), further down.

I'll relabel that section to make the distinction obvious.

mteb/models/model_implementations/qwen3_vl_reranker.py
```diff
@@ -0,0 +1,170 @@
...
+            prompt_type=prompt_type,
+            **kwargs,
+        )
+
+
+qwen3_vl_reranker_2b = ModelMeta(
+    loader=Qwen3VLRerankerWrapper,
+    name="Qwen/Qwen3-VL-Reranker-2B",
+    model_type=["cross-encoder"],
+    languages=None,  # supports 30+ languages
```
isaac-chung: Please fill this from https://huggingface.co/Qwen/Qwen3-VL-Reranker-2B

mteb/models/model_implementations/qwen3_vl_reranker.py
```diff
@@ -0,0 +1,170 @@
...
+    Reference implementation: https://github.com/QwenLM/Qwen3-VL-Embedding
+    """
+
+    def __init__(
+        self,
+        model: str,
+        revision: str | None = None,
+        device: str | None = None,
+        min_pixels: int | None = None,
+        max_pixels: int | None = None,
```
isaac-chung, line 77: Could we use these values https://github.com/QwenLM/Qwen3-VL-Embedding#model-initialization-parameters and see if the model card example inference values match better?

Samoed:
@hubielu Can you reproduce scores from paper?

mteb/models/model_implementations/qwen3_vl_reranker.py
hubielu: Done! Filled with the 33 languages from the Qwen3-VL README, using the same ISO codes as `qwen3_models.py`.

hubielu:
> @hubielu Can you reproduce scores from paper?

@Samoed Partly — here's how far I got.

I ran the authors' own MMEB-v2 harness (QwenLM/Qwen3-VL-Embedding) on an A6000. The first stage reproduces cleanly: **Qwen3-VL-Embedding-2B IMG_RET = 74.9 against 74.8 in Table 2**, over all 12 image retrieval datasets.

| | MSCOCO_i2t | VisualNews_i2t | VisDial | MSCOCO_t2i | VisualNews_t2i | WebQA |
|---|---|---|---|---|---|---|
| | 74.7 | 75.6 | 81.5 | 77.6 | 72.9 | 90.1 |

| | EDIS | Wiki-SS-NQ | CIRR | NIGHTS | OVEN | FashionIQ |
|---|---|---|---|---|---|---|
| | 92.5 | 83.9 | 65.0 | 69.6 | 76.7 | 38.9 |

The reranker stage I could not finish. At top-100 with batch 16 in bf16 it runs ~52 s/query, so one 1000-query dataset is ~14.5 GPU-hours and the full 12-dataset Image aggregate is ~173. Table 5 reports only the aggregate (74.0), with no per-dataset reranker numbers, so a 1–2 task run would have nothing to compare against, and I'd rather report the wall than an incomparable figure. Happy to run it if anyone has spare compute.

Two upstream issues worth recording:

1. `AutoProcessor.from_pretrained("Qwen/Qwen3-VL-Reranker-2B")` fails with `TypeError: expected str, bytes or os.PathLike object, not NoneType`. transformers resolves extra jinja templates under `chat_template/{name}.jinja`, but this checkpoint has `reranker.jinja` at the repo root, so the path resolves to `None`. Passing a local `snapshot_download` path works, since the local branch globs instead. Reproduced on 4.57.3 through 4.57.6.
2. transformers v5 loads the processor fine but breaks in the forward pass: `get_rope_index` raises `TypeError: only integer tensors of a single element can be converted to an index`. 4.57.6 is the last working version.

I also re-did the model-card reproduction on a controlled baseline — the reference is bfloat16, and text and text+image now match exactly. The updated PR description has the details.

mteb/models/model_implementations/qwen3_vl_reranker.py
hubielu: Good prompt to re-check, as my earlier result was measured on a confounded baseline.

The published scores are exact multiples of 1/16, a bfloat16 signature, and the README's commented-out `torch_dtype=torch.bfloat16` confirms it. I'd been running float16 on MPS, where nothing matched. On CPU in bf16:

| config | text only | image only | text + image |
| --- | --- | --- | --- |
| published | **1.8125** | **0.5625** | **1.3125** |
| bf16, checkpoint defaults | **1.8125** | 0.6875 | **1.3125** |
| bf16, `min_pixels=4096` / `max_pixels=1843200` | 1.6875 | 0.7500 | 1.4375 |
| fp32, checkpoint defaults | 1.7737 | 0.6539 | 1.3663 |
| fp16/MPS (original) | 1.8281 | 0.7578 | 1.4609 |

Text and text+image reproduce exactly on the checkpoint's own `preprocessor_config.json`; the override moves all three away, including text-only, which contains no image. Worth noting the linked values are `Qwen3VLEmbedder`'s — the official `Qwen3VLReranker(...)` example passes only `model_name_or_path`.

Keeping checkpoint defaults, these opt-in for memory-bounded GPUs. Image-only is still 0.125 high; dtype, pixel budget and batch composition are ruled out, leaving `attn_implementation` (reference uses flash_attention_2, CUDA-only) as the untested candidate.

hubielu:
The red check on `a0fae5f` is `test_similarity_functions.py::test_hps_collapses_spurious_ties[dtype1]` on `windows-latest, 3.10`, low-precision `cos_sim` over randomly generated tensors, unrelated to this PR and green on all six Ubuntu jobs.

Samoed:
> transformers v5 loads the processor fine but breaks in the forward pass: get_rope_index raises TypeError: only integer tensors of a single element can be converted to an index. 4.57.6 is the last working version.

Can you pin it then and create issue in models repo

> The reranker stage I could not finish. At top-100 with batch 16 in bf16 it runs ~52 s/query, so one 1000-query dataset is ~14.5 GPU-hours and the full 12-dataset Image aggregate is ~173. Table 5 reports only the aggregate (74.0), with no per-dataset reranker numbers, so a 1–2 task run would have nothing to compare against, and I'd rather report the wall than an incomparable figure. Happy to run it if anyone has spare compute.

Why is it possible to run in MMEB, but not your implementation?

hubielu:
> Can you pin it then and create issue in models repo

The v5 break is in the official Qwen eval repo's code path (`Qwen3VLForConditionalGeneration` → their `compute_scores`), not mteb's. Our wrapper goes through `sentence_transformers.CrossEncoder`, which doesn't hit it, just re-confirmed on transformers 5.14.1, image document, score `0.6875`, identical to 4.57.6. So I think that no pin is needed, but I'd be happy to add one if you'd prefer.

For the issues: `get_rope_index` is a transformers bug. The jinja one is arguably Qwen's checkpoint layout, `reranker.jinja` sits at the repo root rather than under `chat_template/`, where transformers looks when resolving from a repo id. Which repo do you want each filed on? I'll link them here.

> Why is it possible to run in MMEB, but not your implementation?

It is possible, the blocker is compute, not the implementation. Two-stage runs fine in mteb via `convert_to_reranking(top_k=100)`:

| task | retriever only | + reranker |
| --- | --- | --- |
| AILAStatutes | 0.5545 | 0.4815 |
| TwitterHjerneRetrieval | 0.7322 | 0.7084 |

I used the official harness only because the paper's numbers are on MMEB-v2 datasets, which aren't in mteb, so a mteb-native run has nothing in Table 2 or 5 to compare to. The ~52 s/query at top-100 with images is the model's cost either way; ~173 GPU-hours for the 12-dataset Image aggregate.

Samoed:
> I used the official harness only because the paper's numbers are on MMEB-v2 datasets, which aren't in mteb

We have vidore datasets that should be same as in mmeb

hubielu:
> We have vidore datasets that should be same as in mmeb

Good point, I'd only checked MMEB's image datasets. Full v1 is ~3,871 queries ≈ 56 GPU-h at the measured rate, more than I can fund. The five ~100-query subsets (SyntheticDocQA ×4 + ShiftProject) are ~500 queries ≈ 7 GPU-h. Not comparable to the 90.5 aggregate, but it'd show the delta's direction and size. Want me to run that?

mteb/models/model_implementations/qwen3_vl_reranker.py
isaac-chung: So there's still a gap for the image only example using the default initialization params from the QwenLM repo?

hubielu: Yes, still 0.6875 vs 0.5625, but flash_attention_2 isn't the cause. A6000, bf16, checkpoint defaults:

| config | text only | image only | text + image |
| --- | --- | --- | --- |
| published | 1.8125 | 0.5625 | 1.3125 |
| CPU bf16 | 1.8125 | 0.6875 | 1.3125 |
| A6000 `sdpa` | 1.6875 | 0.6875 | 1.3750 |
| A6000 `eager` | 1.8125 | 0.6250 | 1.3750 |
| A6000 `flash_attention_2` | 1.7500 | 0.6875 | 1.4375 |

fa2 leaves image-only unchanged and moves the other two further from published.

I believe the text-only column is the tell: that document has no image, yet kernel choice alone spans 1.6875–1.8125. Two ticks, the same size as the image-only gap. fp32 agrees (−0.039 / +0.091 / +0.054, all one grid step). The example can't resolve differences this small.

isaac-chung: Odd. I tried running the [sentence transformers example](https://huggingface.co/Qwen/Qwen3-VL-Reranker-2B#using-sentence-transformers) on CPU:

| config | text only | image only | text + image |
| --- | --- | --- | --- |
| published | 1.8125 | 0.5625 | 1.3125 |
| CPU bf16 | 1.75 | 0.6875 | 1.3750 |
| CPU fp32 | 1.75 | 0.6875 | 1.3750 |

with these packages install via `make install` from main:
```
sentence-transformers  5.5.1
transformers           5.10.2
torch                  2.11.0
torchvision            0.26.0
```

isaac-chung: Maybe @tomaarsen would be able to help rule out a sentence transformers issue? We're trying to repro this model's inference example but with not much luck.

hubielu:
@Samoed Ran `VidoreShiftProjectRetrieval` (100 queries / 999 image docs, A6000, bf16):

| | NDCG@10 |
| --- | --- |
| `Qwen3-VL-Embedding-2B` | 0.8403 |
| + `Qwen3-VL-Reranker-2B` (top_k=100) | **0.8949** (+0.055) |

So the reranker helps clearly in-domain. The earlier text-only regressions (AILAStatutes −0.073, TwitterHjerne −0.024) look like domain mismatch rather than a wiring issue.

Two notes:

`convert_to_reranking` crashes on this task — `load_data()` leaves `self.dataset` as `None` for vidore-style tasks, so `retrieval.py:727` raises `TypeError: 'NoneType' object is not iterable`. I reranked the stage-1 predictions JSON directly instead. Want an issue on the mteb repo?

Also, my recomputed stage-1 NDCG@10 was 0.8403 vs 0.8273 from `mteb.evaluate` on the same predictions (tie handling?), so absolutes may run slightly high — both rows use the same scorer, so the delta holds.

These scans run ~23 s/query, not the ~52 s/query I measured on MMEB, so the other four subsets are only ~2 GPU-h if useful.

mteb/models/model_implementations/qwen3_vl_reranker.py
```diff
@@ -0,0 +1,207 @@
...
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> Array:
+        kwargs.setdefault(
+            "prompt", self.get_task_instruction(task_metadata, prompt_type)
+        )
```
Samoed: I think better to add a flag for enabling/disabling instructions rather than modifing kwargs

mteb/models/model_implementations/qwen3_vl_reranker.py
tomaarsen: Thanks for the ping.
I had an agent have a look at this. As far as I can tell, it's not a Sentence Transformers issue. The wiring in this PR is correct, the model card example just sits below the resolution of bfloat16.

Here's my agent's findings in more detail:

## fp32 is version invariant

ST 5.4.1 and 5.7.0, transformers 5.5.4 and 5.14.1, torch 2.10.0 and 2.11.0 (CUDA, sdpa) all give the
same fp32 answer to five decimals, and batching does not change it:

| | text only | image only | text+image |
| --- | --- | --- | --- |
| fp32, all six combinations | 1.82623 | 0.72302 | 1.46066 |

So nothing regressed. The prompt ST builds also matches the checkpoint's own `format_mm_instruction`
character for character (1247 image tokens, inside the intended 1280 budget), and image
preprocessing is byte identical from transformers 4.57.6 through 5.14.1.

## bf16 cannot resolve this example

The published scores are multiples of 1/16 because the logits are around 10, where the bf16 ULP is
exactly 0.0625. Every difference in this thread is one to three ULPs. Varying only the attention
kernel and the batch composition, on one machine (RTX 3090, bf16, transformers 5.14.1):

| config | text only | image only | text+image |
| --- | --- | --- | --- |
| published | **1.8125** | **0.5625** | **1.3125** |
| sdpa, one at a time | **1.8125** | 0.625 | 1.375 |
| sdpa, batch of 3 | 1.6875 | 0.625 | 1.375 |
| sdpa, batch of 2 | 1.75 | 0.6875 | 1.375 |
| eager, one at a time | 1.75 | **0.5625** | **1.3125** |
| eager, batch of 3 | **1.8125** | 0.625 | 1.375 |
| eager, batch of 2 | 1.875 | 0.6875 | 1.4375 |

Image only took 0.5625, 0.625, 0.6875 and 0.75 across my runs. That band contains the published
value, @hubielu's 0.6875 and the 0.75 from the pixel override. Every published number is reproducible
here, just not all three columns under one configuration. (Tom here: I'm not sure how I managed to get all 3 at the same time when I ran computed the values shown in the README, but I don't think it's an issue at all)

## Where the batch effect comes from

The checkpoint's reference implementation scores one pair at a time:

```python
for pair in pairs:
    inputs = self.tokenize([pair])   # padding=False
```

`CrossEncoder.predict` defaults to `batch_size=32`, so the 113 token text document gets left padded
to about 1362 next to the image documents, moving it from 1.8125 to 1.6875. That is most of what
@isaac-chung saw. `batch_size=1` restores it. This is not a padding bug: fp32 is batch invariant, so
the masking is correct. It is accumulated rounding over a sequence that got 12x longer.

For a deterministic check of the wiring, run the example in fp32 with `batch_size=1`. The real
validation stays `AskUbuntuDupQuestions` and the Vidore run.

- Tom Aarsen

Samoed: Oh, wow. Thanks for response!

mteb/models/model_implementations/qwen3_vl_reranker.py
hubielu: Done in 6bb8d35. Added `use_instructions: bool = True` on the constructor; `predict()` only injects the prompt when it's set. Named to match the existing `ModelMeta.use_instructions` field, which both metas already declare.

mteb/models/model_implementations/qwen3_vl_reranker.py
hubielu: Hey @tomaarsen I reproduced your fp32 numbers exactly on an A6000 (transformers 5.15.0,
torch 2.12.1, sdpa): 1.82623 / 0.72302 / 1.46066 to five decimals, and batch
invariant (bs=1 and bs=32 agree to one ULP on the last value). Independent
hardware and versions, so I'm treating the wiring as confirmed.

bf16 on the same machine:

| config | text only | image only | text + image |
| --- | --- | --- | --- |
| published | 1.8125 | 0.5625 | 1.3125 |
| bf16, bs=1 | 1.75 | 0.625 | 1.375 |
| bf16, bs=32 | 1.6875 | 0.6875 | 1.3125 |

So fp32 is stable and reproducible across machines, but sits
0.16 above published on image-only, so the published triple doesn't appear to be
an fp32 result either. I'll drop the model card table from the PR description
and leave AskUbuntuDupQuestions (0.64856) and the Vidore two-stage run
(0.8403 -> 0.8949 NDCG@10) as the validation.

hubielu:
Anything still blocking this?

mteb/models/model_implementations/qwen3_vl_reranker.py
isaac-chung: Thanks so much, @tomaarsen! Sorry for the late reply, this slipped through the cracks.

isaac-chung (approved):
Think we should be ok. Thanks @hubielu !

Issue #5070, opened 2026-08-02
Audio statistics and collator assume datasets 3.x Audio() shape


hubielu:
### Describe the bug

Two call sites assume that an audio column decodes to a dictionary of the form {"array": np.ndarray, "sampling_rate": int}, and perform no coercion before using it. In mteb/abstasks/_statistics_calculation.py the code calls audio["array"].tobytes(), and in mteb/models/modality_collators.py it calls torch.from_numpy(audio["array"]).

Both break on any audio column that is not a datasets 3.x style Audio() feature. A plain struct of the form {array, sampling_rate} is a reasonable and real-world way to store audio, and is used by at least one public dataset, mulab-mir/lp-music-caps-magnatagatune-3k. In that case datasets decodes array to a Python list rather than a numpy array, and the code raises AttributeError: 'list' object has no attribute 'tobytes'.

There is a second, related fragility. On datasets version 4 and above, an Audio() feature no longer decodes to a dictionary at all; it decodes to a torchcodec.AudioDecoder. This currently still works, but only because datasets itself ships a compatibility shim in datasets/features/_torchcodec.py that gives AudioDecoder a __getitem__ supporting the keys "array" and "sampling_rate". The audio extra in pyproject.toml is declared as datasets[audio] with no upper bound, so audio support is presently load-bearing on that shim rather than on an explicit contract.

### To reproduce

import numpy as np
from datasets import Dataset

ds = Dataset.from_dict({
    "audio": [{"array": np.random.randn(2205).astype("float32"), "sampling_rate": 22050}]
})
print(type(ds[0]["audio"]["array"]))
ds[0]["audio"]["array"].tobytes()

The first print shows <class 'list'> and the final line raises AttributeError. In practice this was hit by calling task.calculate_descriptive_statistics() on an Any2AnyRetrieval task whose audio column was stored as a plain struct.

### Additional information

A minimal fix would be to wrap both call sites in np.asarray(audio["array"], dtype=np.float32). That is one line in each file, and it decouples the code from the datasets version and from the exact storage format of the audio column. An alternative or complementary fix is to pin datasets<4 in the audio extra so the assumption is explicit rather than implicit. Found while adding a music retrieval dataset for issue #2683.

### Are you interested to contribute a fix for this bug?

Yes

Samoed:
> Both break on any audio column that is not a datasets 3.x style Audio() feature

We're expecting that datasets are cast to the audio feature and don't support other formats.

[...]

Issue #5069, opened 2026-08-02
laion/larger_clap_music returns near-constant text embeddings


hubielu:
### Describe the bug

The model laion/larger_clap_music is registered in mteb/models/model_implementations/clap_models.py, but its text tower returns essentially the same embedding regardless of input. Any retrieval or zero-shot result produced with this model is therefore meaningless. Comparing all three LAION CLAP checkpoints on identical code, the same transformers version, the same processor, and the same four unrelated input strings: for laion/larger_clap_music the CLS hidden-state standard deviation is 0.025, the pooler output standard deviation is 0.005, and the mean pairwise text-text cosine similarity is 0.9993. For laion/clap-htsat-unfused those same three numbers are 0.526, 0.329 and 0.180. For laion/larger_clap_general they are 0.567, 0.441 and 0.294. In other words, the two working checkpoints show roughly 20x more variance in their text representations, and produce sensibly distinct embeddings for distinct inputs. larger_clap_music collapses all inputs to a single direction. Expected behaviour is that unrelated captions embed to distinct vectors, as they do for the other two checkpoints.

### To reproduce

Run the following. Every entry of the printed similarity matrix comes out at approximately 1.0.

import torch
from transformers import ClapModel, ClapProcessor
from transformers.modeling_outputs import BaseModelOutputWithPooling

m = ClapModel.from_pretrained("laion/larger_clap_music").eval()
p = ClapProcessor.from_pretrained("laion/larger_clap_music")
ts = ["a heavy metal guitar solo", "a slow sad piano ballad",
      "a female opera singer", "fast electronic dance music"]
with torch.no_grad():
    o = m.get_text_features(**p(text=ts, return_tensors="pt", padding=True, truncation=True))
    e = o.pooler_output if isinstance(o, BaseModelOutputWithPooling) else o
e = torch.nn.functional.normalize(e, dim=-1)
print(e @ e.T)

Substituting laion/clap-htsat-unfused or laion/larger_clap_general in both from_pretrained calls gives normal, varied similarities.

### Additional information

[...]

PR #5071, opened 2026-08-02
dataset: add LPMusicCapsMTT A2T and T2A retrieval tasks

### LINKED ISSUE #2683: Add LP-MusicCaps dataset for retrieval
https://huggingface.co/datasets/mulab-mir/lp-music-caps-magnatagatune-3k
LLM-generated music captioning

hubielu:
Closes #2683.

This adds two tasks, LPMusicCapsMTTA2TRetrieval and LPMusicCapsMTTT2ARetrieval, built from LP-MusicCaps (Doh et al., ISMIR 2023), which provides LLM-generated pseudo-captions for MagnaTagATune music clips.

On the gap this fills: MTEB already includes MusicCaps, but those captions are human-written and the audio comes from AudioSet. LP-MusicCaps captions are LLM-generated from tags, over MagnaTagATune audio, with four differently-styled captions per clip. Having both lets the leaderboard separate model performance on human-written versus pseudo-generated music captions.

I used the test split only: 300 clips of 10 seconds at 22.05 kHz, and 1200 captions. The source train split of 3000 clips is available if a larger evaluation set is preferred.

Results, all hit_rate@5. The random-encoder baseline scores 0.0233 on the audio-to-text direction and 0.0192 on text-to-audio. laion/clap-htsat-unfused scores 0.2367 and 0.2317. laion/larger_clap_general scores 0.2100 and 0.2933. Working models beat random by roughly ten to fifteen times while staying well below saturation.

On licensing: the mulab-mir repackaging declares no license, but the captions originate from LP-MusicCaps, which is CC-BY-NC-4.0 per its project repository at https://github.com/seungheondoh/lp-music-caps. I have recorded cc-by-nc-4.0.

The datasets currently live under my Hugging Face namespace. Happy to have them moved into the mteb organization.

A note on model selection: I did not use laion/larger_clap_music, despite it being the music-specific checkpoint, because its text tower appears non-functional. It returns near-constant embeddings, with a mean pairwise text-text cosine of 0.9993 across unrelated inputs, versus 0.18 and 0.29 for the other two CLAP checkpoints on identical code, and it scores below random on this task at 0.0133 and 0.0167. This is a problem with the upstream checkpoint rather than with mteb (see #5069). The datasets store audio as a datasets Audio() feature, per the contract confirmed in #5070.

Dataset checklist:

- [x] I have tested that the dataset runs with the mteb package
- [x] I have run models on the task and added the results above. I used CLAP rather than a text-only model such as intfloat/multilingual-e5-small, because this is audio-text retrieval and a text-only encoder cannot embed the audio side.
- [x] I have checked that the performance is neither trivial nor random
- [x] I have considered the size of the dataset and used the 300-clip test split rather than the 3000-clip train split
- [ ] Reproducing scores from the original paper is not applicable here, as LP-MusicCaps reports captioning metrics rather than retrieval metrics


Samoed (commented):
Can you add script how did you transformed rhis dataset during upload

hubielu:
> Can you add script how did you transformed rhis dataset during upload

Added it at scripts/data/lp_music_caps/create_data.py, following the pattern in scripts/data/audiocaps_retrieval/.

Samoed (approved):

PR #4984, opened 2026-07-21
task: add CASTELLA audio moment retrieval (t2a)

dukesun99:
Part of the MOEB effort (#4842): first temporal-localization retrieval task family for the audio benchmark. 

Adds `CASTELLAAMRRetrieval` from CASTELLA ([arXiv 2511.15131](https://arxiv.org/abs/2511.15131)), the human-annotated long-audio benchmark used as the DCASE 2026 Task 6 evaluation set: given a caption describing a moment, retrieve the full 1–5 minute recording that contains it. Corpus is the 566 complete recordings (60–300s), 1,347 caption queries, each with its source recording as the single relevant document. Data at [dukesun99/CASTELLA-AMR](https://huggingface.co/datasets/dukesun99/CASTELLA-AMR).

Dataset checklist:
- [x] Fills an existing gap: CASTELLA is not in mteb
- [x] Tested that the dataset runs with the mteb package
- [x] `mteb/baseline-random-encoder`: 0.77 nDCG@10
- [x] Real model — laion/clap-htsat-fused: 9.67 nDCG@10 (R@10 18.1). 
- [x] Corpus at evaluation scale
- [x] Paper-score reproduction: not applicable by construction — see Reproduction below

**Reproduction.** Not applicable — different task type, not just a different setting. The paper measures Audio Moment Retrieval: a DETR-style model regresses a continuous `[start, end]` interval within a recording, scored by Recall@1 at IoU (R1@0.7). This task adapts it to whole-recording retrieval (which recording contains the described moment), which the paper does not measure — R1@IoU is a localization metric with no ranking analogue, so no mteb model+metric matches it. The data (recordings + captions) ports unchanged; only the framing (moment localization → recording retrieval) is the adaptation.

PR #4967, opened 2026-07-19
task: add MomentSeeker composed video retrieval (it2v, vt2v)

dukesun99:
Part of the MOEB effort (#4842), missing modality combinations "image + text → video" and "video + text → video". Closes #4944. Adds the `it2v` TaskCategory code.

Adds `MomentSeekerTI2VRetrieval` and `MomentSeekerTV2VRetrieval` from MomentSeeker ([arXiv 2502.12558](https://arxiv.org/abs/2502.12558), NeurIPS 2025, CC-BY-NC-SA-4.0): long videos are split into 30-second non-overlapping 360p chunks (the corpus); a chunk is relevant iff it overlaps an annotated answer interval by at least half of the shorter of chunk and interval. 400 composed queries per direction. Data at [dukesun99/MomentSeeker-CVR](https://huggingface.co/datasets/dukesun99/MomentSeeker-CVR).

Dataset checklist:
- [x] Fills an existing gap: first composed video retrieval tasks in mteb
- [x] Tested that the dataset runs with the mteb package
- [x] `mteb/baseline-random-encoder`: ti2v 0.0010, tv2v 0.0028 nDCG@10
- [x] Real model — Qwen/Qwen3-VL-Embedding-2B: ti2v 0.5429, tv2v 0.5792. Clearly solvable but far from saturated
- [x] Corpus capped at ~3.1k chunks — evaluation scale
- [x] Paper-score reproduction: N/A — the paper evaluates moment localization; the chunk-corpus retrieval conversion is new

PR #5074, opened 2026-08-02
task: split MomentSeeker full-video retrieval into moment-level subtasks

dukesun99:
Part of the MOEB effort (#4842). Follow-up to the merged MomentSeeker tasks (#4967). Two changes, per the review discussion here:

1. **Full-video retrieval** (originally requested by @Samoed on #4984): the corpus is the complete source videos (avg >500s — movies, cartoons, egocentric, open-domain footage) rather than 30-second chunks. A video is relevant to a query iff it contains the annotated answer moment.
2. **Moment-level split into an aggregate task** (requested by @isaac-chung / @Samoed): instead of flat per-direction tasks, MomentSeeker is now split by the paper's [Section 3.3](https://arxiv.org/pdf/2502.12558) moment taxonomy into **9 leaf subtasks + aggregates**.

This **replaces** the flat `MomentSeekerTI2VRetrieval` / `MomentSeekerTV2VRetrieval`.

### Structure
- **9 leaf tasks** — `MomentSeeker{TI2V,TV2V,T2V}{Global,Event,Object}LevelRetrieval` (three query directions × three moment levels)
- **3 per-level aggregates** — `MomentSeeker{Global,Event,Object}LevelRetrieval` (over the three directions)
- **1 final aggregate** — `MomentSeekerRetrieval` (over the 9 leaves)

Query directions: `t2v` (text → video), `it2v` (image+text → video), `vt2v` (video+text → video). Moment levels from the source `task` field ([mapping](https://huggingface.co/datasets/avery00/MomentSeeker)): global = Causal Reasoning, Spatial Relation; event = Description Location, Action Recognition, Anomaly Detection; object = Object Recognition, Object Location, Attribute Recognition, OCR.

### Corpus
t2v draws on a larger pool of source videos than the composed directions (many anomaly clips the composed queries never use), so the corpus is **unified to all 271 source videos** across directions — every direction retrieves from the same corpus (re-encoded to 360p, audio kept). Data at [dukesun99/MomentSeeker-Full](https://huggingface.co/datasets/dukesun99/MomentSeeker-Full).

Query counts per (direction, level): ti2v/tv2v = 156 / 88 / 156 (global/event/object); t2v = 74 / 699 / 227.

### Baselines — Qwen/Qwen3-VL-Embedding-2B, map@5 (main_score)

| level | ti2v | tv2v | t2v |
|---|---|---|---|
| global | 0.734 | 0.769 | 0.507 |
| event | 0.699 | 0.728 | 0.439 |
| object | 0.726 | 0.745 | 0.290 |

Per-level aggregates: global 0.670 / event 0.622 / object 0.587. **Final `MomentSeekerRetrieval` 0.626.** `mteb/baseline-random-encoder` is ~0.01 across all leaves. The plain text→video direction is clearly hardest (especially object-level, 0.290) — no image/clip to anchor the query.

Dataset checklist:
- [x] Retrieval regime: full-length videos (avg >500s) as corpus documents, split into moment-level subtasks
- [x] Tested that the dataset runs with the mteb package (all 9 leaves + 4 aggregates register and evaluate)
- [x] `mteb/baseline-random-encoder`: ~0.01 map@5 across all leaves
- [x] Real model — Qwen/Qwen3-VL-Embedding-2B: see table above (final aggregate 0.626 map@5)
- [x] Corpus = 271 full videos, shared across directions
- [x] Paper-score reproduction: N/A — the paper evaluates moment localization; the full-video retrieval conversion is new


isaac-chung:
In the #4967 PR, e5v was used, but Qwen/Qwen3-VL-Embedding-2B results were reported here. Could you run the same model with this set here as well as a better comparison?

dukesun99:
> In the #4967 PR, e5v was used, but Qwen/Qwen3-VL-Embedding-2B results were reported here. Could you run the same model with this set here as well as a better comparison?

@isaac-chung Ran both models on both the chunk (#4967) and full-video sets. map@5:

| model | chunk ti2v | full ti2v | chunk tv2v | full tv2v |
|---|---|---|---|---|
| Qwen3-VL-Embedding-2B | 0.543 | 0.777 | 0.579 | 0.823 |
| E5-V (single-frame) | 0.167 | 0.426 | 0.174 | 0.435 |

E5-V is image+text only, so mteb won't score it over a video corpus; the E5-V row reduces each video to its middle frame as a single-image proxy, which is why it trails Qwen3-VL.

isaac-chung:
Thanks!

To match the paper, I wonder if it makes more sense to have this task be an aggregate task of several sub-tasks, where each sub-task is one of `['global-level', 'event-level', 'object-level']` ? The info needed to split this can be found in e.g. https://huggingface.co/datasets/avery00/MomentSeeker/blob/main/t2v.json: `"task": "Object Recognition"`, and these tasks correspond to a certain level in [the paper's Section 3.3](https://arxiv.org/pdf/2502.12558). This would create 9 subtasks. We could then make 3 aggregate tasks, one per moment. Then we could have a final aggregate task that averages across the 3 agg tasks. 

Would love a 2nd opinion on this: @Samoed or @KennethEnevoldsen 

Note:
  t2v.json: 1000 items
  ti2v.json: 400 items
  tv2v.json: 400 items

## Distribution counts

1)  Moment distribution within each type

| type | event-level | global-level | object-level | total |
|------|---|---|---|---|
| t2v | 699 | 74 | 227 | 1000 |
| ti2v | 88 | 156 | 156 | 400 |
| tv2v | 88 | 156 | 156 | 400 |

2) Type distribution within each moment

| moment | t2v | ti2v | tv2v | total |
|--------|---|---|---|---|
| event-level | 699 | 88 | 88 | 875 |
| global-level | 74 | 156 | 156 | 386 |
| object-level | 227 | 156 | 156 | 539 |

Script used: 
[analyze_momentseeker.py](https://github.com/user-attachments/files/30931424/analyze_momentseeker.py)

Samoed:
I think would be better to split it

isaac-chung:
@dukesun99 could you please split the task as described above? Let us know if you have any questions

dukesun99:
> @dukesun99 could you please split the task as described above? Let us know if you have any questions

Thanks @isaac-chung, @Samoed — done. Split it into the full 9-subtask structure as described.

**Levels** — using the `task` → moment mapping from your `analyze_momentseeker.py` (Causal Reasoning / Spatial Relation → global; Description Location / Action Recognition / Anomaly Detection → event; Object Recognition / Object Location / Attribute Recognition / OCR → object). The split reproduces your distribution counts exactly (t2v: global 74 / event 699 / object 227; ti2v & tv2v: 156 / 88 / 156 each).

**Tasks** (these replace the flat `MomentSeekerTI2VRetrieval` / `MomentSeekerTV2VRetrieval`):
- 9 leaf retrieval tasks — `MomentSeeker{TI2V,TV2V,T2V}{Global,Event,Object}LevelRetrieval`
- 3 per-level aggregates over the 3 directions — `MomentSeeker{Global,Event,Object}LevelRetrieval`
- 1 final aggregate — `MomentSeekerRetrieval`

**Corpus.** t2v draws on a larger set of source videos than the composed directions (many anomaly/UCF-Crime clips ti2v/tv2v never use), so I unified the corpus to **all 271 source videos** across the three directions — every direction now retrieves from the same corpus. This is a harder, more consistent setting; it lowers the ti2v/tv2v numbers vs the earlier 116-video corpus, which is expected (more distractors).

One note on the final aggregate: `MomentSeekerRetrieval` is a flat aggregate over the 9 leaves — numerically identical to averaging the 3 per-level aggregates (the levels are equal-sized triples across directions), and it avoids nesting aggregates-of-aggregates. Easy to switch to the literal nested version if you'd prefer.

Qwen3-VL-Embedding-2B, map@5 (random baseline in parens, all ~0.01):

| level | ti2v | tv2v | t2v |
|---|---|---|---|
| global | 0.734 | 0.769 | 0.507 |
| event | 0.699 | 0.728 | 0.439 |
| object | 0.726 | 0.745 | 0.290 |

Per-level aggregates: global 0.670 / event 0.622 / object 0.587. **Final `MomentSeekerRetrieval` 0.626.** The plain text→video direction (t2v) is clearly the hardest, especially object-level (0.290) — no image/clip to anchor the query.

Does this structure look right?

isaac-chung:
Thanks @dukesun99 . The structure looks good to me.

The point of the aggregate tasks is to easily match the paper scores, e.g. for e5v. Are you able to reproduce them?

dukesun99:
> Thanks @dukesun99 . The structure looks good to me.
> 
> The point of the aggregate tasks is to easily match the paper scores, e.g. for e5v. Are you able to reproduce them?

The paper's E5-V protocol has a task-definition gap worth settling before I run it.

The paper evaluates **within-video moment retrieval**: each long video (avg ~1200s) is cut into fixed 10s chunks, and for a query you rank the chunks *of that video*, with relevance by **IoU ≥ 0.3** against the answer moment (so several chunks count as relevant); image models like E5-V use each chunk's **middle frame**. Reported E5-V mAP@5 ≈ **19.5 / 20.7 / 19.8** (global/event/object).

This PR currently formulates MomentSeeker as **whole-video retrieval** — rank the full source videos (271 here), exactly one gold per query. That's a different, much easier task (video identification vs moment localization), so E5-V here won't come near the paper's ~0.20; it isn't measuring the same thing.

To reproduce the paper's E5-V numbers, the task would need to match its protocol: 10s-chunk corpus, **per-video candidate ranking** (mteb supports this via `top_ranked`), IoU-0.3 relevance, middle-frame for image models. That's essentially a moment-localization task — a rebuild from the current full-video version.

Would you like to see the performance metrics of E5-V under the current task formulation?

mteb/tasks/retrieval/eng/moment_seeker.py
```diff
@@ -76,43 +97,69 @@ class MomentSeekerTI2VRetrieval(AbsTaskRetrieval):
...
-        dialect=[],
-        sample_creation="found",
-        bibtex_citation=_BIBTEX,
-        prompt={
-            "query": "Given the question and the reference clip, retrieve the video segment that answers it."
-        },
-        is_beta=True,
-    )
+class MomentSeekerTI2VEventLevelRetrieval(AbsTaskRetrieval):
+    metadata = _meta("MomentSeekerTI2VEventLevelRetrieval", "ti2v", "event")
```
Samoed, line 113: Can you write metadata classes fully?

isaac-chung:
I see, thanks. I'd lean towards matching the paper as much as we can, so I'd like to get @Samoed 's agreement before we proceed. 

On the other hand, I'd be open to keeping the full video part, but only if we append a suffix to the task name. That way in the future, if we want to implement the chunked version, we can use the original name.

Samoed:
I think we create 2 versions with full videos and chunked. Would be good to have some reranking tasks

PR #5075, opened 2026-08-02
model: add cnmoro/static-nomic-384-pten-v2-st (static pt/en Model2Vec)

cnmoro:
Adds [`cnmoro/static-nomic-384-pten-v2`](https://huggingface.co/cnmoro/static-nomic-384-pten-v2), a Model2Vec/Tokenlearn **static** embedding model distilled from `nomic-ai/nomic-embed-text-v2-moe`, targeting **Portuguese and English**.

| | |
|---|---|
| Revision | `08981248ac665f19e71df149a2f2b5ef9b514bd7` |
| Parameters | 12,564,214 |
| Embedding dim | 384 |
| Max tokens | ∞ (static, no context limit) |
| License | apache-2.0 |

Results for `MTEB(por, v1)`, `MTEB(eng, v2)` and `MTEB(Multilingual, v2)` are submitted separately in embeddings-benchmark/results#659.

### Why a dedicated loader

This checkpoint uses **vocabulary quantization**: a 276,214-token tokenizer is mapped onto 32,000 shared embedding rows via a `mapping` tensor with per-token `weights`. `model2vec.StaticModel` reads that layout correctly, but `sentence_transformers.models.StaticEmbedding` does not — it indexes the embedding matrix directly by token id and reads out of bounds (`RuntimeError: Expected i < index_size to be true`). So this adds a small `Model2VecStaticModelWrapper` rather than reusing `SentenceTransformerEncoderWrapper` as the `minishlab/potion-*` entries do.

The wrapper is generic: any Model2Vec checkpoint using vocabulary quantization can reuse it.

### Training data

`training_datasets={"MSMARCO"}` — the model was fine-tuned on [`cnmoro/AllTripletsMsMarco-PTBR`](https://huggingface.co/datasets/cnmoro/AllTripletsMsMarco-PTBR), a pt-BR translation of MS MARCO triplets. Declared explicitly for the zero-shot determination.

### Validation

- `mteb.mock_run(...)` → **`all_passed: True`** (28 text task types ran; 32 image/audio/video tasks correctly reported as incompatible)
- `ruff check` and `ruff format --check` clean
- Full runs completed on 186 real task instances across the three benchmarks with **zero failures**
- `Assin2STS` reproduced at **61.3558**, matching the value published on the model card

### Checklist

### DIFF
--- a/mteb/models/model_implementations/cnmoro_models.py
+++ b/mteb/models/model_implementations/cnmoro_models.py
@@ -1,0 +1,86 @@
+"""Model definitions for cnmoro's static embedding models.
+
+Proposed addition to `mteb/models/model_implementations/cnmoro_models.py` in the
+embeddings-benchmark/mteb repository.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+
+from mteb.models.abs_encoder import AbsEncoder
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+
+if TYPE_CHECKING:
+    from mteb.types import Array
+
+MODEL2VEC_CITATION = """@software{minishlab2024model2vec,
+  author       = {Stephan Tulkens and {van Dongen}, Thomas},
+  title        = {Model2Vec: Fast State-of-the-Art Static Embeddings},
+  year         = {2024},
+  publisher    = {Zenodo},
+  doi          = {10.5281/zenodo.17270888},
+  url          = {https://github.com/MinishLab/model2vec},
+  license      = {MIT}
+}"""
+
+
+class Model2VecStaticModelWrapper(AbsEncoder):
+    """Loads a Model2Vec static model via `model2vec.StaticModel`.
+
+    Needed for checkpoints that use vocabulary quantization (a `mapping` tensor
+    projecting the tokenizer vocabulary onto a smaller set of shared embedding rows),
+    which `sentence_transformers.models.StaticEmbedding` cannot load.
+    """
+
+    def __init__(self, model: str, revision: str | None = None, **kwargs: Any) -> None:
+        from model2vec import StaticModel
+
+        self.model = StaticModel.from_pretrained(model)
+
+    def encode(
+        self,
+        inputs: Any,
+        *,
+        task_metadata: Any = None,
+        hf_split: str | None = None,
+        hf_subset: str | None = None,
+        prompt_type: Any = None,
+        **kwargs: Any,
+    ) -> Array:
+        texts = [text for batch in inputs for text in batch["text"]]
+        embeddings = self.model.encode(
+            texts, show_progress_bar=kwargs.get("show_progress_bar", False)
+        )
+        return np.asarray(embeddings, dtype=np.float32)
+
+
+static_nomic_384_pten_v2 = ModelMeta(
+    loader=Model2VecStaticModelWrapper,
+    name="cnmoro/static-nomic-384-pten-v2",
+    model_type=["dense"],
+    languages=["eng-Latn", "por-Latn"],
+    open_weights=True,
+    revision="08981248ac665f19e71df149a2f2b5ef9b514bd7",
+    release_date="2026-05-29",
+    # 32000 x 384 embedding matrix + 276214 per-token scalar weights
+    n_parameters=32000 * 384 + 276214,
+    n_embedding_parameters=32000 * 384 + 276214,
+    memory_usage_mb=50,
+    max_tokens=np.inf,
+    embed_dim=384,
+    license="apache-2.0",
+    similarity_fn_name=ScoringFunction.COSINE,
+    framework=["NumPy", "Sentence Transformers", "safetensors"],
+    reference="https://huggingface.co/cnmoro/static-nomic-384-pten-v2",
+    use_instructions=False,
+    adapted_from="nomic-ai/nomic-embed-text-v2-moe",
+    superseded_by=None,
+    # Distilled with Tokenlearn; finetuned on a pt-BR translation of MS MARCO triplets.
+    training_datasets={"MSMARCO"},
+    public_training_code=None,
+    public_training_data="https://huggingface.co/datasets/cnmoro/AllTripletsMsMarco-PTBR",
+    citation=MODEL2VEC_CITATION,
+)
--- a/mteb/models/model_implementations/cnmoro_models.py
+++ b/mteb/models/model_implementations/cnmoro_models.py
@@ -1,64 +1,15 @@
-"""Model definitions for cnmoro's static embedding models.
-
-Proposed addition to `mteb/models/model_implementations/cnmoro_models.py` in the
-embeddings-benchmark/mteb repository.
-"""
+"""Model definitions for cnmoro's static embedding models."""
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
-
 import numpy as np
 
-from mteb.models.abs_encoder import AbsEncoder
 from mteb.models.model_meta import ModelMeta, ScoringFunction
 
-if TYPE_CHECKING:
-    from mteb.types import Array
-
-MODEL2VEC_CITATION = """@software{minishlab2024model2vec,
-  author       = {Stephan Tulkens and {van Dongen}, Thomas},
-  title        = {Model2Vec: Fast State-of-the-Art Static Embeddings},
-  year         = {2024},
-  publisher    = {Zenodo},
-  doi          = {10.5281/zenodo.17270888},
-  url          = {https://github.com/MinishLab/model2vec},
-  license      = {MIT}
-}"""
-
-
-class Model2VecStaticModelWrapper(AbsEncoder):
-    """Loads a Model2Vec static model via `model2vec.StaticModel`.
-
-    Needed for checkpoints that use vocabulary quantization (a `mapping` tensor
-    projecting the tokenizer vocabulary onto a smaller set of shared embedding rows),
-    which `sentence_transformers.models.StaticEmbedding` cannot load.
-    """
-
-    def __init__(self, model: str, revision: str | None = None, **kwargs: Any) -> None:
-        from model2vec import StaticModel
-
-        self.model = StaticModel.from_pretrained(model)
-
-    def encode(
-        self,
-        inputs: Any,
-        *,
-        task_metadata: Any = None,
-        hf_split: str | None = None,
-        hf_subset: str | None = None,
-        prompt_type: Any = None,
-        **kwargs: Any,
-    ) -> Array:
-        texts = [text for batch in inputs for text in batch["text"]]
-        embeddings = self.model.encode(
-            texts, show_progress_bar=kwargs.get("show_progress_bar", False)
-        )
-        return np.asarray(embeddings, dtype=np.float32)
-
+from .model2vec_models import Model2VecModel
 
 static_nomic_384_pten_v2 = ModelMeta(
-    loader=Model2VecStaticModelWrapper,
+    loader=Model2VecModel,
     name="cnmoro/static-nomic-384-pten-v2",
     model_type=["dense"],
     languages=["eng-Latn", "por-Latn"],
@@ -73,7 +24,7 @@ static_nomic_384_pten_v2 = ModelMeta(
     embed_dim=384,
     license="apache-2.0",
     similarity_fn_name=ScoringFunction.COSINE,
-    framework=["NumPy", "Sentence Transformers", "safetensors"],
+    framework=["NumPy", "safetensors"],
     reference="https://huggingface.co/cnmoro/static-nomic-384-pten-v2",
     use_instructions=False,
     adapted_from="nomic-ai/nomic-embed-text-v2-moe",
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/cnmoro_models.py
```diff
@@ -0,0 +1,86 @@
...
+class Model2VecStaticModelWrapper(AbsEncoder):
+    """Loads a Model2Vec static model via `model2vec.StaticModel`.
+
+    Needed for checkpoints that use vocabulary quantization (a `mapping` tensor
+    projecting the tokenizer vocabulary onto a smaller set of shared embedding rows),
+    which `sentence_transformers.models.StaticEmbedding` cannot load.
+    """
+
+    def __init__(self, model: str, revision: str | None = None, **kwargs: Any) -> None:
+        from model2vec import StaticModel
```
KennethEnevoldsen: required but unspecified dependency

adding:
```
extra_requirements_groups=["model2vec"]
```

Should solve the trick

mteb/models/model_implementations/cnmoro_models.py
```diff
@@ -0,0 +1,86 @@
...
+    framework=["NumPy", "Sentence Transformers", "safetensors"],
+    reference="https://huggingface.co/cnmoro/static-nomic-384-pten-v2",
+    use_instructions=False,
+    adapted_from="nomic-ai/nomic-embed-text-v2-moe",
+    superseded_by=None,
+    # Distilled with Tokenlearn; finetuned on a pt-BR translation of MS MARCO triplets.
+    training_datasets={"MSMARCO"},
+    public_training_code=None,
+    public_training_data="https://huggingface.co/datasets/cnmoro/AllTripletsMsMarco-PTBR",
+    citation=MODEL2VEC_CITATION,
```
KennethEnevoldsen:
```suggestion
    citation=None,
```
Uses the technique of use of this should not cite this work

mteb/models/model_implementations/cnmoro_models.py
```diff
@@ -0,0 +1,86 @@
...
+
+import numpy as np
+
+from mteb.models.abs_encoder import AbsEncoder
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+
+if TYPE_CHECKING:
+    from mteb.types import Array
+
+MODEL2VEC_CITATION = """@software{minishlab2024model2vec,
```
KennethEnevoldsen: delete or replace with own

mteb/models/model_implementations/cnmoro_models.py
```diff
@@ -0,0 +1,86 @@
...
+  title        = {Model2Vec: Fast State-of-the-Art Static Embeddings},
+  year         = {2024},
+  publisher    = {Zenodo},
+  doi          = {10.5281/zenodo.17270888},
+  url          = {https://github.com/MinishLab/model2vec},
+  license      = {MIT}
+}"""
+
+
+class Model2VecStaticModelWrapper(AbsEncoder):
```
KennethEnevoldsen: there already exists a `Model2VecModel` model in:

https://github.com/cnmoro/mteb/blob/a9a10e0afca62b446dc3c0671a3ea6a1cd562344/mteb/models/model_implementations/model2vec_models.py#L139

mteb/models/model_implementations/cnmoro_models.py
```diff
@@ -0,0 +1,86 @@
+"""Model definitions for cnmoro's static embedding models.
+
+Proposed addition to `mteb/models/model_implementations/cnmoro_models.py` in the
+embeddings-benchmark/mteb repository.
+"""
```
KennethEnevoldsen:
```suggestion
```

cnmoro:
Thanks for the review — all five points addressed in the latest push, which drops 54 lines net.

- **Reuse `Model2VecModel`** — done, the duplicate wrapper is gone. I'd missed the existing one; sorry for the noise.
- **`extra_requirements_groups=["model2vec"]`** — added.
- **`citation=None`** — done, and the copied `MODEL2VEC_CITATION` block is deleted.
- **Placeholder docstring** — removed.
- I also dropped `"Sentence Transformers"` from `framework`, since this checkpoint cannot actually be loaded that way (see below).

Verified after the change: `ruff check` and `ruff format --check` clean, `mteb.mock_run(...)` → `all_passed: True`, and `mteb.get_model(...)` now returns `Model2VecModel`.

I re-ran `Assin2STS` through `Model2VecModel` to confirm the submitted results match the reference implementation: **0.6135576687** vs **0.613558** stored in embeddings-benchmark/results#659 — identical to the precision the results files keep.

### One thing worth flagging

`Model2VecModel` currently emits:

```
DeprecationWarning: Model2VecModel is deprecated. Please use SentenceTransformerEncoderWrapper instead.
```

For this checkpoint that migration path doesn't work. It uses **vocabulary quantization** — a 276,214-token tokenizer mapped onto 32,000 shared embedding rows via a `mapping` tensor plus per-token `weights`. `sentence_transformers.models.StaticEmbedding` indexes the embedding matrix directly by token id, so it reads out of bounds:

```
RuntimeError: Expected i < index_size to be true, but got false.
```

So `Model2VecModel` is the only working loader here, and following its own deprecation notice would break the model. Not something this PR needs to solve — just flagging that retiring it would strand vocabulary-quantized Model2Vec checkpoints, unless `StaticEmbedding` grows support for the `mapping`/`weights` layout first. Happy to open a separate issue if that's useful.

mteb/models/model_implementations/cnmoro_models.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import annotations
+
+import numpy as np
+
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+
+from .model2vec_models import Model2VecModel
+
+static_nomic_384_pten_v2 = ModelMeta(
+    loader=Model2VecModel,
```
Samoed: Seems your model support sentence transformers. Why not to use sentence transformer loader directly?

cnmoro: Good question — the model card does advertise a `SentenceTransformer` path, but that claim is wrong and I'll fix the card. The checkpoint cannot actually be loaded that way. I re-tested both routes just now on the exact submitted revision (`08981248ac66…`):

```python
m = SentenceTransformer("cnmoro/static-nomic-384-pten-v2")   # loads fine
m.encode(["teste"])
# RuntimeError: Expected i < index_size to be true, but got false.
#   sentence_transformers/.../static_embedding.py:122 in forward -> F.embedding_bag(...)
```

and the same via mteb's own wrapper (CPU, to avoid it surfacing as a CUDA device-side assert):

```python
SentenceTransformerEncoderWrapper("cnmoro/static-nomic-384-pten-v2", revision="08981248ac66…", device="cpu")
# constructs OK, then on evaluate:
# RuntimeError: Expected i < index_size to be true, but got false.
```

**Why:** this checkpoint uses Model2Vec's *vocabulary quantization*. `model.safetensors` holds

| tensor | shape |
|---|---|
| `embeddings` | `(32000, 384)` |
| `mapping` | `(276214,)` int64 |
| `weights` | `(276214,)` float32 |

The tokenizer has 276,214 tokens, which are projected onto 32,000 shared embedding rows via `mapping`, scaled by `weights`. `modules.json` declares `StaticEmbedding`, so ST takes that path and indexes the 32,000-row matrix directly by token id — out of bounds as soon as a token id ≥ 32000 appears (the first test sentence hits id 250217). `StaticEmbedding.quantize_to` is dtype quantization (float32/int8) and unrelated to this. I checked 5.6.0 and there's nothing handling `mapping`/`weights`.

So `Model2VecModel` is currently the only loader that reads this layout correctly.

**To actually use the ST loader** the weights would have to be re-exported with the mapping materialized — a full `(276214, 384)` matrix, **424 MB vs the 52 MB published, 8.1×** — which gives up the compression that is the point of the model. It would also mean a new revision on the Hub, and the 160 results in embeddings-benchmark/results#659 are pinned to `08981248ac66…`, so they'd all need re-running.

Happy to go either way — your call:

1. **Keep `Model2VecModel`** (current state of the PR), or
2. I re-export an ST-compatible checkpoint, re-run all 186 tasks against the new revision, and switch to `SentenceTransformerEncoderWrapper`.

Worth noting for (1): `Model2VecModel` currently emits `DeprecationWarning: ... Please use SentenceTransformerEncoderWrapper instead`, which for vocabulary-quantized checkpoints points at a loader that cannot load them. Might be worth either narrowing that warning or teaching `StaticEmbedding` the `mapping`/`weights` layout — happy to open a separate issue if useful.

KennethEnevoldsen: I would go for 2), but would be nice to have a fix for that in sentence transfomers. Is there an issue on this? (cc @tomaarsen in case you are not aware)

tomaarsen: Oh, that's indeed not great. Model2vec models should all load in Sentence Transformers, although there've been a handful of outliers where weights were stored in int8. Based on what I'm reading here, it seems something similar is happening here. I believe @davidmezzetti used that int8 approach.

If this is becoming a common way to store model2vec-like models, then perhaps Sentence Transformers should indeed be updated to support this as well, but it's not something I can work on before the upcoming releases I have scheduled, as time is a little tight right now.

cnmoro: Sounds good, going with 2).

I built the materialized export locally and confirmed it round-trips exactly. Reconstructing `row_i = embeddings[mapping[i]] * weights[i]` gives a plain `(276214, 384)` matrix that `StaticEmbedding` loads without any changes:

```
quantized   : mapping (276214,), embeddings (32000, 384), weights (276214,)   52 MB
materialized: embeddings (276214, 384)                                       424 MB

SentenceTransformer(export).encode(...) vs model2vec StaticModel.encode(...)
  cosine   : min 1.0000000000, mean 1.0000000000
  max diff : 1.49e-08
```

So it is purely a storage-layout change, not a numerical one. I'll upload that as a new revision, re-run all 186 tasks against it, and update this PR to `SentenceTransformerEncoderWrapper` plus embeddings-benchmark/results#659 with the new revision. Will ping when it's pushed.

@tomaarsen for context on whether this is worth supporting natively: it isn't int8 quantization here, it's Model2Vec/Tokenlearn's `vocabulary_quantization` option (`"vocabulary_quantization": 32000` in `config.json`). Instead of one row per token it stores a smaller shared table plus two parallel `(vocab,)` tensors, `mapping` (int64 row index) and `weights` (float32 scale), with the embedding for token `i` being `embeddings[mapping[i]] * weights[i]`. `StaticEmbedding` currently passes raw token ids straight into `embedding_bag` against the smaller table, so anything with id >= 32000 reads out of bounds:

```
RuntimeError: Expected i < index_size to be true, but got false.
```

Supporting it would mean materializing at load time (cheap, one gather + broadcast multiply, exactly the snippet above) or carrying the indirection into the forward pass. Completely understand it's not a priority right now, and no blocker for this PR since the re-export sidesteps it. Happy to open an issue on the Sentence Transformers side so it's tracked.

cnmoro: Done, and a correction to something I said above.

I claimed the export was "purely a storage-layout change, not a numerical one". That is only true up to 512 tokens. `model2vec.StaticModel.encode` defaults to `max_length=512`, while Sentence Transformers' `StaticEmbedding` has `max_seq_length = inf` and does not truncate. My verification used short sentences, so I missed it.

Cosine between the two loaders on varied text:

| tokens | cosine |
|---|---|
| 100 | 1.000000 |
| 400 | 1.000000 |
| 512 | 0.998651 |
| 800 | 0.996333 |
| 2000 | 0.992483 |

The weights themselves round-trip exactly; the divergence is purely truncation. Practically this means the earlier submission was silently evaluated with a 512-token cap while its `ModelMeta` declared `max_tokens=inf`, which was inconsistent. The re-run has no cap, so the declaration and the measurement now agree.

Effect on scores, per-subset against the previous run:

- `MTEB(eng, v2)`: 24 of 41 tasks bit-identical; the rest shift by under 1 point except `STS22.v2` (4.3) and `ArXivHierarchicalClusteringP2P` (1.1)
- long-document tasks move most, e.g. `MultiLongDocRetrieval` pt went 25.23 to 43.98
- benchmark totals (mean of task-type means): por 51.81 to 53.51, eng 47.66 to 47.60, multilingual 38.35 to 38.69

Pushed here: this PR now registers `cnmoro/static-nomic-384-pten-v2-st` (rev `50fed703`) with `SentenceTransformerEncoderWrapper`, no custom loader. `ruff check`, `ruff format --check` and `mteb.mock_run` (`all_passed: True`) all pass.

embeddings-benchmark/results#659 is updated with all 186 tasks re-run against that revision, 160 result files, zero failures.

The original `cnmoro/static-nomic-384-pten-v2` stays on the Hub for anyone wanting the 52MB `model2vec` version; the `-st` repo is the materialized 424MB one and its card links back.

@tomaarsen the `vocabulary_quantization` detail in my earlier comment still stands if you ever want to support the layout natively, but nothing here is blocked on it now.

KennethEnevoldsen (approved):

mteb/models/model_implementations/cnmoro_models.py
tomaarsen: I see! Yeah, I think ST should see if it can adopt that format natively. I'm open to a PR although I won't be able to review it for another week or so.

PR #5078, opened 2026-08-03
Add model: OpenGVLab/InternVideo2-CLIP-1B-224p-f8

### LINKED ISSUE #5060: Add model: OpenGVLab/InternVideo2-CLIP-1B-224p-f8

https://huggingface.co/OpenGVLab/InternVideo2-CLIP-1B-224p-f8

### Arxiv link

https://arxiv.org/abs/2403.15377

### Description of the model

InternVideo2 (OpenGVLab) is a video foundation model family trained with a
progressive scheme combining masked video modeling, cross-modal contrastive
learning, and next-token prediction.

This issue covers the CLIP variant specifically — InternVideo2-CLIP-1B-224p-f8 —
which has aligned video and text towers from the contrastive stage and exposes a
documented video/text similarity path. The Chat/MLLM variants are generative and
aren't suited to embedding evaluation, so they're out of scope here.

Fits the video/omni model priority for MOEB, and because it has a text tower it
supports t2v and v2t retrieval rather than video-to-video only. The paper reports
video retrieval benchmarks, so reproducibility can be checked directly.

One caveat worth recording up front: reported results can fall short of the
paper's numbers unless the ITM re-ranking stage is replicated, and frame count
(4 vs 8) also shifts scores.

hubielu:
Closes #5060. InternVideo2-CLIP-1B (1B video tower plus InternVL-C text tower) as a video/text dual encoder.

Drafted with Claude, and I ran and verified everything below.

Checkpoints: the HF repo holds only the CLIP-stage deltas, which is 47 tensors and about 7M params. The two towers come from `InternVideo2-Stage2_1B-224p-f4` (2.8 GB) and `internvl_c_13b_224px.pth` (25.4 GB), plus a config and tokenizer that only exist in InternVL's git repo. That comes to 7.7B params total across three gated repos.

Results: MSVD zero-shot at 8 frames in bf16 gives t2v ndcg@10 of 0.859 and v2t of 0.858, and `mteb mock-run` passes 37/37 (file included). These are not reproductions, because there is no published MSVD number for the CLIP variant, and mteb's split uses one caption per clip rather than the roughly 27k-caption protocol from the paper.

One correctness note that may be useful to anyone else wrapping this model family: upstream's config sets `use_lora=True`, which under peft 0.7 or newer renames `q_proj.weight` to `q_proj.base_layer.weight`. The InternVL checkpoint still uses the old naming and `load_checkpoint` runs with `strict=False`, so the q_proj and v_proj weights across all 32 layers silently stay randomly initialised. The model still loads and runs, but the outputs are garbage. Setting `use_lora=False` is numerically identical, because the released add-on contains no LoRA tensors at all, and it makes every key match.

Status: the install currently requires a local clone, which isn't acceptable per the discussion below. Waiting on the gated-repo question before building the ST module.

- [X] I reproduced results from the original paper (if applicable) on at least one benchmark, and I am including the results in the PR description.


mteb/models/model_implementations/internvideo2_models.py
```diff
@@ -0,0 +1,413 @@
...
+
+_INSTALL_HINT = """InternVideo2-CLIP requires the upstream OpenGVLab package.
+
+Install it from a local clone rather than a git URL: upstream's .gitmodules is missing
+an entry for InternVideo3/InternVideo3_eval/lmms-eval, so pip's automatic submodule
+init fails.
+
+    git clone --depth 1 https://github.com/OpenGVLab/InternVideo.git
+    pip install --no-deps -e InternVideo/InternVideo2/multi_modality
+    pip install "mteb[internvideo2]"
```
Samoed: I don't think we want to integate model that reqire to clone repo

hubielu: Makes sense. The clone was only needed because pip install git+... fails during automatic submodule init, upstream's .gitmodules is missing an entry for InternVideo3/InternVideo3_eval/lmms-eval.

Pinning to an older commit avoids that. b81a0bb23c48 is the commit that added the installable pyproject ("Allow InternVideo2/multi_modality to be pip installed") and predates InternVideo3, so only the two properly-declared submodules exist:

internvideo2_multi_modality @ git+https://github.com/OpenGVLab/InternVideo.git@b81a0bb23c48#subdirectory=InternVideo2/multi_modality

Same pattern as the ebind entry. I've confirmed that commit has the pyproject and the CLIP model with no broken gitlink, but I haven't yet verified it resolves under uv (upstream pins apex/deepspeed/flash-attn, which inference doesn't need) or re-run the load test against it, since it's Oct 2024 rather than the revision I tested.

If you'd rather avoid a git dependency entirely, the alternative is vendoring the ~800 lines of vision/text tower into the wrapper file. Both are fine by me, I'd rather confirm the direction with you before rebuilding the GPU environment to validate one of them.

Samoed: We can't include git dependencies into our package because otherwise we can't publish it. Also I'm not sure if we want to include a lot of files for models. WDYTH @KennethEnevoldsen

hubielu: > We can't include git dependencies into our package because otherwise we can't publish it. Also I'm not sure if we want to include a lot of files for models. WDYTH @KennethEnevoldsen

Noted, PyPI rejects direct URL deps so that's out.

One thing that might help: `ebind` and `omnivinci` both ship model files with their extras commented out in pyproject, seemingly for this same reason — is that the established pattern? Otherwise vendoring is ~1,300 lines but would actually remove three workarounds (flash-attn stubs, the modeling_utils shim, and the `transformers<5` pin), so I'm happy either way.

Will wait on you and @KennethEnevoldsen

KennethEnevoldsen:
@hubielu a 6000 character long PR message is too much. Please respect reviewers' time. Additionally, do note our [AI policy](https://github.com/embeddings-benchmark/mteb/discussions/4509).

mteb/models/model_implementations/internvideo2_models.py
KennethEnevoldsen: My take:

The model **has** to be in a version repository, preferably huggingface if not, there needs to be a good reason for it. Ideally, it should require minimal or no dependencies and integrate with existing libraries (e.g. sentence transformers). If it does not it provide a reasonable justification on why. Using the model should not require external script, download from github etc. during runtime and should be pip installable with `mteb[...]` without requiring initial setup. Thus adding a model can require adding github package to pypi or uploading a model to huggingface with sentence transformers support.

We have previously been more lenient on this than this, so @Samoed let me know if you agree with this.

KennethEnevoldsen: In general though it seems like this model could be implemented simply using sentence transformers, though it might require a custom module and `trust_remote_code=True`.

You can see examples here:

https://huggingface.co/collections/sentence-transformers/sentence-transformers-v54-integrations

Samoed: >  If it does not it provide a reasonable justification on why. Using the model should not require external script, download from github etc. during runtime and should be pip installable with mteb[...] without requiring initial setup. 

Yes, I agree on this

> In general though it seems like this model could be implemented simply using sentence transformers, though it might require a custom module and trust_remote_code=True.

Of course this can be done, but we would still to publish somewhere, because I don't think authors will answer to pr, but this would worth a shot

KennethEnevoldsen: of course also worth noting that a PR is enough if the author already has it up (we don't need it merged), e.g.: https://huggingface.co/LCO-Embedding/LCO-Embedding-Omni-3B/discussions/1

Samoed: Yes, but we can't use revision properly then. I think it will be then `discussions/1`

hubielu:
> @hubielu a 6000 character long PR message is too much. Please respect reviewers' time. Additionally, do note our [AI policy](https://github.com/embeddings-benchmark/mteb/discussions/4509).

Understood, sorry about that. I cut the description down.

mteb/models/model_implementations/internvideo2_models.py
KennethEnevoldsen: "refs/1" I think, but yes. It is still a unique identifier though (I was also not able to figure out a way to delete it so I think it is stable)

KennethEnevoldsen:
> Understood, sorry about that. I cut the description down.

Thanks for the cut - @hubielu does it make sense to upload it as a sentence transformers-compatible model?

hubielu:
> Thanks for the cut - @hubielu does it make sense to upload it as a sentence transformers-compatible model?

Yes, that would fix most of it. Concretely, a PR on OpenGVLab's repo adding the ST module plus the LlamaConfig and tokenizer, which currently only exist in InternVL's git repo, so nothing gets fetched from GitHub at runtime.

One thing I'd want your read on first. The module would still hf_hub_download the two base towers (2.8 GB and 25.4 GB), and all three repos are gated, so a user has to accept terms on each before it loads. Does that count as "requiring initial setup"? If it does, the alternative is consolidating everything into one repo under my namespace, which would be an unofficial re-upload of about 16 GB.

hubielu:
> > Thanks for the cut - @hubielu does it make sense to upload it as a sentence transformers-compatible model?

@KennethEnevoldsen bumping this one. Happy to build the ST module, just want to confirm the gated-repo point first.

hubielu:
@KennethEnevoldsen going ahead with the ST module. Assuming gated-but-official
is fine, since mteb already has gated models in the registry (NV-Embed-v2 etc.)
and the alternative — re-uploading 16 GB under my namespace — would violate the
terms I accepted for access.

Will open the PR on OpenGVLab's repo and point ModelMeta at refs/pr/1.

hubielu:
@KennethEnevoldsen thanks — starting on the ST module.

On the gated question, I'll take it that gated-but-official is acceptable, since
mteb already ships NV-Embed-v2, Llama-2-7b-hf and embeddinggemma-300m. The 16 GB
re-upload would mean redistributing weights I only hold under those terms, so I'd
rather avoid it.

I've vendored the inference path to ~1.7k lines — no upstream package, no runtime
clone. Next step is a PR on OpenGVLab's repo, with ModelMeta pinned to refs/pr/3.
Will report back once it loads.

hubielu:
@KennethEnevoldsen quick update — the ST module is now working end-to-end from OpenGVLab’s gated repo PR #3.

Verified text/video parity with the previous implementation, and MTEB text, video, and mixed inputs all pass. The MTEB branch is pinned to the HF PR commit, and CI is green.

Upstream PR: https://huggingface.co/OpenGVLab/InternVideo2-CLIP-1B-224p-f8/discussions/3

mteb_mock_run_results.md
Samoed, line 1: CAn you remove this?

Samoed (commented):
Can you reproduce scores of model?

mteb/models/model_implementations/internvideo2_models.py
```diff
@@ -0,0 +1,38 @@
...
+  title={InternVideo2: Scaling Video Foundation Models for Multimodal Video Understanding},
+  author={Wang, Yi and Li, Kunchang and Li, Xinhao and Yu, Jiashuo and He, Yinan and Chen, Guo and Pei, Baoqi and Zheng, Rongkun and Xu, Jilan and Wang, Zun and others},
+  journal={arXiv preprint arXiv:2403.15377},
+  year={2024}
+}"""
+
+internvideo2_clip_1b_224p_f8 = ModelMeta(
+    loader=SentenceTransformerEncoderWrapper,
+    loader_kwargs=dict(num_frames=8, trust_remote_code=True),
+    name="OpenGVLab/InternVideo2-CLIP-1B-224p-f8",
```
Samoed:
```suggestion
    name="OpenGVLab/InternVideo2-CLIP-1B-224p-f8",  # discussions/3
```

hubielu:
> Can you reproduce scores of model?

@Samoed Reproduced DiDeMo. R@1 47.25 t2v / 45.45 v2t against MODEL_ZOO's 47.8 / 46.4, so within 0.6 and 1.0 points. Full: t2v R@1 47.25 / R@5 72.27 / R@10 80.18 / ndcg@10 63.72, v2t R@1 45.45 / R@5 72.77 / R@10 82.48 / ndcg@10 63.32.

Run on a clean A6000 box with just `pip install "mteb[video,internvideo2]"` and an HF login: no clone, no manual setup. Also removed the mock run results file and applied the discussions/3 suggestion.

Samoed (approved):

PR #5081, opened 2026-08-03
fix: add av to the video extra

hubielu:
mteb[video] installs torchcodec but not av, so video decoding fails and mteb mock-run errors on all seven mock video tasks. av>=17.0.0 is already declared in the test extra, which is why CI passes — the dependency is real, just declared in the wrong group. Found while adding InternVideo2 (#5078).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,7 +91,7 @@ api = [
 og = [
     "playwright>=1.49.0",
 ]
-video = ["torchcodec"]
+video = ["torchcodec", "av>=17.0.0"]
 github = ["PyGithub>=1.60"]
 
 # model specific optional dependencies:


Samoed (changes requested):
I don't think we want to add `av` to video. `av` needed for testing only

hubielu:
> I don't think we want to add `av` to video. `av` needed for testing only

You're right, my mistake. av is only imported in mteb/mocks/mock_tasks/create_mock_samples.py, which writes the synthetic mp4s — nothing in the real video path uses it, that's all torchcodec. So the test group is the correct home and this PR is wrong.

The thing I actually hit: adding_a_model.md asks contributors to run mteb mock-run, and for a video model that needs av, which mteb[video] doesn't provide. Would a docs note saying to install the test group be useful, or is that already obvious enough? Happy to send that instead. Closing this either way.

Issue #5021, opened 2026-07-25
MOEB Dataset Audit and Refresh


lxer19:
Related to #4842.

## Goal

Audit existing MOEB datasets for benchmark saturation and data-quality
problems, then prioritize confirmed issues for cleaned or harder dataset
versions.

The audit will consider:

- saturation, including whether top models are approaching the metric ceiling;
- exact and near duplicates;
- noisy, incomplete, or inconsistent labels;
- train/test or cross-split overlap;
- trivial examples and modality-specific shortcuts;
- missing, broken, or duplicated media items.

## Initial Scope

I will first review the datasets included in the current MOEB benchmark
families and propose a small, representative set of pilot datasets.

Before running the full audit, I will confirm the initial dataset scope with
the maintainers.

The first pilot should be:

- explicitly included in the MOEB scope;
- reasonably accessible and reproducible;
- supported by enough public model results for saturation analysis;
- representative of at least one current modality or inter-modality task.

### Dataset defects

- inspect dataset structure, splits, queries, candidates, and relevance labels;
- detect exact duplicates;
- detect near duplicates using modality-appropriate similarity methods;
- check split overlap;
- identify potentially incomplete or inconsistent labels;
- detect missing or corrupted media;
- flag potentially trivial examples or shortcuts;
- manually validate representative findings.

### Saturation

- collect available model results for each audited task;
- examine top-model scores and distance from the metric ceiling;
- measure score separation among leading models;
- inspect model-ranking stability;
- distinguish direct task results from proxy or incomplete result coverage.

## Deliverables

- reproducible dataset-audit scripts;
- an inventory of audited datasets and result coverage;
- a report of flagged datasets;
- manually validated examples;
- proposed remediation for confirmed issues;
- cleaned datasets or harder versions for selected cases;
- focused pull requests for implementation-ready fixes.

## Current Progress

I developed an initial retrieval-audit prototype on
`NanoArguAnaRetrieval`.

The prototype detected:

- exact duplicate corpus entries;
- potential incomplete-qrels cases where identical document content appears
  under multiple IDs but only one ID is labeled relevant.

`NanoArguAnaRetrieval` https://github.com/embeddings-benchmark/mteb/pull/4959 belongs to NanoBEIR rather than MOEB, so I am keeping
that finding separate from this project. I will reuse only the general audit
methodology and restart the MOEB work with an in-scope pilot dataset.

## Next Steps

1. Build an inventory of current in-scope datasets and available results.
[...]

PR #5090, opened 2026-08-05
Update quality check

Samoed:
New checks:
- Text/image/audio/video duplicate detection, extended to all modalities and task types (previously only single-field text tasks)
- Correct duplicate-check denominator for Retrieval (`num_documents`/`num_queries` instead of `num_samples`)
- Pair-level duplicate check via `unique_pairs` (STS/Bitext/PairClassification)
- Small image / video frame size
- Zero-length video clip
- Fewer than 2 unique labels
- Zero score variance (`min_score == max_score`)
- Zero relevant docs per query
- Zero top-ranked candidates per query
- Relevant docs exceeding corpus size
- Audio/video length mismatch between paired fields
- Impossible unique count (`unique_count > expected_count`)
- `min_text_length` now checked unconditionally (previously skipped for retrieval corpora)
Warning:
- Long text/audio/video/frame-size outliers

Ref https://github.com/embeddings-benchmark/mteb/issues/5021

### DIFF
--- a/tests/test_tasks/test_task_quality.py
+++ b/tests/test_tasks/test_task_quality.py
@@ -918,13 +918,6 @@ KNOWN_ISSUES: dict[str, list[str]] = {
 
 
 def _task_known_issues() -> dict[str, set[str]]:
-    """Reverse index (task name -> set of exempted check *kinds*). A task can
-    legitimately need more than one exemption (e.g. both "short_text" and
-    "duplicate_text"), so this doesn't collapse to a single kind per task --
-    only a check whose kind is recorded here for that task is suppressed
-    (kind-level, not per-field/per-hf_subset); any other check kind failing
-    on the same task still surfaces as a real error.
-    """
     task_known_issues: dict[str, set[str]] = {}
     for kind, names in KNOWN_ISSUES.items():
         for name in names:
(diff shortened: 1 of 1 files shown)


tests/test_tasks/test_task_quality.py
```diff
@@ -3,364 +3,1443 @@
...
+    LabelStatistics,
+    RelevantDocsStatistics,
+    ScoreStatistics,
+    SplitDescriptiveStatistics,
+    TextStatistics,
+    TopRankedStatistics,
+    VideoStatistics,
+)
+
+# DO NOT ADD TO THIS DICT WITHOUT VERIFYING THE TASK ACTUALLY FAILS THAT CHECK.
```
KennethEnevoldsen: We can simply have it fail if it is in but doesn't fail check (that way we can remove issues)

KennethEnevoldsen: I see it is already there
```suggestion
```

KennethEnevoldsen (approved):
Generally think it is good, but can you explain what you mean by: "Recursion into multilingual" couldn't find it in the tests

Samoed:
> "Recursion into multilingual" couldn't find it in the tests

This is bad wording. This means that each subset will be checked and it's problem will be reproted

PR #5091, opened 2026-08-05
Add MLX as a framework annotation

choipilkyu:
Adds `MLX` to the `FRAMEWORKS` literal, alongside the other alternative runtimes already there (`GGUF`, `ONNX`, `OpenVINO`), and maps the HuggingFace `mlx` tag in `_get_frameworks_from_hf_tags`.

The Hub derives `mlx` from repository contents the same way it derives `gguf` and `openvino`, so without the mapping `ModelMeta.from_hub()` silently drops it:

```python
ModelMeta._get_frameworks_from_hf_tags("mlx-community/kanana-2-3b-instruct-4bit")
# before: ['safetensors']
# after:  ['safetensors', 'MLX']
```

Three lines, no new dependency, no new models.

<details>
<summary>This PR originally registered three MLX conversions of Nemotron-3-Embed-1B as models; that part has been dropped.</summary>

The original version added `ModelMeta` entries for the bf16/8-bit/4-bit MLX builds together with a custom encoder and an `mlx` optional dependency. Per the discussion below, one reference implementation per model is the invariant worth protecting, and runtime variants of a single checkpoint should not occupy separate leaderboard rows. Those parts are removed; only the annotation remains.

</details>


ayush1298:
@Samoed @KennethEnevoldsen I don't think we should add MLX versions of a model as separate models. What's your thought on this?

Samoed:
I think we can add it as a separete model

mteb/models/model_implementations/nemotron_mlx_models.py
```diff
@@ -0,0 +1,256 @@
...
+            ) * _MASK_BIAS
+            for layer in self.layers:
+                hidden = layer(hidden, attn_scale, mask=padding_mask)
+            hidden = self.norm(hidden).astype(mx.float32)
+
+            mask = attention_mask[:, :, None].astype(mx.float32)
+            pooled = (hidden * mask).sum(axis=1) / mask.sum(axis=1)
+            return pooled / mx.linalg.norm(pooled, axis=-1, keepdims=True)
+
+    return NemotronEmbedModel(ModelArgs.from_dict(config))
```
Samoed: I'm not familiar with mlx. Is it possible to add code to use it with `trust_remote_code=True`

KennethEnevoldsen:
Hmm I am not sure we want multiple implementations of the same models (if they disagree on a task we need a reference. We only want one reference). We can def. add MLX as a framework annotation. 

While MLX, OpenVINO, and vLLM maintainers can use MTEB to ensure their implementations match the reference (and users can as well), I am not sure these are the models we want to maintain. We already have plenty degrees of freedom (OS, flash-attention, torch version etc.) I am not sure adding more is the most productive direction.

I am also not sure it makes for a very pleasant experience as a user of the leaderboard (if the top 5 is simply versions of one model)

Let me know what you think

Edit: one option might be to add it as an experiment

Samoed:
> Edit: one option might be to add it as an experiment

I think about experiments more like parameter configurations. The current setup will be difficult to implement as an experiment, because these are different models.

choipilkyu:
Thanks all — that's a fair call, and the leaderboard point in particular convinces me: three runtime variants of one checkpoint competing for rank is not what the board is for, and "one reference per model" is the right invariant to protect.

I've dropped the three `ModelMeta` entries, the custom loader, and the `mlx` optional dependency. What's left is only the annotation @KennethEnevoldsen said was fine: `MLX` in the `FRAMEWORKS` literal, and the `mlx` HF tag in `_get_frameworks_from_hf_tags` so `ModelMeta.from_hub()` stops silently dropping it. Three lines, no new dependency, no new models.

On using MTEB to check a runtime against the reference — that's exactly what I'll use it for on my side; those agreement numbers belong on the model card, not the leaderboard.

Happy to close this entirely if you'd rather not carry even the annotation.

ayush1298:
Can you also add entry to frameworks in `modelmeta` for MLX

choipilkyu:
@ayush1298 That is what the PR is now — after dropping the model entries it is only that annotation, in two places in `model_meta.py`:

```diff
 FRAMEWORKS = Literal[
     ...
     "OpenVINO",
     "Transformers",
+    "MLX",
 ]
```

```diff
         tag_to_framework: dict[str, FRAMEWORKS] = {
             ...
             "gguf": "GGUF",
             "openvino": "OpenVINO",
+            "mlx": "MLX",
         }
```

The second one is what makes it useful today: the Hub derives an `mlx` tag from repository contents the same way it derives `gguf` and `openvino`, so without the mapping `ModelMeta.from_hub()` drops it silently:

```python
ModelMeta._get_frameworks_from_hf_tags("mlx-community/kanana-2-3b-instruct-4bit")
# before: ['safetensors']
# after:  ['safetensors', 'MLX']
```

Whole diff is 5 added lines, no new dependency and no new models. If you were looking at the original version of this PR, it registered three MLX conversions as models as well — that part is gone. Happy to rebase if it has drifted, and equally happy to close it if the annotation on its own isn't wanted.

ayush1298:
@choipilkyu what I meant was, can you add MLX as a framework in `Nemotron-3-Embed-1B` model's modelmeta, specifically here:  https://github.com/embeddings-benchmark/mteb/blob/98f3b8f24e0803d6ee63c6ddc99d5ba882ab7f36/mteb/models/model_implementations/nvidia_models.py#L721

KennethEnevoldsen (commented):
This updated PR looks good, but if we add it, we should also go through the models and annotate those that support MLX (otherwise this doesn't add much)

PR #5094, opened 2026-08-05
model: add VideoMAE video encoders

hubielu:
Three VideoMAE video-only encoders: videomae-base, videomae-large, videomae-base-finetuned-kinetics. Proposed in #4842. TimeSformer split into #5100 per review.

Drafted with Claude; I ran and verified everything below.

VideoMAE v1 rather than v2, since VideoMAEv2 is not implemented in transformers and I took that as the bar from Samoed's comment on VJepa2.1.

**Results.** HMDB51Clustering, one of the datasets in VideoMAE's transfer table, v_measure: videomae-base (self-supervised) 0.4079, videomae-base-finetuned-kinetics 0.6996. Not a reproduction, since the paper fine-tunes the whole model on HMDB51 while mteb fits on frozen embeddings. Worth flagging for the leaderboard that the self-supervised checkpoint scores well below the finetuned one.

**Weight loading.** transformers v4 stored VideoMAE's query and value biases as separate q_bias and v_bias parameters, which is how every published checkpoint is saved. v5 refactored to standard Linear layers and ships no mapping for the old names, so those biases load as zeros with no error raised. The wrapper copies them back from the checkpoint and skips itself on v4 or on any version that loads them correctly.

**Other details.** VideoMAE has no CLS token and its head averages over tokens, so the wrapper mean-pools and returns the encoder output. num_frames comes from config.num_frames since the temporal position embeddings are not interpolated.

The K400 tasks are declared in training_datasets: pre-training used Kinetics-400 videos without labels, and the finetuned checkpoint is supervised on them.

mteb mock-run passes 4/4 video tasks on all three checkpoints.


mteb/models/model_implementations/videomae_timesformer.py
```diff
@@ -0,0 +1,507 @@
...
+        def tensors() -> dict[str, torch.Tensor]:
+            nonlocal state_dict
+            if state_dict is None:
+                state_dict = _load_checkpoint_tensors(self.model_name, self.revision)
+            return state_dict
+
+        self._restore_qkv_bias(tensors)
+        self._load_fc_norm(tensors)
+
+    def _restore_qkv_bias(self, tensors: Any) -> None:
```
Samoed: Why this is needed? I didn't find this in their examples

mteb/models/model_implementations/videomae_timesformer.py
```diff
@@ -0,0 +1,507 @@
...
+    VideoMAE has no CLS token, so pooling is a mean over the token axis.
+    """
+
+    def _post_load(self) -> None:
+        state_dict: dict[str, torch.Tensor] | None = None
+
+        def tensors() -> dict[str, torch.Tensor]:
+            nonlocal state_dict
+            if state_dict is None:
+                state_dict = _load_checkpoint_tensors(self.model_name, self.revision)
```
Samoed: Why are you not using `_load_checkpoint_tensors`?

mteb/models/model_implementations/videomae_timesformer.py
```diff
@@ -0,0 +1,507 @@
...
+#
+# Every published TimeSformer checkpoint is supervised-finetuned on
+# Kinetics-400, Kinetics-600, or Something-Something V2. There is no
+# pretrained-only variant, so the corresponding mteb tasks are declared in
+# `training_datasets` for all of them.
+# ---------------------------------------------------------------------------
+
+timesformer_base_k400 = ModelMeta(
+    loader=TimesformerWrapper,
+    name="facebook/timesformer-base-finetuned-k400",
```
Samoed: I think this better to move to other file and separete PR

mteb/models/model_implementations/videomae_timesformer.py
hubielu: Because transformers 5.x silently drops those weights, ie, loading videomae-base on 5.14.1 prints:

encoder.layer.{0...11}.attention.attention.query.bias | MISSING
encoder.layer.{0...11}.attention.attention.key.bias   | MISSING
encoder.layer.{0...11}.attention.attention.value.bias | MISSING

4.x stored them as q_bias/v_bias alongside bias=False Linears; 5.x switched to standard Linears with their own biases and shipped no conversion mapping, so every published checkpoint loses them and gets zeros. It isn't in upstream's examples because those predate the 5.x rewrite. Effect is small but real, 0.4037 vs 0.4079 on HMDB51Clustering. The code no-ops when the biases load correctly, so it covers 4.x and any future upstream fix.

fc_norm is the same class of problem: VideoMAEModel only keeps the final LayerNorm when use_mean_pooling is False. The finetuned checkpoints set it True, moving that LayerNorm onto the classification head where AutoModel discards it, so without loading it the self-supervised and finetuned checkpoints would run through different pipelines.

mteb/models/model_implementations/videomae_timesformer.py
hubielu: Done, TimeSformer is now in #5100 with its own file. This PR is now VideoMAE only.

mteb/models/model_implementations/videomae_timesformer.py
hubielu: Good catch, no reason for the wrapper. Fixed in 74bd275 to call _load_checkpoint_tensors directly.

mteb/models/model_implementations/videomae_timesformer.py
Samoed: > Because transformers 5.x silently drops those weights

Can you pin to transformers version that don't need this then?

hubielu: Yep, pinned the VideoMAE models to the existing transformers-v4 requirement (transformers<5) and removed the q/v bias restoration. Verified the checkpoints load on transformers 4.57.6 without it. I kept the fc_norm handling since that's independent of the attention change.

mteb/models/model_implementations/videomae_models.py
```diff
@@ -0,0 +1,283 @@
...
+            num_frames if num_frames is not None else self.model.config.num_frames
+        )
+        if self.num_frames != self.model.config.num_frames:
+            raise ValueError(
+                f"{model_name} was trained with {self.model.config.num_frames} frames "
+                f"per clip, but num_frames={self.num_frames} was requested."
+            )
+
+    def _load_fc_norm(self, state_dict: dict[str, torch.Tensor]) -> None:
+        """Load the final LayerNorm when it lives on the classification head.
```
Samoed: I don't think you need classification head since you're taking embeddings before it

hubielu: Good call, removed.

mteb/models/model_implementations/videomae_models.py
```diff
@@ -0,0 +1,224 @@
...
+        # and transformers does not interpolate it, so the config is the source
+        # of truth rather than a per-ModelMeta constant.
+        self.num_frames = (
+            num_frames if num_frames is not None else self.model.config.num_frames
+        )
+        if self.num_frames != self.model.config.num_frames:
+            raise ValueError(
+                f"{model_name} was trained with {self.model.config.num_frames} frames "
+                f"per clip, but num_frames={self.num_frames} was requested."
+            )
```
Samoed, line 97:
```suggestion
        if self.num_frames != self.model.config.num_frames:
            logger.warning(
                f"{model_name} was trained with {self.model.config.num_frames} frames "
                f"per clip, but num_frames={self.num_frames} was requested."
            )
```

hubielu:
@Samoed anything outstanding here?

mteb/models/model_implementations/videomae_models.py
```diff
@@ -0,0 +1,227 @@
...
+            "cuda"
+            if torch.cuda.is_available()
+            else "mps"
+            if torch.backends.mps.is_available()
+            else "cpu"
+        )
+
+        self.processor = AutoImageProcessor.from_pretrained(
+            model_name, revision=revision
+        )
```
Samoed: Seems that this model supports processing both images and video. For video `VideoProcessor` should be used

hubielu: This conflicts with the v4 pin on this PR

On 5.14.1 you're right, videomae is in VIDEO_PROCESSOR_MAPPING_NAMES and AutoVideoProcessor resolves to VideoMAEVideoProcessor. On 4.x it isn't registered at all, so that call raises

The pin is there because v5 silently drops q_bias/v_bias. So it's either the v4 pin with AutoImageProcessor as it is now, or drop the pin, target v5 with AutoVideoProcessor, and restore the biases in the wrapper.

Which would you prefer?

Samoed: Can you compare performance?

hubielu: Switched to AutoVideoProcessor and dropped the v4 pin

Same score, and faster. HMDB51Clustering v_measure on videomae-base:

| transformers | processor | q/v biases | v_measure | encode |
|---|---|---|---|---|
| 4.52.4 (pinned) | AutoImageProcessor | native | 0.407616 | 12m00 |
| 5.14.1 | AutoVideoProcessor | restored | 0.407899 | 5m48 |
| 5.14.1 | AutoImageProcessor | restored | 0.407899 | 6m21 |

The bottom two are identical to 6dp, so the processor class doesn't affect the score, and VideoProcessor is the fastest. q_bias/v_bias are restored in the wrapper instead of relying on the pin, so the models now run in the default environment.

videomae-base-finetuned-kinetics is 0.699565 under the new config

Samoed: Should you pin then `transformers>5`?

hubielu: Added transformers-v5 = ["transformers>=5"]

Samoed (approved):

PR #5100, opened 2026-08-07
model: add TimeSformer video encoders

hubielu:
Three TimeSformer checkpoints: timesformer-base-finetuned-k400, timesformer-base-finetuned-ssv2, timesformer-hr-finetuned-k400. Split out of #5094 at @Samoed's suggestion.

Drafted with Claude; I ran and verified everything below.

**Results.** HMDB51Clustering v_measure for timesformer-base-finetuned-k400: 0.7287.

TimeSformer prepends a CLS token and its classification head reads last_hidden_state[:, 0], so the wrapper pools the same way rather than averaging.

I use AutoImageProcessor rather than AutoVideoProcessor, since TimesformerVideoProcessor only exists in transformers 5.x while mteb supports 4.40 and up. It also matters for correctness: that class defaults to ImageNet mean and std, while every checkpoint config declares VideoMAEImageProcessor with its own values. Frames are passed as HWC lists because a list of 4D tensors trips make_batched on the v4 slow processor.

num_frames comes from config.num_frames, 8 for base and 16 for hr, since the temporal position embeddings are not interpolated.

Every published TimeSformer checkpoint is supervised on Kinetics-400, Kinetics-600, or Something-Something V2; there is no pretrained-only variant, so those tasks are declared in training_datasets. Worth knowing for the leaderboard, since running these on Kinetics400V measures label memorisation.

mteb mock-run passes 4/4 video tasks on all three.

### DIFF
--- a/mteb/models/model_implementations/timesformer_models.py
+++ b/mteb/models/model_implementations/timesformer_models.py
@@ -94,16 +94,7 @@ class TimesformerWrapper(AbsEncoder):
         embeddings = []
         for batch in tqdm(inputs, desc="Encoding", disable=not show_progress_bar):
             videos = batch["video"]
-            padded = [
-                torch.cat(
-                    [v, v[-1:].expand(self.num_frames - v.shape[0], *v.shape[1:])],
-                    dim=0,
-                )
-                if v.shape[0] < self.num_frames
-                else v[: self.num_frames]
-                for v in videos
-            ]
-            processed = self.processor(padded, return_tensors="pt").to(self.device)
+            processed = self.processor(videos, return_tensors="pt").to(self.device)
 
             outputs = self.model(**processed)
             embeddings.append(outputs.last_hidden_state[:, 0].cpu())
--- a/mteb/models/model_implementations/timesformer_models.py
+++ b/mteb/models/model_implementations/timesformer_models.py
@@ -94,7 +94,11 @@ class TimesformerWrapper(AbsEncoder):
         embeddings = []
         for batch in tqdm(inputs, desc="Encoding", disable=not show_progress_bar):
             videos = batch["video"]
-            processed = self.processor(videos, return_tensors="pt").to(self.device)
+            # Explicit list-of-videos-of-frames in HWC. A list of 4D tensors trips
+            # `make_batched` on the v4 slow processor, which treats the whole batch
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/timesformer_models.py
```diff
@@ -0,0 +1,219 @@
...
+            videos = batch["video"]
+            padded = [
+                torch.cat(
+                    [v, v[-1:].expand(self.num_frames - v.shape[0], *v.shape[1:])],
+                    dim=0,
+                )
+                if v.shape[0] < self.num_frames
+                else v[: self.num_frames]
+                for v in videos
+            ]
```
Samoed: Why padding is needed? This won't by done by processor?

mteb/models/model_implementations/timesformer_models.py
```diff
@@ -0,0 +1,219 @@
...
+    model_type=["dense"],
+    citation=_TIMESFORMER_CITATION,
+    contacts=None,
+    output_dtypes=None,
+    extra_requirements_groups=None,
+)
+
+timesformer_hr_k400 = ModelMeta(
+    loader=TimesformerWrapper,
+    name="facebook/timesformer-hr-finetuned-k400",
```
Samoed, line 188: These models tuned for classification and I don't think this is worth to adding them

mteb/models/model_implementations/timesformer_models.py
hubielu: You're right that the extra padding isn't needed. FramesCollator(num_frames=self.num_frames) already selects exactly the required number of frames, so I removed the redundant padding/truncation before the processor.

mteb/models/model_implementations/timesformer_models.py
hubielu: Yeah, all three available TimeSformer checkpoints are classification-finetuned. Would you prefer that I drop the TimeSformer addition entirely?

hubielu: @Samoed Hi Samoed, just wanted to follow up on this one

Samoed: Yes, I don't think that we want to add it

PR #5102, opened 2026-08-07
fix: leaderboard partial split aggregation

### LINKED ISSUE #5101: Leaderboard handles partly existing results incorrectly
### Describe the bug

> Thanks for pointing out `validate_and_filter_scores=True`. I agree that strict validation is available when results are loaded explicitly with that option.
> 
> My concern is that this does not appear to be the behavior used in the public benchmark summary views:
> 
>  - When filtering directly to [`WebLINXCandidatesReranking`](https://mteb-leaderboard.hf.space/benchmark/MTEB(Multilingual%2C%20v2)?types=Reranking&tasks=WebLINXCandidatesReranking), Querit-4B is shown with a score of **30.87**, even though its result contains only `validation` and `test_iid`.
> 
> - When filtering only by [`Task Type = Reranking`](https://mteb-leaderboard.hf.space/benchmark/MTEB(Multilingual%2C%20v2)?types=Reranking), the displayed aggregate also includes this **30.87** as Querit-4B's WebLINX task score.
> 
> 
> Other models in the same leaderboard column, such as jina-reranker-v3, are represented by an average over all six splits. Therefore, the user-facing summary still compares numeric scores computed over different split sets, even though the task-level split view makes the missing results visible.
> 
> In other words, the missing-split information is available, but it does not prevent the incomplete aggregate from being displayed and used in the benchmark summary. A user looking only at the summary table would reasonably interpret these values as directly comparable.
> 
> Would it make sense for the leaderboard summary to use strict split validation by default? An incomplete result could either be excluded from the aggregate, shown as `NaN`, or displayed with a prominent incomplete-splits indicator instead of a normal numeric score.

By @terryqj0107 from https://github.com/embeddings-benchmark/mteb/issues/5092#issuecomment-5211590079

### To reproduce

In descriptioon

### Additional information

_No response_

No

Samoed:
Close https://github.com/embeddings-benchmark/mteb/issues/5101

Previously polars path didn't check if all splits/subsets are presented and averaged against non missing scores

Test on LB:
image


KennethEnevoldsen (approved):
Nothing to add here I think it looks good!

PR #5103, opened 2026-08-07
test: Remove gradio lb tests and ci

Samoed:
Since we moved to api leaderboard and we don't plan to maintain it I think we can remove tests and CI


isaac-chung (approved):

KennethEnevoldsen:
Do we want to completely remove the leaderboard? I suspect we might break it at some point without knowing?

Samoed:
> Do we want to completely remove the leaderboard?

I don't think so, at least for now. Maybe some users are using them locally for their private evalutations

>  I suspect we might break it at some point without knowing?

Yes, we might

PR #5107, opened 2026-08-08
feat: Functionality to clean task and remove low-quality samples

ayush1298:
related to #3672 

This PR adds `mteb.quality`, a place for filters that remove low-quality samples from a task before it is evaluated. Deduplication is the first filter. 

```python
from mteb.quality import remove_duplicates

cleaned = remove_duplicates(task)

print(cleaned.metadata.name)
# MassiveIntentClassification (remove_duplicates)
print({split: len(data) for split, data in cleaned.dataset["en"].items()})
# {'train': 11468, 'test': 2970, 'validation': 2031}, from 11514 / 2974 / 2033
```

What this adds: 

1. `remove_duplicates(task)` drops repeated samples, keeping the first of each. It does not modify task in-place. It returns a cleaned copy, so the original stays usable. `normalization=` takes any `Callable[[str], str]` and decides when two texts count as equal; `columns=, splits=, subsets= and num_proc= `narrow what it touches.
2. A cleaned task is a new task, named after the filters applied to it: `MassiveIntentClassification (remove_duplicates)`, and `MassiveIntentClassification (remove_duplicates, filter_short`) once there is a second filter. It can be evaluated like any other task, and its scores are recorded against that ID rather than against the published dataset. `unload_data()` restores the published data and the published name.
3. Every modality is covered. Text compares as text; images, audio and video compare by a content hash, reusing the hash functions the descriptive statistics already use. Verified on a real task per modality combination. 1389 tasks declare their content columns; the 9 that do not are aggregates, which forward to their sub-tasks.
4. Symmetric tasks compare pairs order-insensitively, so (a, b) and (b, a) are duplicates. This follows the symmetric flag the pair statistics already use, so only STS opts in. (Not Pair Classification)
5. Retrieval keeps its qrels valid. A judgement pointing at a removed duplicate moves to the copy that was kept, making deduplication lossless, and a query left without a positive is dropped. A corpus document is compared on its title as well as its text, matching how it is encoded - without that, documents the model sees as different were being collapsed.
6.`mteb/_content_hashes.py` now holds the per-modality hashing, shared by the descriptive statistics and by `mteb.quality` instead of one reaching into the other.
7. No change to the `TaskResult` schema, so existing result files are unaffected.
8. New guide at `docs/get_started/advanced_usage/cleaning_task_data.md` covering how to spot these issues in the descriptive statistics and how to contribute a fix as a new task version.


ayush1298:
Adding the rest of the details as a separate comment, as these are necessary, but it can make the PR description too long (more than 2000 chars limit)

Verification and other important things:

1. On verifying these on `MassiveIntentClassification`, 52 duplicates + 46 short documents removed; 6.9 MB peak memory.
2. SciFact (retrieval) was already clean, so 0 removed; corpus/queries/qrels untouched
3. normalize on real data - `strip/casefold → 11468 train rows`, `alphanumeric → 11467`
4. Cache bypass, end-to-end - clean run cached at −0.8660; same task unmodified still served from cache; after `remove_duplicates()` re-evaluated to −1.0000 under `overwrite_strategy="never",` with both warnings raised
5. No-op filters - a filter that removes nothing leaves `data_modified` `false` and the cache working as before
6. Tests are added to specifically check the data cleaning in `tests/test_abstasks/test_data_cleaning.py.`

A few important things that I have not implemented and need opinions on before doing them:

1. Should `data_modified` be recorded in the result JSON?:  A run on a cleaned task currently produces a `TaskResult` that is indistinguishable on disk from a regular one. It records the published `task_name` and `dataset_revision`. Only a warning and the cache bypass guard it. `model_dump()` writes every field, so adding one would add a key to every file in the `results` repo. Do you want it persisted, and if so how (new field, or something narrower)?
2. `remove_train_leakage()` - is in scope, and what does it mean for retrieval?: Dropping eval documents that also appear in train is a common ask, and the existing`_data_filter/filters.py` already has the logic. I have not implemented it yet as it needs some clarifications before stepping forward:
    - For retrieval, a corpus shared across splits is normal rather than leakage. Should it consider queries only, or raise `NotImplementedError` there?
    - "Train split" is not a universal concept, as `train_split` exists on `AbsTaskClassification` but not on STS, pair classification, etc. So, should we add an explicit `train_split="train"` argument?
    - Unlike the two filters here, it changes what the benchmark measures rather than just dropping low-quality rows. So, main question is: should it belong in this API at all?
3. What should happen to `mteb/abstasks/_data_filter/filters.py` and `task_pipelines.py`?: They are classification-specific scripts added, but not used anywhere currently. Three of their filters are reproduced exactly by this PR (`deduplicate`, `filter_empty`, `filter_short`.While below are some filters that are not implemented yet:
    - `filter_train_leakage` - mentioned in question 2.
    - `filter_unclear_label`, which drops texts appearing with contradictory labels across splits. This fits "remove low-quality samples" well; should it become `remove_conflicting_labels()`? It is classification-family only, since the label column has no analogue for STS scores.
    - `split_train_test`, which *creates* a test split by stratified sampling and mutates`metadata.eval_splits`. That reads as dataset restructuring rather than filtering, so I left it out.
    - Their language special-case (skipping the length filter for zho/jpn/tha/mya/cmn) is not needed here. It existed because the old filter counted words, and this one counts characters by default.

4. Should retrieval deduplication compare `title` as well as `text`?: It defaults to `["text"]` only, so two documents with the same body but different titles count as duplicates. `columns=["title", "text"]` gives the stricter behaviour, which should be the default?
5. stale descriptive statistics(Minor): `task.metadata.descriptive_stats` still describes the published dataset after filtering. This PR only warns. Should the cached statistics be invalidated instead?

I know that these become too long, and sorry for that. I tried to be as specific and concise as possible, but I needed to add background context for each of them to make it easier for you. Will appreciate your opinions! @KennethEnevoldsen @Samoed @isaac-chung

mteb/abstasks/abstask.py
```diff
@@ -592,6 +617,213 @@ def filter_languages(
...
+        """
+        if self.dataset is None:
+            raise ValueError(f"Dataset of task '{self.metadata.name}' is not loaded.")
+
+        first_value = next(iter(self.dataset.values()), None)
+        if isinstance(first_value, Dataset):
+            return {"default": cast("DatasetDict", self.dataset)}
+        return self.dataset
+
+    def _filter_dataset_rows(
```
Samoed: I'm not sure about integrating these functions as part of tasks. Maybe implement it similarly to `ReversibleWorkflow`

ayush1298: Any reason for that? I guess if we have to implement it similarly to `ReversibleWorkflow` then many things need to be added, though I am not sure.

Samoed: Not directly like `ReversibleWorkflow`, but just pipeline. I don't think that these function should be a part AbsTask. AbsTask is already quite huge and complex and I don't think we want to add much new functions to it

mteb/abstasks/abstask.py
```diff
@@ -592,6 +617,213 @@ def filter_languages(
...
+            self.load_data(num_proc=num_proc)
+
+        text_columns = (
+            list(columns) if columns is not None else self._get_text_columns()
+        )
+        if not text_columns:
+            raise NotImplementedError(
+                f"`{filter_name}` does not know which columns of '{self.metadata.name}' hold text. This is "
+                "expected for tasks without a text modality; pass `columns=[...]` to filter on specific columns."
+            )
```
Samoed: I think better to expand this function for multimodal datasets

mteb/abstasks/abstask.py
ayush1298: Agree with these. So, it will no longer be used as `task.remove_duplicates()` or `task.filter_short_documents()`. But basically, it will be a pipeline where we can pass a task and add all things we want to do in filtering, with parameters corresponding to each of these functions, and then at the end it will return the updated task object. Are you saying something like that?

Samoed:
> remove_train_leakage() - is in scope, and what does it mean for retrieval?: Dropping eval documents that also appear in train is a common ask, and the existing_data_filter/filters.py already has the logic. I have not implemented it yet as it needs some clarifications before stepping forward:
    For retrieval, a corpus shared across splits is normal rather than leakage. Should it consider queries only, or raise NotImplementedError there?
    "Train split" is not a universal concept, as train_split exists on AbsTaskClassification but not on STS, pair classification, etc. So, should we add an explicit train_split="train" argument?
    Unlike the two filters here, it changes what the benchmark measures rather than just dropping low-quality rows. So, main question is: should it belong in this API at all?

This function should be classification specific I think, but we also can implement function that would check data across splits. 


> stale descriptive statistics(Minor): task.metadata.descriptive_stats still describes the published dataset after filtering. This PR only warns. Should the cached statistics be invalidated instead?

I don't think we should do anything with statistics. Statistics exist for alreaddy implemented tasks and we shoudn't update any of existing tasks

mteb/abstasks/abstask.py
Samoed: Yes, this sounds good

ayush1298: okay, will wait for @KennethEnevoldsen opinion before changing to it, if he had anything further to add here

docs/api/task.md
```diff
@@ -51,6 +51,50 @@ All tasks in `mteb` inherits from the following abstract class.
 
 :::mteb.AbsTask
 
+## Cleaning Task Data
```
KennethEnevoldsen: add reference to this in adding a dataset 

Also I would probably put this under get started > examples

docs/api/task.md
```diff
@@ -51,6 +51,50 @@ All tasks in `mteb` inherits from the following abstract class.
...
+```python
+import mteb
+
+task = mteb.get_task("MassiveIntentClassification")
+task.remove_duplicates()
+task.filter_short_documents(min_length=5)  # or min_length=3, unit="words"
+```
+
+Both methods load the data if needed, modify the dataset in place and return the task, so they can be chained.
+They cover every split and subset by default; `splits=`, `subsets=` and `columns=` narrow that down. For retrieval
```
KennethEnevoldsen: this is API docs - just link to the function signature instead

docs/api/task.md
```diff
@@ -51,6 +51,50 @@ All tasks in `mteb` inherits from the following abstract class.
...
+## Cleaning Task Data
+
+Datasets often contain duplicated or near-empty documents. A task can remove those before it is evaluated:
+
+```python
+import mteb
+
+task = mteb.get_task("MassiveIntentClassification")
+task.remove_duplicates()
+task.filter_short_documents(min_length=5)  # or min_length=3, unit="words"
```
KennethEnevoldsen:
```suggestion
task.filter_short_documents(min_length=5, unit="words")
```
I would write it like this if you want to show the defaults

btw. units=word I think is a poor way to approach it (it assumed we can perfectly split on word)

docs/api/task.md
```diff
@@ -51,6 +51,50 @@ All tasks in `mteb` inherits from the following abstract class.
...
 
+## Cleaning Task Data
+
+Datasets often contain duplicated or near-empty documents. A task can remove those before it is evaluated:
+
+```python
+import mteb
+
+task = mteb.get_task("MassiveIntentClassification")
+task.remove_duplicates()
```
KennethEnevoldsen: I would split this block in two and first explain deduplication and then short text filtering. The example also need to show the descriptive stats as a way of identifying the dataset.

docs/api/task.md
```diff
@@ -51,6 +51,50 @@ All tasks in `mteb` inherits from the following abstract class.
...
 :::mteb.AbsTask
 
+## Cleaning Task Data
+
+Datasets often contain duplicated or near-empty documents. A task can remove those before it is evaluated:
+
+```python
+import mteb
+
+task = mteb.get_task("MassiveIntentClassification")
```
KennethEnevoldsen: this changes the task so of course the task is no longer the same so running on it should not produce `MassiveIntentClassification` results. We need to figure out how to best do this

mteb/abstasks/abstask.py
KennethEnevoldsen: Yeah I can see it both ways, but I agree that I would prefer not to have the complexity on AbsTask.

@Samoed are you suggesting something like:

```
from mteb.quality import remove_duplicates

task = remove_duplicates(task)
```

mteb/abstasks/abstask.py
KennethEnevoldsen: Yeah I would also suggest focusing on a few filters at the start and then we can expand e.g. start with deduplication. That way we can tackle the overall problems first and future additions should be much easier

docs/api/task.md
KennethEnevoldsen: I think maybe a solution could be:

change name to `MassiveIntentClassification*` indicating that it is a transformed version. Or specify the transformation within the name `MassiveIntentClassification (removed duplicates)`.

docs/api/task.md
KennethEnevoldsen: In general I would love the section to read like:

Some dataset have issues, e.g. 
- you can see it here (desc stats, link to desc stats)
- then fix those issues > see how it changes scores (and task name) 
- add a note on that we do not accept scores on manipulated tasks, but that you can submit a updated version of a task, by bumping the version id (link to submission)

thoughout add link to API docs of cleaning functions

KennethEnevoldsen (commented):
Great to see the PR! I would simplify down to one filter and then we can in future PRs add more -- then we can figure out the best structure and then adding new should be very simple.

mteb/abstasks/abstask.py
Samoed: Yes, something like this

mteb/abstasks/abstask.py
ayush1298: done

mteb/abstasks/abstask.py
ayush1298: changing to these

docs/api/task.md
ayush1298: okay,leaving this for now as this I have target this PR with just deduplication filter.

docs/api/task.md
ayush1298: Added for deduplication.

docs/api/task.md
ayush1298: @Samoed mentioned above that we shouldn't change the statistics of the existing task, which I also agree. So, probably just renaming it like `MassiveIntentClassification*` is better. But probably if we want to use these on LB, then we have to see how to do that? Maybe more better naming convention as these will confuse peoples otherwise

Samoed: I think we can just update description of new task

docs/api/task.md
ayush1298: done

docs/api/task.md
ayush1298: done. Can you check updated docs

ayush1298:
I have targeted this PR for only the deduplication filter and other functionality around it. So, once these is there, adding others will be easy. Also, I have mentioned the other filters list [here](https://github.com/embeddings-benchmark/mteb/issues/3672#issuecomment-5290362874). Also, I have updated the docs with some parts left to discuss and finalized.

mteb/evaluate.py
```diff
@@ -369,9 +369,14 @@ def _check_cache(
...
     """
-    # Load results from the cache if the overwrite strategy allows it
+    # Load results from the cache if the overwrite strategy allows it. Cached results describe the published
+    # dataset, so they say nothing about a task whose data was modified locally.
     existing_results: TaskResult | None = None
-    if cache and overwrite_strategy != OverwriteStrategy.ALWAYS:
+    if (
+        cache
+        and overwrite_strategy != OverwriteStrategy.ALWAYS
+        and not task.data_modified
```
Samoed: I think you can just raise error in `evaluate` if any task have `task.data_modified=True`

mteb/quality/_apply.py
```diff
@@ -0,0 +1,261 @@
...
+    from mteb.types import HFSubset, Modalities
+
+    from ._row_filters import KeepIndicesFn, TextNormalization
+
+logger = logging.getLogger(__name__)
+
+
+def _warn(msg: str) -> None:
+    logger.warning(msg)
+    warnings.warn(msg, stacklevel=3)
```
Samoed: I don't think we need warning warn here because this is "expected" behaviour 
```suggestion
```

mteb/quality/_apply.py
```diff
@@ -0,0 +1,261 @@
...
+        return
+
+    for subset, dataset_dict in datasets_by_subset(task).items():
+        if task.train_split not in dataset_dict:
+            continue
+        train_labels = Counter(dataset_dict[task.train_split][task.label_column_name])
+
+        # stratified subsampling and the train/test split both need at least two examples per label
+        too_rare = sorted(
+            str(label) for label, count in train_labels.items() if count < 2
```
Samoed: I think we can move count to arg

mteb/quality/_apply.py
```diff
@@ -0,0 +1,261 @@
...
+        col_modalities = {column: col_modalities[column] for column in columns}
+    if not col_modalities:
+        raise NotImplementedError(
+            f"`{filter_name}` does not know which columns of '{task.metadata.name}' hold its content. Please "
+            "open an issue at https://github.com/embeddings-benchmark/mteb/issues so the task can declare them."
+        )
+    return col_modalities
+
+
+def filter_task_rows(
```
Samoed: Add check on modalities, for now. You currently support only text for now

mteb/quality/_deduplication.py
```diff
@@ -0,0 +1,83 @@
...
+
+    from mteb.abstasks.abstask import AbsTask
+    from mteb.types import HFSubset
+
+    from ._row_filters import TextNormalization
+
+T = TypeVar("T", bound="AbsTask")
+
+
+def remove_duplicates(
```
Samoed: I think we might want to check for pair input tasks cases like
```
row1 sentence1 sentence2 1
row2 sentence2 sentence1 3
```

mteb/quality/_row_filters.py
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+
+def _count_values(dataset: Dataset, column: str, grouped: bool) -> int:
+    if not grouped:
+        return len(dataset)
+    return sum(len(values) for values in dataset[column])
+
+
+def apply_row_filter(
```
Samoed: Shoudn't this be private?
```suggestion
def _apply_row_filter(
```

mteb/quality/_retrieval.py
```diff
@@ -0,0 +1,160 @@
...
+            key = row_key(row)
+            if i in keep_set:
+                canonical.setdefault(key, ids[i])
+            elif (target := canonical.get(key)) is not None:
+                replacements[ids[i]] = target
+
+    return dataset.select(keep), kept_ids, replacements
+
+
+def filter_retrieval_split(  # noqa: PLR0914
```
Samoed:
```suggestion
def _filter_retrieval_split(  # noqa: PLR0914
```

mteb/quality/_retrieval.py
Samoed, line 1: I think would be better to put all filtering functions in same file. And split per task type files task type specific filters. Currently very hard to navigate in code

Samoed (commented):
Can you update example in PR description?

mteb/quality/_retrieval.py
ayush1298: moved to all of them to single `_filters.py` file.

mteb/quality/_apply.py
ayush1298: done

mteb/evaluate.py

mteb/quality/_apply.py
ayush1298: Modality check added to `_resolve_columns` which raises `NotImplementedError` naming the offending modalities and the supported set. Also Text, image, audio, and video all work through `_HASH_FN` (I have verified with `Caltech101)`

mteb/quality/_deduplication.py
ayush1298: Symmetric pairs are already accounted for in MTEB, where `AbsTaskSTS` passes `symmetric=True` to `calculate_pair_modality_statistics', so ("sent1", "sent2") and ("sent2", "sent1") count as a single unique pair in the descriptive statistics.  So, I am following the same rule in Deduplication.

Added `AbsTask._get_symmetric_sides()` returning `None` by default and overridden in `AbsTaskSTS` When set, `_iter_row_keys` orders the two sides within each row, so a pair and its swap hash are identical. Your example : (s1, s2, 1) and (s2, s1, 3) now collapse to the first row.

1. Also, `Pair classification` stays order-sensitive because in statistics symmetric defaults to False.
2. Narrowing with columns= disables it. If the selection doesn't cover both sides exactly, swapping is meaningless, so `_resolve_symmetric_sides` returns None. 
3. The sides are declared as column lists, not single columns, so multimodal STS (side = video + audio) swaps whole sides rather than shuffling individual columns.

docs/api/task.md
ayush1298: We have to change the name as well, right?

ayush1298:
Resolves most of the comments; changes included are: 
Added all filters to a single file, made all functions private, added a modality check and a threshold for min examples per label, updated docs, and added support for pair input tasks
Also, updated PR description with correct example and scope of this PR.

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,95 @@
...
+task = remove_duplicates(
+    task, normalize="alphanumeric"
+)  # "Wake me up!" == "wake  me  up"
+```
+
+Text is compared as text, while images, audio and video are compared by a hash of their content, so the filter
+works on any task. Retrieval tasks keep their relevance judgements valid: a judgement pointing at a removed
+duplicate moves to the copy that was kept.
+
+## Changes after cleaning Task Data
```
ayush1298: @Samoed @KennethEnevoldsen Can you read all parts below these? I have added it temporarily, and some of it needs to be changed

docs/api/task.md
Samoed: Yes

ayush1298: What it should be, just next version like `.v2, .v3` like that?

Samoed: I think you can add cleared function name

docs/api/task.md
```diff
@@ -51,6 +51,12 @@ All tasks in `mteb` inherits from the following abstract class.
 
 :::mteb.AbsTask
 
+## Cleaning Task Data
```
Samoed: I think this should be in the bottom, because firstly this page about tasks

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,95 @@
...
+print(stats["text_statistics"]["unique_texts"])  # 2970
+print(stats["text_statistics"]["min_text_length"])  # 2
+```
+
+Fewer unique texts than samples means the split contains duplicates -- four of them here. A small `min_text_length` points the other way, at documents too short to be meaningful.
+
+For a task you are developing, compute the statistics yourself with
+[`task.calculate_descriptive_statistics()`][mteb.AbsTask.calculate_descriptive_statistics].
+
+## Removing duplicates
```
Samoed, line 36: I'd add a list with supporting clearing functions and add a links to chapters with them

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,95 @@
...
+
+Text is compared as text, while images, audio and video are compared by a hash of their content, so the filter
+works on any task. Retrieval tasks keep their relevance judgements valid: a judgement pointing at a removed
+duplicate moves to the copy that was kept.
+
+## Changes after cleaning Task Data
+
+Cleaning changes the descriptive statistics, and it would change the scores too, because the model would be
+evaluated on different data. What it does *not* change is the task's name or its `dataset_revision`, so such a
+score would be indistinguishable from one computed on the published dataset while not being comparable to it.
```
Samoed:
```suggestion
Cleaning changes the data, and it would change the scores too, because the model would be
evaluated on different data. What it does *not* change is the task's name or its `dataset_revision`, so such a
score would be indistinguishable from one computed on the published dataset while not being comparable to it.
```

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,95 @@
...
+
+For that reason a cleaned task is marked with `data_modified` and [`mteb.evaluate`][mteb.evaluate] refuses to run
+it:
+
+```python
+mteb.evaluate(model, [task])
```
Samoed:
```suggestion
mteb.evaluate(model, task)
```

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,95 @@
...
+# ValueError: The data of ['MassiveIntentClassification'] was modified locally ...
+```
+
+Reloading the task with [`task.unload_data()`][mteb.AbsTask.unload_data] restores the published data and clears the
+mark.
```
Samoed: I don't think we need to mention this
```suggestion
```

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,95 @@
...
+    [leaderboard](https://huggingface.co/spaces/mteb/leaderboard). They are not comparable to any other result, and
+    nothing in the result file would distinguish them.
+
+## Contributing the fix
+
+If a dataset needs cleaning, everyone benefits from fixing it once rather than in each user's script. Submit the
+cleaned data as a **new version of the task** rather than as scores. A new version:
+
+- bumps the version suffix of the task name, so `MassiveIntentClassification` becomes
+  `MassiveIntentClassification.v2`. The suffix replaces rather than accumulates, so a task that is already at `.v2` becomes `.v3`.
```
Samoed: I think we should develop better strategy of naming

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,95 @@
...
+- points `adapted_from` at the task it was derived from, and sets `superseded_by` on every older version so that
+  users of those are warned towards the current one.
+
+See [Adding a Task](../../contributing/adding_a_dataset.md) for the full process, and
+[`push_dataset_to_hub`][mteb.AbsTask.push_dataset_to_hub] for uploading the cleaned data.
+`scripts/data/clean_and_update_tasks.py` automates the version bump, including finding the highest existing
+version.
```
Samoed:
```suggestion
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,552 @@
...
+from collections.abc import (
+    Callable,
+    Iterable,
+    Mapping,  # noqa: TC003
+)
+from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
+
+from datasets import Dataset
+
+from mteb.abstasks._statistics_calculation import _HASH_FN
```
Samoed: I think would be better to moveout from statistics to separete file for reusability

docs/get_started/advanced_usage/cleaning_task_data.md
Samoed, line 1: Ovearall, I don't think we want to add to the docs at least until we support all modalities

mteb/evaluate.py
```diff
@@ -264,6 +264,35 @@ def _evaluate_task(  # noqa: PLR0913, PLR0914
...
+    else:
+        check_tasks = tasks
+
+    modified = sorted(task.metadata.name for task in check_tasks if task.data_modified)
+    if modified:
+        raise ValueError(
+            f"The data of {modified} was modified locally, e.g. by a filter from `mteb.quality`, so these tasks "
+            "cannot be evaluated: the scores would not be comparable to any other result while being "
+            "indistinguishable from one. To contribute the cleaned data, submit it as a new version of the task "
+            "instead; see https://embeddings-benchmark.github.io/mteb/contributing/adding_a_dataset/."
```
Samoed:
```suggestion
            "instead; see https://docs.mteb.org/contributing/adding_a_dataset/."
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,552 @@
...
+    digest = hashlib.blake2b(digest_size=16)
+    for key in keys:
+        encoded = key.encode()
+        digest.update(len(encoded).to_bytes(8, "little"))
+        digest.update(encoded)
+    return digest.digest()
+
+
+def _keep_first_occurrence(keys: Iterable[tuple[str, ...]]) -> list[int]:
+    """Keep the rows whose comparison key has not been seen before.
```
Samoed: I think keys is a bit confusing, because you're working here with just data that available in dataset. When I read `key` I'm thinking about column names

mteb/quality/_classification.py
```diff
@@ -0,0 +1,61 @@
...
+
+from ._filters import _datasets_by_subset, _warn
+
+if TYPE_CHECKING:
+    from mteb.abstasks.classification import AbsTaskClassification
+
+logger = logging.getLogger(__name__)
+
+
+def _warn_about_label_distribution(
```
Samoed:
```suggestion
def _check_label_distribution(
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,552 @@
...
+    task_name: str, available: Mapping[str, Mapping[str, Any]]
+) -> str:
+    """The message raised when the `splits`/`subsets` given to a filter select nothing."""
+    listed = ", ".join(
+        f"{subset}: {sorted(splits)}" for subset, splits in available.items()
+    )
+    return f"The given splits and subsets do not select any data of '{task_name}'. The task has {listed}."
+
+
+def _warn_about_unusable_data(task: AbsTask) -> None:
```
Samoed:
```suggestion
def _check_unusable_data(task: AbsTask) -> None:
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,552 @@
...
+- `"casefold"`: also ignore case, so `"Hello"` and `"hello"` are duplicates.
+- `"alphanumeric"`: also ignore punctuation and repeated whitespace, so `"hello, world!"` and `"Hello world"` are
+    duplicates, as are `"e-mail"` and `"email"`. This is the most aggressive option and can merge documents that a
+    reader would tell apart, e.g. source code or texts where punctuation carries the meaning. Word boundaries are
+    still significant, so `"e mail"` remains distinct from `"email"`.
+
+It only applies to text. Images, audio and video are always compared by the hash of their content.
+"""
+
+KeepIndicesFn = Callable[[Iterable[tuple[str, ...]]], list[int]]
```
Samoed, line 57: I think this won't work with retrieval format which might have `title`, `text` in corpus

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: these was supposed to be version bump for dataset name, like .v2, .v3 like that. But, yeah we have to still decide for how to name. Because, when it will be used with other filters, then we can't add those filters name in the task name

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: agreed

docs/api/task.md
ayush1298: done

mteb/quality/_filters.py
ayush1298: agreed. Rename it to row/content based on what it uses in different places.

mteb/quality/_filters.py
ayush1298: Fixed it now, so it will work with retrieval. Checked with `NFCorpus,` where it removes 40 duplicate documents.
The problem was that `AbsTaskRetrieval._get_content_columns()` returned only `{"text": "text"}` and no title. It now declares `{"title": "text", "text": "text"}` for the text modality.

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: added

mteb/quality/_filters.py
ayush1298: done, moved all hash related things from these file to separate `mteb/_content_hashes.py`

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: Do you mean all filters, because it's now supporting all modalities? Added a comment on results on different modalities task

ayush1298:
Deduplication supports all the modalities, and below I am verifying them on different modality tasks: 


modalities | task | split | actually compared | published | after dedup | removed
-- | -- | -- | -- | -- | -- | --
text | MassiveIntentClassification | train | text | 11514 | 11195 | 319
image | OxfordFlowersClassification | train | image | 7169 | 7165 | 4
audio | GunshotTriangulation | train | audio | 88 | 88 | 0
image+text | BLINKIT2TRetrieval | test corpus | text | 20 | 20 | 0
audio+text | HiFiTTSA2TRetrieval | test corpus | text | 300 | 288 | 12
audio+image | SpeechCocoA2IRetrieval | test corpus | image | 2048 | 2048 | 0
video | mock | test | video | 3 | 2 | 1

docs/get_started/advanced_usage/cleaning_task_data.md
Samoed: I think your filter working with text only and most part of code expect only one sting from one row

ayush1298: Each row is a `tuple[str, ...]` which is one entry per compared column, not one per row. For non-text columns, content-hash strings are created using `MODALITY_HASH_FNS`. Updated table below with actually compared columns.

docs/contributing/adding_a_dataset.md
```diff
@@ -618,6 +618,8 @@ Before creating a pull request, it is important to calculate some descriptive st
 To calculate the descriptive statistics, you can run [`task.calculate_descriptive_statistics()`][mteb.AbsTask.calculate_descriptive_statistics].
 By default, this method caches the results and loads them from cache on subsequent calls. To recalculate the statistics (overwriting the cached results), use the `overwrite_results=True`.
 
+If the statistics show that the dataset does contain duplicates, very short documents or any other dataset cleaning, see [Cleaning task data](../get_started/advanced_usage/cleaning_task_data.md) for how to find and remove them.
```
KennethEnevoldsen:
```suggestion
If the statistics show that the dataset does contain duplicates, very short documents or other quality concerns, see [Cleaning task data](../get_started/advanced_usage/cleaning_task_data.md) for how to find and remove them.
```

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,101 @@
...
+[`task.calculate_descriptive_statistics()`][mteb.AbsTask.calculate_descriptive_statistics].
+
+## Available filters
+
+Each filter takes a task, modifies its dataset in place and returns it, so they can be chained. They cover every split and subset by default; see the linked reference for the arguments.
+
+| filter | removes |
+|---|---|
+| [`remove_duplicates`][mteb.quality.remove_duplicates] | [repeated samples](#removing-duplicates) |
+
```
KennethEnevoldsen, line 35: I think it is better to simply refer to the API section - this table will get outdated

docs/api/task.md
```diff
@@ -88,3 +89,9 @@ Tasks that support any modality (text, image, etc.) inherit from the following a
 ## Image Tasks
 
 :::mteb.abstasks.image.image_text_pair_classification.AbsTaskImageTextPairClassification
+
+## Cleaning Task Data
+
+Filters that remove low-quality samples from a task before it is evaluated. See [Cleaning task data](../get_started/advanced_usage/cleaning_task_data.md) for more details on how to use them.
+
+:::mteb.quality.remove_duplicates
```
KennethEnevoldsen: I think you can specify this so that it just takes all public functions in this module

docs/get_started/advanced_usage/cleaning_task_data.md
KennethEnevoldsen: @Samoed I see that this contradict with your comment

docs/get_started/advanced_usage/cleaning_task_data.md
KennethEnevoldsen: closing in favor of above

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,101 @@
...
+# {'train': 11468, 'test': 2970, 'validation': 2031}, from 11514 / 2974 / 2033
+```
+
+Two documents count as duplicates when they are identical once surrounding whitespace is stripped. `normalize=`
+loosens that to ignore case, or case and punctuation:
+
+```python
+task = remove_duplicates(
+    task, normalize="alphanumeric"
+)  # "Wake me up!" == "wake  me  up"
```
KennethEnevoldsen:
```suggestion
)  # "Wake me up!" == "wake me up"
```

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,101 @@
...
+evaluated on different data. What it does *not* change is the task's name or its `dataset_revision`, so such a
+score would be indistinguishable from one computed on the published dataset while not being comparable to it.
+
+For that reason a cleaned task is marked with `data_modified` and [`mteb.evaluate`][mteb.evaluate] refuses to run
+it:
+
+```python
+mteb.evaluate(model, task)
+# ValueError: The data of ['MassiveIntentClassification'] was modified locally ...
+```
```
KennethEnevoldsen: hmm I def. think we want to be able to evaluate the task. I simply think we need to make sure it is a new task

I think something like:
`MassiveIntentClassification (remove_duplicates)`

is a good new id and then of course with more then `MassiveIntentClassification (remove_duplicates, remove_short)`

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+!!! warning
+    Scores from a locally cleaned task are not accepted on the
+    [leaderboard](https://huggingface.co/spaces/mteb/leaderboard). They are not comparable to any other result, and
+    nothing in the result file would distinguish them.
```
KennethEnevoldsen: I don't think it is a warning more a note

I think I would change the text to:

> As cleaning the dataset changes the score, we do not accept scores from modified datasets on the [leaderboard](link). We do however allow a cleaned version of a dataset to [submitted to MTEB](adding a dataset section)

docs/get_started/advanced_usage/cleaning_task_data.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Contributing the fix
```
KennethEnevoldsen: I am not sure we need this section here (this is for the adding a dataset section)

docs/get_started/advanced_usage/cleaning_task_data.md
KennethEnevoldsen: Adnan also mentioned this - should we start an issue on this (I don't think it is for this PR)

mteb/quality/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+contradictory labels, and creating a test split -- that are implemented but not yet reachable from here; see
+https://github.com/embeddings-benchmark/mteb/issues/3672 for which of them to expose and how.
+"""
+
+from __future__ import annotations
+
+from ._filters import TextNormalization, remove_duplicates
+
+__all__ = [
+    "TextNormalization",
```
KennethEnevoldsen: don't expose

mteb/quality/_filters.py
```diff
@@ -0,0 +1,556 @@
...
+    from collections.abc import Iterator, Sequence
+
+    from datasets import DatasetDict
+
+    from mteb.abstasks.abstask import AbsTask
+    from mteb.types import HFSubset, Modalities
+
+logger = logging.getLogger(__name__)
+
+T = TypeVar("T", bound="AbsTask")
```
KennethEnevoldsen, line 40: Why not simply use AbsTask?

mteb/quality/_filters.py
```diff
@@ -0,0 +1,556 @@
...
+
+    logger.info(
+        f"`{filter_name}` removed {n_removed} samples from '{task.metadata.name}' "
+        f"(columns={sorted(col_modalities)})."
+    )
+    return task
+
+
+def remove_duplicates(
+    task: T,
```
KennethEnevoldsen, line 529: are you trying to indicate it is a same T? I am not sure we want to modify in-place

```
task = get_task(...)
task_cleaned = filter(task)
# these two should not be the same
# task_cleaned != task
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,556 @@
...
+    *,
+    normalize: TextNormalization = "strip",
```
KennethEnevoldsen:
```suggestion
    normalization: Callable[str, str]: lambda x: x.strip()
```
I think this is more general and more transparent. If people are cleaning data I think this should be readable.

Do we only want normalize text?

mteb/quality/_filters.py
```diff
@@ -0,0 +1,556 @@
...
+        subsets: The Huggingface subsets to filter. Defaults to every loaded subset.
+        num_proc: Number of processes to use for loading the dataset and for hashing non-text content.
+
+    Returns:
+        The task, so that calls can be chained.
+
+    Raises:
+        NotImplementedError: If the task does not declare which of its columns hold content, or holds a modality
+            that cannot be compared.
+        ValueError: If `columns`, `splits` or `subsets` select none of the task's data.
```
KennethEnevoldsen: is this a case that we see?

mteb/quality/_filters.py
```diff
@@ -0,0 +1,556 @@
...
+        >>> # also treat "Wake me up!" and "wake  me  up" as duplicates
+        >>> task = remove_duplicates(task, normalize="alphanumeric")
+    """
+    return _filter_task_rows(
+        task,
+        _keep_first_occurrence,
+        filter_name="remove_duplicates",
+        normalize=normalize,
+        remap_duplicates=True,
+        columns=columns,
```
KennethEnevoldsen: this looks kinda odd to me - remap_duplicates + `filter_name="remove_duplicates"` seems to be conditional on eachother

mteb/evaluate.py
```diff
@@ -264,6 +264,35 @@ def _evaluate_task(  # noqa: PLR0913, PLR0914
     return result
 
 
+def _check_data_not_modified(tasks: AbsTask | Iterable[AbsTask]) -> None:
```
KennethEnevoldsen: I don't think we need this

KennethEnevoldsen (commented):
great with multimodal support - still some kinks to iron out be already a lot better

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: I guess yes, but this naming is specifically for task after cleanup

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: yes, that's why I mentioned to read whole these file, like I was not sure on some of the part

ayush1298: Moved contributing to fix in `adding a dataset`?

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: added as a simple list now

mteb/quality/_filters.py
ayush1298: > Do we only want normalize text?

Yes, as Images, audio, and video go through content hashes, not doing any normalization there. 
(Maybe normalization can be applied to other modalities as well though)

mteb/quality/_filters.py
ayush1298: `remove_duplicates(sts_task)` returns `AbsTaskSTS`, not `AbsTask`; otherwise, `cleaned.column_names` fails type-checking, and every caller needs a cast.

mteb/quality/_filters.py
ayush1298: made these change, previously I was modifying in-place.

ayush1298: It's a shallow copy, and `hf_subsets` is shared. Will that be a problem?

mteb/quality/_filters.py
ayush1298: removed `NotImplementedError` and kept only `ValueError`

mteb/quality/_filters.py
ayush1298: yes, they were coupled. Fixed by grouping them in a better way

mteb/evaluate.py
ayush1298: I have removed it

ayush1298:
@Samoed @KennethEnevoldsen can you check it again?

mteb/quality/_filters.py
KennethEnevoldsen: it was more if there are cases where columns`, `splits` or `subsets` select none of the task's data.

docs/contributing/adding_a_dataset.md
```diff
@@ -618,6 +618,29 @@ Before creating a pull request, it is important to calculate some descriptive st
 To calculate the descriptive statistics, you can run [`task.calculate_descriptive_statistics()`][mteb.AbsTask.calculate_descriptive_statistics].
 By default, this method caches the results and loads them from cache on subsequent calls. To recalculate the statistics (overwriting the cached results), use the `overwrite_results=True`.
 
+If the statistics show that the dataset does contain duplicates, very short documents or other quality concerns, see [Cleaning task data](../get_started/advanced_usage/cleaning_task_data.md) for how to find and remove them.
+
+
+### Contributing the fix
```
KennethEnevoldsen:
```suggestion
### Improving or cleaning a task
```

docs/get_started/advanced_usage/cleaning_task_data.md
KennethEnevoldsen: For me it also seems like all modalities are supported for `remove_duplicates`

mteb/abstasks/abstask.py
```diff
@@ -89,6 +110,9 @@ class AbsTask(ABC):  # noqa: PLR0904
         seed: The random seed used for reproducibility.
         hf_subsets: The list of Huggingface subsets to use.
         data_loaded: Denotes if the dataset is loaded or not. This is used to avoid loading the dataset multiple times.
+        data_modified: Denotes if the dataset was modified locally by a filter from `mteb.quality`. Such a task
+            carries a name of its own, e.g. `MassiveIntentClassification (remove_duplicates)`, so that its scores
+            are not confused with results on the published dataset.
```
KennethEnevoldsen: I don't believe we need this

mteb/quality/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+```python
+import mteb
+from mteb.quality import remove_duplicates
+
+task = remove_duplicates(mteb.get_task("MassiveIntentClassification"))
+```
+
+Only deduplication is exposed for now. `mteb/abstasks/_data_filter/` holds further filters -- train/test leakage,
+contradictory labels, and creating a test split -- that are implemented but not yet reachable from here; see
+https://github.com/embeddings-benchmark/mteb/issues/3672 for which of them to expose and how.
```
KennethEnevoldsen: delete

mteb/quality/_filters.py
KennethEnevoldsen: makes sense

mteb/quality/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+from __future__ import annotations
+
+from ._filters import (
+    alphanumeric_text,
+    casefold_text,
+    remove_duplicates,
+    strip_whitespace,
```
KennethEnevoldsen: I would not make these, I think the users can make them just fine. You can show an example in the docs.
```suggestion
    remove_duplicates,
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,614 @@
...
+    task_name: str, available: Mapping[str, Mapping[str, Any]]
+) -> str:
+    """The message raised when the `splits`/`subsets` given to a filter select nothing."""
+    listed = ", ".join(
+        f"{subset}: {sorted(splits)}" for subset, splits in available.items()
+    )
+    return f"The given splits and subsets do not select any data of '{task_name}'. The task has {listed}."
+
+
+def _check_unusable_data(task: AbsTask) -> None:
```
KennethEnevoldsen: I am unsure if we want this, wouldn't we rather just have the task fail when you run it? (this constantly have to match the other code)

KennethEnevoldsen (commented):
Added feature flag, we still need some comments resolved, but I think otherwise I am quite happy with this.

Add feature flag so we will need to announce it in the "whats new section", we should probably state that more filters will be coming in the future.

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: @Samoed Can you clarify then do you meant to be something different here?

mteb/quality/_filters.py
ayush1298: @KennethEnevoldsen Do we want anything else here, or should we resolve the comment?

mteb/quality/__init__.py
ayush1298: These are used in tests as well: `tests/test_quality.py`. So, how should we handle it?

docs/get_started/advanced_usage/cleaning_task_data.md
ayush1298: This is done. Can you check it? Only naming thing is left

mteb/abstasks/abstask.py
ayush1298: removed `data_modified` from every place

mteb/quality/_filters.py
ayush1298: yes, no need. Removed these, and related code including `mteb/quality/_classification.py,` `_check_empty_retrieval_splits` from `_retrieval.py,` `_MIN_TRAIN_EXAMPLES_PER_LABEL` and related tests

mteb/quality/_filters.py
ayush1298: made it more clear like what it means: 

> ValueError: If `columns` names a column the task does not declare, or if `splits` and `subsets` together
>             match none of the task's splits.

ayush1298:
Resolved some more comments and left comments on one that are yet to be resolved. Added the whats_new section.

mteb/quality/_filters.py
KennethEnevoldsen: Hmm yeah I believe it is because then influencing one might unintentionally influence the other

mteb/quality/_filters.py
KennethEnevoldsen: So we don't expect this to happen. Are we checking for this? Why not just have it raise a downstream keyerror?

mteb/quality/__init__.py
KennethEnevoldsen: but the tests tests the functions no? I don't think we need the functions (and then also not the tests)

ayush1298: Removing these from here and updating the test accordingly, define `alphanumeric_text` only for tests

mteb/quality/_filters.py
ayush1298: `hf_subsets, rng_state, np_rng `  all were shared. So, I have defined:

```python
def _independent_copy(task: T) -> T:
    """A copy of `task` that shares no mutable state with it.

    `copy.copy` would leave the two pointing at the same list of subsets and the same random number generators, so
    working with one task could reach into the other: narrowing the copy's languages in place would narrow the
    original's, and drawing from one generator would advance the other's sampling.
    """
    cleaned = copy.copy(task)
    cleaned.hf_subsets = list(task.hf_subsets)
    cleaned.rng_state, cleaned.np_rng = _set_seed(task.seed)
    return cleaned
```

These will give the copy its own subset list and freshly seeded generators.

mteb/quality/_filters.py
ayush1298: Updated docstring:
```
Raises:
    ValueError: If `splits` and `subsets` together match none of the task's splits, which would
        otherwise filter nothing at all.
    KeyError: If `columns` names a column the task does not declare.
```
Sorry, my previous comments and the docstring both confused you a bit here. KeyError will be on columns, so if something like these:
`columns=['nope'] -> KeyError: 'nope'`
But for splits and subsets, we cant check/neither raise any error downstream error because nothing downstream fails:
`splits=['nope'] -> loop ran 0 times; no exception is possible; the copy comes back with all 3 rows.`

ayush1298:
Update the example in docs and define functions like `strip_whitespace, casefold_text and alphanumeric_text` in examples only, not exposing them. Added a fix for shallow copy and made the docstring clearer for ValueError and KeyError

docs/whats_new.md
```diff
@@ -7,6 +7,30 @@ icon: lucide/newspaper
 
 This section is an overview of releases for more information check out the autogenerated [changelog](https://github.com/embeddings-benchmark/mteb/blob/main/pyproject.toml).
 
+## New in v2.20
+
+### Cleaning low-quality samples from a task
+
+Datasets sometimes repeat the same sample many times, which distorts a benchmark by scoring the same document twice. The new `mteb.quality` module removes such samples before a task is evaluated, starting with deduplication.
```
```suggestion
It is sadly far too common to quality concern in modern benchmarks. While MTEB already have test to ensure that new additions does not have typical quality concerns, these tests does not apply to local tasks or earlier submissions. Similarly, we often accept datasets with known quality concerns to ensure compatibility with published work.

However, in an effort to make it easier to make it easier to develop high-quality we now introduce the `mteb.quality` module. It starts out with a simple deduplication function, but builds the foundation for a future addition of quality filters and it is build in a way that ensure reproducibility when changing the task.

For an example of use:
```

docs/whats_new.md
```diff
@@ -7,6 +7,30 @@ icon: lucide/newspaper
...
+from mteb.quality import remove_duplicates
+
+task = mteb.get_task("MassiveIntentClassification", languages=["eng"])
+cleaned = remove_duplicates(task)
+
+print(cleaned.metadata.name)
+# MassiveIntentClassification (remove_duplicates)
+```
+
+The task you pass in is left untouched and a cleaned copy is returned, named after the filters applied to it. That id is what keeps results honest: a cleaned task is evaluated like any other, and its scores are recorded against the cleaned name rather than against the published dataset.
```
```suggestion
The task you pass in is left untouched and a cleaned copy is returned, named after the filters applied to it. That id ensures that the cleaned and current version aren't interchanged: a cleaned task is evaluated like any other, and its scores are recorded against the cleaned name rather than against the published dataset.
```

KennethEnevoldsen (approved):
I think this is good -- @Samoed anything to add?

ayush1298:
Applied changes in `whats_new`

mteb/abstasks/abstask.py
```diff
@@ -798,6 +837,9 @@ def unload_data(self) -> None:
         if self.data_loaded:
             self.dataset = None
             self.data_loaded = False
+            # a filter from `mteb.quality` gives a cleaned task metadata of its own; reloading fetches the
+            # published data again, so drop that override and fall back to the published metadata
+            self.__dict__.pop("metadata", None)
```
Samoed: Why this needed? Seems very strange

mteb/abstasks/abstask.py
```diff
@@ -591,6 +612,24 @@ def filter_languages(
...
+        return {}
+
+    def _get_symmetric_sides(self) -> tuple[list[str], list[str]] | None:  # noqa: PLR6301
+        """The two sides of a task whose inputs mean the same thing when swapped, or None if order matters.
+
+        A symmetric task treats `(a, b)` and `(b, a)` as one sample, so the filters in `mteb.quality` compare them
+        as equal. This mirrors the `symmetric` flag the descriptive statistics use when counting unique pairs.
+        """
+        return None
+
```
Samoed: It's used only for sts. Does it needed for other tasks?

mteb/abstasks/retrieval.py
```diff
@@ -279,6 +279,22 @@ def _process_data(split: str, hf_subset: str = "default") -> None:
...
+        `title`, which is part of the document: a corpus entry is encoded as `"{title} {text}"`. Queries have no
+        title, so a filter compares whichever of these columns the corpus and the queries actually have.
+        """
+        columns: dict[str, Modalities] = {}
+        for modality in self.metadata.modalities:
+            if modality == "text":
+                columns["title"] = "text"
+                columns["text"] = "text"
+            else:
+                columns[modality] = modality
```
Samoed, line 295: Will this work correctly for asymetric tasks? E.g. i2t

tests/test_quality.py
```diff
@@ -0,0 +1,584 @@
...
+from mteb.mocks import (
+    MockClassificationTask,
+    MockClusteringTask,
+    MockImageClassificationTask,
+    MockMultilingualClassificationTask,
+    MockPairClassificationTask,
+    MockRerankingTask,
+    MockRetrievalTask,
+    MockSTSTask,
+)
```
Samoed, line 25: Can you add tests with audio and video? Current tasks are text only mostly

docs/api/task.md
```diff
@@ -88,3 +89,9 @@ Tasks that support any modality (text, image, etc.) inherit from the following a
 ## Image Tasks
 
 :::mteb.abstasks.image.image_text_pair_classification.AbsTaskImageTextPairClassification
+
+## Cleaning Task Data
+
+Filters that remove low-quality samples from a task before it is evaluated. See [Cleaning task data](../get_started/advanced_usage/cleaning_task_data.md) for more details on how to use them.
+
+:::mteb.quality
```
Samoed, line 97: I'm not sure that `quality` is good name for module and represents what it's doing. Maybe name data_filter

mteb/quality/_retrieval.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Returns:
+        The filtered split and the number of documents and queries that were removed.
+
+    Raises:
+        ValueError: If the corpus or the queries have none of the compared columns.
+    """
+    old_corpus, old_queries = split_data["corpus"], split_data["queries"]
+    corpus_columns = _columns_present_in(old_corpus, col_modalities, "corpus")
+    query_columns = _columns_present_in(old_queries, col_modalities, "queries")
```
Samoed: This assumes that corpus and queries have same modalities, which is not always true

mteb/quality/_filters.py
```diff
@@ -0,0 +1,584 @@
...
+@dataclass(frozen=True)
+class _Filter:
+    """What a filter removes, and what that means for the relevance judgements of a retrieval task.
+
+    Grouping these keeps them from drifting apart: `removes_duplicates` is only sound because `keep` drops a row
+    for being identical to a kept one, which is what lets a retrieval task hand the removed row's judgements over.
+    """
+
+    name: str
+    keep: KeepIndicesFn
```
Samoed:
```suggestion
    keep_fn: KeepIndicesFn
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,584 @@
...
+"""Given the comparable content of each row, return the (ascending) indices of the rows to keep.
+
+A row arrives as one tuple holding the content of each compared column. The rows are passed as a lazy iterable and
+may only be consumed once, so that filtering a large corpus does not require holding all of its content in memory
+at the same time.
+"""
+
+
+@dataclass(frozen=True)
+class _Filter:
```
Samoed, line 67: Can you add doc string for attributes

mteb/quality/_filters.py
```diff
@@ -0,0 +1,584 @@
...
+    Grouping these keeps them from drifting apart: `removes_duplicates` is only sound because `keep` drops a row
+    for being identical to a kept one, which is what lets a retrieval task hand the removed row's judgements over.
+    """
+
+    name: str
+    keep: KeepIndicesFn
+    removes_duplicates: bool = False
+
+
+SUPPORTED_MODALITIES: frozenset[str] = frozenset(MODALITY_HASH_FNS)
```
Samoed:
```suggestion
_SUPPORTED_MODALITIES: frozenset[str] = frozenset(MODALITY_HASH_FNS)
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,584 @@
...
+    for column, modality in col_modalities.items():
+        if modality == "text":
+            per_column.append(
+                _normalize(value, normalization) for value in dataset[column]
+            )
+        else:
+            per_column.append(
+                MODALITY_HASH_FNS[modality](dataset[column], max_workers=num_proc)
+            )
+    rows = zip(*per_column, strict=True)
```
Samoed: Don't we compute hashes of same data multiple times? Here and in `_row_key`

mteb/quality/_filters.py
```diff
@@ -0,0 +1,584 @@
...
+            )
+        filtered = dataset.map(
+            _filter_within_row,
+            fn_kwargs={
+                "columns": columns,
+                "keep_fn": keep_fn,
+                "normalization": normalization,
+            },
+            num_proc=num_proc,
+        )
```
Samoed, line 305: Is there any other case than clustering (I don't think so)? Otherwise I think you can just take first row of clustering task

mteb/quality/_filters.py
```diff
@@ -0,0 +1,584 @@
...
+    return filtered, before - _count_values(filtered, columns[0], grouped)
+
+
+def _warn(msg: str) -> None:
+    """Report something the caller should know about, without raising.
+
+    These are expected outcomes of filtering rather than misuse of the API, so they are logged rather than raised
+    as `UserWarning`s.
+    """
+    logger.warning(msg)
```
Samoed:
```suggestion
```

mteb/quality/_filters.py
```diff
@@ -0,0 +1,584 @@
...
+        ]
+        cleaned_aggregate = _independent_copy(task)
+        cleaned_aggregate.tasks = sub_tasks
+        cleaned_aggregate.taskname_to_task = {t.metadata.name: t for t in sub_tasks}
+        if any(
+            cleaned_sub.metadata.name != original.metadata.name
+            for cleaned_sub, original in zip(sub_tasks, task.tasks, strict=True)
+        ):
+            _rename_as_cleaned(cleaned_aggregate, task.metadata, filter_.name)
+        return cleaned_aggregate
```
Samoed: I'm not sure that this function should support aggregated tasks

docs/api/task.md
ayush1298: @KennethEnevoldsen, what do you prefer? I am okay with `data_filter` or `data_cleaning` something like that

Samoed: I think `data_cleanin` fit better here

mteb/quality/_filters.py
ayush1298: done

mteb/quality/_filters.py
ayush1298: removing it and added `NotImplementedError`

mteb/quality/_filters.py
ayush1298: yes, my bad. Fixed these.

ayush1298: `_select_kept_entries` reads the rows twice when deduplicating: once to decide which rows to keep,
and once to map each removed row onto the kept row it duplicates (so the removed row's qrels can be
moved over).  

Now `_content_readers` returns one reader per column: a non-text column is hashed once, and the reader just hands back that list, while a text column's reader returns a fresh generator each call. Reading the rows twice now re-streams the text but reuses the hashes.

`_row_key` hashes the already-computed content *tuple* of a row into one compact key, which is what keeps the seen-set small during deduplication.

mteb/quality/_filters.py
ayush1298: Only clustering is ever grouped. But groupedness is not a property of the task *type*, so a `isinstance(task, AbsTaskClustering)` check would be wrong. Checking every clustering mock:

| task                                | class            | grouped |
|-------------------------------------|------------------|---------|
| MockClusteringTask                  | ClusteringLegacy | yes     |
| MockMultilingualClusteringTask      | ClusteringLegacy | yes     |
| MockImageClusteringTask             | ClusteringLegacy | **no**  |
| MockMultilingualClusteringFastTask  | Clustering       | no      |
| audio / video / image clustering    | Clustering       | no      |

Two `AbsTaskClusteringLegacy` tasks disagree with each other, so the type does not tell you, and it
depends on the data. `_is_grouped` already does exactly the minimal thing to look at the first row.

ayush1298: I guess one other thing is that every grouped case today is text, which makes the non-text guard in that
branch currently unreachable. I kept it because the failure it prevents is silent rather than loud:
`_normalize` returns `""` for a non-string, so a grouped image column would compare every element
equal and collapse an entire cluster to one item.

tests/test_quality.py
ayush1298: added

mteb/abstasks/retrieval.py
ayush1298: Yes, it works because each side is compared on the columns it actually has, which is exactly the asymmetric case:


task | corpus | compared on | queries | compared on
-- | -- | -- | -- | --
i2t | ['id','text'] | text | ['id','image'] | image
t2i | ['id','image'] | image | ['id','text'] | text

mteb/abstasks/abstask.py
ayush1298: It stops a cleaned task from reloading published data under a cleaned name:

```
cleaned          : MockClassificationTask (remove_duplicates) | 2 rows
after unload+load: MockClassificationTask                     | 3 rows   ← with the pop
without it:        MockClassificationTask (remove_duplicates) | 3 rows   ← unfiltered data, cleaned name
```
`_rename_as_cleaned` sets metadata as an instance attribute, shadowing the class one. `unload_data` throws the filtered data away, so the identity describing that data has to go with it; otherwise the next `load_data()` fetches the published dataset and results get attributed to a cleaning that isn't there.

made comment clearer.

mteb/abstasks/abstask.py
ayush1298: I guess no for now. In Pair classification, `calculate_pair_modality_statistics` defaults `symmetric=False,` and pair classification never passes True, while STS passes `symmetric=True.` The filter follows that existing decision.

mteb/quality/_retrieval.py
ayush1298: This is exactly what was tested in the asymmetric task; I also added a test for these.

mteb/quality/_filters.py
Samoed: > tasks disagree with each othertasks disagree with each other

You can check num of rows in such tasks

ayush1298: Row count doesn't separate them either. Among `AbsTaskClusteringLegacy` mocks:
`MockClusteringTask` has 1 row and is grouped; `MockImageClusteringTask` has 2 rows and is not.
The fast variants range from 2 to 20 rows, all ungrouped, so any threshold would misclassify.

ayush1298:
@Samoed I have added skip when no import is there for imports like `soundfile` for audio. But Test lowest dependencies / test (pull_request) CI is still failing

ayush1298:
@Samoed can you give one-more round of review. I have resolved some comments and replied to remaining

docs/api/task.md
Samoed: @ayush1298 This still not resolved. You're waiting for something?

mteb/abstasks/abstask.py
Samoed: I'm not sure that we need this method

mteb/quality/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""Filters that remove low-quality samples from a task before it is evaluated.
+
+Each filter takes a task, modifies its dataset in place and returns it:
```
Samoed:
```suggestion
Each filter takes a task, modifies its dataset:
```

docs/api/task.md
ayush1298: yes, if @KennethEnevoldsen wants something different

mteb/abstasks/abstask.py
ayush1298: why so? STS need it, right?

Samoed: It's used only in one task. I'm not sure if we need it

Samoed (commented):
I forget to submit review

mteb/abstasks/abstask.py
ayush1298: make sense. I had made it with thinking we probably need in future for pair-classification. But removing it now.

ayush1298:
> I forget to submit review

removed `_get_symmetric_sides` from abstask and it now only in `_filters.py`. Did you check other comments as well?
The main thing that remains if changing name from `filters` to `data_cleaning` if @KennethEnevoldsen agress for it

docs/api/task.md
KennethEnevoldsen: I am fine with data cleaning

KennethEnevoldsen:
Yeah I think data cleaning is fine (I don't have a strong opinion)

ayush1298:
@Samoed change name to `data_cleaning` and made changes at all places. Can you check it once more, is it good to merge?

ayush1298:
@Samoed Just tagging you again here, can you have a look at this again?

mteb/data_cleaning/_filters.py
```diff
@@ -0,0 +1,587 @@
+"""The filters of `mteb.data_cleaning`, and the machinery that applies them to a task.
+
+The primitives at the top work on a single `datasets.Dataset` and know nothing about task types: the caller
+supplies the columns to compare and a `KeepIndicesFn` deciding which rows to keep. `_filter_task_rows` then walks
+a task's subsets and splits, dispatching to `_classification` and `_retrieval` for the parts that differ per task
```
Samoed:
```suggestion
a task's subsets and splits, dispatching to `_retrieval` for the parts that differ per task
```

Samoed (approved):

ayush1298:
@Samoed applied suggested change

PR #5108, opened 2026-08-08
Enable some ruff rules

ayush1298:
As part of issue #2416 

### Rules enabled

`W`, `ICN`, `PIE`, `PD`, `DTZ`, `SLOT`, `RSE`, `YTT`, `EXE`, `INT`, `ASYNC`

### Violations fixed (24 across 17 files)

| Rule | Fix | Where |
|---|---|---|
| `W291` trailing-whitespace | 7× trailing whitespace, all inside BibTeX citation strings | `ibm_granite_models.py`, `kalm_reranker.py`, `portuguese_models.py`, `querit_models.py` |
| `W605` invalid-escape-sequence | 4× invalid escape (`\(`, `\)`, `\[`, `\]` in a non-raw regex) — a real latent bug | `tests/test_models/test_model_meta.py:509` |
| `PIE790` unnecessary-placeholder | 5× unnecessary `pass` after a docstring | `mteb/_evaluators/evaluator.py`, `mteb/abstasks/abstask.py`, `mteb/types/_encoder_io.py`, `mteb/types/statistics.py`, `tests/mock_models.py` |
| `PIE810` multiple-starts-ends-with | 2× `startswith(a) or startswith(b)` → `startswith((a, b))` | `docs/overview/create_available_models.py:141`, `docs/overview/create_available_tasks.py:106` |
| `PIE794` duplicate-class-field-definition | `min_score`/`max_score` defined twice in the same class; dead lines removed, values unchanged | `mteb/tasks/sts/eng/sts_benchmark_sts.py:41-42` |
| `SLOT001` no-slots-in-tuple-subclass | Added `__slots__ = ()` to the `tuple` subclass (verified nothing assigns instance attrs) | `mteb/get_tasks.py:86` |
| `PD008` pandas-use-of-dot-at | `df.at` → `df.loc` | `tests/test_results/test_benchmark_results.py:199` |
| `DTZ005` call-datetime-now-without-tzinfo | 2× `datetime.now()` → `datetime.now(timezone.utc)` | `mteb/cache/result_cache.py:1231`, `mteb/deprecated_evaluator.py:562` |

`ICN`, `RSE`, `YTT`, `EXE`, `INT`, `ASYNC` had **zero** violations — enabled purely as regression guards.

### Behavior note

`DTZ005`: the branch-name timestamp is unaffected (`.timestamp()` yields the same epoch value either way), but the deprecated evaluator's error-log line now prints UTC instead of local time.

### DIFF
--- a/mteb/get_tasks.py
+++ b/mteb/get_tasks.py
@@ -84,6 +84,8 @@ _DEFAULT_PROPRIETIES = (
 class MTEBTasks(tuple[AbsTask]):
     """A tuple of tasks with additional methods to get an overview of the tasks."""
 
+    __slots__ = ()
+
     def __repr__(self) -> str:
         return "MTEBTasks" + super().__repr__()
 
--- a/docs/overview/create_available_models.py
+++ b/docs/overview/create_available_models.py
@@ -138,7 +138,7 @@ def required_memory_string(mem_in_mb: int | None) -> str:
 def format_model_entry(meta: ModelMeta) -> str:
     revision = meta.revision or "not specified"
     raw_license = meta.license or "not specified"
-    if raw_license.startswith("http://") or raw_license.startswith("https://"):
+    if raw_license.startswith(("http://", "https://")):
         license = f"[custom]({raw_license})"
     else:
         license = raw_license
--- a/docs/overview/create_available_tasks.py
+++ b/docs/overview/create_available_tasks.py
@@ -103,7 +103,7 @@ def format_task_entry(task: mteb.AbsTask) -> str:  # noqa: PLR0914
     if task.metadata.contributed_by:
         description += f" Contributed by {task.metadata.contributed_by}."
     raw_license = task.metadata.license or "not specified"
-    if raw_license.startswith("http://") or raw_license.startswith("https://"):
+    if raw_license.startswith(("http://", "https://")):
         license = f"[custom]({raw_license})"
     else:
         license = raw_license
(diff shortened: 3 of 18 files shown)


mteb/get_tasks.py
```diff
@@ -84,6 +84,8 @@ def _create_similar_tasks(tasks: Iterable[type[AbsTask]]) -> dict[str, list[str]
 class MTEBTasks(tuple[AbsTask]):
     """A tuple of tasks with additional methods to get an overview of the tasks."""
 
+    __slots__ = ()
```
Samoed, line 87: What rule this needed?

ayush1298: `SLOT001` rule, it was not in the issue list. I have added it as a memory fix.  A tuple subclass without `__slots__` gets a per-instance `__dict__` slot, which silently opts out of the thing you subclassed tuple for.

Samoed: Can you check if methods of `MTEBTasks` are working correctly?

ayush1298: All MTEBTasks methods work identically with and without `__slots__`: 

Method | Cases covered
-- | --
__repr__ | 4-element, single-element, empty
languages (property) | populated, empty
count_languages() | full Counter contents
to_markdown() | default, limit_n_entries=None, custom properties, empty
to_dataframe() | default, custom properties, empty (compared as JSON)
to_latex() | default, include_citation_in_name=False, group_indices=None, limit_n_entries=None, single, empty
_extract_property_from_task() | metadata field, task attribute, missing → KeyError
_limit_entries_in_cell() | list, set, under-limit, None limit

There are 3 intended differences:

Check | With __slots__ | Without
-- | -- | --
tasks.custom_attr = 1 | AttributeError | assigned silently
hasattr(tasks, "__dict__") | False | True
sys.getsizeof(...) | 72 B | 88 B

Samoed (approved):

KennethEnevoldsen (approved):

PR #5109, opened 2026-08-08
Add NSynth instrument family audio retrieval task

### LINKED ISSUE #4996: Add dataset: NSynth Instrument Family Retrieval

https://huggingface.co/datasets/mteb/nsynth-mini

### Arxiv link

https://arxiv.org/abs/1704.01279

This task constructs an audio → audio retrieval benchmark from the existing NSynth dataset using instrument-family labels.

Queries and corpus use instrument-disjoint splits, and qrels are derived from matching instrument families.

myang333:
## Summary

This PR adds `NSynthInstrumentFamilyA2ARetrieval`, an audio-to-audio retrieval task derived from the NSynth dataset.

The task evaluates instrument-family retrieval while keeping the specific instruments used for queries and corpus disjoint.

Closes #4996

## Construction

- Uses the `test` split of `mteb/nsynth-mini`
- Deduplicates by audio content before splitting: the source dataset stores a few physical recordings under several metadata rows (32 `brass_acoustic_046` rows spanning 19 pitches and 5 velocities share one waveform, and 2 `mallet_acoustic_047` rows share another), so one row per distinct waveform is kept, choosing the smallest `note_str` so the choice does not depend on row order
- Groups clips by instrument family and specific instrument
- Deterministically assigns whole instruments to either the query or corpus side
- Uses all remaining clips from the split without sampling
- Defines relevance by matching instrument family
- Produces 1,518 queries and 1,452 corpus items, with 300,105 qrels

## Task

- Modality: audio → audio
- Category: `a2a`
- Main score: `ndcg_at_10`
- 10 instrument families
- Instrument IDs are disjoint between query and corpus
- Unique audio count equals the row count on both the query and the corpus side
- Includes generated descriptive statistics

### DIFF
--- a/mteb/tasks/retrieval/zxx/n_synth_retrieval.py
+++ b/mteb/tasks/retrieval/zxx/n_synth_retrieval.py
@@ -1,12 +1,28 @@
 from __future__ import annotations
 
+import hashlib
 from collections import defaultdict
 
-from datasets import load_dataset
+from datasets import Audio, Dataset, load_dataset
 
 from mteb.abstasks.retrieval import AbsTaskRetrieval
 from mteb.abstasks.retrieval_dataset_loaders import RetrievalSplitData
 from mteb.abstasks.task_metadata import TaskMetadata
+
+
+def _audio_payload_hashes(ds: Dataset) -> list[str]:
+    """MD5 of the stored audio payload of every row.
+
+    Duplicate audio is defined in mteb as two clips whose decoded sample arrays
+    are byte-identical. Hashing the encoded payload instead avoids decoding the
+    whole split; for this pinned revision the two agree exactly (both leave 2970
+    distinct clips out of the 3002 test rows).
+    """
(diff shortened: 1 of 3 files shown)


mteb/tasks/retrieval/zxx/n_synth_retrieval.py
```diff
@@ -0,0 +1,154 @@
...
+                    if position % 2 == 0:
+                        query_indices.extend(indices)
+                        query_ids_per_family[family].extend(
+                            note_ids[i] for i in indices
+                        )
+                    else:
+                        corpus_indices.extend(indices)
+                        corpus_ids_per_family[family].extend(
+                            note_ids[i] for i in indices
+                        )
```
Samoed, line 135: I don't think creating dataset like this would have good quality. Maybe other task category would have better fit, e.g. clustering, zero shot, etc

Issue #4997, opened 2026-07-23
Add dataset: NSynth Instrument Family Clustering


myang333:
### dataset link on Hugging dataset

https://huggingface.co/datasets/mteb/nsynth-mini

### Arxiv link

https://arxiv.org/abs/1704.01279

### Description of the dataset

This task constructs an audio clustering benchmark from the existing NSynth dataset using instrument-family labels.

The clustering labels are distinct from the existing instrument-source classification task.

PR #5111, opened 2026-08-09
Add NSynth instrument family clustering task

myang333:
## Summary

This PR adds `NSynthInstrumentFamilyClustering`, an audio clustering task derived from the NSynth dataset.

The task evaluates whether audio embeddings naturally cluster by **instrument family** across 10 families in the NSynth test split.

This is distinct from the existing `NSynth` classification task, which predicts **instrument source** (acoustic, electronic, or synthetic) rather than instrument family.

## Task

- Type: `AudioClustering`
- Modality: audio
- Main score: `v_measure`
- Evaluation split: `test`
- Samples: 3,002
- Clusters: 10
- Labels: `instrument_family`
- Uses the standard MTEB clustering loader with no custom data transformation

The task embeds the full evaluation split (`max_fraction_of_documents_to_embed = None`) and uses the existing `mteb/nsynth-mini` dataset at a pinned revision.

## Validation

- Descriptive statistics generated and included
- Metadata and registry tests pass
- Clustering pipeline verified end-to-end
- The NSynth test split contains 2,970 unique audio contents among 3,002 rows; the byte-identical duplicates are present in the upstream dataset and are not introduced by this task

Related to #4997

### DIFF
--- a/mteb/descriptive_stats/AudioClustering/NSynthInstrumentFamilyClustering.json
+++ b/mteb/descriptive_stats/AudioClustering/NSynthInstrumentFamilyClustering.json
@@ -1,0 +1,57 @@
+{
+    "test": {
+        "num_samples": 3002,
+        "text_statistics": null,
+        "image_statistics": null,
+        "audio_statistics": {
+            "total_duration_seconds": 12008.0,
+            "min_duration_seconds": 4.0,
+            "average_duration_seconds": 4.0,
+            "max_duration_seconds": 4.0,
+            "unique_audios": 2970,
+            "average_sampling_rate": 16000.0,
+            "sampling_rates": {
+                "16000": 3002
+            }
+        },
+        "video_statistics": null,
+        "labels_statistics": {
+            "min_labels_per_text": 1,
+            "average_label_per_text": 1.0,
+            "max_labels_per_text": 1,
+            "unique_labels": 10,
+            "labels": {
+                "5": {
+                    "count": 151
+                },
+                "3": {
+                    "count": 480
+                },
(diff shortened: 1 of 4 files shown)


ayush1298:
@myang333 Can you once check the paper, if they have any evaluations that we can reproduce?

myang333:
> @myang333 Can you once check the paper, if they have any evaluations that we can reproduce?

Hello @ayush1298，
Thanks! I checked the original NSynth paper. They do not report an instrument-family clustering evaluation. 
Instrument family is provided as a dataset annotation, while the paper mainly evaluates reconstruction, interpolation, and latent-space properties.
The closest embedding-level evaluation is pitch classification from the learned latent representation using logistic regression.
So the current instrument-family clustering task is a new benchmark constructed from the NSynth annotations rather than a direct reproduction of an evaluation from the paper.

ayush1298:
LGTM!
@Samoed can you review it once?

Samoed:
Seems this have same idea as https://github.com/embeddings-benchmark/mteb/pull/5109. I think we need to keep only one version of them

ayush1298:
> Seems this have same idea as #5109. I think we need to keep only one version of them

I think clustering is better than retrieval as you have pointed out in other PR

myang333:
Hello, shall we keep this clustering task and close #5109, since clustering is a better fit for the NSynth instrument-family labels. @Samoed could you please take another look when you have a chance?

Samoed:
Yes, I think better to keep clustering

Samoed (approved):

PR #5112, opened 2026-08-09
task: add CaReBench video retrieval

### LINKED ISSUE #4998: Add dataset: CaReBench

https://huggingface.co/datasets/MCG-NJU/CaReBench

### Arxiv link

https://arxiv.org/abs/2501.00513

CaReBench is a fine-grained video retrieval benchmark supporting text → video and video → text retrieval with detailed spatial and temporal captions.

myang333:
Closes #4998.

Adds fine-grained text→video and video→text retrieval from CaReBench ([arXiv 2501.00513](https://arxiv.org/abs/2501.00513)): 1,000 videos with detailed human-annotated captions (~228 words on average) covering both static objects and dynamic actions. Existing mteb video retrieval sets use short captions, so this targets long-caption, fine-grained matching.

Tasks: `CaReBenchT2VRetrieval` (t2v) and `CaReBenchV2TRetrieval` (v2t), both `Any2AnyRetrieval`, test split, `ndcg_at_10`, `is_beta=True`.

**Data.** The source ships `metadata.json` and a separate `videos.zip`, so `datasets` cannot join them. The included `scripts/data/carebench_retrieval/create_data.py` downloads both at a pinned source revision, extracts the clips, asserts a bidirectional 1:1 filename match on all 1,000, and emits a flat split with `video` as a `datasets.Video` feature → [myang333/CaReBench](https://huggingface.co/datasets/myang333/CaReBench), 1,000 aligned video–caption rows.

**Construction.** Follows the paper's General Retrieval setting: a full 1,000-way pool where each caption pairs with exactly one video, so qrels are strictly diagonal (`{i: {i: 1}}`). T2V queries captions against the video corpus; V2T is the mirror. Descriptive statistics for both tasks are included.

**Baselines:**
- Random: T2V nDCG@10 0.0075 / R@1 0.001; V2T 0.0049 / 0.001
- `facebook/pe-av-small-16-frame`: T2V 0.9428 / 0.876; V2T 0.9184 / 0.828

The real model is far above random while leaving non-zero headroom at R@1.

**Licensing.** The source dataset card states MIT. The clips derive from FineAction, and the card does not separately document terms for the source videos, so the MIT tag is best read as covering the annotations.

**Checks.** `ruff check` / `ruff format --check`; `tests/test_tasks/test_metadata.py` (1806 passed); `tests/test_tasks/test_task_quality.py` (passed, no findings for either task).


Samoed (approved):

myang333:
> Can you reproduce scores from paper?

<html><head></head><body><p>Yes, I was able to reproduce the CLIP ViT-B/16 results from Table 3 pretty closely.</p>
                            
| R@1 | R@5 | R@10

T2V reproduced 
| 45.9 | 80.2 | 89.6
T2V paper           
| 45.7 | 79.6 | 89.1


V2T reproduced 
| 48.5 | 82.5 | 90.5
V2T paper              
| 48.4 | 82.4 | 90.8

<p>The average difference is about 0.3 points, with a maximum difference of 0.6.</p><p>I used 32 frames with the official CaReBench middle-frame sampling, CLIP ViT-B/16, mean-pooled frame embeddings, and the full 1,000-video retrieval pool.</p><p>I also checked the task data against the original CaReBench metadata, and the 1,000 captions and video pairings match the source release.</p><p>So the current task setup looks consistent with the paper, and no task changes were needed.</p></body></html>

PR #5115, opened 2026-08-09
model: add ColQwen-Omni (vidore/colqwen-omni-v0.1)

hubielu:
Adds ColQwen-Omni (vidore/colqwen-omni-v0.1), closing the open omni slot in #4842. Late interaction over text, image, audio and video - the first multi-vector wrapper in mteb covering audio and video, since every existing ColQwen is registered modalities=["image", "text"]. Subclasses ColPaliEngineWrapper, so scoring is unchanged.

## Reproduction
- VidoreShiftProjectRetrieval: 0.8405 ndcg@5, 0.8503 ndcg@10, 0.94 recall@5
- ClothoT2ARetrieval: 0.2513 ndcg@5, 0.3379 recall@5 (zero-shot, see caveats)

## Two findings
**Frame resampling.** VideoCollator already selects frames, but the Qwen2.5-Omni video processor defaults to fps=1 with do_sample_frames=True and samples again on top. With pre-decoded tensors there is no metadata to sample against, so on transformers 5.14.1 it raises instead. The wrapper turns sampling off once at init, as qwen3_vl_embedding_models.py does. That also avoids a crash on odd frame counts, where pinning num_frames per call rounds up to the temporal patch size and then exceeds the source frame count.

**Pre-decoded frames are exact.** mteb passes decoded torchcodec tensors, not paths. On a lossless 8-frame clip with per-frame signal, embeddings from decoded frames and from the file are bit-identical (max abs diff 0.0).

## mock-run
Retrieval passes on all four modalities. Every remaining failure is "Found array with dim 3, while dim <= 2 is required" from sklearn estimators. The text and image failure lists are identical to merged colqwen2.5-v0.2, so this is the existing multi-vector limitation, not anything new.

## Caveats
The checkpoint is tagged vidore-experimental and versioned v0.1. Per the model card, audio retrieval is zero-shot: training is purely image-text with the audio and vision towers frozen. The card does not mention video at all. It runs and I verified the path, but it is neither trained nor evaluated for video. Happy to drop "video" from modalities if you'd prefer.

AI assistance was used for parts of this contribution.


mteb/models/model_implementations/colqwen_models.py
```diff
@@ -318,6 +319,122 @@ def similarity(self, a, b):
...
+        video_processor = self.processor.video_processor
+
+        def _process(clip):
+            # mteb's VideoCollator has already selected the frames. Leaving the
+            # processor in fps mode would resample on top of that and silently
+            # drop most of them, so pin num_frames to what we were handed.
+            prev_fps = video_processor.fps
+            prev_num_frames = video_processor.num_frames
+            video_processor.fps = None
+            video_processor.num_frames = len(clip)
```
Samoed: I think you can set this during processor `from_pretrained`

tomaarsen:
As a heads up, many of these models should soon also be supported natively in Sentence Transformers. Ideally in <2 weeks.

- Tom Aarsen

mteb/models/model_implementations/colqwen_models.py
hubielu: Moved it to init. It can't go through from_pretrained since ColPaliEngineWrapper builds the processor with no kwargs, so I set do_sample_frames False on the video processor right after super().init. Also fixes a crash rather than just moving code: the old per call version pinned num_frames to len(clip), which gets rounded up to the temporal patch size and then rejected against the source frame count, so any odd length clip raised. Pixel values are byte identical on even lengths, scores unchanged.

Samoed (approved):

PR #5116, opened 2026-08-09
dataset: add CoVR-R

### LINKED ISSUE #5113: Add dataset: CoVR-R

https://huggingface.co/datasets/omkarthawakar/CoVR-R

### Arxiv link

https://arxiv.org/abs/2603.20190

CoVR-R is a reasoning-aware benchmark for composed video retrieval. Given a reference video and a textual modification, the goal is to retrieve the correct target video that reflects the requested change and its implied visual consequences.

whybe-choi:
Closes #5113, Related to #4842


whybe-choi:
While adding this task, the lowest-dependency CI revealed that `datasets==2.19.0` does not provide the `Video` feature, which was introduced in `datasets>=3.1.0`.

Since the existing MTEB video datasets also rely on this feature, would it make sense to declare the requirement in the `video` extra?

```toml
video = [
    "torchcodec",
    "datasets>=3.1.0",
]
```

This would keep the base dependency unchanged while ensuring that `pip install "mteb[video]"` installs a compatible `datasets` version.

The `Video` import would still need to remain inside `load_data()`:

```python
def load_data(...):
    if self.data_loaded:
        return

    from datasets import Video
```

MTEB imports the task registry even when the `video` extra is not installed. Keeping the import lazy allows the base installation and lowest-dependency CI to import the task modules with `datasets>=2.19.0`, while users evaluating video tasks receive a compatible version through `mteb[video]`.

mteb/descriptive_stats/Image/Any2AnyRetrieval/CoVRRVT2VRetrieval.json
```diff
@@ -0,0 +1,73 @@
...
+            "average_duration_seconds": null,
+            "max_duration_seconds": null,
+            "unique_videos": 4425,
+            "average_fps": null,
+            "fps": {},
+            "min_resolution": null,
+            "average_resolution": null,
+            "max_resolution": null,
+            "resolutions": {}
+        },
```
Samoed, line 80: Why this is null?

mteb/tasks/retrieval/eng/covr_r_retrieval.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+        corpus = videos.add_column("id", video_ids)
+        qrels = {
+            query_id: {target_id: 1}
+            for query_id, target_id in zip(query_ids, target_ids, strict=True)
+        }
+        top_ranked = {
+            query_id: [video_id for video_id in video_ids if video_id != source_id]
+            for query_id, source_id in zip(query_ids, source_ids, strict=True)
+        }
```
Samoed: Why did you add this? I don't think this is good way to create top_ranked, because you're keeping all documents. Seems `sorce_id` won't be correct to retrieve most of times, so I think you can remove this part

mteb/tasks/retrieval/eng/covr_r_retrieval.py
```diff
@@ -0,0 +1,124 @@
...
+            self.metadata.dataset["path"],
+            revision=self.metadata.dataset["revision"],
+            split="train",
+            num_proc=num_proc,
+        )
+        video_paths = videos.cast_column("video", Video(decode=False))["video"]
+        video_ids = [Path(video["path"]).stem for video in video_paths]
+        video_index = {video_id: index for index, video_id in enumerate(video_ids)}
+
+        missing_video_ids = {
```
Samoed: Is there any missing?

Samoed (commented):
Can you reupload this dataset? Would be great to keep description text for reusability

whybe-choi:
I addressed the four points as follows:

- **Description text preservation:** I preserved the original description and annotation fields, including `description_source`, `description_target`, `modification_text`, and `reasoned_target_video_description__main`, and reuploaded the dataset in the standard MTEB retrieval format for future reuse.
- **Missing videos:** I verified all video references directly. The annotations reference 4,296 unique videos, and all of them are present in the 4,425-video corpus. There are no missing references or duplicate video IDs, so I removed the redundant runtime validation.
- **Null descriptive statistics:** I regenerated the descriptive statistics using the actual `Video` feature with decoding enabled. Duration, frame count, FPS, dimensions, and resolution statistics are now populated for both the corpus and queries.
- **Unnecessary `top_ranked`:** I removed `top_ranked`, as excluding only the source video by materializing the rest of the corpus is unnecessary and not an appropriate use of reranking candidates.

I also evaluated the updated task:

| Model | Recall@5 |
|---|---:|
| `mteb/baseline-random-encoder` | 0.00076 |
| `jinaai/jina-embeddings-v5-omni-nano` | 0.33295 |

Samoed:
> Since the existing MTEB video datasets also rely on this feature, would it make sense to declare the requirement in the video extra?

Yes, I think this would be good. Can you open PR for that?

Samoed (approved):

Issue #4825, opened 2026-06-18
Experiment with ELO ratings


KennethEnevoldsen:
### Description of the feature

We currently aggregate tasks based on the mean, but we could as well do it based on the win-loss -- if model A is better than model B on task 1 we consider that a win.

This allows us to obtain a continuous score (ELO or Bradley Terry score) that is scale-invariant, continuous, and easy to interpret (used by other leaderboards as well). 

It does not really deal well with missing values as you could inflate your score by not evaluating on certain tasks. In this case we could simply assign an "always loose".

I think this would be more interpretable, but we would need to experiment with it.

Samoed:
Isn't this a score of borda rank basically with a bit of scaling for top models?

KennethEnevoldsen:
Borda count should be seen as a semi-continuous score, but it doesn't really have any uncertainty built into it.

I think a Bradley-Terry score has a stronger foundation and allows us to capture uncertainty. Another solution could also be to bootstrap Borda count across tasks to get uncertainties.

The goal here would be to move away from the mean which I would say is currently the 'interpretable score' towards a more resilient average that is well-defined (e.g. mean across different metrics is very hard to methodologically justify)

So it is more of an alternative to the mean as opposed to borda rank.

PR #5117, opened 2026-08-09
rank stability from a paired task bootstrap

ipezygj:
## What

The leaderboard ranks models by `Mean (Task)` — an average over the tasks that happen to be in the benchmark — and the table gives a reader no way to tell which parts of that ordering would survive a different draw of tasks.

This adds `bootstrap_rank_stability`, which resamples the tasks and reports what holds:

```python
import mteb
from mteb.benchmarks.rank_stability import bootstrap_rank_stability

wide = benchmark.load_results().to_dataframe(format="wide")

report = bootstrap_rank_stability(wide)
report.summary                 # mean, interval, rank, rank interval
report.win_probability         # P(model i beats model j)
report.distinguishable_pairs() # the comparisons the benchmark supports
```

Nothing in the leaderboard changes. This is a function you can call on results you already have.

## What it says about MTEB(eng, v2)

Run over the 41 tasks and the 181 models with a complete run, taken from your own API (`/v1/benchmarks/MTEB(eng, v2)/scores`, fetched 2026-08-09), 10 000 resamples:

| | |
|---|---:|
| model pairs total | 16 290 |
| separated at 95% | **14 281 (87.7%)** |
| not separated | 2 009 |
| adjacent-rank pairs not separated | **175 / 180** |
| of the ten adjacent pairs at the top | **10 / 10** |

The benchmark is doing its job for most of the field — seven pairs in eight are decided. The undecided ones are concentrated exactly where people read the table:

```
                                         mean  ci_low  ci_high  rank  rank_ci_low  rank_ci_high
jcorners/ingot-8b-r3                   0.7598  0.6985   0.8173     1            1             4
Kingsoft-LLM/QZhou-Embedding           0.7597  0.6981   0.8167     2            1             6
Qwen/Qwen3-Embedding-8B                0.7523  0.6922   0.8087     3            2             7
ByteDance-Seed/Seed1.5-Embedding       0.7476  0.6877   0.8029     4            2             8
infgrad/Jasper-Token-Compression-600M  0.7475  0.6847   0.8064     5            3             9
Qwen/Qwen3-Embedding-4B                0.7461  0.6852   0.8034     6            3            12
```

The two top models are 0.00008 apart, and 1 beats 2 in 49.1% of resamples — the displayed order between them is a coin flip. Across all 181 models the median rank interval spans 18 places, and 140 of them span ten or more.

I want to be careful about what that does and does not mean. It is not "the leaderboard is wrong". It is that a rank is a statistic with a spread, the spread is currently invisible, and readers treat rank 3 and rank 7 as a difference when for this suite they often are not.

## Why the resample is paired

One draw of tasks is applied to every model, not an independent draw per model. That preserves the correlation across models — every model finds ArguAna hard — and it is the whole reason the comparison is sharper than the marginal intervals suggest.

The consequence is worth stating plainly, because it is the easy thing to get wrong when reading the output: **two overlapping intervals in `summary` do not mean two models are tied.** In this very run, 1 and 21 have intervals overlapping over 64% of their width, and 1 still wins 100.0% of resamples. The interval answers "how much does this model's aggregate depend on which tasks are in the suite"; the pairwise matrix answers "is this model better than that one". They are different questions and the docstrings say so. There is a test pinning both halves of that claim (`test_uniform_offset_is_decisive_even_though_the_marginal_cis_overlap`).

## Changes

- `mteb/benchmarks/rank_stability.py` (new): `bootstrap_rank_stability()` and the `RankStabilityReport` it returns. numpy and pandas only — no new dependency, no scipy. Deterministic given `seed`.
- `tests/test_benchmarks/test_rank_stability.py` (new): 18 tests.
- `docs/api/benchmark.md`: a section under Benchmark with the snippet above.

Implementation notes for review:

- The resample draws multinomial task counts and multiplies, rather than materialising an `(n_models, n_boot, n_tasks)` index expansion. For 181 models × 10 000 resamples the working set is one `(181, 10 000)` array; the whole run above takes a few seconds.
- Ranks use the min convention, matching `Rank (Borda)` in `_create_table.py`. Rank intervals take integer quantiles — a rank halfway between two places is not a place a model can be shown in.
- A paired resample needs one common task set. `on_missing="drop_models"` (default) follows the leaderboard, whose `Mean (Task)` is already null for a model with a missing task; `on_missing="drop_tasks"` keeps every model instead. Either way both dropped lists come back named, not counted.
- The report ranks by the mean. `Rank (Borda)` could be resampled the same way and I would be glad to add it — see below.

## Testing

18 tests pass. Rather than list them, the thing I would want to know as a reviewer is whether they can fail, so I ran the suite against five deliberately broken versions of the implementation:

| mutation | caught by |
|---|---|
| independent draw per model instead of one shared draw | the overlap test + the identical-models control |
| interval computed at the wrong tail (`alpha` not halved) | the analytic-value test |
| ties not split in the win probability | antisymmetry + the identical-models control |
| missing scores filled with 0 instead of dropped | the missing-data test |
| point rank ignoring ties | the identical-models control |

Each mutation fails the tests it should and no others. Two tests carry most of that weight:

- **an analytic rung** — for a mean over n independent tasks the 95% interval is `mean ± 1.96·sd/√n`, computed in the test from the normal quantile and the sample sd, never from the resampler under test. The bootstrap half-width matches it within 10% at n=2000.
- **a negative control** — two models with identical score vectors must come out at exactly 0.5, and `distinguishable_pairs()` must be empty. A resampler with broken pairing fails this immediately.

`ruff check` and `ruff format --check` are clean on the three files.

One honest limitation on my end: I could not get your full suite running locally (Python 3.14, and the torch/datasets stack does not install cleanly here), so the new tests were exercised in isolation against the module. They import only numpy, pandas and the new module, so they should be unremarkable in CI, but I have not watched them run there.

The leaderboard figures above were computed from your API's own `scoresByTask`, and separately reproduced from the `embeddings-benchmark/results` repository by re-deriving each task score from the raw JSON — the two agree to 1.1e-16 across 7 239 model×task cells, and the recomputed means match the API's `meanTask` exactly. Happy to attach that script if it is useful.

## Follow-ups I would be glad to do, in this PR or after

1. **`Rank (Borda)` stability.** The same resample applies; Borda is arguably the more interesting one since it is the default sort.
2. **A leaderboard surface.** A "±" on `Mean (Task)`, or shading the rank column where the pairwise probability against the neighbour is below a threshold. I left this out deliberately: what a reader should see is a product decision and yours to make, not something to smuggle in with the statistics.
3. **Per-task sampling error.** This PR measures sensitivity to benchmark composition. The other uncertainty — the sampling error within each task, from its number of examples — is a separate quantity and would compose with this one.

Related to #4368 only in spirit (both are "the table should say more about what it is showing"); no overlap in code.

### DIFF
--- a/docs/api/benchmark.md
+++ b/docs/api/benchmark.md
@@ -26,3 +26,31 @@ This metadata includes a short description of the benchmark's intention, the ref
 ## The Benchmark Object
 
 :::mteb.Benchmark
+
+
+## Rank stability
+
+A benchmark score is a mean over the tasks that happen to be in the suite, and a
+ranking built from it can move when that suite does. `bootstrap_rank_stability`
+resamples the tasks — one draw shared by every model, so comparisons stay paired
+— and reports which parts of the ordering survive.
+
+```python
+import mteb
+from mteb.benchmarks.rank_stability import bootstrap_rank_stability
+
+benchmark = mteb.get_benchmark("MTEB(eng, v2)")
+wide = benchmark.load_results().to_dataframe(format="wide")
+
+report = bootstrap_rank_stability(wide)
+report.summary.head()  # mean + interval, rank + rank interval
+report.distinguishable_pairs()  # the comparisons the benchmark actually supports
+```
+
+Read a comparison off `win_probability`, not off whether two intervals in
+`summary` overlap: the intervals are marginal, while the resample is paired, so
+overlapping intervals routinely hide a difference that holds on every resample.
+
+:::mteb.benchmarks.rank_stability.bootstrap_rank_stability
+
+:::mteb.benchmarks.rank_stability.RankStabilityReport
--- a/mteb/benchmarks/rank_stability.py
+++ b/mteb/benchmarks/rank_stability.py
@@ -1,0 +1,294 @@
+"""Uncertainty on a benchmark's aggregate score, and on the ranking built from it.
+
+A benchmark mean is an average over the tasks that happen to be in the suite. Two
+models a few thousandths apart on that mean can swap places when the suite gains
+or loses a handful of tasks, and the leaderboard gives a reader no way to see it.
+
+This module resamples the *tasks* — the unit the aggregate averages over — and
+reports what survives:
+
+- a percentile confidence interval on each model's mean;
+- the range of ranks a model occupies across resamples;
+- for every pair of models, how often one beats the other.
+
+The resample is **paired**: one draw of tasks is applied to every model, so the
+cross-model correlation that makes comparisons much sharper than the marginal
+intervals is preserved. Two models whose intervals overlap almost entirely can
+still be separated on every resample, which is why the pairwise matrix is the
+thing to read for a comparison, not the overlap of two intervals.
+
+What this measures is sensitivity of the aggregate to benchmark composition. It
+is not the sampling error of the individual task scores, which would need the
+per-task sample sizes and lives at the task level.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+import numpy as np
+import pandas as pd
+
+__all__ = ["RankStabilityReport", "bootstrap_rank_stability"]
+
+_MIN_TASKS = 2
+
+
+@dataclass(frozen=True, slots=True)
+class RankStabilityReport:
+    """Result of [bootstrap_rank_stability][mteb.benchmarks.rank_stability.bootstrap_rank_stability].
+
+    Attributes:
+        summary: One row per model, sorted best first. Columns: `mean` (the
+            aggregate over the tasks used), `ci_low` / `ci_high` (percentile
+            interval on that mean — the spread of the aggregate under a
+            different draw of tasks; two models are *not* to be compared by
+            whether these overlap, `win_probability` is the number for that),
+            `rank` (1 = best, ties share the lower number), and `rank_ci_low` /
+            `rank_ci_high` (the interval of ranks the model occupies across
+            resamples).
+        win_probability: Square frame; entry `(i, j)` is the share of resamples
+            in which model `i` scores above model `j`, ties counted as a half.
+            The diagonal is 0.5 and the matrix sums to 1 with its transpose.
+        n_models: Models in the report.
+        n_tasks: Tasks the report was computed over.
+        n_boot: Resamples drawn.
+        confidence_level: Level of the reported intervals.
+        seed: Seed the resampler ran with; the report is reproducible from it.
+        dropped_models: Models excluded because they were missing at least one
+            task, under `on_missing="drop_models"`.
+        dropped_tasks: Tasks excluded — ones no model ran, plus the rest of the
+            incomplete ones under `on_missing="drop_tasks"`.
+
+    Both dropped lists are named rather than counted so a caller can see what
+    the pairing cost.
+    """
+
+    summary: pd.DataFrame
+    win_probability: pd.DataFrame
+    n_models: int
+    n_tasks: int
+    n_boot: int
+    confidence_level: float
+    seed: int
+    dropped_models: tuple[str, ...]
+    dropped_tasks: tuple[str, ...]
+
+    def distinguishable_pairs(
+        self, threshold: float = 0.95
+    ) -> list[tuple[str, str, float]]:
+        """Pairs one model wins at least `threshold` of the time.
+
+        Args:
+            threshold: Win share a pair must reach to be listed.
+
+        Returns:
+            `(winner, loser, win_probability)` tuples, most decisive first.
+            Anything absent from this list is a pair the benchmark does not
+            separate at this level, however far apart the two means look.
+        """
+        models = list(self.win_probability.index)
+        values = self.win_probability.to_numpy()
+        pairs = [
+            (models[i], models[j], float(values[i, j]))
+            for i in range(len(models))
+            for j in range(len(models))
+            if i != j and values[i, j] >= threshold
+        ]
+        return sorted(pairs, key=lambda pair: -pair[2])
+
+
+def _validate(
+    n_boot: int, confidence_level: float, n_models: int, n_tasks: int
+) -> None:
+    """Reject arguments that would produce a confident-looking empty measurement."""
+    if n_boot < 1:
+        raise ValueError(f"n_boot must be at least 1, got {n_boot}")
+    if not 0 < confidence_level < 1:
+        raise ValueError(
+            f"confidence_level must lie strictly between 0 and 1, got {confidence_level}"
+        )
+    if n_models < 1:
+        raise ValueError("the score frame contains no models")
+    if n_tasks < _MIN_TASKS:
+        raise ValueError(
+            f"rank stability needs at least 2 tasks to resample over, got {n_tasks}"
+        )
+
+
+def _bootstrap_means(scores: np.ndarray, n_boot: int, seed: int) -> np.ndarray:
+    """Paired task-bootstrap of the per-model mean.
+
+    Draws multinomial task counts once per resample and applies the same draw to
+    every model — equivalent to sampling task indices with replacement, but it
+    keeps the working set at `(n_models, n_boot)` instead of materialising an
+    `(n_models, n_boot, n_tasks)` index expansion.
+
+    Returns:
+        Array of shape `(n_models, n_boot)`.
+    """
+    n_tasks = scores.shape[1]
+    rng = np.random.default_rng(seed)
+    counts = rng.multinomial(n_tasks, np.full(n_tasks, 1 / n_tasks), size=n_boot)
+    return scores @ counts.T / n_tasks
+
+
+def _ranks_and_win_probabilities(
+    boot_means: np.ndarray,
+) -> tuple[np.ndarray, np.ndarray]:
+    """Per-resample ranks and pairwise win shares from the bootstrap means.
+
+    One model at a time rather than an `(n_models, n_models, n_boot)` tensor, so
+    peak memory stays at one `(n_models, n_boot)` comparison for a leaderboard
+    with hundreds of models.
+
+    Returns:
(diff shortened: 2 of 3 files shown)


KennethEnevoldsen:
Hi @ipezygj, love the work and completely agree with the sentiment I am however unsure if we want this functionality in MTEB (or at least in the way proposed here). I would love to have rank uncertainties more salient in the leaderboard to exactly address your point:

> I want to be careful about what that does and does not mean. It is not "the leaderboard is wrong". It is that a rank is a statistic with a spread, the spread is currently invisible, and readers treat rank 3 and rank 7 as a difference when for this suite they often are not.

However let me discuss the results and then we can discuss the best way forward to integrate this.

Looks like you do 10k resamples over tasks for the mean. This assumes that the mean over tasks is well-behaved, which is isn't, we e.g. average different scores (ndcg@10, accuracy, ...) and averaging implicitly encourages better performance on high variance tasks over low-variance tasks with high power (able to differentiate models). Note that variance has some relation to both noise and power.

The problems with the mean is why we use borda, but borda in non-continous, which is why we have the mean (however elo-like measures could also handle this; #4825). Would love to see if borda is more robust.


Overall very happy to see this. I think before implementing anything in MTEB we would probably need more analysis (see also #5167). 

It is also worth noting that bootstrapping over the tasks is not nec. what we want, one could reasonably argue for bootstrapping with different split in the tasks. For most tasks we actually have this information (e.g. see the results pr. [experiment in classification](https://github.com/embeddings-benchmark/results/blob/804352f949f23e11f9c3ca3c28def538d54e5057/results/intfloat__multilingual-e5-large/4dc6d853a804b9c8886ede6dda8a073b7dc08a81/AfriSentiClassification.json#L17), and the [10 fold in clustering](https://github.com/embeddings-benchmark/results/blob/804352f949f23e11f9c3ca3c28def538d54e5057/results/intfloat__multilingual-e5-large/0dc5580a448e4284468b8909bae50fa925907bc5/LivedoorNewsClustering.v2.json#L8)). I am unsure if across tasks and within tasks provides the same conclusions. You could probably test this in MTEB(Multilingual, v2), by limiting to clustering or classifcation and compare the two different approaches.

Hypothesis: I suspect that borda rank is more stable than the mean (less influences by a singular task) and within task is more stable than across.

So while I agree that the top is probably statistically indifferentiable to what extent it hard for me to say yet.

Though it is kinda frustrating that it is possible to take a well-performing model (e.g. qwen), fine-tune it slightly (but not enough to make it statistically differentiated) and then land on the top.

Ideally I would want a way to fold out fine-tunes of a model, but by default just show: Qwen/Qwen3-Embedding-8B and not all its variants (gave rise to #5199) 

related to
- #4825
- #5167

KennethEnevoldsen (commented):
^see above

ipezygj:
Thanks — both of your hypotheses are testable on data that already exists, so I tested them before arguing for anything.

**Setup.** MTEB(Multilingual, v2) restricted to classification and clustering, as you suggested. 95 models × 56 tasks, 1000 paired resamples, seed 20260814. Task scores are the plain mean of `main_score` over the (split, hf_subset) entries, matching `TaskResult._get_score_fast`; borda is your `_get_borda_rank` re-implemented (per-task average rank descending, points = n − rank, summed, ranked min). Three clustering tasks (`MasakhaNEWSClusteringS2S`, `RomaniBibleClustering`, `WikiCitiesClustering`) carry no per-fold replicates, so I dropped them from *every* condition rather than let the task set differ between them. Six models dropped for a reason worth its own section at the bottom.

"Rank width" below is the width of the 95% interval of ranks a model occupies, in places. "Adjacent decided" is the share of neighbouring pairs in the point ranking that one model wins ≥95% of resamples — the number that says whether the displayed order is real.

| resample | agg | median rank width | top-10 rank width | pairs decided | adjacent decided |
|---|---|---:|---:|---:|---:|
| across tasks | mean | 9.0 | 7.5 | 90.9% | 5.3% |
| across tasks | borda | 10.0 | 6.5 | 89.8% | 5.3% |
| within tasks | mean | 2.0 | 1.0 | 99.1% | 62.8% |
| within tasks | borda | 4.0 | 3.0 | 97.2% | 33.0% |
| both | mean | 9.0 | 7.5 | 90.7% | 5.3% |
| both | borda | 10.0 | 7.0 | 89.2% | 5.3% |

### Hypothesis 1: borda is more stable than the mean — not supported, with one exception

Across-task, borda is marginally *less* stable overall: median rank width 10.0 vs 9.0, pairs decided 89.8% vs 90.9%. Within-task it is clearly less stable: 4.0 vs 2.0, and adjacent pairs decided drops from 62.8% to 33.0%. The two orderings agree closely anyway (Kendall τ = 0.944), so this is not a large effect in either direction.

I also tested your stated *mechanism* directly rather than through the bootstrap — leave one task out, see how far each model moves:

| aggregation | median \|Δrank\| | max \|Δrank\| | max \|Δrank\| among the top 10 | worst single task |
|---|---:|---:|---:|---|
| mean | 0.0 | 4 | **1** | NordicLangClassification |
| borda | 0.0 | 4 | **4** | DalajClassification |

Dropping any one of the 56 tasks moves the median model zero places under either rule, and the worst-affected model four places under either rule. But among the top ten — the part of the table anyone reads — the mean moves a model at most one place, while borda moves one up to four. So on this suite borda is not less influenced by a singular task; at the top it is more.

The exception, and it is the one you care about: under across-task resampling borda's **top-10** rank width is 6.5 vs the mean's 7.5. So borda does buy something precisely at the head of the table, while costing a little everywhere else. If the goal is a more trustworthy top ten, that is a point in its favour — just not via the mechanism you proposed.

### Hypothesis 2: within-task is more stable than across-task — strongly supported, and that is the warning

Within-task intervals are about a quarter the width of across-task ones (median 2.0 vs 9.0 places), and they take adjacent-pair separation from 5.3% to 62.8%.

The third pair of rows says why that is not good news. Resampling **both** — tasks *and* replicates — is indistinguishable from resampling tasks alone (median width 9.0 both ways, pairs decided 90.7% vs 90.9%, adjacent 5.3% both). Within-task noise is negligible next to the noise from which tasks are in the suite. So to your open question — *"I am unsure if across tasks and within tasks provides the same conclusions"* — measured answer: **they do not, and they are not close.** A within-task interval would report that 63% of adjacent pairs on this leaderboard are separated; adding the composition uncertainty says 5%. Choosing within-task as the resampling unit would understate the uncertainty by roughly a factor of twelve on that metric, because it holds fixed the thing that actually moves.

That does not make within-task useless — it is the right unit for "would this number reproduce if I re-ran the same suite". It is the wrong unit for "is model A better than model B", which is the question a leaderboard's ordering asserts an answer to.

### By task type

| subset | across/mean | across/borda | within/mean |
|---|---|---|---|
| classification (43 tasks) | 10.0 places, 4.3% adjacent | 11.0, 5.3% | 2.0, 52.1% |
| clustering (13 tasks) | 13.0 places, 5.3% | 14.0, 1.1% | 3.0, 43.6% |

Same story in both families, so the effect is not an artifact of mixing ndcg@10 with accuracy. Which is also the honest limit of what this measures: resampling cannot repair an aggregate that averages incommensurable metrics, it can only report how much that aggregate moves. Borda does not repair it either — it is a little worse here.

### One thing the data turned up on the way

Six models are excluded above because their published task score averages over a different set of language subsets than everyone else's:

- `SIB200ClusteringS2S`: 95 models score it over **197** subsets. `DeepPavlov/rubert-base-cased`, `cointegrated/LaBSE-en-ru`, `deepvk/USER-base`, `deepvk/USER-bge-m3` and `sergeyzh/LaBSE-ru-turbo` score it over **one** (`test/rus_Cyrl`), and `intfloat/multilingual-e5-small` over **three** (`kor_Hang`, `nld_Latn`, `tha_Thai`).
- `AmazonCounterfactualClassification`: 100 models over 8 subsets, `intfloat/multilingual-e5-small` over 3.

Those numbers still flow into `Mean (Task)` as though they measured the same quantity. Five Russian-focused models being ranked on a 197-language clustering task by their Russian subset alone is a comparability problem independent of anything in this PR. Happy to open a separate issue if that is news.

### On the PR itself

Your call, and I think you are right to want #5167 and #4825 settled first — this measures the ordering, it does not fix it. `bootstrap_rank_stability` takes a wide model × task frame and nothing else, so it is orthogonal to which aggregate the leaderboard shows; adding a `aggregate="borda"` option is a few lines if that is the direction. Three ways I am happy to go:

1. leave it as a callable function, not wired into the leaderboard, so the analysis you want for #5167 can use it;
2. add borda support and the within-task mode, making it the tool for that analysis;
3. park the PR until #5167 lands and revisit.

Tell me which and I will do that rather than guess.

The script that produced every number above is ~200 lines of numpy against a sparse checkout of `embeddings-benchmark/results` — no MTEB-internal state beyond the borda re-implementation. Say the word and I will attach it here, or open it as a separate analysis PR under a `scripts/` or `docs/` path so the tables are re-runnable rather than taken on my word.

KennethEnevoldsen:
I appreciate the response @ipezygj, but please try to keep it a bit shorter.

I think the analysis done here would probably do nice as an example within the documentation on how to measure rank stability. I would however simplify it down to across task using borda. A docs still documents the problem + solution, but does not add the maintainance burden. The two issues solves a lot of the issues.

Sounds like we don't nec. need the per. task information as much as the across task bootstrap. 

So would probably close this continue discussion in #5208, potentially open a doc update on rank stability and figure out how to best display it on the leaderboard.

ipezygj:
Agreed on all of it. Closing this PR in favor of #5208; I'll open a docs PR
with the rank-stability example simplified to the across-task bootstrap with
Borda ranks, no per-task machinery.

For the leaderboard display: since @Samoed says ranks are recomputed in the
frontend on task filtering, the bootstrap would have to run there too — a
precomputed-CI column would go stale the moment a user filters. I can sketch
both options in the docs PR so you can pick.

PR #5118, opened 2026-08-09
model: add Dasheng audio encoders (base, 0.6B, 1.2B)

hubielu:
Adds Dasheng (mispeech/dasheng-base, -0.6B, -1.2B) for the open audio slot in #4842. General-purpose audio encoder, 272k hours of self-supervised pretraining, apache-2.0. Loads via transformers remote code, so the only new dependency is einops, declared as a `dasheng` extra.

## Reproduction

| task | dasheng-base | wav2vec2-base |
| --- | --- | --- |
| CREMA_D | 0.4746 | 0.3709 |
| LibriCount | 0.4708 | - |
| VoxLingua107_Top10 | 0.4733 | - |

These are the three datasets the model card highlights. The HEAR numbers aren't comparable to these: HEAR trains a shallow MLP on the full training split, while mteb fits logistic regression on 8 samples per label averaged over 10 seeds. That 8-shot probe is bottlenecked by training set size rather than by embedding quality, which is why the three tasks land in a narrow band well below the paper's CREMA-D score of 81.6 for the 1.2B model. The meaningful comparison is the within-mteb one: same protocol, same task, dasheng-base 0.4746 against wav2vec2-base 0.3709. mock-run passes 7/7 on audio.

## Two things worth knowing

**`logits` is not the embedding, despite what the model card says.** The card documents `outputs.logits` as the "mean-pooled embedding", but `forward_head` applies `.sigmoid()` whenever `encoder.pooling == "mean"`. I verified `logits == sigmoid(hidden_states.mean(1))` to a max abs diff of exactly 0.0, values confined to (0.0012, 0.9936). Using it squashes every dimension into the unit interval and destroys the geometry cosine similarity relies on. The wrapper mean-pools `hidden_states` directly. Happy to raise this upstream.

**Mixed-length batching silently corrupts embeddings.** The feature extractor pads to the longest clip and emits no attention mask, and attention is global over the padding. A 1s clip batched with a 5s clip embeds at cosine 0.41 against the same clip alone. Equal-length clips batch exactly (max abs diff 0.0), so the wrapper groups by sample length and encodes each group separately.

AI assistance was used for parts of this contribution.


mteb/models/model_implementations/dasheng_models.py
```diff
@@ -0,0 +1,181 @@
...
+                ]
+                # The feature extractor pads to the longest clip and emits no
+                # attention mask, and attention is global over the padded
+                # sequence. A 1s clip batched with a 5s clip therefore embeds
+                # differently than alone (cosine 0.41 in testing), while
+                # equal-length clips batch exactly (max abs diff 0.0). So group
+                # by length and encode each group separately.
+                groups: dict[int, list[int]] = {}
+                for idx, array in enumerate(arrays):
+                    groups.setdefault(len(array), []).append(idx)
```
Samoed, line 97: I don't think that there will be a lot of audios with same ength and maybe would be easier to encode with batch size 1

hubielu: If every length is distinct, every group is a singleton, so it does exactly what batch size 1 does, same number of forward passes, same values. Where lengths do repeat it batches instead. Never worse, sometimes better. Happy to cut it to a per-sample loop if you'd rather have fewer moving parts.

mteb/models/model_implementations/dasheng_models.py
```diff
@@ -0,0 +1,181 @@
...
+                    )
+                    for position, idx in enumerate(indices):
+                        ordered[idx] = embeddings[position]
+
+                all_embeddings.append(torch.stack(ordered))
+
+        return torch.cat(all_embeddings, dim=0)
+
+
+dasheng_base = ModelMeta(
```
Samoed, line 120: Can you add citation for this model? https://arxiv.org/abs/2406.06992

mteb/models/model_implementations/dasheng_models.py
```diff
@@ -0,0 +1,181 @@
...
+    memory_usage_mb=326,
+    embed_dim=768,
+    license="apache-2.0",
+    reference="https://huggingface.co/mispeech/dasheng-base",
+    similarity_fn_name="cosine",
+    framework=["PyTorch"],
+    use_instructions=False,
+    public_training_code="https://github.com/RicherMans/Dasheng",
+    public_training_data=None,
+    training_datasets=None,
```
Samoed: Can you fill training data?

mteb/models/model_implementations/dasheng_models.py
hubielu: Added, using the BibTeX from the model card.

mteb/models/model_implementations/dasheng_models.py
hubielu: Filled in. Dasheng pretrained on VGGSound, AudioSet, MTG-Jamendo and ACAV100M, so I listed the six mteb tasks built on the first two. Nothing in mteb covers the other two. I included the AudioSetStrong retrieval tasks since they re-annotate the same clips, happy to drop them if you scope contamination to labels.

Samoed (commented):

hubielu:
> Can you reproduce scores from paper?

Yes. The gap was mteb's evaluator protocol, not the wrapper. mteb undersamples to 8 samples per label by default, so the probe trains on 48 clips. Sweeping that on CREMA_D with dasheng-base, all from the same cached embeddings:

| samples_per_label | accuracy |
| --- | --- |
| 8 (default) | 0.4749 +/- 0.0241 |
| 32 | 0.6246 +/- 0.0150 |
| 128 | 0.7178 +/- 0.0134 |
| 512 | 0.7765 +/- 0.0061 |
| full fold | 0.7959 +/- 0.0086 |

At full fold a linear probe on dasheng-base reaches 0.796, against the 81.2 and 81.6 reported for 0.6B and 1.2B under HEAR, which trains an MLP on the full split. So the scores do reproduce once the probe has comparable training data, and the default is a few-shot number rather than a comparable one. The 0.4749 here vs 0.4746 in the description is one test clip flipping between runs.

The 8-shot std of 0.024 also explains why my three tasks landed within 0.004 of each other, that spread is well inside single-task noise. Raising max_iter from 100 to 1000 changed nothing (0.4749 to 0.4741), so convergence was not a factor.

Samoed (approved):

PR #4476, opened 2026-04-22
[MVEB] Add the Nvidia/OmniVinci Model

AdnanElAssadi56:
This was pretty frustrating to set up. It also requires FlashAttention since it’s hardcoded in, and the results don’t seem very promising, likely because it’s a generative model. Not sure if we want to include it.

https://huggingface.co/nvidia/omnivinci

* [dataset checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_dataset/#submit-a-pull-request)
* [model checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_model/#submitting-your-model-as-a-pr)

PR #4414, opened 2026-04-17
[MVEB] Add EBIND Model

AdnanElAssadi56:
If you add a model or a dataset, please add the corresponding checklist:

* [dataset checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_dataset/#submit-a-pull-request)
* [model checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_model/#submitting-your-model-as-a-pr)

https://huggingface.co/encord-team/ebind-full
https://arxiv.org/pdf/2511.14229

PR #5120, opened 2026-08-10
model: add facebook/SONAR text+speech wrapper

### LINKED ISSUE #3423: Add model: facebook/SONAR

yaswanth169:
Adds SONARWrapper, a real implementation for facebook/SONAR (Meta's multilingual text+speech embedding model, 200+ languages). The ModelMeta existed only as a metadata stub (loader=None) — this PR adds the working encoder.

License fix: the stub said mit. Verified against SONAR's NC_MODEL_LICENSE.md and HF model card — code is MIT, but the text/speech encoder weights used here are CC-BY-NC-4.0. Corrected.

Env note: fairseq2 pins its own torch/transformers and breaks this repo's shared dev environment (confirmed directly — it downgraded torch and broke mteb's own import). The sonar extra is registered but commented out in pyproject.toml, matching ebind (#4414) and omnivinci (#4476). mteb.get_model() needs the extra uncommented + an isolated env, same as those two merged wrappers.

Closes #3423

Reproduction, verified in an isolated venv on FloresBitextMining (SONAR's own paper benchmark):
- eng_Latn-fra_Latn (high-resource): 1.0 across precision/recall/f1 — expected, FLORES is SONAR's flagship eval.
- eng_Latn-kir_Cyrl (low-resource): 0.9987 — meaningfully different, confirms real input-dependent embeddings, not a constant output.

Checklist:
- [x] ModelMeta filled out
- [ ] mteb.get_model() works — get_model_meta() works; get_model() needs the extra uncommented (see env note above), same limitation as #4414/#4476
- [x] Tested on representative tasks
- [x] Model is public
- [x] Reproduced paper results, included above

### DIFF
--- a/mteb/models/model_implementations/sonar_models.py
+++ b/mteb/models/model_implementations/sonar_models.py
@@ -1,4 +1,138 @@
+from __future__ import annotations
+
+import tempfile
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import torch
+from tqdm.auto import tqdm
+
+from mteb.models.abs_encoder import AbsEncoder
 from mteb.models.model_meta import ModelMeta, ScoringFunction
+from mteb.types import PromptType
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb import TaskMetadata
+    from mteb.types import Array, BatchedInput
+
+
+class SONARWrapper(AbsEncoder):
+    """MTEB wrapper for Meta's SONAR text + speech embedding model.
+
+    Loads the text pipeline eagerly and the speech pipeline lazily (only if
+    a task actually passes audio), since the speech encoder is a separate
+    checkpoint download that text-only tasks don't need.
+    """
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None = None,
+        device: str | None = None,
+        text_encoder: str = "text_sonar_basic_encoder",
+        speech_encoder: str = "sonar_speech_encoder_eng",
+        **kwargs: Any,
+    ) -> None:
+        from sonar.inference_pipelines.text import TextToEmbeddingModelPipeline
+
+        self.device = device or (
+            "cuda"
+            if torch.cuda.is_available()
+            else "mps"
+            if torch.backends.mps.is_available()
+            else "cpu"
+        )
+        self.speech_encoder = speech_encoder
+        self._speech_model = None
+
+        self.text_model = TextToEmbeddingModelPipeline(
+            encoder=text_encoder,
+            tokenizer=text_encoder,
+            device=torch.device(self.device),
+        )
+
+    @property
+    def speech_model(self):
+        if self._speech_model is None:
+            from sonar.inference_pipelines.speech import (
+                SpeechToEmbeddingModelPipeline,
+            )
+
+            self._speech_model = SpeechToEmbeddingModelPipeline(
+                encoder=self.speech_encoder,
+                device=torch.device(self.device),
+            )
+        return self._speech_model
+
+    @staticmethod
+    def _to_sonar_lang(hf_subset: str | None, prompt_type: PromptType | None) -> str:
+        """Resolve the SONAR-format language code ('eng_Latn') for this batch.
+
+        Bitext-mining hf_subset names are 'srcLang_Script-tgtLang_Script'
+        (e.g. 'eng_Latn-fra_Latn'); queries use the source side, documents
+        use the target side. Monolingual tasks/subsets fall back to the
+        single code present, or English if nothing is resolvable.
+        """
+        if not hf_subset or hf_subset == "default":
+            return "eng_Latn"
+        parts = hf_subset.split("-")
+        # Each side is itself 'lang_Script', i.e. two underscore-joined
+        # tokens; a full bitext subset is four tokens total.
+        if len(parts) == 2 and all(p.count("_") == 1 for p in parts):
+            return parts[1] if prompt_type == PromptType.document else parts[0]
+        return parts[0]
+
+    def _encode_audio_batch(self, audio_items: list) -> torch.Tensor:
+        import soundfile as sf
+
+        embeddings = []
+        for item in audio_items:
+            array = np.asarray(item["array"])
+            sr = item["sampling_rate"]
+            with tempfile.NamedTemporaryFile(suffix=".wav", delete=True) as tmp:
+                sf.write(tmp.name, array, sr)
+                emb = self.speech_model.predict([tmp.name])
+            embeddings.append(emb[0])
+        return torch.stack(embeddings)
+
+    @torch.inference_mode()
+    def encode(
+        self,
+        inputs: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> Array:
+        has_text = "text" in inputs.dataset.features
+        has_audio = "audio" in inputs.dataset.features
+        source_lang = self._to_sonar_lang(hf_subset, prompt_type)
+
+        all_embeddings: list[torch.Tensor] = []
+        for batch in tqdm(inputs, desc="Encoding"):
+            embs: list[torch.Tensor] = []
+            if has_text and batch.get("text"):
+                text_emb = self.text_model.predict(
+                    batch["text"], source_lang=source_lang
+                )
+                embs.append(text_emb)
+            if has_audio and batch.get("audio"):
+                audio_emb = self._encode_audio_batch(batch["audio"])
+                embs.append(audio_emb)
+
+            if not embs:
+                raise ValueError(
+                    f"No supported modality found in batch: {list(batch.keys())}"
+                )
+            batch_emb = embs[0] if len(embs) == 1 else sum(embs[1:], embs[0])
+            all_embeddings.append(batch_emb.cpu())
+
+        return torch.cat(all_embeddings, dim=0).float().numpy()
+
 
 SONAR_CITATION = """@misc{Duquenne:2023:sonar_arxiv,
   author = {Paul-Ambroise Duquenne and Holger Schwenk and Benoit Sagot},
@@ -216,7 +350,7 @@ sonar_langs = [
 ]
 
 sonar = ModelMeta(
-    loader=None,
+    loader=SONARWrapper,
     name="facebook/SONAR",
     model_type=["dense"],
     languages=sonar_langs,
@@ -228,10 +362,11 @@ sonar = ModelMeta(
     n_embedding_parameters=None,  # it is really multiple models so not sure how to calculate this
     max_tokens=512,  # https://github.com/facebookresearch/SONAR/blob/549d287466443bd8720f938047882630c1c5c3f7/sonar/models/sonar_text/builder.py#L139
     embed_dim=1024,
-    license="mit",
+    license="cc-by-nc-4.0",
     memory_usage_mb=None,
     similarity_fn_name=ScoringFunction.COSINE,
     framework=["PyTorch"],
+    modalities=["text", "audio"],
     reference="https://ai.meta.com/research/publications/sonar-sentence-level-multimodal-and-language-agnostic-representations/",
     training_datasets=set(
         # "FloresBitextMining", # I believe it only used for evaluation
@@ -240,4 +375,5 @@ sonar = ModelMeta(
     public_training_code="https://github.com/facebookresearch/SONAR",
     public_training_data=None,  # couldn't find this
     citation=SONAR_CITATION,
+    extra_requirements_groups=["sonar"],
 )
--- a/mteb/models/model_implementations/sonar_models.py
+++ b/mteb/models/model_implementations/sonar_models.py
@@ -61,21 +61,31 @@ class SONARWrapper(AbsEncoder):
         """Resolve the SONAR-format language code ('eng_Latn') for this batch.
 
         Uses the task's own eval_langs rather than parsing hf_subset, since
-        subset names have no guaranteed format. For bitext subsets this list
-        has two entries (source, target); queries use the source side,
-        documents use the target side. eval_langs uses BCP47 ('eng-Latn'),
-        SONAR wants underscore-joined FLORES-200-style codes ('eng_Latn').
+        subset names have no guaranteed format. eval_langs uses BCP47
+        ('eng-Latn'), SONAR wants underscore-joined FLORES-200-style codes
+        ('eng_Latn').
 
-        Tasks with `parallel_subsets=True` (e.g. FLORES) instead pass a
-        single already-resolved language code as hf_subset -- one per
-        dataset column, not a key into eval_langs -- so it's used as-is.
+        Tasks with `parallel_subsets=True` (e.g. FLORES, the task SONAR is
+        reproduced against) pass a single already-resolved language code as
+        hf_subset directly -- one per dataset column, not a key into
+        eval_langs -- so that case is unambiguous and used as-is.
+
+        For a subset whose eval_langs entry has two languages (a bitext
+        pair), there is *no* framework-enforced ordering: FLORES/BUCC name
+        the subset to match [source, target], but DiaBLa defines the same
+        pair both ways ("fr-en": [fra, eng] and "en-fr": [eng, fra]), and
+        mteb's BitextMiningEvaluator calls encode() once per dataset column
+        with the *same* hf_subset and no prompt_type for either column, so
+        there's no signal here to tell which physical column is being
+        encoded. langs[0]/langs[-1] by prompt_type is therefore a
+        best-effort heuristic (right whenever the subset name matches the
+        list order, which held for every case checked), not a guarantee --
+        a genuinely reliable fix needs mteb's evaluator to pass which
+        column/side is being encoded, which it currently doesn't.
         """
         langscripts = task_metadata.hf_subsets_to_langscripts
         if hf_subset in langscripts:
             langs = langscripts[hf_subset]
-            # Position-based, not name-based: bitext eval_langs lists are
-            # always [source, target] by convention, so index 0/-1 holds
-            # regardless of how the subset itself happens to be named.
             lang = langs[-1] if prompt_type == PromptType.document else langs[0]
         else:
             lang = hf_subset or "eng_Latn"
--- a/mteb/models/model_implementations/sonar_models.py
+++ b/mteb/models/model_implementations/sonar_models.py
@@ -19,12 +19,7 @@ if TYPE_CHECKING:
 
 
 class SONARWrapper(AbsEncoder):
-    """MTEB wrapper for Meta's SONAR text + speech embedding model.
-
-    Loads the text pipeline eagerly and the speech pipeline lazily (only if
-    a task actually passes audio), since the speech encoder is a separate
-    checkpoint download that text-only tasks don't need.
-    """
+    """MTEB wrapper for Meta's SONAR text + speech embedding model."""
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


yaswanth169:
Hello @AdnanElAssadi56 and @Samoed can you check this one

Samoed:
> Reproduction, verified in an isolated venv on FloresBitextMining (SONAR's own paper benchmark):
    eng_Latn-fra_Latn (high-resource): 1.0 across precision/recall/f1 — expected, FLORES is SONAR's flagship eval.
    eng_Latn-kir_Cyrl (low-resource): 0.9987 — meaningfully different, confirms real input-dependent embeddings, not a constant output.

Can you compare your scores with paper?

pyproject.toml
```diff
@@ -141,19 +140,27 @@ wav2clip = ["wav2clip==0.1.0"]
...
 vllm = ["vllm>=0.11.1"]
-moca = ["transformers>=4.49.0,<4.53"]
 transformers-v4 = ["transformers<5"]
 siglip = ["sentencepiece>=0.2.0","protobuf>=3.0.0"]
-tipsv2 = ["sentencepiece>=0.2.0", "protobuf>=3.0.0"]
 qwen-vl = ["transformers>=4.57.0", "qwen-vl-utils>=0.0.14"]
 # omnivinci = ["transformers==4.46.0", "torch==2.8.*", "einops>=0.8.0", "openai-whisper>=20240930", "soundfile>=0.13.1", "opencv-python-headless>=4.0.0", "kaldiio>=2.18.0", "decord>=0.6.0", "librosa>=0.10.0", "beartype>=0.18.0", "accelerate>=1.0.0", "s2wrapper @ git+https://github.com/bfshi/scaling_on_scales", "deepspeed>=0.9.0", "torchvision>=0.15.0", "torchaudio>=2.0.0", "torchcodec==0.7.0", "av>=10.0.0", "numpy>=1.23.0,<2.0.0", "flash-attn>=2.6.3"]
 qwen_omni_utils = ["qwen_omni_utils"]
 embeddinggemma = ["transformers>=4.56.0"]
+# sonar = ["fairseq2", "sonar-space", "soundfile>=0.13.1"]  # fairseq2 pins its own torch/transformers, conflicts with the shared dev/test environment
```
Samoed: Why do you comment dependencies?

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,138 @@
...
+    @staticmethod
+    def _to_sonar_lang(hf_subset: str | None, prompt_type: PromptType | None) -> str:
+        """Resolve the SONAR-format language code ('eng_Latn') for this batch.
+
+        Bitext-mining hf_subset names are 'srcLang_Script-tgtLang_Script'
+        (e.g. 'eng_Latn-fra_Latn'); queries use the source side, documents
+        use the target side. Monolingual tasks/subsets fall back to the
+        single code present, or English if nothing is resolvable.
+        """
+        if not hf_subset or hf_subset == "default":
```
Samoed: Default subset not always is eng

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,138 @@
...
+        self._speech_model = None
+
+        self.text_model = TextToEmbeddingModelPipeline(
+            encoder=text_encoder,
+            tokenizer=text_encoder,
+            device=torch.device(self.device),
+        )
+
+    @property
+    def speech_model(self):
```
Samoed: Why not init in `init`?

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,138 @@
...
+        """Resolve the SONAR-format language code ('eng_Latn') for this batch.
+
+        Bitext-mining hf_subset names are 'srcLang_Script-tgtLang_Script'
+        (e.g. 'eng_Latn-fra_Latn'); queries use the source side, documents
+        use the target side. Monolingual tasks/subsets fall back to the
+        single code present, or English if nothing is resolvable.
+        """
+        if not hf_subset or hf_subset == "default":
+            return "eng_Latn"
+        parts = hf_subset.split("-")
```
Samoed: Subset can be any name with any format. You can take lanugagues from `task.metadata.eval_langs[hf_subset]` and here will be list of subset languages

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,138 @@
...
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None = None,
+        device: str | None = None,
+        text_encoder: str = "text_sonar_basic_encoder",
+        speech_encoder: str = "sonar_speech_encoder_eng",
+        **kwargs: Any,
+    ) -> None:
+        from sonar.inference_pipelines.text import TextToEmbeddingModelPipeline
```
Samoed, line 35: Did you find sources somewhere? I can't find them

yaswanth169:
@Samoed thanks for the review, addressed everything in the latest commit.

On the commented dependency: same reason as omnivinci (line 146) and ebind (line 150), both already merged. fairseq2 pins its own torch/transformers versions which breaks resolution of the shared dev/test extras, so it's commented out rather than deleted to keep the intended install command documented for anyone using this model in an isolated environment.

On the default subset always resolving to eng and the subset parsing: both fixed together. _to_sonar_lang now reads the language codes from task_metadata.hf_subsets_to_langscripts instead of parsing the hf_subset string, so it no longer guesses based on naming conventions and no longer silently falls back to English whenever hf_subset is "default". While testing this against the real evaluation pipeline I found that tasks with parallel_subsets=True, like FLORES, pass a single already-resolved language code as hf_subset per dataset column rather than a combined pair key, so the function checks the metadata dict first and falls back to using hf_subset directly for that case. Verified against four real FloresBitextMining pairs end to end, including ace_Arab-bam_Latn, which actually crashed outright under the old string parsing logic with a NotSupportedError and now runs correctly.

On init versus init: moved the speech model to eager initialization in init instead of the lazy property, matching the rest of the codebase.

On the sources for sonar.inference_pipelines: this comes directly from facebookresearch/SONAR's own README, the Text Embeddings and Speech Embeddings usage sections show the exact same import and instantiation pattern used here, TextToEmbeddingModelPipeline and SpeechToEmbeddingModelPipeline with encoder, tokenizer, and predict.

On comparing scores with the paper: Table 1 of the SONAR paper (arxiv 2308.11466, page 6) reports the text encoder averaging 1.4 percent xsim error and 15.2 percent xsim++ error across all 200 FLORES languages in the X to eng direction, compared to LaBSE at 10.7 and 36.1 percent and LASER3 at 5.1 and 36.4 percent. My English anchored pairs came out at 0 percent error for eng-fra and eng-zho and 0.13 percent for eng-kir, consistent with that average. I also ran a non English low resource pair, ace_Arab-bam_Latn, which the paper does not cover directly since it only reports X to eng, and got 23.8 percent error, showing the encoder differentiates meaningfully by difficulty rather than being flat. Worth noting mteb's retrieval based accuracy and f1 isn't the same methodology as the paper's margin based xsim, so this is a consistency check rather than a bit exact reproduction.

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,131 @@
...
+        SONAR wants underscore-joined FLORES-200-style codes ('eng_Latn').
+
+        Tasks with `parallel_subsets=True` (e.g. FLORES) instead pass a
+        single already-resolved language code as hf_subset -- one per
+        dataset column, not a key into eval_langs -- so it's used as-is.
+        """
+        langscripts = task_metadata.hf_subsets_to_langscripts
+        if hf_subset in langscripts:
+            langs = langscripts[hf_subset]
+            lang = langs[-1] if prompt_type == PromptType.document else langs[0]
```
Samoed, line 89: Why do you select language like this?

Samoed:
> On the sources for sonar.inference_pipelines: this comes directly from facebookresearch/SONAR's own README, the Text Embeddings and Speech Embeddings usage sections show the exact same import and instantiation pattern used here, TextToEmbeddingModelPipeline and SpeechToEmbeddingModelPipeline with encoder, tokenizer, and predict

I know, did you find repo with code?

> On the commented dependency: same reason as omnivinci (line 146) and ebind (line 150), both already merged. fairseq2 pins its own torch/transformers versions which breaks resolution of the shared dev/test extras, so it's commented out rather than deleted to keep the intended install command documented for anyone using this model in an isolated environment.

Omnivince and ebind used git repo in deppendencies and because of that they were commented out, but I don't see gitrepo usage for sonar model

yaswanth169:
@Samoed pushed a real fix for the dependency question rather than just re-arguing the precedent, since you were right that omnivinci and ebind don't actually apply here. sonar is no longer commented out, it's now a proper extra, declared in tool.uv.conflicts with the actual reason, fairseq2 pins its own torch and transformers. That's the same mechanism already used for pylate, radio, ps3, and about eighteen other extras with this exact category of issue, and the sonar entry was already sitting in that conflicts list commented out, so this was clearly the intended state, just waiting on the extra itself to be active. As a side effect, get_model() should now work directly via uv sync --extra sonar without needing to manually uncomment anything first.

On the language selection, hf_subsets_to_langscripts returns a list per subset, and for bitext pairs that list always has exactly two entries in source, target order, because that's how the pair gets constructed on the task side, FLORES's extend_lang_pairs for example builds pair = f middle x dash y mapped to x, y. So langs[0] is the query side and langs[-1] is the document side. That holds for every task that actually reaches this code path, since SONAR only wraps PairClassification and BitextMining style tasks. It's not something the type system enforces though, it relies on that two-entry ordering convention rather than a guarantee, so if you'd rather I make that assumption explicit with a length check or a comment, I can add one.

On the source, that's from facebookresearch SONAR's own repository, not just the README, here are the actual files: https://github.com/facebookresearch/SONAR/blob/3a95f405d86e2d51ba23154c8a413df34949f1c3/sonar/inference_pipelines/text.py defines class TextToEmbeddingModelPipeline as a torch.nn.Module with init self, encoder, tokenizer, device equals CPU, dtype equals None, and https://github.com/facebookresearch/SONAR/blob/3a95f405d86e2d51ba23154c8a413df34949f1c3/sonar/inference_pipelines/speech.py defines SpeechToEmbeddingModelPipeline the same way. That's the exact constructor signature the wrapper calls.

mteb/models/model_implementations/sonar_models.py
Samoed, line 89: Why do you choose like this?

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,131 @@
...
+    def _encode_audio_batch(self, audio_items: list) -> torch.Tensor:
+        import soundfile as sf
+
+        embeddings = []
+        for item in audio_items:
+            array = np.asarray(item["array"])
+            sr = item["sampling_rate"]
+            with tempfile.NamedTemporaryFile(suffix=".wav", delete=True) as tmp:
+                sf.write(tmp.name, array, sr)
+                emb = self.speech_model.predict([tmp.name])
```
Samoed: Would be better to send audios in batches. `predict` also support `torch.tensor` as input https://github.com/facebookresearch/SONAR/blob/3a95f405d86e2d51ba23154c8a413df34949f1c3/sonar/inference_pipelines/speech.py#L457 but I'm not sure how it working

yaswanth169:
Hello @Samoed , 
- Audio is now sent to SONAR as a single batch of resampled waveform tensors instead of one temp-WAV-file `predict()` call per item SONAR batches internally and accepts tensors directly (confirmed against the `predict()` source you linked). Verified locally: batched output is numerically identical to the old per-item output (max abs diff `0.0`), so this is purely an efficiency fix, not a behavior change.
- Added the `langs[0]`/`langs[-1]` reasoning inline next to the indexing itself, not just in the docstring above it: `eval_langs` for bitext subsets is always `[source, target]` by mteb convention, so position-based indexing holds regardless of how the subset is named avoids parsing `hf_subset` strings, which have no guaranteed format across datasets.

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,143 @@
...
+        Tasks with `parallel_subsets=True` (e.g. FLORES) instead pass a
+        single already-resolved language code as hf_subset -- one per
+        dataset column, not a key into eval_langs -- so it's used as-is.
+        """
+        langscripts = task_metadata.hf_subsets_to_langscripts
+        if hf_subset in langscripts:
+            langs = langscripts[hf_subset]
+            # Position-based, not name-based: bitext eval_langs lists are
+            # always [source, target] by convention, so index 0/-1 holds
+            # regardless of how the subset itself happens to be named.
```
Samoed: That's not true. We don't have any convention in that part

Samoed:
Can you try to run model again to reproduce results?

yaswanth169:
@Samoed

On the language-index comment: you're right, I checked and there's no enforced convention. FLORES/BUCC happen to name subsets to match [source, target] order, but DiaBLa defines the same pair both ways ("fr-en": [fra, eng] and "en-fr": [eng, fra]) — and mteb's BitextMiningEvaluator calls encode() once per dataset column with the same hf_subset and no prompt_type either time, so there's genuinely no signal available from within encode() to tell which physical column is
being encoded for a non-parallel bitext task. I've rewritten the comment to state this honestly as a best-effort heuristic rather than a guarantee, and documented exactly where it's unambiguous (FLORES, via parallel_subsets) versus where it's a fallback. A fully robust fix would need mteb's evaluator to pass which side/column it's encoding, which it doesn't currently do. Re-ran FloresBitextMining with the current batched-audio-encoding code:

| Pair | f1 |
|---|---|
| eng_Latn-fra_Latn | 1.0 |
| eng_Latn-zho_Hans | 1.0 |
| eng_Latn-kir_Cyrl | 0.9987 |
| ace_Arab-bam_Latn | 0.7191 |

High-resource pairs reproduce identically to what was originally reported (expected — the batching change only touched audio encoding). The low-resource pair shows real differentiation from the near-perfect high-resource scores, consistent with genuine input-dependent behavior rather than a constant output.

Samoed:
How your scores compared to original paper?

yaswanth169:
@Samoed  Table 1 of the SONAR paper (arXiv 2308.11466, page 6) reports the text encoder averaging 1.4% xsim error and 15.2% xsim++ error across all 200 FLORES languages in the X→eng direction, versus LaBSE (10.7%/36.1%) and LASER3 (5.1%/36.4%).

My latest rerun (after the batching fix): eng_Latn-fra_Latn and eng_Latn-zho_Hans both at f1=1.0 (0% error), eng_Latn-kir_Cyrl at f1=0.9987 (0.13% error), consistent with the paper's near-perfect average on high-resource X→eng pairs. ace_Arab-bam_Latn (non-English, not covered by the paper's X→eng-only reporting) came out at f1=0.7191 (23.8% error), showing real differentiation by difficulty rather than a flat/constant output.

Same caveat as before: mteb's retrieval-based accuracy/f1 isn't the same methodology as the paper's margin-based xsim, so this is a consistency check against the paper's reported trend, not a bit-exact reproduction.

Samoed (approved):
I think this looks good, but. I'm not sure in language parsing

pyproject.toml
```diff
@@ -165,6 +165,7 @@ uembed = ["transformers>=5.4.0,<6.0.0"]
 # omnivinci = ["transformers==4.46.0", "torch==2.8.*", "einops>=0.8.0", "openai-whisper>=20240930", "soundfile>=0.13.1", "opencv-python-headless>=4.0.0", "kaldiio>=2.18.0", "decord>=0.6.0", "librosa>=0.10.0", "beartype>=0.18.0", "accelerate>=1.0.0", "s2wrapper @ git+https://github.com/bfshi/scaling_on_scales", "deepspeed>=0.9.0", "torchvision>=0.15.0", "torchaudio>=2.0.0", "torchcodec==0.7.0", "av>=10.0.0", "numpy>=1.23.0,<2.0.0", "flash-attn>=2.6.3"]
 qwen_omni_utils = ["qwen_omni_utils"]
 embeddinggemma = ["transformers>=4.56.0"]
+sonar = ["fairseq2", "sonar-space", "soundfile>=0.13.1"]
```
KennethEnevoldsen, line 168: add minimal versions

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,153 @@
...
+class SONARWrapper(AbsEncoder):
+    """MTEB wrapper for Meta's SONAR text + speech embedding model."""
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None = None,
+        device: str | None = None,
+        text_encoder: str = "text_sonar_basic_encoder",
+        speech_encoder: str = "sonar_speech_encoder_eng",
```
KennethEnevoldsen, line 31: do we want this to be language specific?

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,153 @@
...
+        task_metadata: TaskMetadata,
+        hf_subset: str | None,
+        prompt_type: PromptType | None,
+    ) -> str:
+        """Resolve the SONAR-format language code ('eng_Latn') for this batch.
+
+        Uses the task's own eval_langs rather than parsing hf_subset, since
+        subset names have no guaranteed format. eval_langs uses BCP47
+        ('eng-Latn'), SONAR wants underscore-joined FLORES-200-style codes
+        ('eng_Latn').
```
KennethEnevoldsen, line 66:
```suggestion
        Uses the task's eval_langs to determine language. eval_langs uses BCP47
        ('eng-Latn'), SONAR wants underscore-joined FLORES-200-style codes
        ('eng_Latn').
```

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,153 @@
...
+
+        Tasks with `parallel_subsets=True` (e.g. FLORES, the task SONAR is
+        reproduced against) pass a single already-resolved language code as
+        hf_subset directly -- one per dataset column, not a key into
+        eval_langs -- so that case is unambiguous and used as-is.
```
KennethEnevoldsen, line 71: I am not sure I get this?

mteb/models/model_implementations/sonar_models.py
```diff
@@ -1,4 +1,153 @@
...
+        For a subset whose eval_langs entry has two languages (a bitext
+        pair), there is *no* framework-enforced ordering: FLORES/BUCC name
+        the subset to match [source, target], but DiaBLa defines the same
+        pair both ways ("fr-en": [fra, eng] and "en-fr": [eng, fra]), and
+        mteb's BitextMiningEvaluator calls encode() once per dataset column
+        with the *same* hf_subset and no prompt_type for either column, so
+        there's no signal here to tell which physical column is being
+        encoded. langs[0]/langs[-1] by prompt_type is therefore a
+        best-effort heuristic (right whenever the subset name matches the
+        list order, which held for every case checked), not a guarantee --
```
KennethEnevoldsen, line 82: Seems like we need an issue + a fix here. I would probably just raise an error in these cases (I wouldn't trust the results at least)

KennethEnevoldsen:
A few questions on language parsing, I think there are some cases that are ambiguous, some like the bitext cases seems like they require MTEB features (we should make an issue, I suspect there is one), but other cases are ambiguous as a part of the task e.g. [NordicLangClassification](https://mteb-leaderboard.hf.space/tasks/NordicLangClassification). In both cases I think the model should just error. In one referring to the issue (feature is missing) and in the other referring to SONAR not supporting the task (at least not without some custom design).

mteb/models/model_implementations/sonar_models.py
Samoed: Sonar don't have multilingual encoder. Users can override if they want

PR #5130, opened 2026-08-10
feat: add LLM retriever wrappers

AdnanElAssadi:
Factored out of #4868 as discussed.

LLM retrieval strategies as wrappers around any base retriever, each implementing `SearchProtocol`, so they compose with each other and run on ordinary retrieval tasks:

- `QueryRewriteRetriever`: LLM rewrites the query into a keyword query
- `HyDERetriever`: LLM writes a hypothetical answer passage to retrieve with
- `RerankRetriever`: base retrieves a candidate pool, LLM listwise reranks it to top_k
- `MultiQueryRetriever`: LLM query variants searched together, rankings fused with RRF
- `MultiHopRetriever`: iterative query-read search loop (the multi-hop agent from OBLIQ-Bench, arXiv 2605.06235)
- `TournamentRerankRetriever`: listwise tournament over large pools (OBLIQ-Bench appendix C)

The LLM side is `ChatModelProtocol` (one `generate(messages, **kwargs)` method) with a LiteLLM client: per-model cost accounting from litellm's local pricing table, lazy import so litellm stays an optional dependency.

Composed-retriever identity follows `HybridSearch` (`ModelMeta.create_empty`, `model_type=["hybrid"]`).

11 offline tests (fakes; one wraps the real BM25 model), including wrapper composition: `RerankRetriever(QueryRewriteRetriever(bm25, llm), llm)`.

Note: on reranking tasks that supply `top_ranked`, the rerankers currently pass it through to the base rather than reranking those candidates directly.


mteb/models/chat_models.py
```diff
@@ -0,0 +1,87 @@
...
+        max_retries: int = 2,
+        **kwargs: Any,
+    ) -> None:
+        try:
+            import litellm  # noqa: F401
+        except ImportError as exc:
+            raise ImportError(
+                "LiteLLMChatModel requires the litellm package. Install it "
+                "with pip install litellm."
+            ) from exc
```
KennethEnevoldsen, line 66: should be something like:

```
pip install mteb[agentic]
# or
pip install mteb[litellm]
```

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+    rows = []
+    for row in queries:
+        out = model.generate(
+            [{"role": "user", "content": template.format(q=row["text"])}]
+        )
+        rows.append({"id": row["id"], "text": out.text.strip() or row["text"]})
+    return Dataset.from_list(rows)
+
+
+class _WrapperBase:
```
KennethEnevoldsen: I am not sure this is needed - I think in needlessly ties implementations together

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+            corpus,
+            task_metadata=task_metadata,
+            hf_split=hf_split,
+            hf_subset=hf_subset,
+            encode_kwargs=encode_kwargs,
+            num_proc=num_proc,
+        )
+
+
+class _QueryTransformRetriever(_WrapperBase):
```
KennethEnevoldsen: I like this one though

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+    """Base for retrievers that LLM-transform the query text before searching."""
+
+    _template: str
+    _kind: str
```
KennethEnevoldsen: would fold them into the init, it should be possible to redefine these and just have defaults for hyde and query rewriter

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+        corpus: CorpusDatasetType,
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        encode_kwargs: EncodeKwargs,
+        num_proc: int | None,
+    ) -> None:
+        """Index the base retriever and cache document text."""
+        self._text = {row["id"]: row.get("text", "") for row in corpus}
```
KennethEnevoldsen: ?

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+        super().index(
+            corpus,
+            task_metadata=task_metadata,
+            hf_split=hf_split,
+            hf_subset=hf_subset,
+            encode_kwargs=encode_kwargs,
+            num_proc=num_proc,
+        )
+
+    def _docs_block(self, doc_ids: list[str]) -> str:
```
KennethEnevoldsen: this is never used in this class (probably a good indication that the subclassing is incorrect)

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+        RetrievalOutputType,
+        TopRankedDocumentsType,
+    )
+
+logger = logging.getLogger(__name__)
+
+_REWRITE = (
+    "Rewrite the question into a concise keyword search query. "
+    "Reply with only the query.\n\nQuestion: {q}"
+)
```
KennethEnevoldsen, line 39: is this really what they use? I would think I would rather expand the other way

"eiffel tower" > "tell me something about the eiffel tower".

Probably worth supporting these prompts with a paper reference

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+        )
+        ranked = [d for d in (_parse_ids(out.text) or []) if d in set(candidates)]
+        if not ranked:
+            logger.warning(
+                "listwise rerank parse failed for query %s; keeping base order", qid
+            )
+        return ranked + [d for d in candidates if d not in set(ranked)]
+
+
+class RerankRetriever(_TextCachingRetriever):
```
KennethEnevoldsen: @Samoed I seem to recall that we had some of these implemented, but couldn't find them do you recall any?

KennethEnevoldsen (commented):
(initial review, I suspect it will take a few rounds)

Would love to see these tested just to see that they produce results in the direction that we expect.

I have also changed it to a feature, which means we also need to add docs this way we can reasonably reuse the results on the docs. E.g. cost x approach using a reference model. 

In general we also need API docs for these

edit: all of these also assume text-only, I am not sure if we want to already now add support for other modalities (it is not too bad to do and it would follow our general pattern)

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+        )
+        results: dict[str, dict[str, float]] = {}
+        for qid, ranking in pool.items():
+            candidates = sorted(ranking, key=lambda d: -ranking[d])
+            ordered = self._listwise(qid, query_text[qid], candidates)
+            results[qid] = _to_scores(ordered, top_k)
+        return results
+
+
+class TournamentRerankRetriever(_TextCachingRetriever):
```
KennethEnevoldsen: can I ask you to add a citation to all the docstrings

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        top_k: int,
+        encode_kwargs: EncodeKwargs,
+        top_ranked: TopRankedDocumentsType | None = None,
+        num_proc: int | None,
+    ) -> RetrievalOutputType:
+        """Search the original query plus LLM variants, then fuse the rankings."""
+        from datasets import Dataset
```
KennethEnevoldsen: top level?

mteb/models/llm_retrievers.py
Samoed: You mean two stage reranking? Or merging scores from multilple models https://github.com/embeddings-benchmark/mteb/blob/f414a066a38482be0188c6a37f59f5f0cc3b90c2/mteb/models/hybrid_wrappers.py#L36

mteb/models/chat_models.py
```diff
@@ -0,0 +1,87 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+
+@dataclass
+class ChatResponse:
+    """One chat completion with cost accounting."""
+
+    text: str
+    cost_usd: float | None = None
```
Samoed: I don't think that this is possible to compute all time. We can keep input/output /resoning tokens that are standart in OpeanAIApi's. This probably LiteLLM specific.
```suggestion
```

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+)
+
+
+def _wrapper_meta(kind: str, base: SearchProtocol) -> ModelMeta:
+    # Composed-retriever identity, mirroring HybridSearch naming.
+    base_meta = getattr(base, "mteb_model_meta", None)
+    base_name = (getattr(base_meta, "name", None) or "unknown").rsplit("/", 1)[-1]
+    return ModelMeta.create_empty(
+        overwrites={
+            "name": f"mteb/baseline-{kind}-{base_name}",
```
Samoed: I don't think we should uses `mteb/baseline` here

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+    "You are searching a corpus for documents relevant to a question. Given "
+    "the question and notes from previous searches, reply with only the "
+    "single most useful next search query.\n\nQuestion: {q}\n\nNotes:\n{notes}"
+)
+_HOP_READ = (
+    "Identify which of the retrieved documents below are relevant to the "
+    "question. Reply with a JSON array of the relevant document ids (may be "
+    "empty), then on a new line one short observation to guide the next "
+    "search.\n\nQuestion: {q}\n\nDocuments:\n{docs}"
+)
```
Samoed, line 65: This should be customizable

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,533 @@
...
+
+def _parse_ids(text: str) -> list[str] | None:
+    match = re.search(r"\[.*\]", text, re.DOTALL)
+    if not match:
+        return None
+    try:
+        value = json.loads(match.group(0))
+    except json.JSONDecodeError:
+        return None
+    return [str(x) for x in value] if isinstance(value, list) else None
```
Samoed: I think we can try to use json scema for responses

mteb/models/llm_retrievers.py
KennethEnevoldsen: sorry no, LLM rerankers

mteb/models/llm_retrievers.py
KennethEnevoldsen: agree, but good to have reasonable defaults (but def. don't want to hide the prompt)

mteb/models/chat_models.py
KennethEnevoldsen: Yeah there is a question whether it is worth adding the litellm vs simply openai for this information

mteb/models/llm_retrievers.py
Samoed: Most close we have some models from Orion https://github.com/embeddings-benchmark/mteb/blob/f414a066a38482be0188c6a37f59f5f0cc3b90c2/mteb/models/model_implementations/promptriever_models.py#L25 but they just finetuned llms

mteb/models/chat_models.py
Samoed: We can add some libraries to handle models, but probably would be impossible to match models from one library (or inference provider) with other

mteb/models/llm_retrievers.py
KennethEnevoldsen: Ahh yes don't output natural text

mteb/models/chat_models.py
KennethEnevoldsen: I am more thinking of a cost x performance plot. cost_usd seems like a valid way to compare models to be (we of course have to make some assumptions on provider)

Samoed: Yes, this would be helpful, but this is not clear how to calculate local models, because there won't be any cost out of the box (of course we can manually approximate this from some provider)

KennethEnevoldsen: yea agree. APIs is still a fairly common way to use LLMs so I still think it is worth considering.

@AdnanElAssadi56 do we get anything else out of litellm?

AdnanElAssadi:
> (initial review, I suspect it will take a few rounds)
> 
> Would love to see these tested just to see that they produce results in the direction that we expect.

while creating these i tested on qwen3.6-27b and got these results on Obliq (some formatting issues happened midway so not final):


ndcg@10 | Congress | Math | Twitter | WildChat | Writing
-- | -- | -- | -- | -- | --
bm25 | 0.004 | 0.072 | 0.003 | 0.002 | 0.074
rewrite | 0.025 | 0.042 | 0.005 | 0.000 | 0.028
hyde | 0.033 | 0.039 | 0.015 | err | 0.028
multi-query | 0.041 | 0.052 | 0.007 | 0.000 | 0.036
rerank | 0.012 | 0.077 | 0.010 | **0.004** | **0.069**
multi-hop | **0.071** | **0.093** | **0.040** | 0.000 | 0.033


> I have also changed it to a feature, which means we also need to add docs this way we can reasonably reuse the results on the docs. E.g. cost x approach using a reference model.

> In general we also need API docs for these

agree


> edit: all of these also assume text-only, I am not sure if we want to already now add support for other modalities (it is not too bad to do and it would follow our general pattern)

I say lets start with text and then we can definitely extend once we get this in. (once we settle the points here it will be straightforward to add other modalities in followups).

mteb/models/llm_retrievers.py
AdnanElAssadi: alright, remvoved

mteb/models/llm_retrievers.py
AdnanElAssadi: added citation in docstring

mteb/models/llm_retrievers.py
AdnanElAssadi: just did

mteb/models/llm_retrievers.py
KennethEnevoldsen: This is a general comment across though

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,695 @@
...
+    rows = []
+    for row in queries:
+        out = model.generate(
+            [{"role": "user", "content": template.format(q=row["text"])}]
+        )
+        rows.append({"id": row["id"], "text": out.text.strip() or row["text"]})
+    return Dataset.from_list(rows)
+
+
+class _QueryTransformRetriever:
```
KennethEnevoldsen: if this is the only base class I would drop it (it is only used for two instances)

KennethEnevoldsen (commented):
@AdnanElAssadi56 do we have a paper to compare these scores with (I am a bit unsure what the expected ranking of them is). I assume one of the papers compare across methods

Can I also ask you to take a look at the unadressed comments

mteb/models/chat_models.py
Samoed: I stumbled on this library from pydantic. I think this could solve some problems here https://github.com/pydantic/genai-prices/blob/main/packages/python/README.md

AdnanElAssadi:
> @AdnanElAssadi56 do we have a paper to compare these scores with (I am a bit unsure what the expected ranking of them is). I assume one of the papers compare across methods
> 
> Can I also ask you to take a look at the unadressed comments

Yes: OBLIQ-Bench (arXiv:2605.06235) table 2 compares BM25, three dense models, a late-interaction model, a GPT-5.2 query rewriter, a GPT-5.2 multi-hop agent, and an oracle tournament reranker.

mteb/models/chat_models.py
AdnanElAssadi: i also agree that pricing is valuable here (even if we have to estimate for local models).

> @AdnanElAssadi56 do we get anything else out of litellm?

I guess running anthropic/gemini/bedrock/etc... directly.

> I stumbled on this library from pydantic. I think this could solve some problems here https://github.com/pydantic/genai-prices/blob/main/packages/python/README.md

nice! this probably covers the cost tracking aspect of this.

mteb/models/llm_retrievers.py
AdnanElAssadi: yes, not sure if all support schemas though

mteb/models/llm_retrievers.py
AdnanElAssadi: done

mteb/models/llm_retrievers.py
AdnanElAssadi: removed from the class

mteb/models/llm_retrievers.py
AdnanElAssadi: yep added elsewhere

mteb/models/llm_retrievers.py
AdnanElAssadi: yes defaults should be the ones from the papers, but now its customizable

mteb/models/llm_retrievers.py
AdnanElAssadi: i dropped the mteb prefix, so now it would look something similar to `hyde-qwen3.6-27b-baseline-bb25`

mteb/models/llm_retrievers.py
AdnanElAssadi: I dropped it

mteb/models/llm_retrievers.py
AdnanElAssadi: Is that what you meant here?

KennethEnevoldsen:
Table included reported in paper in parenthesis) - note paper uses gpt 5:

ndcg@10 | Congress | Math | Twitter | WildChat | Writing
-- | -- | -- | -- | -- | --
bm25 | 0.004 | 0.072 | 0.003 (.002) | 0.002 (.005) | 0.074
rewrite | 0.025 | 0.042 | 0.005 (.116) | 0.000 (.048) | 0.028
hyde | 0.033 | 0.039 | 0.015 | err | 0.028
multi-query | 0.041 | 0.052 | 0.007 | 0.000 | 0.036
rerank | 0.012 | 0.077 | 0.010 | **0.004** | **0.069**
multi-hop | **0.071** | **0.093** | **0.040** (.176) | 0.000 (.077) | 0.033

mteb/models/chat_models.py
KennethEnevoldsen: Sounds like we are on a "remove litellm" add pydantics genai-prices?

AdnanElAssadi: but how will we run anthropic/gemini/bedrock/etc... ?

mteb/models/llm_retrievers.py
KennethEnevoldsen: I would say search model first then approach then llm so:

```
{model} {apprach} ({llm})
# so:
baseline-bm25s hyDE (qwen3.6-27b)
```
I don't think we need to make path friendly.

I could also see the argument for:

```
HyDE (base=baseline-bm25s, llm=qwen3.6-27b)
```
(probably leaning toward approach first)

I don't think I want the llm first.

@Samoed do we want to make it more like an experiment?

mteb/models/llm_retrievers.py
```diff
@@ -0,0 +1,737 @@
...
+class QueryRewriteRetriever:
+    """LLM rewrites each query into a search query, then the base searches.
+
+    Reference: Ma et al., Query Rewriting for Retrieval-Augmented LLMs (arXiv:2305.14283).
+    """
+
+    def __init__(
+        self,
+        base: SearchProtocol,
+        model: ChatModelProtocol,
```
KennethEnevoldsen: Model seems very hard to differentiate.

I would probably state llm

mteb/models/llm_retrievers.py
AdnanElAssadi: is it because it copied whole corpus to memory? `index()` now keeps the corpus reference plus an id-to-row-index map, and document text is read through select(), same as `SearchCrossEncoderWrapper`

mteb/models/chat_models.py
KennethEnevoldsen: but don't they all support the openai API?

AdnanElAssadi: ah didn't know they were all supported but apparently they do

Samoed: Antropic don't support I think. Also we could use openrouter or other providers, but I think we want to prefer official API, but this will reqire a lot of wrappers

mteb/models/llm_retrievers.py
Samoed: This is complicated, because combines multiple paths and would be intresting to filter by llm. I think we need to add some better approach

mteb/models/chat_models.py
AdnanElAssadi: I found this for anthropic: https://platform.claude.com/docs/en/api/openai-sdk.

not sure how they handle `response_format` though

AdnanElAssadi: how about we add an anthropic wrapper for now? and leave others through the openai one?

AdnanElAssadi: and i dont think having a lot of provider wrappers is bad, we can add gradually as needed.

mteb/models/llm_retrievers.py
AdnanElAssadi: Should we have an extra field in model meta like this for filtering purposes:

```
composition: dict[str, str] | None = None
# {"approach": "hyde", "base": "mteb/baseline-bm25s", "llm": "qwen3.6-27b"}
```

mteb/models/llm_retrievers.py
AdnanElAssadi: changed

AdnanElAssadi:
@KennethEnevoldsen @Samoed can you guys take a look here again when you have the time

mteb/models/chat_models.py
KennethEnevoldsen: So I see two options:
1) either force openai SDK (I would say pretty much default by now) or 2) re-add the chat interface protocol

I am leaning 1), but either is fine with me (2) might allow us to add the pydantic pricing layer.

mteb/models/llm_retrievers.py
KennethEnevoldsen: I will move this to slack to resolve it a bit faster (reproducibility is a concern though) (e.g. what kwargs was used at each layer)

mteb/models/chat_models.py
Samoed: I think we need 2 to be able to support all possible models

PR #5131, opened 2026-08-10
add QuestionAnswering task type, answer systems, and BrowseCompPlusQA

AdnanElAssadi:
Second slice of #4868: answer-mode RAG evaluation on BrowseComp-Plus, as discussed. Stacked on #5130; will retarget to main when it merges.

- `AbsTaskQuestionAnswering` and a `QuestionAnswering` task type: the model answers each query from a fixed corpus, a judge grades against the reference answers. Accuracy is the main score, total LLM cost alongside. Runs through the existing `mteb.evaluate` unchanged.
- Systems: `ClosedBookAnswerer` (floor), `RAGAnswerer` (any `SearchProtocol` as retriever), `OracleAnswerer` (gold-context ceiling). Judges: normalized exact match (default), `LLMJudge` (BrowseComp grading scheme).
- `BrowseCompPlusQA`: corpus, queries, and qrels load from the existing mteb mirror; the reference answers join from the pinned source dataset, since the mirror dropped that column.

```python
task = mteb.get_task("BrowseCompPlusQA")  # judge=LLMJudge(llm) for official grading
mteb.evaluate(RAGAnswerer(llm, bm25, top_k=5), task)
```

Follow-ups: the remaining QA datasets, a per-query-context variant, re-uploading the mirror with answers included.


mteb/tasks/__init__.py
```diff
@@ -8,6 +8,7 @@
 from .multichoice import *
 from .multilabel_classification import *
 from .pair_classification import *
+from .QuestionAnswering import *
```
Samoed:
```suggestion
from .question_answering import *
```

mteb/abstasks/question_answering.py
```diff
@@ -0,0 +1,112 @@
...
+
+class QADescriptiveStatistics(SplitDescriptiveStatistics):
+    """Descriptive statistics for QuestionAnswering tasks."""
+
+    num_samples: int
+    num_queries: int
+    num_documents: int
+
+
+class AbsTaskQuestionAnswering(AbsTask):
```
Samoed, line 33: I'm not sure if we want to add new task, since this is just retrieval with extra steps

mteb/models/answer_systems.py
```diff
@@ -0,0 +1,244 @@
...
+    "Grade whether the predicted answer matches the reference answer. Extract "
+    "the final answer from the prediction and judge only whether it matches "
+    "the reference, ignoring phrasing and extra words. Reply with one line, "
+    "exactly 'correct: yes' or 'correct: no'.\n\n"
+    "Question: {q}\nReference answer: {ref}\nPrediction: {pred}"
+)
+_VERDICT_RE = re.compile(r"correct:\s*(yes|no)", re.IGNORECASE)
+
+
+class LLMJudge:
```
Samoed, line 225: Do we need judge? Why we can't just use qrels? E.g. BrowserComp have golden docs

mteb/models/answer_systems.py
```diff
@@ -0,0 +1,244 @@
...
+@dataclass
+class AnswerResult:
+    """One produced answer with cost accounting."""
+
+    text: str
+    cost_usd: float | None = None
+
+
+@runtime_checkable
+class AnswerProtocol(Protocol):
```
Samoed, line 37: Maybe reuse search?

mteb/abstasks/question_answering.py
AdnanElAssadi: but is calling it retrieval suitable here?

Samoed: I think so. It just recive a bit updated SearchWrapper that will be with LLM

mteb/models/answer_systems.py
AdnanElAssadi: thats their official way of judging. 

I didnt understand your point about qrels because we are dealing with answers here

mteb/models/answer_systems.py
AdnanElAssadi: Do we want to keep them coupled that way? An "answer" method feels weird with search

mteb/models/answer_systems.py
Samoed: > thats their official way of judging

Who is "their"? BrowserComp?

> I didnt understand your point about qrels because we are dealing with answers here

Overall, do we need answer? I think we can check only did model found relevant docs or not, answer part is not part of our pipeline

AdnanElAssadi: Their official grading is LLM-as-judge with Qwen3-32B [docs/llm_as_judge.md in texttron/BrowseComp-Plus](https://github.com/texttron/BrowseComp-Plus/blob/main/docs/llm_as_judge.md)

AdnanElAssadi: >Overall, do we need answer? I think we can check only did model found relevant docs or not, answer part is not part of our pipeline

The point of this PR is to enable answer part in our pipeline

PR #5133, opened 2026-08-10
model: add Cosmos-Embed1 (nvidia/Cosmos-Embed1-224p/336p/448p)

hubielu:
Adds NVIDIA Cosmos-Embed1, a joint video-text embedder, in three checkpoints (224p, 336p, 448p). Part of the MOEB video track, see #4842. Video and text share one embedding space, so it covers t2v, v2t and v2v, filling the thin video-to-video direction.

### Implementation notes

- **Wrapper**: follows the XCLIP pattern, with `get_text_embeddings` and `get_video_embeddings` feeding a shared `encode` dispatch. Projections come off `visual_proj` and `text_proj`.
- **transformers 4 pin**: the remote code imports `find_pruneable_heads_and_indices`, removed in transformers 5. Added a `cosmos-embed1` extra pinning `transformers<5` plus einops and torchvision, registered in the uv conflicts list like `moca` and `visrag-ret`.
- **Frames**: the processor requires exactly 8 frames and raises otherwise, so `loader_kwargs` pins `num_frames=8`. `FramesCollator` repeats frames for shorter videos.
- **Metadata**: params and memory measured per checkpoint. `training_datasets` covers all 12 Kinetics 400/600/700 tasks, which the card lists in the training pool. No BibTeX is published, so `citation` is None.

### Validation

Kinetics-400 zero-shot on 224p: **82.05** accuracy in mteb vs **83.06** F1 on the card. Not an exact comparison (accuracy vs F1, subset vs full val). MSR-VTT numbers and the three GPU-path fixes are in the comment below. `mteb mock-run` passes 31/31 text and 7/7 video.

### Checklist

### DIFF
--- a/mteb/models/model_implementations/cosmos_embed_models.py
+++ b/mteb/models/model_implementations/cosmos_embed_models.py
@@ -1,0 +1,210 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import torch
+from tqdm.auto import tqdm
+
+from mteb.models.abs_encoder import AbsEncoder
+from mteb.models.modality_collators import FramesCollator
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+
+
+def _proj(output, attr: str) -> torch.Tensor:
+    """Pull the projected embedding off a Cosmos-Embed1 output object."""
+    value = getattr(output, attr, None)
+    if not isinstance(value, torch.Tensor):
+        raise ValueError(
+            f"Expected a tensor at {type(output).__name__}.{attr}, got {type(value).__name__}"
+        )
+    return value
+
+
+class CosmosEmbed1Model(AbsEncoder):
+    """Wrapper for NVIDIA Cosmos-Embed1 joint video-text embedders."""
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str,
+        device: str = "cuda" if torch.cuda.is_available() else "cpu",
+        num_frames: int = 8,
+        **kwargs: Any,
+    ) -> None:
+        from transformers import AutoModel, AutoProcessor
+
+        self.model_name = model_name
+        self.device = device
+        self.num_frames = num_frames
+        dtype = torch.bfloat16 if device.startswith("cuda") else torch.float32
+        self.model = AutoModel.from_pretrained(
+            model_name,
+            revision=revision,
+            trust_remote_code=True,
+            torch_dtype=dtype,
+        ).to(self.device)
+        self.model.eval()
+        self.processor = AutoProcessor.from_pretrained(
+            model_name,
+            revision=revision,
+            trust_remote_code=True,
+        )
+
+    def _move(self, batch: Any) -> dict[str, Any]:
+        return {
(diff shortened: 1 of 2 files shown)


hubielu:
### Reproducibility

**Kinetics-400 zero-shot classification** (nvidia/Cosmos-Embed1-224p, `Kinetics400ZeroShot`):

| | Score |
|---|---|
| mteb, accuracy | **82.05** |
| model card, F1 | 83.06 |

About one point apart. Not an exact comparison: mteb reports accuracy where the card reports F1, mteb's task is a subset rather than the full val set, and this wrapper samples 8 frames uniformly.

The card's retrieval benchmarks are Agibot, Bridge and OpenDV, none of which are in mteb, so there's no direct retrieval comparison. As a sanity check on the retrieval path, MSR-VTT:

| | R@1 | R@5 | R@10 | nDCG@10 |
|---|---|---|---|---|
| MSRVTTT2V | 0.4960 | 0.7292 | 0.8385 | 0.6580 |
| MSRVTTV2T | 0.5051 | 0.7600 | 0.8487 | 0.6702 |

### Fixes from running on GPU

Three bugs that only appeared outside CPU mock-run, now pushed:

- `--device` is an int in the CLI, so the loader now normalises through `torch.device`
- the processor emits float32 pixels while CUDA weights are bfloat16, so float inputs are cast to the model dtype
- Kinetics clips arrive at mixed resolutions (360x640 and 360x480), so clips are preprocessed individually and concatenated after resize instead of stacked raw

`mteb mock-run` passes 31/31 text and 7/7 video on CUDA.

mteb/models/model_implementations/cosmos_embed_models.py
```diff
@@ -0,0 +1,232 @@
...
+            processed: list[dict[str, torch.Tensor]] = []
+            for video in batch["video"]:
+                clip = (
+                    video.to(torch.uint8)
+                    if isinstance(video, torch.Tensor)
+                    else torch.as_tensor(video, dtype=torch.uint8)
+                )
+                processed.append(
+                    dict(self.processor(videos=clip.unsqueeze(0), return_tensors="pt"))
+                )
```
Samoed, line 106: Can you combine this?

mteb/models/model_implementations/cosmos_embed_models.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def _proj(output, attr: str) -> torch.Tensor:
+    """Pull the projected embedding off a Cosmos-Embed1 output object."""
+    value = getattr(output, attr, None)
+    if not isinstance(value, torch.Tensor):
+        raise ValueError(
+            f"Expected a tensor at {type(output).__name__}.{attr}, got {type(value).__name__}"
+        )
+    return value
```
Samoed: Better to not use `getattr` here
```suggestion
```

hubielu: Done, dropped the helper and accessing output.text_proj and output.visual_proj directly.

mteb/models/model_implementations/cosmos_embed_models.py
hubielu: The processor won't take a batch of clips. It only accepts a single BTCHW tensor, and passing a list of any kind raises ValueError: Processor expects a numpy or torch tensor of shape BTCHW from [0-255], including a list of one element. Stacking before preprocessing isn't an option either, since Kinetics clips arrive at mixed resolutions (360x640 and 360x480) and the resize only happens inside the processor. So each clip is preprocessed individually and the results concatenated after resize.

Samoed (approved):

PR #5136, opened 2026-08-11
fix: break MRR ties by doc id to match pytrec_eval

### LINKED ISSUE #5092: A constant scorer gets mrr_at_10 = 1.0 on WebLINXCandidatesReranking due to inconsistent tie-breaking
## Summary

A "model" that assigns the *same score to every candidate* — no weights, no forward pass — achieves **`mrr_at_10 = 1.0000`** on `WebLINXCandidatesReranking/test_iid`, while scoring **`ndcg_at_10 = 0.0000`** on the same run. Since `mrr_at_10` is this task's `main_score`, and the task is part of MTEB(Multilingual, v2), this directly affects leaderboard numbers.

## Cause

Two metric families disagree on tie-breaking. [`retrieval_metrics.py::mrr`](https://github.com/embeddings-benchmark/mteb/blob/5cadabaefe6859b7cf3e815e3e123b5090fa0978/mteb/_evaluators/retrieval_metrics.py#L25-L50) ranks with a plain score sort ([L35-L37](https://github.com/embeddings-benchmark/mteb/blob/5cadabaefe6859b7cf3e815e3e123b5090fa0978/mteb/_evaluators/retrieval_metrics.py#L35-L37)):

```python
top_hits[query_id] = sorted(
    doc_scores.items(), key=lambda item: item[1], reverse=True
)[0:k_max]
```

Python's `sorted` is **stable**, so tied candidates keep their insertion order. Everything else goes through `pytrec_eval`, which breaks ties by document id — the two are computed on separate code paths in `calculate_retrieval_scores`:

```python
evaluator = pytrec_eval.RelevanceEvaluator(qrels, {map_string, ndcg_string, ...})
scores = evaluator.evaluate(results)
...
mrr_scores = mrr(qrels, results, k_values)   # separate implementation
```

| Metric | Implementation | Tie-break |
| --- | --- | --- |
| `mrr_at_k` | `retrieval_metrics.py::mrr` | Python's **stable** `sorted` → preserves candidate insertion order |
| `ndcg` / `map` / `recall` / `hit_rate` | `pytrec_eval` | document id |

`WebLINXCandidatesReranking` lists the positive candidate **first**, so preserving insertion order under ties hands the positive rank 1. The more ties a model emits, the closer `mrr_at_10` gets to 1.0.

A convenient invariant: `mrr_at_1` and `hit_rate_at_1` measure the same event, but `hit_rate` comes from `pytrec_eval`'s `success.1` while `mrr_at_1` comes from the code above. Their ratio is therefore a direct readout of tie contamination — it should be exactly 1.00.

## Evidence on a real task

`WebLINXCandidatesReranking/test_iid`. The first row is a scorer that returns the same constant for every candidate; the other two are the same checkpoint (`Querit/Querit-4B`), same data, same `mteb` 2.18.7, same dtype, same wrapper, with **only `max_length` changed**:

| Scorer | `max_length` | `mrr@10` | `ndcg@10` | `mrr@1`/`hit@1` |
| --- | ---: | ---: | ---: | ---: |
| constant | — | **1.0000** | 0.0000 | ∞ |
| Querit-4B | 4096 | 0.2172 | **0.2479** | 1.13 |
| Querit-4B | 1024 | **0.3214** | 0.2129 | 2.47 |

Between the last two rows, ranking quality drops 14% while the main score rises 48% — impossible under a single consistent ranking.

(Why ties appear at all here: rerankers put the document last in the input string and tokenizers truncate right; WebLINX queries are long (~622 tokens mean, 9288 max) and documents short (~139). At 1024, 18.2% of pairs lose the document *entirely*, making 14.0% of queries produce byte-identical inputs across all candidates. That is a property of the task, but the metric is what converts it into a higher score.)

## Why this matters beyond one task

Any model that saturates — low precision, aggressive truncation, a bounded score head — is rewarded on this `main_score`. Optimising it is partly optimising for emitting ties.

## Suggested fix

Make `mrr` tie-break consistently with the other metrics — either sort by `(-score, doc_id)`, or simply use `pytrec_eval`'s `recip_rank` so all metrics share one implementation. Optionally, report a tie-rate diagnostic (or assert `mrr_at_1 == hit_rate_at_1`) so this class of problem surfaces automatically.

Note this would change previously submitted `mrr_at_k` numbers, so the fix needs a call on how to handle existing results.

Measurements were taken on 2.18.7; the code linked above is `main`. I'm reporting rather than proposing a patch, since the right call on backwards compatibility is yours to make. Happy to share the repro scripts or run additional checks if useful.

Expected behavior: MRR and the pytrec_eval metrics should use the same deterministic tie-breaking policy, and MRR should not depend on dictionary insertion order.

## Minimal repro (no model, no dataset)

The metric depends on `dict` insertion order alone:

```python
from mteb._evaluators.retrieval_metrics import mrr

qrels = {"q1": {"d_pos": 1}}
tied = {"d_pos": 0.5, "d_x": 0.5, "d_y": 0.5}

mrr(qrels, {"q1": tied}, [10])                                # MRR@10 = 1.0
mrr(qrels, {"q1": dict(reversed(tied.items()))}, [10])        # MRR@10 = 0.333
```

The two calls produce MRR@10 values of 1.0 and approximately 0.333, respectively. Scores and qrels are identical; only dictionary insertion order changes.

### Additional information

Tested with mteb==2.18.7. The linked source code is from the current main branch.

No

serhiizghama:
Fixes #5092. `mrr` ranked candidates with a plain score sort, and since Python's `sorted` is stable, tied docs kept their dict insertion order. Everything else (`ndcg`/`map`/`recall`/…) goes through `pytrec_eval`, which breaks ties by doc id. On tasks that list the positive candidate first — `WebLINXCandidatesReranking` — that handed a constant scorer `MRR@10 = 1.0` while `nDCG@10` stayed at 0.

Switched the sort key to `(score, doc_id)` with `reverse=True`, so ties break by descending doc id and MRR shares one ranking with the pytrec_eval metrics — the direction @Samoed and @KennethEnevoldsen landed on in the thread. Applied the same one-line change to `recall_cap`, `hole` and `top_k_accuracy`, which had the identical stable-sort pattern.

Verified against `pytrec_eval`'s `recip_rank` on a few thousand random tie-heavy queries: the new key matches in every case, the old one diverged on ~45%. Added two tests — one that MRR no longer depends on insertion order, one that it agrees with `recip_rank` when the positive isn't the highest doc id.

This does shift previously reported `mrr_at_k` numbers on the two tasks that use it as `main_score` (`WebLINXCandidatesReranking`, `GermanQuAD-Retrieval`). Left the call on migrating existing results to you.

### DIFF
--- a/mteb/_evaluators/retrieval_metrics.py
+++ b/mteb/_evaluators/retrieval_metrics.py
@@ -32,8 +32,10 @@ def mrr(
     k_max, top_hits = max(k_values), {}
 
     for query_id, doc_scores in results.items():
+        # tie-break by doc id descending to match pytrec_eval; a score-only sort is
+        # stable, so tied docs kept dict insertion order and leaked into the rank (#5092)
         top_hits[query_id] = sorted(
-            doc_scores.items(), key=lambda item: item[1], reverse=True
+            doc_scores.items(), key=lambda item: (item[1], item[0]), reverse=True
         )[0:k_max]
 
     for query_id in top_hits:
@@ -60,9 +62,9 @@ def recall_cap(
     k_max = max(k_values)
 
     for query_id, doc_scores in results.items():
-        top_hits = sorted(doc_scores.items(), key=lambda item: item[1], reverse=True)[
-            0:k_max
-        ]
+        top_hits = sorted(
+            doc_scores.items(), key=lambda item: (item[1], item[0]), reverse=True
+        )[0:k_max]
         query_relevant_docs = [
             doc_id for doc_id in qrels[query_id] if qrels[query_id][doc_id] > 0
         ]
@@ -92,9 +94,9 @@ def hole(
     k_max = max(k_values)
 
     for _, scores in results.items():
-        top_hits = sorted(scores.items(), key=lambda item: item[1], reverse=True)[
-            0:k_max
-        ]
+        top_hits = sorted(
+            scores.items(), key=lambda item: (item[1], item[0]), reverse=True
+        )[0:k_max]
         for k in k_values:
             hole_docs = [
                 row[0] for row in top_hits[0:k] if row[0] not in annotated_corpus
@@ -116,7 +118,7 @@ def top_k_accuracy(
         top_hits[query_id] = [
             item[0]
             for item in sorted(
-                doc_scores.items(), key=lambda item: item[1], reverse=True
+                doc_scores.items(), key=lambda item: (item[1], item[0]), reverse=True
             )[0:k_max]
         ]
 
--- a/tests/test_evaluators/test_evaluation_metrics.py
+++ b/tests/test_evaluators/test_evaluation_metrics.py
@@ -1,4 +1,32 @@
-from mteb._evaluators.retrieval_metrics import calculate_pmrr
+import pytrec_eval
+
+from mteb._evaluators.retrieval_metrics import calculate_pmrr, mrr
+
+
+def test_mrr_tiebreak_independent_of_insertion_order():
+    # regression for #5092: with tied scores a stable score-only sort ranked docs by
+    # dict insertion order, so a constant scorer scored MRR@10 = 1.0 on tasks that list
+    # the positive candidate first.
+    qrels = {"q1": {"d_pos": 1}}
+    tied = {"d_pos": 0.5, "d_x": 0.5, "d_y": 0.5}
+
+    forward = mrr(qrels, {"q1": tied}, [10])["MRR@10"][0]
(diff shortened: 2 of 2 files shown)


ayush1298:
Seems good. @Samoed, can you check it?

Samoed:
This looks good, but we haven't decided in main issue

KennethEnevoldsen (approved):
I think this is a reasonable change, we of course need to update the results repo fairly quickly after merging this

KennethEnevoldsen:
Seems like we have a windows error - will rerun to see if it spurious (looks like it might be)

Samoed:
Yes, this test is flacky on windows for some reason

Issue #4159, opened 2026-02-24
Allow for interleave modality dataset


KennethEnevoldsen:
Currently datasets with interleaved modality samples with various modality coverage would lead to errors. E.g. a dataset with samples containing only text and samples contains only images:

```
{"texts":["text 1", "", "text 3"}, images: [[], [image1], []], ...}
```

would be great to support any arbitrary construction.

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/pull/4148#discussion_r2849030885_

tommasocerruti:
MixBench (#4953, PR #5144) is a concrete use case for this feature. Each corpus contains text-only, image-only, and image+text rows in the same split, so it cannot be evaluated faithfully until per-row optional modalities are supported.

isaac-chung:
I think one constraint for this is that we will NOT need to rerun all models after this implementation. So this needs to be done in a backwards compatible manner.

PR #5144, opened 2026-08-11
dataset: add MixBench mixed-modality retrieval

### LINKED ISSUE #4953: Add dataset: MixBench

https://huggingface.co/datasets/mixed-modality-search/MixBench2026

### Arxiv link

https://arxiv.org/abs/2507.19054

MixBench is a mixed-modality retrieval benchmark containing text, image, and image-text queries over heterogeneous corpora composed of text, images, and multimodal documents.

The benchmark contains four subsets based on MSCOCO, Google WIT, VisualNews, and OVEN. Each subset provides queries, corpora, and relevance judgments using a shared evaluation structure.

I plan to integrate the four subsets as separate MTEB tasks grouped under a single MixBench benchmark. This would provide useful coverage for mixed-modality retrieval within MOEB.

tommasocerruti:
## Summary

Adds four `Any2AnyRetrieval` tasks—`MixBenchMSCOCO`, `MixBenchGoogleWIT`, `MixBenchVisualNews`, and `MixBenchOVEN`—and the `MIXBENCH` group. It fills a MOEB gap: retrieval where text-only, image-only, and image+text documents compete in one pool. Combined rows reach encoders as one multimodal input.

## Data and metrics

Uses `mixed-modality-search/MixBench2026@17a9e705b2346b118a63f163f10e47325f9e9ecc`. Exact qrels come from `MixBench25@88e3916036ea0bdb205f4da885d6e947a565c1a0`; the current release omits them. No derivative dataset is used.

Counts (queries/corpus/qrels): MSCOCO 984/984/984; WIT 1000/4421/1000; VisualNews 981/981/981; OVEN 1000/1000/1000. Corpora preserve the near-even text/image/image+text mix. Main score: `ndcg_at_10`.

## Results

| Model | MSCOCO | WIT | VisualNews | OVEN |
| --- | ---: | ---: | ---: | ---: |
| `mteb/baseline-random-encoder` | .00599 | .00030 | .00663 | .00528 |
| `Qwen/Qwen3-VL-Embedding-2B` | .88198 | .79488 | .85505 | .54664 |
| Paper: VLM2Vec-Qwen2VL-7B | .753 | .632 | .734 | .412 |

Qwen is above random and below saturation. The paper row is not an exact reproduction: it uses another checkpoint, and public WIT has 4421 documents versus 4423 reported.

## Dataset checklist

- [x] I outlined why this dataset fills an existing gap in `mteb`.
- [x] I tested all four tasks with the `mteb` package.
- [x] I ran models and added results above:
  - [x] `Qwen/Qwen3-VL-Embedding-2B` (multimodal replacement for the suggested text-only model)
- [x] Performance is neither trivial nor random.
- [x] I retained the evaluation-scale subsets; no downsampling was needed.
- [x] I added paper scores; exact reproduction is not applicable for the reasons above.

## Validation

`make lint`, `make typecheck`, focused tests, and the full suite pass. Every qrel resolves and image decodes. Mixed-media licensing is heterogeneous; metadata uses `not specified`.

Closes #4953
Related to #4842
Blocked by #4159


mteb/abstasks/retrieval.py
```diff
@@ -532,13 +532,23 @@ def _calculate_descriptive_statistics_from_split(  # noqa: PLR0914
         # Build corpus col_inputs — text needs special mapping from the corpus dict format.
         corpus_col_inputs: dict[Modalities, list[Any]] = {}
         if "text" in corpus_modalities:
-            corpus_col_inputs["text"] = corpus.map(_corpus_to_dict)["text"]
+            corpus_col_inputs["text"] = [
+                text
+                for text in corpus.map(_corpus_to_dict)["text"]
+                if text is not None and text.strip()
+            ]
```
Samoed, line 537: Why these changes needed?

mteb/benchmarks/benchmarks/benchmarks.py
```diff
@@ -3571,3 +3571,28 @@
...
+    citation=r"""
+@article{li2025closing,
+  author = {Li, Binxu and Zhang, Yuhui and Wang, Xiaohan and Liang, Weixin and Schmidt, Ludwig and Yeung-Levy, Serena},
+  journal = {arXiv preprint arXiv:2507.19054},
+  title = {Closing the Modality Gap for Mixed Modality Search},
+  url = {https://arxiv.org/abs/2507.19054},
+  year = {2025},
+}
+""",
+    contacts=["tommasocerruti"],
```
Samoed: There shoud be authors of benchmarks
```suggestion
    contacts=None,
```

mteb/models/model_implementations/random_baseline.py
```diff
@@ -141,56 +141,29 @@ def _batch_to_embeddings(
...
-                image_embeddings,
-                audio_embeddings,
-                video_embeddings,
-            ]:
-                if i < len(embeddings_list):
-                    combined_embedding += embeddings_list[i]
-                    count += 1
-            if count > 0:
-                combined_embedding /= count
-            embeddings.append(combined_embedding)
```
Samoed, line 193: Why this is needed?

mteb/types/statistics.py
```diff
@@ -515,6 +517,14 @@ class RetrievalDescriptiveStatistics(SplitDescriptiveStatistics):
...
     num_queries: int
     num_documents: int
+    num_documents_with_text: NotRequired[int]
+    num_documents_with_image: NotRequired[int]
+    num_documents_with_audio: NotRequired[int]
+    num_documents_with_video: NotRequired[int]
+    num_queries_with_text: NotRequired[int]
+    num_queries_with_image: NotRequired[int]
+    num_queries_with_audio: NotRequired[int]
+    num_queries_with_video: NotRequired[int]
```
Samoed, line 527: This should be just equal num_queries or num_documents

mteb/tasks/retrieval/eng/mixbench_retrieval.py
Samoed, line 1: Seems some corpus and queries just have empty image column. In that case would be easier to just drop during load, rather than changing models and other parts of code

Samoed (changes requested):

mteb/abstasks/retrieval.py
tommasocerruti: MixBench rows can contain text, an image, or both. This filters out empty values so models only receive the modalities present in each row.

mteb/types/statistics.py
tommasocerruti: These counts are used for per-modality statistics, and can differ from the total (e.g., MSCOCO has 984 documents, but only 656 contain text and 656 contain an image).

mteb/tasks/retrieval/eng/mixbench_retrieval.py
tommasocerruti: I can drop columns that are empty for the whole split. However, corpus rows are mixed (some have text, some images, and some both), so we still need to filter empty values per row.

mteb/models/model_implementations/random_baseline.py
tommasocerruti: MixBench needs the baseline to skip missing modalities, but I agree this rewrite is too broad. I’ll preserve the existing logic, add only the missing-value checks, and test that normal inputs produce exactly the same embeddings as before. Would that be fine? @Samoed

mteb/tasks/retrieval/eng/mixbench_retrieval.py
Samoed: We don't support this kind of input for now

Samoed:
Can you revert changes other than tasks code? To fully support partly evaluation of modalities we need more work and this is not scope of this PR

tommasocerruti:
Ok I’ll keep this PR limited to the MixBench task code. Since MixBench requires per-row missing modalities, should I open a separate issue/PR for that support and keep this PR blocked until it lands, or what do you suggest? @Samoed

Samoed:
You can update this PR and just create an issue for this dataset

tommasocerruti:
Just to confirm: after reverting the framework changes, MixBench can be added as a beta task but cannot be evaluated faithfully until the sparse-modality issue is resolved. Should this PR remain blocked until then, or should the beta task definitions be merged first?

Samoed:
> but cannot be evaluated faithfully until the sparse-modality issue is resolved

Why so?

tommasocerruti:
Because it has missing modalities per row, not per split. It's the same limitation already described in #4159

Samoed:
I thought only one task task from these reqired this format. In that case we can't integrate these tasks

tommasocerruti:
@AdnanElAssadi56, do you agree that MixBench cannot currently be integrated? It was approved in the main MOEB issue, but all four subsets require per-row optional modalities. This limitation is already tracked in #4159, and this PR includes a focused implementation with regression tests.

Samoed:
We can't merge implementaation like this

tommasocerruti:
@Samoed, please clarify why the current implementation cannot be merged together with the fix for the framework limitation. MixBench was approved for MOEB, and I’ve invested significant effort implementing and testing it. Feedback such as

> We can't merge implementation like this

is too vague to constructively act on, you should clarify better the design you expect. 
@AdnanElAssadi56, I’d appreciate your perspective as well.

Samoed:
Currently we're expecting that all rows in columns always contain data. We do not support datasets with sparse inputs, adding it's support would require a lot of work which is out of scope of this pr

AdnanElAssadi:
@tommasocerruti Sorry about how this went, and sorry it took getting this far in before the blocker
surfaced.

Unfortunately I think @Samoed is right here. The sparse-modality support needs to land in the framework first (#4159), and it shouldn't ride along inside a dataset PR. So I think its best if we keep this on hold for now. The task code itself looks good and I'd like to see it land once the framework side is there.

MixBench will be valuable for testing the feature when it lands.

isaac-chung:
Closing this for now. Feel free to reopen later if still relevant and actively worked on.

PR #5148, opened 2026-08-11
dataset: add Mars-VL-Pairs bidirectional retrieval

### LINKED ISSUE #4955: Add dataset: Mars-VL-Pairs

https://huggingface.co/datasets/SUSTech/Mars-VL-Pairs

### Arxiv link

https://arxiv.org/abs/2602.13961

Mars-VL-Pairs is the paired image-text retrieval dataset from the MarsRetrieval benchmark. It contains 2,287 curated Martian image-text pairs spanning orbital, rover, atmospheric, geomorphic, and instrument imagery.

The dataset supports bidirectional text-to-image and image-to-text retrieval and is already released in Parquet format with images and refined captions. This would add a scientific multimodal retrieval task to MOEB.

tommasocerruti:
## Summary

Adds `MarsVLPairsT2IRetrieval` and `MarsVLPairsI2TRetrieval`, bidirectional scientific-caption↔Mars-image retrieval. It fills a MOEB gap: MTEB has no planetary-science retrieval task.

## Data and metrics

Uses expert-validated `refined_caption` from `SUSTech/Mars-VL-Pairs@1cac8885e481256d3752ad0d3a0f8f9681c5f206`. Frozen data: `Cerru02/Mars-VL-Pairs-MTEB@f0084ab0ba2f584b15dc72a82502b38ee490f58d`.

Recovered 2,249/2,287 images: 2,203 direct, 1 HTTPS, 42 Wayback, and 3 Wikimedia; 38 were unavailable. Removed the lower-resolution item from two resize-equivalent pairs, leaving 2,247 unique one-to-one pairs with provenance. The source declares CC-BY-4.0; external-image rights may vary.

Both tasks rank the full gallery with main score `mrr_at_2247`.

## Results (%)

| Model | Direction | R@1 | R@10 | MRR | MedR |
| --- | --- | ---: | ---: | ---: | ---: |
| random | T→I | 0.05 | 0.49 | 0.34 | — |
| random | I→T | 0.00 | 0.45 | 0.34 | — |
| ViT-L-16-SigLIP2-512 (exact OpenCLIP) | T→I | 12.73 | 28.48 | 18.24 | 78 |
| ViT-L-16-SigLIP2-512 (exact OpenCLIP) | I→T | 13.53 | 29.28 | 19.11 | 53 |
| Qwen3-VL-Embedding-2B | T→I | 13.04 | 29.37 | 18.68 | 76 |
| Qwen3-VL-Embedding-2B | I→T | 18.51 | 34.53 | 23.97 | 61 |

Using the paper’s exact OpenCLIP checkpoint improves over the equivalent Transformers checkpoint. Matching the paper’s 512px center-crop preprocessing gives similar R@10 results of 29.02 T→I and 29.77 I→T.

The paper reports SigLIP2 R@10 of 42.37 T→I and 37.12 I→T on all 2,287 pairs. The remaining difference likely comes from the underlying images, since the paper’s exact processed snapshot is unavailable and the source URLs are mutable.

The paper reports Qwen T→I 16.97/33.45/22.55/62 and I→T 15.96/33.71/22.24/47. Our 2,247-pair run uses MTEB prompts, so it is metric-equivalent rather than an exact protocol reproduction.

## Dataset checklist

- [x] Fills an existing gap: first planetary-science retrieval tasks in MTEB.
- [x] Both tasks load and run with the `mteb` package.
- [x] Ran random plus `Qwen/Qwen3-VL-Embedding-2B` (multimodal substitute for the text-only suggestion).
- [x] Qwen is clearly above random and below saturation in both directions.
- [x] Kept all 2,247 valid pairs; the dataset is already evaluation-scale.
- [x] Compared with the paper above; the exact OpenCLIP checkpoint is clearly above random, although the paper’s exact processed image snapshot is unavailable.

## Validation

6,724 practical tests, Ruff, typos, mypy, media decoding, ID/qrel integrity, and descriptive-stat checks pass.

Closes #4955
Related to #4842


mteb/tasks/retrieval/eng/mars_vl_pairs.py
```diff
@@ -0,0 +1,148 @@
...
+        revision=task.metadata.dataset["revision"],
+        split="test",
+        num_proc=num_proc,
+    )
+    if len(pairs) != _FROZEN_PAIRS:
+        raise ValueError(f"Expected {_FROZEN_PAIRS} frozen pairs, found {len(pairs)}")
+
+    keys = pairs["key"]
+    query_ids = [f"query-{key}" for key in keys]
+    corpus_ids = [f"corpus-{key}" for key in keys]
```
Samoed: You can keep query and corpus without prefixes to simplify a bit

mteb/tasks/retrieval/eng/mars_vl_pairs.py
```diff
@@ -0,0 +1,148 @@
...
+        dialect=[],
+        sample_creation="multiple",
+        bibtex_citation=_BIBTEX,
+        prompt={
+            "query": "Retrieve the Mars image that matches this scientific description."
+        },
+        is_beta=True,
+    )
+    k_values = (1, 3, 5, 10, 20, 100, 1000, _FROZEN_PAIRS)
+    _top_k = _FROZEN_PAIRS
```
Samoed:
```suggestion
```

mteb/tasks/retrieval/eng/mars_vl_pairs.py
```diff
@@ -0,0 +1,148 @@
...
+        annotations_creators="LM-generated and reviewed",
+        dialect=[],
+        sample_creation="multiple",
+        bibtex_citation=_BIBTEX,
+        prompt={
+            "query": "Retrieve the Mars image that matches this scientific description."
+        },
+        is_beta=True,
+    )
+    k_values = (1, 3, 5, 10, 20, 100, 1000, _FROZEN_PAIRS)
```
Samoed:
```suggestion
    k_values = (1, 3, 5, 10, 20, 100, 1000)
```
I don't think it would be useful, because it will always output 1

mteb/tasks/retrieval/eng/mars_vl_pairs.py
tommasocerruti: We need this to rank the full gallery, as the official benchmark does, without it, MTEB only retrieves the top 1,000 results.

mteb/tasks/retrieval/eng/mars_vl_pairs.py
tommasocerruti: True for recall, but MRR@2247 is not always 1, since it depends on the correct match’s rank. I’ll compute full-gallery MRR separately so we don’t generate unnecessary metrics at 2247

mteb/tasks/retrieval/eng/mars_vl_pairs.py
```diff
@@ -0,0 +1,162 @@
...
+    def task_specific_scores(
+        self,
+        scores: dict[str, dict[str, float]],
+        qrels: RelevantDocumentsType,
+        results: dict[str, dict[str, float]],
+        hf_split: str,
+        hf_subset: str,
+    ) -> dict[str, float]:
+        full_gallery_mrr = mrr(qrels, results, (_FROZEN_PAIRS,))[f"MRR@{_FROZEN_PAIRS}"]
+        return {f"mrr_at_{_FROZEN_PAIRS}": fmean(full_gallery_mrr)}
```
Samoed: Won't mrr automatically compute across all `top_k` values?

mteb/tasks/retrieval/eng/mars_vl_pairs.py
```diff
@@ -0,0 +1,162 @@
...
+                relevant_docs=qrels,
+                top_ranked=None,
+            )
+        }
+    }
+    task.data_loaded = True
+
+
+class _MarsVLPairsRetrieval(AbsTaskRetrieval):
+    _top_k = _FROZEN_PAIRS
```
Samoed: You shoudn't change private values

mteb/tasks/retrieval/eng/mars_vl_pairs.py
tommasocerruti: `top_k` controls how many results are retrieved, while `k_values` controls which MRR scores are calculated. Since `k_values` stops at 1000, we need to calculate MRR@2247 separately.

mteb/tasks/retrieval/eng/mars_vl_pairs.py
tommasocerruti: Got it. The official benchmark computes MRR over all 2247 candidates, and MTEB only retrieves 1000 by default, so matches below rank 1000 would be missing, making the score differ from the official protocol. I used `_top_k` because the current retrieval task API doesn't expose a public way to change this limit. So I see two options:

 - Keep `_top_k = 2247` to follow the official protocol.
 - Use MRR@1000, following MTEB’s default but differing from the paper.
 
Would you prefer the second option, or is there another supported way to request full-gallery retrieval?

mteb/tasks/retrieval/eng/mars_vl_pairs.py
Samoed: You can just set k_values to include `_FROZEN_PAIRS` and this also would solve https://github.com/embeddings-benchmark/mteb/pull/5148#discussion_r3772587966

tommasocerruti: That would remove the need for the separate MRR calculation, but `_top_k` is inherited from the base class as 1000 and isn’t recalculated when a task overrides `k_values`, so only 1000 results would still be retrieved. We would therefore still need `_top_k = 2247` and it would also bring back `recall_at_2247 = 1` which I removed based on your earlier feedback

Samoed: If this is metric from paper, then better to use it

tommasocerruti: Agree, done.

mteb/tasks/retrieval/eng/mars_vl_pairs.py
Samoed: Can you remove `_top_k = _FROZEN_PAIRS`? Also would be easier to just add `k_values` explicitly to classes, rather than inherit from other class

tommasocerruti: These two requests conflict with the current API: reproducing the paper requires ranking all 2,247 candidates, but removing `_top_k = 2247` makes MTEB retrieve only 1,000. In that case, `mrr_at_2247` would not reproduce the paper’s metric. I moved `k_values` into both classes, but kept `_top_k = 2247` for correctness. Is there another supported way to request all 2,247 results?

Samoed: Submited fix in https://github.com/embeddings-benchmark/mteb/pull/5148/commits/f761ff94718853e3d86cd7fed74bd1fffc14ecf0

tommasocerruti:
@Samoed It overwrites the existing `_top_k` values in subclasses, which is causing the test failures in CI

Samoed:
I run `google/siglip2-large-patch16-512` (I think this is from paper `ViT-L-16-SigLIP2-512`)

and got these results

| | MTEB R@10 | PAPER R@10 |
| - | - | - |
| MarsVLPairsI2TRetrieval | 0.21762 | 0.3712 |
| MarsVLPairsT2IRetrieval | 0.23676 | 0.4237 |

Can you investigate score difference?

tommasocerruti:
@Samoed I ran the exact OpenCLIP checkpoint from the paper on all 2247 frozen pairs: it improves R@10 to 28.48 T2I / 29.28 I2T, and matching the paper’s preprocessing gives similar results, confirming that the dataset and evaluation behave as expected. The remaining difference likely comes from the underlying images, since the exact processed snapshot is not publicly available and the source URLs are mutable. I added these reproducible results to the PR description

Samoed:
Where did you find that `ViT-L-16-SigLIP2-512` is open clip?

tommasocerruti:
@Samoed The official MarsRetrieval code uses `hf-hub:timm/ViT-L-16-SigLIP2-512` through OpenCLIP, so I used the same implementation as the paper.

PR #5150, opened 2026-08-11
Add KiteFishAI/Nano-Em1-0.6B-v2.1

anuj0456:


anuj0456:
@Samoed releasing a new version

mteb/models/model_implementations/kite_fish_models.py
```diff
@@ -23,41 +36,30 @@
     model_type=["dense"],
     public_training_code=None,
     public_training_data=None,
-    training_datasets={
-        "MSMARCO",
-        "HotpotQA",
-        "FEVER",
-        "NQ",
-        "DBPedia",
```
ayush1298, line 31: lot of these from old `training_datasets` are removed in the newly defined one; were they incorrect before?

ayush1298:
Can you also change PR description and add model checklist

mteb/models/model_implementations/kite_fish_models.py
anuj0456: i kept the ones that are common in both. added all of them again

anuj0456:
@ayush1298 updated the model checklist as well

mteb/models/model_implementations/kite_fish_models.py
ayush1298: If training datasets are different, then you can define them accordingly instead of a single dict

anuj0456: yes, but i dont see any harm in keeping one dict.

ayush1298:
Can you also try to reproduce any results if there are

mteb/models/model_implementations/kite_fish_models.py
ayush1298: It's just to keep metadata perfect to the extent possible, so if both models have different training datasets, it would be good if you mention them accordingly.

anuj0456: done

anuj0456:
> Can you also try to reproduce any results if there are

yes, i will publish the results in results repo once this is merged. just like i did in previous version

Samoed (approved):

PR #5163, opened 2026-08-12
ci: add Copilot dataset PR review via label trigger

isaac-chung:
Adds a GitHub Actions workflow and a Copilot instructions file to automate dataset PR reviews.

**How it works:** when the `new dataset` label is added to a PR, the workflow posts a comment that triggers a Copilot review against `.github/instructions/dataset-pr.instructions.md`.

Checks enforced:
- Descriptive stats present in the PR description
- Paper reproduction results in a table (models as rows, score source as columns)
- Random encoder baseline reported
- No AI-generated boilerplate in the description
- Upload script in `scripts/data/` when the HF dataset username matches the PR author

### DIFF
--- a/.github/instructions/dataset-pr.instructions.md
+++ b/.github/instructions/dataset-pr.instructions.md
@@ -1,0 +1,51 @@
+---
+applyTo: "**"
+---
+
+# New Dataset PR Guidelines
+
+## Purpose
+
+These instructions apply only when the PR title starts with `task:` or `dataset:`, or when the PR adds or modifies files under `mteb/tasks/`. If neither condition is met, ignore all instructions in this file.
+
+For matching PRs, ensure the PR description includes the evidence and context reviewers need to evaluate correctness and quality.
+
+## Descriptive Statistics
+
+- The PR description must include descriptive statistics for the dataset directly (e.g. number of samples per split, average text length, label distribution, language breakdown).
+- A reference to the dataset card is not sufficient — the stats must appear in the PR itself.
+
+## Results Table Format
+
+- Reproduction results from the paper must be presented as a table, not as prose.
+- Columns: score source (e.g. "PR" and "Paper").
+- Rows: one row per model.
+- Flag the PR if results are described only in paragraph form, or if the table is transposed (models as columns).
+
+Example of correct format:
+
+| Model | PR | Paper |
+|---|---|---|
+| model-a | 42.1 | 42.3 |
+| model-b | 38.7 | 39.0 |
+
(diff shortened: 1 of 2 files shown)


isaac-chung:
@Samoed @KennethEnevoldsen FYI

isaac-chung:
Maybe this solution with the current * file path doesn't make sense - all PRs would trigger this review. 

Perhaps this works better as a GitHub bot workflow of some sort instead.

ayush1298:
I was discussing these with Kenneth today. Can we also create a skills file to automatically review particularly 2 types of PRs: adding a new model and a new task, as they mostly follow a structural pattern.
The issue with copilot is, in general, its limit is too low

isaac-chung:
I think I'll give the following a try:
- trigger GH action when a new tag (new dataset) is added to a PR
- the workflow adds a comment that @'s copilot with a prompt similar to the instruction in the md file

isaac-chung:
@copilot review this PR for instruction formatting and ambiguity, to make sure there are no contradicting instructions.

KennethEnevoldsen:
I am mostly concerned about:

"Upload script in scripts/data/ when the HF dataset username matches the PR author"

I would rather just put that in the dataset repo (we will just end up deleting it if it is here)

~~(cc @ayush1298 you also mentioned this)~~ you already saw

isaac-chung:
> I am mostly concerned about:
> 
> "Upload script in scripts/data/ when the HF dataset username matches the PR author"
> 
> I would rather just put that in the dataset repo (we will just end up deleting it if it is here)
> 
> ~(cc @ayush1298 you also mentioned this)~ you already saw

What's the concern? If it's not there, then how will we know how it was uploaded? What's the dataset repo?

.github/instructions/dataset-pr.instructions.md
```diff
@@ -0,0 +1,47 @@
+# New Dataset PR Guidelines
```
KennethEnevoldsen, line 1: I would also want the actual checklist to be present

isaac-chung: Checklist is translated into instructions. I don't think simply duplicating the checklist here will give desired results consistently. e.g. for `I have tested that the dataset runs with the `mteb` package.`, Copilot reviews the PR description and diff, it can't execute code or check CI logs. The only signal available at review time is whether model scores are present in the description, which already implies the task ran.

isaac-chung:
Let's merge to see how it goes, then re-assess.

PR #5168, opened 2026-08-13
dataset: add MM-BRIGHT retrieval tasks

### LINKED ISSUE #4685: Add dataset: MM-BRIGHT

https://huggingface.co/datasets/mm-bright/MM-BRIGHT

### Arxiv link

https://arxiv.org/abs/2601.09562

_No response_

guangyusong:
Closes #4685.

Adds MM-BRIGHT Tasks 1–3 as separate retrieval tasks for each of the 29 domains: T2T, IT2T, and IT2I. T2T and IT2T exclude each query’s `negative_ids`; IT2I uses the full image corpus because those IDs refer to text documents. Multi-image queries are concatenated vertically following the paper. Task 4 was removed following the earlier review discussion.

Mean nDCG@10 across the 29 domains:

| Model | T2T | IT2T | IT2I |
|---|---:|---:|---:|
| Random | 0.0000 | 0.0000 | 0.0323 |
| CLIP ViT-L/14 | 0.0957 | 0.1024 | 0.3246 |
| Paper CLIP | not reported | 0.1040 | 0.3360 |

- [x] Fills an existing MTEB gap
- [x] Runs with MTEB
- [x] Evaluated with random and CLIP
- [x] Performance is non-trivial and non-random
- [x] Uses the paper-aligned evaluation setup
- [x] Compared with published CLIP results


Samoed:
Can you compare scores with paper?

mteb/tasks/retrieval/eng/mm_bright_retrieval.py
```diff
@@ -0,0 +1,511 @@
...
+    no_image_ids = pc.binary_join_element_wise(
+        document_ids, f"{PAIR_SEPARATOR}{NO_IMAGE}", ""
+    )
+    no_image_table = pa.table(
+        {
+            "id": no_image_ids,
+            "text": document_texts,
+            "image": pa.array(
+                [_BLANK_IMAGE] * len(documents), type=_IMAGE_FEATURE.pa_type
+            ),
```
Samoed:
```suggestion
```

mteb/tasks/retrieval/eng/mm_bright_retrieval.py
```diff
@@ -0,0 +1,511 @@
...
+
+
+def _pair_corpus(
+    domain: str, documents: Dataset, images: Dataset, examples: Dataset
+) -> tuple[Dataset, set[str]]:
+    document_ids = documents.data.column("id")
+    document_texts = documents.data.column("text")
+    no_image_ids = pc.binary_join_element_wise(
+        document_ids, f"{PAIR_SEPARATOR}{NO_IMAGE}", ""
+    )
```
Samoed: Why this needed?

Samoed (changes requested):
Seems this dataset using `gold_ids` and `negative_ids`, but I don't know how many values they have since preview is broken

mteb/tasks/retrieval/eng/mm_bright_retrieval.py
guangyusong: These are the text-only Task 4 candidates. The dataloader supports null images, but stats and CLIP don’t yet.

mteb/tasks/retrieval/eng/mm_bright_retrieval.py
guangyusong: Task 4 uses relevance 1 for a gold passage without an image and relevance 2 with the corresponding positive image. The suffix gives the text-only version its own ID.

guangyusong:
> Can you compare scores with paper?

Compared with the paper:
- IT2T: 0.096 vs 0.104
- IT2I: 0.322 vs 0.336
- IT2IT: 0.256 vs 0.280

guangyusong:
Across 1,218 multimodal queries, gold_ids has 1–11 per query, averaging 2.40. negative_ids has 0–2,073 per query, averaging 171.25, with 101 empty.

mteb/tasks/retrieval/eng/mm_bright_retrieval.py
Samoed: > These are the text-only Task 4 candidates

Then why do you fill it with blank images?

Samoed:
I think you should also use negative_ids and convert this to reranking task. This can produce changes in scores

guangyusong:
> @Samoed Thanks for the review. I updated all four tasks to use the released candidate sets, including `negative_ids`, for reranking. Task 4 now uses `image=None` for text-only candidates instead of blank images.
>
> I also added narrow `None` handling to the statistics calculator and random baseline. Both previously failed on missing images with `AttributeError: 'NoneType' object has no attribute 'tobytes'`.

Samoed:
How scores looks like now after your changes?

guangyusong:
> How scores looks like now after your changes?

@Samoed Here are the CLIP ViT-L/14 mean nDCG@10 scores before and after the reranking update:

| Task | Before | After |
|---|---:|---:|
| T2T | 0.088 | **0.536** |
| IT2T | 0.096 | **0.555** |
| IT2I | 0.322 | **0.820** |
| IT2IT | 0.256 | **0.600** |

Mean across all 29 subsets. Before used full-corpus retrieval and after uses the released reranking candidates.

Samoed:
I'm not sure how to compare your results with results from paper. I'm not sure that this is same model and I can't find resultls from description in paper

mteb/abstasks/_statistics_calculation.py
```diff
@@ -428,16 +428,33 @@ def calculate_single_input_modality_statistics(
...
+                "Expected one image hash per input, got "
+                f"{len(image_hashes)} hashes for {len(images)} inputs"
+            )
+        present_indices = [
+            index for index, image in enumerate(images) if image is not None
+        ]
+        images = [images[index] for index in present_indices]
+        if image_hashes is not None:
+            image_hashes = [image_hashes[index] for index in present_indices]
+
```
Samoed: Why this is needed? Tasks shoudn't have such rows. Which subset contains this?

Samoed (changes requested):

mteb/abstasks/_statistics_calculation.py
guangyusong: Task 4 intentionally includes text-only candidates in all 29 subsets. The paper scores them as rel=1, and the official loader uses image=None. Since MTEB does not support this, I can remove Task 4 and the shared change.

Samoed: Yes, better to remove it for now. You can create issue for this task

guangyusong:
> I'm not sure how to compare your results with results from paper. I'm not sure that this is same model and I can't find resultls from description in paper

It is the same CLIP ViT-L/14 model. Tables 4 and 5 report mean nDCG@10 across all 29 domains: 0.104 for IT2T and 0.336 for IT2I. Our full-corpus means were 0.096 and 0.322. I can restore the earlier full-corpus setup to match the paper.

Samoed:
> It is the same CLIP ViT-L/14 model

Agree, I found it in code https://github.com/mm-bright/MM-BRIGHT/blob/2d37a4ddcbb3437aee83c285bddd2cb13bf22dbf/src/retrievers/task2_multimodal.py#L34

> Our full-corpus means were 0.096 and 0.322. I can restore the earlier full-corpus setup to match the paper.

Yes, we need to match the paper. I tried to follow code in their repo but didn't understand fully, do they use `negative_ids`. 


Also I think would be good to split dataset per domains

guangyusong:
Thanks a lot! This one was tricky to untangle. I updated it to match the paper’s full-corpus and vertical-concatenation setup, split each of Tasks 1–3 across all 29 domains, and removed Task 4 and the shared changes. `negative_ids` are excluded for T2T and IT2T. They point to text documents, so they do not apply to IT2I.

CLIP ViT-L/14 mean nDCG@10:

| Task | This PR | Paper |
|---|---:|---:|
| IT2T | 0.1024 | 0.1040 |
| IT2I | 0.3246 | 0.3360 |

Samoed:
What about other modality combinations? Scores from https://github.com/embeddings-benchmark/mteb/pull/5168#issuecomment-5404591026 are differ from paper

mteb/tasks/retrieval/eng/mm_bright_retrieval.py
```diff
@@ -0,0 +1,1005 @@
...
+                image_lookup[path]
+                for path in example["image_paths"]
+                if path in image_lookup
+            ]
+        )
+        if image is None:
+            from PIL import Image as PILImage
+
+            missing_query_image_ids.add(example["id"])
+            image = PILImage.new("RGB", (224, 224), "white")
```
Samoed, line 203: Why this is needed?

Samoed (commented):
Is it possible to get your scores for `IT2T` from results of current tasks or they reqire some aggregations?

guangyusong:
> What about other modality combinations? Scores from [#5168 (comment)](https://github.com/embeddings-benchmark/mteb/pull/5168#issuecomment-5404591026) are differ from paper

Those were reranking scores and no longer apply. Current scores are T2T 0.0957, IT2T 0.1024, and IT2I 0.3246. IT2IT was removed earlier.

Samoed:
Can you update table?

guangyusong:
> What about other modality combinations? Scores from https://github.com/embeddings-benchmark/mteb/pull/5168#issuecomment-5404591026 are differ from paper

Those were reranking scores and no longer apply. Current scores are T2T 0.0957, IT2T 0.1024, and IT2I 0.3246. IT2IT was removed earlier.

> Is it possible to get your scores for `IT2T` from results of current tasks or they reqire some aggregations?

They require a simple aggregation: the IT2T score is the unweighted mean across the 29 domain tasks.

> Why this is needed?

Twelve queries have no usable image. The official [Task 2](https://github.com/mm-bright/MM-BRIGHT/blob/2d37a4ddcbb3437aee83c285bddd2cb13bf22dbf/src/retrievers/task2_multimodal.py#L82-L97) and [Task 3](https://github.com/mm-bright/MM-BRIGHT/blob/2d37a4ddcbb3437aee83c285bddd2cb13bf22dbf/src/retrievers/task3_image.py#L252-L276) evaluators use the same white-image fallback.

> Can you update table?

Updated.

isaac-chung:
@Samoed instead of asking Kenneth, can you give a verdict and propose a way forward? Accept or close, or fill in the gaps ourselves.

Samoed:
Currently I'm reuploading data for this task for future merge

Samoed (approved):

PR #5165, opened 2026-08-13
[MOEB]fix: restore combined OVEN/InfoSeek candidate pool for retrieval

lxer19:
## Summary

OVEN and InfoSeek are intended to use a combined candidate pool in the original M-BEIR preprocessing, but the current MTEB exports retain cross-dataset qrels while only including the task-local corpus.

### Evidence

The original M-BEIR preprocessing explicitly builds a combined OVEN + InfoSeek candidate pool:

```python
combined_pool_dict = {**infoseek_cand_pool, **oven_1m_cand_dict}
````

The current OVEN export instead contains only OVEN (`5:`) corpus IDs, while its qrels also reference InfoSeek (`6:`) documents:

```text
corpus prefixes: {'5': 676667}
qrel prefixes:   {'5': 50004, '6': 442650}
```

This leaves valid qrel targets absent from the searchable corpus.

### Fix

This PR adds `.v2` tasks that restore the combined corpus while keeping queries and qrels unchanged.

Before the fix:

* OVEN: 21,689 missing qrel corpus IDs
* InfoSeek: 941 missing qrel corpus IDs

After the fix:

* 1,288,318 corpus entries
* 0 ID collisions
* 0 missing qrel corpus IDs for both tasks

The existing v1 tasks are preserved so previously published results are not silently changed.

Validation: metadata tests, task-quality tests, and ruff all pass.

Related to #5021.

```
```

PR #5172, opened 2026-08-13
ci: Fix Copilot review comment command

isaac-chung:
Updated the comment command to use '@copilot' instead of '@github-copilot'.


Samoed:
I tried to call it in https://github.com/embeddings-benchmark/mteb/pull/5165 but got warning `Copilot isn't available for cross-repository pull requests `

isaac-chung:
Let me look into it a bit more

PR #5173, opened 2026-08-13
fix: require datasets>=4.0.0 for video extra

whybe-choi:
Related to #5116

I used `datasets>=4.0.0` rather than `>=3.1.0`. Although the `Video` feature was introduced in 3.1.0, it only switched to the `torchcodec` backend in datasets 4.0.0. This keeps the `video` extra internally consistent.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,7 +91,7 @@ api = [
 og = [
     "playwright>=1.49.0",
 ]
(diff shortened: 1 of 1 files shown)


Samoed (commented):
Can you regenerate `uv.lock`

Samoed (approved):

PR #5179, opened 2026-08-13
fix: Add support for more metrics in ZeroShot classification

Samoed:
Currently ZeroShotClassification reports only accuracy, but after https://github.com/embeddings-benchmark/mteb/pull/4790 is possible to use any sklearn metric. Added same metrics as in Classification task

### DIFF
--- a/mteb/abstasks/zeroshot_classification.py
+++ b/mteb/abstasks/zeroshot_classification.py
@@ -240,7 +240,6 @@ class AbsTaskZeroShotClassification(AbsTask):
             ap_weighted=None,
         )
 
-        # if binary classification, also report average precision using the
         if len(np.unique(labels)) == 2:
             positive_scores = probs[:, 1]
             scores["ap"] = metrics.average_precision_score(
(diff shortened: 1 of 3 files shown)


mteb/abstasks/zeroshot_classification.py
```diff
@@ -153,11 +172,14 @@ def _evaluate_subset(
                 hf_split=hf_split,
             )
 
+        probs_array = (
+            probs.cpu().numpy() if torch.is_tensor(probs) else np.asarray(probs)
+        ).astype(np.float64)
```
KennethEnevoldsen: is this normalization required? I believe sklearn also tries to adopt the array API spec (#941), would be great to make as few assumptions as possible.

KennethEnevoldsen (commented):
Generally looks good, but would be great is we could be more array agnostic (it depends on how sklearn treats the input)

mteb/abstasks/zeroshot_classification.py
Samoed: I don't think so, but previosly we converted to tensor. I'll test that

Samoed: Removed conversion

KennethEnevoldsen (approved):

PR #5189, opened 2026-08-14
Enables `B` (flake8-bugbear) ruff rule

ayush1298:
refs #2416.

Takes the bug classes and solves them. `B905` (114 occurrence places) and `B028` (39 occurrence places) are deferred to a follow-up with a TODO in the ignore list — each needs a per-site `strict=` / `stacklevel=` decision.

### 84 fixes

| Rule | n | Pattern applied |
|---|---|---|
| `B007` unused-loop-control-variable | 26 | `for k, v in d.items()` → `.values()` when the key is unused (and `for k in d` when the value is); dropped `enumerate()` where the index was never read |
| `B023` function-uses-loop-variable | 25 | Bound the loop variable as a default arg (`lambda x, split=split: ...`) |
| `B904` raise-without-from | 14 | `except ... as e` → `from e`; unbound optional-dependency guards → `from None` |
| `B006` mutable-argument-default | 7 | List defaults → tuples, annotated `Sequence[...]` (matches the existing convention elsewhere in the package) |
| `B008` function-call-in-default | 4 | Module-level singletons |
| `B020` loop-var-overrides-iterator | 2 | Renamed the shadowing loop variable |
| `B026` star-arg-after-keyword | 2 | Moved `*args` ahead of the keyword arguments |
| `B011` assert-false | 2 | `assert False, msg` → `raise AssertionError(msg)` |
| `B009` getattr-with-constant | 1 | Direct attribute access |
| `B027` empty-method-in-ABC | 1 | `noqa` — deliberate optional hook, not abstract |

### Notes

- **`B023` sites are not live bugs.** `datasets.map()`/`.filter()` are eager, so each iteration already sees its own value. change makes the capture explicit rather than incidental.
- **`B006`** required widening `list[str]` → `Sequence[str]` . Backward-compatible. None of the defaults are mutated.
- **`B008`** uses module-level singletons rather than a `None` sentinel. Both are constructed at import time, so behaviour is unchanged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -312,6 +312,7 @@ select = [
     "INT",     # gettext f-string misuse
     "ASYNC",   # blocking calls in async functions
     "ISC",     # implicit string concatenation
+    "B",       # flake8-bugbear
 ]
 
 ignore = [
@@ -334,6 +335,9 @@ ignore = [
     "PLR0915",  # too many statements
     "PLW0717",  # too-many-statements-in-try-clause
     "ISC001",   # single-line implicit concat; conflicts with `ruff format`
+    # TODO: enable in a follow-up PR; each needs a per-site `strict=` / `stacklevel=` decision
+    "B905",     # zip-without-explicit-strict
+    "B028",     # no-explicit-stacklevel
 ]
 future-annotations = true  # For TC rules
 typing-modules = ["typing_extensions", "mteb.types", "collections.abc"]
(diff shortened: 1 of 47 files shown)


pyproject.toml
```diff
@@ -334,6 +335,9 @@ ignore = [
     "PLR0915",  # too many statements
     "PLW0717",  # too-many-statements-in-try-clause
     "ISC001",   # single-line implicit concat; conflicts with `ruff format`
+    # TODO: enable in a follow-up PR; each needs a per-site `strict=` / `stacklevel=` decision
+    "B905",     # zip-without-explicit-strict
```
Samoed, line 339: I think would be better to enable this, but we have a bit strange case with BUCC which should be checked

Samoed (approved):

pyproject.toml
ayush1298: Yes, as I mentioned in the description, B905 (114 occurrence places) and B028 (39 occurrence places), so these need more thorough checking, so I kept it for a separate PR.

PR #5193, opened 2026-08-14
model: add VideoPrism video encoders (base, large, LVT base, LVT large)

hubielu:
Adds the four public VideoPrism checkpoints, closing the video-encoder gap alongside V-JEPA 2 and InternVideo2 (#5078). In transformers since 5.15.

`VideoPrismVisionWrapper` covers base and large: no text tower, so classification/clustering only, same shape as vjepa2. `VideoPrismClipWrapper` covers lvt-base and lvt-large: video plus text, so t2v/v2t too.

## Reproduction

MSR-VTT zero-shot retrieval, R@1, vs the model cards:

| checkpoint | task | measured | card |
| --- | --- | --- | --- |
| lvt-base | MSRVTTT2V | 50.06 | 50.1 |
| lvt-base | MSRVTTV2T | 49.83 | 49.8 |
| lvt-large | MSRVTTT2V | 51.54 | 50.1 |
| lvt-large | MSRVTTV2T | 50.97 | 50.6 |

Our split has 1,758 samples where the cards report MSRVTT-1K. K400 zero-shot on lvt-base gives 67.56 vs 69.2.

Linear probes are weaker and training data seems to explain it: K400 gives 51.65 vs 82.9 on ~10 train clips per class where the paper probes on ~600, while UCF101 at ~96 per class scores 86.92 on the same wrapper. Diving48 has ~313 per class and still gives 4.46 vs 64.3, which is the pooling point below.

## Worth flagging

**Weights are pinned to commits inside unmerged Hub PRs.** No repo has transformers-format weights on `main`, only the Flax `.npz`; they sit in `refs/pr/4` (open) and `refs/pr/2`/`refs/pr/3` (drafts). Bare SHAs resolve and all four load with zero missing or mismatched keys, but if Google force-pushes a draft the SHA stays valid while no longer matching the PR. Happy to wait for a merge.

**The text tower has 64 position embeddings.** Longer input raises a shape mismatch rather than truncating, so the wrapper pins `max_length=64`. Detailed-caption tasks like CaReBench lose most of the caption.

**Pooling differs by class.** The video pooler returns `(batch, 1, hidden)` vs text's `(batch, hidden)`, so video is squeezed. Vision-only checkpoints have no pooler and mean over `last_hidden_state` like vjepa2, discarding temporal order, so Diving48 looks like the task not the wrapper.

### DIFF
--- a/mteb/models/model_implementations/videoprism_models.py
+++ b/mteb/models/model_implementations/videoprism_models.py
@@ -1,0 +1,311 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import torch
+from tqdm.auto import tqdm
+
+from mteb.models.abs_encoder import AbsEncoder
+from mteb.models.modality_collators import FramesCollator
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+
+# The LVT text tower has exactly 64 learned position embeddings and adds them
+# elementwise, so a longer sequence raises a shape mismatch instead of being
+# truncated internally. Every text batch is padded and truncated to this length.
+MAX_TEXT_TOKENS = 64
+
+
+def _resolve_device(device: str | int | torch.device | None) -> torch.device:
+    if device is not None:
+        # mteb's CLI may pass an int index, torch.device normalises it
+        return torch.device(device)
+    if torch.cuda.is_available():
+        return torch.device("cuda")
+    if torch.backends.mps.is_available():
+        return torch.device("mps")
+    return torch.device("cpu")
+
+
+class VideoPrismVisionWrapper(AbsEncoder):
+    """VideoPrism video-only encoders (Base, Large).
+
+    These checkpoints have no text tower, so they support video clustering and
+    classification only, the same shape as the vjepa2 checkpoints.
+    """
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None,
+        device: str | int | torch.device | None = None,
+        num_frames: int = 16,
+        **kwargs: Any,
+    ) -> None:
+        from transformers import AutoVideoProcessor, VideoPrismVisionModel
+
+        self.model_name = model_name
+        self.num_frames = num_frames
+        self.device = _resolve_device(device)
+        self.dtype = torch.bfloat16 if self.device.type == "cuda" else torch.float32
+        self.processor = AutoVideoProcessor.from_pretrained(
+            model_name, revision=revision
+        )
+        self.model = VideoPrismVisionModel.from_pretrained(
+            model_name, revision=revision, dtype=self.dtype
+        ).to(self.device)
+        self.model.eval()
+
+    @torch.inference_mode()
+    def encode(
+        self,
+        inputs: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        show_progress_bar: bool = True,
+        **kwargs: Any,
+    ) -> Array:
+        inputs.collate_fn = FramesCollator(num_frames=self.num_frames)
+
+        all_embeddings = []
+        for batch in tqdm(inputs, disable=not show_progress_bar, desc="Video Encoding"):
+            # Source clips vary in resolution, so they cannot be stacked before
+            # preprocessing. The processor resizes each clip to 288x288, after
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/videoprism_models.py
```diff
@@ -0,0 +1,311 @@
...
+        from transformers import AutoVideoProcessor, VideoPrismVisionModel
+
+        self.model_name = model_name
+        self.num_frames = num_frames
+        self.device = _resolve_device(device)
+        self.dtype = torch.bfloat16 if self.device.type == "cuda" else torch.float32
+        self.processor = AutoVideoProcessor.from_pretrained(
+            model_name, revision=revision
+        )
+        self.model = VideoPrismVisionModel.from_pretrained(
```
Samoed, line 65: Were did you find examples with usage transformers? All I can find examples with cloning google's repo

hubielu: The transformers docs page: https://huggingface.co/docs/transformers/en/model_doc/videoprism

It was contributed by @MHRDYN7 and reviewed by @vasqu and @zucchini-nlp, landed 2026-06-19, and documents `VideoPrismClipModel`, `VideoPrismVisionModel`, `VideoPrismForVideoClassification` and `VideoPrismProcessor` with usage examples.

I believe the reason you only find clone-google's-repo examples is the same thing I flagged in the description: the HF-format weights aren't on `main` in any of the four repos yet, only the Flax `.npz`. The transformers docs example itself loads with `revision="refs/pr/4"`. So the integration exists but the converted weights are still sitting in Hub PRs, which is why every public example still points at the JAX repo. That's also why the metas here pin commit SHAs from those PR refs.

Samoed: Oh, I see. Can you add comment to this page then?

MHRDYN7: Yeah the google team is lazy, still haven't merged those weights 🤷

mteb/models/model_implementations/videoprism_models.py
```diff
@@ -0,0 +1,311 @@
...
+    These checkpoints have no text tower, so they support video clustering and
+    classification only, the same shape as the vjepa2 checkpoints.
+    """
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None,
+        device: str | int | torch.device | None = None,
+        num_frames: int = 16,
```
Samoed: Typically we're adding extra 
```
    fps: float | None,
    max_frames: int | None,
    num_frames: int | None,
```
args for future experimentation with different sampling techniques

mteb/models/model_implementations/videoprism_models.py
```diff
@@ -0,0 +1,311 @@
...
+            # Source clips vary in resolution, so they cannot be stacked before
+            # preprocessing. The processor resizes each clip to 288x288, after
+            # which they concatenate cleanly.
+            processed = []
+            for video in batch["video"]:
+                clip = (
+                    video.to(torch.uint8)
+                    if isinstance(video, torch.Tensor)
+                    else torch.as_tensor(video, dtype=torch.uint8)
+                )
```
Samoed: I think this always be torch.Tensor since we're using torcodec for video decoding

mteb/models/model_implementations/videoprism_models.py
```diff
@@ -0,0 +1,311 @@
...
+    if torch.backends.mps.is_available():
+        return torch.device("mps")
+    return torch.device("cpu")
+
+
+class VideoPrismVisionWrapper(AbsEncoder):
+    """VideoPrism video-only encoders (Base, Large).
+
+    These checkpoints have no text tower, so they support video clustering and
+    classification only, the same shape as the vjepa2 checkpoints.
```
Samoed:
```suggestion
```

mteb/models/model_implementations/videoprism_models.py
hubielu: Yes, added in the commit above!

Thanks @MHRDYN7 for confirming

Samoed: Yes, this is pretty common behavior for huge orgs. They just release and mostly abandon the repo.

mteb/models/model_implementations/videoprism_models.py
hubielu: Added `fps` and `max_frames`, and `num_frames` is now `int | None`. All three pass through to the collator, same as the vjepa2 wrapper. Kept `num_frames` at 16 for the f16 checkpoints and 8 for the f8 ones in `loader_kwargs`

mteb/models/model_implementations/videoprism_models.py
hubielu: You're right, simplified to `video.to(torch.uint8)`.

Samoed (approved):

PR #5196, opened 2026-08-14
model: add UNITE (friedrichor/Unite-Base-Qwen2-VL-2B)

### LINKED ISSUE #5185: Add model: UNITE (friedrichor/Unite-Base-Qwen2-VL-2B)
Claimed in #4842 (Video row).

**UNITE** is a universal multimodal embedder over text, image, and video, plus combined queries.

- Paper: Modality Curation: Building Universal Embeddings for Advanced Multimodal Information Retrieval, arXiv:2505.19650 (Kong et al., Northeastern University and Kuaishou)
- Code: https://github.com/friedrichor/UNITE
- Weights: https://huggingface.co/collections/friedrichor/unite-682da30c4540abccd3da3a6b

**Details**

Qwen2-VL backbone fine-tuned for retrieval with Modal-Aware Masked Contrastive Learning. Embeddings come from E5-V style prompting ("Summary above sentence in one word:"), last-token pooling, L2 normalized. `friedrichor/Unite-Base-Qwen2-VL-2B` outputs 1536-d. There are both Base and Instruct families in the collection; Base is recommended for fine-grained retrieval and Instruct for instruction-based retrieval.

Covers t2i/i2t, t2v/v2t, and composed image+text queries, so it can run across the image and video parts of the matrix rather than one direction.

**Implementation note**

The custom class lives in `modeling_unite.py` and only subclasses `Qwen2VLForConditionalGeneration`. There is no `auto_map` on the HF repos, so `trust_remote_code` is not an option and the class needs to be vendored into the wrapper. Upstream requirements pin `transformers==4.47.1`, but I have it working on 4.52.4 with a two-line change (`embed_tokens` and the decoder call moved under `model.language_model` in the Qwen2VL refactor), so no version pin should be needed. Weight remapping is handled by the existing `_checkpoint_conversion_mapping`.

Planning to reproduce results on 1-2 benchmarks from the paper (CaReBench or WebVid-CoVR) as the reference run.

hubielu:
Closes #5185. Claimed in #4842 (Video row).

Adds UNITE (arXiv:2505.19650), a universal embedder for text, image, video, and fused queries. Qwen2-VL-2B backbone, E5-V style prompting, last-token pooling, L2 normalization, 1536-d embeddings.

## Reference run

CaReBench, using 32-frame video sampling:

| Task | nDCG@10 | R@1 | R@5 | R@10 |
|---|---:|---:|---:|---:|
| CaReBenchT2VRetrieval | 89.03 | 79.00 | 95.80 | 97.80 |
| CaReBenchV2TRetrieval | 90.13 | 80.30 | 96.50 | 98.60 |

`Unite-Base-Qwen2-VL-2B` @ `850056e7`, bf16, 1x A6000, 32 sampled frames, video capped at `360*420`.

MTEB uses the CaRe-General queries. The paper reports:

- T→V: 78.1 / 95.5 / 97.8
- V→T: 80.8 / 96.4 / 98.7

for R@1/R@5/R@10, so the reproduction is within 0.9 R@1 points and essentially matches R@5/R@10. The earlier gap came from fps-based sampling.

Separately reproduced upstream demo similarities closely: text-image 0.7539 (0.7500 reported), text-video 0.5625 (0.5664 reported).

## Note on loading

UNITE ships `UniteQwen2VL`, a `Qwen2VLForConditionalGeneration` subclass, without an `auto_map`.

On transformers 5.x, the base class loaded the checkpoint correctly, but the UNITE subclass did not receive the same key remapping and loaded only 391/729 weights, producing invalid embeddings.

The final implementation restores UNITE's subclass/forward behavior and uses MTEB's existing `transformers-v4` requirement group.

## Video / batching

Video resizing is applied through the processor with `max_pixels=360*420`. Default sampling is 32 frames, while `fps`, `max_frames`, and `num_frames` remain configurable.

Batch encoding is supported. On 16 CaReBench videos, bs=1 was fastest (3.59 s/item vs. 3.92/4.21/5.16 for bs=2/4/8), so batch size remains configurable.

## Follow-up

Instruct checkpoints can be added separately as additional `ModelMeta` entries.

### DIFF
--- a/mteb/models/model_implementations/unite_models.py
+++ b/mteb/models/model_implementations/unite_models.py
@@ -117,15 +117,16 @@ def _build_unite_model(model_name: str, revision: str | None, **kwargs):
 
 
 class UniteWrapper(AbsEncoder):
-    def __init__(
+    def __init__(  # noqa: PLR0913, PLR0917 - video sampling options are configurable
         self,
         model_name: str,
         revision: str,
         device: str | None = None,
         min_image_tokens: int = 256,
         max_image_tokens: int = 1280,
-        fps: float = 1.0,
-        max_frames: int = 32,
+        fps: float | None = None,
+        max_frames: int | None = None,
+        num_frames: int | None = 32,
         target_sampling_rate: int = 16000,
         video_max_pixels: int = 360 * 420,
         **kwargs: Any,
@@ -135,6 +136,7 @@ class UniteWrapper(AbsEncoder):
         self.device = device or ("cuda" if torch.cuda.is_available() else "cpu")
         self.fps = fps
         self.max_frames = max_frames
+        self.num_frames = num_frames
         self.target_sampling_rate = target_sampling_rate
         self.video_max_pixels = video_max_pixels
 
@@ -235,7 +237,7 @@ class UniteWrapper(AbsEncoder):
                 target_sampling_rate=self.target_sampling_rate,
                 fps=self.fps,
                 max_frames=self.max_frames,
-                num_frames=None,  # fps mode; both set would let num_frames win
+                num_frames=self.num_frames,
             )
 
         all_embeddings = []
--- a/mteb/models/model_implementations/unite_models.py
+++ b/mteb/models/model_implementations/unite_models.py
@@ -46,9 +46,13 @@ def _fit(video, max_pixels: int):
     scale = (max_pixels / (h * w)) ** 0.5
     nh = max(28, int(h * scale) // 28 * 28)
     nw = max(28, int(w * scale) // 28 * 28)
-    return torch.nn.functional.interpolate(
-        video.float(), size=(nh, nw), mode="bilinear", align_corners=False
-    ).clamp(0, 255).to(torch.uint8)
+    return (
+        torch.nn.functional.interpolate(
+            video.float(), size=(nh, nw), mode="bilinear", align_corners=False
+        )
+        .clamp(0, 255)
+        .to(torch.uint8)
+    )
 
 
 def _unwrap(out):
@@ -91,13 +95,21 @@ def _unite_embed(model, inputs) -> torch.Tensor:
 
     if inputs.get("pixel_values") is not None:
         inputs_embeds = _scatter(
-            model, inputs_embeds, input_ids, inputs["pixel_values"],
-            inputs["image_grid_thw"], model.config.image_token_id,
+            model,
+            inputs_embeds,
+            input_ids,
+            inputs["pixel_values"],
+            inputs["image_grid_thw"],
+            model.config.image_token_id,
         )
     if inputs.get("pixel_values_videos") is not None:
         inputs_embeds = _scatter(
-            model, inputs_embeds, input_ids, inputs["pixel_values_videos"],
-            inputs["video_grid_thw"], model.config.video_token_id,
+            model,
+            inputs_embeds,
+            input_ids,
+            inputs["pixel_values_videos"],
+            inputs["video_grid_thw"],
+            model.config.video_token_id,
         )
 
     out = model.model.language_model(
--- a/mteb/models/model_implementations/unite_models.py
+++ b/mteb/models/model_implementations/unite_models.py
@@ -269,5 +269,4 @@ unite_base_qwen2vl_2b = ModelMeta(
     training_datasets=unite_training_datasets,
     adapted_from="Qwen/Qwen2-VL-2B-Instruct",
     citation=UNITE_CITATION,
-    contacts=["hubielu"],
 )
--- a/mteb/models/model_implementations/unite_models.py
+++ b/mteb/models/model_implementations/unite_models.py
@@ -27,8 +27,10 @@ UNITE_CITATION = """@article{kong2025modality,
 
 # Vendored from https://github.com/friedrichor/UNITE/blob/main/inference/modeling_unite.py
 # The HF repos ship no auto_map, so trust_remote_code is not available.
-# Two lines differ from upstream, both required by the transformers >=4.52
-# Qwen2VL refactor; each is marked inline.
+# UNITE ships modeling_unite.py as a Qwen2VLForConditionalGeneration subclass.
+# On transformers 5.x that subclass misses the checkpoint key remapping and the
+# language model loads randomly initialized without erroring, so we use the base
+# class and reproduce UNITE's pooling in _unite_embed.
 
 
 def _unwrap(out):
--- a/mteb/models/model_implementations/unite_models.py
+++ b/mteb/models/model_implementations/unite_models.py
@@ -1,0 +1,250 @@
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+import torch
+from tqdm.autonotebook import tqdm
+
+from mteb.models.abs_encoder import AbsEncoder
+from mteb.models.modality_collators import VideoCollator
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+from mteb.types import PromptType
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput
+
+logger = logging.getLogger(__name__)
+
+UNITE_CITATION = """@article{kong2025modality,
+  title={Modality Curation: Building Universal Embeddings for Advanced Multimodal Information Retrieval},
+  author={Kong, Fanheng and Zhang, Jingyuan and Liu, Yahui and Zhang, Hongzhi and Feng, Shi and Yang, Xiaocui and Wang, Daling and Tian, Yu and W., Victoria and Zhang, Fuzheng and Zhou, Guorui},
+  journal={arXiv preprint arXiv:2505.19650},
+  year={2025}
+}"""
+
+# Vendored from https://github.com/friedrichor/UNITE/blob/main/inference/modeling_unite.py
+# The HF repos ship no auto_map, so trust_remote_code is not available.
+# Two lines differ from upstream, both required by the transformers >=4.52
+# Qwen2VL refactor; each is marked inline.
+
+
+def _build_unite_model(model_name: str, revision: str | None, **kwargs):
+    from transformers import Qwen2VLForConditionalGeneration
+
+    class UniteQwen2VL(Qwen2VLForConditionalGeneration):
+        def __init__(self, config):
+            super().__init__(config)
+            self.normalize = True
+
+        def forward(
+            self,
+            input_ids=None,
+            attention_mask=None,
+            position_ids=None,
+            past_key_values=None,
+            inputs_embeds=None,
+            pixel_values=None,
+            pixel_values_videos=None,
+            image_grid_thw=None,
+            video_grid_thw=None,
+            pooling_mask=None,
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/unite_models.py
```diff
@@ -0,0 +1,273 @@
...
+    scale = (max_pixels / (h * w)) ** 0.5
+    nh = max(28, int(h * scale) // 28 * 28)
+    nw = max(28, int(w * scale) // 28 * 28)
+    return (
+        torch.nn.functional.interpolate(
+            video.float(), size=(nh, nw), mode="bilinear", align_corners=False
+        )
+        .clamp(0, 255)
+        .to(torch.uint8)
+    )
```
Samoed: I think this will be done in processor https://github.com/huggingface/transformers/blob/60eefbc8141a028609875951f5170086631d9605/src/transformers/models/qwen2_vl/video_processing_qwen2_vl.py#L295-L307

mteb/models/model_implementations/unite_models.py
```diff
@@ -0,0 +1,273 @@
...
+    # language model silently loads with random weights. We keep the base class and
+    # reproduce UNITE's pooling in _unite_embed instead.
+    try:
+        return Qwen2VLForConditionalGeneration.from_pretrained(
+            model_name, revision=revision, dtype=torch.bfloat16
+        )
+    except TypeError:
+        return Qwen2VLForConditionalGeneration.from_pretrained(
+            model_name, revision=revision, torch_dtype=torch.bfloat16
+        )
```
Samoed: How this solves this problem? Maybe would be easier to pin transformers version?

mteb/models/model_implementations/unite_models.py
```diff
@@ -0,0 +1,273 @@
...
+        video_max_pixels: int = 360 * 420,
+        **kwargs: Any,
+    ) -> None:
+        from transformers import AutoProcessor
+
+        self.device = device or ("cuda" if torch.cuda.is_available() else "cpu")
+        self.fps = fps
+        self.max_frames = max_frames
+        self.target_sampling_rate = target_sampling_rate
+        self.video_max_pixels = video_max_pixels
```
Samoed, line 141: Shoudn't this be in processor?

mteb/models/model_implementations/unite_models.py
```diff
@@ -0,0 +1,273 @@
...
+            texts = batch.get("text")
+            images = batch.get("image")
+            videos = batch.get("video")
+            n = len(texts or images or videos)
+            for i in range(n):
+                emb = self._embed_one(
+                    text=texts[i] if texts is not None else None,
+                    image=images[i] if images is not None else None,
+                    video=videos[i] if videos is not None else None,
+                )
```
Samoed: Why your're not batch this?

mteb/models/model_implementations/unite_models.py
```diff
@@ -0,0 +1,273 @@
...
+    reference="https://huggingface.co/friedrichor/Unite-Base-Qwen2-VL-2B",
+    similarity_fn_name=ScoringFunction.COSINE,
+    framework=["PyTorch", "safetensors", "Transformers"],
+    use_instructions=False,
+    public_training_code="https://github.com/friedrichor/UNITE",
+    public_training_data="https://huggingface.co/datasets/friedrichor/Unite-Base-Retrieval-Train",
+    training_datasets=unite_training_datasets,
+    adapted_from="Qwen/Qwen2-VL-2B-Instruct",
+    citation=UNITE_CITATION,
+    contacts=["hubielu"],
```
Samoed: There should be model's authors
```suggestion
```

mteb/models/model_implementations/unite_models.py
hubielu: You're right, this belongs in the processor. I'll drop `_fit` and configure the video processor's size instead

mteb/models/model_implementations/unite_models.py
hubielu: The remapping is applied to the known class, so a subclass doesn't get it. On 5.15.0: base loads 729/729, subclass 391/729 silently, and upstream's demo similarity drops 0.7500 → 0.056.

Pinning is awkward since UNITE wants 4.47.1 and mteb resolves 5.15.0. Base class needs no pin, verified on 4.52.4 and 5.15.0.

The try/except is only for the torch_dtype → dtype rename; happy to drop it if the floor already covers it.

mteb/models/model_implementations/unite_models.py
hubielu: video_max_pixels yes. fps and max_frames go to VideoCollator, not the processor, so those stay unless you'd rather they be loader_kwargs.

mteb/models/model_implementations/unite_models.py
hubielu: No good reason. Will batch and re-run CaReBench to confirm scores hold.

mteb/models/model_implementations/unite_models.py
Samoed: > Pinning is awkward since UNITE wants 4.47.1 and mteb resolves 5.15.0

We don't pin high transformers version. A lot of models pin lower version of transformers, e.g. https://github.com/embeddings-benchmark/mteb/blob/687d5ecad9e2f57d94beeaefd94077c9140a7125/pyproject.toml#L147-L148

mteb/models/model_implementations/unite_models.py
hubielu: Batched! Measured on 16 CaReBench videos (A6000, fps=1, max_frames=32): bs=1 3.59s/item, bs=2 3.92, bs=4 4.21, bs=8 5.16 — clips run 9-32 frames, so padding to batch max costs more than it saves. Left to the CLI.

Re-ran at bs=1, with the video cap now in the processor:

| Task | nDCG@10 | R@1 |
|---|---|---|
| CaReBenchT2VRetrieval | 88.36 | 77.70 |
| CaReBenchV2TRetrieval | 89.62 | 79.40 |

Was 87.41/76.00 and 90.15/80.40. The ~1pt drift is the processor's smart-resize landing on a different grid than the manual resize, same 360*420 budget.

mteb/models/model_implementations/unite_models.py
hubielu: That makes sense. Though I don't think a pin is needed here either way since the base class loads fine on both 4.x and 5.x, so pinning <5 would only be to keep the subclass. Happy to pin instead if you'd prefer that shape.

mteb/models/model_implementations/unite_models.py
```diff
@@ -0,0 +1,266 @@
...
+    # language model silently loads with random weights. We keep the base class and
+    # reproduce UNITE's pooling in _unite_embed instead.
+    try:
+        return Qwen2VLForConditionalGeneration.from_pretrained(
+            model_name, revision=revision, dtype=torch.bfloat16
+        )
+    except TypeError:
+        return Qwen2VLForConditionalGeneration.from_pretrained(
+            model_name, revision=revision, torch_dtype=torch.bfloat16
+        )
```
Samoed:
```suggestion
    return Qwen2VLForConditionalGeneration.from_pretrained(
        model_name, revision=revision, torch_dtype=torch.bfloat16
    )
```

mteb/models/model_implementations/unite_models.py
Samoed: But you said 
> On 5.15.0: base loads 729/729, subclass 391/729 silently, and upstream's demo similarity drops 0.7500 → 0.056.

So transformers can't load this model correctly on v5 version or I miss something?

mteb/models/model_implementations/unite_models.py
Samoed: Seems perforamcne is a bit lower that in paper https://arxiv.org/pdf/2505.19650. Can you investigate this a bit more why this happen? Would be great to reproduce scores

mteb/models/model_implementations/unite_models.py
hubielu: Sorry, that was unclear. transformers loads it fine on v5, the base class gets 729/729. Only the subclass fails, because the key remapping is registered against the known class and doesn't reach a user subclass. So no pin is needed; the wrapper just uses the base class.

mteb/models/model_implementations/unite_models.py
hubielu: The paper's Table 1 seems to report R@1 separately for CaRe-General / Spatial / Temporal, but mteb ships CaReBench as a single task, so I'm not sure which row I should be comparing against. Do you know which one it's meant to match?

My best guess at the gap is frame sampling. I'm using fps=1 to match UNITE's inference demo, but CaReBench's own paper says their experiments use 32 uniformly sampled frames, and since the clips are mostly 5-30s, fps=1 probably gives fewer than that. Could easily be something else though. I'll try num_frames=32 and post both sets of numbers.

Samoed: > CaRe-General / Spatial / Temporal, but mteb ships CaReBench as a single task, so I'm not sure which row I should be comparing against

These are just different queries. We have general implementation

mteb/models/model_implementations/unite_models.py
Samoed: > Only the subclass fails, because the key remapping is registered against the known class and doesn't reach a user subclass

Which subclass fails? Shoudn't we pin then?

hubielu: The failing subclass is UNITE's UniteQwen2VL(Qwen2VLForConditionalGeneration), vendored from their inference/modeling_unite.py. Agreed on constraining the transformers version. I restored that subclass and added MTEB's existing transformers-v4 requirement group in 7b713204, so this model stays off transformers v5 instead of using the base-class workaround.

mteb/models/model_implementations/unite_models.py
hubielu: video_max_pixels is now applied through the processor. fps and max_frames stay on VideoCollator since they control frame sampling rather than processor resizing.

mteb/models/model_implementations/unite_models.py
hubielu: Thanks, that clarifies it. I compared against the CaRe-General row.

I reran CaReBench with 32 uniformly sampled frames. For UNITEbase 2B I get:

T→V: R@1 79.0, R@5 95.8, R@10 97.8
V→T: R@1 80.3, R@5 96.5, R@10 98.6

The paper reports CaRe-General:

T→V: 78.1 / 95.5 / 97.8
V→T: 80.8 / 96.4 / 98.7

So the reproduction is within 0.9 R@1 points and essentially exact at R@5/R@10. The earlier discrepancy was sensitive to the video sampling setup.

mteb/models/model_implementations/unite_models.py
```diff
@@ -0,0 +1,287 @@
...
+        with torch.inference_mode():
+            return self.model(**inputs)
+
+    def _embed_one(self, text=None, image=None, video=None) -> torch.Tensor:
+        """Single-item convenience wrapper, used by tests."""
+        return self._embed_batch(
+            texts=[text] if text is not None else None,
+            images=[image] if image is not None else None,
+            videos=[video] if video is not None else None,
+        )
```
Samoed: Not used
```suggestion
```

Samoed (approved):

PR #5202, opened 2026-08-14
model: add OmniRet (chuonghm/OmniRet)

### LINKED ISSUE #5201: model: add OmniRet (chuonghm/OmniRet)
Adds `chuonghm/OmniRet` (CVPR 2026), a trained omni-modal retriever over text,
image, audio and video. One normalized 4096-d vector per input, instruction
aware.

### Reproduction: ACM (the paper's own benchmark)

Run via their `omniret.evaluate_acm`. As integer query counts, since the
percentages are dense:

| direction | metric | README | this run |
|---|---|---|---|
| A,T→A (4,251 q) | R@1 / R@5 / R@10 | 290 / 950 / 1396 | 284 / 955 / **1396** |
| A→I (1,292 q) | R@1 / R@5 / R@10 | 98 / 284 / 423 | **98** / 286 / 422 |

Largest deviation is 6 queries out of 4,251. Consistent with near-tie reordering
under bf16, not a preprocessing difference.

### Wrapper is bit-identical to their reference path

Same inputs through `OmniRetEmbedder.process()` and through `encode()`:

| case | cosine | max abs diff |
|---|---|---|
| text | 1.000000 | 0.00e+00 |
| text + image | 1.000000 | 0.00e+00 |
| text + audio | 1.000000 | 0.00e+00 |
| text + video | 1.000000 | 0.00e+00 |

Noise floor (same input twice) is also 1.000000, so these are exact, not within
tolerance.

`NIGHTSI2IRetrieval` (an extended M-BEIR task from the paper) runs end to end
through mteb at batch 8: ndcg@10 0.21673, R@10 0.44340.

### Implementation notes

Calls `encode_raw_media_batch` + `encode_batch` rather than `process()`, which
only accepts file paths. Two container mismatches this required fixing:
`_prepare_qwen_audio` reads `item["audio"]`, so mteb's `{"array": ...}` resolved
to None and silently embedded **zeros**; and `VideoCollator` yields a uint8 NCHW
tensor where `_raw_video_frames` evaluates `video or []`, which raises.

### Caveats

Python 3.10 + CUDA 12.8 + `transformers==5.13.0`, not on PyPI, and its
`torch==2.9.1+cu128` pin cannot resolve from PyPI, so the extra carries only
resolvable deps and the loader raises with git install instructions. torchcodec
also needs `libnppicc.so.12`, which the cu128 torch wheels do not ship — that
hits any video task, not just this model. One media item per input max, video
capped at 8 frames.

hubielu:
Adds `chuonghm/OmniRet` (CVPR 2026), a trained omni-modal retriever over text,
image, audio and video. One normalized 4096-d vector per input. Closes #5201.

### Reproduction: ACM, the paper's own benchmark

Via their `omniret.evaluate_acm`. Integer query counts, since the percentages
are dense:

| direction | R@1 / R@5 / R@10 (README) | this run |
|---|---|---|
| A,T→A (4,251 q) | 290 / 950 / 1396 | 284 / 955 / **1396** |
| A→I (1,292 q) | 98 / 284 / 423 | **98** / 286 / 422 |

Largest deviation is 6 queries out of 4,251, consistent with near-tie reordering
under bf16.

### Wrapper is bit-identical to their reference path

Same inputs through `OmniRetEmbedder.process()` and through `encode()`, cosine
and max abs diff:

| text | image | audio | video |
|---|---|---|---|
| 1.000000 / 0e+00 | 1.000000 / 0e+00 | 1.000000 / 0e+00 | 1.000000 / 0e+00 |

Noise floor (same input twice) is also 1.000000, so exact, not within tolerance.
`NIGHTSI2IRetrieval`, an extended M-BEIR task from the paper, runs end to end at
batch 8: ndcg@10 0.21673, R@10 0.44340.

### Implementation notes

Uses `encode_raw_media_batch` + `encode_batch` rather than `process()`, which
only accepts file paths. Two container mismatches needed handling:
`_prepare_qwen_audio` reads `item["audio"]`, so mteb's `{"array": ...}` resolved
to None and silently embedded zeros; and `VideoCollator` yields a uint8 NCHW
tensor where `_raw_video_frames` evaluates `video or []`, which raises.

### Caveats

Python 3.10, CUDA 12.8, `transformers==5.13.0`, and not on PyPI. Its
`torch==2.9.1+cu128` pin cannot resolve from PyPI, so the extra carries only
resolvable deps and the loader raises with git install instructions. torchcodec
also needs `libnppicc.so.12`, which the cu128 torch wheels don't ship, affecting
any video task. One media item per input max, video capped at 8 frames.

### DIFF
--- a/mteb/models/model_implementations/omniret.py
+++ b/mteb/models/model_implementations/omniret.py
@@ -1,0 +1,220 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import torch
+from tqdm.auto import tqdm
+
+from mteb.models.abs_encoder import AbsEncoder
+from mteb.models.modality_collators import AudioCollator, VideoCollator
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+
+# OmniRet's Qwen-Audio tower expects 16 kHz mono; its whisper-style frontend
+# consumes a 30 second window.
+OMNIRET_SAMPLING_RATE = 16_000
+OMNIRET_MAX_AUDIO_SECONDS = 30
+# omniret.config.OmniRetModelConfig.max_video_frames is 8 and the model
+# re-subsamples above that, so sample 8 at the collator to save decode work.
+OMNIRET_MAX_VIDEO_FRAMES = 8
+
+
+class OmniRetWrapper(AbsEncoder):
+    """Wrapper for OmniRet, an instruction-aware text/image/audio/video retriever.
+
+    Every input maps to one L2-normalized 4096-d vector via Attention Sliced
+    Wasserstein Pooling, so cosine similarity is the scoring function.
+
+    This calls ``OmniRetModel.encode_raw_media_batch`` and ``encode_batch``
+    directly rather than ``OmniRetEmbedder.process``, because ``process`` only
+    accepts filesystem paths or HTTP URLs and would force a temp-file write per
+    item. The two-call path accepts in-memory PIL images, frame lists, and audio
+    arrays, which is what mteb's collators already produce.
+    """
+
+    def __init__(
+        self,
+        model_name: str,
+        revision: str | None = None,
+        device: str | None = None,
+        max_audio_length_seconds: int = OMNIRET_MAX_AUDIO_SECONDS,
+        num_frames: int = OMNIRET_MAX_VIDEO_FRAMES,
+        **kwargs: Any,
+    ) -> None:
+        from huggingface_hub import snapshot_download
+
+        try:
+            from omniret import OmniRetEmbedder
+        except ImportError as error:
+            raise ImportError(
+                "OmniRetWrapper requires the `omniret` package, which is not on PyPI. "
+                "Install it into a Python 3.10 CUDA 12.8 environment with "
+                "`uv pip install --no-deps git+https://github.com/hmchuong/OmniRet` "
+                "after installing torch 2.9.1+cu128 from the pytorch cu128 index."
+            ) from error
+
+        self.device = device or ("cuda" if torch.cuda.is_available() else "cpu")
+        self.num_frames = num_frames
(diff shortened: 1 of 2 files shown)


mteb/models/model_implementations/omniret.py
```diff
@@ -0,0 +1,310 @@
...
+    ) -> None:
+        from huggingface_hub import snapshot_download
+
+        try:
+            from omniret import OmniRetEmbedder
+        except ImportError as error:
+            raise ImportError(
+                "OmniRetWrapper requires the `omniret` package, which is not on PyPI. "
+                "Install it into a Python 3.10 CUDA 12.8 environment with "
+                "`uv pip install --no-deps git+https://github.com/hmchuong/OmniRet` "
```
Samoed, line 58: I don't think we want to add models that dependencies we can't add directly

hubielu: That makes sense. `omnivinci` and `ebind` look like the closest precedent, both merged with their extras commented out in `pyproject.toml`.

I've matched that in 8927b50: the `omniret` extra and its conflicts entry are commented out, and `extra_requirements_groups` is kept as those two have it. The `ImportError` gives the git install command.

Would you rather prefer a different approach, or would you rather not take the model at all given the install constraints?

Samoed: > precedent, both merged with their extras commented out in pyproject.toml

Yes, but turns out that we can't publish with git optional dependencies and we commented out them

hubielu: Good to know, thanks. The extra doesn't need a git dependency though, it can carry just the PyPI-resolvable deps (`transformers`, `peft`, `imageio-ffmpeg`) and leave `omniret` itself to the `ImportError`, which gives the git install command. Nothing to publish then.

It also means `get_model` raises something actionable rather than "Unknown extras group". That's what was passing CI in 2ba77e0e. Want me to restore it that way?

Samoed: > leave omniret itself to the ImportError, which gives the git install command

This is not solution. Model should declare reqired dependencies in pyproject

hubielu: Understood, thanks for clarifying. I’ll close this for now

PR #5205, opened 2026-08-15
fix: Enable B905 and B028 Ruff rules

ayush1298:
Completes `B` (flake8-bugbear) from #2416 by enabling the two rules deferred in #5189. The ignore entries are removed, so `B` Ruff rules are fully enabled. 

| Rule | n | Pattern applied |
|---|---|---|
| `B905` zip-without-explicit-strict | 118 | `strict=True` at every site |
| `B028` no-explicit-stacklevel | 39 | `stacklevel=2` — points warnings at the caller, not at mteb internals |

### Why `strict=True` everywhere: They fall into a few families that are equal-length by construction:

- two columns of the same dataset or group — the largest family
- `zip(*pairs)` transposition (8 sites)
- parallel top-k index/value arrays from the same `topk` call
- embeddings zipped against the inputs they were computed from

Several already assert the invariant by hand - a `ValueError` on length mismatch immediately above the zip, a `titles` list padded to `[None] * len(texts)`, an XOR key derived to exactly `len(encrypted)` bytes. `strict=True` encodes what those guards already state.

### DIFF
--- a/mteb/abstasks/text/bitext_mining.py
+++ b/mteb/abstasks/text/bitext_mining.py
@@ -245,7 +245,10 @@ class AbsTaskBitextMining(AbsTask):
 
         text1_statistics = calculate_text_statistics(sentence1)
         text2_statistics = calculate_text_statistics(sentence2)
-        unique_pairs = len(set(zip(sentence1, sentence2, strict=True)))
+        # Not strict: BUCC (superseded by BUCC.v2) filters `sentence1` down to the
+        # gold pairs while keeping the full `sentence2` corpus, so the two sides are
+        # deliberately ragged and the zip truncates to the shorter one.
+        unique_pairs = len(set(zip(sentence1, sentence2, strict=False)))
 
         return BitextDescriptiveStatistics(
             num_samples=len(sentence1),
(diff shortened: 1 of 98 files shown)


ayush1298:
@Samoed Can you check it?

mteb/abstasks/text/bitext_mining.py
```diff
@@ -155,7 +155,7 @@ def _evaluate_subset(  # type: ignore[override]
         )
         # NOTE: used only by BUCC
         gold = (
-            list(zip(range(len(data_split)), range(len(data_split))))
+            list(zip(range(len(data_split)), range(len(data_split)), strict=True))
```
Samoed, line 158: Can you check how bucc will work here?

Samoed (approved):

mteb/abstasks/text/bitext_mining.py
ayush1298: its giving error, I have changed it to `strict=False` and added comment above it.

KennethEnevoldsen (approved):

Issue #5204, opened 2026-08-15
Add model: Amazon Nova Multimodal Embeddings


hubielu:
**Model:** `amazon.nova-2-multimodal-embeddings-v1:0` (Amazon Bedrock)
**Tech report:** https://assets.amazon.science/ba/f2/d0af272848748a24ba6ba45af3a7/nova-mme-technical-report-14.pdf
**Docs:** https://docs.aws.amazon.com/nova/latest/userguide/nova-embeddings.html

A trained embedder, not a pooled generative model. The tech report describes a contrastive training stage and Matryoshka Representation Learning giving 3072/1024/384/256 output dims. Supports text, image, video and audio in one space, with an `embeddingPurpose` parameter that provides an asymmetric query/document split (`GENERIC_RETRIEVAL` vs `GENERIC_INDEX`).

**Scope:** I'd like to do this in two PRs. The first covers text and image, both of which go through the synchronous `InvokeModel` endpoint. Audio and video are deferred because mteb decodes both before they reach the encoder (audio as a float array, video as a torchcodec frame tensor), so sending them to Bedrock would mean re-encoding decoded media back into a container at a frame rate the wrapper picks. That's a design question worth its own discussion rather than burying it here.

**Status:** wrapper is working. SciFact nDCG@10 = 0.736. MSCOCOT2IRetrieval running now for the reproducibility check against the report's text-to-image evaluation.

**Caveats:** proprietary, no weights, Bedrock API only (us-east-1), same terms as the Marengo slot. Requires the `boto3` extras group, which is currently missing from pyproject and breaks all four existing `bedrock/*` models, fixed separately in #5203.

Relates to #4842.

PR #5203, opened 2026-08-15
fix: add missing boto3 extras group

hubielu:
`mteb.get_model("bedrock/amazon-titan-embed-text-v1")` raises `ValueError: Unknown extras group(s) for mteb: ['boto3']`.

The four `bedrock/*` metas declare `extra_requirements_groups=["boto3"]`, but there is no `boto3` group in `[project.optional-dependencies]`, so `_validate_extras_groups` rejects it. All four are currently unloadable:

- `bedrock/amazon-titan-embed-text-v1`
- `bedrock/amazon-titan-embed-text-v2`
- `bedrock/cohere-embed-english-v3`
- `bedrock/cohere-embed-multilingual-v3`

This adds the missing group. `botocore` comes in transitively, which covers the `from botocore.exceptions import ValidationError` in `BedrockModel._encode_amazon`.

Found while working on a Nova Multimodal Embeddings wrapper for #4842.

PR #5206, opened 2026-08-15
model: add Amazon Nova Multimodal Embeddings (text, image, audio, video)

hubielu:
Adds `bedrock/amazon-nova-2-multimodal-embeddings` (`amazon.nova-2-multimodal-embeddings-v1:0`), a trained multimodal embedder from Amazon, via the Bedrock synchronous `InvokeModel` endpoint. Text, image, audio and video.

Discussed in #5204, part of #4842.

### Results

| Task | Metric | Score |
|---|---|---|
| SciFact | nDCG@10 | 0.7364 |
| MSCOCOT2IRetrieval | nDCG@10 | 0.7309 |
| MSVDT2VRetrieval | nDCG@10 | 0.8462 |

Run at `embeddingDimension=3072`. The tech report evaluates text-to-text on MMTEB and text-to-image on MSCOCO, so those two directions correspond to the paper's evaluation, though not to directly comparable numbers: the report uses Recall on the standard MSCOCO 5k protocol, while MIEB's `MSCOCOT2IRetrieval` scores nDCG@10 over 24,809 captions against a 5,000-image pool.

### Notes

The model exposes an `embeddingPurpose` parameter, so the wrapper maps `prompt_type` onto it: `GENERIC_RETRIEVAL` for queries, `GENERIC_INDEX` for documents, plus `CLASSIFICATION` and `CLUSTERING` for those task types. The asymmetry is worth using; on a query/document probe it widened the relevant-vs-irrelevant cosine margin from +0.359 to +0.425.

Images are sent as base64 JPEG. `png` and `webp` are also accepted, but JPEG keeps the payload small on photographic corpora.

Audio arrives as a float array plus sampling rate, so it's encoded to WAV via the stdlib `wave` module. Lossless and no extra dependency.

### Video

Frames are re-encoded to mp4 at the source frame rate (`metadata.average_fps`) and sent inline as base64. Pass-through of the source container isn't possible: Nova validates the container against the declared `format`, and MSVD ships AVI which isn't in the accepted enum (`mp4`, `mov`, `mkv`, `webm`, `flv`, `mpeg`, `mpg`, `wmv`).

Uses `AUDIO_VIDEO_COMBINED`, which returns one vector per item. `AUDIO_VIDEO_SEPARATE` returns one vector per stream and would break the one-embedding-per-item contract.

Two decode guards: `metadata.num_frames` from the container header can overshoot what actually decodes, so trailing frames are dropped until the read succeeds (same approach as `FramesCollator`), and frames are cropped to even dimensions since h264 requires them. Segments are capped at 30s, Nova's limit.

### Dependency

Requires the `boto3` extras group, added in #5203, plus `video` for the torchcodec decode path.


mteb/models/model_implementations/nova_multimodal_models.py
Samoed, line 1: I think better to move to `bedrock_models` file and maybe reuse some functions from it

hubielu: Moved into `bedrock_models.py`. Shared helpers: `get_bedrock_runtime_client()`, `CHARS_PER_TOKEN`, `read_response_body()`.

Kept Nova as its own class rather than a third `provider` branch, since `BedrockModel.encode` is text-only and Nova's request/response shapes differ. Can fold it in further if you'd prefer.

Titan still 1536-d after the refactor.

mteb/models/model_implementations/bedrock_models.py
```diff
@@ -280,3 +303,161 @@ def _to_numpy(self, embedding_response) -> Array:
...
+        inputs: DataLoader[BatchedInput],
+        *,
+        task_metadata: TaskMetadata,
+        hf_split: str,
+        hf_subset: str,
+        prompt_type: PromptType | None = None,
+        **kwargs: Any,
+    ) -> Array:
+        purpose = self._purpose(task_metadata, prompt_type)
+        show_progress_bar = kwargs.pop("show_progress_bar", False)
```
Samoed: You can move directly to args

mteb/models/model_implementations/bedrock_models.py
```diff
@@ -280,3 +303,161 @@ def _to_numpy(self, embedding_response) -> Array:
...
+    memory_usage_mb=None,
+    public_training_code=None,
+    public_training_data=None,
+    training_datasets=None,
+    license=None,
+    reference="https://docs.aws.amazon.com/nova/latest/userguide/nova-embeddings.html",
+    similarity_fn_name=ScoringFunction.COSINE,
+    framework=["API"],
+    use_instructions=False,
+    modalities=["text", "image"],
```
Samoed: This model also support image and video

mteb/models/model_implementations/bedrock_models.py
hubielu: Done

mteb/models/model_implementations/bedrock_models.py
hubielu: Added audio, `modalities` is now `["text", "image", "audio"]`. Verified against Bedrock.

I left video out though as the sync endpoint rejects inline `bytes` for video and requires an S3 source, and mteb hands the encoder a frame tensor with no fps.  What do you think?

hubielu: Correction on the video part above: inline bytes is fine, I misread an earlier validation error. `embeddingMode` accepts `AUDIO_VIDEO_COMBINED` and `AUDIO_VIDEO_SEPARATE`, and with a valid mode the schema asks for `format` rather than rejecting `bytes`. `s3Location` is the one it refuses.

The real blocker is on our side: mteb decodes video to a frame tensor before the encoder sees it, with no fps and no audio, so I'd have to re-encode to a container at an arbitrary frame rate and drop the audio stream, which also makes `AUDIO_VIDEO_*` meaningless.

Samoed: > with no fps

You can set it in collator or you can not use collator, then you'd recive `VideoEncoder` object with all info 

> and no audio

Not all videos have audio stream. We're limited on datasets side, because this is not possible to extract audio from video easlity. On datasets with both `audio` and `video` you can combine them to get audio with video

hubielu: Video's in and Pr description is updated. `MSVDT2VRetrieval` nDCG@10 = 0.846 over 660 videos

Re-encodes frames to mp4 at `metadata.average_fps` rather than passing the source through, since Nova validates the container against the declared `format` and AVI isn't in the enum. Using `AUDIO_VIDEO_COMBINED` to keep one vector per item.

Samoed (approved):

Issue #5157, opened 2026-08-12
Korean BM25 baseline results are not reproducible from their recorded metadata


sixijsu77-hub:
### Describe the bug

The published `mteb/baseline-bm25s` scores for the Korean retrieval tasks cannot be
reproduced using the configuration their result files record.

To be clear: I am not claiming the published numbers are wrong. Every one I tried
reproduces. The problem is that a result file does not identify the configuration that
produced it, and for one task the recorded `mteb_version` could not have produced its own
score.

**Expected:** running the baseline at the `mteb_version` recorded in a result file
reproduces that file's score.

**Actual:** four Korean tasks, reproduced against `results/mteb__baseline-bm25s`:

| Task | recorded version | Published | Measured | Diff | Configuration that reproduces it |
|---|---|---|---|---|---|
| `MIRACLRetrieval` ko (dev) | 2.12.7 | 0.24521 | 0.24521 | **0.00000** | word split, English stopwords, English stemmer |
| `AutoRAGRetrieval` (test) | 2.12.30 | 0.65022 | 0.65022 | **0.00000** | character unigram **+ frequency stopwords at 0.9** |
| `Ko-StrategyQA` (dev) | 2.10.8 | 0.37808 | 0.37807 | −0.00001 | word split, English stopwords, English stemmer |
| `MIRACLReranking` ko (dev) | 2.12.7 | 0.33380 | 0.33979 | +0.00599 | word split, English stopwords, English stemmer — *remainder not accounted for, see below* |

Two configurations, and only one of them is consistent with the version its file records.

**`AutoRAGRetrieval` is the inconsistent one.** At 2.12.30 the BM25 model has no
`_ISO3_TO_LANG` table, no `_unicode_tokenize` function and no `freq_threshold` — none of
the three things its published score requires. All three were introduced in a single
commit, `4012ed57` (#4405), first released in **2.14.2**; Korean was routed to character
tokenization later still, in `db49990f` (#4870), first released in **2.18.12**. The result
file also sits in the `0_3_0` folder, and that revision string arrived with the same 2.14.2
commit.

**A second configuration element is recorded nowhere.** `BM25Tokenizer` removes every token
present in at least `freq_threshold` of the corpus (default 0.9) whenever `_ISO3_TO_LANG`
assigns the language no stopword key, which is the case for Korean. Character unigrams
alone give 0.64342 on `AutoRAGRetrieval`; with that step the difference against the
[...]

PR #5209, opened 2026-08-15
record the resolved BM25 search configuration in retrieval results

sixijsu77-hub:
Follows up on #5157: published `mteb/baseline-bm25s` results can't be traced back to the
configuration that produced them.

`BM25Tokenizer` resolves stopwords, stemmer and tokenizer from the task's language at index
time, and those defaults have changed across releases, so `mteb_version` doesn't recover
them. The values already exist — computed in `__init__`, formatted into a log line no
result file keeps. This writes them into the result instead.

`AutoRAGRetrieval` on `mteb/baseline-bm25s`, `ndcg_at_10` unchanged at 0.65022:

```json
"search_config": {
  "stopwords_key": null, "stemmer_language": null, "tokenizer": "char",
  "freq_threshold": 0.9, "n_freq_stopwords": 7
}
```

`n_freq_stopwords` was previously only a log line: seven tokens dropped from corpus and
queries for appearing in ≥90% of documents.

One `mteb_version` covers many resolutions, since they follow the task's language:

| language | stopwords_key | stemmer_language | tokenizer |
|---|---|---|---|
| `eng` | `en` | `english` | — |
| `tur` | — | `turkish` | — |
| `kor` | — | — | `char` |
| `swa` | — | — | — |

`freq_threshold` only applies where `stopwords_key` is None (`bm25.py:206`, `:254`).

`SearchProtocol` is unchanged and `_resolved_search_config` is duck-typed, so a model
without `resolved_config` gets `None` and nothing is written. `search_config` is merged
only when present, exactly like `previous_results_model_meta`, and rides the existing
`**hf_scores` pass-through in `TaskResult.from_task_results` — no new `TaskResult` field,
and it lands per split and per subset, as it has to.

No score changes. `tests/test_models/test_bm25.py` passes, `ruff check` and
`ruff format --check` are clean, `mypy` reports nothing new.

Only BM25 is wired. If the shape is right the same three lines would let any search model
report what it resolved. Naming and shape are both guesses; happy to change either.

### DIFF
--- a/mteb/_evaluators/retrieval_metrics.py
+++ b/mteb/_evaluators/retrieval_metrics.py
@@ -428,7 +428,7 @@ def robustness_at_10(
     return sum(robustness_scores) / len(robustness_scores)
 
 
-def make_score_dict(
+def make_score_dict(  # noqa: PLR0913
     *,
     ndcg: dict[str, float],
     _map: dict[str, float],
@@ -440,6 +440,7 @@ def make_score_dict(
     hit_rate: dict[str, float],
     task_scores: dict[str, float],
     previous_results_model_meta: dict[str, Any] | None = None,
+    search_config: dict[str, Any] | None = None,
 ) -> dict[str, Any]:
     return {
         **{f"ndcg_at_{k.split('@')[1]}": v for (k, v) in ndcg.items()},
@@ -465,6 +466,7 @@ def make_score_dict(
             if previous_results_model_meta
             else {}
         ),
+        **({"search_config": search_config} if search_config else {}),
     }
 
 
--- a/mteb/abstasks/retrieval.py
+++ b/mteb/abstasks/retrieval.py
@@ -64,6 +64,18 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _resolved_search_config(search_model: Any) -> dict[str, Any] | None:
+    """Settings a search model resolved for this task, if it reports any.
+
+    Optional and duck-typed: a model that does not implement `resolved_config` is
+    unaffected, and `SearchProtocol` is unchanged.
+    """
+    fn = getattr(search_model, "resolved_config", None)
+    if not callable(fn):
+        return None
+    return fn() or None
+
+
 def _filter_queries_without_positives(
     relevant_docs: RelevantDocumentsType, queries: QueryDatasetType
 ) -> tuple[RelevantDocumentsType, QueryDatasetType]:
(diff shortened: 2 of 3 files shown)


mteb/_evaluators/retrieval_metrics.py
```diff
@@ -465,6 +466,7 @@ def make_score_dict(
             if previous_results_model_meta
             else {}
         ),
+        **({"search_config": search_config} if search_config else {}),
```
Samoed, line 469: This is not standart field and shoudn't be in results file. You have mteb version. This is not most clear variant for saving info, but we don't want to add such values to reusults

Samoed (changes requested):

mteb/_evaluators/retrieval_metrics.py
sixijsu77-hub: Understood — the schema call is yours, and I won't argue the field into the result file.

One correction on `mteb_version`, checkable without running anything. The `0_3_0` revision
string enters `bm25.py` in `4012ed57` (#4405), first released in 2.14.2. **29 of the 80
results published under `mteb__baseline-bm25s/0_3_0/` record an older `mteb_version`** —
2.3.0 (×22), 2.12.30 (×4), 2.12.7 (×2), 2.2.0 (×1). A re-homed batch explains it, so this
isn't anyone's mistake, but the field can't be the thing that recovers the configuration.

A better place for it than the result file: `run_settings.jsonl`. `save_task_result` already
writes it beside the results (`result_cache.py:413-433`) with `version` and `encode_kwargs`,
keyed per (task, split, subset) — the granularity this needs, since `BM25Tokenizer` resolves
from the task's language. It's incomplete today (30 of those 80 files have no row), but
where it covers a task it agrees with the result file.

If `search_config` there is acceptable I'll rewrite the PR that way — smaller diff,
`make_score_dict` and `retrieval.py` untouched. If not, I'll close this.

sixijsu77-hub:
Replied in the thread on `retrieval_metrics.py` rather than here, so briefly at top level:
I'm not arguing the field into the result file. The proposal there is to write it to
`run_settings.jsonl` instead — `save_task_result` already writes that beside the results,
and the file is keyed per (task, split, subset).

Either answer is fine. Say so and I'll rewrite the PR that way, or close it.

isaac-chung:
@sixijsu77-hub  Apologies for the delay. We've been quite swamped by the number of new dataset/model additions lately. 

Run settings would indeed be a better option. Thanks for working on this!

Samoed:
I don't think we need to add anything. We already have revision and mteb version. Run settings not for model paramets, but for environment version. Model parameters should be in `model_meta`

isaac-chung:
Run settings has encode_kwargs, which I believe the best place for runtime configs that can be changed. 
Unless you are suggesting that we create a ModelMeta for each combination of the bm25 config parameters?

Samoed:
It's already support this. When you pass different `encode_kwargs` new experiment folder will be created

isaac-chung:
@Samoed what is your suggestion here? Should we close this PR then?

Samoed:
Yes, I think we can close pr

PR #5210, opened 2026-08-15
Add source-aligned v2 metrics for M-BEIR tasks

lxer19:
## Summary
- Add v2 versions for all 16 M-BEIR-derived retrieval tasks.
- Preserve existing v1 tasks and historical leaderboard results.
- Align metrics with the M-BEIR source evaluation protocol:
  - 13 tasks use `hit_rate_at_5`
  - 3 Fashion tasks use `hit_rate_at_10`
  - following the paper https://arxiv.org/pdf/2311.17136

## Implementation
- v1 tasks keep their existing `main_score=ndcg_at_10`.
- v2 tasks reuse the same dataset paths/revisions and data.
- v2 tasks use source-aligned hit-style metrics.
- No evaluator, qrels, corpus, or dataset changes.

## Validation
- `test_metadata.py` passed
- `ruff check` passed
- `ruff format` passed
- v1/v2 metadata consistency verified


Samoed (changes requested):
I understand that it would be great to integrate more benchmarks like MBEIR and UniIR, but we can't change metrics in existing tasks which are part of existing benchmarks. Also, there can be some deviations in data and tasks that are not 1-to-1 equal to those used in MBEIR and UniIR (at least our tasks not equal tasks in MMEB, so there can be also some deviations). You can create an issue for the integration of these. 

As precedents for LOTTE we didn't have any of results and this tasks wasn't included in any of benchmarks. As for `cv_recall` -> `hit_rate` is just renaming of same metric to be more clear what it mean.

lxer19:
> I understand that it would be great to integrate more benchmarks like MBEIR and UniIR, but we can't change metrics in existing tasks which are part of existing benchmarks. Also, there can be some deviations in data and tasks that are not 1-to-1 equal to those used in MBEIR and UniIR (at least our tasks not equal tasks in MMEB, so there can be also some deviations). You can create an issue for the integration of these

I try to do something similar to https://github.com/embeddings-benchmark/mteb/pull/4128. What's your suggested way to do this? I will create an issue to discuss this.

Samoed:
I updated my comment about precedents. For now, I don't see a way, except for creating new tasks as copy of them

lxer19:
> I understand that it would be great to integrate more benchmarks like MBEIR and UniIR, but we can't change metrics in existing tasks which are part of existing benchmarks. Also, there can be some deviations in data and tasks that are not 1-to-1 equal to those used in MBEIR and UniIR (at least our tasks not equal tasks in MMEB, so there can be also some deviations). You can create an issue for the integration of these.
> 
> As precedents for LOTTE we didn't have any of results and this tasks wasn't included in any of benchmarks. As for `cv_recall` -> `hit_rate` is just renaming of same metric to be more clear what it mean.

when I tried to reproduce the results as suggested in https://github.com/embeddings-benchmark/mteb/pull/5165, I saw the metric is not aligned. That's why i created this PR.

lxer19:
@Samoed Thanks for your feedback! I have created an issue https://github.com/embeddings-benchmark/mteb/issues/5214

Samoed:
I don't think that we want to migrage only these 2 datasets, because they still reqire https://github.com/embeddings-benchmark/mteb/pull/5210 to be exact as in mbeir

lxer19:
**Paper reproduction with UniIR's own evaluator** (`mbeir_retriever.compute_recall_at_k`, hit-style)

Two full end-to-end reproductions using the trained UniIR CLIP_SF checkpoint (`TIGER-Lab/UniIR/checkpoint/CLIP_SF/clip_sf_large.pth`) and UniIR's genuine `compute_recall_at_k` — one per metric path:

| Task (M-BEIR) | Query representation | R@1 | R@5 | R@10 | Paper (UniIR CLIP_SF, M-BEIR<sub>local</sub>) |
| -- | -- | -- | -- | -- | -- |
| InfoSeek `q_i,q_t→c_t` (task6) | image + instruction-text (raw-add fusion) | 0.1365 | **0.2744** | 0.3522 | R@5 = 0.279 ✓ |
| Fashion200K `q_i→c_t` (task3, I2T) | image + instruction-text (raw-add fusion) | 0.0481 | 0.1264 | **0.1808** | R@10 = 0.183 ✓ |

- **InfoSeek** → primary metric **R@5** → `hit_rate_at_5`. Full test set: 11,323 queries / 611,651-doc corpus.
- **Fashion200K I2T** → primary metric **R@10** → `hit_rate_at_10` (Fashion200K/FashionIQ are reported at R@10 in the paper). Full test set: 4,889 queries / 61,707-doc corpus. Mean R@10 over the 4 instruction variants UniIR samples = 0.1814 (range 0.1806–0.1820).

**MTEB metric equivalence:** feeding the identical rankings + qrels to MTEB's `hit_rate` (pytrec_eval `success_k`) reproduces UniIR's `compute_recall_at_k` **exactly, per query** — 0 mismatches (InfoSeek 0/11,323 at k=5 and k=10; Fashion200K 0/4,889 at k=5 and k=10).

Together these validate **both** v2 metric paths (`hit_rate_at_5` and `hit_rate_at_10`) end-to-end against the paper. This is a representative validation of the metric protocol on the two paths, not a reproduction of all 16 M-BEIR tasks.

lxer19:
> I don't think that we want to migrage only these 2 datasets, because they still reqire #5210 to be exact as in mbeir

I have migrate all datasets related to MBEIR @Samoed

isaac-chung:
@Samoed it seems like we haven't got a benchmark for M-BEIR yet. Should we introduce it in this PR as well?

Other than that, @lxer19 I think all that's left is to resolve the merge conflict.

Samoed:
I think we can add benchmark object in this PR, but we need to reproduce results firstly. Also https://github.com/embeddings-benchmark/mteb/pull/5165 should be done here if we want to integrate mbeir tasks in this PR

KennethEnevoldsen:
Hmm, to me, this feels more like benchmarks need the option to use a different than the main when the fetch a score.

@Samoed would this be possible in the current leaderboard structure?

Samoed:
We can add to leaderboard, but main blocker is that we don't have such option in library code

lxer19:
That makes sense. I'll hold off on further changes until we decide on the preferred approach. If benchmark-specific metric selection is the preferred direction, would you like me to adapt #5210 to add that library support, or should that be a separate PR? Also, since #5165 already contains the OVEN/InfoSeek candidate-pool fix, should that land separately first, or should those changes be folded into #5210? @Samoed @isaac-chung @KennethEnevoldsen

KennethEnevoldsen:
Focusing on the use of a different metric on the leaderboard for now: I think we need to outline in an issue how to best add this (here we can also discuss potential issues that might come up).

Let me make an issue on it

lxer19:
Not stale — this is currently blocked on the benchmark-specific metric design in #5349. I’ve posted a concrete proposal there; if that direction is accepted, I’ll update this PR accordingly and the v2 task copies should no longer be needed.

PR #5217, opened 2026-08-16
model: add EarthSpeciesProject/esp-aves2-naturelm-audio-v1-beats

yaswanth169:
Checklist:

Mock-run results (`mteb mock-run -m EarthSpeciesProject/esp-aves2-naturelm-audio-v1-beats`):

| Task | Modality | Pass | Reason |
| --- | --- | --- | --- |
| MockAudioClusteringTask | audio | ✓ | - |
| MockAudioMultilabelClassification | audio | ✓ | - |
| MockAny2AnyRetrievalA2A | audio | ✓ | - |
| MockAudioReranking | audio | ✓ | - |
| MockAudioClassification | audio | ✓ | - |
| MockAudioClassificationCrossVal | audio | ✓ | - |
| AbsTaskAudioPairClassification | audio | ✓ | - |

Summary: 7/7 audio checks passed. Text, image, and video correctly skipped (this model is audio-only).


yaswanth169:
Hello @Samoed  and @AdnanElAssadi56 , can you look into this PR.

mteb/models/model_implementations/naturelm_audio_models.py
```diff
@@ -0,0 +1,150 @@
...
+        with torch.no_grad():
+            for batch in tqdm(
+                inputs, disable=not show_progress_bar, desc="Encoding audio"
+            ):
+                max_samples = int(self.max_audio_length_seconds * self.sampling_rate)
+                arrays = [
+                    (item["array"] if isinstance(item, dict) else item)[:max_samples]
+                    for item in batch["audio"]
+                ]
+
```
Samoed, line 103: This will be done in audio collator
```suggestion
```

mteb/models/model_implementations/naturelm_audio_models.py
```diff
@@ -0,0 +1,150 @@
...
+
+naturelm_audio_beats = ModelMeta(
+    loader=NatureLMAudioBEATsWrapper,
+    name="EarthSpeciesProject/esp-aves2-naturelm-audio-v1-beats",
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="03f84874dacdbab04d67350fd8c0e5ee1dc04086",
+    release_date="2025-08-15",
+    max_tokens=float("inf"),
```
Samoed, line 134: Why max_tokens inf?

Samoed:
> - [X] I reproduced results from the original paper (if applicable) on at least one benchmark, and I am including the results in the PR description.

Can you share results that you reproduced?

mteb/models/model_implementations/naturelm_audio_models.py
```diff
@@ -0,0 +1,150 @@
...
+    embedding, verified directly), so same-length clips must be grouped
+    and encoded separately, same issue as Dasheng's wrapper.
+
+    BEATs' attention/embedding memory scales with sequence length, and
+    some real-world field recordings (e.g. in BirdCLEF) run many minutes
+    long -- one such clip triggered a 24GB single-allocation request on
+    real GPU hardware, more than any single GPU here has. Clips are
+    truncated to max_audio_length_seconds to keep memory bounded, same
+    pattern as Qwen2AudioWrapper's max_audio_length_seconds.
+    """
```
AdnanElAssadi, line 48: too verbose

mteb/models/model_implementations/naturelm_audio_models.py
```diff
@@ -0,0 +1,150 @@
...
+            filename="esp-aves2-naturelm-audio-v1-beats.safetensors",
+            revision=revision,
+        )
+        self.model = load_model(
+            "esp_aves2_naturelm_audio_v1_beats",
+            checkpoint_path=checkpoint_path,
+            return_features_only=True,
+            device=self.device,
+        )
+        self.model.register_hooks_for_layers([0, -1])
```
AdnanElAssadi, line 78: why is this needed?

mteb/models/model_implementations/naturelm_audio_models.py
```diff
@@ -0,0 +1,150 @@
...
+                for indices in groups.values():
+                    audio_tensor = torch.stack(
+                        [
+                            torch.as_tensor(arrays[i], dtype=torch.float32)
+                            for i in indices
+                        ]
+                    ).to(self.device)
+                    embeddings = self.model.extract_embeddings(
+                        audio_tensor, aggregation="mean"
+                    )
```
AdnanElAssadi, line 118: why not:
`embeddings = model(audio_tensor)`

mteb/models/model_implementations/naturelm_audio_models.py
```diff
@@ -0,0 +1,150 @@
...
+        revision: str | None = None,
+        device: str | None = None,
+        max_audio_length_seconds: float = 30.0,
+        **kwargs: Any,
+    ) -> None:
+        from avex import load_model
+        from huggingface_hub import hf_hub_download
+
+        self.device = device or ("cuda" if torch.cuda.is_available() else "cpu")
+        self.sampling_rate = 16_000
```
AdnanElAssadi, line 62: did you check this?

mteb/models/model_implementations/naturelm_audio_models.py
```diff
@@ -0,0 +1,150 @@
...
+    name="EarthSpeciesProject/esp-aves2-naturelm-audio-v1-beats",
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="03f84874dacdbab04d67350fd8c0e5ee1dc04086",
+    release_date="2025-08-15",
+    max_tokens=float("inf"),
+    n_parameters=90_717_055,
+    n_embedding_parameters=0,
+    memory_usage_mb=346,
+    embed_dim=1536,
```
AdnanElAssadi, line 138: https://huggingface.co/EarthSpeciesProject/esp-aves2-naturelm-audio-v1-beats
Source shows 768

mteb/models/model_implementations/naturelm_audio_models.py
AdnanElAssadi, line 118: I think padding and then masking would be simpler.

Is that how we are doing things @Samoed ?

Samoed: We can, but padding can decrease performance of models https://github.com/embeddings-benchmark/mteb/pull/5118#discussion_r3749875529

AdnanElAssadi: we should probably have something like this in ci to not accidently have this behavior for models:
encode a 1s clip alone, then batched with a 5s clip, and assert the embeddings are allclose.

Samoed: I don't think that this is possible to do CI because this requires a lot of GPU. All models have this problem, just some of them more or less https://github.com/embeddings-benchmark/mteb/pull/5057#discussion_r3766719125

AdnanElAssadi: why not on cpu? we are just checking one sample.

Samoed: We could have pretty big models. I don't think that we want such checks on CI

AdnanElAssadi: Don't we already load models [here](https://github.com/embeddings-benchmark/mteb/blob/main/tests/test_models/model_loading.py). It also checks if the model is too large and skips in that case. I think only difference is just doing a forward pass.

yaswanth169:
@Samoed @AdnanElAssadi  pushed fixes for everything actionable here.

CI fix: the failures were all the same root cause, avex==1.3.0 requires Python >=3.11,<3.14 but the project allows >=3.10,<3.15, so uv sync was failing outright before any job could even run. Restricted the avex extra with a version marker (same pattern already used for sauerkrautlm-colpali in this file). Also rebased onto main to clear the pyproject.toml conflict and pick up newer lint rules.

embed_dim: fixed, 1536 → 768. Confirmed against the model card, you were right.

Why max_tokens=float("inf") (Samoed): matches the already-merged Dasheng wrapper, same pattern for audio-only encoders with no text tower and no meaningful token-length concept.

Audio truncation duplicating the collator (Samoed): fixed, now passes max_samples into AudioCollator directly and drops the redundant manual slicing.

Docstring too verbose (Adnan): trimmed. max_audio_length_seconds=30.0 is the standard default across roughly 20 other audio wrappers in this codebase, and none of them document it further, so the removed paragraph was more explanation than the norm here.

Why register_hooks_for_layers([0, -1]) is needed (Adnan): checked avex's source directly, extract_embeddings() explicitly raises ValueError("No hooks are registered in the model.") if you skip this. It's a hard precondition, not optional.

Why not just model(audio_tensor) (Adnan): also checked directly. With return_features_only=True (how the model is loaded), plain forward() returns only the raw final-backbone features. extract_embeddings() with hooks on layers [0, -1] deliberately pulls and mean-aggregates both an early and the final transformer layer, a materially different representation, not something model(audio_tensor) gives you for free.

Reproduction results (Samoed): the NatureLM-audio paper only reports numbers for the full pipeline (BEATs + Q-Former + LLM), it never isolates this standalone BEATs encoder, so a literal paper reproduction isn't possible for this component, similar to the HEAR-vs-mteb-protocol gap on the Dasheng PR. What I can show instead: real accuracy on BirdCLEF (species classification, the task this encoder is actually built for), mteb/baseline-random-encoder gets 1.3%, this model gets 18.4%, about 14x random. Genuine, non-trivial signal on a real bioacoustics benchmark, same style of evidence that was accepted for Dasheng.

Device selection line (Adnan): not sure what specifically you were flagging on did you check this?, the pattern matches how every other wrapper in this codebase picks a device. Can you say more about what you want double-checked there?

mteb/models/model_implementations/naturelm_audio_models.py
```diff
@@ -107,9 +112,8 @@ def encode(
                             for i in indices
                         ]
                     ).to(self.device)
-                    embeddings = self.model.extract_embeddings(
-                        audio_tensor, aggregation="mean"
-                    )
+                    features = self.model(audio_tensor)
+                    embeddings = features.mean(dim=1)
```
Samoed, line 116: Match the model card's documented usage:
```
embeddings = model(audio)
embedding = embeddings.mean(dim=1)
```
i.e. temporal mean pool of the final-layer features -> 768-dim. (Earlier revisions hooked layers [0, -1] and concatenated them via extract_embeddings(aggregation="mean"), which mixes the pre-transformer projection with a mid-block activation and yields 1536-dim vectors -- not the published representation.)

Samoed:
Can you rerun model with my fixes?

mteb/models/model_implementations/naturelm_audio_models.py
isaac-chung: https://huggingface.co/EarthSpeciesProject/esp-aves2-naturelm-audio-v1-beats/blob/main/train_config.yaml

isaac-chung:
Mock runs pass.

isaac-chung:
@YashDThapliyal could you run it on BirdSet?

PR #4841, opened 2026-06-20
Wave 7b integration

debashishc:
If you add a model or a dataset, please add the corresponding checklist:

* [dataset checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_dataset/#submit-a-pull-request)
* [model checklist](https://embeddings-benchmark.github.io/mteb/contributing/adding_a_model/#submitting-your-model-as-a-pr)

Close https://github.com/embeddings-benchmark/mteb/issues/4613

PR #5218, opened 2026-08-16
model: add tsinghua-ee/WAVE-7B omni embedding model

yaswanth169:
WAVE (arXiv:2509.21990) produces prompt-aware embeddings for text, audio, silent video, and synchronized audio-visual inputs, via a Qwen2.5-Omni-Thinker backbone with a BEATs audio encoder and an all-layer feature-fusion head.

The prior PR (#4841) stalled because WAVE's checkpoint shipped no modeling code. That's since been resolved upstream (trust_remote_code support merged into the HF repo), so this loads directly with no vendored copy of the training repo.

**Verification**

mteb mock-run, all four modalities pass:

| Modality | Pass |
|---|---|
| text | 41/41 |
| image | 15/15 |
| audio | 10/10 |
| video | 7/7 |

Real reproduction on AudioCapsA2TRetrieval (matches WAVE's paper benchmark "AudioCaps A-RET"), 883 audio queries against a 4411 item text corpus:

| Metric | Score |
|---|---|
| hit_rate@1 | 0.528 |
| hit_rate@5 (main) | 0.836 |
| hit_rate@10 | 0.913 |
| recall@100 | 0.958 |
| MRR@10 | 0.655 |

Not a strict reproduction of the paper's own protocol (different corpus size, different metric), reported as a health check rather than a paper exact number.

**Checklist**
- [x] ModelMeta filled out
- [x] mteb.get_model() and get_model_meta() work
- [x] Tested on a representative set of tasks
- [x] Model is public
- [x] Reproduced paper results, included above

### DIFF
--- a/mteb/models/model_implementations/wave_models.py
+++ b/mteb/models/model_implementations/wave_models.py
@@ -151,8 +151,11 @@ class Wave7BWrapper(AbsEncoder):
         # standalone checkpoint after `from_pretrained`, regardless of what
         # `from_pretrained` already populated it with (see train_qwen.py's
         # `run_test` path). Replicated here for a faithful reproduction.
-        # load_state_dict copies onto whatever device the target params are
-        # already on, so this works fine with the model already placed.
+        # No explicit .to(device) needed: self.model is already placed by
+        # device_map above, and Module.load_state_dict copies each tensor
+        # onto its *target* parameter's existing device (self.model.beats'
+        # params), regardless of what device the source state dict (loaded
+        # here on CPU via map_location="cpu") is on.
         beats_ckpt = torch.load(beats_path, map_location="cpu")
         self.model.beats.load_state_dict(beats_ckpt["model"])
 
@@ -164,6 +167,12 @@ class Wave7BWrapper(AbsEncoder):
         processor = copy.deepcopy(self.processor.image_processor)
         image = image.convert("RGB")
         width, height = image.size
+        # Qwen2VLImageProcessor's own smart_resize already upscales images
+        # below its min_pixels bound, so this rarely fires on real inputs --
+        # kept anyway since it's WAVE's own process_image_unified behavior
+        # (harmless no-op when unneeded, and cheap insurance against a
+        # min_pixels configuration where smart_resize's upscale alone
+        # wouldn't reach the 28px patch-size floor).
         if width < 28 or height < 28:
             pad_width = max(0, 28 - width)
             pad_height = max(0, 28 - height)
@@ -191,16 +200,20 @@ class Wave7BWrapper(AbsEncoder):
         WAVE computes fps = sampled_frames / video_length, which can differ
         from the nominal rate whenever the frame cap binds or a short video
         keeps all its frames, and derives video_second_per_grid from that
-        actual fps. Replicated here using the sampled frame count and the
-        source video's duration.
+        actual fps (qwenvl/data/data_qwen.py's video_decord, line 274 in the
+        WAVE GitHub repo: `fps = len(frame_idx) / video_length`). Replicated
+        here using the sampled frame count and the source video's duration.
         """
         if frames.ndim == 4 and frames.shape[1] == 3 and frames.shape[-1] != 3:
             frames = frames.permute(0, 2, 3, 1)
         frames = frames.contiguous()
-        image_processor = self.processor.image_processor
-        video_proc = image_processor(images=None, videos=frames, return_tensors="pt")
+        video_proc = self.processor.image_processor(
+            images=None, videos=frames, return_tensors="pt"
+        )
         fps = frames.shape[0] / duration if duration else self.fps
-        video_second_per_grid = [image_processor.temporal_patch_size / fps]
+        video_second_per_grid = [
+            self.processor.image_processor.temporal_patch_size / fps
+        ]
         return (
             video_proc["pixel_values_videos"],
             video_proc["video_grid_thw"],
@@ -219,7 +232,6 @@ class Wave7BWrapper(AbsEncoder):
         not part of WAVE's own preprocessing.
         """
         array = np.asarray(array, dtype=np.float32)
-        feature_extractor = self.processor.feature_extractor
         segment_samples = 300 * _SAMPLING_RATE
         segments = [
             array[k : k + segment_samples]
@@ -234,7 +246,7 @@ class Wave7BWrapper(AbsEncoder):
                 segment = np.pad(  # noqa: PLW2901
                     segment, (0, _SAMPLING_RATE - segment.shape[0])
                 )
-            features = feature_extractor(
+            features = self.processor.feature_extractor(
                 segment,
                 sampling_rate=_SAMPLING_RATE,
                 padding="max_length",
@@ -300,6 +312,13 @@ class Wave7BWrapper(AbsEncoder):
         )
         text_input = text_input[0].split("<|im_start|>user\n")[-1].strip()
 
+        # replace_multimodal_special_tokens (processing_qwen2_5_omni.py, bundled
+        # in the checkpoint repo) calls next() on each of these arguments
+        # internally -- it expects iterators, not plain lists/values, hence the
+        # explicit iter() wrapping. position_id_per_seconds=25 and the
+        # seconds_per_chunk formula are copied verbatim from WAVE's own
+        # qwenvl/data/data_qwen.py (process_omni_conversations, lines 449-450
+        # in the WAVE GitHub repo), not derived independently.
         text_input = self.processor.replace_multimodal_special_tokens(
             [text_input],
             iter([audio_length]) if audio_inputs is not None else iter([]),
@@ -465,7 +484,11 @@ wave_7b = ModelMeta(
     release_date="2026-02-11",
     max_tokens=131_072,
     n_parameters=9_410_651_007,
-    n_embedding_parameters=0,
+    # classify_linear, the all-layer fusion head that actually produces the
+    # embedding (two Linear layers: hidden_size*num_hidden_layers -> hidden_size
+    # -> hidden_size, i.e. 3584*28*3584 + 3584 + 3584*3584 + 3584 for WAVE-7B's
+    # config), not zero like a plain last-hidden-state pooling model.
+    n_embedding_parameters=372_513_792,
     memory_usage_mb=17949,
     embed_dim=3584,
     license="apache-2.0",
--- a/mteb/models/model_implementations/wave_models.py
+++ b/mteb/models/model_implementations/wave_models.py
@@ -1,0 +1,484 @@
+from __future__ import annotations
+
+import copy
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import torch
+from tqdm.auto import tqdm
+
+from mteb.models.abs_encoder import AbsEncoder
+from mteb.models.modality_collators import VideoCollator
+from mteb.models.model_meta import ModelMeta, ScoringFunction
+
+if TYPE_CHECKING:
+    from torch.utils.data import DataLoader
+
+    from mteb.abstasks.task_metadata import TaskMetadata
+    from mteb.types import Array, BatchedInput, PromptType
+
+_SAMPLING_RATE = 16_000
+
+# Per-modality prompt used when a task provides no instruction, mirroring the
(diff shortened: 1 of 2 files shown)


yaswanth169:
Hello @AdnanElAssadi56 and @Samoed , can you check this one now.

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+            device_map=self.device,
+        )
+        # WAVE's own eval entrypoint reloads the BEATs submodule from the
+        # standalone checkpoint after `from_pretrained`, regardless of what
+        # `from_pretrained` already populated it with (see train_qwen.py's
+        # `run_test` path). Replicated here for a faithful reproduction.
+        # load_state_dict copies onto whatever device the target params are
+        # already on, so this works fine with the model already placed.
+        beats_ckpt = torch.load(beats_path, map_location="cpu")
+        self.model.beats.load_state_dict(beats_ckpt["model"])
```
Samoed, line 160: Shodn't be `to(device)` here?

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+        width, height = image.size
+        if width < 28 or height < 28:
+            pad_width = max(0, 28 - width)
+            pad_height = max(0, 28 - height)
+            left, top = pad_width // 2, pad_height // 2
+            image = ImageOps.expand(
+                image,
+                border=(left, top, pad_width - left, pad_height - top),
+                fill=(0, 0, 0),
+            )
```
Samoed, line 184: I think processor inside do the same

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+wave_7b = ModelMeta(
+    loader=Wave7BWrapper,
+    name="tsinghua-ee/WAVE-7B",
+    languages=["eng-Latn"],
+    open_weights=True,
+    revision="7d51cdaecfaabb9c529a447249cd4c2a6df8ce5b",
+    release_date="2026-02-11",
+    max_tokens=131_072,
+    n_parameters=9_410_651_007,
+    n_embedding_parameters=0,
```
Samoed: This can't be none for this model

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+        actual fps. Replicated here using the sampled frame count and the
+        source video's duration.
+        """
+        if frames.ndim == 4 and frames.shape[1] == 3 and frames.shape[-1] != 3:
+            frames = frames.permute(0, 2, 3, 1)
+        frames = frames.contiguous()
+        image_processor = self.processor.image_processor
+        video_proc = image_processor(images=None, videos=frames, return_tensors="pt")
+        fps = frames.shape[0] / duration if duration else self.fps
+        video_second_per_grid = [image_processor.temporal_patch_size / fps]
```
Samoed: Why this needed to be done like this? Can you share source of this code?

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+
+        feature_attention_masks = []
+        input_features = []
+        audio_length = 0
+        for segment in segments:
+            if segment.shape[0] < _SAMPLING_RATE:
+                segment = np.pad(  # noqa: PLW2901
+                    segment, (0, _SAMPLING_RATE - segment.shape[0])
+                )
+            features = feature_extractor(
```
Samoed: Why not
```suggestion
            features = self.processor.feature_extractor(
```

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+            video_proc["video_grid_thw"],
+            video_second_per_grid,
+        )
+
+    def _process_audio(
+        self, array: np.ndarray
+    ) -> tuple[dict[str, torch.Tensor], int, np.ndarray]:
+        """Mirror WAVE's process_audio: chunk into 300s segments for BEATs.
+
+        WAVE never truncates audio itself; long clips are chunked into 300s
```
Samoed, line 228: Maybe better just truncate to 300s?

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+        from the nominal rate whenever the frame cap binds or a short video
+        keeps all its frames, and derives video_second_per_grid from that
+        actual fps. Replicated here using the sampled frame count and the
+        source video's duration.
+        """
+        if frames.ndim == 4 and frames.shape[1] == 3 and frames.shape[-1] != 3:
+            frames = frames.permute(0, 2, 3, 1)
+        frames = frames.contiguous()
+        image_processor = self.processor.image_processor
+        video_proc = image_processor(images=None, videos=frames, return_tensors="pt")
```
Samoed:
```suggestion
        video_proc = self.processor.image_processor(videos=frames, return_tensors="pt")
```

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+        text_input = self.processor.apply_chat_template(
+            conversation, add_generation_prompt=False, tokenize=False
+        )
+        text_input = text_input[0].split("<|im_start|>user\n")[-1].strip()
+
+        text_input = self.processor.replace_multimodal_special_tokens(
+            [text_input],
+            iter([audio_length]) if audio_inputs is not None else iter([]),
+            iter([grid_thw]) if grid_thw is not None else iter([]),
+            iter([video_grid_thw[0]]) if video_grid_thw is not None else iter([]),
```
Samoed, line 326: Why this need to be `iter`?

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,484 @@
...
+            video_second_per_grid=iter(second_per_grid_ts)
+            if second_per_grid_ts is not None
+            else iter([]),
+            use_audio_in_video=audio_inputs is not None and video is not None,
+            position_id_per_seconds=25,
+            seconds_per_chunk=2.0 * second_per_grid_ts[0]
+            if second_per_grid_ts is not None
+            else None,
```
Samoed, line 334: Where did you find this?

tomaarsen:
@yaswanth169 Have you also considered e.g. integrating the model with Sentence Transformers in the model repository itself? Then it can be more trivially tested, but also more trivially used. 

- Tom Aarsen

yaswanth169:
@Samoed thanks for the review, addressed everything in the latest commit.

On the device question: no explicit .to(device) needed, self.model is already placed by device_map before this runs, and load_state_dict copies each tensor onto its target parameter's existing device regardless of what device the source checkpoint was loaded on.

On the image padding: you're right that Qwen2VLImageProcessor's own smart_resize already upscales images below its min_pixels bound, so this rarely fires on real inputs. I kept it since it's WAVE's own process_image_unified behavior and it's a harmless no-op when unneeded, cheap insurance for a min_pixels configuration where smart_resize's upscale alone wouldn't reach the 28px patch size floor.

On n_embedding_parameters: fixed, it's classify_linear's real parameter count, 372,513,792 (two Linear layers, hidden_size times num_hidden_layers to hidden_size, then hidden_size to hidden_size).

On the video fps source and the iter() question: both cited directly now. The fps derivation matches qwenvl/data/data_qwen.py's video_decord (line 274 in the WAVE repo), and position_id_per_seconds/seconds_per_chunk match process_omni_conversations in the same file (lines 449 to 450). The iter() wrapping is required because replace_multimodal_special_tokens, bundled in the checkpoint repo's processing_qwen2_5_omni.py, calls next() on those arguments internally.

On truncating audio to 300s instead of chunking: kept the chunking since it matches WAVE's own eval behavior exactly (they never truncate, only chunk and concatenate). In practice this rarely triggers since max_audio_length_seconds already caps most real inputs well under 300s.

On the two suggested diffs: applied both. One caught a real bug along the way, the image_processor(videos=frames) suggestion dropped the images=None argument, which is actually required by BaseImageProcessor.__call__() and crashed immediately on any video input. Fixed to keep images=None.

@tomaarsen thanks for the suggestion. Native Sentence Transformers integration would need changes in WAVE's own HF repo (or a full separate ST wrapper here), and WAVE's dual audio encoder plus all-layer fusion head don't map cleanly onto standard ST architectures. Open to revisiting if the WAVE authors add ST support upstream, but treating it as out of scope for this PR for now.

yaswanth169:
Quick follow-up: re-verified these fixes on real GPU hardware. text, audio, image, and video all encode correctly end to end with proper shapes and normalized embeddings, no issues.

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,507 @@
...
+            if audio_inputs is not None
+            else None,
+            "feature_attention_mask": audio_inputs["feature_attention_mask"]
+            if audio_inputs is not None
+            else None,
+            "input_raw_wav": [torch.from_numpy(raw_wav)]
+            if raw_wav is not None
+            else None,
+            "use_audio_in_video": video_grid_thw is not None
+            and audio_inputs is not None,
```
Samoed, line 364: Not possible to extract audio from our video
```suggestion
```

mteb/models/model_implementations/wave_models.py
```diff
@@ -0,0 +1,507 @@
...
+        }
```
Samoed, line 365: Is this really reqired? Why not just `processor.apply_chat_template`?

Samoed (commented):

tomaarsen:
> thanks for the suggestion. Native Sentence Transformers integration would need changes in WAVE's own HF repo (or a full separate ST wrapper here), and WAVE's dual audio encoder plus all-layer fusion head don't map cleanly onto standard ST architectures. Open to revisiting if the WAVE authors add ST support upstream, but treating it as out of scope for this PR for now.

Or a pull request, and the MTEB configs uses `revision="refs/pr/..."`.

- Tom Aarsen

yaswanth169:
@Samoed ran real reproductions on the paper's own benchmarks instead of just the earlier health check (AudioCapsT2ARetrieval and ClothoT2ARetrieval, both already in mteb).

Results: AudioCaps recall@1 = 30.3% (paper reports R@1 = 44.2%), Clotho recall@1 = 20.7% (paper reports R@1 = 25.6%).
[wave_paper_repro_results.json](https://github.com/user-attachments/files/31947363/wave_paper_repro_results.json)

@tomaarsen thanks, the refs/pr/ pattern is good to know in general. It doesn't quite solve this specific case though: the blocker isn't where the ST config would live, it's that WAVE's dual audio encoders plus all-layer fusion head don't fit Sentence-Transformers' pooling-based single-tower interface. Even with an ST config PR upstream, mteb would still need the same custom encode logic this wrapper already has, so I don't think it reduces the integration work here. Happy to revisit if I'm missing something about how ST handles multi-encoder architectures.

Samoed:
> Or a pull request, and the MTEB configs uses revision="refs/pr/...".

We can use commit hash from pr branch, but this can break after merge I think

> Results: AudioCaps recall@1 = 30.3% (paper reports R@1 = 44.2%), Clotho recall@1 = 20.7% (paper reports R@1 = 25.6%).

This seems very low. Can you investigate what is wrong?

tomaarsen:
> but this can break after merge I think

That is possible, yeah. I'm not sure how that works. I know `refs/pr/...` does stay working, but it is obviously not fixed like a commit hash is.

And the WAVE model looks very custom, but it overall might be compatible with Sentence Transformers' multimodality. It's fairly flexible nowadays. Either way, I'm also fine with having implementations in MTEB, but there's a lot of value in having accessible implementations for each of these models. Otherwise we're back to like 2018 when a lot of research was essentially unusable, poorly accessible, and often stale. 

- Tom Aarsen

yaswanth169:
@Samoed this run used 8-bit quantization, not full precision. No single GPU on the available hardware had the ~19GB free needed for bf16, and a multi-GPU split isn't viable for this model since WAVE's fusion head requires all layers on one device. That's a likely contributor to the gap versus the paper's numbers.

Haven't isolated exactly how much of the drop is quantization versus mteb's protocol differing from the paper's own eval script. Let me know if you'd like a genuine bf16 run to isolate it.

Samoed:
Yes, would be good to see evaluation with full prescision

ayush1298:
@yaswanth169 any update here?

Issue #5208, opened 2026-08-15
Should the leaderboard report rank resolution? Most adjacent ranks aren't statistically separable


ipezygj:
@Muennighoff suggested moving this from email to here — thanks, makes sense.

**Short version.** Treating the MTEB (eng, v2) leaderboard as a paired experiment — every model is scored on the same tasks and per-task scores are published — I tested each adjacent-rank pair with a paired bootstrap over tasks, rather than trusting the printed order. **176 of 180 adjacent pairs are not separable at 5%, and 4 models are statistically tied with the top.**

This isn't a knock on MTEB. 85% of *all* pairs **are** separated, so the benchmark discriminates distant models cleanly. It's the neighbours it can't order — which is exactly the part at the top that people read as a ranking.

The same test gives the same shape on two unrelated boards: SWE-bench (SWE-bench/experiments#466) and HELM Lite (stanford-crfm/helm#4336). LMArena is the positive control, since it already ships bootstrap CIs and gives tied models the same rank. Code, all boards, and a ~200-line tool with self-tests: https://github.com/ipezygj/swebench-rank-audit

**The question, really for the maintainers/authors:** would it make sense for the leaderboard to surface this — e.g. show the group statistically tied for the top, or a rank interval per model — instead of a strict 1..N order? The per-task data to compute it is already published, so it costs nothing.

#5117 is one implementation (a `rank_stability` module), but @KennethEnevoldsen had fair reservations about that specific form, so I'd rather settle the concept here first: is a resolution / tie-group view something MTEB would want, and if so, in what shape? Happy to adapt the implementation to whatever fits the leaderboard.

KennethEnevoldsen:
Hi @ipezygj!

We are def. not taking it as a knock (we already agree that robustness needs to be factored in #5167). 

I would be happy to see uncertainty similar to Arena.ai:

<img width="604" height="610" alt="Image" src="https://github.com/user-attachments/assets/7a042e57-f026-4a00-987a-2b42e5eae323" />

As noted in #5117, it sounds like we can compute our ranks a bootrap across tasks using borda.

 @Samoed rank is computed at the frontend I assume? I am unsure how hard it would be to add the CI (I don't assume we want that computed in the frontend?)

Samoed:
It computed on backend and frontend. I think we need to compute in on frontend to update scores when users filter tasks

ipezygj:
[...]

PR #5233, opened 2026-08-17
docs: measuring rank stability with an across-task bootstrap (Borda)

ipezygj:
The docs example promised in #5117, scoped per the discussion there: across-task bootstrap with Borda ranks, no per-task machinery, no library surface — one advanced-usage page plus a nav line.

The snippet was run as-is against the public results for `MTEB(eng, v2)` and the example output in the page is from that run: 122 models with a complete task set, 5 of them a defensible 1, 120/121 adjacent rank intervals overlapping. The page ends with the two placement options for a leaderboard display (precomputed vs on-filter recompute) for the #5208 discussion.

Closes nothing; complements #5208.

### DIFF
--- a/docs/get_started/advanced_usage/rank_stability.md
+++ b/docs/get_started/advanced_usage/rank_stability.md
@@ -1,0 +1,98 @@
+---
+title: "Rank stability"
+icon: lucide/list-ordered
+---
+
+# Measuring rank stability
+
+A benchmark aggregate is a mean over the tasks that happen to be in the suite, and
+a ranking built from it can move when that suite does. Nothing in a score table
+shows how much of the ordering would survive a different draw of tasks. This page
+shows how to measure that with a task bootstrap: resample the tasks, one draw
+shared by every model so comparisons stay paired, rank models by Borda (mean rank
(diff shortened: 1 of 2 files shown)


Samoed (changes requested):
I don't think that we want to add this as part of documentation, because this is just usage outside of our main scores

ipezygj:
No argument from me. Kenneth pointed at docs in #5117, but if these numbers belong in the ranks rather than beside them, then docs is the wrong home and I would rather do it once in the right place. Happy to close this — say the word, or leave it open as a reference implementation while the placement is decided.

The frontend cost question from #5208 is measurable, so I measured it and put the numbers there.

KennethEnevoldsen:
> I don't think that we want to add this as part of documentation, because this is just usage outside of our main scores

Yeah I think ideally I would have it as a blogpost (but it is not yet supported in zensical), but @ipezygj seems like you are more interested in adding it to the leaderboard if so I think we should simply target that change.

A docs change would more focus on informing about the importance of robustness and rank stability. The current version of this needs a bit more work to be an introductory piece on the importance. I will close for now and focus on the discussing the leaderboard change

PR #5238, opened 2026-08-17
fix: Enable PT ruff rule

ayush1298:
refs https://github.com/embeddings-benchmark/mteb/issues/2416.

Enables the `PT` (flake8-pytest-style) Ruff rule.

`PT018` (composite assertions) is **not** enabled — splitting `assert a and b` is not always clearer, particularly when the assert carries a custom message and the split forces you to duplicate it. The reasonable splits are applied anyway; the rule is in `ignore` so it isn't enforced.

| Rule | n | Fix |
|---|---|---|
| `PT006` names-wrong-type | 27 | `"a, b"` → `("a", "b")` |
| `PT011` raises-too-broad | 10 | Added `match=` to bare `pytest.raises(ValueError)` |
| `PT018` composite-assertion | 22 | Split where it improves the failure output; ignored as a rule |
| `PT001` / `PT003` / `PT007` | 3 | Bare `@pytest.fixture`; dropped implied `scope="function"`; values tuple → list |

`PT011` turned out to be a needed Ruff rule because several of these tests previously passed on any `ValueError`. Two assertions were checking a different error than intended: the `overwrite_strategy` cases hit the "results file exists/missing splits" branch rather than "no results found in cache", and the two MRL cases raise different messages. The `match=` patterns are now branch-specific.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -315,6 +315,7 @@ select = [
     "ASYNC",   # blocking calls in async functions
     "ISC",     # implicit string concatenation
     "B",       # flake8-bugbear
+    "PT",      # flake8-pytest-style
 ]
 
 ignore = [
@@ -391,6 +392,11 @@ preview = true
     "TC002",
 ]
 
+
+[tool.ruff.lint.flake8-pytest-style]
+# The suite already writes `parametrize("a, b", ...)` at every multi-name call
+# site, so enforce that rather than rewriting them all to tuples.
+parametrize-names-type = "csv"
 
 [tool.ruff.lint.pep8-naming]
 ignore-names = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -338,6 +338,7 @@ ignore = [
     "PLR0915",  # too many statements
     "PLW0717",  # too-many-statements-in-try-clause
     "ISC001",   # single-line implicit concat; conflicts with `ruff format`
+    "PT018",    # composite assertions; splitting `assert a and b` is not always clearer
 ]
 future-annotations = true  # For TC rules
 typing-modules = ["typing_extensions", "mteb.types", "collections.abc"]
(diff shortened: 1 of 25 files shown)


tests/test_benchmarks/test_benchmark_score.py
```diff
@@ -359,7 +363,10 @@ def _assert_score_parity(
                 f"{model_name} summary[{summary_col}] should be None, got {sm!r}"
             )
             continue
-        assert gs is not None and sm is not None, (
+        assert gs is not None, (
+            f"{model_name} {gs_key}/{summary_col}: get_score={gs!r} vs summary={sm!r}"
+        )
+        assert sm is not None, (
             f"{model_name} {gs_key}/{summary_col}: get_score={gs!r} vs summary={sm!r}"
```
KennethEnevoldsen: I am not sure this rule is needed

KennethEnevoldsen (commented):
While I agree that the "and" can sometimes make for quite long asserts, I am not sure enforcing it to not be there is better (I would do the fix where reasonable and then not add the rule

ayush1298:
> While I agree that the "and" can sometimes make for quite long asserts, I am not sure enforcing it to not be there is better (I would do the fix where reasonable and then not add the rule

So, should I avoid PT018, which `Split assert a and b into separate asserts`, and keep others?

pyproject.toml
```diff
@@ -392,6 +393,11 @@ preview = true
 ]
 
 
+[tool.ruff.lint.flake8-pytest-style]
+# The suite already writes `parametrize("a, b", ...)` at every multi-name call
+# site, so enforce that rather than rewriting them all to tuples.
+parametrize-names-type = "csv"
```
Samoed: I think we can make tuples

ayush1298: done, removing these

ayush1298:
PT018 is not enforced now, only made changes at relevant places

Samoed (approved):

Issue #5231, opened 2026-08-17
Add total number of parameters to "performance by model size"


KennethEnevoldsen:
### Description of the feature

Currently you only see active, it would be nice to see total

<img width="250" height="119" alt="Image" src="https://github.com/user-attachments/assets/a1f450a5-c5fb-449a-8038-41f92ab3b755" />

PR #5245, opened 2026-08-19
add total parameters to performance size plot

Damadimo:
## Summary

- Add total parameter counts to the performance-by-model-size hover data.
- Keep active parameters as the logarithmic x-axis.
- Add regression coverage for total parameter values and models with unknown totals.

## Validation

- `make lint-check`
- `make typecheck`
- `make test`: 6,713 passed, 238 skipped
#5231


tests/test_leaderboard/test_figures.py
ayush1298, line 1: I dont think we need these test

ayush1298:
Can you run leaderboard locally and add image of plot to verify its coming as expected

Samoed (commented):
This change for old leaderboard that we don't modify much, but I think we still can add, but PR won't close an issue

Damadimo:
Thanks for clarifying! I realized I updated the older Gradio leaderboard rather than the current frontend. I'll take a look at the frontend implementation for issue #5231 and include a screenshot when I follow up.

PR #5247, opened 2026-08-20
dataset: add EDIR

### LINKED ISSUE #5244: Add dataset: EDIR

https://huggingface.co/datasets/EDIR-BENCH/EDIR

### Arxiv link

https://aclanthology.org/2026.acl-long.2144/

EDIR is a composed image retrieval benchmark where a model retrieves an edited target image from a reference image and a natural-language modification instruction. The streamlined public release contains 5,000 queries and a shared corpus of 28,645 edited images across 15 fine-grained edit categories.

whybe-choi:
Closes #5244, Related to ##4842

## Checklist
  - [x] `jinaai/jina-embeddings-v5-omni-nano`
- [x] I have checked that the performance is neither trivial (close to perfect scores) nor random.
- [x] I have considered the size of the dataset and reduced it if it is too big (e.g. 2048 examples for binary classification)
- [x] I reproduced scores from the original paper (if applicable) and added them to the PR description for reference.

## Result

| Model Name | Recall@1 | Recall@5 |
|---|---:|---:|
| jinaai/jina-embeddings-v5-omni-nano | 0.1588 | 0.7558 |

## Note

The public release is intentionally a streamlined version of the corpus used in the paper. The original EDIR evaluation uses a 178,645-image corpus, including a large number of additional distractor images. The released version removes these extra distractors and retains the core database of edited target images, reducing the corpus to 28,645 images.

### DIFF
--- a/mteb/tasks/retrieval/eng/edir_it2i_retrieval.py
+++ b/mteb/tasks/retrieval/eng/edir_it2i_retrieval.py
@@ -1,0 +1,42 @@
+from mteb.abstasks.retrieval import AbsTaskRetrieval
+from mteb.abstasks.task_metadata import TaskMetadata
+
+
+class EDIRIT2IRetrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="EDIRIT2IRetrieval",
+        description=(
+            "EDIR evaluates composed image retrieval: retrieve an edited image from "
+            "a reference image and a textual modification instruction. This task uses "
+            "the public streamlined release, whose corpus removes the paper's extra "
+            "distractor images."
+        ),
+        reference="https://aclanthology.org/2026.acl-long.2144/",
+        dataset={
+            "path": "whybe-choi/EDIRIT2IRetrieval",
+            "revision": "6b371e3b3c37fffeb02bf029fc9beba3515ee7ce",
+        },
+        type="Any2AnyRetrieval",
+        category="it2i",
+        modalities=["image", "text"],
+        eval_splits=["test"],
+        eval_langs=["eng-Latn"],
+        main_score="recall_at_1",
+        date=("2026-01-22", "2026-01-22"),
+        domains=["Scene", "Constructed"],
+        task_subtypes=["Image Text Retrieval"],
+        license="mit",
+        annotations_creators="derived",
+        dialect=[],
+        sample_creation="LM-generated and verified",
+        bibtex_citation=r"""
+@inproceedings{song-etal-2026-rethinking-composed,
+  title = {Rethinking Composed Image Retrieval Evaluation: A Fine-Grained Benchmark from Image Editing},
+  author = {Song, Tingyu and Zhang, Yanzhao and Li, Mingxin and Guo, Zhuoning and Long, Dingkun and Xie, Pengjun and Zhang, Siyue and Zhao, Yilun and Wu, Shu},
+  booktitle = {Proceedings of the 64th Annual Meeting of the Association for Computational Linguistics (Volume 1: Long Papers)},
+  pages = {46224--46242},
+  year = {2026},
+  url = {https://aclanthology.org/2026.acl-long.2144/},
+}
+""",
+    )
--- a/mteb/tasks/retrieval/eng/edir_it2i_retrieval.py
+++ b/mteb/tasks/retrieval/eng/edir_it2i_retrieval.py
@@ -31,12 +31,12 @@ class EDIRIT2IRetrieval(AbsTaskRetrieval):
         sample_creation="LM-generated and verified",
         bibtex_citation=r"""
 @inproceedings{song-etal-2026-rethinking-composed,
-  title = {Rethinking Composed Image Retrieval Evaluation: A Fine-Grained Benchmark from Image Editing},
   author = {Song, Tingyu and Zhang, Yanzhao and Li, Mingxin and Guo, Zhuoning and Long, Dingkun and Xie, Pengjun and Zhang, Siyue and Zhao, Yilun and Wu, Shu},
   booktitle = {Proceedings of the 64th Annual Meeting of the Association for Computational Linguistics (Volume 1: Long Papers)},
   pages = {46224--46242},
-  year = {2026},
+  title = {Rethinking Composed Image Retrieval Evaluation: A Fine-Grained Benchmark from Image Editing},
   url = {https://aclanthology.org/2026.acl-long.2144/},
+  year = {2026},
 }
 """,
     )
--- a/mteb/tasks/retrieval/eng/edir_it2i_retrieval.py
+++ b/mteb/tasks/retrieval/eng/edir_it2i_retrieval.py
@@ -40,7 +40,6 @@ class EDIRIT2IRetrieval(AbsTaskRetrieval):
 }
 """,
         prompt={
-            "query": "Given an image, find a similar everyday image with the described changes: ",
-            "document": "Represent the given image. ",
+            "query": "Given an image, find a similar image satisfying the query.",
         },
     )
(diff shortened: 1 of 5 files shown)


tests/test_tasks/test_task_quality.py
ayush1298, line 1: Is it required to add it? Does it have `duplicate_text` and `duplicate_image?`

ayush1298:
We have `gme-Qwen2-VL-2B-Instruct` integrated into MTEB, and its results are reported in the paper. Can you run an evaluation on it and add a comparison of the reported and obtained results?

tests/test_tasks/test_task_quality.py
whybe-choi: Yes. among the 5,000 queries, there are 4,784 unique reference images and 4,936 unique texts. However, all 5,000 (reference image, text) query pairs are unique, so there are no duplicate query pairs.

whybe-choi:
I encountered an issue while running `gme-Qwen2-VL-2B-Instruct` on this task and opened #5249 to fix it

mteb/tasks/retrieval/eng/edir_it2i_retrieval.py
```diff
@@ -0,0 +1,42 @@
...
+        name="EDIRIT2IRetrieval",
+        description=(
+            "EDIR evaluates composed image retrieval: retrieve an edited image from "
+            "a reference image and a textual modification instruction. This task uses "
+            "the public streamlined release, whose corpus removes the paper's extra "
+            "distractor images."
+        ),
+        reference="https://aclanthology.org/2026.acl-long.2144/",
+        dataset={
+            "path": "whybe-choi/EDIRIT2IRetrieval",
```
Samoed, line 16: Can you add script how you processed dataset?

Samoed (commented):
> The public release is intentionally a streamlined version of the corpus used in the paper. The original EDIR evaluation uses a 178,645-image corpus, including a large number of additional distractor images. The released version removes these extra distractors and retains the core database of edited target images, reducing the corpus to 28,645 images.

I think would be good to still include full version 

You might be intrested @SighingSnow

whybe-choi:
> I think would be good to still include full version

Only the streamlined version is publicly accessible via the official Hugging Face repository ([EDIR-BENCH/EDIR](https://huggingface.co/datasets/EDIR-BENCH/EDIR)), and it seems the full version with the additional distractor images hasn't been made publicly available yet.

SighingSnow:
> I think would be good to still include full version

Hi, thank you for your interest in EDIR.

The full version contains too many images, which consumes a large disk quota (~170 GB). I think the streamlined version is sufficiently challenging, when measured using Recall@1 and Recall@3. 

Additionally, the open-source version can also provide multiple-choice evaluation, which is also very interesting.

Samoed:
@whybe-choi Can you reproduce scores then?

whybe-choi:
I'm currently running the evaluation, but it's taking a bit of time. I'll update the results here.

whybe-choi:
The reproduced GME-2B results show a substantial gap compared with the scores reported in the paper:

| Metric | Paper | Reprod | Difference |
|---|---:|---:|---:|
| Recall@1 | 42.40 | 34.64 | −7.76 pp |
| Recall@3 | 66.10 | 54.36 | −11.74 pp |

The difference is larger than expected, so further investigation is needed to identify the cause.

whybe-choi:
@Samoed 

I also evaluated VLM2Vec-V2.0 and observed a substantial performance gap similar to the one found with GME-2B.

| Model | Metric | Paper | Reprod | Difference |
|---|---|---:|---:|---:|
| GME-2B | Recall@1 | 42.40 | 34.64 | −7.76 pp |
| GME-2B | Recall@3 | 66.10 | 54.36 | −11.74 pp |
| VLM2Vec-2B | Recall@1 | 32.40 | 24.38 | −8.02 pp |
| VLM2Vec-2B | Recall@3 | 61.70 | 47.74 | −13.96 pp |

I initially suspected a dataset conversion issue, but the queries, corpus IDs, qrels, and image files appear consistent with the original release. Although the corpus composition differs from the paper and may naturally lead to some performance variation, the observed gaps still seem unexpectedly large. They may instead stem from differences in inference settings or model implementations. Do you have any thoughts on what might be causing these gaps?

Samoed:
I don't have ideas. Maybe you can try to evaluate from authors repository to try to reproduce?

SighingSnow:
I will try to check the code and current implementation.

Samoed:
Maybe you can try to apply instructions to documents side, maybe this can be a case here?

whybe-choi:
I tested applying instructions to the document side as well, but there was almost no difference in performance.

whybe-choi:
I investigated several factors that could affect the reproduction results, including batch size, numerical precision (`float16` vs. `bfloat16`), and implementation settings. Among these factors, batch size appeared to have the strongest impact.

changing only the batch size produced the following results:

| Batch Size | Recall@1 | Difference vs. Paper |
|---:|---:|---:|
| 16 | 34.64 | -7.76 |
| 1 | 46.08 | +3.68 |
| Paper | 42.40 | — |

Reducing the batch size from 16 to 1 improved Recall@1 by **11.44 percentage points**, bringing the result much closer to the paper performance. I think the reproduced result is higher than the paper result because the streamlined dataset removes additional distractor images from the original benchmark.

SighingSnow:
@whybe-choi hi, the new results is updated at https://huggingface.co/datasets/EDIR-BENCH/EDIR. 
But the batch size should not affect the final results. 

Here is the avg performance of the evaluated models on the streamlit version. I will update the hf official repo of EDIR.

| Model | Recall@1 |
| --- | --- |
| GME-7B | 45.7 | 
| GME-2B | 46.9 |
| BGE-VL-MLLM-S1 | 47.4 | 
| UniME-V2-Qwen2VL-7B | 39.5 | 
| UniME-V2-Qwen2VL-2B | 35.4 |
| EDIR-MLLM-Qwen2-VL-7B-Instruct | 64.8 |

whybe-choi:
@SighingSnow Thank you for sharing the updated results on the streamlined version!

I'll double-check my evaluation setup and implementation settings again. 

While checking [the original repository code](https://github.com/SighingSnow/edir/blob/main/models/qwen2vl_model.py#L29), I noticed the following line during inference:

```python
inputs = self.processor(
    texts=tmp_texts,
    images=tmp_images,
    videos=[None] * len(tmp_texts),
    instruction=[instruction] * len(tmp_texts),
    is_query=False,
    max_length=1024,
    truncation=False,
)
```

Because `is_query` is set to `False`, it looks like the instruction isn't being applied even when encoding queries. Could you clarify if this is the intended behavior?

SighingSnow:
> Because `is_query` is set to `False`, it looks like the instruction isn't being applied even when encoding queries. Could you clarify if this is the intended behavior?

Hi, this is not the intend version. Actually, we apply 'Given an image, find a similar image satisfying the query.' to the query and no instruction is applied to the doc.

whybe-choi:
Thanks for the clarification! I also noticed some differences in other preprocessing parameters between the EDIR repository and the current MTEB implementation:

| Parameter | EDIR Repo | MTEB Implementation |
| :--- | :--- | :--- |
| `max_length` | `1024` | `1800` |
| `min_pixels` | `4 * 32 * 32` | `4 * 28 * 28` |
| `max_pixels` | `1280 * 32 * 32` | `1280 * 28 * 28` |

Could you also share which specific settings were used for your evaluation on the streamlined version?

whybe-choi:
@Samoed 

After revisiting the prompt configuration and evaluation settings, I found that the previous setup was likely incorrect. With the corrected prompt(`Given an image, find a similar image satisfying the query.`) and a batch size of 4, the result is consistent with the expected GME performance.

| Model | Official | Reprod | Diff |
|---|---:|---:|---:|
| GME-Qwen2-VL-2B | 46.90 | 46.48 | -0.42 |

Samoed:
Great! Should we merge then?

whybe-choi:
I think it's ready to be merged

Samoed (approved):

Issue #4941, opened 2026-07-18
Add dataset: MSRVTT-I2V


dukesun99:
### dataset link on Hugging dataset

https://huggingface.co/datasets/Alibaba-NLP/UVRB

### Arxiv link

https://arxiv.org/abs/2510.27571

### Description of the dataset

Image-to-video retrieval built on MSR-VTT, from the UVRB benchmark: the query is a frame held out from a video and the corpus contains 1,000 MSR-VTT videos; 1,000 queries with instance-level ground truth. The reverse video-to-frame direction uses the same assets. A dataset added for #4842: covers the missing image ↔ video combination. Claimed by me in #4842.

dukesun99:
Update after full implementation and evaluation: I built the image↔video tasks from MSRVTT-I2V (1,000 held-out-frame queries vs 1,000 videos, both directions, plus a derived video-frame pair classification) and verified them end to end. Qwen3-VL-Embedding-2B reaches 0.9465 and 0.9284 nDCG@10 on the two retrieval directions and 0.9987 average precision on the pair classification — held-out-frame matching on MSR-VTT is effectively saturated for modern video embedders. I am therefore not proposing these tasks for mteb. The packaged data stays at https://huggingface.co/datasets/dukesun99/MSRVTT-I2V if anyone wants it; the image↔video direction likely needs a harder construction. Suggestions welcome — I'd keep this issue open for that discussion.

PR #5251, opened 2026-08-20
task: add UCF101I2VRetrieval image-to-video retrieval (i2v)

### LINKED ISSUE #5250: Add dataset: UCF101-I2V

https://huggingface.co/datasets/hubxrt/UCF101-I2V (built from https://huggingface.co/datasets/mteb/UCF101-51VA)

### Arxiv link

https://arxiv.org/abs/1212.0402

Image-to-video retrieval covering the image ↔ video direction, which has 0 merged tasks. Built from the UCF101 official split-1 test set.

The held-out-frame construction used by previous i2v attempts saturates — @dukesun99 reported 0.9465 nDCG@10 on MSRVTT-I2V in #4941, and I measured 0.9909 on AVE-Dataset with the same approach. The failure is that the query frame is content of the positive video, so frame-sampling video embedders reduce it to near-duplicate matching.

UCF101 clips are scene-grouped (v_<Class>_g<NN>_c<NN>), which allows a leak-free version: the corpus holds one clip per group, and each query is an interior frame from a non-adjacent sibling clip of the same group. The query frame therefore appears in no corpus video, but scene identity is still recoverable. 337 queries / 337 corpus videos across all 51 classes, one positive each.

Results (via mteb run): ebind-audio-vision 0.9080 nDCG@10, Qwen3-VL-Embedding-2B 0.8687, random baseline 0.0145.

Claimed by me in #4842. PR incoming.

hubielu:
Closes #5250. First task in the image to video direction, which has 0 merged.

**Why a new construction.** Held-out-frame (query is a frame of the positive video) saturates: @dukesun99 measured 0.9465 nDCG@10 on MSRVTT-I2V in #4941 and declined to submit. I got 0.9909 with the same construction on `mteb/AVE-Dataset` against a 0.014 floor, and moving the query frame changed nothing, so it looks structural. Not submitting an AVE i2v task either.

**Construction.** UCF101 clips are scene-grouped (`v_<Class>_g<NN>_c<NN>`), and 100% sit in a multi-clip group (6837 clips / 1274 groups, no singletons), so instance-level qrels over all clips are unanswerable: a frame cannot distinguish its positive from 4 to 6 near-identical siblings. Instead the corpus holds one clip per group (`c01`), and each query is an interior frame (25 to 75% of duration) from a sibling at least 3 clip indices away, since frame-contiguity at group boundaries is adjacency-specific (non-adjacent pairs cross that threshold 1/70 times vs 5/16 adjacent). The query frame therefore appears in no corpus video, while scene identity stays recoverable (+0.418 mean centered correlation over same-class-different-group).

Groups whose query frame is still near-identical to its positive are dropped (raw min RMS < 7): 20 of 357, the largest threshold keeping all 51 classes at 3+ groups. Final size 337 queries / 337 videos, test split only.

**Results**, via `mteb run`:

| model | params | nDCG@10 | R@1 |
|---|---|---|---|
| `encord-team/ebind-audio-vision` | 764M | 0.9080 | 0.8309 |
| `Qwen/Qwen3-VL-Embedding-2B` | 2.1B | 0.8687 | 0.7626 |
| `mteb/baseline-random-encoder` | - | 0.0145 | 0.0000 |

Roughly 60x the floor and well below saturation. Difficulty is graded: R@1 falls 0.977 to 0.400 as query-to-positive frame distance grows.

Dataset is `hubxrt/UCF101-I2V` @ `f252bcab`, happy to move to the mteb org. Query-side prompt only, following `StanfordI2VRetrieval`.


hubielu:
Separate from this PR, but worth flagging before the full benchmark run: all three `encord-team/ebind` models are currently unloadable. `ModelMeta` declares `extra_requirements_groups=["ebind"]`, but that extra is commented out in `pyproject.toml` (L158, L498), so `_validate_extras_groups` raises before any loading. Affects `ebind-audio-vision`, `ebind-points-vision`, `ebind-full`. The ebind result above required patching that call out. Happy to send a one-line PR.

Samoed:
Hm, not sure why we want to match videos by frames of other videos, while we're using this task as classification already

hubielu:
> Hm, not sure why we want to match videos by frames of other videos, while we're using this task as classification already

Good point. My thinking was that the existing UCF101 tasks are all category-level (51 classes), whereas this is instance-level: 337 scenes, so a model has to separate v_Archery_g01 from v_Archery_g02, same class different recording. I don't think classification covers that, but let me know if you see it differently. On matching by frames of other clips: I tried the more natural version first, where the query frame comes from the target video itself, and it saturates (0.9465 on MSRVTT in #4941, 0.9909 on AVE). The non-adjacent sibling was the only way I found to keep it leak-free with objective ground truth. I went this route mainly because the direction has 0 merged and I haven't found a native i2v dataset that isn't saturated or license-blocked. If you'd rather hold out for one, let me know if I should close this

Samoed:
> My thinking was that the existing UCF101 tasks are all category-level (51 classes), whereas this is instance-level: 337 scenes, so a model has to separate v_Archery_g01 from v_Archery_g02, same class different recording

Why you do you then changed from classification to retrieval?

hubielu:
> Why you do you then changed from classification to retrieval?

None of the existing UCF101 tasks are cross-modal in this direction: they're i2c, v2c, va2c, i2t, va2t, so none put an image query against a video corpus. Classification tests whether video embeddings separate by action label. This tests whether image and video embeddings share a space, which is what the i2v cell measures. Instance-level classification also wouldn't work here: 337 scenes with one clip each leaves no training examples per class.

hubielu:
@Samoed following up on this — #5284 (ABO i2v) merged last week. Is the remaining concern the UCF101 construction specifically, or something else?

Samoed:
> Is the remaining concern the UCF101 construction specifically, or something else?

Yes, I'm not sure in construction. WDYT @AdnanElAssadi

AdnanElAssadi:
I think we should close this one.  I'd rather not add a construction derived from a dataset we already use in other ways, especially with models at 0.87-0.91.

PR #5261, opened 2026-08-21
fix: apply explicit document prompts in GME

whybe-choi:
Previously, GME always discarded document instructions by setting them to `None`, even when a task explicitly defined a document prompt. This change applies document prompts defined in task metadata or `prompts_dict`, while preserving the existing instruction-free behavior when no document prompt is specified.


Samoed (commented):
I don't think that we need to add prompts to the `document` part, because in their paper they say `The instructions for different tasks, we only use the instructions for query encoding.`, so not prompts are used for the document part. Do you have other reference or mismatch score?

whybe-choi:
I missed that. Thanks for pointing it out.

PR #5275, opened 2026-08-23
task: Add Breakfast Clustering and Breakfast Pair Classification datasets

wissam-sib:
- [x] I have outlined why this dataset is filling an existing gap in the MOEB benchmark
  - [x] `LCO-Embedding-Omni-3B`

Results on Breakfast Clustering:
```
metric                 random LCO-Embedding-Omni-3B
v_measure             0.04418        0.46434
ami                   0.04314        0.46375
```

Results on Breakfast Pair Classification:
```
metric                 random LCO-Embedding-Omni-3B
max_accuracy          0.53926        0.72286
max_f1                0.66821        0.75097
max_precision         0.50174        0.64874
max_recall            1.00000        0.90300
max_ap                0.53817        0.77764
cosine_accuracy       0.53349        0.71940
cosine_f1             0.66718        0.75048
cosine_precision      0.50058        0.64204
cosine_recall         1.00000        0.90300
cosine_ap             0.53817        0.77647
```

### DIFF
--- a/mteb/tasks/pair_classification/eng/breakfast_pc.py
+++ b/mteb/tasks/pair_classification/eng/breakfast_pc.py
@@ -16,8 +16,8 @@ class BreakfastPairClassification(AbsTaskPairClassification):
         ),
         reference="https://ieeexplore.ieee.org/document/6909500",
         dataset={
-            "path": "Wissam42/Breakfast-PC-V",
-            "revision": "9ff45d989f5280ba90075694ad0060d79184410d",
+            "path": "mteb/Breakfast-PC-V",
+            "revision": "5c29a53b9045d7bc862d68ff56db3eeb7aaed069",
         },
         type="VideoPairClassification",
         category="v2v",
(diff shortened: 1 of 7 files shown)


mteb/tasks/pair_classification/eng/breakfast_pc.py
```diff
@@ -0,0 +1,50 @@
...
+        description=(
+            "Pair classification on the Breakfast Actions dataset: determining "
+            "whether two video clips depict the same breakfast activity category "
+            "(10 classes). Same-class / different-class pairs are sampled with a "
+            "fixed seed. Built by "
+            "scripts/data/breakfast_pair_classification/create_data.py."
+        ),
+        reference="https://ieeexplore.ieee.org/document/6909500",
+        dataset={
+            "path": "Wissam42/Breakfast-PC-V",
```
isaac-chung: are you able to move this dataset to under the mteb org on hf?

wissam-sib: Yes, let me do this now (and will do the same for my other PRs)

isaac-chung (approved):
Thanks. 

I suspect the PC task would be too easy. But it's not a blocker.

PR #5280, opened 2026-08-23
dataset: Add MUSE-VA A2I and I2A retrieval tasks

wissam-sib:
- [x] I have outlined why this dataset is filling an existing gap in the MOEB benchmark
- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [x] `LCO-Embedding-Omni-3B`
  - [x] `mteb/baseline-random-encoder`
- [x] I have checked that the performance is neither trivial (both models gain close to perfect scores) nor random (both models gain close to random scores).
- [x] I have considered the size of the dataset and reduced it if it is too big (2048 examples is typically large enough for most tasks)

Results on MUSE-VA-A2I:
```
metric                 random LCO-Embedding-Omni-3B
ndcg_at_1             0.00000        0.02240
ndcg_at_5             0.00365        0.05676
ndcg_at_10            0.00569        0.07577
ndcg_at_20            0.01040        0.09928
ndcg_at_100           0.02946        0.17167
map_at_10             0.00355        0.05294
recall_at_1           0.00000        0.02240
recall_at_5           0.00640        0.09120
recall_at_10          0.01280        0.15200
recall_at_20          0.03200        0.24640
mrr_at_10             0.00355        0.05294
```

Results on MUSE-VA-I2A:
```
metric                 random LCO-Embedding-Omni-3B
ndcg_at_1             0.00160        0.02400
ndcg_at_5             0.00570        0.05861
ndcg_at_10            0.00662        0.07911
ndcg_at_20            0.00978        0.10377
ndcg_at_100           0.03156        0.19050
map_at_10             0.00472        0.05576
recall_at_1           0.00160        0.02400
recall_at_5           0.00960        0.09280
recall_at_10          0.01280        0.15680
recall_at_20          0.02560        0.25600
mrr_at_10             0.00472        0.05576
```

PR #5281, opened 2026-08-23
[MOEB] Add Reflectra I2A retrieval task

wissam-sib:
- [ ] I have outlined why this dataset is filling an existing gap in the MOEB benchmark
- [ ] I have tested that the dataset runs with the `mteb` package.
- [ ] I have run the following models on the task (adding the results to the pr). These can be run using the `mteb run -m {model_name} -t {task_name}` command.
  - [ ] ...
  - [ ] `mteb/baseline-random-encoder`
- [ ] I have considered the size of the dataset and reduced it if it is too big (2048 examples is typically large enough for most tasks)

### DIFF
--- a/scripts/data/reflectra_retrieval/create_data.py
+++ b/scripts/data/reflectra_retrieval/create_data.py
@@ -1,0 +1,131 @@
+#!/usr/bin/env python3
+"""Package AraNge/reflectra-benchmark into MTEB image→audio retrieval.
+
+Reflectra contains 1,000 images each rated against six candidate music clips on a
+0–10 scale. We treat scores >= 7 as relevant for binary qrels (image query →
+audio corpus).
+
+Usage:
+  export HF_TOKEN=...
+  uv run python scripts/data/reflectra_retrieval/create_data.py \\
+      --repo-id Wissam42/Reflectra-I2A \\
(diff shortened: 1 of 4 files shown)


scripts/data/reflectra_retrieval/create_data.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python3
+"""Package AraNge/reflectra-benchmark into MTEB image→audio retrieval.
+
+Reflectra contains 1,000 images each rated against six candidate music clips on a
+0–10 scale. We treat scores >= 7 as relevant for binary qrels (image query →
```
Samoed, line 5: Why do you choose 7?

Samoed (commented):
Can you run some models? This is hard to judge anything about datasets because there is no paper about it

scripts/data/reflectra_retrieval/create_data.py
wissam-sib: 7 corresponds to strong match in dataset description here: https://huggingface.co/datasets/AraNge/reflectra-benchmark

wissam-sib:
Closing, see comment here: https://github.com/embeddings-benchmark/mteb/pull/5280#issuecomment-5410287527

PR #5282, opened 2026-08-24
task: add ImagePairClassification task type with 5 tasks

shriyasudhakar:
Adds the `ImagePairClassification` task type and five tasks built from image classification datasets already in mteb. Image had 0 pair classification tasks, while audio has 5 and video 7+.

Pairs are balanced same/different class with a fixed seed, and the construction script is included. The generic `AbsTaskPairClassification` handles the image modality via column pointers, so only the task type registration was needed.

Datasets are under my HF namespace for now, can move to mteb when ready.

### Results (max_ap, test)

| Task | pairs | random | siglip2-base-16-224 | dinov3-vitb16 |
|---|---|---|---|---|
| DTD (textures) | 3000 | 0.5007 | 0.8473 | 0.8872 |
| FGVCAircraft (variants) | 2000 | 0.5239 | 0.9427 | 0.9067 |
| OxfordFlowers (species) | 2000 | 0.5063 | 0.9931 | 0.9989 |
| EuroSAT (land use) | 4000 | 0.5118 | 0.8022 | 0.8399 |
| FER2013 (expressions) | 4000 | 0.5009 | 0.5839 | 0.5959 |

OxfordFlowers looks too easy. I kept it for domain coverage, but happy to swap it for something harder or add another task if preferred.

### Checklist
- [x] Outlined why this fills an existing gap in mteb
- [x] Tested that the datasets run with the `mteb` package
- [x] Ran the following models (results above):
  - [x] another small model: `google/siglip2-base-patch16-224` and `facebook/dinov3-vitb16-pretrain-lvd1689m` (e5-small is text only)
- [x] Performance is neither trivial nor random: random sits at ~0.50, real models 0.58 to 0.94, except OxfordFlowers noted above
- [x] Considered dataset size: 2000 to 4000 pairs per task
- [x] Reproduced paper scores (if applicable): N/A, these are newly constructed pair tasks and no source paper reports pair classification scores


Samoed (commented):
We already have these tasks as classification tasks and I'm not sure if these would have good quality as pair classification tasks

shriyasudhakar:
Editing my response, hadn't read your message properly before @Samoed.

Two reasons I think the pair framing adds signal beyond the existing classification tasks:

1. **They measure different things.** The classification tasks fit a logistic probe on embeddings, so they measure linear separability with supervision. Pair classification is probe free (cosine similarity vs a threshold), so it measures the zero shot geo of the embedding space; same-class instances can be far apart even when classes are linearly separable. This is also how the audio pair tasks were built as well.

2. **Results allude to the fact that they're not redundant**: model ordering flips across tasks (DINOv3 > SigLIP2 on DTD/EuroSAT pairs, reversed on FGVC-Aircraft), and FER2013 pairs sit at 0.58 - 0.60 max_ap.

If you'd prefer the task type built on natively paired data instead of derived pairs. I just looked into some, which include Totally-Looks-Like (perceptual pairs), Copydays (copy detection), or animal re-ID, which are some instance-level similarity with no classification counterpart. Let me know what you think

isaac-chung:
@shriyasudhakar thanks for the effort, but I agree with @Samoed here. These are too similar to the classification variants, and we see similar high scores on most of these tasks, e.g. [Oxford Flower clf](https://mteb-leaderboard.hf.space/tasks/OxfordFlowersClassification). Closing this now.

Issue #5276, opened 2026-08-23
dataset-loading-check CI "install dependencies" step consistently times out


isaac-chung:
### Describe the bug

Most recent runs:
1. https://github.com/embeddings-benchmark/mteb/actions/runs/32648625307/job/97216699177
2. https://github.com/embeddings-benchmark/mteb/actions/runs/32648625307/job/97216699177

### To reproduce

See above section.

### Are you interested to contribute a fix for this bug?

Yes

isaac-chung:
After the above PR, it seems like the timeout persists in [dataset-loading-check for the main branch](https://github.com/embeddings-benchmark/mteb/actions/runs/32660186718/job/97245069693). Documentation and lint workflows have dramatically sped up tho.

isaac-chung:
[Debug mode](https://github.com/embeddings-benchmark/mteb/actions/runs/32713003763/job/97388244984) shows the uv sync step is stuck at the venv creation step. Trying out creating the venv step manually.

PR #5290, opened 2026-08-24
ci: fix dataset_loading workflow hang by replacing uv sync with uv pip install

isaac-chung:
[agent assisted analysis]

`uv sync` was hanging indefinitely on every run. A series of diagnostic steps isolated the root cause.

**Ruled out:**
- uv-managed Python interpreter — both `python -m venv` and `uv venv` completed fine in isolation
- uv cache lock contention — `UV_LOCK_TIMEOUT=60` had no effect; trace confirmed the cache lock was acquired and released normally
- Missing/incompatible `.venv` — trace confirmed uv accepted the environment

**Root cause (via `RUST_LOG: uv=trace`, [artifacts](https://github.com/embeddings-benchmark/mteb/actions/runs/32718923659/artifacts/9517592285)):**

Both `uv sync --group test --no-install-project` and `uv sync --no-install-project` hung at the identical line:
```
uv::commands::project::lock_target::toml::from_str lock path=.../uv.lock
```
`uv.lock` is 54 MB / 19k lines. uv always parses the full lockfile before doing anything else, regardless of which group or extras are requested. Rust serde deserialization of this file into typed structs requires significantly more memory than the raw file size, consistent with exhausting available memory on the runner.

**Why lint always completes but dataset-loading always hangs:**

Both workflows trigger on the same push and run `uv sync` against the same lockfile. Lint always finishes first (~25s). The trace shows uv acquires an **exclusive filesystem lock on the project directory** (`/tmp/uv-<hash>.lock`, hash derived from project path) during sync. If both jobs land on the same physical runner, dataset-loading blocks waiting for the lock held by lint. Once lint releases it, dataset-loading acquires it and begins parsing `uv.lock` — at which point the runner has elevated memory pressure from lint's installed packages still resident, causing the deserialization to stall or OOM.

**Fix:**

Bypass `uv.lock` entirely. Create a venv via `python -m venv` (stdlib, confirmed hang-free in diagnostics) and install only the packages this workflow actually needs via `uv pip install`, which resolves without reading the lockfile. Also reverts the Python 3.10 experiment back to 3.11, and scopes the `pull_request` trigger to changes to the workflow file itself.

#5276

### DIFF
--- a/.github/workflows/dataset_loading.yml
+++ b/.github/workflows/dataset_loading.yml
@@ -19,48 +19,8 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v6
 
-    - name: Install uv
-      uses: astral-sh/setup-uv@v8.2.0
-      with:
-        enable-cache: true
-        version: "0.10.12"
-        python-version: '3.10'
-
-    - name: Trace uv sync
-      timeout-minutes: 25
-      continue-on-error: true
-      env:
-        RUST_LOG: uv=trace
-        UV_LOG_CONTEXT: "1"
-        UV_NO_PROGRESS: "1"
-        UV_CACHE_DIR: ${{ runner.temp }}/uv-cache-${{ github.run_id }}-${{ github.run_attempt }}
-        UV_LOCK_TIMEOUT: "60"
-      run: |
-        set -o pipefail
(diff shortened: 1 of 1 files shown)


isaac-chung:
Merging now to test whether the passing can be replicated on the main branch.

PR #5293, opened 2026-08-24
ci: limit PR dataset checks to added task files

### LINKED ISSUE #5288: dataset-loading-check CI: PR checks only run for the added dataset

We already run full dataset sweeps at push to main. No point rerunning again in the PR. Can be a lot more specific. Might need to break out into a separate workflow, but not needed. 

e.g. when there is no dataset added like in [this PR](https://github.com/embeddings-benchmark/mteb/actions/runs/32723489392/job/97419669407), 1468 tests were still selected.

nightcityblade:
Fixes embeddings-benchmark/mteb#5288

PR dataset-loading checks now keep only task files added relative to the base branch before extracting dataset revisions. When a PR adds no task file, the helper exports an explicit empty selection so pytest does not fall back to the full main-branch sweep.

Tests:
- PYTHONPATH=. .venv/bin/pytest -q --confcutdir=tests/test_workflow tests/test_workflow/test_extract_datasets.py
- CUSTOM_DATASET_REVISIONS=__EMPTY__ PYTHONPATH=. .venv/bin/pytest -q --confcutdir=tests/test_tasks tests/test_tasks/test_dataset_on_hf.py -m test_datasets (expected empty-parameter skip)
- .venv/bin/ruff check scripts/extract_datasets.py tests/test_workflow/test_extract_datasets.py
- .venv/bin/ruff format --check scripts/extract_datasets.py tests/test_workflow/test_extract_datasets.py

N/A — this PR changes CI only and adds neither.


isaac-chung:
Thanks. I'm aware of this. Just wanted to double check why we see deviated behaviour in that test before I move on to reviewing.

scripts/extract_datasets.py
```diff
@@ -63,9 +66,25 @@ def extract_datasets(files: list[str]) -> list[tuple[str, str]]:
         logging.debug(f"Set CUSTOM_DATASET_REVISIONS={custom_revisions}")
 
         print(f'export CUSTOM_DATASET_REVISIONS="{custom_revisions}"')
+    else:
+        # Keep the test module from falling back to the full dataset sweep.
+        print('export CUSTOM_DATASET_REVISIONS="__EMPTY__"')
```
isaac-chung, line 68: I think this is the only change needed to be kept in this PR.

isaac-chung (changes requested):
Thanks for jumping on this. I realize I did not motivate the problem too much in the issue.   But it seems like we've converged to the core problem as well, which is:
> When a PR adds no task file, the helper exports an explicit empty selection so pytest does not fall back to the full main-branch sweep.

I've updated the issue with an example workflow run from yesterday. 

Based on this, the key should be setting CUSTOM_DATASET_REVISIONS to some string that does not include `:` (as you have done here), and let [the list be [] in the pytest on the reading step](https://github.com/embeddings-benchmark/mteb/blob/d3bc5aec881c676bf666235e7f6aafc2ce03f5b0/tests/test_tasks/test_dataset_on_hf.py#L40-L44). 

To me, the rest of the PR, i.e. the extract_added_datasets bits, would be redundant as the functionality already exists, and should be removed.

nightcityblade:
Thanks for the clarification. I narrowed the PR in `5d4e185` to the empty-selection fallback in `scripts/extract_datasets.py` and removed the redundant added-file filtering and its test.

Validation: Ruff check and format passed, and a focused invocation confirmed that `extract_datasets([])` emits `CUSTOM_DATASET_REVISIONS="__EMPTY__"` and returns an empty list.

isaac-chung (approved):

PR #5298, opened 2026-08-25
fix: mock_tasks not supported by get_instruction

### LINKED ISSUE #5297: Mock tasks are not supported by `get_instruction`
In `get_instruction` we're using `get_task`, but this is not possible to get info about mock task with `mteb.get_task`

ayush1298:
closes #5297 

`AbsEncoder.get_instruction` fell back to `get_task(task_name=task_metadata.name).abstask_prompt`,
which only works for tasks in the registry. Mock tasks (and any custom/unregistered task) raised `KeyError`

I have added `mteb.get_tasks.get_abstask_prompt(task_type)`, which resolves the default prompt from the
task **type** instead of the task **name**:

1. a type → `abstask_prompt` mapping built once from the task registry;
2. a fallback via `_TASKTYPE2SIMPLIFIEDTASKTYPE` for task types no registered task implements yet
   (e.g. `VideoMultilabelClassification`);
3. otherwise an empty string plus a one-time warning.

`get_instruction` now calls it. Prompt precedence is unchanged. `prompts_dict` and `TaskMetadata.prompt` are still checked first; this only affects the final fallback.

- No task type has conflicting `abstask_prompt` values, so registered tasks that resolved before return the identical string.
- Every value in `_TASK_TYPE` now resolves to a non-empty prompt, and each mock task's resolved prompt equals its own class's `abstask_prompt`.
- verified with `mteb.mock_run` on `codefuse-ai/F2LLM-v2-80M` and `intfloat/multilingual-e5-small` before and after fix

### DIFF
--- a/mteb/get_tasks.py
+++ b/mteb/get_tasks.py
@@ -77,8 +77,6 @@ def _create_task_type_to_prompt_mapping(
     """Create a mapping from task type to the prompt defined by its abstask (base) class."""
     type_to_prompt: dict[str, str] = {}
     for cls in tasks:
-        # `abstask_prompt` is defined on the abstask (base) class, e.g. `AbsTaskRetrieval`, and is
-        # only annotated on `AbsTask` itself, so it can be missing (e.g. zero-shot classification)
         prompt = getattr(cls, "abstask_prompt", None)
         if prompt:
             type_to_prompt.setdefault(cls.metadata.type, prompt)
--- a/mteb/get_tasks.py
+++ b/mteb/get_tasks.py
@@ -17,7 +17,7 @@ from mteb.abstasks import (
 from mteb.filter_tasks import filter_tasks
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, Sequence
+    from collections.abc import Iterable, Iterator, Sequence
 
     from mteb.abstasks.task_metadata import TaskCategory, TaskDomain, TaskType
     from mteb.types import Modalities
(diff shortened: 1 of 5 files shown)


mteb/get_tasks.py
```diff
@@ -71,6 +71,57 @@ def _create_similar_tasks(tasks: Iterable[type[AbsTask]]) -> dict[str, list[str]
...
+_TASK_TYPE_TO_PROMPT = _create_task_type_to_prompt_mapping(TASK_LIST)
+
+# fallback for task types that no task in the registry implements (yet)
+_SIMPLIFIED_TASK_TYPE_TO_PROMPT: dict[str, str] = {
+    "retrieval": "Retrieve text based on user query.",
+    "clustering": "Identify categories in user passages.",
+    "classification": "Classify user passages.",
+    "semantic-similarity": "Retrieve semantically similar text.",
+    "pair-classification": "Retrieve text that are semantically similar to the given text.",
+}
```
Samoed: I don't think that we need that

mteb/get_tasks.py
Samoed, line 1: I don't think that prompt related functions should be in `get_task`

mteb/get_tasks.py
ayush1298: As I mentioned in the PR description, this is for final fallback only. 1st it will check `prompts_dict` and `TaskMetadata.prompt` only

Samoed: I think they will fallback to prompt from base class and we don't need to handle simplified task types

ayush1298: so should we do it like, if `task_type` is not there in task-to-prompt mapping, then we just logged warning and return empty string?

Samoed: I think we can return prompt from base class

ayush1298: added

mteb/get_tasks.py
```diff
@@ -71,6 +71,68 @@ def _create_similar_tasks(tasks: Iterable[type[AbsTask]]) -> dict[str, list[str]
...
 
 
+def _create_task_type_to_prompt_mapping(
+    tasks: Iterable[type[AbsTask]],
+) -> dict[str, str]:
+    """Create a mapping from task type to the prompt defined by its abstask (base) class."""
+    type_to_prompt: dict[str, str] = {}
+    for cls in tasks:
+        # `abstask_prompt` is defined on the abstask (base) class, e.g. `AbsTaskRetrieval`, and is
+        # only annotated on `AbsTask` itself, so it can be missing (e.g. zero-shot classification)
```
Samoed: I think better to add to zero-shot classification then. All tasks should have abs prompt

ayush1298: added `abstask_prompt` to zero-shot classification and Image-Text Pair classification

mteb/get_tasks.py
Samoed: This is still not resolved I think

mteb/get_tasks.py
```diff
@@ -71,6 +71,66 @@ def _create_similar_tasks(tasks: Iterable[type[AbsTask]]) -> dict[str, list[str]
...
+        task_type: The type of the task, e.g. "Retrieval".
+
+    Returns:
+        The prompt of the abstask class of the task type, or an empty string if its abstask class
+            doesn't define one (e.g. zero-shot classification).
+    """
+    if task_type in _TASK_TYPE_TO_PROMPT:
+        return _TASK_TYPE_TO_PROMPT[task_type]
+
+    # no task in the registry uses this task type, fall back to the imported task classes
```
Samoed: When this can be possible?

ayush1298: for a task type declared in `TaskType` that no registered task implements. Currently its applicable for `VideoMultilabelClassification,` implemented only by `MockVideoMultilabelClassificationTask` and `MockVideoAudioMultilabelClassificationTask`

mteb/get_tasks.py
ayush1298: Will it be okay if I implement it in `abstask.py` and then import from there in `abs_encoder.py`

mteb/get_tasks.py
```diff
@@ -71,6 +71,66 @@ def _create_similar_tasks(tasks: Iterable[type[AbsTask]]) -> dict[str, list[str]
...
+def _create_task_type_to_prompt_mapping(
+    tasks: Iterable[type[AbsTask]],
+) -> dict[str, str]:
+    """Create a mapping from task type to the prompt defined by its abstask (base) class."""
+    type_to_prompt: dict[str, str] = {}
+    for cls in tasks:
+        prompt = getattr(cls, "abstask_prompt", None)
+        if prompt:
+            type_to_prompt.setdefault(cls.metadata.type, prompt)
+    return type_to_prompt
```
Samoed: I don't think type is good filed for separete prompts. Won't be better to use base class as key?

ayush1298: Made some changes. Instead of using the base class as the key, I am using now the cache dict `_task_name_to_prompt` which is keyed by task name

mteb/get_tasks.py
Samoed: Yes

Samoed (approved):

KennethEnevoldsen (approved):

PR #5306, opened 2026-08-26
fix: drop unlabelled documents at a level instead of scoring them

arthi-arumugam-git:
## What changed

In `_evaluate_clustering_bootstrapped`, documents whose label path does not reach the level being scored get the sentinel `-1`, and the next lines try to filter them out. The filter never removes anything.

Labels are strings, so `np.array(['Fysikk', -1])` gives a `<U21` array and the sentinel becomes the string `'-1'`. The comparison is then `np.str_('-1') != -1`, always `True`. The in-code comment records that the vectorised form raised a `FutureWarning`; rewriting it as a loop silenced the warning but not the cause.

So those documents stay in with `'-1'` as their gold class. `n_clusters=np.unique(...).size` is one larger than the number of real classes, and `v_measure`/`ami` are scored against a gold partition holding a class that means "no label at this level".

The fix builds the mask from the same condition that produced the sentinel, so the sentinel is never created.

## Why it matters

Measured on the test splits at the pinned revisions, both `SNLHierarchicalClusteringP2P` and `...S2S` (1300 docs, `max_depth=5`):

| Level | 1 | 2 | 3 | 4 |
|---|---|---|---|---|
| documents given the sentinel | 0% | 5% | 26% | **73%** |

`_evaluate_subset` averages across levels before reporting `v_measure`, so level 4 reaches the reported number.

I did not measure `ArXivHierarchicalClustering` or the VG tasks, so I make no claim about those.

## Scope

Only datasets whose label paths vary in depth are affected. I ran both versions over a depth-1 dataset and the score dictionaries are identical. Scores for the hierarchical tasks will change, so existing numbers for them were produced under the old behaviour. I have not touched anything about published results.

## Tests

Two added to `tests/test_abstasks/test_clustering_metrics.py`, both failing on `main`: `v_measure` is 1.0 here and 0.3632 on main, and `k` is 2 here and 3 on main. `pytest tests/test_abstasks`: 166 passed, 46 skipped (all unrelated optional media deps).

### DIFF
--- a/mteb/abstasks/clustering.py
+++ b/mteb/abstasks/clustering.py
@@ -46,7 +46,7 @@ def _evaluate_clustering_bootstrapped(
     max_depth: int | None,
     rng_state: random.Random,
     seed: int,
-    drop_unlabelled_documents: bool = True,
+    drop_unlabelled_documents: bool = False,
 ) -> tuple[dict[str, dict[str, list[float]]], dict[str, list[list[int]]]]:
     """Bootstrapped evaluation of clustering performance.
 
@@ -86,9 +86,8 @@ def _evaluate_clustering_bootstrapped(
             if not valid_idx.all():
                 logger.warning(
                     "Level %d: dropping %d of %d documents that have no label at "
-                    "this level. Set drop_unlabelled_documents=False on the task to "
-                    "score them as one shared class instead, which is the older "
-                    "behaviour.",
+                    "this level, because drop_unlabelled_documents is set on this task. "
+                    "The older behaviour scores them as one shared class.",
                     i_level,
                     len(labels) - int(valid_idx.sum()),
                     len(labels),
@@ -167,10 +166,11 @@ class AbsTaskClustering(AbsTask):
         drop_unlabelled_documents: For a hierarchical task, whether a document whose label
             path is shorter than the level being scored is dropped at that level rather than
             scored. Those documents have nothing in common except a missing label, so keeping
-            them asks the model to find a class that is not really there. Defaults to True.
-            The hierarchical tasks that existed before this option was added set it to False
-            so their published scores stay reproducible. A task whose label paths all reach
-            the same depth is unaffected either way, since no document is ever dropped.
+            them asks the model to find a class that is not really there. Defaults to False,
+            which is the behaviour every published hierarchical score was produced under.
+            New tasks that want the corrected behaviour set it to True; the .v2 hierarchical
+            tasks do. A task whose label paths all reach the same depth is unaffected either
+            way, since no document is ever dropped.
         input_column_name: Name of the column(s) containing the input sentences or data points. Default is "sentences".
             Can be a string for single-column tasks or a list of strings for multimodal tasks (e.g. ["video", "audio"]).
             When specified as a list, values must be the default column names as defined in the encoder I/O types
@@ -185,7 +185,7 @@ class AbsTaskClustering(AbsTask):
     n_clusters: int = 10
     k_mean_batch_size: int = 512
     max_depth = None
-    drop_unlabelled_documents: bool = True
+    drop_unlabelled_documents: bool = False
     abstask_prompt = "Identify categories in user passages."
     input_column_name: str | Sequence[Modalities] = "sentences"
     label_column_name: str = "labels"
--- a/mteb/abstasks/clustering.py
+++ b/mteb/abstasks/clustering.py
@@ -46,11 +46,17 @@ def _evaluate_clustering_bootstrapped(
     max_depth: int | None,
     rng_state: random.Random,
     seed: int,
+    drop_unlabelled_documents: bool = True,
 ) -> tuple[dict[str, dict[str, list[float]]], dict[str, list[list[int]]]]:
     """Bootstrapped evaluation of clustering performance.
 
     The bootstrapping is done by sampling N samples from the corpus and clustering them. It is done without replacement to get a diverse set of
     samples.
+
+    `drop_unlabelled_documents` decides what happens at a given level to the documents
+    whose label path is shorter than that level: drop them, or keep them under the
+    sentinel gold label -1, which is the older behaviour. See
+    `AbsTaskClustering.drop_unlabelled_documents`.
 
     Returns:
         A tuple containing:
@@ -72,15 +78,45 @@ def _evaluate_clustering_bootstrapped(
         max_depth = max(map(len, labels))
     # Evaluate on each level til max depth
     for i_level in range(max_depth):
-        # Drop the documents whose label path does not reach this level, rather
-        # than give them a sentinel gold label and filter on it afterwards. An
-        # earlier version appended -1 for those documents, but np.array() on a
-        # mixed list of str and int returns a string array, so the sentinel
-        # became the string "-1" and the filter kept every document.
-        valid_idx = np.array([len(label) > i_level for label in labels])
-        np_level_labels = np.array(
-            [label[i_level] for label in labels if len(label) > i_level]
-        )
+        if drop_unlabelled_documents:
+            # Drop the documents whose label path does not reach this level,
+            # rather than give them a sentinel gold label and filter on it
+            # afterwards.
+            valid_idx = np.array([len(label) > i_level for label in labels])
+            if not valid_idx.all():
+                logger.warning(
+                    "Level %d: dropping %d of %d documents that have no label at "
+                    "this level. Set drop_unlabelled_documents=False on the task to "
+                    "score them as one shared class instead, which is the older "
+                    "behaviour.",
+                    i_level,
+                    len(labels) - int(valid_idx.sum()),
+                    len(labels),
+                )
+            np_level_labels = np.array(
+                [label[i_level] for label in labels if len(label) > i_level]
+            )
+        else:
+            # The older behaviour, kept so that published results stay
+            # reproducible. Note that it is not one consistent rule: np.array()
+            # on a mixed list of str and int returns a string array, so for
+            # string labels the sentinel becomes "-1", the filter below never
+            # matches, and every document is kept and scored as one shared
+            # class. For integer labels the filter does fire and they are
+            # dropped. Which of the two happens is decided by the dataset's
+            # label dtype.
+            level_labels: list[str | int] = []
+            # Assign -1 to gold label if the level is not there
+            for label in labels:
+                if len(label) > i_level:
+                    level_labels.append(label[i_level])
+                else:
+                    level_labels.append(-1)
+            np_level_labels = np.array(level_labels)
+            valid_idx = np.array(
+                [level_label != -1 for level_label in np_level_labels]
+            )  # Could be level_labels != -1 but fails with FutureWarning: elementwise comparison failed
+            np_level_labels = np_level_labels[valid_idx]
         level_embeddings = embeddings[valid_idx]
         clustering_model = MiniBatchKMeans(
             n_clusters=np.unique(np_level_labels).size,
@@ -128,6 +164,13 @@ class AbsTaskClustering(AbsTask):
         n_clusters: Number of clustering experiments to run.
         k_mean_batch_size: Batch size to use for k-means clustering.
         max_depth: Maximum depth to evaluate clustering. If None, evaluates all levels.
+        drop_unlabelled_documents: For a hierarchical task, whether a document whose label
+            path is shorter than the level being scored is dropped at that level rather than
+            scored. Those documents have nothing in common except a missing label, so keeping
+            them asks the model to find a class that is not really there. Defaults to True.
+            The hierarchical tasks that existed before this option was added set it to False
+            so their published scores stay reproducible. A task whose label paths all reach
+            the same depth is unaffected either way, since no document is ever dropped.
         input_column_name: Name of the column(s) containing the input sentences or data points. Default is "sentences".
             Can be a string for single-column tasks or a list of strings for multimodal tasks (e.g. ["video", "audio"]).
             When specified as a list, values must be the default column names as defined in the encoder I/O types
@@ -142,6 +185,7 @@ class AbsTaskClustering(AbsTask):
     n_clusters: int = 10
     k_mean_batch_size: int = 512
     max_depth = None
+    drop_unlabelled_documents: bool = True
     abstask_prompt = "Identify categories in user passages."
     input_column_name: str | Sequence[Modalities] = "sentences"
     label_column_name: str = "labels"
@@ -250,6 +294,7 @@ class AbsTaskClustering(AbsTask):
                 max_depth=self.max_depth,
                 rng_state=self.rng_state,
                 seed=self.seed,
+                drop_unlabelled_documents=self.drop_unlabelled_documents,
             )
 
         if prediction_folder:
--- a/mteb/abstasks/clustering.py
+++ b/mteb/abstasks/clustering.py
@@ -72,18 +72,15 @@ def _evaluate_clustering_bootstrapped(
         max_depth = max(map(len, labels))
     # Evaluate on each level til max depth
     for i_level in range(max_depth):
-        level_labels: list[str | int] = []
-        # Assign -1 to gold label if the level is not there
-        for label in labels:
-            if len(label) > i_level:
-                level_labels.append(label[i_level])
-            else:
-                level_labels.append(-1)
-        np_level_labels = np.array(level_labels)
-        valid_idx = np.array(
-            [level_label != -1 for level_label in np_level_labels]
-        )  # Could be level_labels != -1 but fails with FutureWarning: elementwise comparison failed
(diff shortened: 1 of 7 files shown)


arthi-arumugam-git:
Trimmed to 1,963 characters. The check does not re-run on a description edit, so the message above still quotes the original length.

Samoed:
How scores changes for tasks?

arthi-arumugam-git:
Good question, I had only measured the sentinel share before. I have now run both affected tasks end to end.

Method: `all-MiniLM-L6-v2`, the real test splits at the pinned revisions, the same embeddings scored twice so only the filter differs. Task defaults, `max_depth=5`, `n_clusters=10`, `seed=42`.

**SNLHierarchicalClusteringP2P** (1300 docs)

| level | docs given the sentinel | main | this PR |
|---|---|---|---|
| 0 | 0% | 0.2543 | 0.2543 |
| 1 | 0.2% | 0.5455 | 0.5448 |
| 2 | 5% | 0.6821 | 0.6972 |
| 3 | 26% | 0.7279 | 0.8560 |
| 4 | 73% | 0.3881 | 0.8979 |
| **v_measure** | | **0.5196** | **0.6500** |
| **ami** | | 0.2601 | 0.3759 |

**SNLHierarchicalClusteringS2S** (1300 docs)

| level | main | this PR |
|---|---|---|
| 0 | 0.2602 | 0.2602 |
| 1 | 0.5429 | 0.5437 |
| 2 | 0.6779 | 0.6942 |
| 3 | 0.7329 | 0.8607 |
| 4 | 0.3719 | 0.8952 |
| **v_measure** | **0.5172** | **0.6508** |
| **ami** | 0.2555 | 0.3724 |

So about +0.13 v_measure on both, and the per-level change tracks the sentinel share almost exactly: level 0 has no sentinel and does not move at all, level 4 is 73% sentinel and moves +0.51.

One thing worth correcting from my own description. The PR body illustrates the effect with a constructed case where the score is inflated. On this real data it goes the other way: scores go up once the fix is in. The `'-1'` group is documents that have nothing in common except a missing label, so it is not recoverable in embedding space, and the model was being marked down for failing to find a class that is not really there. Whether the effect inflates or suppresses depends on whether that group happens to be separable, so I would not want to state a direction as a general rule. What is consistent is that the number moves by a lot at the levels where the sentinel is common.

Happy to run other models or the ArXiv and VG tasks if that would help.

KennethEnevoldsen (commented):
Ahh this is a great find. Thanks!

So agree that this is the ideal way to do it. I am not sure what is the best way to implement it. Just doing this as is would leave a lot of our results dated. What we could do it enable it as a setting which we set to this behaviour as default, but for all existing datasets set to current behaviour.

This ensure that all results are reproducible, while also ensuring that new dataset gets the update. 

The problem is that this can break local datasets that people have for internal benchmarks. I am not sure how common it is a have a hierarchical clustering task, so the damage is probably minor but it will be breaking.

The other option is to create an entire "v2" of clustering and deprecate the old one, which also seems problematic.

Regardless we probably have to run this through some tasks to get an idea about the impact (I suspect it is minor).

If it is minor I would probably lean 1). Let me know what you think

arthi-arumugam-git:
Thanks Kenneth. Option 1 sounds right to me too, and I can rework the PR that way.

On impact, two measurements. The score change on the two SNL tasks is in my comment from the 27th above: about +0.13 v_measure on both, and the per-level move tracks the share of documents that have no label at that level. This morning I measured that share for all six hierarchical tasks, labels only, no model:

- ArXivHierarchicalClustering P2P and S2S: 53.7% of documents have no level-1 label
- VGHierarchicalClustering P2P and S2S: 13.7%
- SNLHierarchicalClustering P2P and S2S: 0.2% / 4.7% / 26.3% / 72.6% at levels 1 to 4

So the affected set is exactly the six hierarchical tasks, and none of them lightly; ArXiv would move the most. A task where every label path reaches the same depth selects the same documents under both versions of the code, so the rest of the clustering suite does not change.

One nuance for the reproducibility side: on main the outcome depends on the label dtype. The valid_idx filter is already there and does fire when the labels are integers; it only no-ops when they are strings, because the mixed list becomes a string array and the sentinel turns into "-1". So current behaviour is the accident of each dataset's dtype rather than one consistent rule. I still think pinning the existing tasks is right for published results, it just means the pin preserves that accident deliberately.

For the implementation I would propose: a flag on the task, something like drop_unlabelled_documents defaulting to True, threaded into _evaluate_clustering_bootstrapped, set to False on the six existing hierarchical tasks, plus a logged warning whenever documents are actually dropped so anyone with a local hierarchical task can see why a number moved. If that shape works for you I will push it to this PR.

KennethEnevoldsen:
Thanks @arthi-arumugam-git for the overview - I think it sounds reasonable, let me just get a second opinion before we implement anything

@Samoed would love you thoughts on this as well before we start implementing

Samoed:
I think flag for class would work fine without score reproduction problems. I think would be intresting to see performance of non-hierarchical clustering tasks

arthi-arumugam-git:
Thanks both. I have pushed the flag version.

`drop_unlabelled_documents` is a task attribute defaulting to `True`, threaded into `_evaluate_clustering_bootstrapped`, and set to `False` on the six existing hierarchical tasks so their published numbers do not move. The `False` branch is the old code left as it was, and it warns when documents are actually dropped, so anyone with a local hierarchical task can see why a number changed.

@Samoed on the non-hierarchical tasks: I went and measured rather than reasoned about it, because I had asserted earlier that only the six were affected and I wanted to check that was actually true.

There are 59 clustering tasks. 43 have a scalar label column, so once `_evaluate_subset` wraps each label they are depth 1 for every document and nothing is ever dropped. The other 16 have a sequence-typed column, which is the only way a task could be ragged, so I loaded those through their own `dataset_transform` and counted the label path lengths that actually reach the scorer:

| task | label path depths (test) | ragged |
|---|---|---|
| ArXivHierarchicalClustering P2P / S2S | 1: 1099, 2: 949 | yes |
| VGHierarchicalClustering P2P / S2S | 1: 281, 2: 1767 | yes |
| SNLHierarchicalClustering P2P / S2S | 1: 3, 2: 58, 3: 281, 4: 602, 5: 231, 6: 70, 7: 27, 8: 15, 9: 9, 10: 3, 12: 1 | yes |
| TwentyNewsgroupsClustering.v2 | 1: 59545 | no |
| StackExchangeClusteringP2P.v2 | 1: 74914 | no |
| BlurbsClusteringP2P.v2 | 1: 18084 | no |
| TenKGnadClustering P2P / S2S | 1: 10275 / 1: 10267 | no |
| BlurbsClusteringS2S.v2 | 1: 2048 | no |
| CLSClusteringS2S.v2 | 1: 2048 | no |
| StackExchangeClustering.v2 | 1: 2048 | no |
| ThuNewsClustering P2P / S2S | 1: 2048 | no |

That is 14 of the 16 measured end to end. The remaining two are the Reddit pair, and they cannot be ragged by construction: `dataset_transform` flattens their labels with `itertools.chain.from_iterable`, so what reaches the scorer is a plain string per document.

So every non-hierarchical task is depth 1 throughout. Both settings then select the same rows and the score is not merely close, it is identical, which is why I added a test asserting that rather than a tolerance.

The same conclusion falls out of the code: only three modules build a label path longer than one element, `arxiv_hierarchical_clustering`, `snl_hierarchical_clustering` and `vg_hierarchical_clustering`. The non-hierarchical SNL and VG tasks join their parts back into a single string, so they stay depth 1.

Two caveats worth stating. The sequence-typed column on the hub is not a reliable guide on its own, because several tasks store labels per cluster and flatten them in `dataset_transform`, so the shape on the hub is not the shape the scorer sees. And these depths line up with the sentinel shares I posted on the 27th, ArXiv at 1099/2048 and VG at 281/2048, which is a useful sign the two measurements agree.

Happy to run models over any of these if you would rather see scores than row counts, though for the non-hierarchical ones I do not think a run tells you anything the counts do not.

arthi-arumugam-git:
The Reddit pair finished downloading, so that is now measured rather than argued from the code: `RedditClustering.v2` is 1: 2048 and `RedditClusteringP2P.v2` is 1: 459389, both depth 1 throughout.

So all 16 sequence-typed tasks are measured, and the ragged set is exactly the six hierarchical ones.

mteb/tasks/clustering/eng/arxiv_hierarchical_clustering.py
```diff
@@ -14,6 +14,10 @@ def split_labels(record: dict) -> dict:
 
 
 class ArXivHierarchicalClusteringP2P(AbsTaskClustering):
+    # Pinned to the behaviour these scores were published under. New
+    # hierarchical tasks should leave this at the default.
+    drop_unlabelled_documents = False
```
Samoed: I don't think we want to update existing tasks. I think better to create a new version for them

arthi-arumugam-git:
Good call, and you are right that pinning them was still changing them. Reworked as `.v2` tasks and pushed.

What changed:

- **`drop_unlabelled_documents` now defaults to `False`**, which is the behaviour every published hierarchical score was produced under. This is the part that actually answers the concern: a task that predates the option is unaffected by it and needs no attribute at all, so nothing can move by accident.
- **The six originals are untouched apart from `superseded_by`.** The diff on those three files is `+44/-0`, so nothing was removed or edited.
- **Six new `.v2` tasks** subclass the originals, set the flag to `True`, and reuse the same dataset path and revision. A `.v2` here is a scoring change, not a data change, so a difference in score cannot be blamed on the dataset moving. There is a test asserting that.

I followed the `BlurbsClusteringP2P` / `.v2` pattern already in the repo. Tell me if you would rather they were full standalone classes than subclasses; I went with subclassing so the metadata stays single-sourced and the two versions cannot drift apart.

Tests are 11 passing, including one that the two versions of a task select identical rows when every label path reaches the same depth, so the non-hierarchical suite is provably untouched.

One thing worth flagging since it is a judgement call rather than a fact. With the default at `False`, a brand new hierarchical task written next year inherits the old behaviour unless its author knows to opt in. The alternative is defaulting to `True` and setting `False` on the six, which is what I had, and you have said no to that. I think your version is right for reproducibility, and the warning fires whenever documents are actually dropped, so the failure mode is a silently unchanged number rather than a silently changed one. Happy either way, just noting it so the choice is deliberate.

mteb/tasks/clustering/nob/snl_hierarchical_clustering.py
```diff
@@ -83,3 +85,45 @@ class SNLHierarchicalClusteringS2S(AbsTaskClustering):
...
+    """
+
+    drop_unlabelled_documents = True
+
+    metadata = SNLHierarchicalClusteringP2P.metadata.model_copy(
+        update={"name": "SNLHierarchicalClusteringP2P.v2", "superseded_by": None},
+    )
+
+
+class SNLHierarchicalClusteringS2SV2(SNLHierarchicalClusteringS2S):
```
KennethEnevoldsen, line 111: Don't subclass, it is only a metadata class anyway. What is important is updating the metadata such that the description described why the update (see other v2 datasets as examples, e.g. [this one](https://leaderboard.mteb.org/tasks/SummEvalSummarization.v2))

tests/test_abstasks/test_clustering_metrics.py
```diff
@@ -67,3 +76,162 @@ def test_end_to_end_result_dict_contains_both_metrics():
...
+
+
+def test_v2_hierarchical_tasks_drop_unlabelled_documents():
+    # The new version is where the corrected behaviour lives.
+    for name in HIERARCHICAL_TASKS:
+        task = mteb.get_task(f"{name}.v2")
+        assert task.drop_unlabelled_documents is True, name
+
+
+def test_v2_keeps_the_same_data_as_the_task_it_supersedes():
```
KennethEnevoldsen, line 227: delete

tests/test_abstasks/test_clustering_metrics.py
```diff
@@ -67,3 +76,162 @@ def test_end_to_end_result_dict_contains_both_metrics():
...
+def test_existing_hierarchical_tasks_are_untouched_and_superseded():
+    # These six were scored before the option existed. They keep their behaviour
+    # and point at the .v2 that has the fix.
+    for name in HIERARCHICAL_TASKS:
+        task = mteb.get_task(name)
+        assert task.drop_unlabelled_documents is False, name
+        assert task.metadata.superseded_by == f"{name}.v2", name
+
+
+def test_v2_hierarchical_tasks_drop_unlabelled_documents():
```
KennethEnevoldsen, line 220: delete - versioning doesn't correlate with this attribute

tests/test_abstasks/test_clustering_metrics.py
```diff
@@ -67,3 +76,162 @@ def test_end_to_end_result_dict_contains_both_metrics():
...
+
+
+def test_default_keeps_the_published_behaviour():
+    # Default False, so a task that predates the option is unaffected by it.
+    from mteb.abstasks.clustering import AbsTaskClustering
+
+    assert AbsTaskClustering.drop_unlabelled_documents is False
+
+
+def test_existing_hierarchical_tasks_are_untouched_and_superseded():
```
KennethEnevoldsen, line 211: delete

tests/test_abstasks/test_clustering_metrics.py
```diff
@@ -67,3 +76,162 @@ def test_end_to_end_result_dict_contains_both_metrics():
...
+    )
+    kept, kept_assignments = _run_with(
+        embeddings, labels, drop_unlabelled_documents=False
+    )
+
+    assert dropped == kept
+    assert dropped_assignments == kept_assignments
+
+
+def test_default_keeps_the_published_behaviour():
```
KennethEnevoldsen, line 204: delete

tests/test_abstasks/test_clustering_metrics.py
```diff
@@ -9,6 +9,15 @@
...
 from tests.mock_models import MockSentenceTransformer
 
+HIERARCHICAL_TASKS = [
+    "ArXivHierarchicalClusteringP2P",
+    "ArXivHierarchicalClusteringS2S",
+    "SNLHierarchicalClusteringP2P",
+    "SNLHierarchicalClusteringS2S",
+    "VGHierarchicalClusteringP2P",
+    "VGHierarchicalClusteringS2S",
+]
```
KennethEnevoldsen, line 19: please don't add any specific tests on these

tests/test_abstasks/test_clustering_metrics.py
KennethEnevoldsen, line 1: Can you please rereview these tests. I think a LLM might have gone a little rogue here. The tests should only test the implemented functionality

mteb/abstasks/clustering.py
```diff
@@ -131,6 +163,14 @@ class AbsTaskClustering(AbsTask):
...
         k_mean_batch_size: Batch size to use for k-means clustering.
         max_depth: Maximum depth to evaluate clustering. If None, evaluates all levels.
+        drop_unlabelled_documents: For a hierarchical task, whether a document whose label
+            path is shorter than the level being scored is dropped at that level rather than
+            scored. Those documents have nothing in common except a missing label, so keeping
+            them asks the model to find a class that is not really there. Defaults to False,
+            which is the behaviour every published hierarchical score was produced under.
+            New tasks that want the corrected behaviour set it to True; the .v2 hierarchical
+            tasks do. A task whose label paths all reach the same depth is unaffected either
+            way, since no document is ever dropped.
```
KennethEnevoldsen, line 173: This description is wrong

KennethEnevoldsen (changes requested):
thanks for the fix @arthi-arumugam-git, but this PR could use a read through. Currently, there seems too many mistakes for me properly review - can I ask you to do a critical read through

Also perfectly fine to use a AI but do disclose it and to what extent you have reviewed the code. Otherwise, it can end up taking up a lot of our time

KennethEnevoldsen:
@arthi-arumugam-git friendly ping here

arthi-arumugam-git:
Sorry for the silence, work has been stormy the last couple of weeks and this slipped. Thanks for the ping.

I went back through the whole thing rather than just patching the lines you marked.

The four tests at the bottom and the `HIERARCHICAL_TASKS` list: agreed, deleting them. They assert things about the task registry and about `superseded_by`, which is not what this file tests and not what this PR implements. That was scope creep and you were right to pull me up on it.

The `.v2` classes: I will redo them as plain metadata classes instead of subclasses, and put the reason for the version bump into the description field, following `SummEvalSummarization.v2`.

On the `drop_unlabelled_documents` description in `clustering.py`, I would rather fix the right thing than guess. Do you mean something in it is factually wrong, or that it is far too long for an API docstring? Reading it back, it is eight lines of prose sitting next to one-line entries, so my instinct is to cut it right down to a sentence or two. If there is a specific inaccuracy I have missed, tell me which part and I will fix that instead.

I will push the rework over the next few days and ping you here.

PR #5307, opened 2026-08-26
fix: add recording-level FleursT2ARetrieval.v2 / FleursA2TRetrieval.v2

### LINKED ISSUE #5270: FLEURS retrieval tasks use sentence-level IDs that collapse multiple audio recordings
## Observation
FLEURS `id` identifies a sentence, not a unique recording. Multiple speakers can have different audio files sharing the same `id`. `FleursT2ARetrieval` / `FleursA2TRetrieval` use this `id` as both corpus/query ID and qrel key: ([fleurs.py:245-261](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/retrieval/multilingual/fleurs.py#L245-L261)),

relevant_docs_ = {str(row["id"]): {str(row["id"]): 1} for row in split_dataset}
As a result, multiple recordings with the same sentence ID collapse into a single retrieval item.

## Evidence
Across all 102 FLEURS languages:
- 77,809 audio recordings
- 33,018 unique IDs
meaning many recordings share the same retrieval ID.

Example (kn_in, id=1956):

3 recordings
same transcript
different speakers
durations: 10.08s / 0.84s / 8.28s

## Impact
This may be harmless for sentence-level retrieval because recordings with the same sentence are interchangeable positives. However, it means the current implementation evaluates at sentence-ID level rather than recording level and silently drops recording diversity.

## Question
Should FLEURS retrieval tasks:
1. keep sentence-level evaluation intentionally, or
2. switch to recording-level IDs with multi-positive qrels?
Clotho handles similar one-to-many cases by creating unique IDs for each retrieval item, so wanted to discuss the intended granularity before proposing a PR.

lxer19:
## Summary
FLEURS id identifies a sentence, not a recording. Multiple recordings of the same sentence therefore share one id, causing both retrieval tasks to collapse recording-level examples.
The original FLEURS paper defines:
- T2A: retrieve any speaker recording of the queried sentence.
- A2T: retrieve the correct text segment for each recording.
v2 matches those semantics:
- T2A.v2: one sentence query → all recordings of that sentence are explicit positives.
- A2T.v2: every recording is an independent query → one unique sentence document.

### T2A — text query → find the audio
Before the fix, v1:
queries : id='1956'  (×3 identical rows, one per recording)
corpus  : id='1956'  ×3   ← three different .wav files, same id
qrels   : {'1956': {'1956': 1}}
issues: multiple recordings sharing the same corpus ID can produce multiple retrieval scores, but the final result dictionary can retain only one score for that ID.

After the fix, v2:
queries : id='1956'  (one row)
corpus  : id='1956-0'  17653886544018491782.wav  10.08s
          id='1956-1'  18023272269047514633.wav   0.84s
          id='1956-2'   5382544524698160008.wav   8.28s
qrels   : {'1956': {'1956-0': 1, '1956-1': 1, '1956-2': 1}}

Fixes #5270


lxer19:
Why I use load_data here:
1)  there is no transformed data to materialize. The audio is byte-identical between v1 and v2, so v2 is a deterministic relabeling, not a data transformation. 
2) Everything else follows from that — materializing would upload copies two and three of audio that already exists unchanged at mteb/fleurs (~130 GB after a ~65 GB download)

mteb/descriptive_stats/Image/Any2AnyRetrieval/FleursA2TRetrieval.v2.json
```diff
@@ -0,0 +1,3918 @@
...
+            }
+        },
+        "queries_video_statistics": null,
+        "relevant_docs_statistics": {
+            "num_relevant_docs": 77812,
+            "min_relevant_docs_per_query": 1,
+            "average_relevant_docs_per_query": 1.0000385559511111,
+            "max_relevant_docs_per_query": 2,
+            "unique_relevant_docs": 33018
+        },
```
Samoed, line 39: How did you calculate this? This statistics looks a bit odd and don't have new fields

mteb/tasks/retrieval/multilingual/fleurs.py
```diff
@@ -114,6 +114,87 @@
...
+        t2a_relevant_docs[sentence_id][recording_id] = 1
+
+    # A2T: the recording's own sentence. Two distinct sentence ids can carry identical
+    # text (one pair in `ln_cd`); those documents are indistinguishable, so both count.
+    text_to_sentence_ids: dict[str, list[str]] = defaultdict(list)
+    for sentence_id, text in sentence_text.items():
+        text_to_sentence_ids[text].append(sentence_id)
+    a2t_relevant_docs = {
+        recording_id: {s: 1 for s in text_to_sentence_ids[text]}
+        for recording_id, text in zip(recording_ids, transcriptions)
```
Samoed: Is there recordings with same id?

mteb/tasks/retrieval/multilingual/fleurs.py
```diff
@@ -114,6 +114,87 @@
...
+    ids = [str(i) for i in split_dataset["id"]]
+    transcriptions = split_dataset["transcription"]
+    recording_ids = _recording_ids(ids, _audio_paths(split_dataset))
+
+    audio_ds = split_dataset.select_columns(["audio"]).add_column("id", recording_ids)
+
+    # one document per unique sentence, in first-seen order
+    sentence_text: dict[str, str] = {}
+    for sentence_id, text in zip(ids, transcriptions):
+        sentence_text.setdefault(sentence_id, text)
```
Samoed: Why not?
```suggestion
        sentence_text[sentence_id] = text
```

mteb/tasks/retrieval/multilingual/fleurs.py
```diff
@@ -114,6 +114,87 @@
...
+    text_ds = Dataset.from_dict(
+        {"id": list(sentence_text), "text": list(sentence_text.values())}
+    )
+
+    # T2A: every recording of the sentence is a correct answer.
+    t2a_relevant_docs: dict[str, dict[str, int]] = defaultdict(dict)
+    for sentence_id, recording_id in zip(ids, recording_ids):
```
Samoed: Ids are not unique and this will be incorrect

Samoed (changes requested):

mteb/tasks/retrieval/multilingual/fleurs.py
lxer19: update.

mteb/tasks/retrieval/multilingual/fleurs.py
lxer19: sentence_id is not unique, but recording_id is: each recording gets {sentence_id}-{rank}, so each (sentence_id, recording_id) pair is unique. The code intentionally builds a one-to-many mapping from one sentence to all of its recordings. For example, three rows with id=1956 produce {"1956": {"1956-0": 1, "1956-1": 1, "1956-2": 1}}.

Samoed: Yes, I checked more and this looks good

mteb/descriptive_stats/Image/Any2AnyRetrieval/FleursA2TRetrieval.v2.json
lxer19: Recomputed via MTEB's own _calculate_descriptive_statistics_from_split, run once per language because compute_overall=True would materialize all 77,809 decoded clips at once (~65 GB of audio) and exceed available RAM. Script: https://github.com/lxer19/mteb/blob/31a30f46/scripts/fleurs_v2_descriptive_stats/rerun_official_stats.py
max = 2 and average = 1.0000385559511111 come from one edge case in ln_cd: two distinct sentence IDs have byte-identical transcriptions, so the 3 recordings for those sentences are correctly marked relevant to both indistinguishable documents.
Also rebased onto main; the newer stats schema adds num_missing_query_ids / num_missing_corpus_ids, and both are 0.

lxer19: While most `audio_id`s have only one relevant `sentence_id`, this edge case maps one `audio_id` to two `sentence_id`s:

```python
a2t_relevant_docs = {
    # most cases: {audio_id: {sentence_id: 1}}
    "1703-0": {"1703": 1, "1966": 1},  # edge case: 1 recording -> 2 sentence IDs
    "1966-0": {"1703": 1, "1966": 1},
    "1966-1": {"1703": 1, "1966": 1},
}
```
This is why `max_relevant_docs_per_query = 2` and the average is slightly above 1.

lxer19: This odd number is calculated by average_relevant_docs_per_query = 77,812 / 77,809 = 1.0000385559511111

lxer19:
Done @Samoed Samoed

mteb/descriptive_stats/Image/Any2AnyRetrieval/FleursA2TRetrieval.v2.json
Samoed: I just thought that this number should be much higher because we should have much more relevant documents. E.g. for `af_za` split in `mteb/fleurs` there are 264 rows and 233 unique id, so `average_relevant_docs_per_query` shouldn't be 1?

Samoed: I investigated a bit and only `t2a` should have more that 2 average_relevant_docs_per_query. I think this is correct then

Samoed (approved):

mteb/descriptive_stats/Image/Any2AnyRetrieval/FleursA2TRetrieval.v2.json
lxer19: Right. I checked the paper, it says "After this first recording step, each recording is evaluated by additional workers to assess whether the recording corresponds to the input sentence. Invalid recordings are discarded, leaving us between zero and three recordings per sentence in the final dataset." Now matches the statistics of average_relevant_docs_per_query in t2a.

PR #5313, opened 2026-08-27
Add DNA-VL-STEER-2B

dhk1349:
Adds `dnotitia/DNA-VL-STEER-2B`, a language-bias-calibrated variant of
`Qwen/Qwen3-VL-Embedding-2B`. Identical architecture and loader, so the meta inherits the
base entry's capacity fields; it differs in declaring all 36 calibrated languages.
`training_datasets=None` is accurate — the calibration corpus is translated CC12M
captions and contains no MTEB task data.

### Checklist

- [x] The model is public, i.e., the weights are publicly available to download
- [ ] I reproduced results on at least one benchmark — full `MIEB(Multilingual)` run in
  progress; table to follow.


dhk1349:
Description is now 1787 characters, under the 2000 limit. The `long PR desc` label is from the original submission — the check only runs on `opened`/`reopened` and doesn't remove it, so it needs clearing manually.

mteb/models/abs_encoder.py
Samoed, line 1: We know about this issue https://github.com/embeddings-benchmark/mteb/issues/5297. You shoudn't change this part of code

dhk1349:
Thanks — reverted the `abs_encoder.py` change and dropped the tests that went with it. This is now just the model meta.

mteb/models/abs_encoder.py
dhk1349: Thanks for your comment.

mteb/models/model_implementations/qwen3_vl_embedding_models.py
```diff
@@ -141,3 +141,71 @@ def get_task_instruction(
     citation=QWEN3_VL_EMBEDDING_CITATION,
     extra_requirements_groups=["multimodal-sbert"],
 )
+
+
+dna_vl_steer_2b = ModelMeta(
+    # Same architecture and loader as Qwen/Qwen3-VL-Embedding-2B; only the early
+    # residual-stream layers differ, so every capacity field is inherited unchanged.
```
ayush1298: there is `adapted_from` field in ModelMeta where you can add these

dhk1349: added adapted_from="Qwen/Qwen3-VL-Embedding-2B" param.

Samoed (approved):

PR #5315, opened 2026-08-28
Add papr-embed-v1 API models

shawkatkabbara:
Adds six closed-weights `papr-embed-v1` entries scored through the public Papr embeddings API (`https://memory.papr.ai`). Companion to embeddings-benchmark/results#691. Harness and raw run artifacts: https://github.com/Papr-ai/papr-embed-evals

| name | encoder | variant |
|---|---|---|
| `papr/papr-embed-v1-0.6b-scifact` | Qwen3-0.6B | SciFact-specialized |
| `papr/papr-embed-v1-0.6b-scifact-reasoning-frontier-max` | Qwen3-0.6B | + request-time teacher bands |
| `papr/papr-embed-v1-0.6b` | Qwen3-0.6B | general |
| `papr/papr-embed-v1-0.6b-reasoning-frontier-max` | Qwen3-0.6B | general + teacher bands |
| `papr/papr-embed-v1-4b` | Qwen3-4B | general |
| `papr/papr-embed-v1-4b-reasoning-frontier-max` | Qwen3-4B | general + teacher bands |

Notes:
- Auth is `PAPR_API_KEY` (https://dashboard.papr.ai).
- The `*-reasoning-frontier-max` entries are systems rather than plain encoders: they call a hosted reasoning model at request time, so they are registered under separate names.
- `training_datasets` lists the six MTEB tasks whose **train** splits are in the fine-tune mix. No test split is trained on.

SciFact NDCG@10 (mteb 2.10.0, test split, production API):

| model | NDCG@10 |
|---|---:|
| 0.6b-scifact | 0.72990 |
| 0.6b-scifact-reasoning-frontier-max | 0.73147 |
| 0.6b | 0.71476 |
| 0.6b-reasoning-frontier-max | 0.71516 |
| 4b | 0.78777 |
| 4b-reasoning-frontier-max | 0.78326 |

- [x] Loads via `mteb.get_model(name, revision)` and `mteb.get_model_meta(name, revision)`
- [x] The model is public (available as an API)
- [x] Results included above

### DIFF
--- a/mteb/models/model_implementations/papr_models.py
+++ b/mteb/models/model_implementations/papr_models.py
@@ -21,10 +21,10 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-# Per-task holographic schema served by the public API. The instruction is
-# applied server-side from this schema, so the encoder MUST send raw query
-# text (batch["query"]), never mteb's already-instructed batch["text"].
-# Reproducing the submitted SciFact numbers requires this mapping.
+# Band schema per task. This is model configuration, not a prompt: the schema
+# names the 14 frequency bands the encoder conditions its export gate on. The
+# two -scifact models pin their schema server-side and ignore this; the general
+# models have no default and require one.
 TASK_SCHEMA_IDS: dict[str, str] = {
     "SciFact": "biomedical:scifact:2.0.0",
 }
@@ -51,11 +51,6 @@ class PaprEmbedAPIModel(AbsEncoder):
 
     Auth: ``PAPR_API_KEY`` (https://dashboard.papr.ai). Optional
     ``PAPR_BASE_URL`` overrides the default production endpoint.
-
-    Query batches use ``batch["query"]`` (raw text). Sending ``batch["text"]``
-    would double-instruct every query because the server already applies the
-    schema instruction. That is the same rule the public eval harness uses
-    (https://github.com/Papr-ai/papr-embed-evals).
     """
 
     def __init__(
@@ -97,7 +92,7 @@ class PaprEmbedAPIModel(AbsEncoder):
         input_type: Literal["query", "document"] = (
             "query" if _is_query(prompt_type) else "document"
         )
-        texts = _collect_texts(inputs, input_type)
+        texts = _collect_texts(inputs)
         if not texts:
             return np.zeros((0, self._embed_dim), dtype=np.float32)
 
@@ -139,9 +134,7 @@ class PaprEmbedAPIModel(AbsEncoder):
             except urllib.error.URLError as exc:
                 last_error = f"transport error: {exc}"
             time.sleep(min(60.0, 2.0**attempt))
-        raise RuntimeError(
-            f"Papr embeddings API failed after retries ({last_error})"
-        )
+        raise RuntimeError(f"Papr embeddings API failed after retries ({last_error})")
 
 
 def _is_query(prompt_type: PromptType | None) -> bool:
@@ -151,15 +144,10 @@ def _is_query(prompt_type: PromptType | None) -> bool:
     return str(value).lower() == "query"
 
 
-def _collect_texts(
-    inputs: DataLoader[BatchedInput], input_type: str
-) -> list[str]:
+def _collect_texts(inputs: DataLoader[BatchedInput]) -> list[str]:
     texts: list[str] = []
     for batch in inputs:
-        if input_type == "query" and batch.get("query") is not None:
-            texts.extend(str(text) for text in batch["query"])
-        else:
-            texts.extend(str(text) for text in batch["text"])
+        texts.extend(str(text) for text in batch["text"])
     return texts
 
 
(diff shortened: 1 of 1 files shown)


mteb/models/model_implementations/papr_models.py
```diff
@@ -0,0 +1,263 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Per-task holographic schema served by the public API. The instruction is
+# applied server-side from this schema, so the encoder MUST send raw query
+# text (batch["query"]), never mteb's already-instructed batch["text"].
+# Reproducing the submitted SciFact numbers requires this mapping.
+TASK_SCHEMA_IDS: dict[str, str] = {
+    "SciFact": "biomedical:scifact:2.0.0",
+}
```
Samoed: I don't think that instructions should be applied on server side

mteb/models/model_implementations/papr_models.py
```diff
@@ -0,0 +1,263 @@
...
+
+def _collect_texts(
+    inputs: DataLoader[BatchedInput], input_type: str
+) -> list[str]:
+    texts: list[str] = []
+    for batch in inputs:
+        if input_type == "query" and batch.get("query") is not None:
+            texts.extend(str(text) for text in batch["query"])
+        else:
+            texts.extend(str(text) for text in batch["text"])
```
Samoed: I don't think you need quuery field. In most cases query field is same as text
```suggestion
        texts.extend(str(text) for text in batch["text"])
```

mteb/models/model_implementations/papr_models.py
```diff
@@ -0,0 +1,263 @@
...
+            batches.append(current)
+            current, chars = [], 0
+        current.append(text)
+        chars += len(text)
+    if current:
+        batches.append(current)
+    return batches
+
+
+def _papr_meta(
```
KennethEnevoldsen: no need to make a function to make the metadata, that just complicated things on our end

mteb/models/model_implementations/papr_models.py
```diff
@@ -0,0 +1,263 @@
...
+            reasoning=reasoning,
+        ),
+        max_tokens=2048,
+        embed_dim=embed_dim,
+        open_weights=False,
+        n_parameters=n_parameters,
+        n_embedding_parameters=None,
+        memory_usage_mb=None,
+        license=None,
+        reference="https://github.com/Papr-ai/papr-embed-evals",
```
KennethEnevoldsen: Models need a model card or release blogpost. I don't believe this constitute a model card.

shawkatkabbara:
Thanks both! the instruction point turned out to be a real design bug on our
side rather than a wrapper detail, so we fixed it at the API level.

@Samoed re: server-side instructions. You were right, and it was worse than
a style issue. The API resolved the query instruction from `band_schema_id`,
which conflated two unrelated things: the schema id is a task/domain identifier
for the encoder, while the instruction is the retrieval prompt. The failure mode
was silent — a schema with no instruction mapping encoded the bare query and
still returned a perfectly normal-looking vector.

`instruction` is now an explicit request field: required for `input_type="query"`,
rejected for documents, and applied exactly as given. The server never chooses
it. This wrapper passes MTEB's own `self.get_instruction(task_metadata, prompt_type)`,
so the prompt a model is scored with is the task's registered one, visible here
rather than picked server-side.

No submitted number moves. Passing the string the schema would have resolved to
reproduces the previously formatted text byte for byte, which we pin with a test.
Verified end to end against the deployed service: a different instruction changes
the embedding (cos 0.955), `""` gives a genuine bare encode (cos 0.899), and
repeated calls are byte-identical.

**@Samoed re: `batch["query"]`.** Applied — `_collect_texts` now reads `batch["text"]`.

**@KennethEnevoldsen re: the metadata factory.** Removed; the six `ModelMeta`
objects are inlined.

**@KennethEnevoldsen re: the model card.** Agreed — `reference` currently points
at an evals repo and that isn't a model card. We're preparing a release post to
point at instead, which matches how the other closed-weight API models here are
referenced (OpenAI, Cohere and Voyage each link a company post with
`open_weights=False`). It's gated on an internal review, so realistically a few
weeks rather than days.

Two ways to handle that, whichever you prefer: keep this open until the post is
live, or merge the code now and we follow up with a one-line `reference` change
once it publishes. The second would also unblock
embeddings-benchmark/results#691, which is otherwise waiting on these models
existing in the registry.

Also addressed: the PR description is now under the 2000-character limit.

mteb/models/model_implementations/papr_models.py
```diff
@@ -0,0 +1,344 @@
...
+logger = logging.getLogger(__name__)
+
+# Band schema per task. This selects the 14 frequency bands the encoder
+# conditions its export gate on, and nothing else -- the query instruction is a
+# separate request field that mteb fills from the task metadata. The two
+# -scifact models pin their schema server-side and ignore this; the general
+# models have no default and require one.
+TASK_SCHEMA_IDS: dict[str, str] = {
+    "SciFact": "biomedical:scifact:2.0.0",
+}
```
Samoed, line 31: > instruction is now an explicit request field: required for input_type="query",
rejected for documents, and applied exactly as given. The server never chooses
it. This wrapper passes MTEB's own self.get_instruction(task_metadata, prompt_type),
so the prompt a model is scored with is the task's registered one, visible here
rather than picked server-side.

Why do you still have this then?

KennethEnevoldsen:
I will close this as stale, given that there has been no response for two weeks

PR #5322, opened 2026-08-29
dataset: add Multi30k multilingual image-text retrieval (t2i, i2t)

nahatav:
Adds `Multi30kT2IRetrieval` and `Multi30kI2TRetrieval` (en/cs/de/fr). Part of #4842.

## Why

Multi30k is not in `mteb` in either direction. It adds to the multilingual coverage discussed in #4842. The de/fr/cs captions are independent human translations of the same Flickr30k image rather than machine translations, so the four subsets are directly comparable.

## Data

`romrawinjp/multi30k` @ `110e827`, MIT. Test split: 1000 images, one caption per language, 340 MB. The image side is identical across subsets, so the loader builds it once per split and varies only the caption side; `add_column` is used so image bytes are never decoded. 1:1 qrels in both directions.

Both tasks are added to `KNOWN_ISSUES["duplicate_image"]` for the same reason as `XM3600T2IRetrieval` and `XFlickr30kCoT2IRetrieval`: sharing one image set across language subsets is by design.

## Results

nDCG@10:

| Model | dir | en | fr | de | cs |
| --- | --- | ---: | ---: | ---: | ---: |
| clip-vit-base-patch32 | t2i | 0.7433 | 0.3767 | 0.2297 | 0.0461 |
| clip-vit-base-patch32 | i2t | 0.7598 | 0.4292 | 0.3148 | 0.0880 |
| random-encoder | t2i | 0.0063 | 0.0067 | 0.0042 | 0.0037 |
| random-encoder | i2t | 0.0051 | 0.0064 | 0.0048 | 0.0041 |

English scoring high confirms the task is wired correctly. Czech sits 16x below English while still ~12x above the random floor, so the task is discriminative rather than saturated for the lower-resource languages.

## Checklist

- [x] I have tested that the dataset runs with the `mteb` package
- [x] `openai/clip-vit-base-patch32` (substituted for `intfloat/multilingual-e5-small`, which is text-only)

### DIFF
--- a/mteb/tasks/retrieval/multilingual/multi30k_retrieval.py
+++ b/mteb/tasks/retrieval/multilingual/multi30k_retrieval.py
@@ -1,6 +1,7 @@
-from datasets import Dataset, DatasetDict, Image, load_dataset
+from datasets import Dataset, Image, load_dataset
 
 from mteb.abstasks.retrieval import AbsTaskRetrieval
+from mteb.abstasks.retrieval_dataset_loaders import RetrievalSplitData
 from mteb.abstasks.task_metadata import TaskMetadata
 
 _LANGUAGES = {
@@ -36,13 +37,7 @@ _BIBTEX = r"""
 """
 
 
-def _load_multi30k_data(
-    path: str,
-    langs: list[str],
-    splits: list[str],
-    direction: str,
-    revision: str | None = None,
-):
+def _load_multi30k(task: AbsTaskRetrieval, direction: str) -> None:
     """Load Multi30k into MTEB retrieval format.
 
     Multi30k stores one row per image with parallel captions in four columns
@@ -51,11 +46,14 @@ def _load_multi30k_data(
     per split and shared. `direction` selects which side is the query: "t2i"
     retrieves images from captions, "i2t" retrieves captions from images.
     """
-    corpus = {lang: dict.fromkeys(splits) for lang in langs}
-    queries = {lang: dict.fromkeys(splits) for lang in langs}
-    relevant_docs = {lang: dict.fromkeys(splits) for lang in langs}
+    if task.data_loaded:
+        return
 
-    for split in splits:
+    path = task.metadata.dataset["path"]
+    revision = task.metadata.dataset["revision"]
+    task.dataset = {}
+
+    for split in task.metadata.eval_splits:
         data = load_dataset(path, split=split, revision=revision)
         row_ids = [str(i) for i in range(len(data))]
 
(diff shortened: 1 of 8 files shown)


mteb/tasks/retrieval/multilingual/multi30k_retrieval.py
```diff
@@ -0,0 +1,177 @@
...
+        modalities=["text", "image"],
+        sample_creation="found",
+        bibtex_citation=_BIBTEX,
+    )
+
+    def load_data(self, num_proc: int | None = None, **kwargs) -> None:
+        if self.data_loaded:
+            return
+
+        self.corpus, self.queries, self.relevant_docs = _load_multi30k_data(
```
Samoed: Can you change to use `self.dataset` instead with format https://github.com/embeddings-benchmark/mteb/blob/1c9514dcbf9190d32d78df2b8d76563ca98c016b/mteb/abstasks/retrieval_dataset_loaders.py#L28

mteb/tasks/retrieval/multilingual/multi30k_retrieval.py
```diff
@@ -0,0 +1,177 @@
...
+        image_side = image_side.add_column("text", [None] * len(row_ids))
+        image_side = image_side.cast_column("image", Image())
+
+        for lang in langs:
+            text_side = Dataset.from_dict(
+                {
+                    "id": [f"text-{i}" for i in row_ids],
+                    "text": data[lang],
+                    "modality": ["text"] * len(row_ids),
+                    "image": [None] * len(row_ids),
```
Samoed:
```suggestion
```

mteb/tasks/retrieval/multilingual/multi30k_retrieval.py
```diff
@@ -0,0 +1,177 @@
...
+    for split in splits:
+        data = load_dataset(path, split=split, revision=revision)
+        row_ids = [str(i) for i in range(len(data))]
+
+        # Built once and reused across languages. add_column is used rather
+        # than map so the image bytes are never decoded and re-encoded.
+        image_side = data.select_columns(["image"])
+        image_side = image_side.add_column("id", [f"image-{i}" for i in row_ids])
+        image_side = image_side.add_column("modality", ["image"] * len(row_ids))
+        image_side = image_side.add_column("text", [None] * len(row_ids))
```
Samoed:
```suggestion
```

nahatav:
Thanks @Samoed — both addressed.

**`self.dataset` / `RetrievalSplitData`:** the loader now writes
`task.dataset[lang][split] = RetrievalSplitData(queries=..., corpus=...,
relevant_docs=..., top_ranked=None)` instead of returning the three
`corpus`/`queries`/`relevant_docs` dicts. `load_data` on both classes is now a one-liner
delegating to the shared `_load_multi30k(self, direction)`, and the `DatasetDict`
wrapping is gone.

**The two suggestions:** dropped the `modality` columns and the `None` filler columns on
both sides. They were only there to satisfy the old format; with `RetrievalSplitData` the
modality follows from the columns actually present, so the image side is now just
`id` + `image` and the text side just `id` + `text`.

Scores are unchanged.

Samoed (commented):
Where did you found sources for datast? Can you provide script for your dataset transformations

nahatav:
Source is `romrawinjp/multi30k` on the Hub, pinned at revision `110e827`. It is a third-party upload of Multi30k (Elliott et al. 2016; Barrault et al. 2018), and there is no transformation script because there is no transformation step: the task loads it directly and reshapes into retrieval format at runtime in `load_data`. Happy to materialise that into an mteb-owned dataset with a `scripts/data/` script instead if you would prefer

Samoed:
Yes, script for that would help

nahatav:
Done, added `scripts/data/multi30k_retrieval/create_data.py` and published the reshaped form in the standard retrieval format, so the task file is now metadata only with no custom `load_data`. Scores are unchanged, clip-vit-base-patch32 on en still gives 0.7433 t2i and 0.7598 i2t, and the descriptive statistics are byte-identical

ayush1298:
@Samoed I guess good to merge?

mteb/tasks/retrieval/multilingual/multi30k_retrieval.py
```diff
@@ -0,0 +1,100 @@
...
+        },
+        type="Any2AnyMultilingualRetrieval",
+        category="i2t",
+        eval_splits=["test"],
+        eval_langs=_LANGUAGES,
+        main_score="ndcg_at_10",
+        date=("2016-01-01", "2018-12-31"),
+        domains=["Scene", "Web", "Written"],
+        task_subtypes=["Image Text Retrieval"],
+        license="mit",
```
Samoed: In `romrawinjp/multi30k` wrong license. 
```suggestion
        license="cc-by-sa-4.0",
```

mteb/tasks/retrieval/multilingual/multi30k_retrieval.py
```diff
@@ -0,0 +1,100 @@
...
+@inproceedings{elliott2016multi30k,
+  address = {Berlin, Germany},
+  author = {Elliott, Desmond and Frank, Stella and Sima'an, Khalil and Specia, Lucia},
+  booktitle = {Proceedings of the 5th Workshop on Vision and Language},
+  pages = {70--74},
+  publisher = {Association for Computational Linguistics},
+  title = {Multi30K: Multilingual English-German Image Descriptions},
+  year = {2016},
+}
+"""
```
Samoed, line 46: You should also add https://aclanthology.org/W17-4718/ (french language addition)

isaac-chung:
Merging, based on @Samoed 's last 2 comments.

PR #5325, opened 2026-08-29
dataset: add XM3600 image-to-text multilingual retrieval

guangyusong:
Adds `XM3600I2TRetrieval`, the missing image-to-text direction for multilingual XM3600: given an image, retrieve any matching caption. It reuses the pinned [`mteb/xm3600`](https://huggingface.co/datasets/mteb/xm3600/tree/536cd45bbfe53de9b08c0483bb4a76a4bd3673fa) test set unchanged: 36 languages with 3,600 image queries per language.

| Model | nDCG@10 | hit@1 |
|---|---:|---:|
| Random | 0.0007 | 0.0002 |
| Multilingual SigLIP | 0.6091 | 0.5627 |

The original [XM3600 paper](https://aclanthology.org/2022.emnlp-main.45/) evaluates caption generation rather than retrieval. The [SigLIP paper](https://arxiv.org/abs/2303.15343) reports **54.1%** mean image-to-text R@1 on XM3600; this run reaches **56.27%**. Both count any matching caption at rank 1.

- [x] I have outlined why this dataset is filling an existing gap in `mteb`.
- [x] I have run the following models on the task and added the results to the PR:
  - [x] `google/siglip-base-patch16-256-multilingual`
- [x] I have considered the dataset size; the existing fixed 3,600-image evaluation set per language is retained.
- [x] I compared with the closest published same-data retrieval result above.

Related to embeddings-benchmark/mteb#4842


mteb/tasks/retrieval/multilingual/xm3600_i2t_retrieval.py
Samoed, line 1: Better to move to other xm3600 dataset version to keep in same file

guangyusong: Thanks. Updated

Samoed (approved):

Issue #5063, opened 2026-08-01
Make a low-dependency version of of MTEB


KennethEnevoldsen:
### Description of the feature

Currently, for results analysis or setting up the API you need to install ~2gb of stuff (e.g. torch).

My suggestion would be to split out `mteb-core`, which includes everything that doesn't require torch. That way the `mteb-api` could also be split out.

I am not sure this is possible, but I suspect it could be, though it would require some work.

ayush1298:
Below is some analysis that I have done (measured on main, macOS CPU venv): 

- import `mteb cold`: 8.6 s, 6684 modules, torch in sys.modules
- Cumulative import cost: `mteb.abstasks` 3.65 s · `mteb.models` 2.53 s (sentence_transformers 2.51 s) · mteb.types 1.50 s (torch)
- Disk: torch 363 MB (CPU-only wheel; ~2 GB with the CUDA nvidia-* wheels), transformers 52 MB, scipy 81 MB, sklearn 35 MB
- `mteb/api/`: zero real torch/ST usage 

So the API and results-analysis paths genuinely don't need torch today; they only inherit it transitively.

So, i think we can package `mteb` as a `metapackage` so import mteb never changes
1. `mteb-core` - distribution that ships the `mteb` package. Light deps only:` numpy, pydantic, polars, pandas, huggingface_hub, requests, pyyaml, packaging, tqdm, rich.`
2. `mteb-core[run]` - adds `torch, sentence-transformers, transformers, datasets, scikit-learn, scipy, pytrec-eval-terrier.`
3. `mteb` - becomes a `metapackage` with no code: `dependencies = ["mteb-core[run]==<version>"].`

Net effect: `pip install mteb` behaves exactly as today,` import mteb` keeps working for everyone, no import paths move, and `pip install mteb-core` gives the light install. `mteb-api` then just depends on `mteb-core` + the api extra. This avoids the fragile alternative of two distributions writing into the same `mteb/` directory.

Broadly, 3 types of changes are required to avoid `torch` being reachable from `import mteb` directly: 
1. Move heavy top-level imports to function-local / TYPE_CHECKING imports. `torch, SentenceTransformer, CrossEncoder, AutoConfig `are imported at module scope in places where they're only used inside a function body or only in an `TYPE_CHECKING` annotation. Modules like `mteb.types` and `mteb.models.model_meta` don't need torch at import time.
[...]

PR #5329, opened 2026-08-29
 fix: Make it such that torch is not a required import of mteb.types

ayush1298:
First PR in a series making mteb importable without torch, so results analysis, the leaderboard and the API don't need a ~2 GB install. Refs #5063.

This one only touches mteb.types. No public API changes, no behaviour changes — import mteb still pulls in torch through `mteb.models` and `mteb.abstasks,` which are follow-ups.

`mteb/types/_encoder_io.py` had three separate torch dependencies, each needing different handling:

1. import torch → moved under TYPE_CHECKING.
2. OutputDType.get_dtype() → function-local import, so torch is only needed when the method is actually called.
3. Array → a runtime assignment, so from __future__ import annotations doesn't defer it. It now has a numpy-only fallback in the else branch of TYPE_CHECKING, keeping the precise | torch. Tensor type for type checkers.

-  `import mteb.types` no longer loads torch: 1.21 s → 0.64 s, 2265 → 1536 modules.
- No runtime cost: a cached import torch inside get_dtype() measures ~30 ns of overhead, against ~75 µs for the tensor cast in the same function.
- `mkdocs.yml` — filters `__getattr__` out of the generated API docs. Verified with griffe that Array still renders with its docstring; __getattr__ would otherwise have appeared as a public member.
- import mteb.types no longer loads torch — 1536 modules, 1.14s (was pulling the full torch stack).
- New tests/test_lazy_imports.py guards this: one test asserts torch stays out of sys.modules, another asserts Array still equals its previous definition. Confirmed the guard actually fails by temporarily re-adding the eager import torch.


mteb/types/_encoder_io.py
```diff
@@ -31,8 +31,23 @@ class EncodeKwargs(TypedDict):
...
 
 # --- Output types ---
-Array = NDArray[np.floating | np.integer | np.bool_] | torch.Tensor
-"""General array type, can be a numpy array (float, int, or bool) or a torch tensor."""
+if TYPE_CHECKING:
+    Array: TypeAlias = NDArray[np.floating | np.integer | np.bool_] | torch.Tensor
+    """General array type, can be a numpy array (float, int, or bool) or a torch tensor."""
+
+
+def __getattr__(name: str) -> Any:
```
Samoed: I don't think we should customize getattr (I even didn't this is possible to do that`). I think you can create a fallback in else branch of `TYPE_CHECKING` for array, e.g. `Array: TypeAlias = NDArray[np.floating | np.integer | np.bool_]`

ayush1298: made changes

mteb/types/_encoder_io.py
```diff
@@ -31,8 +31,14 @@ class EncodeKwargs(TypedDict):
 
 
 # --- Output types ---
-Array = NDArray[np.floating | np.integer | np.bool_] | torch.Tensor
-"""General array type, can be a numpy array (float, int, or bool) or a torch tensor."""
+if TYPE_CHECKING:
+    Array: TypeAlias = NDArray[np.floating | np.integer | np.bool_] | torch.Tensor
```
KennethEnevoldsen, line 35: how does this influence the docs?

KennethEnevoldsen (approved):

tests/test_ensure_no_torch_at_import.py
KennethEnevoldsen, line 1: I would rename to:

```
tests/test_ensure_no_torch_at_import.py
```

mteb/types/_encoder_io.py
KennethEnevoldsen: (I think this might be the biggest remaining issue)

tests/test_ensure_no_torch_at_import.py
ayush1298: done

mteb/types/_encoder_io.py
ayush1298: I checked this by building the docs both ways and diffing the HTML and the rendered docs are unchanged.

`mkdocstrings` builds its model via `griffe,` and `griffe` resolves `TYPE_CHECKING` as `true.` So it reads the if branch and ignores the else — `Array` is still recorded as `NDArray[...] | torch.Tensor` with its original docstring. The else branch is invisible to the docs; it only exists for the interpreter.

But, again all this depends on `griffe` treating `TYPE_CHECKING` as `true` which is current behaviour. If it ever changed, `Array` would render as the numpy-only alias with no description, since the docstring sits in the if branch.

ayush1298:
Lintter CI failure is not relevant to change in these PR

mteb/types/_encoder_io.py
KennethEnevoldsen: great - then I will resolve this (if it changes we can handle it then)

KennethEnevoldsen:
@Samoed anything to add?

Samoed (approved):

ayush1298:
Can I merge?

KennethEnevoldsen:
@ayush1298 can you fix the linting (such that we don't introduce it in main) - otherwise you are free to merge

ayush1298:
> @ayush1298 can you fix the linting (such that we don't introduce it in main) - otherwise you are free to merge

https://github.com/embeddings-benchmark/mteb/blob/main/mteb/tasks/retrieval/dan/danrag_retrieval.py#L65-L70
Its already fixed in main, maybe I have to merge main. Should I or we can merge?

KennethEnevoldsen:
yeah update from main  should solve this issue here (we don't want to merge a failing test)

ayush1298:
merged main, it would be fixed now

PR #5331, opened 2026-08-30
dataset: add GLAMI-1M retrieval

guangyusong:
Adds `GLAMI1MT2IRetrieval` and `GLAMI1MI2TRetrieval` for multilingual product image-text retrieval. They use all unique image-text pairs from the authors’ [unmodified test archive](https://huggingface.co/datasets/glami/glami-1m/blob/befda45d8d4e8b8082bb8a1912d1f9eb9483991c/GLAMI-1M-dataset--test-only.zip) across 13 languages. The I2T task keeps every matching product text as relevant.

| Direction | Model | nDCG@10 | hit@1 |
|---|---|---:|---:|
| T2I | Random | 0.0007 | 0.0001 |
| T2I | Multilingual SigLIP | 0.6820 | 0.5168 |
| I2T | Random | 0.0008 | 0.0003 |
| I2T | Multilingual SigLIP | 0.7025 | 0.5390 |

The paper evaluates classification rather than retrieval, so there is no same-task paper score.

- [x] I have outlined why this dataset is filling an existing gap in `mteb`.
- [x] I have run the following models on the task and added the results above:
  - [x] `google/siglip-base-patch16-256-multilingual`
- [x] I have considered the dataset size and retained the official test set.
- [x] Original-paper reproduction is not applicable because the paper evaluates classification.

Related to embeddings-benchmark/mteb#4842.


mteb/tasks/classification/multilingual/glami_1m.py
```diff
@@ -0,0 +1,69 @@
...
+    "lav-Latn",
+    "lit-Latn",
+    "ron-Latn",
+    "slk-Latn",
+    "slv-Latn",
+    "spa-Latn",
+    "tur-Latn",
+]
+
+_DATASET_REVISION = "93bfcc674ea6e3837fe3e404ee2f0ebfb2b429ed"
```
Samoed: Move inside metadata

scripts/data/glami_1m/create_data.py
Samoed, line 1: What is source of data?

scripts/data/glami_1m/create_data.py
```diff
@@ -0,0 +1,328 @@
...
+    indices = np.arange(len(labels))
+    np.random.RandomState(SEED).shuffle(indices)
+    counts: Counter[int] = Counter()
+    selected = []
+    for index in indices:
+        label = labels[int(index)]
+        if counts[label] < TRAIN_PER_LABEL:
+            selected.append(int(index))
+            counts[label] += 1
+    return sorted(selected)
```
Samoed: Is train big? I think you can upload it fully and we would sample it in classification class

mteb/tasks/classification/multilingual/glami_1m.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class GLAMI1MMultimodalClassification(AbsTaskClassification):
```
Samoed: Wodn't be better to use this dataset as retrieval task with descriptions?

Samoed:
Can you upload these zip archives to hf without modifications?

guangyusong:
> Can you upload these zip archives to hf without modifications?

Thanks, updated this to text-to-image retrieval using the product names and descriptions. The authors already host the unmodified test archive [here](https://huggingface.co/datasets/glami/glami-1m/blob/befda45d8d4e8b8082bb8a1912d1f9eb9483991c/GLAMI-1M-dataset--test-only.zip), so the preparation script pins that exact revision. The formatted dataset is [here](https://huggingface.co/datasets/artist/glami-1m-t2i-mteb).

Samoed:
> using the product names and descriptions

I think better to split them to separete columns. You can put name to the title and description as text, but this would work for corpus split only, for queries you can combine them during loading. I see there problem in bg split that alsmost all text have `2-4 работни дни` which just "2-4 work days"

guangyusong:
> I think better to split them to separete columns. You can put name to the title and description as text, but this would work for corpus split only, for queries you can combine them during loading. I see there problem in bg split that alsmost all text have `2-4 работни дни` which just "2-4 work days"

Done! Queries now store the product name as `title` and the description as `text`, then combine them during loading. I kept the source descriptions unchanged.

mteb/tasks/retrieval/multilingual/glami_1m_retrieval.py
Samoed, line 1: I think would be good to have also `i2t` version

guangyusong: Added the i2t version and updated the PR description with its Random and SigLIP results.

Samoed (approved):

PR #5335, opened 2026-08-30
Enables `RET` (flake8-return) ruff rule

ayush1298:
refs #2416.
Enables `RET` (flake8-return) ruff rule. `RET504` (unnecessary-assign) is **not** enabled — it accounts for 117 of the 197 hits and removes named intermediate results that often aid readability. It's in `ignore` with a comment.

| Rule | n | Fix |
|---|---|---|
| `RET505` superfluous-else-return | 62 | Dropped `else` after `return` |
| `RET503` implicit-return | 7 | See below |
| `RET507` superfluous-else-continue | 6 | Autofix |
| `RET506` superfluous-else-raise | 4 | Autofix |
| `RET502` implicit-return-value | 1 | Autofix |

118 fixes were applied rather than 80: de-indenting an `else` block exposes more of the same pattern nested within, and ruff iterates to a fixpoint.

The 7 `RET503` were fixed at the cause rather than by appending `return None`:

- A `-> None` method was propagating `super()`'s return value.
- Two functions dispatching on `Literal["wide", "long"]` used two separate `if`s, so an unexpected value silently returned `None`. The second branch is now the exhaustive fall-through.
- Two untyped retry loops get an explicit `return None` (behaviour-identical).
- One `-> list[float]` coroutine raises instead, since `None` would contradict the annotation and the tail is unreachable.
- A `-> str` test helper returned `None` for models with no `n_parameters`, serialising as `null`; it now returns a descriptive string.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -335,6 +335,7 @@ select = [
     "B",       # flake8-bugbear
     "PT",      # flake8-pytest-style
     "A",       # flake8-builtins, don't shadow builtins
+    "RET",     # flake8-return
 ]
 
 ignore = [
@@ -358,6 +359,7 @@ ignore = [
     "PLW0717",  # too-many-statements-in-try-clause
     "ISC001",   # single-line implicit concat; conflicts with `ruff format`
     "PT018",    # composite assertions; splitting `assert a and b` is not always clearer
+    "RET504",   # unnecessary-assign; named intermediate results often aid readability
 ]
 future-annotations = true  # For TC rules
 typing-modules = ["typing_extensions", "mteb.types", "collections.abc"]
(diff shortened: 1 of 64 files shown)


pyproject.toml
```diff
@@ -358,6 +359,7 @@ ignore = [
     "PLW0717",  # too-many-statements-in-try-clause
     "ISC001",   # single-line implicit concat; conflicts with `ruff format`
     "PT018",    # composite assertions; splitting `assert a and b` is not always clearer
+    "RET504",   # unnecessary-assign; named intermediate results often aid readability
```
Samoed, line 362: Is this produce a lot of changes?

ayush1298: yes, around 117

ayush1298: and also ruff will not autofix them I guess

Samoed (approved):

PR #5338, opened 2026-08-30
feat: add sentence transformers multi vector models 

Samoed:
Add support for `MultiVectorEncoder` which was released in [sentence transformers v6](https://github.com/huggingface/sentence-transformers/releases/tag/v6.0.0)
Example:
```python
import mteb
from sentence_transformers import MultiVectorEncoder

model = MultiVectorEncoder("lightonai/GTE-ModernColBERT-v1")
tasks = mteb.get_tasks(tasks=["AskUbuntuDupQuestions"])
results = mteb.evaluate(model, tasks=tasks)
```

Created 2 classes like for pylate models:
- MultiVectorWrapper for wrapping aroung encode functions 
- MultiVectorSearchEncoderWrapper for wrapping around searh pipeline


docs/get_started/usage/defining_the_model.md
```diff
@@ -63,10 +63,42 @@ on `MTEB`:
...
+    model = MultiVectorEncoder("lightonai/GTE-ModernColBERT-v1")
+
+    # MultiVectorEncoder (ColBERT-style late interaction) produces a sequence of per-token
+    # embeddings per input rather than a single fixed-size vector, so it only supports
+    # retrieval-shaped tasks (Retrieval, Reranking)
+    tasks = mteb.get_tasks(tasks=["AskUbuntuDupQuestions"])
+    results = mteb.evaluate(model, tasks=tasks)
+    ```
+
+    !!! warning
```
KennethEnevoldsen, line 97:
```suggestion
    !!! note
```

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -687,3 +713,352 @@ def encode(
...
+            num_proc=num_proc,
+        )
+
+
+class MultiVectorWrapper(MultiVectorSearchEncoderWrapper):
+    """Loads and encodes sentence-transformers' native `MultiVectorEncoder` models.
+
+    Combines with the inherited `MultiVectorSearchEncoderWrapper` mixin to implement
+    [SearchProtocol][mteb.models.SearchProtocol] -- mirroring `MultiVectorModel(PylateSearchEncoder)`
+    for PyLate models. Registered as the `loader` for `MultiVectorEncoder`-backed models
```
KennethEnevoldsen, line 914: Does it make sense to unify sentence transformers and pylate models to reduce dependencies? (similar to what we did to model2vec)

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -687,3 +713,352 @@ def encode(
             concatenate_batches=_concatenate_sparse_batches,
             **kwargs,
         )
+
+
+class MultiVectorSearchEncoderWrapper:
+    """Mixin class to add brute-force MaxSim indexing and search to a MultiVectorEncoder-backed encoder.
```
KennethEnevoldsen, line 719: I am not sure I get why it is "brute force"?

KennethEnevoldsen: (plaid I assume?)

mteb/models/sentence_transformer_wrapper.py
```diff
@@ -687,3 +713,352 @@ def encode(
...
+    (used the same way by `MultiVectorModel` for PyLate models): `index()`/`search()` call
+    `self._encode(...)` on the same object, so the concrete class combines this mixin with its own
+    model-loading and `_encode` -- see `MultiVectorWrapper`, the class that does so for
+    sentence-transformers' native `MultiVectorEncoder`.
+
+    Unlike `PylateSearchEncoder` (which builds a PLAID approximate-nearest-neighbor index via the
+    `pylate` package), this scores the full corpus -- or pre-ranked candidates -- directly against
+    each query with the model's own MaxSim `similarity()`, chunked over the corpus (via
+    `corpus_chunk_size`) to bound memory. This trades PLAID's approximate, sub-linear search for an
+    exact one with no dependency beyond sentence-transformers.
```
KennethEnevoldsen, line 731: Hmm it makes it hard to compare two models in different frameworks (but agree that the no-dependency is nice).

how big of a performance/memory difference is this?

KennethEnevoldsen (commented):
We will need "what's new" docs. Otherwise I am happy to see these changes -- maybe worth making some comparisons with pylate to figure out how to best handle the different implementations

mteb/models/sentence_transformer_wrapper.py
tomaarsen: For context, all PyLate models load out of the box with Sentence Transformers, and ST MVE models don't load in PyLate, as of right now. Furthermore, ST requires ST v6.0.0+ for MVE, and PyLate requires ST v5.3.0 exactly.

Ideally, I'd like to help update PyLate so they both use the new ST MVE backbone instead of 2 different save formats. Then it might also be easier to loosen PyLate's strict ST version requirement.

mteb/models/sentence_transformer_wrapper.py
tomaarsen: It might honestly be too slow / memory costly to go non-PLAID. For context, one of the ColBERT authors himself, Omar Khattab, only evaluates late interaction models with PLAID. 

In my opinion, `fast-plaid` (https://github.com/lightonai/fast-plaid) is a very solid project that lowers the memory/disk space costs a lot, while being a lot faster than PLAID, while PLAID itself is already like 5x faster than exact brute force.

It'll always be interesting to see the difference in speed / reported NDCG, but I believe PLAID with nbits=4 (fast-plaid default) is effectively lossless.

mteb/models/sentence_transformer_wrapper.py
Samoed: Yes, we can do that, but pylate utilize fastplaid index, which speedup evaluations

mteb/models/sentence_transformer_wrapper.py
Samoed: Agree, fastplaid is used in pylate out of the box and faster. I can run some tests with lateon models

tomaarsen: It might be worthwhile to use fastplaid everywhere.

KennethEnevoldsen: Might be worth looking into (we have to reproduce some results, would prefer not to rerun everything if we do that), but for now I think we can start with it here and then add an issue on it

tomaarsen: I agree.

tomaarsen:
Bumping to try and get rid of the stale label 🤗 

- Tom Aarsen

PR #5340, opened 2026-08-31
Add CrisisMMD image-text classification tasks

PranitChawla:
## Summary

- add CrisisMMD informative (2-class) and humanitarian (5-class) image+tweet tasks
- introduce the `it2c` image+text-to-category direction
- register both tasks and add descriptive statistics

The official `text_img_agreed_lab` partitions retain only samples whose independently annotated image and tweet labels agree.

## Dataset

https://huggingface.co/datasets/pranitchawla/crisismmd-mteb at `a935fda3cee8a5250746a8d1d6278214d615cc89`

| Config | Classes | Train | Validation | Test |
|---|---:|---:|---:|---:|
| informative | 2 | 9,601 | 1,573 | 1,534 |
| humanitarian | 5 | 6,126 | 998 | 955 |

The artifact pins QCRI's source revision and agreed-label archive checksum. It verified 12,708 unique images and checked image/tweet IDs for cross-split leakage.

## Reference results

Accuracy / macro-F1 (%):

| Protocol | Model | Informative | Humanitarian |
|---|---|---:|---:|
| 16-shot | MTEB random | 50.10 / 48.01 | 19.52 / 15.81 |
| 16-shot | CLIP ViT-B/32 | 89.32 / 87.92 | 79.56 / 68.69 |
| 16-shot | SigLIP Base | 88.85 / 87.47 | 78.99 / 67.74 |

Scores are means over five 16-shot experiments using frozen embeddings and a fitted classifier. Macro-F1 is important for the imbalanced binary split.

## Validation

- metadata, registry, and dataset-quality tests passed
- Ruff, formatting, and diff checks passed
- both tasks load and all three reference models run at the pinned revision

## License

QCRI's official Hub release declares CC BY-NC-SA 4.0. Its legacy CrisisNLP page also links separate research-data terms.

Related to #4842.


mteb/tasks/classification/eng/crisismmd.py
```diff
@@ -0,0 +1,103 @@
...
+  title = {Analysis of Social Media Data using Multimodal Deep Learning for Disaster Response},
+  year = {2020},
+}
+"""
+
+
+class _CrisisMMDClassification(AbsTaskClassification):
+    input_column_name: ClassVar[tuple[str, str]] = ("image", "text")
+    samples_per_label: int = 16
+    n_experiments: int = 5
```
Samoed: Please remove class and list these parameters for each class

mteb/tasks/classification/eng/crisismmd.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class CrisisMMDInformativeClassification(_CrisisMMDClassification):
```
Samoed: Why do you separete them into 2 tasks?

PranitChawla: They represent two distinct official CrisisMMD annotation tasks: informativeness is binary, while humanitarian categorization has five classes. They also use different dataset configurations and example sets. Keeping them separate gives each taxonomy its own classifier and interpretable score instead of aggregating unrelated label spaces. This also follows the existing MTEB pattern for datasets with multiple prediction targets. Would you prefer a different structure here?

Samoed (approved):

Issue #2669, opened 2025-05-07
Add Vaani dataset


silky1708:

KennethEnevoldsen:
@silky1708 are you taking this? Otherwise I think we need more info

silky1708:
@anime-sh has been working on this related PR: https://github.com/embeddings-benchmark/mteb/pull/2367

PR #5342, opened 2026-08-31
dataset: add Vaani multilingual audio-image retrieval

nahatav:
Adds `VaaniA2IRetrieval` and `VaaniI2ARetrieval` over **62 Indian languages** (Indo-Aryan, Dravidian, Tibeto-Burman, Munda). Part of #4842; #2669.

## Gap

`mteb/tasks/retrieval/multilingual/` has 42 files: 9 image, 5 audio, 0 video. There is no multilingual audio↔image task in the repo.

Vaani records *image-prompted* speech: a speaker is shown a photograph and describes it aloud, so every clip is grounded in a specific image and both directions need no extra annotation. arXiv:2603.28714, CC-BY-4.0.

## Construction

Audio and images sit in separate trees (~3.8 TB) linked only by a `referenceImage` filename, so `create_data.py` indexes all 289,676 images by name (path column only, never the bytes) and fetches just the referenced ones. Sampling is speaker-capped and spread across row groups, since Vaani records long sessions per speaker. Languages under 25 usable clips are dropped: **62 of 84 kept, 6,056 clips, 3,552 images**, an evaluation sample of a 31,255-hour source.

**Split:** Vaani ships only `train`. These are eval-only tasks, so the sample is published as `test`; no training data is retained.

**Corpora:** a2i uses one shared image corpus across subsets (each recording maps to exactly one image, so labels are unambiguous). i2a keeps the corpus **per language**, because 709 of the 3,552 images are described in more than one language, so a pooled recording corpus would leave correct answers from other subsets labelled irrelevant.

Both are added to `KNOWN_ISSUES["duplicate_image"]`, as `XM3600T2IRetrieval` is: the shared corpus is by design.

## Checklist

- [x] `mteb/baseline-random-encoder` on the full task → ~0.00, the floor for a 3,552-item corpus
- [ ] Second encoder not run yet: 3,552 images across 62 subsets. Happy to add one before merge.
- [x] No duplicate audio, images or pairs; all qrels resolve


Samoed (commented):
This seems more like traning dataset rather than evaluation

nahatav:
Vaani ships only a `train` split, so there is nothing pre-held-out to use. This is a 6,056-clip evaluation sample drawn from a 31,255-hour source, published as `test` because these are eval-only tasks and no training data is retained. Around 105 tasks already in mteb evaluate on a `train` split, though happy to carve the sample by district if you want it separated from data likely used in ASR training

isaac-chung:
> Second encoder not run yet: 3,552 images across 62 subsets. Happy to add one before merge.

Could you please run an omni model?

PR #5343, opened 2026-08-31
dataset: add VATEX multilingual (en/zh) video retrieval

nahatav:
Adds `VATEXMultilingualT2VRetrieval` and `VATEXMultilingualV2TRetrieval`. Part of #4842.

## Gap

Every video task in `mteb` sits under `eng/` or `zxx/`, so there is no multilingual video task. @AdnanElAssadi asked me to check whether the existing VATEX covers Chinese; it does not.

The existing English VATEX tasks use the **test** split. VATEX publishes Chinese captions only for **validation**, and the two are disjoint, with zero id overlap against the 5,162 hosted test clips, so the Chinese side needs its own build. **993 validation clips** carry both languages and share one video corpus, so the two subsets differ only in caption language.

## Two source issues corrected

**Untrimmed videos.** The archive contains whole source videos, not the annotated 10s windows: sampled clips ran 15s–406s against a 10s annotation, so a caption would describe a few percent of the retrieved video. Every clip is cut to its annotated range. 993 of 1,000 trimmed cleanly; 7 raised a muxer error on re-encode and were dropped.

**Malformed URLs.** The `url` column of `lmms-lab/vatex_from_url` appends the whole clip id to a watch URL, where only the leading part is the video id and the rest is the second range. The id is parsed instead.

## Licence

`not specified`, matching the existing VATEX tasks. VATEX's CC-BY-4.0 covers captions and annotations only. The authors state they cannot redistribute the videos, which remain subject to YouTube terms.

Both tasks join `KNOWN_ISSUES["duplicate_video"]`: the en and zh subsets share one video corpus by design, as `XM3600T2IRetrieval` does for images.

X-CLIP is English-only, and its 32x en/zh gap on t2v is what this task exists to measure.

## Checklist

- [x] Runs with `mteb` (993 queries, corpus and qrels per subset)
- [x] `mteb/baseline-random-encoder`: t2v 0.0050, v2t 0.0044
- [x] `microsoft/xclip-base-patch32`: t2v en 0.4716, zh 0.0146; v2t en 0.6135, zh 0.0349


Samoed (changes requested):
I don't think that we want to redistribute videos without clear license

nahatav:
Worth noting the repo already does this: `mteb/VATEX_test_1k` is in the mteb org, hosts 1,000 VATEX clips, declares no license, and the eight existing English VATEX tasks use it with `license="not specified"`, which is what I set here for the same reason.

I can't reuse those clips for the Chinese side, unfortunately: they're the test split, and VATEX only publishes Chinese captions for validation, so there's zero id overlap.

Happy to go whichever way you prefer. Just let me know.

Samoed:
I think we missed that. CC @KennethEnevoldsen @AdnanElAssadi

KennethEnevoldsen:
Yeah, license concerns was brought up in the review of MVEB (fairly so and addressed). I can't see VATEXV2TRetrieval used in an existing benchmark, but I would not want to knowingly include unlicensed datasets into MOEB. 

@AdnanElAssadi let me know what you think?

AdnanElAssadi:
I found this on their [website](https://eric-xw.github.io/vatex-website/about.html).

```
The dataset is under a [Creative Commons Attribution 4.0 International License](https://creativecommons.org/licenses/by/4.0/).
Contact the VATEX team by [vatex.org@gmail.com](mailto:vatex.org@gmail.com).
```

I don't think most datasets have specific licenses to video; I would assume most of them are public youtube videos. Is the license above not enough for us?

Samoed:
Yes, but they don't have videos in their dataset

AdnanElAssadi:
> Yes, but they don't have videos in their dataset

yeah, that's a limitation for most video datasets; they dont wanna have copyright issues because of youtube

isaac-chung:
@AdnanElAssadi @KennethEnevoldsen @Samoed 

Are we a go or no-go for this dataset?

Samoed:
I don't think we want to upload videos without clear license

AdnanElAssadi:
We could maybe just have it in mteb but not include as part of final benchmark?

isaac-chung:
> We could maybe just have it in mteb but not include as part of final benchmark?

I'd prefer to lean this way too

PR #5258, opened 2026-08-20
task: add visual-only Stanford I2V retrieval

tommasocerruti:
## Summary

Adds `StanfordI2VVisualRetrieval`, a pure image-to-video (`i2v`) retrieval task that fills the zero-merged image→video gap tracked in #4842. Sources: [official release](https://purl.stanford.edu/zx935qw7203) and [paper](https://andrefaraujo.github.io/files/papers/2015-03-18-mmsys-stanfordi2v.pdf).

The original relevance annotations were visual-only: annotators marked the video intervals where the query image was visible in the frames. Audio was not used, so `i2v` faithfully represents the original benchmark.

This is a visual-only sibling of `StanfordI2VRetrieval`, merged in #5151 as `i2va`. The existing task remains unchanged. This additional visual-only direction was discussed and approved by @AdnanElAssadi56.

## Data and metrics

Uses the same official 600K variant as #5151: 229 image queries, 3,401 newscast videos, and 1,280 binary judgments. The source’s 2,260 temporal annotations are collapsed to unique scene-level judgments following its scorer. Main score: `map_at_100`.

Frozen data: [`Cerru02/Stanford-I2V-600K@30ba417`](https://huggingface.co/datasets/Cerru02/Stanford-I2V-600K/tree/30ba417f4ed92c7035abd552cdef1ebfb3542186).

The task reuses the existing queries, corpus, and qrels without rebuilding or re-uploading the dataset. It removes the corpus audio column before encoding, leaving only image queries and video documents.

## Results

| Model | MAP@100 |
| --- | ---: |
| `mteb/baseline-random-encoder` | 0.00213 |
| `Qwen/Qwen3-VL-Embedding-2B` | Not reported — full local run blocked by compute |

The random baseline completed the full task over all 229 queries and 3,401 videos in 49m40s.

The full-task Qwen command completed all 229 query images but reached only 3 of 3,401 corpus videos after 4m01s on Apple MPS. The measured remaining time ranged from 35 to 76 hours, so the process was stopped. No reduced, sampled, or partial result is reported as a full score. I am happy to run the complete evaluation if suitable GPU compute is available.

The encoder-facing data was explicitly checked: queries contain only `id,image`, and the corpus contains only `id,video`. Qwen never receives the audio column.

The two remaining checklist limitations follow the precedent established by the merged sibling task in #5151. This PR uses the same dataset and evaluation protocol; its only intentional difference is removing audio before encoding to expose the original visual-only `i2v` direction.

## Dataset checklist

[...]

Issue #5223, opened 2026-08-16
Add dataset: AVCaps


yaswanth169:
### dataset link on Hugging dataset

https://huggingface.co/datasets/TUT-ARG/AVCaps

### Arxiv link

https://ieeexplore.ieee.org/document/11029114/

### Description of the dataset

Part of the MOEB tracker (#4842) — audio+image+text+video (all four modalities) open slot.

AVCaps (Sudarsanam, Martin-Morato, Hakala, Virtanen — IEEE Open Journal of Signal Processing 2025) is derived from VidOR: 2,061 video clips (28.8 hours), each with up to 5 audio-only captions, 5 visual-only captions, and 5 audio-visual captions (crowdsourced separately — annotators for the single-modality captions worked without access to the other modality), plus 3 GPT-4-synthesized audio-visual captions. CC-BY-NC-4.0, not gated — real hosted video confirmed (test_videos.zip 1.02GB / train_videos.zip 8.05GB, actual MP4 content verified, not just filenames).

Plan: build from the 200-clip test split (1GB, no need to touch the larger train split) — video clips natively provide the video+audio modalities, with a keyframe extraction covering image (same approach as other MOEB video tasks that derive image data from video), and the modality-specific captions covering text. Will report exact task formulation (retrieval direction(s), corpus/query split) and construction script in the PR.

PR #5344, opened 2026-08-31
dataset: add AVCaps audio-visual retrieval (6 directions)

nahatav:
Adds six directions over AVCaps: a2t/t2a, v2t/t2v and va2t/t2va. Part of #4842; issue #5223.

@yaswanth169 — you opened that issue; happy to hand this back if you're already on it.

## Gap

AVCaps captions each clip **three separate ways** — from the audio alone, from the visuals alone, and from both together. That is what makes it useful here: the audio-only, video-only and combined directions can be scored independently on identical clips, instead of being inferred from one caption set that mixes the modalities.

## Construction

Official **test** split only — 200 clips; train and validation are untouched, so the evaluation set matches the split the authors held out.

The published clips are muxed mp4s. Audio is demuxed to a separate mono 16 kHz stream, and **each task exposes only the media columns its captions actually describe**, so an audio-caption task cannot be quietly solved off the video track.

Repeated caption text is dropped (725/788/808 → 715/780/802). Annotators sometimes write the same short sentence for different clips, which would leave an identical query marked relevant to only one of the clips it describes — this was failing `test_task_quality.py` before the dedup.

Paper: IEEE OJSP 2025, 6:691–704. Licence **CC-BY-NC-4.0**, matching the source.

## Checklist

- [x] Tested that it runs with `mteb` (all six directions)
- [x] `mteb/baseline-random-encoder` on the full task → a2t 0.0188, t2a 0.0388, v2t 0.0083, t2v 0.0171, va2t 0.0096, t2va 0.0210
- [ ] Second encoder — not run: built on CPU-only hardware, where an omni encoder over six directions runs to hours. Same rationale as #5258, which merged with the random baseline only. Happy to add one before merge.
- [x] Size considered — official test split, no reduction needed

### DIFF
--- a/mteb/descriptive_stats/Image/Any2AnyRetrieval/AVCapsA2TRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyRetrieval/AVCapsA2TRetrieval.json
@@ -1,0 +1,42 @@
+{
+    "test": {
+        "num_samples": 915,
+        "num_queries": 200,
+        "num_documents": 715,
+        "number_of_characters": 44045,
+        "documents_text_statistics": {
+            "total_text_length": 44045,
+            "min_text_length": 19,
+            "average_text_length": 61.6013986013986,
+            "max_text_length": 270,
+            "unique_texts": 715
+        },
+        "documents_image_statistics": null,
+        "documents_audio_statistics": null,
+        "documents_video_statistics": null,
+        "queries_text_statistics": null,
+        "queries_image_statistics": null,
+        "queries_audio_statistics": {
+            "total_duration_seconds": 10432.394125,
+            "min_duration_seconds": 20.0145625,
+            "average_duration_seconds": 52.161970625,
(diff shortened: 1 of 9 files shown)


Samoed (approved):

Samoed:
Can you run some models to see scores?

nahatav:
nDCG@10 on the full task:

| direction | random | CLAP | X-CLIP |
|---|---|---|---|
| a2t | 0.0188 | **0.0639** | |
| t2a | 0.0388 | **0.1357** | |
| v2t | 0.0083 | | **0.4133** |
| t2v | 0.0171 | | **0.3128** |
| va2t | 0.0096 | | **0.4060** |
| t2va | 0.0210 | | **0.3165** |

`laion/clap-htsat-fused` on the audio directions and `microsoft/xclip-base-patch32` on the visual ones, each being the matching model class. X-CLIP has no audio encoder, so its va2t/t2va numbers are video-only and a lower bound for those two directions

PR #5347, opened 2026-08-31
fix: count single-colour images and flag unanswerable queries

lxer19:
## Issue

Some image retrieval tasks contain fully black or fully white images. These can
be valid content, but they are useful diagnostic signals for potential data
quality issues.

An offline audit of 2.2M images found examples in WIT and VisualNews:
https://github.com/lxer19/mteb/blob/c242be8e427e9ae28c3a002f7ff44d5f94c4af70/scripts/validation/validate_constant_image_claims.py

Examples include:
- `VisualNewsI2TRetrieval`: 14 black/white query images.
- `WITT2IRetrieval`: 21 black/white corpus images that are positive qrels.
- `VisualNewsT2IRetrieval`: 14 queries whose positive image qrels are all black/white.

These findings are treated as warnings rather than correctness failures because
a black or white image may still be valid benchmark content.

## Fix

Add two diagnostic statistics:

- `black_or_white_images`: counts fully black or fully white images.
- `queries_with_all_gold_black_or_white`: counts retrieval queries whose
  positive image qrels (`score > 0`) all point to black/white images.

Both checks are warnings in `test_dataset_quality`; they do not fail the task.


mteb/abstasks/_statistics_calculation.py
```diff
@@ -125,10 +125,61 @@ def calculate_text_statistics(
...
+    constant_doc_ids: Container[str],
+) -> int:
+    """Number of queries whose *entire* gold set is constant images.
+
+    A constant document is only an evaluation defect when the query has nothing
+    else to retrieve. Corpora routinely carry a handful of blank images that no
+    qrel references, and class-judged tasks give a query hundreds of positives
+    where one blank changes nothing; neither is a broken query. Requiring the
+    whole gold set to be constant is what separates those from a query that
+    genuinely cannot be answered.
```
Samoed: Why do you thing that they cannot be answered. There might be cases when constant are expected and correct

mteb/abstasks/_statistics_calculation.py
```diff
@@ -418,13 +475,19 @@ def calculate_relevant_docs_statistics(
         unique_relevant_docs=len(qrel_corpus_ids),
         num_missing_query_ids=len(qrel_query_ids.difference(query_ids)),
         num_missing_corpus_ids=len(qrel_corpus_ids.difference(corpus_ids)),
+        queries_with_all_gold_constant=(
```
Samoed: This field actually computes only images, not any modality. Name is a bit misleading

Samoed (changes requested):

KennethEnevoldsen (commented):
couldn't a constant color image be a valid target, e.g. "give the yellow used for the minions in the minion movie", but probably rarely.

I think the special cases are probably black and white, which are very likely just "blank" (maybe worth just limiting to those two?) 

I am not sure this is the most valuable metric to compute, but maybe akin to empty or single word docs.

lxer19:
Good point. I’ll narrow the quality check to pure black/white images, while keeping the qrels-aware logic for corpus images. This avoids treating valid constant-colour targets as defects.

mteb/abstasks/_statistics_calculation.py
```diff
@@ -125,10 +125,80 @@ def calculate_text_statistics(
...
+        if gold and all(doc_id in black_or_white_doc_ids for doc_id in gold):
+            broken += 1
+    return broken
+
+
 def calculate_image_statistics(
     images: list[Image.Image],
     hashes: list[str] | None = None,
     max_workers: int | None = None,
+    black_or_white_flags: list[bool] | None = None,
```
KennethEnevoldsen, line 201: do we need a flag here? Shouldn't we just always do it?

tests/test_tasks/test_task_quality.py
```diff
@@ -1099,6 +1098,23 @@ def _image_field_quality(
...
 ) -> tuple[int, list[tuple[str, str]]]:
     errors: list[tuple[str, str]] = []
+
+    # A pure black or white image carries no visual signal, but whether that breaks
+    # the evaluation depends on what the column *is*. A query or a labelled sample
+    # is itself the unit being evaluated, so a blank one can never be answered or
+    # learned from. A document is only unretrievable when it is the whole of some
+    # query's gold set, which qrels decide, not this column -- that case is
+    # reported separately as `queries_with_all_gold_black_or_white`. `.get` because
+    # stats generated before this field existed simply omit it.
```
KennethEnevoldsen, line 1108:
```suggestion
```

tests/test_tasks/test_task_quality.py
```diff
@@ -1099,6 +1098,23 @@ def _image_field_quality(
...
+    black_or_white_images = stats.get("black_or_white_images")
+    if black_or_white_images and not _is_document_field(field):
```
KennethEnevoldsen, line 1110:
```suggestion
    if black_or_white_images:
```

tests/test_tasks/test_task_quality.py
```diff
@@ -1332,6 +1348,21 @@ def _relevant_docs_integrity_quality(
             )
         )
 
+    # A black/white document only breaks evaluation when a query has nothing else
+    # to retrieve. Corpora carry blank images no qrel references, and class-judged
+    # tasks give a query hundreds of positives where one blank changes nothing;
+    # neither is a defect. Only a wholly-black/white gold set is.
```
KennethEnevoldsen, line 1354:
```suggestion
```

tests/test_abstasks/test_image_statistics.py
KennethEnevoldsen, line 1: There are way too many tests here. Please rereview this part of the code yourself. I suspect you would need just 1 good test

KennethEnevoldsen (changes requested):
still some work to be done here

mteb/abstasks/_statistics_calculation.py
lxer19: I updated them as warnings

mteb/abstasks/_statistics_calculation.py
lxer19: remove them

tests/test_tasks/test_task_quality.py
lxer19: remove

tests/test_tasks/test_task_quality.py
lxer19: clean

tests/test_tasks/test_task_quality.py
lxer19: clean

tests/test_abstasks/test_image_statistics.py
lxer19: Thanks for the feedback. keep only 1 test, just make sure the count of black/white images is as expected

lxer19:
> still some work to be done here

@KennethEnevoldsen I have cleaned the codes

lxer19:
The Windows job failure is unrelated to this PR: `test_model_implementations_declare_no_import_time_torch_dtypes` (added in #5441) calls `path.read_text()` without `encoding="utf-8"`, so on Windows it hits `UnicodeDecodeError` on files with non-ASCII prompts such as `bge_models.py`. All Linux jobs and lint pass.

ayush1298:
@lxer19 can you merge main here, failing windows CI is fixed now

KennethEnevoldsen (approved):
I think this is now good to merge

lxer19:
@Samoed any feedback?

PR #5350, opened 2026-08-31
fix: make model_meta and _set_seed torch-free

ayush1298:
Second PR in the series making mteb importable without torch. Refs #5063. Follows #5329.

import mteb is still unchanged — torch now enters via `mteb/models/__init__.py` and the model registry, which is the next PR. This is a prerequisite step.

`mteb/models/model_meta.py`: Dropped the eager `sentence_transformers` and `transformers` imports. Of the nine uses of `SentenceTransformer / CrossEncoder / AutoConfig`, only three are runtime uses and the rest are annotations, already deferred by `from __future__ import annotations`:

1. `SentenceTransformer / CrossEncoder` → moved to `TYPE_CHECKING` satisfies six annotation sites (dropping it gives 6 mypy name-defined errors), while a function-local import in `_get_n_embedding_parameters_from_sentence_transformers` for its two isinstance checks (dropping it passes mypy but raises NameError at runtime). 
2. `AutoConfig` → function-local import in `_from_hub.`
3. Checked that none of the affected methods are pydantic validators, so pydantic never resolves their signatures.
4. Performance improvement: 3.19 s → 0.20 s, 4709 → 697 modules. model_meta now imports with no heavy dependencies at all — that's this PR composing with #5385, which deferred the datasets import in mteb.types; neither change reaches zero on its own.

`mteb/_set_seed.py`: `AbsTask.__init__` seeds every task it constructs and `get_tasks()` constructs them all, so this sits on the metadata-only path. A plain lazy import wasn't enough — the function had to become torch-optional, or a torch-free install would crash on `get_tasks().` Used the existing `_is_package_available` helper so that an installed-but-broken torch still raises, instead of silently leaving a run unseeded. 
Performance improvement: 0.53 s → 0.05 s, 1116 → 278 modules.

Testing:
Generalised the helper in tests/test_ensure_no_torch_at_import.py to stub arbitrary parent packages, and added five tests (7 total). Beyond the import guards, these check that the lazy imports didn't break behaviour.


mteb/_set_seed.py
Samoed, line 1: I'm not sure that without pytorch set_seed make sense

ayush1298: We're not removing torch. We are just changing when it's imported (module scope → first call), so the else branch is currently unreachable. On whether seeding makes sense without torch: what callers actually consume is the return value, `(random.Random, np.random.Generator) — self.rng_state, self.np_rng`. Seeding torch and TF are global side effects on top of that. So without torch, it still seeds everything that exists and skips the one thing that doesn't, same as the `tensorflow` block already does just below.

`AbsTask.__init__` seeds every task, and `get_tasks()` constructs them all — so this sits on the metadata-only path and would be the first thing to break once `mteb-core` drops torch. The availability check (rather than except ImportError) is so an installed-but-broken torch still raises instead of silently leaving a run unseeded.

ayush1298:
@KennethEnevoldsen Can you have a look whenever you get time?

isaac-chung (approved):
Looks good to me. 

Would be great if you could also share how one can reproduce the performance improvements.

ayush1298:
script is: 
<details>
<summary> script </summary>

```py
"""Measure how long importing parts of `mteb` takes, and which heavy deps they pull in.

Used to verify the lazy-import work (https://github.com/embeddings-benchmark/mteb/issues/5063).

Each measurement runs in a fresh subprocess, because imports are cached process-wide. Parent
packages are stubbed out with an empty module of the same `__path__`, so that e.g. `mteb.types`
can be imported without executing `mteb/__init__.py` (which still pulls torch in via other
modules). That isolates the module under test from the rest of the package.

Usage:
    # measure the working tree
    python scripts/bench_import_time.py

    # A/B the working tree against another ref (checked out into a temp git worktree)
    python scripts/bench_import_time.py --compare-to origin/main
"""

import argparse
import json
import shutil
import statistics
import subprocess
import sys
import tempfile
from pathlib import Path

REPO_ROOT = Path(__file__).resolve().parent.parent

# (module to import, parent packages to stub out so the module can be imported on its own)
TARGETS: list[tuple[str, list[str]]] = [
    ("mteb.types", ["mteb"]),
    ("mteb._set_seed", ["mteb"]),
    ("mteb.models.model_meta", ["mteb", "mteb.models"]),
    ("mteb", []),  # the full package, for context
]

HEAVY_DEPS = (
    "torch",
    "sentence_transformers",
    "transformers",
    "datasets",
    "sklearn",
    "scipy",
)

_CHILD = """
import json, pathlib, sys, time, types

root, module, stubs, heavy = pathlib.Path(sys.argv[1]), sys.argv[2], json.loads(sys.argv[3]), json.loads(sys.argv[4])
sys.path.insert(0, str(root))
for name in stubs:
    stub = types.ModuleType(name)
    stub.__path__ = [str(root.joinpath(*name.split(".")))]
    sys.modules[name] = stub

start = time.perf_counter()
__import__(module)
elapsed = time.perf_counter() - start

print(json.dumps({
    "seconds": elapsed,
    "modules": len(sys.modules),
    "heavy": [d for d in heavy if d in sys.modules],
}))
"""


def measure(root: Path, module: str, stubs: list[str], repeats: int) -> dict | None:
    """Import `module` from `root` in a fresh interpreter, `repeats` times; keep the fastest."""
    runs = []
    for _ in range(repeats + 1):  # +1 warm-up: the first run pays cold filesystem cache
        proc = subprocess.run(
            [
                sys.executable,
                "-c",
                _CHILD,
                str(root),
                module,
                json.dumps(stubs),
                json.dumps(list(HEAVY_DEPS)),
            ],
            capture_output=True,
            text=True,
            cwd=root,
        )
        if proc.returncode != 0:
            print(f"  {module}: FAILED\n{proc.stderr.strip()[:400]}", file=sys.stderr)
            return None
        runs.append(json.loads(proc.stdout.strip().splitlines()[-1]))
    best = min(runs[1:], key=lambda r: r["seconds"])
    best["median"] = statistics.median(r["seconds"] for r in runs[1:])
    return best


def fmt(result: dict | None) -> str:
    if result is None:
        return "failed"
    heavy = ", ".join(result["heavy"]) or "none"
    return f"{result['seconds']:5.2f}s  {result['modules']:5d} modules  heavy: {heavy}"


def main() -> int:
    parser = argparse.ArgumentParser(description=__doc__)
    parser.add_argument(
        "--compare-to", metavar="REF", help="git ref to use as the baseline"
    )
    parser.add_argument(
        "--repeats", type=int, default=3, help="measured runs per target (default: 3)"
    )
    args = parser.parse_args()

    if not args.compare_to:
        print(f"working tree: {REPO_ROOT}\n")
        for module, stubs in TARGETS:
            print(
                f"  {module:32s} {fmt(measure(REPO_ROOT, module, stubs, args.repeats))}"
            )
        return 0

    # A branch that is behind the baseline will show *regressions* for anything the baseline
    # already improved, which looks like a bug in this script but is really a stale branch.
    behind = subprocess.run(
        ["git", "rev-list", "--count", f"HEAD..{args.compare_to}"],
        cwd=REPO_ROOT,
        capture_output=True,
        text=True,
        check=False,
    ).stdout.strip()
    if behind.isdigit() and int(behind) > 0:
        print(
            f"warning: the working tree is {behind} commits behind {args.compare_to}.\n"
            f"         Numbers may be confounded by changes already on {args.compare_to}.\n"
            f"         Merge it in, or compare against your branch point instead.\n",
            file=sys.stderr,
        )

    tmp = Path(tempfile.mkdtemp(prefix="mteb-bench-"))
    worktree = tmp / "baseline"
    subprocess.run(
        [
            "git",
            "worktree",
            "add",
            "--detach",
            "--quiet",
            str(worktree),
            args.compare_to,
        ],
        cwd=REPO_ROOT,
        check=True,
    )
    try:
        print(f"baseline: {args.compare_to}   vs   working tree\n")
        for module, stubs in TARGETS:
            before = measure(worktree, module, stubs, args.repeats)
            after = measure(REPO_ROOT, module, stubs, args.repeats)
            print(f"  {module}")
            print(f"    before  {fmt(before)}")
            print(f"    after   {fmt(after)}")
            if before and after and after["seconds"] > 0:
                print(
                    f"    -> {before['seconds'] / after['seconds']:.1f}x faster, "
                    f"{before['modules'] - after['modules']} fewer modules"
                )
            print()
    finally:
        subprocess.run(
            ["git", "worktree", "remove", "--force", str(worktree)],
            cwd=REPO_ROOT,
            check=False,
        )
        shutil.rmtree(tmp, ignore_errors=True)
    return 0


if __name__ == "__main__":
    raise SystemExit(main())
```

</details>

to run: 

`python scripts/bench_import_time.py --compare-to origin/main`

and adding results again, a bit detail: 


```md
baseline: origin/main   vs   working tree

  mteb._set_seed
    before   0.53s   1116 modules  heavy: torch
    after    0.05s    278 modules  heavy: none
    -> 10.1x faster, 838 fewer modules

  mteb.models.model_meta
    before   3.19s   4709 modules  heavy: torch, sentence_transformers, transformers, datasets, sklearn, scipy
    after    0.20s    697 modules  heavy: none
    -> 16.2x faster, 4012 fewer modules

  mteb.types
    before   0.08s    365 modules  heavy: none
    after    0.08s    365 modules  heavy: none

  mteb
    before   4.15s   6799 modules  heavy: torch, sentence_transformers, transformers, datasets, sklearn, scipy
    after    3.96s   6799 modules  heavy: torch, sentence_transformers, transformers, datasets, sklearn, scipy

```

isaac-chung:
Nice, but I don't see this script in this PR or in main. Could you add it here please?

ayush1298:
> Nice, but I don't see this script in this PR or in main. Could you add it here please?

I thought we dont have to include it in the PR. So added script in above comment only, foldable dropdown

isaac-chung:
The ultimate all-time goal here is reproducibility. Anything we do here is to further that goal.

The script documents an important caveat (stubs parent packages and does not follow real mmteb import path). I think here it's ok not to commit it. 

Generally, for import times, we can create flame graphs, e.g. with https://pypi.org/project/flameprof/ A script that produces such graphs and analyses the full path could be committed in the future.

PR #5257, opened 2026-08-20
dataset: add ManiSkill i2v/v2i robot manipulation retrieval (MOEB)

ZhixuLi-the-Runner:
Part of MOEB (#4842). Adds two **image ↔ video** tasks:

- `ManiSkillI2VRetrieval` — goal-state image → manipulation videos (https://huggingface.co/datasets/ZhixuLi/ManiSkill-I2V)
- `ManiSkillV2IRetrieval` — the reverse (https://huggingface.co/datasets/ZhixuLi/ManiSkill-V2I)

**Gap**: second pair for the sparse image ↔ video direction; queries and documents never share a viewpoint.

**Construction** (`scripts/build_maniskill_datasets.py`): built from scratch from the official ManiSkill3 motion-planning demos — 8 tabletop tasks × 150 successful episodes, replayed via env states, rendered at 256×256 from both viewpoints. Per-task disjoint query (10) / corpus (140) pools; task-level multi-positive qrels: 80 queries / 1,120 docs / 11,200 qrels per direction.

**Design note**: 1:1 qrels measure at chance on this near-duplicate corpus (jina 0.015, Qwen 0.028, random 0.004); we ship the task-level design mirroring LIBERO. Adds `v2i` + `Robotics`, same one-line edits as #5255.

**Results (ndcg@10)** — floor is high: 140/1,120 docs relevant per query:

| Model | i2v | v2i |
|---|---|---|
| `baseline-random-encoder` | 0.1248 | 0.1161 |
| `jina-embeddings-v5-omni-nano` | 0.4193 | 0.4787 |
| `Qwen3-VL-Embedding-2B` | 0.7540 | 0.6975 |

## Checklist

- [x] I have outlined why this dataset is filling an existing gap in `mteb`
- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr).
  - [x] `mteb/baseline-random encoder`
  - [x] `jinaai/jina-embeddings-v5-omni-nano` (smallest image+video model)
- [x] I have checked that the performance is neither trivial (close to perfect scores) nor random.
- [x] I have considered the size of the dataset and reduced it if it is too big
- [ ] I reproduced scores from the original paper (N/A — new task construction)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Vue9R33gcLojTcxp8ZDkW3

PR #5255, opened 2026-08-20
dataset: add LIBERO i2v/v2i robot manipulation retrieval (MOEB)

ZhixuLi-the-Runner:
Part of MOEB (#4842). Adds two **image ↔ video** tasks:

- `LIBEROI2VRetrieval` — goal-state image → manipulation video (https://huggingface.co/datasets/ZhixuLi/LIBERO-I2V)
- `LIBEROV2IRetrieval` — the reverse (https://huggingface.co/datasets/ZhixuLi/LIBERO-V2I)

**Gap**: image ↔ video had no merged tasks; this adds the missing `v2i` category and a `Robotics` domain (first robotics dataset in MTEB). Distractors are same-scene episodes doing *different* goals, so scene recognition alone scores near random.

**Construction** (`scripts/build_libero_datasets.py`): built from scratch from physical-intelligence/libero (LIBERO benchmark, 40 tasks, 1,693 teleop episodes, 256×256 @ 10 fps). Each episode → H.264 mp4 + final-frame goal image; per-task disjoint query/corpus pools (a query's source episode is never in the corpus, so frame matching can't solve it); task-level multi-positive qrels: 400 queries / 1,293 docs / 12,930 qrels per direction.

**Results (ndcg@10)** — well above chance, far from ceiling:

| Model | i2v | v2i |
|---|---|---|
| `baseline-random-encoder` | 0.0297 | 0.0261 |
| `jina-embeddings-v5-omni-nano` | 0.3537 | 0.3312 |
| `Qwen3-VL-Embedding-2B` | 0.4753 | 0.4027 |

## Checklist

- [x] I have outlined why this dataset is filling an existing gap in `mteb`
- [x] I have tested that the dataset runs with the `mteb` package.
- [x] I have run the following models on the task (adding the results to the pr).
  - [x] `mteb/baseline-random encoder`
  - [x] `jinaai/jina-embeddings-v5-omni-nano` (smallest image+video model; `e5-small` is text-only)
- [x] I have checked that the performance is neither trivial (close to perfect scores) nor random.
- [x] I have considered the size of the dataset and reduced it if it is too big
- [ ] I reproduced scores from the original paper (N/A — new task construction)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Vue9R33gcLojTcxp8ZDkW3

PR #5351, opened 2026-09-01
dataset: add Jaco Play image-video retrieval (i2v, v2i)

nahatav:
Adds `JacoPlayI2VRetrieval` and `JacoPlayV2IRetrieval` over the CLVR Jaco Play dataset, filling two of the open image↔video slots in #4842.

## Construction

Follows the construction already merged for LIBERO (#5255) and ManiSkill (#5257): the query is the goal state of a held-out episode (its final frame) and the corpus is other episodes' videos, with a corpus item relevant iff it comes from the same task.

Held-out episodes never appear in the corpus, so identical-frame matching cannot solve it, and episodes recorded in the same scene that accomplish a different goal stay non-relevant, so matching the scene alone is not enough.

Tasks with fewer than 4 demonstrations are dropped, leaving **65 of 89 tasks**, so every query has several distinct relevant videos rather than a single one. That gives 130 queries, 904 corpus videos and 1,808 qrels. `v2i` is the same media with the roles swapped.

## Source detail worth noting

LeRobot concatenates all episodes into one mp4 per camera and addresses each episode by a `[from_timestamp, to_timestamp]` range, so clips are cut by timestamp rather than read as separate files. The scene camera is used rather than the wrist camera, which is too close-up to identify a task from a single frame.

Source: `lerobot/jaco_play`, **cc-by-4.0**, citation from the project's `CITATION.cff` (Dass et al. 2023).

## Checklist

- [x] `mteb/baseline-random-encoder` on the full task → i2v 0.017, v2i 0.038
- [ ] Second encoder not run yet. Happy to add one before merge
- [x] Verified 0 dangling qrels, 0 query/corpus id overlap, 0 duplicate media, and 0 episode leakage between queries and corpus

### DIFF
--- a/mteb/descriptive_stats/Image/Any2AnyRetrieval/JacoPlayI2VRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyRetrieval/JacoPlayI2VRetrieval.json
@@ -1,0 +1,66 @@
+{
+    "test": {
+        "num_samples": 1034,
+        "num_queries": 130,
+        "num_documents": 904,
+        "number_of_characters": 0,
+        "documents_text_statistics": null,
+        "documents_image_statistics": null,
+        "documents_audio_statistics": null,
+        "documents_video_statistics": {
+            "total_duration_seconds": 6555.900000000002,
+            "total_frames": 65559,
+            "min_width": 224,
+            "average_width": 224.0,
+            "max_width": 224,
+            "min_height": 224,
+            "average_height": 224.0,
+            "max_height": 224,
+            "min_duration_seconds": 2.199999999999818,
+            "average_duration_seconds": 7.252101769911507,
+            "max_duration_seconds": 15.100000000000364,
+            "unique_videos": 904,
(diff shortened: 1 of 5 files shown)


Samoed (commented):
I'm not sure that we want robotics dataset without publication

nahatav:
Fair, the LeRobot card says "Paper: More Information Needed". Jaco Play is a constituent dataset of Open X-Embodiment, published at ICRA 2024 (O'Neill et al., doi:10.1109/ICRA57147.2024.10611477), so I can cite that alongside the project's own `CITATION.cff` if that resolves it

isaac-chung:
No, Open X-Embodiment actually cites Jaco as:
```
[128] S. Dass, J. Yapeter, J. Zhang, J. Zhang, K. Pertsch,
S. Nikolaidis, and J. J. Lim, “CLVR jaco play
dataset,” 2023. [Online]. Available: https://github.
com/clvrai/clvr jaco play dataset
```

PR #5352, opened 2026-09-01
dataset: add Vaani multilingual speech-text retrieval (45 languages)

nahatav:
Adds `VaaniA2TRetrieval` and `VaaniT2ARetrieval` over **45 Indian languages** (4,001 utterances), from Vaani's transcribed release. Part of #4842.

## Why this release

The main Vaani corpus ships only a `train` split. The transcribed release ships an **official `test` split**, so here the evaluation set is held out at source rather than sampled out of training data. 45 of the 64 languages with a test split are kept; the rest fall under 25 usable utterances.

Languages span Indo-Aryan, Dravidian and Tibeto-Burman: Assamese, Bengali, Bhojpuri, Chakma, Garo, Gujarati, Hindi, Kannada, Kashmiri, Kokborok, Malayalam, Marathi, Mizo, Nepali, Odia, Punjabi, Tamil, Telugu, Tulu, Urdu and others.

## Scripts were verified, not assumed

Script tags come from inspecting the transcripts. Two are not what the language name implies:

- **Chakma is romanised** in this release despite having its own script, so `ccp-Latn`
- **Tulu is written in Kannada**, so `tcy-Knda`

A naive pass would have mislabelled Assamese, Nepali, Telugu and Urdu too, whose transcripts carry Latin markup that skews automatic detection.

## Cleaning

Transcripts contain `<noise>`, `<pause>`, `<static>` event tags and `{...}` braces marking code-switched English. Tags are removed as annotation artefacts; braces are unwrapped so the code-switched word survives, since it was actually spoken.

Byte-identical audio and repeated transcript text are dropped per language, so an identical query is never marked relevant to only one of the clips it matches.

`main_score` is `hit_rate_at_5`, matching every existing audio↔text retrieval task in mteb.

## Checklist

- [x] `mteb/baseline-random-encoder` on the full task → a2t 0.0702, t2a 0.0703
- [ ] Second encoder not run yet: 4,001 utterances across 45 subsets. Happy to add one before merge

### DIFF
--- a/mteb/descriptive_stats/Image/Any2AnyRetrieval/VaaniA2TRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyRetrieval/VaaniA2TRetrieval.json
@@ -1,0 +1,1844 @@
+{
+    "test": {
+        "num_samples": 8002,
+        "num_queries": 4001,
+        "num_documents": 4001,
+        "number_of_characters": 334597,
+        "documents_text_statistics": {
+            "total_text_length": 334597,
+            "min_text_length": 12,
+            "average_text_length": 83.62834291427143,
+            "max_text_length": 456,
+            "unique_texts": 4000
+        },
+        "documents_image_statistics": null,
+        "documents_audio_statistics": null,
+        "documents_video_statistics": null,
+        "queries_text_statistics": null,
+        "queries_image_statistics": null,
+        "queries_audio_statistics": {
+            "total_duration_seconds": 22663.5333125,
+            "min_duration_seconds": 0.965,
+            "average_duration_seconds": 5.66446721132217,
+            "max_duration_seconds": 24.188,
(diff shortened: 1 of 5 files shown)


Samoed:
How it's different from https://github.com/embeddings-benchmark/mteb/pull/5342?

nahatav:
Different modality pair: #5342 is audio↔image, retrieving the photograph someone is describing, while this is audio↔text, retrieving the transcript of what they said. It also uses Vaani's transcribed release rather than the main corpus, which ships an official `test` split, so the training-data point you raised on #5342 does not apply here. The language sets barely overlap either: #5342 is mostly dialects, this is Hindi, Bengali, Tamil, Telugu, Kannada, Marathi and similar

isaac-chung (approved):
Thanks!

PR #5355, opened 2026-09-01
dataset: add MCIF multilingual video and audio retrieval (t2v, v2t, t2a, a2t)

nahatav:
Adds four directions over [MCIF](https://arxiv.org/abs/2507.19634) (Papi et al. 2025): t2v/v2t and t2a/a2t, each across English, German, Italian and Chinese. Part of #4842.

## Gap

mteb has no multilingual video task and no multilingual audio+video task. Every existing video and audio+image task is English-only retrieval, so these are the first in both cells. Talks are spoken in English while the questions are parallel across all four languages, so the non-English subsets score cross-lingual grounding.

## Construction

Official **test** split only; MCIF ships nothing else.

Only the QA samples bind one question to one clip; recognition and translation samples group ~33 clips behind one reference. Unanswerable questions and ones about the speaker or affiliation are dropped, leaving **133 content-specific questions per language** against the full **755-clip corpus**. The fixed instruction prefix is stripped.

Media is stored once with both tracks; each task exposes only the column it may see, so video directions cannot be solved off the audio track.

License **CC-BY-4.0**: the paper states MCIF "is released under CC-BY 4.0 license", and FBK hosts the media itself.

## Results

nDCG@10 mean over subsets:

| direction | random | X-CLIP |
|---|---|---|
| v2t | 0.0342 | **0.1234** |
| t2v | 0.0041 | **0.0250** |
| a2t | 0.0374 | |
| t2a | 0.0083 | |

By subset, v2t: en 0.2288, de 0.0918, it 0.1095, zh 0.0634. t2v: en 0.0541, de 0.0084, it 0.0243, zh 0.0131.

## Checklist

- [x] Two encoders on both video directions, see table. The audio directions need a speech-text model; CLAP is trained on audio captions, not speech, and scores at the random floor here.
- [x] `test_task_quality.py` passes; exempted for `duplicate_audio`/`duplicate_video` as one corpus is reused across the four language subsets, as XM3600 is for `duplicate_image`

### DIFF
--- a/scripts/data/mcif_retrieval/create_data.py
+++ b/scripts/data/mcif_retrieval/create_data.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+"""Build the MCIF multilingual audio-visual retrieval tasks for MTEB.
+
+Source is `FBK-MT/MCIF` at a pinned revision, the MCIF benchmark (Papi et al. 2025) of
+recorded ACL conference talks. MCIF ships only a `test` split and hosts its own media,
+both under CC-BY-4.0, so nothing needs to be scraped or re-licensed here.
+
+MCIF is an instruction-following benchmark rather than a retrieval one. The conversion
+keeps only the question-answering samples, which are the sole entries that bind one
+question to one clip; the ASR and translation samples group ~33 clips behind a single
+reference and cannot be aligned per clip. Within the QA samples two further filters
+apply: `qa_type="NA"` is dropped because those questions are deliberately unanswerable,
+and `qa_origin="General"` is dropped because those ask about the speaker or affiliation
+and so match many talks rather than one. That leaves 133 content-specific questions,
+each unique in all four languages, against the full 755-clip corpus.
+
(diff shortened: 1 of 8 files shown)


scripts/data/mcif_retrieval/create_data.py
```diff
@@ -0,0 +1,200 @@
...
+def stage_build(work: Path) -> dict:
+    work.mkdir(parents=True, exist_ok=True)
+    meta = pq.read_table(
+        hf_hub_download(
+            _SOURCE_REPO,
+            "short_fixedprompt/test-00000-of-00001.parquet",
+            repo_type="dataset",
+            revision=_SOURCE_REV,
+        )
+    ).to_pylist()
```
Samoed, line 71: Why do you choose only short version?

Samoed (changes requested):
I think we can add, but I don't think this provide useful signal, because this is QA task without given answer and to answer a question model reqired to know document

PR #5356, opened 2026-09-01
dataset: add Afri-MCQA multilingual speech-image retrieval (a2i, i2a)

nahatav:
Adds a2i/i2a over [Afri-MCQA](https://arxiv.org/abs/2601.05699) (Tonja et al. 2026) across **16 African languages**. Part of #4842.

## Gap

mteb has **no multilingual audio+image task**. All six existing audio+image tasks are English-only, so this is the first, and it covers languages with almost no non-text representation in the benchmark.

Questions are spoken by native speakers and grounded in culturally relevant photographs, so the task is speech-to-image grounding rather than transcript matching.

## Construction

Official **test** split only. Only the question audio is kept; the four spoken answer options are dropped because they describe candidate answers, not the image.

Three cleanups, all in `scripts/data/afri_mcqa_retrieval/create_data.py`:

- **432 rows dropped.** The source has questions with no recording and a few missing images, which fail on load. Detected by reading both columns undecoded.
- **Duplicates removed.** A few photographs are byte-identical within a language (twi 179 to 178, lin 152 to 150) and a few questions share a recording (sot 151 to 148). Questions on a collapsed image are remapped to the surviving one, so no qrel is orphaned.
- **Images held per language**, not pooled. One photograph can carry questions in several languages, and a pooled corpus would score a correct retrieval wrong when the model returned the twin from another language.

Result: **3,855 spoken questions over 2,430 images**. License **CC-BY-NC-4.0**, matching the source.

## Checklist

- [x] Tested that it runs with `mteb` (both directions)
- [x] `mteb/baseline-random-encoder` mean over 16 subsets gives a2i 0.0286, i2a 0.0208
- [ ] Second encoder not run: CPU-only hardware. Same rationale as #5344.
- [x] Size considered, official test split, no reduction

### DIFF
--- a/mteb/descriptive_stats/Image/Any2AnyMultilingualRetrieval/AfriMCQAA2IRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyMultilingualRetrieval/AfriMCQAA2IRetrieval.json
@@ -1,0 +1,718 @@
+{
+    "test": {
+        "num_samples": 6285,
+        "num_queries": 3855,
+        "num_documents": 2430,
+        "number_of_characters": 0,
+        "documents_text_statistics": null,
+        "documents_image_statistics": {
+            "min_image_width": 47,
+            "average_image_width": 1150.8810699588478,
+            "max_image_width": 5760,
+            "min_image_height": 70,
+            "average_image_height": 1159.2526748971193,
+            "max_image_height": 5700,
+            "unique_images": 2429
+        },
+        "documents_audio_statistics": null,
+        "documents_video_statistics": null,
+        "queries_text_statistics": null,
+        "queries_image_statistics": null,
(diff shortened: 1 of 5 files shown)


isaac-chung (approved):

Samoed (commented):
I'm not sure that this format of task make sense. It takes from original queries audio and image and try to match parts of questions. I think better to implemented as originally inteded as QA task @isaac-chung

isaac-chung:
> I'm not sure that this format of task make sense. It takes from original queries audio and image and try to match parts of questions. I think better to implemented as originally inteded as QA task @isaac-chung

Hmm true. I don't mind this formulation. Ultimately task selection should filter things out. But yes let's add a QA task please, @nahatav

PR #5357, opened 2026-09-01
dataset: add Afri-MCQA cross-modal category classification (ia2c, 16 African languages)

nahatav:
Adds `AfriMCQACategoryClassification` over [Afri-MCQA](https://arxiv.org/abs/2601.05699) (Tonja et al. 2026): classify an entry's cultural category from its photograph and a question about it spoken by a native speaker, across **16 African languages**. Part of #4842.

Distinct from #5356, a2i/i2a retrieval on the same source: there image and audio sit on opposite sides of a pair, here they are one joint input classified into a category.

## Gap

mteb has **no audio+image classification task at all**; every audio+image task is retrieval. The taxonomy had no direction either, so this adds `ia2c` (image+audio to category) beside the existing `it2c` and `va2c`. It is also the first multilingual task there.

## Construction

Official splits: `dev` trains the evaluator, `test` is scored. In `scripts/data/afri_mcqa_classification/create_data.py`:

- Entries kept only when the photograph and recording are both present and the entry has a **single** category; about 13% carry several. Category strings are casefolded, as the source uses two cases.
- **Leakage removed.** The official dev and test splits share photographs and the label is a property of the photograph, so any photograph in test is dropped from train, leaving train small (19 to 64 per language).
- **Identical recordings dropped**, being repeats of an example.

Main score is **macro-f1**: the smallest subsets are skewed enough that a majority-class predictor reaches 0.79 accuracy on `sot`.

License **CC-BY-NC-4.0**, as in the source.

## Checklist

- [x] `mteb/baseline-random-encoder` macro-f1 mean over 16 subsets: 0.0836 (range 0.041 to 0.220)
- [ ] Second encoder not run: no `mteb` encoder takes image and audio jointly, short of the optional EBind backbones.
- [x] Size considered, official splits
- [x] `test_task_quality.py` passes; exempted for `duplicate_image`, as several questions share one photograph, like WorldSense across QA rows


mteb/tasks/classification/multilingual/afri_mcqa_category_classification.py
```diff
@@ -0,0 +1,73 @@
...
+            "recording are present and the entry carries a single category, since "
+            "around 13% are tagged with several at once. Construction script: "
+            "scripts/data/afri_mcqa_classification/create_data.py."
+        ),
+        reference="https://arxiv.org/abs/2601.05699",
+        dataset={
+            "path": "vnahata/AfriMCQA-category-classification",
+            "revision": "1581d446105e451a1fc8d2c62fd1f3d54d75fb5e",
+        },
+        type="ImageClassification",
```
isaac-chung: Let's create a new AudioImageClassification for now in `_TASKTYPE2SIMPLIFIEDTASKTYPE` in `mteb/abstasks/task_metadata.py`

isaac-chung (commented):
Looks good, just a small thing:
1. Let's create a new task type. This is not pure image clf
2. move the descriptive stats file into the new folder

mteb/descriptive_stats/AudioImageClassification/AfriMCQACategoryClassification.json
isaac-chung, line 1: Apologies for not being clearer. I'd prefer the path to be `mteb/descriptive_stats/AudioImageClassification/AfriMCQACategoryClassification.json`

isaac-chung: Otherwise LGTM

Samoed: This is done automatically and to match exactly main code for handling descriptive stats should be updated

Samoed (commented):
You can also add `t2c` variant (including english queries)

nahatav:
Both done. AudioImageClassification moved from MIEB_TASK_TYPE to MAEB_TASK_TYPE, which is the group that doesn't get the Image/ prefix, so the path is derived rather than moved by hand. Also fixed a bug I found while looking at the t2c idea, the source spells five categories a second way, including "Tradition, Art, and History" beside the source's own typo "tranditions", so 133 test and 68 train rows were being silently dropped. Both spellings now map to the same label and the stats are regenerated

Can someone else also take a look at t2c please? I left it out for now because 91 rows seem to have identical question text with different gold labels ("Who is this person?" maps to two categories), so the text only variant isn't super well posed without the image.

Samoed:
> the text only variant isn't super well posed without the image.

I agree, and because of that I don't think that task from https://github.com/embeddings-benchmark/mteb/pull/5356 is good measure for embedders

mteb/abstasks/task_metadata.py
```diff
@@ -208,6 +208,10 @@
     "AudioZeroshotClassification",
     "AudioClassification",
     "AudioPairClassification",
+    # Grouped with the audio types rather than the image ones so that its descriptive
+    # statistics land in descriptive_stats/AudioImageClassification/ rather than nested
+    # under descriptive_stats/Image/, which only the MIEB types are.
+    "AudioImageClassification",
```
isaac-chung, line 214: This is not part of MAEB. Better to create a new `MOEB_TASK_TYPE`.

isaac-chung (commented):
Let's remove the t2c variant then

PR #5364, opened 2026-09-01
dataset: add Indic DiarBench multilingual speaker retrieval (a2a, 20 Indian languages)

nahatav:
Adds `IndicDiarBenchSpeakerA2ARetrieval` over [Indic DiarBench](https://arxiv.org/abs/2607.23808) (Mehendale et al., Interspeech 2026): given a clip of one speaker, retrieve other clips of that speaker, across **20 scheduled languages of India**. Part of #4842.

## Gap

`a2a` has 26 tasks and only **one** is multilingual, so this is where multilingual coverage is thinnest. The source is new to mteb and none of these 20 languages have an existing a2a task.

Speaker labels are numbered per session, so identity pairs session with speaker. The corpus keeps other speakers of the same session as the hardest distractors: same room, same channel, different voice.

## Construction

Official **test** split only. In `scripts/data/indic_diarbench_speaker/create_data.py`:

- Annotated turns cut to clips of 2 to 15 seconds; shorter turns carry too little voice to identify a speaker.
- A turn overlapping a different speaker by more than 0.5s is dropped, since the corpus has a 12.8% overlap ratio and such a clip holds two voices. Keeping brief boundary overlap roughly doubles usable clips.
- A speaker needs at least 5 clips, capped at 12, with 3 as queries. Languages capped at 60 speakers so subsets stay comparable.
- **Maithili and Sindhi excluded**: the source gives them only four speakers each, too few to rank meaningfully.

Result: **1,884 queries over 5,100 clips**. License **CC-BY-4.0**, as the source.

## Results

nDCG@10, mean over the 20 subsets:

| model | mean | range |
|---|---|---|
| `mteb/baseline-random-encoder` | 0.1006 | 0.015 to 0.263 |
| `laion/clap-htsat-fused` | 0.4948 | 0.320 to 0.719 |

CLAP is 5x random, so the task is learnable and far from saturated. Random tracks how many speakers a language has, which is a property of the source.

## Checklist

- [x] Two encoders run, see table

### DIFF
--- a/mteb/tasks/retrieval/multilingual/indic_diarbench_speaker_retrieval.py
+++ b/mteb/tasks/retrieval/multilingual/indic_diarbench_speaker_retrieval.py
@@ -1,0 +1,118 @@
+from __future__ import annotations
+
+from datasets import load_dataset
+
+from mteb.abstasks.retrieval import AbsTaskRetrieval
+from mteb.abstasks.retrieval_dataset_loaders import RetrievalSplitData
+from mteb.abstasks.task_metadata import TaskMetadata
+
+_DATASET_PATH = "vnahata/IndicDiarBench-speaker-retrieval"
+_DATASET_REVISION = "0e426b97c9a792fa6c10efc12b2d0800a1ff4257"
+
+_LANGUAGES = {
+    "asm": ["asm-Beng"],
+    "ben": ["ben-Beng"],
+    "brx": ["brx-Deva"],
+    "doi": ["doi-Deva"],
+    "guj": ["guj-Gujr"],
+    "hin": ["hin-Deva"],
+    "kan": ["kan-Knda"],
+    "kas": ["kas-Arab"],
+    "kok": ["gom-Deva"],
+    "mal": ["mal-Mlym"],
+    "mar": ["mar-Deva"],
+    "mni": ["mni-Beng"],
+    "npi": ["npi-Deva"],
+    "ory": ["ory-Orya"],
+    "pan": ["pan-Guru"],
+    "san": ["san-Deva"],
+    "sat": ["sat-Olck"],
+    "tam": ["tam-Taml"],
+    "tel": ["tel-Telu"],
+    "urd": ["urd-Arab"],
+}
+
+_BIBTEX = r"""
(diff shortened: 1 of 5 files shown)


mteb/tasks/retrieval/multilingual/indic_diarbench_speaker_retrieval.py
```diff
@@ -0,0 +1,118 @@
...
+            "Speaker labels are numbered per recording session, so identity pairs the "
+            "session with the speaker, and the corpus holds the other speakers of the "
+            "same session as the hardest distractors: same room, same channel, different "
+            "voice. Built from the official test split by cutting annotated turns to "
+            "clips of 2 to 15 seconds, dropping turns that overlap a different speaker "
+            "by more than 0.5s, and keeping speakers with at least five clips. Maithili "
+            "and Sindhi are excluded because the source gives them only four speakers "
+            "each. Construction script: "
+            "scripts/data/indic_diarbench_speaker/create_data.py."
+        ),
```
Samoed, line 60: Isn't this more classification/clustering task? I'm not sure that this is good retrieval task

isaac-chung: `LibriSpeechSpeakerA2ARetrieval` provides a precedence for such as task. I think it's fine.

mteb/tasks/retrieval/multilingual/indic_diarbench_speaker_retrieval.py
```diff
@@ -0,0 +1,118 @@
...
+        dataset={"path": _DATASET_PATH, "revision": _DATASET_REVISION},
+        type="Any2AnyMultilingualRetrieval",
+        category="a2a",
+        modalities=["audio"],
+        eval_splits=["test"],
+        eval_langs=_LANGUAGES,
+        main_score="ndcg_at_10",
+        date=("2025-01-01", "2026-07-26"),
+        domains=["Spoken"],
+        task_subtypes=["Speaker Identification"],
```
isaac-chung, line 71: In MAEB, `LibriSpeechSpeakerA2ARetrieval` used "Speech Retrieval"

PR #5365, opened 2026-09-02
dataset: add Omnilingual ASR speech-text retrieval (a2t, t2a, 50 new languages)

nahatav:
Adds `OmnilingualASRA2TRetrieval` and `OmnilingualASRT2ARetrieval` over [Omnilingual ASR](https://arxiv.org/abs/2511.09690) (Meta, 2025), covering **50 languages that no existing mteb audio task reaches**. Part of #4842.

## Gap

The audio tasks in mteb reach 165 distinct languages between them. **302 of this corpus's 308 test languages are in none of them**, so the point here is reach into low-resource languages rather than another benchmark on well-served ones. The 50 published span 7 writing systems: Latin, Devanagari, Arabic, Hebrew, Sinhala, Malayalam and Odia.

## Construction

Official **test** split only. In `scripts/data/omnilingual_asr_retrieval/create_data.py`:

- Languages picked by a fixed rule: absent from every existing mteb audio task, test shard between 120MB and 320MB so each has enough recordings without one language dominating, then taken round-robin across writing systems so the set is not all one script.
- Recordings resampled to 16 kHz and re-encoded FLAC to Opus, about 11x smaller, which keeps the published data near a gigabyte.
- Repeated transcripts dropped, since one would otherwise be relevant to several recordings while only one is marked correct.

Result: **6,236 recordings, 50 languages**, 55 to 431 per language. License **CC-BY-4.0**, as the source.

## Checklist

- [x] `mteb/baseline-random-encoder` nDCG@10: a2t 0.0485, t2a 0.0510 (mean over 50 subsets)
- [ ] Second encoder not run. CLAP is the obvious candidate but does not fit: transcripts reach 2,523 characters against its 512 token limit, and recordings average 72s against its 10s design point. Runs segfault rather than producing a usable number. Happy to add a long-form speech encoder if one lands.

### DIFF
--- a/mteb/descriptive_stats/Image/Any2AnyMultilingualRetrieval/OmnilingualASRA2TRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyMultilingualRetrieval/OmnilingualASRA2TRetrieval.json
@@ -1,0 +1,2044 @@
+{
+    "test": {
+        "num_samples": 12472,
+        "num_queries": 6236,
+        "num_documents": 6236,
+        "number_of_characters": 3168582,
+        "documents_text_statistics": {
+            "total_text_length": 3168582,
+            "min_text_length": 16,
+            "average_text_length": 508.11128928800514,
+            "max_text_length": 2523,
+            "unique_texts": 6227
+        },
+        "documents_image_statistics": null,
+        "documents_audio_statistics": null,
+        "documents_video_statistics": null,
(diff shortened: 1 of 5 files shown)


isaac-chung:
Generally looks good. I'm going to validate this task on LCO-embedding-3B first.

isaac-chung (approved):

Issue #5254, opened 2026-08-20
Omnimodal wikipedia


KennethEnevoldsen:
Add an omnimodal version of wikipedia, which generated queries of varying modalities 

queries:
- audio: "what did niel armstrong say on the moon"
- text: "how tall is the eiffel tower"
- text + image: "what does this image show" 
- ...

Especially targeting stuff in 
- tables
- audio clips
- images
- gifs
- ...

related to #4842

ayush1298:
citing couple of already done researches: 
1. https://arxiv.org/pdf/2503.19065
2. https://www.kaggle.com/datasets/jacksoncrow/extended-wikipedia-multimodal-dataset
3. https://arxiv.org/abs/2305.05432

lxer19:
Interested in working on this. I’ve been doing multimodal benchmark/data-integrity audits in MTEB recently. I’d like to first scope the task design and existing related datasets/work before proposing an implementation plan.

whybe-choi:
I'm very interested in working on this. Let me know if I can help !

lxer19:
I started with WikiWeb2M. It looks like a useful starting point for the text/image corpus construction: it re-scrapes the ~2M English Wikipedia articles represented in WIT and preserves page/section structure, text, images, and captions. We could likely reuse this construction approach and its basic validity filtering.

My current scoping plan is:

1. Prior art / reusable data— inspect WikiWeb2M, Extended Wikipedia Multimodal Dataset, WikiAutoGen, and OVEN.
2. Modality matrix — map existing vs. missing query→target combinations and identify high-value gaps.
3. Benchmark semantics — define retrieval units, positives/qrels, multi-positive cases, metrics, hard negatives, and versioning.
4. Pilot → scale — build a small v0 for 2–3 modality combinations, validate the semantics and difficulty, then scale only after review.

lxer19:
One design question from the prior work is retrieval granularity.

WikiWeb2M gives a deterministic image → section mapping, but co-location does not necessarily make that section a valid semantic target. The Extended Wikipedia Multimodal Dataset instead associates images at the article level, while WikiSeek filters underspecified topics rather than assuming a fine-grained target.

So for v0, I’d first test whether `image → section` is actually well-defined. If not, `image → page` or `image + text → section` may be better.

AdnanElAssadi:
Cool @whybe-choi @lxer19, I think you can both work on this. This is fairly large in scope.

[...]

PR #5368, opened 2026-09-02
dataset: add Spoken Wikipedia speech-text retrieval (a2t, t2a, 5 languages)

nahatav:
Adds `SpokenWikipediaA2TRetrieval` and `SpokenWikipediaT2ARetrieval`: volunteer readings of Wikipedia articles paired with the article lead, in **Dutch, English, German, Spanish and French**. Part of #4842, and a first piece of the audio side of #5254.

## Gap

No task in mteb uses Wikimedia audio. #5254 asks for an omnimodal Wikipedia covering audio and video as well as text and images. This is a self-contained slice of that: the recordings exist, they are free by Commons policy, and each already points at an article.

## Construction

Built from the Commons and Wikipedia APIs, in `scripts/data/spoken_wikipedia/create_data.py`.

- The article for a recording comes from the file's **global usage**, restricted to main-namespace pages so project pages are excluded. Where a file is no longer embedded in its article, the title is parsed from the filename, whose shape differs per language (`En-Title-article.oga`, `FR-Title.ogg`). A guess that is not a real article is dropped when its lead fails to resolve.
- Readings run to tens of minutes, so **only the opening 60 seconds is kept**. Readers start at the lead, so "which article is this a recording of" holds however much is kept.
- Leads under 200 characters are dropped as too thin to identify, and **one recording per article**.

Result: **150 pairs per language, 750 total, all texts unique**. Audio is 16 kHz Opus. License **CC-BY-SA-4.0**: Commons media is free by policy, Wikipedia text is CC-BY-SA.

## Checklist

- [x] `mteb/baseline-random-encoder` nDCG@10: a2t 0.0383, t2a 0.0421, against a 0.030 floor for a 150 item corpus
- [x] `jina-embeddings-v5-omni-nano` nDCG@10: a2t **0.5533**, t2a **0.7169**, 14x and 17x the floor. a2t by language: eng 0.8432, deu 0.5809, nld 0.5163, fra 0.4815, spa 0.3448, a 2.4x spread, so not uniformly easy
- [x] Size considered, one recording per article

### DIFF
--- a/mteb/descriptive_stats/Image/Any2AnyMultilingualRetrieval/SpokenWikipediaA2TRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyMultilingualRetrieval/SpokenWikipediaA2TRetrieval.json
@@ -1,0 +1,244 @@
+{
+    "test": {
+        "num_samples": 1500,
+        "num_queries": 750,
+        "num_documents": 750,
+        "number_of_characters": 903278,
+        "documents_text_statistics": {
+            "total_text_length": 903278,
+            "min_text_length": 204,
+            "average_text_length": 1204.3706666666667,
+            "max_text_length": 6305,
+            "unique_texts": 750
+        },
+        "documents_image_statistics": null,
+        "documents_audio_statistics": null,
+        "documents_video_statistics": null,
+        "queries_text_statistics": null,
+        "queries_image_statistics": null,
+        "queries_audio_statistics": {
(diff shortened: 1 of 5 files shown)


Samoed:
Can you run some models? I don't think that this is good task for now, because it seems pretty easy match recording and text

nahatav:
@Samoed Done. Check out the Jina performance

isaac-chung (approved):

PR #5369, opened 2026-09-02
dataset: add MTVQA multilingual image+text retrieval (it2t, 9 languages)

nahatav:
Adds `MTVQAIT2TRetrieval` over [MTVQA](https://arxiv.org/abs/2405.11985) (Tang et al. 2024): given an image and a question about the text inside it, retrieve the answer, in 9 non-English languages. Part of #4842.

## Gap

mteb has **13 `it2t` tasks and none is multilingual**. This scores whether a model reads text in a picture in a non-English language.

## Construction

Official **test** split, in `scripts/data/mtvqa_retrieval/create_data.py`.

- `qa_pairs` is a Python literal, not JSON, and answers contain apostrophes, so it is read with `ast.literal_eval`.
- **Answers repeat**, since different images share short answers such as a price, so the corpus is deduplicated by answer text and every question with that answer points at the surviving document.

**6,778 queries, 5,973 answers, 2,116 images.** License **CC-BY-NC-4.0**, as the source.

## Candidate sets

Following @isaac-chung suggestion, each question now ranks against **its own answer plus the 10 nearest answers**, chosen by `multilingual-e5-small` comparing question text to answer text, exposed as `top_ranked`. Mean 10.6 candidates.

Selection is **text only, on purpose**: using an image-text model would bake its view of the images into the benchmark.

The selector surfaces the gold answer on its own for only **25% to 64%** of questions, so the negatives are hard rather than trivially separable. The gold is unioned in, so every question stays answerable.

## Results

`main_score` is now **accuracy**. With ~11 candidates nearly everything falls inside a cutoff of 10, putting the nDCG@10 random floor at 0.43; accuracy lands on the theoretical 1/11 floor.

| model | accuracy |
|---|---|
| `mteb/baseline-random-encoder` | 0.0865 |
| `openai/clip-vit-base-patch32` | **0.1087** |

## Checklist

- [x] Two encoders run, see table
- [x] `test_task_quality.py` passes, exemptions inline

### DIFF
--- a/mteb/tasks/retrieval/multilingual/mtvqa_retrieval.py
+++ b/mteb/tasks/retrieval/multilingual/mtvqa_retrieval.py
@@ -31,7 +31,7 @@ class MTVQAIT2TRetrieval(AbsTaskRetrieval):
         reference="https://arxiv.org/abs/2405.11985",
         dataset={
             "path": "vnahata/MTVQA-it2t-retrieval",
-            "revision": "72f2324b9ed1e46ca2e41576a3b81b7df480cf3c",
+            "revision": "f2d2ab47cf576e94ef89807d9c2dd8012515cf84",
         },
         type="Any2AnyMultilingualRetrieval",
         category="it2t",
@@ -48,7 +48,7 @@ class MTVQAIT2TRetrieval(AbsTaskRetrieval):
             "tha": ["tha-Thai"],
             "vie": ["vie-Latn"],
         },
-        main_score="ndcg_at_10",
+        main_score="accuracy",
         date=("2023-01-01", "2024-05-20"),
         domains=["Scene", "Written"],
         task_subtypes=["Rendered Texts Understanding"],
--- a/mteb/tasks/retrieval/multilingual/mtvqa_retrieval.py
+++ b/mteb/tasks/retrieval/multilingual/mtvqa_retrieval.py
@@ -1,0 +1,62 @@
+from __future__ import annotations
+
+from mteb.abstasks.retrieval import AbsTaskRetrieval
+from mteb.abstasks.task_metadata import TaskMetadata
+
+_BIBTEX = r"""
+@article{tang2024mtvqa,
+  author = {Tang, Jingqun and Liu, Qi and Ye, Yongjie and Lu, Jinghui and Wei, Shu and Lin, Chunhui and Li, Wanqing and Mahmood, Mohamad Fitri Faiz Bin and Feng, Hao and Zhao, Zhen and Wang, Yanjie and Liu, Yuliang and Liu, Hao and Bai, Xiang and Huang, Can},
+  journal = {arXiv preprint arXiv:2405.11985},
+  title = {{MTVQA}: Benchmarking Multilingual Text-Centric Visual Question Answering},
+  year = {2024},
+}
+"""
+
+
+class MTVQAIT2TRetrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="MTVQAIT2TRetrieval",
+        description=(
+            "Retrieve the answer to a question about the text inside an image, across "
+            "nine languages. The query is the image together with the question, and the "
+            "corpus is the answers for that language. mteb has thirteen it2t tasks and "
+            "none of them is multilingual, so this scores whether a model reads text in "
+            "a picture in a language other than English. Built from the official test "
+            "split. Answers repeat across questions, because different images can share "
+            "a short answer such as a price or a name, so the corpus is deduplicated by "
+            "answer text and every question with that answer points at the surviving "
+            "document rather than dropping the question. Construction script: "
+            "scripts/data/mtvqa_retrieval/create_data.py."
+        ),
+        reference="https://arxiv.org/abs/2405.11985",
+        dataset={
+            "path": "vnahata/MTVQA-it2t-retrieval",
+            "revision": "72f2324b9ed1e46ca2e41576a3b81b7df480cf3c",
+        },
+        type="Any2AnyMultilingualRetrieval",
+        category="it2t",
+        modalities=["image", "text"],
+        eval_splits=["test"],
+        eval_langs={
+            "ara": ["ara-Arab"],
+            "deu": ["deu-Latn"],
+            "fra": ["fra-Latn"],
+            "ita": ["ita-Latn"],
+            "jpn": ["jpn-Jpan"],
+            "kor": ["kor-Hang"],
+            "rus": ["rus-Cyrl"],
+            "tha": ["tha-Thai"],
+            "vie": ["vie-Latn"],
+        },
+        main_score="ndcg_at_10",
+        date=("2023-01-01", "2024-05-20"),
+        domains=["Scene", "Written"],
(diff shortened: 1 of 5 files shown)


isaac-chung:
You can probably run a CLIP model as well

mteb/tasks/retrieval/multilingual/mtvqa_retrieval.py
```diff
@@ -0,0 +1,62 @@
...
+            "a picture in a language other than English. Built from the official test "
+            "split. Answers repeat across questions, because different images can share "
+            "a short answer such as a price or a name, so the corpus is deduplicated by "
+            "answer text and every question with that answer points at the surviving "
+            "document rather than dropping the question. Construction script: "
+            "scripts/data/mtvqa_retrieval/create_data.py."
+        ),
+        reference="https://arxiv.org/abs/2405.11985",
+        dataset={
+            "path": "vnahata/MTVQA-it2t-retrieval",
```
Samoed, line 33: This is QA datasets and should have `top_ranked` split

nahatav:
CLIP added, nDCG@10 0.0616 against a 0.0082 floor, 7.5x. 

On top_ranked, MTVQA only ships the answer with no distractors, so any candidates would be sampled negatives I believe.

Samoed:
I don't think this is good task then.

isaac-chung:
> I don't think this is good task then.

I agree with @Samoed . Perhaps this task can be improved if we constructed top_rank ourselves?

nahatav:
@isaac-chung @Samoed Agreed. How do we go about that?

isaac-chung:
I haven't looked at the data, but I'd try limiting each question to its answer + deduplicated top 10 answers, via a tiny multilingual model only in the text modality (compare question and answers).

nahatav:
@isaac-chung @Samoed I gave it an attempt. Kindly take a look (as well as the other open PRs, when you get the chance. I revised a bunch). 

Each question now ranks against its answer plus the 10 nearest picked by multilingual-e5-small on text only. The selector finds the gold unaided just 25-64% of the time depending on language so the negatives aren't trivial. Also switched main_score to accuracy with ~11 candidates nDCG@10's floor was 0.43 and CLIP came out below random on it. Accuracy sits at the theoretical 1/11, random 0.0865 and CLIP 0.1087

Thanks!

isaac-chung:
I can run lco-embed-3B to see what the scores are like. Wdyt @Samoed ?

Samoed:
Would be good, but I still not sure in this task

isaac-chung:
@Samoed what do you propose we do?

Samoed:
Close this pr

isaac-chung:
Why? We support VQA task types, and indeed all are English only so far.

Samoed:
I don't like that we're created `top_ranked` from other answers

isaac-chung:
Do you have any suggestions that can improve this framing?

Samoed:
No. I don't think that this task fit for evaluating embedding models, because originally it was created to answer based on image. Maybe we can keep as an retrieval task, but this is still strange a bit

isaac-chung:
In MIEB, we had [vision-centric question answering](https://arxiv.org/pdf/2504.10471v1) tasks that are in retrieval format (prev. Any2AnyMultiChoice). Same thing here.

Samoed:
I don't have questions abot QA tasks in general, my comments realated to this specific task. All existing `Any2AnyMultiChoice` tasks have `top_ranked` split to select across some candidates and originally they had answers to select from, but this dataset don't have predifined answers and that make evaluation not clear. Also answers in this dataset are sometimes a "general" and `qrels` sometime will have only partly coverage without manual checks. E.g. `yes/no`, numbers etc

isaac-chung:
I see. This is much more detailed and very helpful to understand the context. 

I took another look at the [original dataset. One image mostly have multiple question/answer pairs](https://huggingface.co/datasets/ByteDance/MTVQA/viewer/default/test?f%5Blang%5D%5Bvalue%5D=%27IT%27&row=845), e.g. 

```
[
{'question': 'Qual è il nome del negozio?', 'answer': 'Vergelio'}, 
{'question': 'Qual è la percentuale di sconto?', 'answer': 'Fino al 50%'}, 
{'question': 'Cosa vende il negozio?', 'answer': 'Scarpe e borse'}, 
{'question': 'Quali sono i prodotti in saldo?', 'answer': 'Scarpe e borse'}
]
```

I think a reasonable top_ranked list for each question can be the unique answers from the same image.

Samoed:
We can do like this, but most of times there are 3-4 questions per images that not much. In some cases only 1

isaac-chung:
Maybe you have a solution for such cases?

Samoed:
We can remove such examples, but again I don't think that this task would give informative signal about quality

isaac-chung:
@deep9539 would be great to get another opinion here, thanks!

deep9539:
Yeah I agree with what @Samoed and @isaac-chung mentioned. I looked at CVBenchCount where the answers were very close to each other like `How many object in this image?` and the answer is 4 options `1, 2, 5, or 7`.

Using embedding to come up with close ones make sense, thanks @nahatav for trying this. 

I found few issues in this task:
**A) What if there are very close candidates and multiple being correct answer?**
I found cases where there are candidates such as "Arabic.", "An excerpt from an Arabic novel.", "Arabic" (second time :) ).. This will impact your metrics heavily, right?

**B) Not sure if there are closer possible and hard negatives for the query.**
For example, I found that in the corpus [here](https://huggingface.co/datasets/vnahata/MTVQA-it2t-retrieval/viewer/ara-corpus/corpus?row=12), the candidates are "Her face is not visible in the picture.", "Jumping on one leg, running, and climbing trees.", etc. I'm not sure if there will be other candidates which will be closer to this in similarity and also make sense for the given query. 

**C) I think sometimes the answer doesn't make sense unless other context is provided.**
This is different from CVBenchCount where it's exact count (i.e. numerical digit) which have single and well-defined meaning. I found documents which are saying "Dabo", "Mohamed Esmat", "UNIM" (English translated) which I think only make sense when query is given. 

**Possible solutions:**
A) Maybe call some LLM to discard close matches? - I suspect that your accuracy will improve with this.
B) Maybe use LLM to generate hard options? - This will drop accuracy but it's okay, we can even only keep 4 option.
C) TBH, I personally don't know how to solve this.

IMO, the solutions might be hard to implement and there is room for error or missing out on edge cases, but I'll let @isaac-chung @Samoed add their thoughts as well based on the latest findings..

isaac-chung:
Thanks @deep9539 !

> I think sometimes the answer doesn't make sense unless other context is provided.

100% agree. If an LLM/VLM is used to generate hard negatives, the query must have both the image and text as context. 

Ultimately, the construction of a meaningful task from this dataset is more involved than simply adding it in retrieval format. This is how I see this proceeding. Either:

1. We keep trying, using LLMs to help generate answer options that make sense, or
2. We close this PR, note down findings on the issue, and try separately

I lean towards option 2.

isaac-chung:
Closing this now. @nahatav I'd encourage you to open a specific issue for MTVQAIT2TRetrieval and continue the investigation there first, if you wish to.

PR #5373, opened 2026-09-03
dataset: add Afri-MCQA multiple choice visual QA (it2t, 16 African languages)

nahatav:
Adds `AfriMCQAVisionCentricQA`: given a photograph and a question about it in the native language, pick the correct answer from that question own four options, in 16 African languages. The QA formulation asked for in #5356. Part of #4842.
## Gap
`mteb` has **35 multiple choice tasks and every one is English only**, so this is the first multilingual one. #5356 added a2i/i2a over the same source, where image and audio sit on opposite sides of a pair.
## Construction
Script: `scripts/data/afri_mcqa_multiple_choice/create_data.py`
**Only `dev` can be scored.** The `*_test` configs ship the four options already shuffled into `native_option_1..4` with no key; `*_dev` keeps `correct_native` apart from `wrong_native_o1..o3`.
**Options are shuffled by a hash of the question.** The source lists the correct answer first in every row, so source order would leave it at index 0 every time. Hashing keeps order stable across rebuilds without leaking position.
Scored as `mteb` scores its other multiple choice sets: each question candidates become its `top_ranked`, so it ranks against its own four options, and the metric is accuracy.

**2,792 questions, 11,168 options.** License **CC-BY-NC-4.0**, as in the source.
## Results
Accuracy over the 16 subsets; a four way choice puts the floor at 0.25.
| model | mean |
|---|---|
| `mteb/baseline-random-encoder` | 0.2543 |
| `openai/clip-vit-base-patch32` | 0.2796 |
| `jina-embeddings-v5-omni-nano` | 0.2641 |
Both sit near the floor, which looks like the format rather than this dataset. As a control, the already merged `BLINKIT2TMultiChoice` is also `it2t`, also accuracy, and English: CLIP scores **0.3985 against its 0.3871 floor**, a ratio of 1.03 against 1.04 here.
## Checklist
- [x] Three encoders run, see table
- [x] Size considered, full official dev split
- [x] `test_task_quality.py` passes, three exemptions noted inline

### DIFF
--- a/mteb/tasks/multichoice/multilingual/afri_mcqa_multiple_choice.py
+++ b/mteb/tasks/multichoice/multilingual/afri_mcqa_multiple_choice.py
@@ -1,0 +1,112 @@
+from __future__ import annotations
+
+from datasets import Dataset, load_dataset
+
+from mteb.abstasks.retrieval import AbsTaskRetrieval
+from mteb.abstasks.retrieval_dataset_loaders import RetrievalSplitData
+from mteb.abstasks.task_metadata import TaskMetadata
+
+_DATASET_PATH = "vnahata/AfriMCQA-multiple-choice"
+_DATASET_REVISION = "3ed9383894a3ff0261af4097d1e4e9d593868f1d"
+
+_LANGUAGES = {
+    "twi": ["twi-Latn"],
+    "amh": ["amh-Ethi"],
+    "nya": ["nya-Latn"],
+    "hau": ["hau-Latn"],
+    "ibo": ["ibo-Latn"],
+    "kik": ["kik-Latn"],
+    "kin": ["kin-Latn"],
+    "lin": ["lin-Latn"],
+    "lug": ["lug-Latn"],
+    # Afri-MCQA's Oromo is the West Central variety, so `gaz` rather than the `orm`
+    # macrolanguage
+    "orm": ["gaz-Latn"],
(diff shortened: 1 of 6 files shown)


Samoed (changes requested):
Why do you include only dev? I think test split should be better

nahatav:
@Samoed The test split has no answer key. Its configs ship the options pre shuffled with no indication which is correct, while dev keeps correct native apart from wrong native. Same for all_test, and there's no key file in the repo

mteb/tasks/multichoice/multilingual/afri_mcqa_multiple_choice.py
```diff
@@ -0,0 +1,112 @@
...
+            "test split ships the four options already shuffled with no key. Options are "
+            "shuffled by a hash of the question, because the source lists the correct "
+            "answer first in every row. Questions with no photograph, questions listing "
+            "their own answer among their distractors, and repeats are dropped. "
+            "Construction script: scripts/data/afri_mcqa_multiple_choice/create_data.py."
+        ),
+        reference="https://arxiv.org/abs/2601.05699",
+        dataset={"path": _DATASET_PATH, "revision": _DATASET_REVISION},
+        type="VisionCentricQA",
+        category="it2t",
```
Samoed, line 54: Maybe also add audio variants?

PR #5376, opened 2026-09-04
fix: break ties by doc id when skipping the first retrieval result

serhiizghama:
Same #5092 tie-break bug, fourth site. `calculate_retrieval_scores` drops the top hit for `skip_first_result` tasks with a score-only sort, and Python's sort is stable — so when the top two scores tie, it's dict insertion order that decides which document gets discarded. The rest of the module has been breaking ties by `(score, doc_id)` since #5136 and #5241; this call was missed.

It affects the seven tasks that set `skip_first_result = True` (Covers80, SHS100K, and the i2i ones — CUB200, MET, RP2K, SOP, StanfordCars). Those are exactly the setups where the query also sits in the corpus, so an exact duplicate scoring 1.0 next to the self-match is the tie you'd expect. Feeding identical scores in two different orders on main:

```
self listed first        NDCG@1 1.0   Recall@3 1.0
duplicate listed first   NDCG@1 0.0   Recall@3 0.5
```

Same data, different dict order. With the tie-break both orders agree.

I haven't measured how often this fires on the real datasets — it needs an exact float tie at rank 1, so in practice duplicate items. And since it can move numbers on those seven tasks, it carries the same results-repo caveat as #5136.

Regression test sits next to the existing tie-break ones and fails on main.

### DIFF
--- a/mteb/_evaluators/retrieval_metrics.py
+++ b/mteb/_evaluators/retrieval_metrics.py
@@ -584,8 +584,12 @@ def calculate_retrieval_scores(  # noqa: PLR0914
     skip_first_result: bool = False,
 ) -> RetrievalEvaluationResult:
     if skip_first_result:
+        # tie-break by doc id descending, like every other ranking in this module (#5092);
+        # a score-only sort is stable, so which doc got dropped followed dict insertion order
         results = {
-            qid: dict(sorted(doc_scores.items(), key=lambda x: x[1], reverse=True)[1:])
+            qid: dict(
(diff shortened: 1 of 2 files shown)


Samoed (commented):
Can you compare results? E.q. `openai/clip-vit-base-patch16` on `CUB200I2IRetrieval`

serhiizghama:
Closing this — I checked on `openai/clip-vit-base-patch16` / `CUB200I2IRetrieval` and all 149 metrics come out identical, because no query on that task has a tie at rank 1 (the corpus has no exact duplicates). The impact I described isn't there. Sorry for the noise.

PR #5377, opened 2026-09-04
fix: update stale tokenizer revisions for Transformers v5

Pdevadiga45:
## Summary
The pinned revisions for three models predate tokenizer fixes by their authors.
Under Transformers v5 (verified with 5.12.0), the old revisions load a fallback
tokenizer: `paraphrase-multilingual-MiniLM-L12-v2` has no `tokenizer_class` set,
and the mmlw models declare `RobertaTokenizer` instead of `XLMRobertaTokenizer`
(with a broken `model_max_length`). This silently changes token IDs and embeddings.

## Changes
- Bump the pinned revision of 3 models to the authors' tokenizer-correction commits.
- Verified each target commit is tokenizer-config-only; model weights are
  bit-identical across the old and new pins for all three (LFS hashes checked).
- Expected effect: MTEB scores for these three models will shift, since the
  corrected tokenizers produce different token IDs.

## Validation
- Reproduced old vs. corrected tokenizer behavior with Transformers 5.12.0
- Verified target commits are tokenizer-config-only via the HF API
- Weights bit-identical across pins for all three model pairs
- Ruff checks pass


ayush1298:
good catch!
Did you check are there any other models as well whose revision gets changed?

Pdevadiga45:
Yes, I checked the registry more broadly, but I couldn’t find any other instances of this specific tokenizer mismatch under Transformers v5 beyond these three models

Samoed (approved):

PR #5378, opened 2026-09-04
fix: repair broken Vultron revision pins

Pdevadiga45:
## Summary

Update the pinned revisions for the three `vultr/VultronRetriever*` models. The current revisions belong to pre-migration repository histories and return HTTP 404, so the registered models cannot load.

The replacements are the earliest surviving commits in the current repositories, avoiding later model and preprocessing changes.

## Validation

- Confirmed the old revisions fail to resolve `config.json`
- Confirmed the replacement revisions resolve `config.json`
- `ruff check mteb/models/model_implementations/colqwen_models.py`
- `git diff --check`


ayush1298:
there are new revisions for all 3 models. Can you change revisions to them?

tests/test_models/test_colqwen_models.py
ayush1298, line 1: I dont think we need this test

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -835,7 +835,7 @@ def __init__(
     name="vultr/VultronRetrieverPrime-Qwen3.5-8B",
     model_type=["late-interaction"],
     languages=["eng-Latn", "fra-Latn", "deu-Latn", "spa-Latn", "ita-Latn", "por-Latn"],
-    revision="e8f3104b743a04b0d5f715b67117d687ae99ce51",
+    revision="37f9944abe9140f26b5d89809f6aad8132abea6e",
```
ayush1298: I was saying to update these revision with latest one at all 3 places

mteb/models/model_implementations/colqwen_models.py
```diff
@@ -96,13 +96,16 @@ def __init__(
 
         self.model = ColQwen3_5.from_pretrained(
             model_name,
+            revision=revision,
             device_map=self.device,
             adapter_kwargs={"revision": revision},
```
ayush1298: its already passed in `adapter_kwargs`. I dont think we need to pass explicitely again. @Samoed is this correct?

tests/test_models/test_colqwen_models.py
Pdevadiga45: yes, I agree. will drop this

mteb/models/model_implementations/colqwen_models.py
Samoed: Actually we don't need `adapter_kargs` at all, becuase this model is presented fully in repo and previosly we load it without any revision at all

Samoed (approved):

ayush1298:
@Samoed should we remove `adapter_kwargs` from https://github.com/embeddings-benchmark/mteb/blob/dca9d97e2deb44199b1270ee9fad70d5f6cae89c/mteb/models/model_implementations/colpali_models.py#L46 as well?

Samoed:
No. E.g. `vidore/colpali-v1.3` is real adapted only repo. It has only [adapter_model.safetensors](https://huggingface.co/vidore/colpali-v1.3/blob/main/adapter_model.safetensors) without full `model.safetensors`. I didn't check others, but there might be similar cases where model exists fully in repo and we don't really pin it's revision

ayush1298:
> No. E.g. `vidore/colpali-v1.3` is real adapted only repo. It has only [adapter_model.safetensors](https://huggingface.co/vidore/colpali-v1.3/blob/main/adapter_model.safetensors) without full `model.safetensors`. I didn't check others, but there might be similar cases

oh, I see. there is another https://huggingface.co/nomic-ai/nomic-embed-multimodal-7b/tree/main but it also has just `adapter_model.safetensors`. so nothing to change

Samoed:
Yes, but there might be other models that use `ColpaliWrapper`, but actually have full weights in repo

PR #5379, opened 2026-09-05
fix: isolate embedding caches by prompt type

Pdevadiga45:
## Summary

Separate `CachedEmbeddingWrapper` storage by prompt type. Previously, identical text encoded as a query and document within one task shared a cache entry, so the first role's embedding was silently reused for the other.

Prompt-less tasks retain the existing layout; prompt-aware caches use `<cache>/<task>/<prompt_type>`.

## Validation

- Added a regression test with role-dependent embeddings
- Existing NumPy and FAISS cache tests pass
- Retrieval cache integration test passes
- `ruff check` and `ruff format --check`


isaac-chung:
Hi, please first raise an issue, describe the issue in detail, then after a discussion with the maintainers, raise a PR if needed. Closing this now.

PR #5381, opened 2026-09-05
Enables `ANN401` (flake8-annotations: `any-type`)

ayush1298:
Enables `ANN401` (flake8-annotations: `any-type`), from #2416. All 50 sites resolved: 26 got a real type, 24 keep `Any` with a documented `noqa` explaining why it is correct there.

### mypy found two real bugs in the new annotations:
- `_iter_test_indices` — `y` is required by `_prepare_stratification` but was optional with no guard; now raises a clear `ValueError` instead of failing obscurely downstream.
- `_resolve_model_prompts` — `SentenceTransformer.prompts` is `dict[str, str | None]`, not `dict[str, str]`; two sites needed a `cast`, including a mutation into an invariant `dict`.

Remaining from `ANN`: `ANN001` (426), `ANN202` (132), `ANN201` (120).


mteb/abstasks/pair_classification.py
```diff
@@ -338,7 +339,10 @@ def _find_best_acc_and_threshold(  # noqa: PLR6301
         return max_acc, best_threshold
 
     def _find_best_f1_and_threshold(  # noqa: PLR6301
-        self, scores: Any, labels: NDArray[np.int64], high_score_more_similar: bool
+        self,
+        scores: npt.ArrayLike,
```
Samoed:
```suggestion
        scores: list[float],
```

Samoed (approved):

mteb/abstasks/pair_classification.py
ayush1298: made change, can you check it?

PR #5384, opened 2026-09-05
task: add COCO Modality Equivalence retrieval tasks

### LINKED ISSUE #5358: MOEB [Paper Experiment] - Modality Equivalence
We can measure per-direction retrieval scores, but we can't currently separate "this modality pair is hard" from "this dataset is hard": the content differs across our directional tasks.

One way we can do this comparison is by holding the semantic content constant and varying only the modality, and retrieving over a shared candidate pool. Any difference is then attributable to the modality.

MMEB-V3 does this with OmniSET (§ in https://arxiv.org/abs/2604.23321): semantically equivalent
tuples across text/image/video/audio, 12 directional tasks, shared pool with human-verified hard
negatives, kept out of the leaderboard and used only for diagnosis. Worth reading before starting;
their construction and its limitations are both instructive.

MMEB-V3 built OmniSET on MSCOCO. We currently have these:

- `MSCOCOI2TRetrieval` / `MSCOCOT2IRetrieval` - image ↔ text
- `SpokenCOCOA2IRetrieval` / `SpokenCOCOI2ARetrieval` - image ↔ real human speech
- `SpeechCocoA2IRetrieval` / `SpeechCocoI2ARetrieval` - image ↔ synthetic TTS speech

That's the full {image, text, audio} triple over the same content. For video, `VATEX*` already
covers 8 directions over {video, audio, text}.

Note their audio is TTS-generated, and they flag that synthetic generation "creates asymmetric
dependencies between modalities." We have both TTS (SpeechCoco) and human (SpokenCOCO) speech over
identical images and captions, so we can test that directly.

### Open for whoever takes this

- which base and which modalities to include
- how to build the shared candidate pool and hard negatives
- how many queries and how exactly to construct the set.

Rakshitha-Ireddi:
Adds six retrieval tasks over a shared 120-image MSCOCO candidate pool where every
  item is simultaneously available as an image, text caption, human-spoken caption
  (SpokenCOCO), and TTS caption (SpeechCoco). Because the pool is identical across
  all directions, score differences reflect modality difficulty rather than content
  difficulty. Closes #5358.

  **Tasks added:**
  - `COCOModalEquivT2IRetrieval` — text → image (baseline)
  - `COCOModalEquivA2IHumanRetrieval` — human speech → image
  - `COCOModalEquivA2ITTSRetrieval` — TTS speech → image
  - `COCOModalEquivI2TRetrieval` — image → text
  - `COCOModalEquivI2AHumanRetrieval` — image → human speech
  - `COCOModalEquivI2ATTSRetrieval` — image → TTS speech

  **Dataset:** `rakshi719/coco-modality-equivalence` (temporary namespace; will be
  transferred to `mteb/` on merge). Revision pinned in task file.

  Pool size: 120 images at the intersection of SpokenCOCO, SpeechCoco, and MSCOCO
  MBEIR — images present in all three sources with image, human audio, TTS audio,
  and text caption available.

  Data creation script: `scripts/data/coco_modality_equivalence/create_data.py`
  
  ## Test plan
  - [x] Built dataset, verified 120-item corpus ID alignment across all three sources
  - [x] Uploaded to HF, `_DATASET_REVISION` filled in
  - [x] All six tasks import and pass metadata validation


mteb/tasks/retrieval/eng/coco_modality_equivalence.py
```diff
@@ -0,0 +1,162 @@
...
+        category="t2i",
+        modalities=["text", "image"],
+        prompt={"query": "Find the image described by the caption."},
+        **_COMMON,
+    )
+
+
+class COCOModalEquivA2IHumanRetrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="COCOModalEquivA2IHumanRetrieval",
```
Samoed, line 102: Doesn't `SpokenCOCOA2IRetrieval` already test that?

mteb/tasks/retrieval/eng/coco_modality_equivalence.py
```diff
@@ -0,0 +1,162 @@
...
+    dialect=[],
+    sample_creation="found",
+    bibtex_citation=_BIBTEX,
+    is_beta=True,
+)
+
+
+class COCOModalEquivT2IRetrieval(AbsTaskRetrieval):
+    metadata = TaskMetadata(
+        name="COCOModalEquivT2IRetrieval",
```
Samoed, line 87: Doesn't original coco tests that?

mteb/tasks/retrieval/eng/coco_modality_equivalence.py
Rakshitha-Ireddi: Yes, SpokenCOCOA2IRetrieval tests audio-to-image retrieval, but it uses a different candidate pool than any text or TTS task. The problem is that if SpokenCOCOA2IRetrieval scores lower than a text-to-image task,
  you can't tell whether that's because speech is harder as a query modality, or because the SpokenCOCO candidate pool is harder than the MSCOCO pool. These tasks fix that by using the same ~1k image corpus for
  all six directions, so score differences are purely attributable to the modality and not the dataset.

mteb/tasks/retrieval/eng/coco_modality_equivalence.py
Rakshitha-Ireddi: Same reasoning — the existing COCO text-to-image task uses a different corpus than SpokenCOCO or SpeechCoco. Here, all six tasks share an identical candidate pool (the intersection of images present in all three
  sources), so t2i, a2i_h, and a2i_s scores are directly comparable. The original COCO task is not comparable against the SpokenCOCO task because the two corpora are different.

PR #5385, opened 2026-09-05
fix: defer datasets import to avoid slow cold-start when reading model metadata

### LINKED ISSUE #5382: [CI] Update Leaderboard Model List workflow running longer than before

From https://github.com/embeddings-benchmark/mteb/actions/workflows/update_leaderboard_models.yml?page=2, prior to Aug 18, 2026 this workflow would finish around 5min. After Aug 19 inclusive, run time triples and went upwards of 20min. The bottleneck is in the `Generate model list` step. 

We want:
- to understand what is causing that step to take so much more time than before
- to find a concise and suitable fix for that root cause

isaac-chung:
Fixes https://github.com/embeddings-benchmark/mteb/issues/5382

`scripts/generate_leaderboard_models.py` (and any process that builds `MODEL_REGISTRY`) regressed from ~5 min to 20+ min after #2970 merged on Aug 18, because `mteb/types/_encoder_io.py` gained a bare `from datasets import Dataset`, pulling `datasets`+`pyarrow` into every process that merely reads model metadata. This PR fixes that.

## Fixes

**`mteb/types/_encoder_io.py`** — move `from datasets import Dataset` inside `if TYPE_CHECKING:` and convert the three type aliases to string form (`QueryDatasetType: TypeAlias = "Dataset"` etc.). `datasets` is only needed for annotations, never at runtime.

**`bm25.py`, `bb25.py`, `random_baseline.py`, `pylate_models.py`, `cde_models.py`** — remove module-level `from mteb._create_dataloaders import …`; move each import inside the method where it's called. `_create_dataloaders` imports `datasets` at the top level, so hoisting these into the method body keeps `datasets` out of the registry-build path.

## Tests (`tests/test_import_hygiene.py`)

Three fast tests (total ~0.8 s) that guard this path without loading the full model registry:

- **AST (×2, static)** — parse `_encoder_io.py` and the five model files with the `ast` module and assert that `from datasets import …` / `from mteb._create_dataloaders import …` only appear inside `if TYPE_CHECKING:` blocks or method bodies, never at module scope. Zero imports; runs in milliseconds.
- **Subprocess (×1, runtime)** — load `_encoder_io.py` in isolation via `importlib.util` (stubbing parent packages to bypass `mteb/__init__.py`) and assert `datasets` is absent from `sys.modules`.


Samoed:
I don't think that slow down for 15 min comes from imports

isaac-chung:
What do you think?

Samoed:
I think problem somewhere else

isaac-chung:
Based on?

You're more than welcome to take a jab at it. I think it'd be much more productive once there's something concrete to point back to and chat about.

PR #5386, opened 2026-09-05
fix: isolate embedding caches by prompt type

### LINKED ISSUE #5383: `CachedEmbeddingWrapper` can reuse a query embedding for identical document text
### Describe the bug

`CachedEmbeddingWrapper` checks the cache before calling the wrapped model. Its cache identity includes the task and raw text, but not `prompt_type`.

This causes a collision when the same raw text appears as both a query and a document and the model encodes those roles differently. For example, MTEB configures E5 with `query:` or `passage:` selected by `prompt_type`. If the query is encoded first, its vector is cached under the raw text. When the same text is later requested as a document, the cache returns the query vector without calling the model, so the document prefix is never applied.

As a result, the cached document embedding can differ from the document embedding produced without the cache.

### To reproduce

Tested at commit `56fa2928dd43ac1c2695abe73c041fb51a37159b` with `mteb.get_model("intfloat/e5-small-v2")`, `QuoraRetrieval` task metadata, and the raw text `"How does the stock market work?"`:

1. Encode the text directly with `prompt_type=PromptType.query` and `prompt_type=PromptType.document`, giving `direct_query` and `direct_document`.
2. Wrap the same model in `CachedEmbeddingWrapper` with an empty cache directory. Encode the same text as a query, then as a document, giving `cached_query` and `cached_document`. Keep the task, split, and subset unchanged.

Comparing the returned vectors with NumPy (`import numpy as np`):

```python
np.allclose(direct_query, direct_document)    # False
np.allclose(cached_query, cached_document)    # True
np.allclose(cached_document, direct_query)    # True
np.allclose(cached_document, direct_document) # False
```

The document request returns the query embedding rather than its own encoding.

### Additional information

In the pinned `mteb/quora` dataset, 18 query strings also occur verbatim in the corpus, including the text above.

Would including `prompt_type` in the cache identity be the intended fix? Opening this for discussion as requested in #5379.

Yes

Pdevadiga45:
## What does this PR do?

Separates `CachedEmbeddingWrapper` caches by `prompt_type`, preventing identical raw input encoded as a query and document from sharing the first role's vector.

In-memory caches are keyed by `(task_name, prompt_type)`. Prompt-aware disk caches use `<cache>/<task>/<prompt_type>`, while calls with `prompt_type=None` retain the existing `<cache>/<task>` layout for compatibility.

This supersedes #5379, which was closed to follow the issue-first workflow.

Fixes #5383

## Tests

- `python -m pytest -q tests/test_models/test_cached_embedding_wrapper.py` — 5 passed
- Regression test fails on current `main` for the intended query/document collision
- `ruff check` and `ruff format --check` on changed Python files
- `git diff --check origin/main...HEAD`


docs/get_started/advanced_usage/cache_embeddings.md
```diff
@@ -22,6 +22,10 @@ model_with_cached_emb = CachedEmbeddingWrapper(model, cache_path="path_to_cache_
 results = mteb.evaluate(model_with_cached_emb, tasks=[task])
 ```
 
+For tasks that use query and document prompts, embeddings are stored separately
+in `<path_to_cache_dir>/<task_name>/<prompt_type>` so that inputs encoded with
+different prompts cannot share cached vectors.
```
Samoed:
```suggestion
For tasks that use query and document prompts, embeddings are stored separately in `<path_to_cache_dir>/<task_name>/<prompt_type>` so that inputs encoded with different prompts cannot share cached vectors.
```

tests/test_models/test_cached_embedding_wrapper.py
```diff
@@ -252,6 +253,44 @@ def test_caching_functionality(  # noqa: PLR0914
...
+            document_embeddings = cached_encode(PromptType.document)
+            cached_query_embeddings = cached_encode(PromptType.query)
+            cached_document_embeddings = cached_encode(PromptType.document)
+        finally:
+            wrapped_model.close()
+
+        np.testing.assert_allclose(query_embeddings, 1.0)
+        np.testing.assert_allclose(document_embeddings, 2.0)
+        np.testing.assert_allclose(cached_query_embeddings, query_embeddings)
+        np.testing.assert_allclose(cached_document_embeddings, document_embeddings)
```
Samoed, line 286: Shoudn't you also test here that document != query

tests/test_models/test_cached_embedding_wrapper.py
```diff
@@ -252,6 +253,44 @@ def test_caching_functionality(  # noqa: PLR0914
...
 
+    def test_cache_isolated_by_prompt_type(self, cache_dir: Path, monkeypatch):
+        model = DummyModel("test_model", revision=None)
+        encode = Mock(
+            side_effect=[
+                np.full((1, model.embedding_dim), 1.0, dtype=np.float32),
+                np.full((1, model.embedding_dim), 2.0, dtype=np.float32),
+            ]
+        )
+        monkeypatch.setattr(model, "encode", encode)
```
Samoed: Better to update in class than set mock

Samoed (approved):

PR #5389, opened 2026-09-05
Support datasets with interleaved modality coverage

### LINKED ISSUE #4159: Allow for interleave modality dataset

ayush1298:
Closes #4159

Problem:  A multimodal dataset had to give every sample the same modalities. A dataset mixing text-only, image-only, and both — {"text": ["text 1", None, "text 3"], "image": [None, img, None]}  is not supprted and will give error at multiple places.

None marks an absent value in an input column; "" remains a genuinely-empty text. Models receive "" for absent text (so no text encoder sees a None) and None for every other absent modality. Absence is only allowed in input columns; a None in id, a label, or a score still raises.

Changes: 
1. `_create_dataloaders.py` — `_corpus_to_dict,` `_combine_queries_with_instruction_text,` `_convert_conv_history_to_query` and `_transform_image_to_rgb` tolerate absence; the collate function normalizes missing input values; the conversation-vs-text branch no longer decides from a row that may carry no text.
2. `_statistics_calculation.py` — `compute_*_hashes` emit a `MISSING_MODALITY_HASH` sentinel so hash lists stay row-aligned (_count_samples_in_train zips them positionally); `calculate_*_statistics` exclude absent rows and handle a fully-absent modality.
3. `modality_utils.py` — `is_modality_present,` `get_present_indices,` `is_interleaved` for models.
4. `clip_models.py` — `get_fused_embeddings` encodes each row from the modalities it carries, as the reference pattern.
5. `random_baseline.py` — averages over present modalities.
6. `_hash_utils.py` — absent modalities contribute nothing to the cache key.
7. Mock task `MockAny2AnyRetrievalInterleavedIT2IT` (in the MIEB grid, so it runs through every model–task integration test), `tests/test_interleaved_modalities.py,` and sections in both contributing guides.

Backwards compatibility:
Nothing changes for existing data; every edit is gated on None, which no working task can contain. Verified against a HEAD worktree that  Descriptive statistics identical, Model scores identical, Embedding caches stay valid, CLIP fusion unchanged. So nothing needs re-running.

- 9 existing t2it tasks (MMDocIRT2ITRetrieval, DanRAGT2ITRetrieval, SDSKoPubVDRT2ITRetrieval, six Vidore3*.v2) are already semantically interleaved. page-image corpora where some pages have no extractable text expressed with "". They work today and not touched in this PR: keeping "" distinct from None is what preserves their min_text_length: 0 statistics and what their models are fed.


Samoed (changes requested):
I'm not sure that we want this for now, because this would reqire massive refactoring almost every model implementation and every evaluation part. Also I don't think that putting `None` inside columns is a godd idea. I think this issue reqires a lot more discussion

ayush1298:
> I'm not sure that we want this for now, because this would reqire massive refactoring almost every model implementation and every evaluation part. Also I don't think that putting `None` inside columns is a godd idea. I think this issue reqires a lot more discussion

agree on refactoring of model implementations and later we have to rerun everything. I will add these on issue and we can discuss there. Will marked this as a draft. 

On Putting `None` inside columns. `None` is `HF/Arrow's` native nullability and needs no schema change; and 9 existing tasks (MMDocIRT2ITRetrieval, DanRAGT2ITRetrieval, six Vidore3*.v2, SDSKoPubVDRT2ITRetrieval) already ship documents with no text, using "". So we already have tasks with interleaved data, it's just implicit and undocumented

ayush1298:
@Samoed should I close these, what do you think?

Samoed:
We can close this for now

ayush1298:
> We can close this for now

okay, sure closing it then

PR #5390, opened 2026-09-05
enable ANN001, ANN201 and ANN202 for `mteb/leaderboard/` and `mteb/tasks/`

ayush1298:
Annotates `mteb/leaderboard/` and `mteb/tasks/`, from #2416

`ANN001`, `ANN201` and `ANN202` move out of the global `ignore` into two scoped per-file exemptions, so they are now enforced across the rest of the package

| Path | Status |
|---|---|
| `mteb/tasks/*` | **enforced** — 133 annotations here |
| `mteb/leaderboard/*` | **enforced** — 45 annotations here |
| `tests/*` | exempt — not type-checked at all (`make mypy` covers `mteb` only), so annotations there would be unverified |

### `mteb/tasks` (133)

- **34 repeated signature shapes** — `dataset_transform` overrides (23) and the row-mapping helpers passed to `Dataset.map` / `.filter` (11).
- **~60 retrieval loaders** — parameters from their (unambiguous) names, and return types derived per function from how `corpus` is actually built, so the nesting matches the code: `dict[str, Dataset]` where split-keyed, `dict[str, dict[str, Dataset]]` where lang-keyed. Two loaders (`jina_vdr`, `vidore2`) genuinely branch on whether `langs` is passed and produce either shape, so those get an explicit union.
- **12 outliers** by hand, including four module-level functions that take `self` and are assigned as methods (`self: AbsTaskRetrieval`).
- **39 imports** moved into `TYPE_CHECKING` blocks, since the new annotations are annotation-only.

### `mteb/leaderboard` (45): 
Gradio callbacks, plot helpers and table formatters.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -400,10 +400,6 @@ preview = true
 "mteb/tasks/*" = [
     "D",
     "DOC",
-    # TODO: annotate and remove -- needs a type inferred per call site
-    "ANN001",
-    "ANN201",
-    "ANN202",
 ]
 "mteb/models/model_implementations/*" = [
     "D",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -386,6 +386,11 @@ preview = true
     "S603",  # allow subprocess calls in tests
     "PLC2701",  # Private name import
     "PLR6301",  # Method could be a function - pytest test methods are instance methods
+    # tests are not type-checked (`make mypy` covers `mteb` only), so annotations
+    # here would be unverified;
+    "ANN001",
+    "ANN201",
+    "ANN202",
 ]
 "mteb/mocks/*" = [
     "RUF012",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -363,11 +363,6 @@ ignore = [
     "PT018",    # composite assertions; splitting `assert a and b` is not always clearer
     "RET504",   # unnecessary-assign; named intermediate results often aid readability
     "SIM108",   # ternary over if/else is a style preference, and displaces inline comments
-    # TODO: burn down in follow-ups. These need a type inferred per call site (ruff can fix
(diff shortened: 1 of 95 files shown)


pyproject.toml
```diff
@@ -400,10 +395,18 @@ preview = true
...
+    "ANN202",
 ]
 "mteb/models/model_implementations/*" = [
     "D",
     "DOC",
+    # TODO: annotate and remove -- needs a type inferred per call site
+    "ANN001",
+    "ANN201",
+    "ANN202",
 ]
```
Samoed, line 411: I think this parts can ignore `ann`. Tests also can ignore `ANN` mostly

ayush1298: okay, then should I ignore for tests and leaderboard and add for mteb/tasks (133 changes) and mteb/models/model_implementations (340 changes).

ayush1298: or are you saying ignoring for all test, tasks, leaderboard, model_implementations?

Samoed: I think for models would be good to activate since we're updating their implementations sometimes. I don't think we need for tests. You can keep for leaderboard for consistency

ayush1298: okay, will remove tests thing. keep LB and for model will add in a separate PR. What about tasks?

Samoed: If this is easy to do you can activate this, if reqires a lot of time, just ignore

ayush1298: okay, will just check. Most of them follows under certain catagory, so easy to check them atleast

ayush1298:
@Samoed also added for `mteb/tasks` here. Will add 1 more PR for `model_implementations`

Samoed (approved):

Issue #5243, opened 2026-08-19
Dataset Versioning (v2 suffix)


AdnanElAssadi:
### Description of the feature

The Clotho fix (#5062) highlighted an issue with MTEB task versioning.

We had to introduce `ClothoT2ARetrieval.v2` because removing 905 invalid samples would otherwise change the task under which existing model results were produced. While this preserves reproducibility, `.v2` is ambiguous: it can refer to an upstream dataset release or an MTEB-side task construction change.

### Proposal

Investigate whether MTEB should store **stable sample-level identifiers/provenance** for task data and submissions.

If results were tied to stable sample IDs, removing invalid samples could potentially be handled without creating an entirely new task version, while still preserving compatibility with existing submissions.

This could also allow us to distinguish between:

* upstream dataset versions
* MTEB task-construction revisions
* genuinely new tasks

### Some questions/issues

* What sample-level information would need to be stored?
* How would existing submissions/results remain compatible after samples are removed?
* Should upstream dataset versions and MTEB task revisions use separate versioning schemes?
* What changes would this require to the data and submission infrastructure?

Rakshitha-Ireddi:
Hi @AdnanElAssadi , I have opened a PR that partially addresses this issue.

PR #5398, opened 2026-09-05
add version_reason to TaskMetadata to partially address .vN versioning ambiguity (#5243)

Rakshitha-Ireddi:
Partially addresses #5243.

## Problem

The `.vN` suffix is ambiguous: it can indicate either that the upstream dataset released a new version or that MTEB fixed a bug in its own task construction.

This makes it impossible for tooling and leaderboards to programmatically distinguish between upstream dataset updates and MTEB-side task revisions.

## What this PR does

Adds `version_reason` to `TaskMetadata`:

```python
version_reason: Literal[
    "upstream_dataset_update",
    "mteb_task_correction",
] | None
```

The two values distinguish:

* `mteb_task_correction` — MTEB fixed an issue in task construction, such as incorrect labels, empty segments, or incorrect splits. The upstream dataset itself was unchanged.
* `upstream_dataset_update` — the upstream dataset published a new version.

This directly addresses question 3 from #5243 by separating upstream dataset versions from MTEB task-construction revisions at the metadata level.

## Analysis

After auditing all 204 existing `.vN` tasks in the repository, every versioned task examined was found to represent an MTEB-side correction rather than an upstream dataset release.

This confirms the versioning ambiguity identified in #5243.

## What this PR does NOT do

The core proposal in #5243 is to store stable sample-level identifiers so that removing invalid samples (for example, Clotho's 905 empty segments) would not require creating a new task name. Existing results could instead remain valid by anchoring them to the original sample IDs.

Implementing this requires changes to the submission and leaderboard infrastructure and is therefore left for a follow-up.

## Changes

* `mteb/abstasks/task_metadata.py`

  * Adds the `TaskVersionReason` type and `version_reason` field.
* 153 task files

  * Backfills `version_reason="mteb_task_correction"` for all `.v2`/`.v3` tasks with `adapted_from` (203 tasks total).
  * `LexRetrieval.v1` is intentionally skipped because it represents an initial-edition label rather than a correction.
* `tests/test_tasks/test_versioned_tasks.py`

  * Enforces that future `.vN` tasks with `adapted_from` declare a `version_reason`.


Samoed:
I don't think this is good solution, because not clear what's changed. This issue require more discussion

PR #5399, opened 2026-09-05
feat: add model_type and similarity_fn_name overrides to ModelMeta.load_model

### LINKED ISSUE #5300: Make possible to customize model type during evaluation

We have some models that have multiple working format, e.g. dense and late-interaction - jina-v4, dense and sparse - UEmbed https://github.com/embeddings-benchmark/mteb/pull/5200. 

I've added custom ModelMeta for jina v4 to change `model_type` based on kwargs https://github.com/embeddings-benchmark/mteb/blob/145988125f8d438285aefc64629c519167591168/mteb/models/model_implementations/jina_models.py#L384-L402 and in UEmbed have setter for such case https://github.com/embeddings-benchmark/mteb/blob/145988125f8d438285aefc64629c519167591168/mteb/models/model_implementations/uembed_models.py#L483-L509 Maybe we can create more general approach for this with possibility to override by model

Rakshitha-Ireddi:
Closes #5300.

## Problem

Models that support multiple operating modes (dense/sparse, single-vector/multi-vector) previously had to use ad-hoc patterns to update `model_type` in result metadata:

* `Jinav4ModelMeta` — a custom `ModelMeta` subclass that overrides `load_model()`.
* `UEmbed` — a custom `mteb_model_meta` property setter.

There was no general, documented mechanism for handling these overrides.

## Solution

This PR supports two clean, general patterns.

### 1. Caller-specified overrides (new)

`ModelMeta.load_model()` now accepts `model_type` and `similarity_fn_name` as explicit parameters, consistent with the existing `embed_dim` parameter:

```python
model = meta.load_model(model_type=["late-interaction"])
model = meta.load_model(similarity_fn_name="dot")
```

Both values are recorded in `experiment_kwargs` for tracking, but are **never forwarded to the loader function**.

### 2. Model-internally-determined overrides (existing, now documented)

Models that determine their type from internal state (for example, based on a pooling or `vector_type` argument) can override the `mteb_model_meta` setter.

This is the same pattern already used by `UEmbed` and is now documented as the general mechanism for model-internal metadata overrides.

## Cleanup

`Jinav4ModelMeta`, a `ModelMeta` subclass that existed solely to update `model_type` after loading, is removed.

`JinaV4Wrapper` now uses the `mteb_model_meta` setter pattern instead, making it consistent with `UEmbed`.

## Changes

* `mteb/models/model_meta.py`

  * Adds `model_type` and `similarity_fn_name` parameters to `load_model()`.
* `mteb/models/model_implementations/jina_models.py`

  * Removes `Jinav4ModelMeta`.
  * Adds the `mteb_model_meta` setter to `JinaV4Wrapper`.
* `tests/test_models/test_model_meta.py`

  * Adds 3 tests covering the new override behavior.


Samoed (changes requested):
In this pr a lot of changes from other prs which is improssible to review

Issue #3232, opened 2025-10-01
Observed potential issues in `MLDR` code


AlejandroBaron:
While reviewing the code, I noticed the following potential problems:

1. **Positive examples handling**  
   - Currently, only the first positive is retained.  
   - This isn’t an issue if there’s guaranteed to be only one positive, but it could be misleading for future readers of the code.  

2. **Negative examples handling**  
   - Negatives are being dropped, which might have an impact, especially for the training split.  
   - This behavior could reduce the effectiveness of the dataset for certain use cases.  
https://github.com/embeddings-benchmark/mteb/blob/867105f60be27fb62d134b457a116ded39172c3c/mteb/tasks/Retrieval/multilingual/MultiLongDocRetrieval.py#L53

3. **Efficiency of dataset loading**  
   - The dataset splits specified by the user are not respected during loading.  
   - Instead, **all splits are loaded during each iteration of the loop**, which is inefficient.  
   - Ideally, only the required subset should be loaded to improve performance.  
https://github.com/embeddings-benchmark/mteb/blob/867105f60be27fb62d134b457a116ded39172c3c/mteb/tasks/Retrieval/multilingual/MultiLongDocRetrieval.py#L48

Samoed:
Recently I've reuploaded this dataset and noticed these problems too. 

> Currently, only the first positive is retained.
This isn’t an issue if there’s guaranteed to be only one positive, but it could be misleading for future readers of the code.

Mostly it is only one positive, but I'm not totally sure

> Negative examples handling

For now, this is a retrieval task and all documents from the corpus are used during retrieval. This task should be converted to reranking task for proper usage of hard negatives

PR #5402, opened 2026-09-07
test: cover MLDR split and positive qrels loading

oliarmak-web:
## Summary
- add deterministic offline MLDR loading coverage for French, Spanish, and Portuguese
- verify a requested `test` split does not load `dev`
- verify all positive qrels survive when a query has multiple positives
- verify hard-negative documents remain retrieval candidates in the corpus

## Context
#3232 was filed against MLDR's pre-v2 custom loader. That loader fetched a complete split mapping and indexed only `positive_passages[0]`. Current main uses the shared `RetrievalDatasetLoader`, which passes the requested split explicitly and groups every qrels row by query.

The reported negative-list behavior is intentional for retrieval: negatives stay in the corpus but are not marked relevant. MLDR now also has a separate reranking task for candidate-list evaluation. This PR therefore locks the current behavior with focused regression tests instead of reintroducing a custom loader or changing published scoring semantics.

## Validation
- `pytest tests/test_tasks/test_load_data.py -q` (10 passed)
- full non-dataset suite (7,058 passed, 300 skipped)
- repository-wide Ruff format/check and typos checks
- mypy (1,717 source files)

## Compatibility and limitations
No production code, dataset revision, public API, or scoring behavior changes. The published MLDR revision currently has one positive qrel per query; the miniature multi-positive fixture protects future compatible data.

Related to #3232.

### DIFF
--- a/tests/test_tasks/test_load_data.py
+++ b/tests/test_tasks/test_load_data.py
@@ -50,3 +50,109 @@ def test_multilingual_retrieval_load_data(task):
     assert mock_load.called
     assert task.dataset is not None
     assert len(task.dataset) == 1
+
+
+@pytest.mark.parametrize(
+    ("language", "query", "positive_passages", "negative_passage"),
+    [
+        (
+            "fr",
+            "Pourquoi l’élève étudie-t-il à l’université de Montréal ?",
+            [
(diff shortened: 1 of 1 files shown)


Samoed:
We don't add tests for tasks specificly. We're pin revisions to not have unexpected changes in dataset

PR #5403, opened 2026-09-07
Add model: Singaraj/morisien-embed-v1.5

LK-maker-007:
`Singaraj/morisien-embed-v1.5` is a Mauritian Creole (`mfe`) sentence embedding model fine-tuned from LaBSE. It succeeds `Singaraj/morisien-embed`, already in `mteb`, which was fine-tuned from multilingual-e5-base. That entry now carries `superseded_by`.

On a FLORES+ pool with [xSIM++](https://arxiv.org/abs/2306.12907) distractors (996 queries, 45,029 passages) the older model reaches an error rate of 0.3504 against untrained LaBSE's 0.3343. This one reaches 0.2932, McNemar exact p = 0.00083. Preprint: https://doi.org/10.5281/zenodo.21877805

MorisienMTBitextMining F1, run through the `ModelMeta` on this branch. LaBSE and the existing entry reproduce their published means in the same runs.

| model | mean F1 |
|---|---|
| LaBSE | 0.848 |
| Singaraj/morisien-embed | 0.925 |
| **Singaraj/morisien-embed-v1.5** | **0.928** |


mteb/models/model_implementations/morisien_models.py
```diff
@@ -33,3 +33,35 @@
...
+    memory_usage_mb=1799,
+    embed_dim=768,
+    license="mit",
+    max_tokens=256,
+    reference="https://huggingface.co/Singaraj/morisien-embed-v1.5",
+    similarity_fn_name=ScoringFunction.COSINE,
+    framework=["Sentence Transformers", "PyTorch", "safetensors"],
+    use_instructions=False,
+    superseded_by=None,
+    adapted_from="sentence-transformers/LaBSE",
```
ayush1298, line 56: add model link: `https://huggingface.co/sentence-transformers/LaBSE`

LK-maker-007: `adapted_from` is resolved as a lookup in `model_meta.py:1331` to inherit the base model's training datasets. `sentence-transformers/LaBSE` resolves; the URL raises KeyError, and the surrounding try/except swallows it, so it would fail silently. No effect today since LaBSE's `training_datasets` is empty, but it breaks if that is ever populated. Happy to change it if you'd still prefer the link.

Samoed: It's correct now

Samoed (approved):

Issue #5362, opened 2026-09-01
MOEB [Paper Experiment] - Sampling Budget


AdnanElAssadi:
How much media do we actually need per task?

Video and audio encoding dominate MOEB's evaluation cost. 

Question: what is the smallest frames-per-video and seconds-per-audio budget that preserves model
rankings?

We can report speedup plus Spearman against the full setting. If for example 8 frames matches 32, that's 4x off our most expensive modality.

Needs a small number of models run at several sampling budgets.

Rakshitha-Ireddi:
Pilot results: audio sampling budget on COCOModalEquiv

  Tested 5 budgets (1s, 2s, 3s, 5s, full) on a2i_h (human audio, avg 4.2s clips) and a2i_s (TTS audio, avg 3.6s clips). Two omni models: jina-v5-nano and jina-v5-small. Metric: nDCG@10 + Spearman rank correlation
  vs full-budget baseline (query-level).

  Script: scripts/analysis/sampling_budget.py

  Results

  Human audio (a2i_h, avg clip 4.2s)

| Budget | Speedup | Jina-v5-nano NDCG@10 | Jina-v5-small NDCG@10 | Spearman (avg) |
|---|---:|---:|---:|---:|
| 1s | **4.2×** | 0.251 | 0.254 | 0.36 |
| 2s | **2.1×** | 0.502 | 0.658 | 0.79 |
| 3s | **1.4×** | 0.665 | 0.807 | 0.91 |
| 5s | 1.0× | 0.735 | **0.861** | 0.99 |
| Full | 1.0× | 0.734 | 0.857 | 1.00 |

  TTS audio (a2i_s, avg clip 3.6s)

  | Budget | Speedup | Jina-v5-nano NDCG@10 | Jina-v5-small NDCG@10 | Spearman (avg) |
|---|---:|---:|---:|---:|
| 1s | **3.6×** | 0.460 | 0.591 | 0.65 |
| 2s | **1.8×** | 0.593 | 0.754 | 0.88 |
| 3s | **1.2×** | 0.658 | 0.831 | 0.96 |
| 5s | 1.0× | 0.673 | **0.847** | 1.00 |
| Full | 1.0× | 0.676 | **0.847** | 1.00 |

  Answer
  
[...]

PR #5404, opened 2026-09-07
[audio] Fix MSCLAP wrapper sample-rate handling and MuQ NumPy buffer assignment

outerheaven199X:
**Summary**

While running a center-crop sampling-budget study (issue #5362) at commit
`61129c370bb9a38275ac39da5ba5a6861b4c9056`, we hit two wrapper defects that
block truthful native baselines:

1. **MSClapWrapper rate mismatch.** The wrapper hardcodes 48 kHz, writes
   48 kHz WAVs and calls msclap with `resample=False`, but msclap 1.3.4's
   config sets `sampling_rate: 44100`, so the model receives 6.43 s of
   actual 48 kHz waveform treated as 7 s at 44.1 kHz (308,700 samples).
   Fix: derive the truthful rate from the package config
   (`self.model.args.sampling_rate`) so the AudioCollator resamples and the
   bridge WAV is written at the rate msclap expects (package preprocessing
   path unchanged).

2. **MuQ wrapper NumPy assignment failure.** The wrapper assigns a NumPy
   array directly into a Torch buffer, which raises before inference.
   Minimal fix: `torch.from_numpy(arr)` before the same destination
   assignment; no change to casting semantics, resampling, padding,
   windowing, pooling, or similarity.

Also documents (no code change): the LAION CLAP wrappers process a fixed
10-second window regardless of supplied duration — crop savings cannot
propagate through a fixed window.

**Testing**: source-level isolation tests reproduce both defects and verify
the fixed patterns (rate/tone/duration through the pinned package methods;
NumPy bridge). Corrected paths pass full embedding parity against the
explicitly corrected reference. Defective unchanged wrappers are preserved
and separately labeled, so existing cached results remain interpretable.

**Evidence**: preregistered 4-model × 2-task crop study (24 primary
comparisons, paired cluster bootstrap, matched GPU timing) against the
corrected/compatibility references; see the issue #5362 comment for results
and limitations.

**Requested review focus**: corrected MSCLAP path as wrapper fix vs separate
model entry; whether the MuQ fix should carry a compatibility-reference
disclosure in model metadata.

### DIFF
--- a/tests/test_models/test_audio_wrapper_fixes.py
+++ b/tests/test_models/test_audio_wrapper_fixes.py
@@ -1,0 +1,127 @@
+"""Source-level tests for the MSCLAP wrapper rate fix and the MuQ NumPy bridge.
+
+These tests mirror the pilot evidence for the audio sampling-budget study
+(issue #5362): the MSCLAP wrapper must derive its sampling rate from the
+msclap package config (not hardcode 48000), and the MuQ wrapper must not
+assign a NumPy array directly into a Torch buffer.
+
+They do not require model weights: the MuQ test runs the real wrapper code
+path on a stubbed model, and the MSCLAP test inspects the wrapper source for
+the defect pattern (the rate must come from the package config before the WAV
+bridge, and resample must stay False to preserve the package's own
+preprocessing path).
+"""
+
+from __future__ import annotations
+
+import ast
+import inspect
+from pathlib import Path
+
+import numpy as np
+import pytest
+import torch
+
+WRAPPERS_DIR = Path(__file__).resolve().parents[2] / "mteb" / "models" / "model_implementations"
+
+
+class TestMuQNumpyBridge:
+    def test_numpy_array_into_torch_buffer_no_longer_raises(self):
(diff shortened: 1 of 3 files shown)


tests/test_models/test_audio_wrapper_fixes.py
ayush1298, line 1: no need of tests for models

ayush1298:
good catch. @Samoed can you check it? Is it bug or did we reduce `sampling_rate` for some reason?

Samoed (commented):
Yes, seems like a bug. In the paper they set the sampling rate to `44100`. @AdnanElAssadi you've added this model, maybe you remember why you set the sampling rate like this?

outerheaven199X:
If there's any other way I can help, lmk~!

AdnanElAssadi:
oh let me take a look and ill get back to you

tests/test_models/test_audio_wrapper_fixes.py
ayush1298: can you remove these test file

ayush1298:
@AdnanElAssadi did you get a chance to look at this?

PR #5414, opened 2026-09-08
fix: defer torch-backed re-exports in mteb.models

ayush1298:
Third PR in the series making mteb importable without torch. Refs #5063. Follows #5329, #5350.

#### The problem: ####
`mteb/abstasks/abstask.py` needs three protocols (EncoderProtocol, CrossEncoderProtocol, SearchProtocol) for `isinstance` checks. They live in `mteb/models/models_protocols.py,` which is already torch-free; but importing them pulled in torch anyway, putting it on the import path of `AbsTask` and everything downstream `(task_result,` `benchmark,` `result_cache).`

My first attempt was to import from the submodule directly rather than the package. But importing `mteb.models.models_protocols` still executes `mteb/models/__init__.py` first. That's why the fix has to live in the `__init__` itself, and required module-level `__getattr__` mechanism: there is no import-site workaround, and the only alternative is dropping the re-exports, which breaks the public API across the 91 files that `import from mteb.models`.

#### The change: ####
ModelMeta and the protocols stay eager and both are torch-free. 6 torch-backed wrapper modules are listed in `_LAZY_ATTRIBUTES` and resolved on first access by `__getattr__,` which caches into globals() so each name pays the lookup once. `__all__` is unchanged.  Result is `mteb.abstasks.abstask's` own dependency closure drops from 3061 to 2313 modules, with torch no longer among them.

`import mteb `is unchanged, and to be clear: no user sees a speedup from this yet. `mteb/__init__.py` is still eager, so importing any mteb submodule today still loads torch and all ~6800 modules. This is a prerequisite step, as were PRs 1 and 2. `task_result` and `benchmark` are still not clean either — they import mteb.abstasks, whose __init__ pulls in every AbsTask subclass, and those import `_evaluators` / `_create_dataloaders` at module scope. That's the next PR.

#### Testing:  ####
Four new tests added all confirmed to fail when one eager wrapper import is restored.

### DIFF
--- a/mteb/models/__init__.py
+++ b/mteb/models/__init__.py
@@ -1,6 +1,7 @@
-from .cache_wrappers import CacheBackendProtocol, CachedEmbeddingWrapper
-from .compression_wrappers import CompressionWrapper
-from .hybrid_wrappers import HybridSearch
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
 from .model_meta import ModelMeta
 from .models_protocols import (
     CrossEncoderProtocol,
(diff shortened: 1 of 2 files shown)


ayush1298:
On the profiling script and flame graphs, its stubbing measures something no user experiences. Concretely, on the real path today, `import mteb.types`, `import mteb.models.model_meta` and `import mteb.abstasks.abstask` all load torch and ~6800 modules, because `mteb/__init__.py` runs first. So none of the isolated timings in this series are user-visible yet, they describe a module's own dependency closure.

I did try `flameprof.` It crashes with ZeroDivisionError on the` import mteb` profile, in both --format svg and --format log, at every --threshold.  

So my suggestion is to do profiling and actual performance improvement analysis once all module-level changes land and `mteb/__init__.py` is lazy.

ayush1298:
also @Samoed you mentioned not to use `__getattr__` type thing, but here in PR description, I mentioned that other option will cause APIs to change that we dont want. And I didnt found anything else apart from these

mteb/models/__init__.py
```diff
@@ -1,28 +1,73 @@
...
+    "OpenAIAPITokenEmbedWrapper": ".openai_wrappers",
+    "SearchCrossEncoderWrapper": ".search_wrappers",
+    "SearchEncoderWrapper": ".search_wrappers",
+    "SentenceTransformerEncoderWrapper": ".sentence_transformer_wrapper",
+    "SparseEncoderWrapper": ".sentence_transformer_wrapper",
+    "sentence_transformers_loader": ".sentence_transformer_wrapper",
+}
+
+
+def __getattr__(name: str) -> object:
```
Samoed, line 55: I really don't like changing `getattr` of module. You can just move imports inside implementations

Samoed (changes requested):

mteb/models/__init__.py
ayush1298: any reason why not to change `getattr`, just for my understanding.
Will changing import inside implementations will solve issue fully?

Samoed: If we import any class from this file this module would still load torch. For examle in retrieval file https://github.com/embeddings-benchmark/mteb/blob/e98503e72c538e1f901cba40a22db0bfc98c79d0/mteb/abstasks/retrieval.py#L18-L24

ayush1298: got it. So, I made some more analysis: 

```
import mteb.models                            -> torch: False
from mteb.models import SearchEncoderWrapper  -> torch: True
import mteb.abstasks                          -> torch: True (3242 modules)
```

`__getattr__` only defers until the first access, and `retrieval.py` accesses at module scope, so `mteb/abstasks/__init__.py` pulls torch anyway. The isolated `abstask.py` improvement never propagates to anything real. 

I think I should work in reverse way like in order:

1. `_evaluators/,` `similarity_functions.py,` `_create_dataloaders.py` — imports into functions
2. `abstasks/` `subclasses` — the 18 module-level `mteb.models` / `mteb._evaluators` imports into the methods that use them. `mteb/abstasks/__init__.py` goes torch-free with no changes to the `__init__` itself
3. models/ wrappers — same thing, but will done once above both are there. 

@Samoed what do you think, will it be better approach?

Samoed: Yes, I think this is good

ayush1298: okay then I will close this PR and open new one targeting 1st step above

PR #5420, opened 2026-09-08
Update FlowVec commit hash

Tirth-021:


mteb/models/model_implementations/sentence_transformers_models.py
```diff
@@ -913,7 +913,7 @@
 FlowVec_v1 = ModelMeta(
     loader=SentenceTransformerEncoderWrapper,
     name="LingoIITGN/FlowVec-v1",
-    revision="53671cb1a58225b91c0469d411500aef082c74a1",
+    revision="3c00af32fa13bee4c7b19b159b9dd4fe50a49e1e",
```
Samoed, line 916: What was changed?

Tirth-021: The file structure was changed which created a new hash, I want the results repo to access this new hash.

Samoed (approved):

PR #5421, opened 2026-09-08
fix: isolate precision-specific result caches

### LINKED ISSUE #5394: Precision-specific evaluations can reuse default cached results
### Describe the bug

`mteb.evaluate(..., encode_kwargs={"precision": "int8"})` can save a precision-specific result in the default model cache namespace. A later evaluation without `precision` then returns the quantized result instead of running the model at its default precision.

`evaluate()` captures the model metadata and checks the cache before encoding. `SentenceTransformerEncoderWrapper` and `InstructSentenceTransformerModel` only add `output_dtypes` to their experiment metadata inside `encode()`, after that lookup. The final save still uses the earlier metadata snapshot.

Expected: evaluations using different precision settings should not reuse each other’s cached results.

### To reproduce

Reproduced on current `main` (`03e73a4c83e46c0398753b2a3bce4cb66368cd99`) using a local, no-download Sentence Transformers BoW model and a deterministic classification task:

```text
INT8 score:                0.583333
Default, same cache:       0.583333
Default, clean cache:      0.916667
```

The default evaluation using the same cache reused the INT8 result. A separate counting-model reproduction confirmed that it made zero encoder calls.

The INT8 result was written to `results/<model>/<revision>/<task>.json` rather than an `experiments/<experiment-name>/` path. The production-supported precision values overlapping MTEB's `OutputDType` are `int8`, `uint8`, and `binary`.

### Additional information

Suggested fix: set precision metadata before cache lookup, remove the late metadata updates from both wrappers, and test that evaluations with different settings do not share cached results.

Yes

Pdevadiga45:
## Purpose

- fix: resolve Sentence Transformers precision into evaluation metadata before result-cache lookup and reuse it for all saves
- keep backend-specific option handling outside generic cache code, including for lazy factory loaders
- add regression coverage for default and quantized cache isolation and supported precision values

Fixes #5394

## Tests

- python -m pytest tests/test_evaluate.py -q
- python -m pytest tests/test_import_hygiene.py -q
- Ruff check and format check

### DIFF
--- a/mteb/evaluate.py
+++ b/mteb/evaluate.py
@@ -16,6 +16,7 @@ from mteb.abstasks.abstask import AbsTask
 from mteb.abstasks.aggregated_task import AbsTaskAggregate
 from mteb.benchmarks.benchmark import Benchmark
 from mteb.cache import ResultCache
+from mteb.models._evaluation_meta import resolve_evaluation_model_meta
 from mteb.models.model_meta import ModelMeta
 from mteb.models.sentence_transformer_wrapper import (
     CrossEncoderWrapper,
@@ -95,6 +96,7 @@ def _sanitize_model(
 
 def _evaluate_task(  # noqa: PLR0913, PLR0914
     model: MTEBModels,
+    model_meta: ModelMeta,
     task: AbsTask,
     *,
     splits: dict[SplitName, list[HFSubset]],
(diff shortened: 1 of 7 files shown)


Samoed:
Hi! Already created pr for that issue https://github.com/embeddings-benchmark/mteb/pull/5418. I'll finish it soon

PR #5426, opened 2026-09-09
fix: make _evaluators and _create_dataloaders torch-free

ayush1298:
Third step toward making mteb importable without torch. Refs #5063. Follows #5329 and #5350.

Per review feedback, this removes the torch dependency rather than deferring it at a package boundary,  no module `__getattr__,` no changes to any `__init__.py.`

Changes:
1. `mteb/_create_dataloaders.py` — four function-local imports plus one `TYPE_CHECKING` entry.
2. `mteb/similarity_functions.py` — ten function-local import torch. Also dtype: torch.dtype = torch.float32 → dtype: torch.dtype | None = None, resolved in the body: a default argument is evaluated at import time, so it kept torch loaded regardless of where the import sat.
3. `mteb/_evaluators/` — two files `(bitext_mining_evaluator.py,` `summarization_evaluator.py)` move import torch into the method that uses it.
4. imagetext_pairclassification_evaluator.py needed real restructuring: `class CustomImageDataset(torch.utils.data.Dataset[...])` executes at import. It's now built inside `_build_image_dataset(),` the only place it was used.

Results: 
 module | Before | After
-- | -- | --
mteb._create_dataloaders | 2266 modules, torch | 1537 modules, no torch
mteb._evaluators | could not be imported standalone | 2461 modules, no torch

mteb.abstasks and import mteb are unchanged (6799 modules, torch loaded). 

Next steps:
PR 4 — `mteb/models/` wrappers + `abstasks/image/image_text_pair_classification.py` (~11 files). Also folds in similarity_functions.py's two consumers (abs_encoder.py, random_baseline.py), since they live here. This will make `mteb.abstasks` and `mteb.benchmarks` go torch-free.

PR 5 — the model registry. `mteb.results.task_result` is still not clean, via `results/benchmark_results.py` → `models.get_model_meta` → `model_implementations` (217 modules, each importing a wrapper). That's the `ModelMeta.loader` work, and it's what finally moves import mteb itself.

Testing: 
Two new guards (9 total): mteb._create_dataloaders imports without torch, and _build_image_dataset still returns a real torch.utils.data.Dataset with unchanged len/items/features and DataLoader batching. Both confirmed to fail when an eager import is restored.

### DIFF
--- a/mteb/_evaluators/text/summarization_evaluator.py
+++ b/mteb/_evaluators/text/summarization_evaluator.py
@@ -106,9 +106,9 @@ class SummarizationEvaluator(Evaluator):
         encode_kwargs: EncodeKwargs,
         num_proc: int | None = None,
     ) -> SummarizationDistances:
-        # Get the human & machine summaries for the text in one go for all
         import torch
 
+        # Get the human & machine summaries for the text in one go for all
         human_lens = [len(human_summaries) for human_summaries in self.human_summaries]
         machine_lens = [
             len(machine_summaries) for machine_summaries in self.machine_summaries
--- a/mteb/_evaluators/text/summarization_evaluator.py
+++ b/mteb/_evaluators/text/summarization_evaluator.py
@@ -8,7 +8,9 @@ import numpy as np
 from scipy.stats import pearsonr, spearmanr
 from tqdm.auto import tqdm
 
+from mteb._create_dataloaders import _create_dataloader_from_texts
 from mteb._evaluators.evaluator import Evaluator
+from mteb.similarity_functions import cos_sim, dot_score
 
 if TYPE_CHECKING:
     from mteb.abstasks.task_metadata import TaskMetadata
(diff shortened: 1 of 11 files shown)


ayush1298:
@Samoed can you check it?

mteb/_evaluators/text/summarization_evaluator.py
```diff
@@ -108,6 +105,11 @@ def __call__(  # noqa: PLR0914
         num_proc: int | None = None,
     ) -> SummarizationDistances:
         # Get the human & machine summaries for the text in one go for all
+        import torch
+
+        from mteb._create_dataloaders import _create_dataloader_from_texts
+        from mteb.similarity_functions import cos_sim, dot_score
```
Samoed: I don't think that mteb imports should be moved

ayush1298: made change. actually it was mixed because `_create_dataloader.py` had changes of torch but `similarity_functions.py` doesnt. To remove these I need to move torch in each function in `similarity_functions.py`

Samoed (approved):

PR #5427, opened 2026-09-09
audio/video hashing + shared cross-task embedding cache

Rakshitha-Ireddi:
## Summary
  
  - **`_hash_item` now supports all modalities**: audio (numpy array bytes + sampling rate) and video (first-frame fingerprint with metadata fallback). Previously raised `TypeError` for video/audio items, making
  `CachedEmbeddingWrapper` unusable with those tasks.
  - **`shared_cache=True` flag on `CachedEmbeddingWrapper`**: all tasks write to a single `_shared/` directory keyed only by content hash. The same video/audio corpus shared by N tasks is encoded once instead of N
  times.
  - **5 new tests**: `TestHashItem` (audio hashing, content diff, unsupported raises) and `TestSharedCache` (cross-task hit, flat directory layout).
  
  ## Motivation
  
  Part of MOEB Track 3 (efficiency — faster evaluation infrastructure). Video/audio encoding is the dominant runtime cost. When multiple tasks evaluate the same corpus (e.g. MSR-VTT appears in retrieval, classification, and clustering tasks), `shared_cache=True` gives up to N× speedup on encoding with zero change to scores.
  
  ## Test plan
  - [x] `pytest tests/test_models/test_cached_embedding_wrapper.py` — 8 passed, 1 skipped

### DIFF
--- a/mteb/models/cache_wrappers/cache_backend_protocol.py
+++ b/mteb/models/cache_wrappers/cache_backend_protocol.py
@@ -10,12 +10,12 @@ if TYPE_CHECKING:
 
 @runtime_checkable
 class CacheBackendProtocol(Protocol):
-    """Protocol for a vector cache map (used to store text/image embeddings).
+    """Protocol for a vector cache map (used to store embeddings for any modality).
 
     Implementations may back the cache with different storage backends.
 
-    The cache maps an input item (text or image) to its vector embedding,
-    identified by a deterministic hash.
+    The cache maps an input item (text, image, audio, or video) to its vector
+    embedding, identified by a deterministic content hash.
     """
 
     def __init__(self, directory: Path | None = None, **kwargs: Any) -> None:
@@ -30,7 +30,7 @@ class CacheBackendProtocol(Protocol):
         """Add a vector to the cache.
 
         Args:
-            item: Input item containing 'text' or 'image'.
+            item: Input item containing 'text', 'image', 'audio', or 'video'.
             vectors: Embedding vector of shape (dim,) or (1, dim).
         """
 
--- a/mteb/models/cache_wrappers/cache_backends/_hash_utils.py
+++ b/mteb/models/cache_wrappers/cache_backends/_hash_utils.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Any
 if TYPE_CHECKING:
     from collections.abc import Mapping
 
+    import numpy as np
     from PIL import Image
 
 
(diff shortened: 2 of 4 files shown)


Samoed (changes requested):
There are too much unlrelated changes

Rakshitha-Ireddi:
> There are too much unlrelated changes

Thanks for the review @Samoed  Sorry about the noise — the other 3 commits (TAU/VGGSound datasets, version_reason, ModelMeta overrides) are from separate tracks and shouldn't have been in here.

  Should I:
  1. Create a clean new branch with only the cache commit and open a fresh PR, or
  2. Rebase/reset this branch to drop the unrelated commits?

  Happy to do whichever you prefer!

Samoed:
You can reset this branch

Rakshitha-Ireddi:
Done @Samoed  - done with resetting the branch to contain only the cache commit (ec9db0e). Now 1 commit, 4 files changed (+153 / −10). The unrelated changes were removed.

ayush1298:
A few issues:

1. `shared_cache=True` silently returns wrong embeddings because`_hash_item` keys on content only, so dropping the per-task/per-prompt directories removes the isolation. The same text encoded as `query` and as `document` now returns the same embeddings. Same root cause makes cross-task sharing unsound for instruction models, since `encode` gets `task_metadata` and the prompt is built from it. I guess we can include `prompt_type` (and the task prompt) in the cache key.
2. Video hash collides on the first frame. Two mp4s identical only in frame 0 hash the same, so the second silently gets the first's embedding. Hash the encoded file bytes, or several frames plus duration/resolution.
3. MockAudio/video tasks still fail.
4. The fallback can raise from inside the `except` (`_hash_utils.py:40-44`). If `get_frames_at` fails, the handler calls `video.metadata`, which throws for anything that isn't a decoder. `num_frames: duration` also collides for any two clips of equal length.

@Samoed can you also check this one?

Samoed:
Yes, I agree. I'm rewriting this pr. I also run a small benchmark for hash functions and I descided to switch from md5 to sha256

```
text_short (200 B)
  sha1             0.3 us/call      760.1 MB/s *
  sha256           0.3 us/call      637.5 MB/s
  blake2b          0.3 us/call      587.5 MB/s
  blake2s          0.4 us/call      493.6 MB/s
  md5              0.5 us/call      442.8 MB/s

image_frame_224 (147 KB)
  sha1            45.2 us/call     3331.3 MB/s *
  sha256          48.1 us/call     3126.3 MB/s
  blake2b        103.4 us/call     1455.3 MB/s
  md5            160.6 us/call      937.1 MB/s
  blake2s        174.5 us/call      862.4 MB/s

audio_1s_16k (62.5 KB)
  sha1            19.2 us/call     3325.4 MB/s *
  sha256          19.6 us/call     3272.1 MB/s
  blake2b         43.9 us/call     1457.7 MB/s
  md5             68.4 us/call      935.4 MB/s
  blake2s         73.2 us/call      874.3 MB/s

audio_10s_16k (625 KB)
  sha1           190.7 us/call     3356.1 MB/s *
  sha256         193.9 us/call     3301.0 MB/s
  blake2b        438.8 us/call     1458.6 MB/s
  md5            686.9 us/call      931.8 MB/s
  blake2s        757.4 us/call      845.0 MB/s

video_frame_720p (2.6 MB)
  sha1           820.5 us/call     3369.5 MB/s *
  sha256         823.1 us/call     3358.9 MB/s
  blake2b       1891.4 us/call     1461.8 MB/s
  md5           2956.6 us/call      935.1 MB/s
  blake2s       3223.5 us/call      857.7 MB/s
```

PR #5434, opened 2026-09-09
[MOEB] task: add YouCook2 I2V/V2I Retrieval

RishabhPatil:
## YouCook2 image↔video retrieval

Adds two MOEB cross-modal retrieval tasks:

- `YouCook2I2VRetrieval`: plated-dish image → cooking tutorial video
- `YouCook2V2IRetrieval`: cooking tutorial video → plated-dish image

The tasks add instructional cooking image↔video retrieval. For each recipe, one held-out video provides the query and 2–3 demonstrations by different cooks form the corpus; the query video is never a corpus item.

Datasets: [I2V](https://huggingface.co/datasets/iamfortytwo/YouCook2-I2V) / [V2I](https://huggingface.co/datasets/iamfortytwo/YouCook2-V2I)

Descriptive statistics:

- 84 queries, 240 corpus videos/images, 240 qrels
- 2–3 relevant corpus items per query
- No query/corpus video overlap

Reference runs (nDCG@10):

| Model | I2V | V2I |
| --- | ---: | ---: |
| `mteb/baseline-random-encoder` | 0.02512 | 0.02369 |
| `jinaai/jina-embeddings-v5-omni-nano` | 0.04809 | 0.06971 |

The reference model exceeds chance while remaining far from saturation. The task also restores decoding for published video columns before model evaluation.

Validation:

- [x] `test_metadata.py` passes for both tasks
- [x] `test_task_quality.py` passes
- [x] Dataset runs with MTEB and both reference encoders
- [x] No paper retrieval result to reproduce: this is a novel plated-dish-image↔tutorial-video construction; YouCook2’s source paper evaluates dense video captioning.

Reproducibility: `python scripts/run_youcook2_retrieval.py`

PR #5435, opened 2026-09-09
[MOEB] task: add MetaWorld MT50 I2V/V2I Retrieval

RishabhPatil:
## MetaWorld MT50 image↔video retrieval

Adds two MOEB cross-modal retrieval tasks:

- `MetaWorldMT50I2VRetrieval`: goal-state image → robot-manipulation video
- `MetaWorldMT50V2IRetrieval`: robot-manipulation video → goal-state image

The tasks fill an image↔video retrieval gap for robotics. They use 49 MetaWorld MT50 tabletop-manipulation tasks, with held-out episodes as queries and different episodes of the same task as relevant corpus items.

Datasets: [I2V](https://huggingface.co/datasets/iamfortytwo/MetaWorld-MT50-I2V) / [V2I](https://huggingface.co/datasets/iamfortytwo/MetaWorld-MT50-V2I)

Descriptive statistics:

- 245 queries, 490 corpus videos, 2,450 qrels
- 10 relevant corpus items per query
- No query/corpus episode overlap

Reference runs (nDCG@10):

| Model | I2V | V2I |
| --- | ---: | ---: |
| `mteb/baseline-random-encoder` | 0.02397 | 0.01877 |
| `jinaai/jina-embeddings-v5-omni-nano` | 0.23006 | 0.29589 |

The random baseline is near chance; the multimodal reference model is substantially higher without saturating, indicating a discriminative task.

Validation:

- [x] `test_metadata.py` passes for both tasks
- [x] Dataset runs with MTEB and both reference encoders
- [x] No paper retrieval result to reproduce: this is a novel goal-image↔demonstration-video construction; MetaWorld’s source paper evaluates RL.

Reproducibility: `python scripts/run_metaworld_mt50_retrieval.py`

### DIFF
--- a/mteb/descriptive_stats/Image/Any2AnyRetrieval/MetaWorldMT50I2VRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyRetrieval/MetaWorldMT50I2VRetrieval.json
@@ -1,0 +1,64 @@
+{
+    "test": {
+        "num_samples": 735,
+        "num_queries": 245,
+        "num_documents": 490,
+        "number_of_characters": 0,
+        "documents_text_statistics": null,
+        "documents_image_statistics": null,
+        "documents_audio_statistics": null,
+        "documents_video_statistics": {
+            "total_duration_seconds": 3948.6,
+            "total_frames": 39486,
+            "min_width": 256,
+            "average_width": 256.0,
+            "max_width": 256,
+            "min_height": 256,
+            "average_height": 256.0,
+            "max_height": 256,
+            "min_duration_seconds": 1.7,
+            "average_duration_seconds": 8.058367346938775,
+            "max_duration_seconds": 50.0,
+            "unique_videos": 490,
+            "average_fps": 10.0,
+            "fps": {
+                "10": 490
+            },
+            "min_resolution": [
+                256,
+                256
+            ],
+            "average_resolution": [
+                256.0,
+                256.0
+            ],
+            "max_resolution": [
+                256,
+                256
+            ],
+            "resolutions": {
+                "256x256": 490
+            }
+        },
+        "queries_text_statistics": null,
+        "queries_image_statistics": {
+            "min_image_width": 256,
+            "average_image_width": 256.0,
+            "max_image_width": 256,
+            "min_image_height": 256,
+            "average_image_height": 256.0,
+            "max_image_height": 256,
+            "unique_images": 245
+        },
+        "queries_audio_statistics": null,
+        "queries_video_statistics": null,
+        "relevant_docs_statistics": {
+            "num_relevant_docs": 2450,
+            "min_relevant_docs_per_query": 10,
+            "average_relevant_docs_per_query": 10.0,
+            "max_relevant_docs_per_query": 10,
+            "unique_relevant_docs": 490
+        },
+        "top_ranked_statistics": null
+    }
+}
--- a/mteb/descriptive_stats/Image/Any2AnyRetrieval/MetaWorldMT50V2IRetrieval.json
+++ b/mteb/descriptive_stats/Image/Any2AnyRetrieval/MetaWorldMT50V2IRetrieval.json
@@ -1,0 +1,64 @@
+{
+    "test": {
+        "num_samples": 735,
+        "num_queries": 245,
+        "num_documents": 490,
+        "number_of_characters": 0,
+        "documents_text_statistics": null,
+        "documents_image_statistics": {
+            "min_image_width": 256,
+            "average_image_width": 256.0,
+            "max_image_width": 256,
+            "min_image_height": 256,
+            "average_image_height": 256.0,
+            "max_image_height": 256,
+            "unique_images": 490
+        },
+        "documents_audio_statistics": null,
+        "documents_video_statistics": null,
+        "queries_text_statistics": null,
+        "queries_image_statistics": null,
+        "queries_audio_statistics": null,
+        "queries_video_statistics": {
+            "total_duration_seconds": 2007.0,
+            "total_frames": 20070,
+            "min_width": 256,
+            "average_width": 256.0,
+            "max_width": 256,
+            "min_height": 256,
+            "average_height": 256.0,
+            "max_height": 256,
+            "min_duration_seconds": 2.6,
+            "average_duration_seconds": 8.191836734693878,
+            "max_duration_seconds": 50.0,
+            "unique_videos": 245,
+            "average_fps": 10.0,
+            "fps": {
+                "10": 245
+            },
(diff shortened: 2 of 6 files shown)


RishabhPatil:
@AdnanElAssadi @Samoed Any thoughts about this one? I'm working on evaluting with better models. In the meantime, please let me know if you expect any other issues.

RishabhPatil:
Added reproducibility results for both retrieval datasets using Qwen3-VL-Embedding-2B and 8B (`mteb.get_model`, CUDA, batch size 1).

| Dataset | Model | I2V nDCG@10 | V2I nDCG@10 |
|---|---:|---:|---:|
| YouCook2 | Qwen3-VL-Embedding-2B | 0.16995 | 0.15872 |
| YouCook2 | Qwen3-VL-Embedding-8B | 0.18362 | 0.19336 |
| MetaWorld MT50 | Qwen3-VL-Embedding-2B | 0.37400 | 0.38459 |
| MetaWorld MT50 | Qwen3-VL-Embedding-8B | 0.40237 | 0.42014 |

The JSON files are available here:

- [Qwen3-VL-Embedding-2B results](https://github.com/RishabhPatil/results/tree/codex/youcook2-metaworld-qwen-results/results/Qwen__Qwen3-VL-Embedding-2B/9f2f7e710d6d81056aa5c0a4f04764fec6bb7bda)
- [Qwen3-VL-Embedding-8B results](https://github.com/RishabhPatil/results/tree/codex/youcook2-metaworld-qwen-results/results/Qwen__Qwen3-VL-Embedding-8B/2c4565515e0f265c6511776e7193b22c0968ddc7)

Samoed (changes requested):
Same as https://github.com/embeddings-benchmark/mteb/pull/5434 I don't think that matching images with videos from same sources is good task even if frames are not matching

RishabhPatil:
@Samoed Just to clarify, this is not matching an image to the video it was extracted from.

For each MetaWorld task. The I2V query is the final-frame image from a query episode, while the candidate videos come only from the 10 different corpus episodes excluding the query episode. Likewise, V2I queries use held-out videos and retrieve images from different corpus episodes.

For a model to succeed at this task it would be expected to understand the context and truly learn the domain and not simply match a frame between an image and video.

I followed the same logic for #5434 as well.

Totally okay if you still think this is a poor quality task to add, I can close both the PRs

isaac-chung:
> @Samoed Just to clarify, this is not matching an image to the video it was extracted from.
> 
> For each MetaWorld task. The I2V query is the final-frame image from a query episode, while the candidate videos come only from the 10 different corpus episodes excluding the query episode. Likewise, V2I queries use held-out videos and retrieve images from different corpus episodes.
> 
> For a model to succeed at this task it would be expected to understand the context and truly learn the domain and not simply match a frame between an image and video.
> 
> I followed the same logic for #5434 as well.
> 
> Totally okay if you still think this is a poor quality task to add, I can close both the PRs

From this description, I think this is a fine task to include. Maybe @AdnanElAssadi can help with the tie break here.

Samoed (changes requested):
> while the candidate videos come only from the 10 different corpus episodes excluding the query episode

I don't think that this is true, because there is no `top_ranked` split in dataset.

> For a model to succeed at this task it would be expected to understand the context and truly learn the domain and not simply match a frame between an image and video.

How are your prevent this? In qrels you're matching image and video from same source video

RishabhPatil:
> > while the candidate videos come only from the 10 different corpus episodes excluding the query episode
> 
> I don't think that this is true, because there is no `top_ranked` split in dataset.

I may be misunderstanding the top_ranked concern. I followed the current standard retrieval layout of queries, corpus, and qrels; the loader only reads top_ranked when that optional config is present, as it is used for reranking. If you expect a top_ranked config for this task, could you clarify its intended contents? I’m happy to revise the dataset format accordingly.


> > For a model to succeed at this task it would be expected to understand the context and truly learn the domain and not simply match a frame between an image and video.
> 
> How are your prevent this? In qrels you're matching image and video from same source video

Just to clarify the leakage point: the qrels aren’t matching an image to the video from the same episode.
For each task, I use 5 episodes for queries and a separate 10 episodes for the corpus ([split logic](https://github.com/RishabhPatil/mteb/blob/feat/metaworld-retrieval/scripts/data/metaworld_retrieval/create_data.py#L149-L155)). The qrels are then only created between q-ep... query IDs and c-ep... corpus IDs ([qrel construction](https://github.com/RishabhPatil/mteb/blob/feat/metaworld-retrieval/scripts/data/metaworld_retrieval/create_data.py#L233-L256)).
For example, q-ep000350 is paired with c-ep000355 through c-ep000364, not its own episode ([viewer](https://huggingface.co/datasets/iamfortytwo/MetaWorld-MT50-I2V/viewer/qrels/test?row=0)). I also checked the full dataset: 245 query episodes, 490 corpus episodes, and no episode overlap.
There can definitely be task/scene cues, but it is cross-episode retrieval rather than matching a frame back to its source video.

Samoed:
On `top_ranked`. It's a correction of the earlier claim that `candidate videos come only from the 10 different corpus episodes excluding the query episode.` `top_ranked` is the *only* mechanism MTEB has for restricting candidates per query, and it's optional/reranking-only. So every query is scored against all 490 corpus items across all 49 tasks, not a 10-item shortlist. 

> There can definitely be task/scene cues, but it is cross-episode retrieval rather than matching a frame back to its source video.

I think this claim actually contradicts the task as built, not just the leakage argument. The query is a single frame — literally the last frame of an episode, saved as a PNG. The corpus document is a video, i.e. a sequence of frames from the same rendering setup. So "match a frame between an image and video" isn't something the task avoids — it's a description of exactly what I2V does by construction. The episode-level split means the specific frame isn't duplicated verbatim in the corpus, but the task is still, structurally, frame-to-frame appearance matching, not action or language understanding.

And for MetaWorld specifically there's a concrete, simple strategy that would work here without any real "domain understanding": each task has a fixed goal region (button location, peg-hole position, drawer/door position, etc.), so the object/end-effector ends up in roughly the same place in the frame for every successful episode of that task. A model — or honestly a non-learned heuristic — only needs to localize where the relevant object/target is in the query's goal frame, then find the video whose frames bring that object to the same location. That solves the retrieval without tracking the manipulation trajectory, understanding the action being performed, or aligning image and video semantically in any deeper sense — it's closer to spatial/appearance matching than to what "truly learn the domain" is meant to rule out

RishabhPatil:
@Samoed Okay, that makes sense to me. The concern is more with the way the image is selected from the videos and how that specific objective can be easily hacked by agents, hence the agents may not end up looking for other hints or cues.

I think its safe to close this.

PR #5441, opened 2026-09-10
fix: declare ModelMeta load dtypes with OutputDType instead of torch.dtype

ayush1298:
Refs #5063. Follows #5329, #5350, #5426, #5441.

Model metadata currently needs torch imported just to define it like `ModelMeta(loader_kwargs=dict(torch_dtype=torch.float16))` 
which evaluates at module scope, so building `MODEL_REGISTRY` pulls in torch. This PR changed it and names the `dtype` with an `OutputDType` member instead of holding a `torch.dtype` object.

Changes:

1. `OutputDType:` added FLOAT32 (the enum had no member) and update docstring as it covers both output-embedding dtypes and model load dtypes.
2. Changes in 85 import-time sites across 30 model files: `torch.float16 → OutputDType.FLOAT16`.
3. 8 wrapper `__init__` whose default changed now resolve at the boundary like:
```
if isinstance(torch_dtype, OutputDType):
        torch_dtype = torch_dtype.get_dtype()
```

`OutputDType` is a str enum, so `getattr(torch, member)` works and transformers resolves it natively.
4. Docs: documented the convention in `adding_a_model.md` (and added the OutputDType import its existing example was missing); corrected the compression doc, which described the enum as only compression levels.

Follow-up PR: Files under model_implementations/ needing restructuring: 94 → 68. No dtype sites remain.
1. 50 @torch.no_grad() / @torch.inference_mode() decorators
2. 48 device: str = "cuda" if torch.cuda.is_available() else "cpu" defaults
3. 4 module- or class-scope `torch.nn.* / torch.Tensor` references and the 161 module-level `import torch / transformers `statements which must move to make is genuinely torch-free

Testing: 
2 new tests: one checks a ModelMeta dtype resolves through both `get_dtype()` and `getattr(torch, ...);` the other is an AST check that no model file evaluates a `torch.<dtype>` in a signature default, so the convention doesn't rot as new models land. Confirmed red when a torch.bfloat16 default is restored.

### DIFF
--- a/mteb/models/model_implementations/colpali_models.py
+++ b/mteb/models/model_implementations/colpali_models.py
@@ -8,6 +8,7 @@ from tqdm.auto import tqdm
 
 from mteb.models.abs_encoder import AbsEncoder
 from mteb.models.model_meta import ModelMeta, ScoringFunction
+from mteb.types import OutputDType
 
 if TYPE_CHECKING:
     from PIL import Image
@@ -205,7 +206,7 @@ COLPALI_TRAINING_DATA = {
 colpali_v1_1 = ModelMeta(
     loader=ColPaliWrapper,
     loader_kwargs=dict(
-        torch_dtype=torch.float16,
+        torch_dtype=OutputDType.FLOAT16,
     ),
     name="vidore/colpali-v1.1",
     model_type=["late-interaction"],
@@ -234,7 +235,7 @@ colpali_v1_1 = ModelMeta(
 colpali_v1_2 = ModelMeta(
     loader=ColPaliWrapper,
     loader_kwargs=dict(
-        torch_dtype=torch.float16,
+        torch_dtype=OutputDType.FLOAT16,
     ),
     name="vidore/colpali-v1.2",
     model_type=["late-interaction"],
@@ -263,7 +264,7 @@ colpali_v1_2 = ModelMeta(
 colpali_v1_3 = ModelMeta(
     loader=ColPaliWrapper,
     loader_kwargs=dict(
-        torch_dtype=torch.float16,
+        torch_dtype=OutputDType.FLOAT16,
     ),
     name="vidore/colpali-v1.3",
     model_type=["late-interaction"],
--- a/tests/test_ensure_no_torch_at_import.py
+++ b/tests/test_ensure_no_torch_at_import.py
@@ -233,32 +233,40 @@ def test_model_meta_dtypes_are_named_not_torch_objects() -> None:
 
 
 def test_model_implementations_declare_no_import_time_torch_dtypes() -> None:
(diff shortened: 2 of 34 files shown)


Samoed (approved):

tests/test_ensure_no_torch_at_import.py
```diff
@@ -211,3 +211,54 @@ def test_importing_benchmarks_does_not_import_torch() -> None:
...
     )
+
+
+def test_model_meta_dtypes_are_named_not_torch_objects() -> None:
+    """`ModelMeta` declares load dtypes by name, so metadata stays importable without torch.
+
+    `OutputDType` is a `str` enum, so it resolves through both `getattr(torch, ...)` (the path
+    transformers takes for a string dtype) and `OutputDType.get_dtype()`.
+    """
+    import torch
```
isaac-chung, line 222: What happens to this test when we eventually have the `mteb-core` lib?

isaac-chung (commented):
Can you address the copilot comments as well please?

tests/test_ensure_no_torch_at_import.py
ayush1298: Most of the tests in this whole file will be simplified or removed, as we no longer need these module-level tests. We can directly check with whole package

ayush1298:
Address copilot comments

mteb/models/model_implementations/colpali_models.py
Samoed, line 1: @ayush1298 Can you check will colapli models would correctly load with updated dtype? They will pass our enum, which I'm not sure will work correctly

ayush1298: @Samoed I checked with `vidore/colSmol-256M` and it's loading perfectly from main. Also, I checked the dtype and weights before and after this change, and both are identical.

Samoed: Great! Not sure how this works, but good

ayush1298: The main reason was the wrappers never touch the `dtype` So, `ColPaliWrapper` and others just pass `torch_dtype` through `**kwargs` into `model_class.from_pretrained(...).` `OutputDType` is a str enum, and transformers already accepts string dtypes (`dtype = getattr(torch, dtype) if isinstance(dtype, str)`. So `OutputDType.FLOAT16` resolves to `torch.float16` exactly as passing `torch.float16` did before.

Samoed: Ah, I see. Thanks. Forgot that they support str

isaac-chung:
CI windows python 3.10 failing: https://github.com/embeddings-benchmark/mteb/actions/runs/34755656152/job/103719465634

PR #5449, opened 2026-09-12
fix: make raise_error=False cover the whole per-task lifecycle

### LINKED ISSUE #2345: `raise_on_error = False` does not
`raise_on_error = False` in mteb.MTEB does not cover all cases. E.g. if a dataset fails to load an error is still raised.

AnonymousAgarwal:
Fixes #2345.

`raise_error=False` only guarded the `_evaluate_task` call, so the rest of the per-task work still aborted the whole run: the modality check (done once for the whole task list), `_check_cache` under `never`/`only-cache`, `model.load_model()`, `cache.save_to_cache()`, and the aggregate combine/merge/save. These now become `TaskError`s in `ModelResult.exceptions` like any other task failure.

A failed cache write no longer discards the scores of a task that ran to completion. `raise_error=True` is unchanged.

Two patchsets:

1. **`fix:`** closes the leaks by widening the guard in place, and adds six tests in `tests/test_evaluate.py`.
2. **`refactor:`** moves the per-task work into an always-raising `_run_task` (aggregates via `_run_aggregate_task`) and leaves a single guard in the task loop, so one task and a list of tasks take the same path and the recursion for plain lists is gone. No behaviour change over patchset 1; it stops the guard drifting out of sync with the task lifecycle again.

Happy to drop the second patchset if you would rather keep the fix minimal.


AnonymousAgarwal:
Sorry about that — description trimmed to well under the limit.

AnonymousAgarwal:
Pushed a second patchset (738c82e) with the refactor described in the updated description: the per-task work moves into an always-raising `_run_task` and the task loop keeps the only guard, so the fix in the first patchset cannot drift out of sync with the task lifecycle again. No behaviour change between the two.

Happy to drop it and keep this PR to the fix alone if you prefer.

Muennighoff (commented):
thats a lot of code changes and a lot of tests for a very simple change - is it really all needed?

Samoed (changes requested):
Can you write why so much change is needed?

AnonymousAgarwal:
Pushed a much smaller version: 

Why it isn't a one-liner — the guard only wrapped `_evaluate_task`, but four other per-task steps sat outside it and still aborted the whole run:

- the modality check, run once over the entire task list before anything starts
- `_check_cache`, which raises under `never` / `only-cache`
- `cache.save_to_cache`
- the aggregate combine / merge / save — a separate branch the single-task guard never reaches

The code change mostly wraps the try catch across all these sub tasks.

The second patchset (the larger refactor you've mentioned) was there to move the per-task work into an always-raising `_run_task`, instead of adding try catch across individual blocks.
That big of a refactor is not necessary here.

One main change is - a failing subtask now fails its aggregate, instead of returning and caching a result with `main_score: None`. That null result makes `_get_score_fast` raise later. Happy to revert it or split it out.

The one test added fails on `main` and passes here.

Samoed:
I added more guards to `evaluate`

KennethEnevoldsen:
Hmm, I am not sure this really addresses the parts that I want it to. This is a very sensitive part of the code that everyone moves through. It def. shouldn't be a TaskError that is saved in all cases (e.g. when cache write fails).

In general, I don't think this is the best space for first-time contributors to start, though I appreciate the PR.

@Samoed, I see that you have made some changes. Is it okay if I close this PR?

Samoed:
Yeah, I agree.

PR #5452, opened 2026-09-12
move Nano retrieval tasks to the v2 dataset format

aradhyagnov:
Part of #3424 — takes the 13 `Nano*Retrieval` tasks.

All 13 carried an identical v1 `load_data()` override reading the `zeta-alpha-ai` copies
into `self.corpus` / `self.queries` / `self.relevant_docs`. The re-uploads under the
`mteb` org are already in the v2 layout (`corpus/`, `queries/`, `qrels/`, `train` split),
so the override goes and the default `AbsTaskRetrieval` loader handles them. -702/+26.

## Verification

For each task I loaded the original through the v1 logic and the migrated task through
mteb's API, then compared corpus ids and text, query ids and text, and qrels.
**13/13 identical.**

Worth flagging: the v1 override hardcoded every relevance score to `1`. I checked the
uploaded qrels for Touche2020, DBPedia and ArguAna — all scores are `1` there too, so
this is score-neutral rather than silently regrading.

(`NanoTouche2020Retrieval` has 49 queries rather than 50. That is true of the original
too, not introduced here.)

Lint clean. `test_validate_metadata`, `test_get_tasks`, `test_filter_tasks` and
`test_benchmarks` pass (1602).

I used `refactor:` rather than `dataset:` since no content changes — happy to switch if
you'd rather this triggered a release.

### DIFF
--- a/mteb/tasks/retrieval/eng/nano_argu_ana_retrieval.py
+++ b/mteb/tasks/retrieval/eng/nano_argu_ana_retrieval.py
@@ -1,8 +1,3 @@
-from collections import defaultdict
-from typing import Any
-
-from datasets import load_dataset
-
 from mteb.abstasks.retrieval import AbsTaskRetrieval
 from mteb.abstasks.task_metadata import TaskMetadata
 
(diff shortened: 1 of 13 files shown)


Samoed (approved):

Samoed:
I run `sentence-transformers/all-MiniLM-L6-v2` and got same results

PR #5453, opened 2026-09-12
Create aurola_omni_models

deep9539:
(Thank you @nehal309 for the help, raising PR after testing it)

| Task / Direction | Metric | Paper (arXiv:2602.18010[1]) | MTEB Run | Difference |
| :--- | :--- | :---: | :---: | :---: |
| AudioCaps T2A (Text → Audio) | R@10 (hit_rate_at_10) | 88.3% | 88.01% (0.88007) | -0.29% |
| AudioCaps A2T (Audio → Text) | R@10 (hit_rate_at_10) | 92.9% | 92.53% (0.92525) | -0.37% |

---------------

We also discussed and ran it for Clotho, but Clotho task in MTEB is very different from the actual one hence the numbers are not comparable.

| Task / Direction | Metric | Paper (arXiv:2602.18010[1])* | MTEB Run (Clotho.v2) | Difference (Expected) |
| :--- | :--- | :---: | :---: | :---: |
| Clotho T2A (Text → Audio) | R@10 (hit_rate_at_10) | 64.7% | 80.04% (0.80043) | +15.34% |
| Clotho A2T (Audio → Text) | R@10 (hit_rate_at_10) | 73.8% | 77.99% (0.77990) | +4.19% |

@AdnanElAssadi @Samoed This one looks important and leading the MMEBv3 Audio tasks.


mteb/models/model_implementations/aurola_omni_models.py
```diff
@@ -0,0 +1,161 @@
...
+            if (
+                Qwen2_5OmniThinkerConfig not in MODEL_FOR_MULTIMODAL_LM_MAPPING
+                and hasattr(AutoModelForMultimodalLM, "register")
+            ):
+                AutoModelForMultimodalLM.register(
+                    Qwen2_5OmniThinkerConfig,
+                    Qwen2_5OmniThinkerForConditionalGeneration,
+                )
+        except (ImportError, AttributeError):
+            pass
```
Samoed: Why do you need this?

mteb/models/model_implementations/aurola_omni_models.py
```diff
@@ -0,0 +1,161 @@
...
+                0
+            ].processor.feature_extractor.sampling_rate
+        if hasattr(self.model[0], "processing_kwargs"):
+            self.model[0].processing_kwargs.setdefault("video", {}).update(
+                {
+                    "max_pixels": 64 * 28 * 28,
+                    "do_sample_frames": False,
+                    "fps": 1,
+                }
+            )
```
Samoed: You can pass this to processor kwargs

mteb/models/model_implementations/aurola_omni_models.py
deep9539: Yeah, not needed, removed.

mteb/models/model_implementations/aurola_omni_models.py
deep9539: only max_pixels kept, other removed

AdnanElAssadi: are you sure `do_sample_frames: False` not needed?

deep9539: Hey, this code is outdated, it's not there in latest snapshot.

The answer is no.

AdnanElAssadi: i know its outdated, im asking to see if you know whats default so that no double sampling happens.

deep9539: Ohh, good question.

https://huggingface.co/Jazzcharles/AuroLA-Omni-7B/blob/main/video_preprocessor_config.json 

Here, the default value for do_sample_frames is False and fps is null, so we should be good.

ayush1298:
license is not specified for these models, what should we do? @Samoed

Samoed:
I don't think missing license is problem for models

Samoed:
> but Clotho task in MTEB is very different from the actual one hence the numbers are not comparable.

What do you mean by `actual` task?

deep9539:
> > but Clotho task in MTEB is very different from the actual one hence the numbers are not comparable.
> 
> What do you mean by `actual` task?

The paper evaluates the official benchmark of 5,225 intact human captions without any text splitting, reporting 64.7% R@10. In MTEB v1, naively splitting concatenated captions on periods generated 905 empty-string queries that scored 0 automatically, artificially dragging the score down to ~67% and creating a false illusion of matching the paper. In MTEB Clotho.v2, those 905 empty queries were filtered out to 4,680 clean segments, removing the penalty and naturally raising the score to 80.04%. Neither MTEB version matches the paper's original 5,225 queries.
  
On AudioCaps (which has no splitting artifacts), our MTEB run reproduces the paper almost identically (88.01% vs 88.3%)

@nehal309 Please feel free to correct me.

Samoed (approved):

nehal309:
Thanks Roman, should we merge this to main?

PR #5462, opened 2026-09-14
Add MUSIC-AVQA audio-to-video and video-to-audio retrieval tasks

RishabhPatil:
## Summary

Adds two cross-modal retrieval tasks derived from the MUSIC-AVQA test split:

- `MusicAVQAA2VRetrieval`: audio query → retrieve videos featuring the same instrument
- `MusicAVQAV2ARetrieval`: video query → retrieve audio featuring the same instrument

The construction uses all 22 instrument classes with a fixed seed:

- 110 queries (5 clips per class)
- 220 corpus items (10 different clips per class)
- 1,100 qrels per direction

Query and corpus source `video_id`s are unique and strictly disjoint. This is therefore cross-clip, class-level instrument retrieval—not matching a modality to the synchronous counterpart from the same clip.

## Included

- Task metadata and registration
- Reproducible data-construction script
- MTEB runner using `mteb.get_model`
- Source/license provenance in the generated dataset cards
- Local validation for both directions and Ruff checks

Note: Currently its a retrieval task, so a guitar audio query should retrieve a different guitar-performance video even when its soundtrack is different. Another option would be to have an AV sync task, where we expect retrieval of the exact same video, which would be a much harder task.


mteb/tasks/retrieval/eng/music_avqa_retrieval.py
```diff
@@ -0,0 +1,84 @@
...
+
+_REFERENCE = "https://arxiv.org/abs/2203.14072"
+_BIBTEX = r"""
+@article{li2022musicavqa,
+  author = {Li, Bo and others},
+  title = {Music Audio-Visual Question Answering},
+  year = {2022},
+  url = {https://arxiv.org/abs/2203.14072},
+}
+"""
```
Samoed, line 16: Other music qa tasks cite different paper https://github.com/embeddings-benchmark/mteb/blob/de206dc17391d5bc67aaeb7db840dfc05fa6c31a/mteb/tasks/clustering/eng/music_avqa_clustering.py#L8-L15

Samoed (changes requested):
I don't think we need to convert classification task into retrieval

RishabhPatil:
> I don't think we need to convert classification task into retrieval

Should we add it as classification task then? do we already have a lot of those?

Samoed:
We already have https://github.com/embeddings-benchmark/mteb/blob/f9b5cf2947f87e1a8f961fa78a850b291e6ba83c/mteb/tasks/classification/eng/music_avqa_classification.py#L46

RishabhPatil:
Gotcha, closing this one

PR #5436, opened 2026-09-09
fix: make abstasks and model wrappers torch-free

ayush1298:
Fourth step toward making mteb importable without torch. Refs #5063. Follows #5329, #5350, #5426.

`AbsTask` and `Benchmark`: the classes carrying the metadata the leaderboard and API read no longer need torch to import.

Changes: 
10 files, each with a single eager import torch and no import-time execution:
- nine under `mteb/models/` — cache, compression, instruct, modality collator, openai, faiss index, search, sentence-transformer and vllm wrappers
- `mteb/abstasks/image/image_text_pair_classification.py,` the one file under `abstasks/` with its own import torch
- Each gets import torch in its `TYPE_CHECKING` block for annotations, plus function-local imports in the methods that use it at runtime.

No `__init__.py` is touched, and no mteb imports are moved. Also, `from mteb.models import EncoderProtocol` imports across `abstasks/` needed no change at all as once the wrappers stopped pulling torch, they became harmless.

Result: 

Module  | Before | After
-- | -- | --
mteb.abstasks | 3260 modules, torch | 2534 modules, no torch
mteb.benchmarks.benchmark | 3439 modules, torch | 2714 modules, no torch

import mteb is still unchanged (6801 modules, torch loaded). It reaches torch through `mteb/__init__.py` and the `model_implementations` registry, which is the remaining step.

Testing
Two new tests added on `mteb.abstasks` and `mteb.benchmarks.benchmark;` both confirmed to fail when an eager import is restored. ruff and mypy clean.

PR #5463, opened 2026-09-14
fix: make model implementations importable without torch

ayush1298:
Refs #5063. Follows #5329, #5350, #5426, #5436, #5441

Building `MODEL_REGISTRY` imports every model implementation file, and 126 of 250 imported `torch,` `transformers` or `sentence-transformers` at import time. After this PR, `import mteb` and `get_model_metas()` work on an install without `torch.` No `mteb imports` are moved and no `__init__.py` is touched. 

Changes:

1. New `mteb/_torch_utils.py` with three helpers that only import torch when called:
    - `no_grad / inference_mode` apply `torch.no_grad()` / `torch.inference_mode()` at call time. Since torch's own decorators do the work, behaviour carries over exactly, including generators. 55 decorators now use these helpers.
    - 58 device defaults of `"cuda" if torch.cuda.is_available() else "cpu"` became `device: str | None = None`, resolved with `device = get_device(device)`. This removes 50 function-local `import torch` statements.
2. 165 module-level imports moved into the functions that use them, or under `TYPE_CHECKING` for annotation-only uses.
3. 3 `nn.Module` subclasses `(gme_v_models,` `semantic_router_multimodal_models)` moved into cached factory functions, because class bases are evaluated at import.
4. `mixedbread` rerankers: `activation_fn=torch.nn.Identity()` became a plain `_identity` function. It returns the same object, and sentence-transformers only ever calls it.
5. Docs: new "Keeping your implementation importable without torch" section in `adding_a_model.md.`
6. Test fix: the `dtype` guard reads files as UTF-8, which fixes a Windows failure (cp1252 can't decode 14 model files).

Verification:

1. Without torch installed: all 250 model files import (124 before), get_model_metas() returns all 922 models, and get_tasks / get_benchmark work. Also, the model registry is identical (922 models, no metadata differences).
2. The decorators match torch's across 17 cases: plain functions, methods, generators, exceptions, nesting, threads, inference tensors and metadata.
3. Testing: runs `import mteb + get_model_metas()` in a subprocess with torch hidden from the import system, so it works in normal CI. It fails if any model file imports torch at module scope.

What's next:
1. version("mteb") / distribution("mteb") / requires("mteb") break in an `mteb-core-only` install and need fixing before the split.
2. The package split itself.
3. Optional: torch still loads when installed, because `cqadupstack_retrieval.py` creates tasks at import and `_set_seed` imports torch. This doesn't affect `mteb-core.`


ayush1298:
@Samoed For most of the files, changes are trivial. But, I am tagging 3 files with some non-trivial changes, can you have a look at them?

mteb/models/model_implementations/mixedbread_ai_models.py
ayush1298, line 1: this

mteb/models/model_implementations/gme_v_models.py
ayush1298, line 1: this

mteb/models/model_implementations/semantic_router_multimodal_models.py
ayush1298, line 1: this

mteb/_torch_utils.py
```diff
@@ -0,0 +1,59 @@
...
+    return wrapper
+
+
+def get_device(device: str | None = None) -> str:
+    """Return `device` if given, otherwise `"cuda"` when available and `"cpu"` if not."""
+    if device is not None:
+        return device
+    import torch
+
+    return "cuda" if torch.cuda.is_available() else "cpu"
```
Samoed, line 63: What about `mps`?

mteb/models/model_implementations/mixedbread_ai_models.py
```diff
@@ -7,6 +9,15 @@
     SentenceTransformerEncoderWrapper,
 )
 
+if TYPE_CHECKING:
+    import torch
+
+
+def _identity(logits: torch.Tensor) -> torch.Tensor:
+    """Keep the raw logit as the score, like ``torch.nn.Identity()`` but without importing torch."""
+    return logits
```
Samoed, line 18: Why not simply import torch inside function?

mteb/_torch_utils.py
ayush1298: `get_device` returns what the default it replaced returned and on main 58 converted defaults was:
` "cuda" if torch.cuda.is_available() else "cpu"`. None of them selected `mps.`

The 26 models that do use mps pick it in their own function bodies like:
`self.device = device or ("cuda" if torch.cuda.is_available() else "mps" if torch.backends.mps.is_available() else "cpu")`

mteb/models/model_implementations/mixedbread_ai_models.py
ayush1298: because torch is used for annotations:
`def _identity(logits: torch.Tensor) -> torch.Tensor:`
Are you referring to something else?

mteb/_torch_utils.py
Samoed: I think would be better to always check here

mteb/models/model_implementations/mixedbread_ai_models.py
Samoed:
```suggestion
def _identity(logits: torch.Tensor) -> torch.Tensor:
    """Keep the raw logit as the score, like ``torch.nn.Identity()`` but without importing torch."""
    import torch
    
    return torch.nn.Identity()
```

ayush1298: we have to return `logits` only, not `Identity()` module. The original `torch.nn.Identity()` is an object, and calling it on the `logits` returns the `logits.` In your version, `_identity` is a function that creates and returns a new `torch.nn.Identity()` and ignores `logits.` So `activation_fn(logits)` returns an Identity module rather than the scores, which is why it breaks. The current `_identity` simply returns `logits,` which is what Identity does, without needing torch.

mteb/_torch_utils.py
ayush1298: updated for `mps` check.

mteb/models/model_implementations/mixedbread_ai_models.py
Samoed: Agree then

KennethEnevoldsen (commented):
Hmm I am unsure if this is the approach I would go for here.

This is both slightly problematic for people to implement (it creates a lot of indirection and will be hard to debug). Also the MPS case will break some models while would be nice for others.

there is a few things here: 

Do we want to be able to work with ModelMetas without loading torch? or do we simply want mteb-core to not require torch?
One way to avoid having mteb-core require torch is to have the mteb-models as its own module. However I suppose a lot of the benefits of not having torch you could handle all info with it (and ModelMetas are info)

So assuming we want ModelMetas without requiring loading torch: I then I would consider if simply wrapping the loaders:

```
def create_model(**kwargs)
  import torch

  class ...

  return class(**kwargs)
```

would be better, though I could foresee some parallelization issues with this. @Samoed would love to hear your thoughts on this?

KennethEnevoldsen:
Same critique probably also applies to https://github.com/embeddings-benchmark/mteb/pull/5441 (but lets discuss it here before we figure out what to do with that)

Samoed:
I think format with wrapping is good

ayush1298:
@KennethEnevoldsen Thanks for reviewing. I checked each point that you raised, so a bit longer and detail reply. 

1. CUDA/MPS: `get_device` only replaces the `"cuda" if torch.cuda.is_available() else "cpu"` defaults (58 places). The 19 models that already pick `mps` keep their own code, so every model keeps its previous device. I checked this against main for all three machine types and by loading real models on a Mac. This PR doesn't change MPS behaviour.

Machine | 58 cuda/cpu models | 19 mps models
-- | -- | --
NVIDIA GPU | cuda → cuda | cuda → cuda
CPU-only | cpu → cpu | cpu → cpu
Apple Silicon | cpu → cpu | mps → mps

`get_device` uses an explicitly passed device as-is, and otherwise picks `cuda` if available, else `cpu.` A new model that should default to `mps` picks it itself. Also added this in `adding_a_model.md`.

2. Debugging and what will change for contributors: The decorator adds one stack frame, and imports inside functions don't change tracebacks. Also, the test explains what to fix when they break one and everything is documented in `adding_a_model.md,.`
    - `import torch, transformers and sentence_transformers` inside functions, not at the top of the file 
    - use `@no_grad / @inference_mode` from `mteb._torch_utils` instead of `@torch.no_grad() / @torch.inference_mode()`
    - write `device: str | None = None` with `device = get_device(device)` instead of a `torch.cuda.is_available()` default
    - use `OutputDType.BF16` instead of `torch.bfloat16` or similar other dtypes in `ModelMeta`
3. Wrapper approach: I checkd it. It does let contributors write normal torch code, but:
    - every load creates a new class, and instances can't be pickled, so worker processes and `torch.save` fail
    - results would store the loader name as `create_model` and loading a stored result picks the first loader with that name across the whole registry, so it could silently load the wrong model; existing results store class names, which would stop resolving
    - 13 classes subclass classes from other model files, so they can't all be nested
    - it re-indents ~173 classes a huge refactor
4. ModelMetas without torch: We need them because `API,` the `leaderboard` and `BenchmarkResults` all read them, and a separate `mteb-models` package would put them behind torch again. Another way to this is to split each model file in two: one file with only the ModelMeta, which points to the model class by name (e.g. `loader="...clip_models:CLIPModel"`), and a separate file with the model class, written as normal torch code. The class file would then only be imported when a model is loaded. I don't think that's worth it, though, because every model's metadata and code would live in separate files.

@Samoed also would love your opinions on the above points

KennethEnevoldsen:
@ayush1298, this is quite a large comment (and maybe a bit AI-generated). I am not sure that is very conducive to the discussion here. The actual point of discussion here is the ModelMeta. I see the argument for needing them, but 1) it is not like these aspects doesn't work currently and 2) I am not sure it is worth the large changes this would cause.

It is not like there aren't other ways to handle torch imports (for example, using a model registry, e.g. using something like [catalogue](https://github.com/explosion/catalogue)). Where we register the model in one place and the metadata in another. Only when loading the models is torch loaded.

This would require only a reference with requiring custom mteb wrappers around torch. It could also be argued that it would be better to use [lazy imports](https://peps.python.org/pep-0810/). The current path seems a bit costly on developers and it is not like the upside is huge. I consider this more a process of refactoring which shouldn't impose extra developer cost (otherwise it is not worth it).

ayush1298:
@KennethEnevoldsen comment is longer which I mentioned at start of comment because there are 3-4 things that you raised which I want to be clarified, else I keep back-and-forth clarifying same thing.

Also, the last point that you mentioned, related to cost on devlopers, I have also mentioned that in above comment, 2nd point. 

> Where we register the model in one place and the metadata in another. 

Also, mentioned in 4th point in my comment, the downside here is it would cause too much changes

Its not that its changing huge things for them. Still, let me know if you get the time to read that comment!

mteb/_torch_utils.py
```diff
@@ -0,0 +1,63 @@
...
+from typing import TYPE_CHECKING, ParamSpec, TypeVar
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+def no_grad(func: Callable[P, R]) -> Callable[P, R]:
```
KennethEnevoldsen: The developers can just use 

```
with torch.no_grad():
```
 inside the function

mteb/_torch_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    @functools.wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+        import torch
+
+        return torch.no_grad()(func)(*args, **kwargs)
+
+    return wrapper
+
+
+def inference_mode(func: Callable[P, R]) -> Callable[P, R]:
```
KennethEnevoldsen: Same here:

```
with torch.inference_mode()
```

should avoid a custom MTEB function

mteb/_torch_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    @functools.wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+        import torch
+
+        return torch.inference_mode()(func)(*args, **kwargs)
+
+    return wrapper
+
+
+def get_device(device: str | None = None) -> str:
```
KennethEnevoldsen: I don't think this function is needed either we can just use it directly in the class:

```
class:
  def __init__(..., device=None):
    device = "cuda" if torch.cuda.is_available() else "cpu"
```

KennethEnevoldsen (commented):
[will avoid discussing whether I read comments here and whether something was mentioned or not -- I think the below is more constructive] 

Took a time to reach through this. I think we can reduce a lot of the custom functionality a ton

docs/contributing/adding_a_model.md
```diff
@@ -182,6 +182,52 @@ your_model = ModelMeta(
 ```
 
 
+### Keeping your implementation importable without torch
```
KennethEnevoldsen: I think this need to be rewritten to be slightly more general and helpful.

I think the points I would like to highlight is:

1) We keep MTEB fast by only loading packages required for models when actually evaluating models (this applied to torch as much as any other custom package)
2) Here are a few design patterns to help you (use with torch.no_grad) instead of decorator. Here you can use the tabbed for each pattern so make it a bit more condensed.

ayush1298:
I have made all changes which are asked

ayush1298:
@Samoed @KennethEnevoldsen can you check this one whenever you get time

Samoed (approved):
Overall looks good, but I'm a bit afraid that we create a bit of friction for users who are adding models.

ayush1298:
> Overall looks good, but I'm a bit afraid that we create a bit of friction for users who are adding models.

Agree to some extent. Main thing would be `import torch` adding at function-level. `torch.no_grad()`/`torch.inference_mode()`  is used relatively by less models, so I guess maynot be a big problem

KennethEnevoldsen (commented):
yeah also mostly worried about the cost for users when adding models, but otherwise I think it looks good. However I think as it is now we can merge (@Samoed let me know I assume you agree given that you have approved)

KennethEnevoldsen (approved):

PR #5464, opened 2026-09-14
dataset: add JuriFindITRetrieval

albertobarnabo:
Adds `JuriFindITRetrieval` ([Findings of EACL 2026](https://aclanthology.org/2026.findings-eacl.221/)):
895 questions written by four Italian legal professionals over 23,113 statutory articles.

**Gap it fills**: mteb's only other Italian legal retrieval task is `MuPLeR-retrieval`,
whose Italian side is translated EU legislation with 200 LM-generated queries. This one
is Italian national law in the original, with expert-written queries.

The 169,301 LLM-generated questions shipped upstream are dropped, so the task is fully
human-authored; 159 empty placeholder rows and 345 duplicate articles go too, per the
docstring.

**Reproduction of the paper.** Table 3's "Pre-trained" block is zero-shot, on the paper's
179-question validation set, whose questions keep a `validation-` id prefix in `test`
here. Filtering to them and running `Qwen/Qwen3-Embedding-0.6B` through `mteb.evaluate`:

| | R@5 | R@20 | R@100 | nDCG@5 | nDCG@100 | mAP@100 |
|---|---|---|---|---|---|---|
| paper | 58.82 | 70.74 | 79.52 | 54.27 | 59.04 | 50.51 |
| here | 60.81 | 74.12 | 81.32 | 55.56 | 60.24 | 51.40 |

Within 1–3.4 points and consistently slightly above, as the deduplication predicts: a
relevant article's twin can no longer take a rank ahead of it.

**Neither trivial nor random**, nDCG@10 over all 895 queries: random 0.000, potion-128M
0.308, bm25s 0.478, e5-small 0.487, e5-large 0.594. Runtime 3.8 min.

- [x] I have run the following models on the task (adding the results to the pr).
- [x] I have considered the size of the dataset and reduced it if it is too big
- [x] I reproduced scores from the original paper and added them above.

Drafted with Claude Code; I ran and verified every number above.


Samoed:
Can you reproduce scores from paper? You reprort `ndcg@10`, but in paper there is no results with this metric

albertobarnabo:
> Can you reproduce scores from paper? You reprort `ndcg@10`, but in paper there is no results with this metric

Hi Samoed, 
Sorry I made a mistake. 
I am working on a fix: their Table 3 "Pre-trained" block is zero-shot, so it is comparable, and it is measured on their 179-question validation set — which this task exposes as validation for exactly that reason. I'm running Qwen3-Embedding-0.6B on it now and will post the comparison at the paper's own cutoffs.

albertobarnabo:
Reproduced with Qwen3-Embedding-0.6B on the 179 validation questions from the paper: recall@5 60.81 vs their 58.82, nDCG@5 55.56 vs 54.27. A bit higher everywhere, probably because I collapse 345 duplicate articles onto one id.

Also dropped the extra validation split, it was a copy of part of test.

Samoed:
Can you run jina-v3 or snowlake, since they're not use prompts and they're easier to reproduce

albertobarnabo:
Hi, yes I tried it earlier and ran into some dependencies issues in my venv, but I will try again later today.

albertobarnabo:
> Can you run jina-v3 or snowlake, since they're not use prompts and they're easier to reproduce

Hi, I Tried both again and neither runs for me on mteb 2.x: jina-v3 hits a cache_dir signature clash in its remote code, snowflake-arctic-embed-m-v2.0 a device-side index assert on two words of input. Every result I sampled for those two in the results repo is from mteb 1.18-1.31, so I don't think either has been run since transformers 5. Can you double check this? Maybe I am missing something, I don't get it. 

What I did is I measured the prompt effect instead. With the task's query prompt Qwen3-0.6B gets nDCG@5 55.56 and recall@5 60.81, without it 52.13 and 58.14. The paper is at 54.27 and 58.82, in between. So the prompt is worth about 3 points either way and the paper sits between the two runs.

Samoed:
Yes, I think they are not compatible withh transformers v5. I'll investigate compatible versions

Samoed:
I created pr with dependencies fix, hope this would help

albertobarnabo:
> I created pr with dependencies fix, hope this would help

Thanks! I'm rebuilding an env with the pins from your PR (transformers<5, xformers, einops) and running arctic-embed-m-v2.0 and jina-v3 on the paper's 179 validation questions, at the paper's cutoffs. I will upload the numbers when done :)

albertobarnabo:
> I created pr with dependencies fix, hope this would help

Hi, I am done. 

Results for all nine models on the full split are here: https://github.com/embeddings-benchmark/results/pull/719

On the paper's 179 validation questions arctic-embed-m-v2.0 matches Table 3 almost exactly (recall@5 63.49 vs 63.77, nDCG@5 58.02 vs 59.31); jina-v3 comes out 4-7 points higher, probably because mteb applies its retrieval LoRA adapters.

Two notes on the env: jina only loaded with sentence-transformers 4.1, with 5.x its custom_st.py passes cache_dir twice, so <5.7 isn't enough. And I ran without flash-attn, the torch 2.8 wheel is 2.8.3 which xformers 0.0.32 refuses.

Samoed (approved):
Great! Thanks for testing dependencies, I updated PR

PR #5465, opened 2026-09-14
fix: don't place pair classification thresholds between tied scores

RizgarOzan:
## What
`_find_best_acc_and_threshold` and `_find_best_f1_and_threshold` now skip candidate thresholds that fall between two rows with the same score.

## Why
The threshold search places a cut between every pair of consecutive sorted rows, including rows with identical scores. No real threshold can separate those, so the order of tied rows in the dataset leaks into `accuracy`/`f1` (the same class of issue as #5092). With a constant scorer, labels `[1, 1, 0, 0]` give accuracy 1.0 / f1 1.0, while `[0, 0, 1, 1]` gives 0.25 / 0.4. Ties are common for integer-valued distances (e.g. manhattan on binary/int8 embeddings) and duplicate pairs. `max_accuracy` is the main score of 4 PairClassification tasks; `ap` is unaffected.

## How
Counts are still accumulated for every row, but a threshold is only evaluated where the next score differs. Results for inputs without ties are unchanged. If *all* scores are tied, no threshold is evaluated and accuracy/f1 are 0.0; the all-positive/all-negative cuts were never evaluated before either, so I left that as is.

## How I tested
- New `tests/test_abstasks/test_pair_classification_metrics.py`: row-order independence for constant and partially tied scores (both `high_score_more_similar` directions), plus an untied case. 4 of 5 fail before the fix, all pass after.
- `pytest -n auto -m "not (test_datasets or leaderboard_stability or test_reference_models)"`: 7192 passed; the only failures are environment-related on Windows (missing torchcodec, cp1254 decoding) and also fail on `main`.
- `ruff format --check .`, `ruff check .`, `typos` clean.


Samoed:
Can you run some models to see if performance would change?

KennethEnevoldsen:
Yeah we need some indication that this is a problem and to what extent it is a problems (we need to know what it affects)

RizgarOzan:
Ran it: `main` vs this PR on identical cached embeddings, full default splits of TwitterSemEval2015, SprintDuplicateQuestions, TwitterURLCorpus and LegalBenchPC (plus Cmnli/Ocnli/TalemaaderPC for e5).

- **Float embeddings** (all-MiniLM-L6-v2, bge-small-en-v1.5, multilingual-e5-small, 15 model × task pairs): no metric changes. Ties do exist (TwitterURLCorpus repeats 973 pairs with conflicting labels), but the best threshold never lands inside one.
- **Binary embeddings** (`CompressionWrapper(OutputDType.BINARY)`): LegalBenchPC main score 0.6543 → 0.6519 (MiniLM) and 0.6431 → 0.6411 (bge-small); max_f1 moves by at most −0.0047 (Sprint, MiniLM). int8 is unchanged, and AP never changes.
- On `main` those binary numbers depend on row order: over 200 row shuffles, bge-small binary LegalBenchPC max_accuracy ranges 0.6411–0.6460. With this PR the result is the same for every order.

So it only matters for heavily tied scores (binary/quantized outputs); no score for a float model changes. If that's too niche for you, I'm happy to close it.

The Windows 3.10 failure is unrelated: `test_ensure_no_torch_at_import` reads the model files without `encoding="utf-8"`, and it fails the same way on `main` (e.g. run 34895765902).

ayush1298:
@RizgarOzan failing windows CI is fixed on main, can you merge main here

RizgarOzan:
Thanks — merged `main` in b1e6efbd (no conflicts). `tests/test_abstasks` passes locally (193 passed; the MAEB mock tasks are skipped here only because torchcodec can't load without FFmpeg on this machine), ruff clean.

KennethEnevoldsen (approved):
No thanks for specifying. I think this is good to change.

It will affect a few results only, but not main results. If it were to affect main results we would have to rerun or reversion all pair classification tasks (which is not easy)

KennethEnevoldsen:
cc @Samoed: I think this is good to merge, but wanted your eyes on it as well. I think the biggest issue is that we need to be aware that binary models needs to be rerun. I could only find these two cases:

https://leaderboard.mteb.org/models/mongodb/voyage-3.5%20(output_dtype%3Dbinary)
https://leaderboard.mteb.org/models/Cohere/Cohere-embed-v4.0%20(output_dtype%3Dbinary)

which doesn't seem to have been run on pair clf tasks

Samoed:
Yes, agree

Samoed (approved):

PR #5466, opened 2026-09-14
Add model: Singaraj/sante-embed

LK-maker-007:
`Singaraj/sante-embed` is a medical retrieval model fine-tuned from `Alibaba-NLP/gte-Qwen2-1.5B-instruct` with LoRA, merged into the base weights. Nine languages, 1536 dimensions, apache-2.0.

`loader_kwargs` follows the base entry but sets `max_seq_length=512`. The checkpoint truncates there; leaving it unset makes the wrapper read `max_position_embeddings` (131072) and the long-document tasks run out of memory.

`training_datasets` declares NFCorpus and SciFact, both used as replay data. That resolves to 75% zero-shot on MTEB(Medical, v1), SciFact-PL included.

Results follow in a results-repo PR.

No paper, so the last item does not apply.


ayush1298:
seems good. @Samoed

Samoed (approved):

PR #5450, opened 2026-09-12
Add tencent/EVIE-8B and tencent/EVIE-4.5B

officea1t:
Adds `tencent/EVIE-8B` and `tencent/EVIE-4.5B`, ColQwen3.5-based late-interaction visual document retrievers.

### Implementation
`colpali_engine` builds `ColQwen3_5` causal. EVIE was trained with bidirectional attention on full-attention layers, so `EvieWrapper` flips `is_causal` after load. It also pins `max_num_visual_tokens=1024` (released default 16384) and `attn_implementation="flash_attention_2"` (bf16 sdpa vs FA2 moves a ViDoRe v3 domain by ~0.2). Install with `pip install mteb[evie]`.

### Reproduction — ViDoRe(v3), nDCG@10
| domain | 8B card | 8B measured | Δ | 4.5B card | 4.5B measured | Δ |
|---|---:|---:|---:|---:|---:|---:|
| ComputerScience | 81.86 | 82.13 | +0.27 | 81.72 | 81.96 | +0.24 |
| Energy | 72.51 | 72.23 | −0.28 | 72.32 | 72.01 | −0.31 |
| FinanceEn | 71.23 | 71.15 | −0.08 | 70.00 | 69.91 | −0.09 |
| FinanceFr | 56.40 | 55.81 | −0.59 | 54.90 | 54.66 | −0.24 |
| Hr | 69.29 | 68.86 | −0.43 | 67.82 | 67.58 | −0.24 |
| Industrial | 59.77 | 59.11 | −0.66 | 59.40 | 58.78 | −0.62 |
| Pharmaceuticals | 70.81 | 70.31 | −0.50 | 70.27 | 70.04 | −0.23 |
| Physics | 52.11 | 51.95 | −0.16 | 51.69 | 52.09 | +0.40 |
| **mean (8 public)** | **66.75** | **66.45** | **−0.30** | **66.02** | **65.88** | **−0.14** |

`measured` is mteb nDCG@10 (`pytrec_eval.ndcg_cut`, linear gain). `card` uses exponential gain (`2^r−1`, protocol `ndcg2r`). ViDoRe v3 has graded qrels, so the two diverge; we submit the linear-gain numbers. Leaderboard mean also includes 2 private tasks we cannot run locally.

### Mock-run
19/19 search tasks pass. 26 single-vector mock tasks fail as expected (one vector per token, `ScoringFunction.MAX_SIM`).

### Checklist
- [x] ModelMeta filled
[...]

PR #5469, opened 2026-09-15
model: restore bidirectional-attention fallback for PyPI colpali_engine

### LINKED ISSUE #5451: Evaluation request — tencent/EVIE-8B and tencent/EVIE-4.5B on ViDoRe(v3) private tasks

https://huggingface.co/tencent/EVIE-8B ;https://huggingface.co/tencent/EVIE-4.5B

### What do you want it to be evaluated on?

Models: `tencent/EVIE-8B`, `tencent/EVIE-4.5B`
Benchmark: ViDoRe(v3)

Only the two private tasks are needed:

- `Vidore3NuclearRetrieval`
- `Vidore3TelecomRetrieval`

The 8 public domains have already been run locally and submitted via
embeddings-benchmark/results#711.

The implementation is in embeddings-benchmark/mteb#5450, so
`mteb.get_model("tencent/EVIE-8B")` works once that PR is merged.

Dependencies: `pip install mteb[evie]`. Note that flash_attention_2 is a hard
requirement for these models — the wrapper hard-codes it and raises ImportError
if flash-attn is missing, because under bf16 the sdpa reduction order moves a
ViDoRe v3 domain by ~0.2 nDCG@10 and we do not want the score to depend on
whether flash-attn happened to be installed.

Protocol the wrapper pins: bf16, `max_num_visual_tokens=1024` , bidirectional attention, MaxSim scoring.

### Are you interested in contributing to the evaluation of this model?

No

### Does this model already exist in MTEB?

Yes

xiaoxiaoshadiao:
Follow-up to #5450, fixes #5451. Same team as @officea1t (different account, sorry for the mismatch).

### Bug

`ColQwen3_5.enable_bidirectional_attention` ships with the EVIE release of `colpali_engine` ([source](https://github.com/Tencent/EVIE/blob/main/colpali/colpali_engine/models/qwen3_5/colqwen3_5/modeling_colqwen3_5.py#L20)) but not with the PyPI wheels the `evie` extra resolves to, so a real load raises `AttributeError`. #5450 originally had a `hasattr` guard; it was dropped during review in favour of calling the method directly — my fault for not flagging the packaging split. Mock tasks never reach `from_pretrained`, so CI couldn't catch it.

### Fix

Restore the guard with a local fallback. Both switches matter and are not interchangeable:

- `config.is_causal=False` — what `create_causal_mask()` reads; on padded batches this mask is what sdpa honours.
- `Qwen3_5Attention.is_causal=False` — what flash_attention_2 reads, as FA2 only gets the 2D padding mask.

Module flag alone leaves sdpa **silently causal**, which matters since the wrapper falls back to sdpa without flash-attn.

### Verification

Both implementations applied to the same upstream `ColQwen3_5` instance, `last_hidden_state` compared bit for bit on a padded batch (reduced 4-layer build of the released config):

| | sdpa | FA2 |
|---|---|---|
| causal baseline | `+25.9662` | `+26.2500` |
| released impl | `−43.1758` | `−9.4375` |
| this PR | `−43.1758` ✅ | `−9.4375` ✅ |
| module flag only | `+25.9662` ❌ no-op | `−9.4375` ✅ |

Last row is the workaround from #5451, shown to explain why the config flag is needed.


mteb/models/model_implementations/evie_models.py
```diff
@@ -53,7 +80,13 @@ def __init__(
             model_name=model_name, revision=revision, device=device, **kwargs
         )
 
-        self.model.enable_bidirectional_attention()
+        # `enable_bidirectional_attention` ships with the EVIE release of
+        # colpali_engine; the published PyPI wheels do not carry it yet, so fall
+        # back to an equivalent local implementation instead of raising.
```
KennethEnevoldsen:
```suggestion
```

mteb/models/model_implementations/evie_models.py
```diff
@@ -11,6 +11,33 @@
 logger = logging.getLogger(__name__)
 
 
+def _enable_bidirectional_attention(model: Any) -> None:
+    """Encoder-ize the full-attention layers of a ColQwen3.5 backbone.
+
+    Mirrors `ColQwen3_5.enable_bidirectional_attention` from the EVIE release.
```
KennethEnevoldsen: can you add a direct link here?

mteb/models/model_implementations/evie_models.py
xiaoxiaoshadiao: Done in 0d575e9 - comment removed.

mteb/models/model_implementations/evie_models.py
xiaoxiaoshadiao: Added in 0d575e9 - links to the function in the EVIE release: https://github.com/Tencent/EVIE/blob/main/colpali/colpali_engine/models/qwen3_5/colqwen3_5/modeling_colqwen3_5.py#L20

KennethEnevoldsen (approved):

KennethEnevoldsen:
@xiaoxiaoshadiao can I ask you to lint then I think we are good

Samoed:
> but not with the PyPI wheels the evie extra resolves to, so a real load raises AttributeError. https://github.com/embeddings-benchmark/mteb/pull/5450 originally had a hasattr guard; it was dropped during review in favour of calling the method directly — my fault for not flagging the packaging split. Mock tasks never reach from_pretrained, so CI couldn't catch it.

Do we need `self.model.enable_bidirectional_attention()` in the first place then?

PR #5470, opened 2026-09-15
fix: reopen existing NumpyCache without a -1 memmap shape

RizgarOzan:
## What

`NumpyCache.load()` fails on any cache directory that already exists, and `CachedEmbeddingWrapper` calls it for every task. A second run with the same `cache_path` raises instead of reusing the stored embeddings.

## Why

Both places that reopen `vectors.npy` call `np.memmap(..., mode="r+", shape=(-1, dim))`, and `np.memmap` does not infer a `-1` dimension. Standalone check (Windows, Py 3.12):

- numpy 1.26.4, 2.0.2, 2.1.3: `OverflowError: memory mapped length must be positive`
- numpy 2.2.6, 2.3.5, 2.4.0, 2.4.6: `ValueError: negative dimensions are not allowed`

So every numpy in the supported range (`>=1.23,<3`) is affected. No test reopened a cache from a new wrapper, so CI missed it.

## How

A small `_open_vectors_file()` helper derives the row count from the file size (`st_size // (4 * dim)`) and opens the memmap with that explicit shape. `load()` and `_initialize_vectors_file()` both use it. The on-disk format is unchanged.

## Tests

- `test_cache_reused_by_new_wrapper`: a second wrapper on the same directory returns the same embeddings without calling the model.
- `test_numpy_cache_reload_and_grow[1|3]`: empty dir -> add 1 or 3 vectors -> reopen -> add enough to double the file -> reopen and check every vector.

`pytest tests/test_models/test_cached_embedding_wrapper.py`
- before: 3 failed, 4 passed, 1 skipped (faiss not installed)
- after: 7 passed, 1 skipped

I ran the same reload/grow check on numpy 1.26.4, 2.1.3, 2.2.6 and 2.4.6. It fails on `main` with all four and passes on this branch with all four. `ruff check` and `ruff format --check` are clean; mypy reports nothing in `numpy_cache.py`.

`pytest tests -n 8` (Windows): no failures touch the cache. I reran the non-network ones serially: `main` and this branch have the same 6 failures (torchcodec video, `MTEB(slk, v1)` coverage, Windows `read_text` encoding). The rest are HF Hub network errors. The `windows-latest, 3.10` CI job is already red on `main` (same `read_text` failure).


Samoed (approved):
Good catch!

PR #5474, opened 2026-09-15
fix typing

Samoed:


RizgarOzan:
Sorry about that, and thanks for the quick fix — the annotation on `_open_vectors_file` was mine in #5470 and I only ran the test suite, not `mypy`. Running `uv run --no-sync mypy mteb` before opening is now part of my checklist for this repo.

PR #5475, opened 2026-09-15
fix: fix jina-v3 and arctic-embed dependencies

Samoed:
Pin jina-v3 and arctic-embed to `transformers<5`. I didn't change `snowflake-arctic-embed-l-v2.0`, becuase this variant don't use `remote_code`. Ref https://github.com/embeddings-benchmark/mteb/pull/5464#issuecomment-5673742543. I tried to regenerate `uv.lock`, but this takes a lot of time (more than 30 min)


KennethEnevoldsen (approved):
do we need to generate uv.lock?

Samoed:
Ideally yes, but I think we can merge without it

Issue #5468, opened 2026-09-15
New windows test is unstable or fails


KennethEnevoldsen:
@ayush1298 seems like we have a test that fails for windows (or atleast sometime fails), can I ask you to look into that one? (it seems to be related to your recent refactors)

_Originally posted by @KennethEnevoldsen in https://github.com/embeddings-benchmark/mteb/issues/5467#issuecomment-5677209693_

ayush1298:
@KennethEnevoldsen yes, and I have already added its fix in: https://github.com/embeddings-benchmark/mteb/pull/5463
Do you want me to separate it out as a separate PR, so we can merge quickly?

PR #5476, opened 2026-09-16
fix: apply the n_parameters lower bound when there is no upper bound

RizgarOzan:
## What

`get_model_metas(n_parameters_range=(lower, None))` now filters. The lower-bound check sat
inside `if upper is not None`, so a range without an upper bound was a no-op.

## Why

It is reachable from the public API, not just the Python call: `/models?min_params_b=1`
maps straight onto this tuple (`mteb/api/routes.py:651`), so it returns the whole registry
instead of the 1B+ models. `benchmark_results.py:236` forwards the same tuple.

On `main`:

```
all models        : 929
(1e9, None)       : 929   <- 567 of them are under 1B
(1e9, 1e13)       : 296   <- same bound + a redundant upper bound works
(None, 1e8)       : 143   <- upper-only works
```

The docstring already promises the behaviour: "A tuple of lower and upper bounds... If
(None, None), this filter is ignored."

## How

```python
if lower is not None or upper is not None:
    if n_parameters is None:
        continue
    if upper is not None and n_parameters > upper:
        continue
    if lower is not None and n_parameters < lower:
        continue
```

Upper-only is byte-identical to before, including dropping models whose `n_parameters` is
unknown. A lower-only range now drops those 66 models too, which is the same rule.

## How I tested

Windows 11, Python 3.14.

- 3 new tests in `tests/test_models/test_model_meta.py`, red before / green after: the
  lower-bound one fails on `main` and passes with the fix, while the upper-only and
  `(None, None)` tests pass both ways, so they pin the unchanged behaviour.
- `pytest tests/test_models/test_model_meta.py tests/test_get_tasks.py` in one process:
  **2707 passed**, 0 failures.
- After the fix `(1e9, None)` returns 296, matching `(1e9, 1e13)`.
- `ruff check` / `ruff format --check` clean.


RizgarOzan:
Superseded by my next comment below, which has the same analysis plus the tracking issue for the encoding failure. Sorry for the duplicate.

RizgarOzan:
The `test (windows-latest, 3.10)` job is red for two reasons, neither from this PR:

- `test_ensure_no_torch_at_import.py::test_model_implementations_declare_no_import_time_torch_dtypes` — `read_text()` without `encoding=` hits cp1252 (`'charmap' codec can't decode byte 0x8f`). Already fixed in #5463 (tracked by #5468), and it fails the same way on unrelated PRs, e.g. run 35055411494.
- `test_similarity_functions.py::test_hps_collapses_spurious_ties[dtype1]` — native crash, `Windows fatal exception: code 0xc000001d` (illegal instruction) at `test_similarity_functions.py:70`, not a Python assertion.

This PR only touches the `n_parameters_range` guard in `mteb/models/get_model_meta.py` and adds three tests in `tests/test_models/test_model_meta.py`; neither failing test exercises `get_model_metas`. All 11 other jobs are green, including every Ubuntu matrix entry.

isaac-chung (approved):
Thanks!

Issue #5416, opened 2026-09-08
Quality Concerns in 11 MMTEB Datasets


shuangli-z:
I analyzed all 131 MMTEB tasks by collecting scores from the 172 models that have complete results across all tasks on the leaderboard, computing per-dataset statistics (Kendall's τ across model tiers, μ, σ), and checking the official BM25 baseline (`mteb__baseline-bm25s` in [embeddings-benchmark/results](https://github.com/embeddings-benchmark/results)) for retrieval tasks. For flagged datasets, I spot-checked sampled examples to verify the issues. I found **11 datasets** with concerns in 3 categories. Details below. Happy to discuss any of these individually.

---

## 1. Reasoning Tasks Mislabeled as Retrieval (3)

These are labeled as retrieval but actually test reasoning. Model rankings reflect reasoning ability, not embedding quality.

| Dataset | Type | τ | μ | Issue | Notes |
|---------|------|:---:|:---:|-------|-------|
| **TempReasonL1** | Retrieval | 0.37 | 2.0 | Temporal reasoning, not retrieval | Queries require date arithmetic (e.g. "What is the time 3 year and 3 month before Mar, 1827" → "Dec, 1823"). μ=2.0 across 172 models. |
| **SpartQA** | Retrieval | 0.41 | 9.0 | Spatial reasoning, not retrieval | Multi-paragraph spatial reasoning puzzles (e.g. "There are three blocks … What is to the right of the blue object?"), not search. |
| **WinoGrande** | Retrieval | 0.42 | 39.4 | Commonsense reasoning, not retrieval | Coreference resolution testing world knowledge (e.g. "The computer of Victoria ran faster than that of Carrie because _ downloaded less files." → "Victoria"). |

---

## 2. Linguistic Acceptability Beyond Embedding Reach (2)

Correct and incorrect sentences can have identical meaning. The difference is purely grammatical. Embeddings encode semantics, not grammatical well-formedness, so this task is structurally beyond what embeddings measure.

| Dataset | Type | τ | μ | σ | Notes |
|---------|------|:---:|:---:|:---:|-------|
| **DalajClassification** | Classification | 0.08 | 50.4 | 1.7 | Swedish linguistic acceptability. τ=0.08, virtually no model can solve this. |
[...]

PR #5479, opened 2026-09-16
benchmark: add MTEB(ita, v1) Italian benchmark

albertobarnabo:
Italian is one of the last large European languages without an MTEB benchmark.
This adds `MTEB(ita, v1)`: 23 tasks over 7 task types.

**Selection.** I took every task with an Italian subset (80, 52 text-only), kept the
monolingual ones, then measured each against the score it gives away for free —
1/n_classes, the positive rate, or BM25. Four failed that and are out.

| type | n | tasks |
|---|---|---|
| Classification | 9 | DadoEvalCoarse, ItaCasehold, MassiveIntent, MassiveScenario, MultiHate, SardiStance, SIB200.v2, SwissJudgement, TweetSentiment |
| Multilabel | 2 | Emit, MultiEURLEX |
| PairClassification | 1 | DisCoTex |
| Clustering | 1 | SIB200ClusteringS2S |
| Retrieval | 6 | Belebele, JuriFindIT, Mintaka, WebFAQ, WikipediaRetrievalMultilingual, XPQA |
| Reranking | 2 | WikipediaRerankingMultilingual, XGlueWPR |
| STS | 2 | STS22.v2, STSBenchmarkMultilingualSTS |

Five are natively Italian: DadoEval, SardiStance, Emit, DisCoTex (EVALITA) and
ItaCasehold. Bitext mining, `MultilingualNano*` and cross-lingual pairs are also out.

**Dropped on measurement**: `MMarco` (training distribution of several Italian
retrieval models, mine included); `XStance` (every model under its 0.539
random-ranking AP, and the best threshold labels every pair positive); `Itacola.v2`
(three models at 0.493–0.500 on a binary task); `IlPostRetrieval`, which I built for
this benchmark and dropped when BM25 came within 0.019 of the best model.

**Runtime** 41 min for `multilingual-e5-large` on one A100, 21 of them WebFAQ. `MultiLongDoc` retrieval and reranking dropped in review: widely trained on.

Two things worth flagging. `MultiEURLEX` is on your list in
#5416 and no model beats its constant baseline here either — glad to drop it, I kept
it only because Europe/Multilingual/nld carry it. And `reference_models` stays red
until embeddings-benchmark/results#720 lands, as it has for slk since August.

Drafted with Claude Code; I ran and verified every number above.


mteb/benchmarks/benchmarks/benchmarks.py
```diff
@@ -1213,6 +1213,72 @@
...
+                "MultiEURLEXMultilabelClassification",
+                # Clustering (1)
+                "SIB200ClusteringS2S",
+                # PairClassification (1)
+                "DisCoTexPairClassification",
+                # Retrieval (7)
+                "BelebeleRetrieval",
+                "JuriFindITRetrieval",
+                "MintakaRetrieval",
+                "MultiLongDocRetrieval",
```
Samoed: MultiLongDocRetrieval is used for training a lot and don't have good quality

albertobarnabo: Good point, removed. I also dropped MultiLongDocReranking since it's built from the same data.

isaac-chung:
@Samoed @KennethEnevoldsen wondering if we should accept benchmarks without citations here. I think we should not.

Samoed:
I think we might want to add some benchmarks without citations, e.g. vidore/rteb, but not in this case

albertobarnabo:
Hi guys, I understand where this comes from. Thank you anyway for the review!

PR #5483, opened 2026-09-18
fix: accept language-script and programming-language codes in filter_tasks

RizgarOzan:
## What
`filter_tasks` (and therefore `mteb.get_tasks(languages=...)` without `tasks=`) rejected two kinds of codes that the rest of the language handling accepts:

- language-script codes: `mteb.get_tasks(languages=["eng-Latn"])` -> `ValueError: Invalid language code: 'eng-Latn'`, although the docstrings of `get_tasks` and `filter_tasks` list `"eng-Latn"` as a valid input, and `get_tasks(tasks=[...], languages=["eng-Latn"])` works (it goes through `filter_languages`/`LanguageScripts`).
- programming languages that appear in task metadata (`python`, `java`, `c++`, ... 14 in total), e.g. `get_tasks(languages=["python"])`.

## How
- `_check_is_valid_language` validates the language part against `ISO_TO_LANGUAGE` or `PROGRAMMING_LANGS`, then calls the existing `check_language_code` so the script part of `eng-Latn` is validated too (`eng-Xxxx`, `zzz` still raise).
- The task-level match also checks `metadata.bcp47_codes`, so `eng-Latn` keeps tasks that have `eng-Latn`.

Plain 3-letter codes behave exactly as before: they can never match a `bcp47` code (those always contain `-`).

## How I tested
- New tests in `tests/test_filter_tasks.py`: `eng-Latn`, `rus-Cyrl`, `python` fail on main (3 failed), pass with the fix; invalid codes still raise.
- `pytest tests/test_filter_tasks.py tests/test_get_tasks.py`: 143 passed; `tests/test_benchmarks/test_get_benchmarks.py`: passed.
- Over all 1993 tasks: for each of the 1136 language codes in metadata, the result is identical to the old rule (0 differences); all 1166 language-script codes now return a non-empty list (all raised before).
- `ruff format --check`, `ruff check`, `typos` clean; `mypy mteb` reports nothing in `filter_tasks.py` (the remaining local errors are in `mteb/api/*` from missing optional deps).


Samoed (approved):
Good catch! I simplified a bit

PR #5478, opened 2026-09-16
fix: Add tests to validate citations and update citations

KennethEnevoldsen:
A little less than 100 cases, where found. I manually went though these.

If a case was found missing (even if just by surface features then i replaced it with a google scholar reference, ACL, etc.. If you came up a second time I wrote an exception.

- [x] Updated citation that did not have a match, these error were various (most are mild a few are bad):
  - shortened journal articles
  - incorrectly referencing the repo
  - draft citations (where no paper existed), e.g. "Extending the Massive Text Embedding Benchmark to French"
  - some where using github handles as author names
  - where arxiv id wsa incorrect
  - master thesis that have since been taken down but published as an article (wtf!)
- [x] Deleted a few citations, these include due to:
  - where articles (e.g. [this](https://huggingface.co/blog/thebajajra/rexrerankers)) were taken down
  - where I couldn't reconstruct the article ([this on](https://github.com/embeddings-benchmark/mteb/blob/629aa89df69a48fb43920bb42d2f55a29f432c2c/mteb/tasks/clustering/zho/cmteb_clustering.py#L334-L339) seems to refer to the full conference proceedings
  - where they refer to it refers to the wrong article ([here](https://github.com/embeddings-benchmark/mteb/blob/629aa89df69a48fb43920bb42d2f55a29f432c2c/mteb/tasks/retrieval/dan/tv2_nordretrieval.py#L30-L31))
- [x] updated description for bibtex citations

Please take care to review this in detail.

ref #5471

PR #5488, opened 2026-09-19
model: add albertobarnabo/bge-m3-italian

albertobarnabo:
Adds `albertobarnabo/bge-m3-italian`, an Italian fine-tune of `BAAI/bge-m3` (same architecture: 1024 dims, 8192 tokens, MIT). Trained on the Italian split of mMARCO, which is declared in `training_datasets`; the base model's data is inherited via `adapted_from`.

It loads with the standard `SentenceTransformerEncoderWrapper`. I evaluated it through `mteb.get_model` / `mteb.evaluate` on the Italian tasks (JuriFindITRetrieval 0.617 nDCG@10, vs 0.594 for multilingual-e5-large); results will follow in a separate PR to the results repo once this is in.

Drafted with Claude Code; I ran and verified every number.


ayush1298:
seems good. @Samoed can you check it?
@albertobarnabo if you have results can you open PR in results repo as well

mteb/models/model_implementations/italian_models.py
```diff
@@ -0,0 +1,35 @@
...
+    # Fine-tuned on the Italian split of mMARCO (train queries and passages), so
+    # the mteb task built on that data is listed here; the bge-m3 base data is
+    # inherited through adapted_from.
+    training_datasets={"MMarcoRetrievalMultilingual"},
+    citation="""@misc{bge-m3-italian-2026,
+      author = {Barnabo, Alberto},
+      title = {bge-m3-italian: Italian retrieval fine-tune of bge-m3},
+      year = {2026},
+      url = {https://huggingface.co/albertobarnabo/bge-m3-italian}
+}""",
```
Samoed: Agree metadata looks good, but I'm not sure if we want ot keep citation, because we clear HF citations in https://github.com/embeddings-benchmark/mteb/pull/5478

albertobarnabo:
@ayush1298 done, results are in embeddings-benchmark/results#722
@Samoed I have updated the citation

let me know :)

Samoed (approved):

PR #5490, opened 2026-09-19
fix: apply scripts filter in TaskResult.get_score

RizgarOzan:
## What

`TaskResult.get_score(scripts=[...])` ignores `scripts`. It is passed to `LanguageScripts.from_languages_and_scripts`, but only `contains_language` is checked. With no `languages`, that returns `True` for every subset, so the script filter does nothing:

```python
# subsets srp-Cyrl (0.2) and srp-Latn (0.8)
tr.get_score(scripts=["Cyrl"])                    # 0.5, expected 0.2
tr.get_score(languages=["srp"], scripts=["Latn"]) # 0.5, expected 0.8
```

It has been like this since `get_score` was added in #658. `ModelResult._get_scores` sends any `scripts` call down this slow path, so the parameter is clearly meant to work. 165 of the 216 multi-subset tasks cover more than one script.

## How

A subset now also has to match the script (`LanguageScripts.contains_script`) when `scripts` is given. When it isn't, nothing changes. An empty list is treated as no filter, like `filter_tasks(script=...)` does.

## How I tested

- New `test_task_results_get_score_scripts` fails on main and passes with the fix. The existing `get_score` test is unchanged.
- Full suite with the CI marker filter: 7241 passed, 3 failed. The same 3 fail on main locally (two video tests need video deps; `test_ensure_no_torch_at_import` is the known Windows encoding failure).
- `ruff check`, `ruff format`, `typos` are clean. mypy reports no errors in `task_result.py`.


mteb/results/task_result.py
```diff
@@ -574,7 +574,9 @@ def get_score(
             for scores in self.scores[split]:
                 eval_langs = scores["languages"]
                 for lang in eval_langs:
-                    if lang_scripts.contains_language(lang):
+                    if lang_scripts.contains_language(lang) and (
+                        not scripts or lang_scripts.contains_script(lang.split("-")[-1])
```
Samoed:
```suggestion
                        not scripts or lang_scripts.contains_script(language.rsplit("-", maxsplit=1)[-1])
```

Samoed (approved):

PR #5491, opened 2026-09-19
task: add VATEX Modality Equivalence retrieval tasks

### LINKED ISSUE #5358: MOEB [Paper Experiment] - Modality Equivalence

Rakshitha-Ireddi:
Closes #5358 (video modality extension).
  
  Adds six retrieval directions over a shared 150-clip VATEX pool where every item is simultaneously available as a video clip, audio track, and English text caption. Because the candidate pool is identical across all directions, score differences reflect modality difficulty rather than content difficulty.
  
  Tasks added:
  - VATEXModalEquivV2TRetrieval — video → text
  - VATEXModalEquivT2VRetrieval — text → video
  - VATEXModalEquivV2ARetrieval — video → audio
  - VATEXModalEquivA2VRetrieval — audio → video
  - VATEXModalEquivA2TRetrieval — audio → text
  - VATEXModalEquivT2ARetrieval — text → audio
  
  Dataset: mteb/VATEX_test_1k (existing, no new upload needed).
  Pool is constructed at eval time by selecting the first 150 items — same pool across all six directions.
  
  Extends the COCO Modality Equivalence tasks from PR #5384 to cover the {video, audio, text} triple, completing the modality equivalence experiment requested by @AdnanElAssadi in #5358.


Samoed (changes requested):
How this task is different from merged vatex task?

Rakshitha-Ireddi:
> How this task is different from merged vatex task?

 @Samoed  — The existing VATEX tasks (VATEXV2TRetrieval, VATEXT2VRetrieval, etc.) have different content per query the retrieval pool changes across tasks. In these modality equivalence tasks, all 6 directions share the exact same 150-clip pool, so score differences isolate modality difficulty rather than content difficulty. This mirrors the COCO Modality Equivalence tasks from #5384 which established this pattern for {image, text, audio} — this extends it to {video, text, audio}. For example, comparing VATEXModalEquivV2TRetrieval vs VATEXModalEquivA2TRetrieval directly answers "is video or audio a harder query modality for text retrieval?" — something the existing VATEX tasks cannot answer.

Samoed (changes requested):
8 already-merged VATEX tasks already draw from the **same underlying 1000-row pool, in the same order**, for every direction — they only differ in which columns are picked as query vs. corpus. The "pool changes across tasks" claim doesn't match the code: the pool was already identical/shared across those directions before this PR.

So for `VATEXModalEquivV2TRetrieval`, `T2VRetrieval`, `V2ARetrieval`, and `A2VRetrieval`, the only actual difference from the existing merged tasks is restricting to the first 150 of the same 1000 shared rows — that's a strict subset with less statistical power, not a new capability. The "compare scores to isolate modality difficulty" analysis this PR is meant to enable was already possible today using the full-pool `VATEXV2TRetrieval`/`VATEXV2ARetrieval`/etc.

The genuinely new contribution here is `A2T` and `T2A`

Rakshitha-Ireddi:
Thanks for the detailed feedback — you are right that V2T, T2V, V2A, and A2V are already covered by the merged pool. I can open a new minimal PR with only the two genuinely new directions: A2T (audio → text) and T2A (text → audio), following the same pattern as the existing VATEX tasks and using the full 1k pool. Does that work?

PR #5496, opened 2026-09-20
Add RuLawRetrieval (Russian legal article retrieval)

alekseevpavel04:
## Add RuLawRetrieval (Russian legal article retrieval)

Retrieve the article of law answering a question, over 3,786 articles of 6 Russian codes (Labour,
Civil, Housing, Administrative Offences, Family, Consumer Rights Protection Law).

**Gap.** The legal domain is well covered in `mteb` — in every major language except Russian:

```python
legal = [t.metadata for t in mteb.get_tasks()
         if t.metadata.type == "Retrieval" and "Legal" in (t.metadata.domains or [])]
len(legal)                                             # 41, across 10 languages
[m.name for m in legal if "rus" in str(m.eval_langs)]  # []
```

Same shape as `AILAStatutes`, `GerDaLIR`, `LawIRKo`, `ZacLegalTextRetrieval`.

**Data.** Law texts are public domain (Civil Code art. 1259 p. 6). 728 test queries in three styles
(everyday / short search / lawyer), generated by YandexGPT-5-Lite-8B, verified by a judge of another
family (Qwen3-8B, rejected 12 of 1,039); queries naming the article number or copying 6+ consecutive
words are filtered. Card and licences: https://huggingface.co/datasets/alekseevpavel04/ru-law-retrieval

**Checklist**
- [x] Outlined why this fills an existing gap in `mteb`
- [x] Runs with `mteb.evaluate` (2.21); `test_metadata.py`, `test_get_tasks.py` pass
- [x] Models run: random encoder 0.0010, `multilingual-e5-small` 0.7564,
      `multilingual-e5-small-ru-law` 0.8226 — neither trivial nor random
- [x] Size: 3,786 documents, 728 queries; descriptive statistics added, no duplicates

**Reference ndcg_at_10** (512-token truncation): FRIDA 0.861, bge-m3 0.847, e5-large-instruct 0.837,
USER-bge-m3 0.837, e5-large 0.828, e5-small 0.757, BM25 0.673.

**Limitations** (in the card): queries are LM-generated — 7 text retrieval tasks already use the
same `sample_creation`; one relevant article per query, a re-annotated 174-query subset shows ~21%
have a second; single edition of the laws; `license="multiple"` — happy to change if another value
fits better.


alekseevpavel04:
Shortened the description to 1,968 characters to fit the 2,000-character limit — apologies for the initial version. The `long PR desc` label is from that first revision.

Samoed (changes requested):
Did you check quality of this dataset? I'm not sure that we want to integate task that done by claude without publication

alekseevpavel04:
Fair question.

I designed and ran this project end to end, I wrote the annotation guide the pass followed (docs/GOLDEN_GUIDE.md), and I've read the annotation output myself, including every question the agent rewrote or dropped. The agent did the labelling at scale, I reviewed what it did. I put that in the card because leaving it out would have been worse.

The corpus isn't generated. 3,786 articles of six Russian codes, verbatim public domain text from legalacts.ru, parsed with regression tests.

The queries are LM-generated, and that's in the metadata. Train and eval use different model families on purpose (Qwen3-8B for train, YandexGPT-5-Lite-8B for dev/test), so eval questions don't share the style of the training ones. Every eval query was checked by a judge from a different family than the one that wrote it, which rejected 12 of 1,039. Queries that name the article number or copy 6+ words in a row from the article are filtered out.

On quality, the 180-query golden subset is the part to look at. Each query was re-checked against its gold article plus 5 candidates from BM25 and FRIDA. Raw counts, so nothing hinges on a percentage: 180 annotated, 15 questions rewritten, 6 removed (4 the annotator flagged as bad questions, 2 where the gold article was judged wrong and the candidate pool had nothing better), leaving 174 in the split. Of those 174 the gold article is relevant in 173, and 37 have a second relevant article that the one-positive automatic labelling missed. That 37 is where the "21% have more than one answer" line in the card comes from, and it means absolute scores here are understated the way they are in any one-positive retrieval set.

I've pushed the per-item record to the repo, https://github.com/alekseevpavel04/ru-law-retrieval/blob/main/data/golden/annotations.jsonl, so you can read the verdict and the reasoning for each of the 180 instead of taking my word for it. The qids join to the golden split of the dataset. Two caveats on that file: the ts field is the time the batch was written, so it's identical on every row and says nothing about per-item timing, and share_gold_relevant = 0.983 in golden_stats.json uses a 176 denominator (180 minus the 4 the annotator dropped), which I've now spelled out in the card instead of leaving it implied.

Discrimination check: random encoder 0.0010, BM25 0.673, multilingual-e5-small 0.757, e5-large 0.828, FRIDA 0.861.

If the bar is evidence of human review, I think that record clears it. If the bar is that an LM-generated task needs an accompanying paper, tell me straight and I'll close the PR rather than take up your time.

PR #5497, opened 2026-09-20
fix: Simplify category `ModelMeta`

Samoed:
Currently each modificaiton of categories reqire to update `TaskCategory` literal. I added `field_validator` that would check based on modalities automatically


mteb/abstasks/task_metadata.py
```diff
@@ -496,7 +396,7 @@ class TaskMetadata(BaseModel):
     prompt: str | PromptDict | None = None
     type: TaskType
     modalities: list[Modalities] = ["text"]
-    category: TaskCategory | None = None
+    category: str | None = None
```
KennethEnevoldsen, line 399: Hmm right now it is quite hard to know how this should be specified without the validator

KennethEnevoldsen (commented):
hmm I get that this is frustrating when adding new modalities, but I think we are mostly done on that front? I am not sure this is solves a problem (let me know if I am wrong)

Samoed:
Yes, this might be a bit confusing. I just wanted to simplify a bit

PR #5498, opened 2026-09-20
fix: return the prompt itself from AbsEncoder.get_prompt

RizgarOzan:
## What
`AbsEncoder.get_prompt` delegated to `get_prompt_name`, so it returned the prompt *key* instead of the prompt *text*. It now calls the module-level `get_prompt` that already lives in the same file.

## Why
The method's only caller is `CDEWrapper.encode` (`cde_models.py:98`), and the value goes straight into `self.model.encode(prompt=prompt)`. So `jxm/cde-small-v1` and `jxm/cde-small-v2` prefixed every input with `"query"` / `"document"` / `"Classification"` instead of `"search_query: "` / `"search_document: "` / `"classification: "` from `cde_model_prompts`. The other two call sites (`pylate_models.py:332`, `openai_wrappers.py:1255`) already use the module-level function and are unaffected.

Introduced in #4151, which extracted the module-level helpers — `b0e5f5a0^` still had the correct body. Same bug class as #1665.

## How
`return get_prompt(...)` instead of `return get_prompt_name(...)`, plus a regression test next to the other prompt tests.

## How I tested
Windows 11, CPython 3.14.3.

- Repro on `main` (an `AbsEncoder` subclass with `model_prompts=cde_model_prompts`): `get_prompt` returned `'query'`, `'document'`, `'Classification'`; with the fix `'search_query: '`, `'search_document: '`, `'classification: '`.
- New `test_get_prompt_returns_the_prompt_not_its_name`: **2 failed** on `main`, passes with the fix. No marker, so `make test` picks it up. `tests/test_integrations/test_prompts.py` → 74 passed.
- `ruff check`, `ruff format --check`, `typos` clean; `mypy mteb` → no issues in 1741 files.
- Base `main` (`5d3212e4`) is green except the pre-existing `test (windows-latest, 3.10)` charmap failure.

I did not run the CDE models end to end (needs the model download plus the dataset-embedding pass), so I can't quote a score delta; any cached CDE results were produced with the wrong prefix.


Samoed:
This will break compatability and this method is not used a lot

RizgarOzan:
Understood — thanks for looking. I'll leave the `AbsEncoder.get_prompt` signature alone.

For context on how I ran into it: the single call site is `CDEWrapper.encode` (`mteb/models/model_implementations/cde_models.py:98`), and it passes the result straight into `self.model.encode(..., prompt=prompt)`, where `prompt=` is literal text for SentenceTransformers. Since `get_prompt` returns the *key* of `model_prompts`, `jxm/cde-small-v1` / `v2` currently prepend `"query"` / `"classification"` instead of the configured `"search_query: "` / `"classification: "` (`cde_models.py:204-215`).

`sentence_transformer_wrapper.py:136` already has `_resolve_prompt()`, which does the name -> text lookup (and the logging) that the CDE path is missing.

Would a fix confined to `cde_models.py` — using that existing helper, no change to `AbsEncoder` — be welcome? Happy to open it if so.

PR #5502, opened 2026-09-21
fix: remaining changes before making import mteb torch-free and mteb-core compatible

ayush1298:
Refs #5063. Follows #5463, #5329, #5350, #5426, #5436, #5441

This PR contains the last changes needed before splitting `mteb` into `mteb-core + mteb`

Changes: 

1. Distribution lookup: the version, extras check and requirements check now read from whichever of `mteb-core` or `mteb` is installed `(mteb-core` first), via `_mteb_distribution().` Currently import mteb fails with `PackageNotFoundError` when only `mteb-core` is installed.
2. CLI: moved `import torch` into `run()` / `_mock_run(),` so `mteb available-tasks` etc. work without `torch.`
3. Image evaluator: with `num_proc > 1`, `ImageTextPairClassificationEvaluator` crashed on macOS/Windows (Can't get local object) because spawn can't pickle a class or function defined inside a function. The collate fn is now module-level, and the local `CustomImageDataset` defines `__reduce__,` so it is pickled as a call to the module-level `_build_image_dataset(images).`
4. Seeding: aggregated tasks create tasks at `import,` and `_set_seed` imported `torch` (and `tensorflow)` to seed them, so every import mteb loaded `torch.` Now `torch/tensorflow` are seeded only if already imported. Evaluators seed again when they're created, after the model is loaded, so evaluation results are unchanged.

Result: `import mteb` 2.29s → 1.76s, 5058 → 4332 modules, torch no longer imported.

Tests: import mteb doesn't import torch, the image dataset works with spawn workers, the distribution lookup prefers mteb-core, and the CLI imports without torch.

Next steps will be to split the package. `mteb-core` without `torch,` `transformers` and `sentence-transformers,` and `mteb` as a metapackage depending on `mteb-core[run].` I will report the full results: install size, install time and import mteb time for mteb-core vs mteb.


mteb/_evaluators/image/imagetext_pairclassification_evaluator.py
```diff
@@ -25,38 +26,32 @@
 logger = logging.getLogger(__name__)
 
 
-def _build_image_dataset(
-    images: list[Image],
-) -> torch.utils.data.Dataset[dict[str, Any]]:
-    """Wrap `images` in a torch map-style dataset.
+class _ImageDataset:
```
Samoed: I think we can simply keep `torch.Dataset`

ayush1298: I think there will be a problem. Because subclassing `torch.utils.data.Dataset` needs torch when the class is defined. At the module level, that brings torch back into the `import mteb`, and inside a function it can't be pickled into spawn workers (the current crash with num_proc > 1).

Samoed: You can keep it inside function as was previously

ayush1298: done

isaac-chung (approved):

Issue #4764, opened 2026-06-03
Quality issues in MTEB(Code, v1)


Geralt-Targaryen:
Recently I did a little research on the quality of MTEB(Code, v1) test set, and here are some findings I'd like to share:

| | # queries | # unique queries | unique query % | Quality Score Distribution | Low-quality % |
|-|-|-|-|-|-|
|CosQA | 500 | 500 | 100% | {4: 251, 2: 116, 3: 94, 1: 39} | 31% |
|SyntheticText2SQL | 5,851 | 5,851 | 100% | {4: 5068, 3: 454, 2: 296, 1: 33} | 6% |
|CodeFeedbackST | 31,306 | 30,749 | 98% | {4: 22229, 3: 5393, 2: 2794, 1: 886} | 12% |
|CodeFeedbackMT | 13,277 | 13,277 | 100% | {4: 9256, 3: 2360, 2: 1136, 1: 523} | 12% |
|StackOverflowQA | 1,994 | 1,994 | 100% | {4: 1574, 3: 315, 2: 91, 1: 14} | 5% |
|AppsRetrieval | 3,765 | 3,765 | 100% | {4: 3277, 3: 214, 2: 193, 1: 45} | 6% |
|CodeEditSearchRetrieval | 21,495 | 21,253 | 99% | {4: 20530, 3: 491, 2: 308, 1: 166} | 2% |
|CodeSearchNetCCRetrieval | 52,561 | 52,439 | 99% | {4: 50810, 3: 906, 2: 670, 1: 174} | 2% |
|CodeSearchNetRetrieval | 6,000 | 5,766 | 96% | {4: 5446, 3: 270, 2: 145, 1: 137} | 5% |
|CodeTransOceanContest | 221 | 221 | 100% | {4: 144, 3: 60, 2: 11, 1: 6} | 8% |
|CodeTransOceanDL | 180 | 50 | 28% | {4: 131, 3: 49} | 0% |
|COIRCodeSearchNetRetrieval | 52,561 | 52,561 | 100% | {4: 38648, 3: 9761, 2: 3468, 1: 683} | 8% |

The quality score is annotated by Qwen3-32B (prompt given below). Samples receiving a score of 1 or 2 are counted as low-quality here.

The two most notable problems are:

[...]

Issue #5508, opened 2026-09-21
Reference Model Coverage Tests fail: missing MTEB(slk, v1) results for reference models


isaac-chung:
## What's happening

The \`Reference Model Coverage Tests\` CI workflow (\`.github/workflows/reference_models.yml\`) has been failing on every recent run, across unrelated PRs and branches. It is not caused by any specific PR.

## Why

\`tests/test_benchmarks/test_reference_models.py\` downloads the published remote result cache and checks that each reference model has results for every public task in every benchmark. \`MTEB(slk, v1)\` (the Slovak benchmark) was added in #4788 (merged 2026-08-23), but the reference models have not been run against it yet, so the remote result cache has no scores for these tasks.

Latest failure: https://github.com/embeddings-benchmark/mteb/actions/runs/35610223163/job/106367182321

```
FAILED tests/test_benchmarks/test_reference_models.py::test_reference_model_coverage[minishlab/potion-multilingual-128M-MTEB(slk, v1)] - Failed: 'minishlab/potion-multilingual-128M' missing 21/31 tasks in 'MTEB(slk, v1)'
FAILED tests/test_benchmarks/test_reference_models.py::test_reference_model_coverage[mteb/baseline-bm25s-MTEB(slk, v1)] - Failed: 'mteb/baseline-bm25s' missing 6/8 tasks in 'MTEB(slk, v1)'
FAILED tests/test_benchmarks/test_reference_models.py::test_reference_model_coverage[codefuse-ai/F2LLM-v2-80M-MTEB(slk, v1)] - Failed: 'codefuse-ai/F2LLM-v2-80M' missing 24/31 tasks in 'MTEB(slk, v1)'
```

The last successful run of this workflow on \`main\` was 2026-07-26, before the Slovak benchmark existed.

## Fix

Run the reference models (\`intfloat/multilingual-e5-small\`, \`sentence-transformers/static-similarity-mrl-multilingual-v1\`, \`minishlab/potion-multilingual-128M\`, \`mteb/baseline-bm25s\`, \`codefuse-ai/F2LLM-v2-80M\`) on the \`MTEB(slk, v1)\` tasks and publish the results to the remote result cache.

RizgarOzan:
I'm running the three models that are still missing here (`potion-multilingual-128M`, `baseline-bm25s`, `F2LLM-v2-80M`; e5-small and static-similarity already cover all 31 tasks) and will open a PR to `results`, same shape as results#720 for Italian.

[...]

PR #5511, opened 2026-09-22
benchmark: add MTEB(Code, v1.1)

HiveYuan:
## Summary

- preserve `MTEB(Code, v1)` and mark it as superseded by `MTEB(Code, v1.1)`
- add `MTEB(Code, v1.1)` with only `CodeTransOceanDL` removed
- point the leaderboard menu to v1.1 while keeping the existing `MTEB(code)` alias on v1
- add regression coverage for the task delta, supersession, alias, and leaderboard visibility

## Rationale

`CodeTransOceanDL` contains duplicate query and corpus texts with conflicting single-positive relevance labels. In the test split, 180 query rows collapse to 50 unique texts, while 816 corpus rows collapse to 266 unique texts. Since identical queries cannot produce different rankings, this creates an artificial score ceiling.

This implements the scoped quick fix proposed in #4764. It does not change the underlying task or dataset, and it leaves CoIR, RTEB, CosQA, and the legacy alias unchanged.

## Validation

- `uv run --no-sync python -m pytest tests/test_benchmarks/test_get_benchmarks.py -q` — 8 passed
- `uv run --no-sync python -m pytest tests/test_benchmarks -q -m "not test_reference_models"` — 27 passed, 390 deselected
- full `tests/test_benchmarks` run — 414 passed; 3 unrelated failures in `MTEB(slk, v1)` reference-model coverage
- `make lint-check`
- `uv run --no-sync python -m mypy mteb`
- `uv run --no-sync python scripts/format_citations.py benchmarks --error-on-change`

Refs #4764


HiveYuan:
@KennethEnevoldsen @Samoed, this is ready for review.

The only failing check is `reference-models`. I checked the full log:

- all five reference-model checks for `MTEB(Code, v1.1)` pass;
- the three failures are all for `MTEB(slk, v1)`;
- the same three Slovak failures are present on `main`: https://github.com/embeddings-benchmark/mteb/actions/runs/33060948254

All other PR checks pass, including the full Python 3.10–3.14 and Windows test matrix. Could you confirm whether the `reference-models` failure is an existing CI/result-cache issue and can be treated as unrelated to this PR?

tests/test_benchmarks/test_get_benchmarks.py
```diff
@@ -45,3 +45,22 @@ def test_benchmark_on_leaderboard():
...
+    assert v1_tasks - v1_1_tasks == {"CodeTransOceanDL"}
+    assert not v1_1_tasks - v1_tasks
+    assert v1.superseded_by == [v1_1.name]
+    assert mteb.get_benchmark("MTEB(code)").name == v1.name
+
+    leaderboard_names = {
+        benchmark.name for benchmark in mteb.get_benchmarks(display_on_leaderboard=True)
+    }
+    assert v1.name not in leaderboard_names
+    assert v1_1.name in leaderboard_names
```
KennethEnevoldsen, line 66:
```suggestion
```

KennethEnevoldsen (approved):
I think this looks good!

I practice it seems like we should add the reference models when updating the benchmark (should be very doable they are all small). The only reason MTEB(Code, v1) doesn't have coverage is because it is from before they were required.

but do feel free to add the exception to the test for v1.1 and add an issue on adding reference models to mteb(code, v1.1) and then I think we can merge.

isaac-chung:
FYI There's an existing issue tracked here for the reference models test: https://github.com/embeddings-benchmark/mteb/issues/5508

Issue #5504, opened 2026-09-21
Multilabel classification: the 2000-row test cap never applies (stratify_by_column fails on List labels)


RizgarOzan:
`AbsTaskMultilabelClassification._evaluate_subset` tries to cut the test set to 2000 rows (`multilabel_classification.py:156-165`):

```python
split_dataset = eval_split.train_test_split(
    test_size=2000, seed=42, stratify_by_column="label"
)
```

For multilabel data `label` is a `List` column, and `datasets` only stratifies on `ClassLabel`, so this raises every time:

```
ValueError: Stratifying by column is only supported for ClassLabel column, and column label is List.
```

The `except ValueError` then logs "Couldn't subsample, continuing with the entire test set." In practice the 2000 cap never applies. (It also uses a hard-coded `"label"` instead of `self.label_column_name`.)

Which tasks this touches: the tasks' own `descriptive_stats` list 10 eval splits above 2000 rows. For multilingual tasks that number is summed over subsets. MultiEURLEX (115000), EmotionAnalysisPlus (43882), SwedishPatentCPCGroup/Subclass (8575 / 8011, `train` as eval split), VABBMultiLabel (3235), SciRepEvalFoS (2458), MalteseNews (2297), and BrazilianToxicTweets / SensitiveTopics / KorHateSpeechML at 2048 / 2048 / 2037 (just over the cap).

I see two ways to fix this, and they have different consequences, so I'd like to check which one you want before opening a PR:

1. **Make the cap work** by calling the existing `_multilabel_subsampling` (`abstask.py:50`) with `n_samples=2000`. Scores for the tasks above would change, and runs would be cheaper.
2. **Drop the eval-time cap** and the warning. The full eval split is what gets scored today anyway, so scores stay the same, and what the code does matches what it says.

I can open a PR for either one, with a test.

isaac-chung:
I believe we want the cap to work.

Would be great if you could also:
1. Confirm if there are existing results for these tasks
2. If 1) is true, we should then update the affected tasks to v2

RizgarOzan:
Thanks. I checked, and yes, most of them have results. First a correction to my count above: the cap works per subset, and counting that way it misses on 16 tasks, not 10. I had left out the audio and image ones.

Result files in `embeddings-benchmark/results` (GitHub code search on the file name, so treat the 0s as "none found", not a proof):

| task | eval rows over 2000 | result files | benchmarks |
|---|---|---|---|
[...]

PR #5514, opened 2026-09-22
fix: make the multilabel classification evaluation cap work

RizgarOzan:
Part of #5504.

The 2000-row cap on multilabel eval splits never ran: `train_test_split(stratify_by_column="label")` raises on a list column and the bare `except ValueError` swallowed it, so every task scored its full split.

What changed:
- The cap uses the existing `_multilabel_subsampling` with `self.label_column_name`, through a new class attribute `max_eval_samples` (default 2000, `None` = whole split).
- `_multilabel_subsampling` uses `select` instead of `Dataset.from_dict(ds[idx])`, so image and audio features survive.
- 16 tasks go over 2000 rows. The 12 with results in `embeddings-benchmark/results` keep `max_eval_samples = None`, so their scores don't change, and each gets a `<Name>.v2` with the cap on: same dataset and revision, v1 points to it with `superseded_by`. Benchmarks still use v1. The other 4 (SciRepEvalFoS, SwedishPatentCPCGroup/Subclass, AudioSet) have no results and just get the cap.
- The cap only applies at evaluation, so v2 descriptive stats are the same as v1. The v2 names are added next to their v1 entries in the quality test's `KNOWN_ISSUES`.

How I tested:
- New test in `tests/test_abstasks/test_multilabel_classification.py`: the two capped cases fail on main, all 4 pass here.
- all-MiniLM-L6-v2, VABBMultiLabelClassification: v1 scores 3235 rows (accuracy 0.3393), v2 2011 rows (0.3357).
- v2 stats recomputed for 5 text tasks, same values as v1. The other 7 are copied from v1 (MultiEURLEX is 4 GB; the audio ones need torchcodec, which doesn't load here).
- `ruff check`, `ruff format --check`, `mypy mteb` clean. `pytest` on test_abstasks, test_tasks, test_evaluators, test_benchmarks, test_get_tasks, test_filter_tasks with the Makefile's `-m` filter: 2471 passed, 30 failed, the same 30 as main (audio/video mock stats, torchcodec DLL).


Samoed (commented):
Can you create v2 versions of tasks?

RizgarOzan:
Yes, added in 96a08d57. Each of the 12 tasks with results now has a `<Name>.v2`: it subclasses v1, copies the metadata with a new name and `adapted_from`, and turns the cap on. It's the same dataset and revision (like `FleursA2TRetrieval.v2`), so nothing new to upload. v1 keeps scoring the full split and gets `superseded_by`; benchmarks still point to v1.

On VABBMultiLabelClassification with all-MiniLM-L6-v2: v1 scores 3235 rows (accuracy 0.3393), v2 2011 rows (0.3357).

One thing to flag: seven of these are only a little over 2000 (KorHate 2037, BrazilianToxicTweets, SensitiveTopics and FSD50K 2048, BirdSet 2050, AudioSetMini 2153, MalteseNews 2297), so their v2 is almost the same as v1. I kept them to follow the rule in #5504, but I can drop those seven if you'd rather not have the extra tasks.

isaac-chung:
> One thing to flag: seven of these are only a little over 2000 (KorHate 2037, BrazilianToxicTweets, SensitiveTopics and FSD50K 2048, BirdSet 2050, AudioSetMini 2153, MalteseNews 2297), so their v2 is almost the same as v1. I kept them to follow the rule in #5504, but I can drop those seven if you'd rather not have the extra tasks.


That's also a lot to rerun, so maybe we dont bump these 7 tasks somehow e.g. raise the cap to 3000.

RizgarOzan:
Thanks, done in 2e279038. The default cap is now 3000, so those seven tasks (all at or under 2297 rows) are never cut and no longer get a v2. Five v2 tasks remain: MultiEURLEX, EmotionAnalysisPlus, FSD2019Kaggle, VOC2007 and VABBMultiLabelClassification. VABB is 3235 rows, so its v2 is close to v1 as well; happy to drop it too if you prefer. SciRepEvalFoS (2458) also falls under the new cap.

PR #5516, opened 2026-09-22
feat: Add optional float relevance gains for retrieval (ndcg_float_at_k)

fabianschmidt-cohere:
## What

With this PR, retrieval tasks can optionally ship a `gains` HF config (`query-id`, `corpus-id`, `gain`, float64) next to `qrels`. When present, `ndcg_float_at_{k}` and its nAUC variants are reported alongside the usual integer-qrels metrics. It is discovered by name exactly like the existing optional `top_ranked` and `instruction` configs, so datasets without one are untouched.

## Why

`_load_qrels` casts scores to `int32` and pytrec_eval only accepts integer labels, so a dataset with continuous relevance annotations (LLM-judged, IRT-calibrated, pooled human means -- more to come very soon 🙂) has no way to be evaluated against them. @Samoed hit this issue in #3479:

> Overall, I think this should be float, but `pytrec_eval` changed scores from float to int in recent release. I left as int32 for compatibility[...]

Query filtering is left alone, so the float metric covers the same queries as the integer ones and published `ndcg_at_k` stays comparable; this only adds keys.

## Ties

Crucially, this PR deviates from #5136 / #5241, which broke MRR and p-MRR ties by doc id. Each tied document here is credited its tie group's mean gain, representing `E[NDCG]` over the tie's orderings. Doc-id ordering would incorrectly score the corpus's naming scheme rather than the system. Those PRs were also a different case: they were second computations of quantities pytrec_eval also produces, so they had to agree with it; `ndcg_float_at_k` has no counterpart to disagree with. Happy to share the supporting numbers if useful.

## How I tested

15 new tests across two files; `tests/test_evaluators/ tests/test_abstasks/` goes from 200 passed / 46 skipped on `main` to 215 / 46 here. `make lint` clean; `make typecheck` reports no new errors (identical error set to `main`).

Written with AI assistance (Opus 5 xhigh via Cursor); I've reviewed every line. The tie semantics were checked against internal implementations and test cases.


Samoed:
I'm not sure that we want to duplicate metrics. And I also don't think that we have tasks with float qrles

fabianschmidt-cohere:
> And I also don't think that we have tasks with float qrles

A forthcoming preprint of ours will introduce RCP-nDCG that comes with fine-grained cross-query calibrated document relevance probabilities (floats). We have already discussed our work with @KennethEnevoldsen who is very happy to have RCP-nDCG in MTEB.

> I'm not sure that we want to duplicate metrics. 

Can you please clarify? pytrec_eval does not support qrels as floats. The other question is whether MTEB would prefer `-gains` or floating qrels in the HF dataset spec.

docs/contributing/adding_a_dataset.md
```diff
@@ -227,6 +227,31 @@ Once we have decided on task, we can implement them as follows:
...
+    (float) — uploaded alongside `queries`/`corpus`/`qrels`. Config name: `gains` for the default subset,
+    `{subset}-gains` for multilingual subsets (e.g. `en-gains`), matching the `{subset}-qrels` convention.
+    Gains are used directly as NDCG gains (identity, non-negative); apply any transform (e.g. a sigmoid)
+    at dataset creation time. Descriptive statistics do not yet cover gains.
+
+    ```python
+    class MyGainsTask(AbsTaskRetrieval):
+        metadata = mteb.TaskMetadata(
+            # ...
+            main_score="ndcg_float_at_10",  # NDCG over the float gains
```
KennethEnevoldsen: shouldn't this instead refer to specifcally the `RCP-nDCG` or is the RCP- referring more to the method of generatin (the preprint i saw didn't use the new term)

docs/contributing/adding_a_dataset.md
```diff
@@ -227,6 +227,31 @@ Once we have decided on task, we can implement them as follows:
     # 0.021194685839832323
     ```
 
+    #### Retrieval tasks with continuous relevance gains
```
KennethEnevoldsen: Overall I would probably suggest adding a blog post on this to start with (#5521) and keep it outside the main docs until it has had some time to prove itself. That is also a way for us to argue why we add the metric and why we add it.

docs/whats_new.md
```diff
@@ -7,6 +7,22 @@ icon: lucide/newspaper
 
 This section is an overview of releases for more information check out the autogenerated [changelog](https://github.com/embeddings-benchmark/mteb/blob/main/pyproject.toml).
 
+## New in v2.22
+
+### Continuous relevance gains for retrieval
```
KennethEnevoldsen: I would probably ship this with at least one actual task

mteb/types/_encoder_io.py
```diff
@@ -277,6 +277,10 @@ def get_dtype(self) -> torch.dtype:
 RelevantDocumentsType = Mapping[str, Mapping[str, int]]
 """Relevant documents for each query, mapping query IDs to a mapping of document IDs and their relevance
 scores. Should have columns `query-id`, `corpus-id`, `score`."""
+GainsType = Mapping[str, Mapping[str, float]]
+"""Continuous relevance gains for each query, mapping query IDs to a mapping of document IDs and their
+gain values. Should have columns `query-id`, `corpus-id`, `gain`. Gains are used directly as NDCG gains
+(identity, non-negative); apply any transform (e.g. sigmoid) at dataset creation time."""
```
KennethEnevoldsen: I am not sure what is meant by:

> apply any transform (e.g. sigmoid) at dataset creation time.

Should I do that?

tests/test_abstasks/test_retrieval_float_gains.py
KennethEnevoldsen: In general for the tests I would implement it as a mock tasks:

https://github.com/embeddings-benchmark/mteb/blob/main/mteb/mocks/mock_tasks/retrieval.py

That should already get it through many of our existing tests and allow you to simplify here

KennethEnevoldsen: You can add 1 task with different splits for the different things you test below (so that we don't need multiple mock tasks)

mteb/_evaluators/retrieval_metrics.py
KennethEnevoldsen: Would love to see a match up against the scores in the paper if that is possible?

mteb/_evaluators/retrieval_metrics.py
```diff
@@ -55,6 +57,97 @@ def mrr(
...
 
 
+def _dcg(gains: Sequence[float], k: int) -> float:
+    """DCG@k = sum_{r=1}^{k} gain(r) / log2(r + 1)."""
+    return sum(
+        gain / math.log2(rank + 1) for rank, gain in enumerate(gains[:k], start=1)
+    )
+
+
+def ndcg_float_scores(
```
KennethEnevoldsen: Would love to get an idea about the speed here - some might wish to scale it to larger datasets - e.g. compare it using smaller static model (or mteb random baseline models) and see the speed diff.

KennethEnevoldsen (commented):
So mostly a general review here to figure out the direction of the fix

So we can def. add this in "stealth" first if that is needed, but I think the documentation on it right now is to vague for users to actually use it.

I would encourage the blog-post route + datasets, let me know if this is possible? If we want to keep it over multiple PRs then I would suggest not adding any noteworthy changes to the docs in this PR.

In general throughout I would consider adding references

mteb/types/_encoder_io.py
Samoed: Can we keep this in qrels? I don't think that we need to add a new split for that

